PR #3, opened 2017-03-19
Update README.md

jmatsushita:
For some reason the parenthesis are needed which is consistent to the syntax in the vending machine example.


davidkpiano:
Thanks! I'll merge this. Just FYI, I'm currently working on developing a simpler, less ad-hoc version of this library. If you know what might be useful, in terms of features etc., please let me know!

PR #14, opened 2017-09-24
Types

zigomir:
Main change here is to add `types` property into `package.json` file as [described here](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html).

What I also find strange though, is that currently published npm package has different `lib/index.d.ts` from the one I get when running `npm run build` locally. Maybe new typescript version? The one that is generated locally actually works well while one on npmjs.org doesn't.

Sorry about updating packages, but I couldn't get `npm run build` working after `npm install` on current master. 

I was always getting this from webpack:

```
module.js:529
    throw err;
    ^

Error: Cannot find module 'locate-path'
```

I'm aware I didn't update `yarn.lock` which probably should be updated too. Fell free to just take `"types": "lib/index.d.ts",` change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,59 +1,117 @@
 {
   "name": "xstate",
-  "version": "0.0.1",
+  "version": "1.2.1",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@types/chai": {
-      "version": "3.5.2",
-      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-3.5.2.tgz",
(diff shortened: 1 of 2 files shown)


davidkpiano:
Makes sense, merged! I'm considering moving (back) to `Rollup` using this library template: https://github.com/alexjoverm/typescript-library-starter for v2, let me know what you think or if you have any suggestions about that.

zigomir:
I unfortunately don't have any experience regarding that. But I remember reading about that rollup might be better for libraries and webpack for apps. [(Found it)](https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c).

PR #17, opened 2017-11-04
Throw a more helpful error when transitioning to an undefined state

akonwi:
Thanks for the great library. This has been immensely helpful for what I'm working on right now.

The only pain point I've had in about a week of using this is that when transitioning to a state that is not defined:

``` javascript
states: {
  a: {
    on: {
       continue: 'b'
    }
  // 'b' should be declared here but I forgot
}
```

the error logged seems completely unrelated to the actual problem at first glance.

screen shot 2017-11-04 at 2 43 07 pm

When I read "cannot read property 'initial'", I thought my chart provided to the `Machine()` function was invalid somehow. This was confusing because I transitioned to 3 other states before seeing this error and spent 2 hours thinking I was crazy and the stacktrace was utterly useless.

Anyway, I hope this is a good start to the solution so others don't run into this problem.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -165,6 +165,10 @@ class StateNode {
       currentHistory = currentHistory[subPath] as IHistory;
     });
 
+    if (currentState === undefined) {
+      throw Error(`Action '${action}' on state '${history.$current}' leads to undefined state '${nextPath}'.`)
+    }
+
     while (currentState.initial) {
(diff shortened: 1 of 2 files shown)


davidkpiano:
Awesome, thank you! I'm working on providing more helpful error messages and stricter typing help, so this is great. I appreciate it!

PR #20, opened 2017-11-12
Include Prettier configs and run with tests

### LINKED ISSUE #18: Prettier configs
It may be a good idea to include a prettier config file or put it in the package.json. I use atom and prettier was changing the files but I didn't know which settings to use so I turned it off.

akonwi:
There was a failure while running tests before I made any changes. I'm not quite sure what to do about it. 

Here is the message: 

``` bash
TSError: ⨯ Unable to compile TypeScript
test/graph.ts (177,9): Object is possibly 'undefined'. (2532)
```

Fixes #18

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,9 @@
+{
+  printWidth: 80,
+  tabWidth: 2,
+  useTabs: false,
+  semi: true,
+  singleQuote: false,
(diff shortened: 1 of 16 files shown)


davidkpiano:
Weird, I'm not getting that error. Do you have latest of everything installed?

PR #56, opened 2018-02-08
Activities

davidkpiano:
This PR adds support for activities.

Main file to look at is `activities.test.ts` to ensure that the behavior is correct. cc. @mogsie

### DIFF
--- a/test/activities.test.ts
+++ b/test/activities.test.ts
@@ -1,5 +1,6 @@
 import { assert } from 'chai';
 import { Machine } from '../src/index';
+import { start, stop } from '../src/actions';
 
 const lightMachine = Machine({
   key: 'light',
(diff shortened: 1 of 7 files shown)


test/activities.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        stop: {}
+      }
+    }
+  }
+});
+
+describe('activities', () => {
+  it('identifies start activities', () => {
+    assert.deepEqual(lightMachine.transition('yellow', 'TIMER').activities, {
+      activateCrosswalkLight: true
```
mogsie: Should `blinkCrosswalkLight` be false here?  Or do they only appear in the map if they've ever been true?

mogsie: I was thinking that as part of the constructor, and building the State hierarchy, that it could aggregate all the known activities and initialize them all as `false` before calculating `initialState`.

davidkpiano: Hmm... depends on the use case. I assumed that the developer would use the `.activities` object as a way of starting and stopping activities, e.g.:

```js
Object.keys(state.activities).forEach(activity => {
  if (state.activities[activity]) {
    start(activity); // yes I know this is imperative and terrible
  } else {
    stopIfRunning(activity);
  }
});
```

For `stopIfRunning(activity);`, if the activity wasn't running in the first place, this is an inert function call - one that we can avoid by just not having the activity show up until it is "born".

Also, if the developer is just checking if a _specific_ activity is active:

```js
state.activities[activity] ? 'active' : 'not active'
```

would return the correct value regardless of if the activity is `false` or the activity does not exist.

mogsie: It's no biggie.  I'm not sure if it's important that they're `false` or simply not present; In general, a stateful counterpart will be able to know what to do by diff'ing the two activities anyway.  It's just a bit odd that the the activities contains one property after `initial` and then two properties when it's gone via `red` to `green`.  It's no biggie.  It's just a documentation issue, I guess.

PR #59, opened 2018-02-10
getShortestPaths + condition

MicheleBertoli:
The `getShortestPaths` method is currently failing when a condition is present because the `cond` function is called with `undefined`. 
This PR fixes the issue, the relevant changes are in https://github.com/davidkpiano/xstate/commit/52fb0e236f8795e222d99f624e9cd56b4912b783.


davidkpiano:
Thanks!

PR #60, opened 2018-02-20
Fix extended state on transition

Julienng:
The `extendedState` isn't pass when we call the transition on the parent.


davidkpiano:
Thank you!

PR #66, opened 2018-03-07
Fix for initialStateValue returning a value if the state shares the s…

### LINKED ISSUE #65: Error when a leaf state shares a name with a property on the String prototype
## Bug
If a leaf state has the same name as a method on the String prototype, the initial state is set as the String property.  I ran into this when naming a state "search".

## Potential fix:
Check for whether the resolvedStateValue is a string, rather than assuming this.resolvedStateValue[this.key] will return undefined.  Return undefined if so.

https://codepen.io/noahblon/pen/9f3604f6720eaa0153fdd8be5ceef440

noahblon:
…ame name as a property on the String prototype and it has no initial value.

fixes #65

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -594,7 +594,9 @@ class StateNode implements StateNodeConfig {
             this.states as Record<string, StateNode>,
             state => state.initialStateValue
           )
-        : this.resolvedStateValue[this.key]) as StateValue);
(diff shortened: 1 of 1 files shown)


davidkpiano:
Thank you, great catch!

zfrankdesign:
Nice work @noahblon !!!

Issue #71, opened 2018-03-26
Problem with local transitions


lmatteis:
If I have this statechart:

```js
{
  initial: 'Hidden',
  states: {
    Hidden: {
      on: {
        PUBLISH_FAILURE: 'Failure',
      },
    },
    Failure: {
      on: {
        CLICKED_CLOSE: 'Hidden',
      },
    },
  },
}
```

When I'm in 'Failure' and the `CLICKED_CLOSE` event is fired, I properly go into the `Hidden` state. The problem is that I want to put the `CLICKED_CLOSE` event one level higher:

```js
{
  initial: 'Hidden',
  on: {
    CLICKED_CLOSE: '.Hidden',
  },
  states: {
    Hidden: {
      on: {
        PUBLISH_FAILURE: 'Failure',
      },
    },
    Failure: {
    },
  },
}
```

When I do this, the transition doesn't happen. I even tried doing `CLICKED_CLOSE: 'Alert.Hidden',` -- where Alert is the name of the parent state.

Any idea what I'm doing wrong?

davidkpiano:
Ah yeah, this might be a bug. I'll look into it.

mogsie:
This seems to be a problem with events declared at the top level. I added a couple of test cases, one where the `on CLICKED_CLOSED` was declared at the top level, the other in an `outer` top level state; otherwise identical statecharts. Only the first one fails.  Digging to see if there's a general problem with top level states, or if it only has to do with local transitions.

mogsie:
Further testing seems to indicate that it's simply top level machine events that don't seem to work.  I moved the failing test to `machine.test.ts` instead.

mogsie:
It fails in the `scxml` branch too.

AnimaLupi:
Hello, sorry to bother you guys, but the fix is going to be backported to the incoming 3.2 release too?
And by the way, is there any approximative release date for 3.2?
Thank you!

davidkpiano:
@AnimaLupi This will be fixed into 3.2, which I plan on releasing at React Finland (April 24-26).

AnimaLupi:
@davidkpiano Thank you very much for the release info and fast reply! Nice work by the way!

PR #73, opened 2018-03-29
Top level events aren't handled, exposes bug #71

mogsie:
When a machine declares event handlers at the top level, they seem to
be ignored altogether.


davidkpiano:
Have you tested this in the SCXML branch?

mogsie:
Yes, it's also in the scxml branch.

PR #76, opened 2018-03-31
Verify and harden deep history

mogsie:
History is shallow by default (as exposed by this test) and supports deep histories with specific depths using `$history.$history.$history` syntax.   This PR adds tests to verify this behaviour.

This test also checks if a state in question _has_ substates, and ignores `$history` elements if they are present.  If you happen to hit on a $history.$history state where the first level doesn't have any substate (but e.g. another possible $history has substates) then it would previously throw 

```
Error: Cannot read '$history' from state 'history.on.second.A': missing 'initial' 
```

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -493,7 +493,9 @@ class StateNode implements StateNodeConfig {
       }
 
       if (subPath === HISTORY_KEY) {
-        if (currentHistory) {
+        if (Object.keys(currentState.states).length == 0) {
+          subPath = NULL_EVENT;
+        } else if (currentHistory) {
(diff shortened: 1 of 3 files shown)


mogsie:
There is a couple of pending tests added here:

If you have a parallel state with a couple of regions, then the history actually only looks at the first "path" and quits.  It doesn't extrapolate it into _many paths_.  I am not sure xstate at the moment allows a transition to target multiple states (in different regions of a parallel state). I think solving these pending tests require that functionality.

mogsie:
I tried these out on the scxml branch too, it's not working there either.

PR #81, opened 2018-04-02
Reject invalid state targets

mogsie:
This change makes it possible to provide multiple targets instead of a single target in a transition, for example

    event : [ { target : "foo.foo" }, {target: "bar.bar" } ]

This is useful when foo and bar are both orthogonal regions, and that foo.foo and bar.bar is a respectively states in foo and bar.

There is some verification that the targets themselves are in fact non-conflicting, i.e. that you cannot target `foo.foo` _and_ `foo.baz`, since in region foo, only one of substates `foo.foo` and `foo.baz` can be active.  An exception is thrown in those cases.

Most of this diff is actually moving some code into a foreach loop, so the diff should be inspected with whitespace ignored (use ?w=1 on github).

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -515,11 +515,13 @@ class StateNode implements StateNodeConfig {
     }
 
     const transition = this.on[eventType] as Transition;
-    let nextStateString: string | undefined;
+
+    let nextStateStrings: string[] = [];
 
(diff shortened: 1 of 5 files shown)


mogsie:
Note that this PR incorporates PR 80.

PR #83, opened 2018-04-03
Added a pending test for transitioning to a # state

mogsie:
transitioning to #foo.bar.baz and to baz should yield the same
results.  This test exposes that it doesn't: It actually enters the
states that it's already in.

### DIFF
--- a/test/actions.test.ts
+++ b/test/actions.test.ts
@@ -230,13 +230,14 @@ describe('onEntry/onExit actions', () => {
       ]);
     });
 
-    it('should ignore same-parent state actions (sparse)', () => {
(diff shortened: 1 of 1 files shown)


davidkpiano:
Hmm, by default, transitions are external, meaning that without otherwise specifying, the transition type is **external**, causing an external self-transition (and entering itself).

Can you confirm? https://www.w3.org/TR/scxml/#transition

mogsie:
But in the test, the event `TACK` and `ABSOLUTE_TACK` are both defined in the `machine.foo` state, so it doesn't matter if it's an local or external transition, because `foo` _will be exited_.  Local vs external only applies when a state defines a transition to one of its children.

PR #92, opened 2018-04-18
Add documentation for the array of state transitions mappings

karl:
This PR updates the documentation to show that you can pass an ordered array of state transition mappings to support transitioning to different states from a single event based on external state.

Based on a quick discussion with @davidkpiano at https://github.com/davidkpiano/xstate/issues/88#issuecomment-380880232

### DIFF
--- a/docs/guides/guards.md
+++ b/docs/guides/guards.md
@@ -111,7 +111,7 @@ If you want to have a single event transition to different states in certain sit
 For example you can model a door that listens for an `OPEN` event, and opens if you are an admin and error if you are not:
 
 ```js
-var machine = new Machine({
+const machine = new Machine({
   id: 'door',
   initial: 'closed',
   states: {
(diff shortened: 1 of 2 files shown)


docs/guides/guards.md
```diff
@@ -106,6 +106,50 @@ console.log(searchAttempt3.actions);
...
+    },
+    opened: {
+      on: {
+        CLOSE: 'closed',
+      }
+    },
+  }
+});
+
+var fullState = { isAdmin: true };
```
davidkpiano: Hmm, can you change these `var`s to `const` and give each state a unique name? Otherwise this looks great!

davidkpiano:
```js
const firstState = ...
const secondState = ...
const thirdState = ...
```

Just to discourage mutation.

karl: Ah, I get you!

I'll fix it up this evening when I get a moment.

karl: Done!

PR #95, opened 2018-04-27
Leverage fact that enumerable: false is the default

Andarist:


davidkpiano:
Oh, good to know. Thanks!

PR #96, opened 2018-04-27
Distribute module build

Andarist:
This should allow modern bundlers to do tree-shaking (although not sure if there is much to tree-shake here, but it **also** allows for using other techniques such as "scope hoisting" a.k.a module concatenation.

Inspecting your umd bundle (minified & gzipped):
- before PR - 4739 bytes
- after PR - 4475 bytes

This is nearly 6% reduction and similar amount of bytes should be removed from application bundles using `xstate` with `"module"` entry (ofc when using new versions of webpack)


davidkpiano:
Great, thanks!

PR #102, opened 2018-05-11
Fix typo describing the purpose of onExit actions

dstockto:
The onExit indicates that they run when entering a state instead of when leaving a state.


davidkpiano:
Thanks!

PR #115, opened 2018-05-29
Fixes #114

mogsie:
Fixes #114.

State.inert didn't copy over any activites from the old state.

This is not the best way to deal with activities; ideally it should go through the current configuration of states and build up a correct view of all of the running activities based on which states are active, instead of relying on input like this.

But it gets the job done for normal use cases, no need to protect people from shooting themselves in the foot, eh?  Jolly good!


davidkpiano:
LGTM, thanks!

PR #127, opened 2018-06-04
Use import helpers from tslib package

Andarist:
This should allow bundlers to dedupe typescript helpers across packages using `tslib`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,5 +44,7 @@
     "webpack": "^3.5.6",
     "xml-js": "^1.6.2"
   },
-  "dependencies": {}
+  "dependencies": {
+    "tslib": "^1.9.2"
+  }
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,6 +8,7 @@
     "strictNullChecks": true,
     "outDir": "./lib",
     "moduleResolution": "node",
+    "importHelpers": true,
     "lib": [
       "es6",
       "dom"
--- a/yarn.lock
+++ b/yarn.lock
@@ -2858,6 +2858,10 @@ tslib@^1.7.1:
   version "1.8.0"
   resolved "https://registry.yarnpkg.com/tslib/-/tslib-1.8.0.tgz#dc604ebad64bcbf696d613da6c954aa0e7ea1eb6"
 
+tslib@^1.9.2:
+  version "1.9.2"
+  resolved "https://registry.yarnpkg.com/tslib/-/tslib-1.9.2.tgz#8be0cc9a1f6dc7727c38deb16c2ebd1a2892988e"
+
 tslint@^5.7.0:
   version "5.8.0"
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -44,5 +44,7 @@
     "webpack": "^3.5.6",
     "xml-js": "^1.6.2"
   },
-  "dependencies": {}
+  "dependencies": {
+    "tslib": "^1.9.2"
```
davidkpiano: Can this be a devDependency?

Andarist: Not quite, the point of this PR is to output this:
```js
import * as tslib_1 from "tslib";

// ...

var stateNode = new StateNode(tslib_1.__assign({}, stateConfig, { key: key, parent: _this }));
```

instead of this:
```js
var __assign = (this && this.__assign) || Object.assign || function(t) {
    for (var s, i = 1, n = arguments.length; i < n; i++) {
        s = arguments[i];
        for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p))
            t[p] = s[p];
    }
    return t;
};

// ...

var stateNode = new StateNode(__assign({}, stateConfig, { key: key, parent: _this }));
```

`tslib` is a real dependency that if used is supposed to be added to `"dependencies"` as it's imported/required

Andarist: @davidkpiano friendly 🏓 😉

davidkpiano: Hey, sorry for the delay. I want to avoid adding any dependencies to this package; is there any way this can be done without an extra dependency?

Andarist: > Hey, sorry for the delay. 

Not a problem 😃 

> I want to avoid adding any dependencies to this package; is there any way this can be done without an extra dependency?

There is not, because actual source code (and ofc compiled files) will refer to this dependency and we want people having it installed automatically to avoid missing deps problems.

Either way (with this PR merged or not) you still have a dependency on that code - it only may be explicit or implicit one, bundled in or kept in separate package.

mogsie: Would it be possible to add this at build time instead?  Perhaps xstate could be persuaded to build two artifacts, one with this dependency and one without?  Different npm package names, but from the same source?  That way, the code base is clean, but bundlers have an option that helps bundlers do their deduping.

Andarist: yes, it's doable - but IMHO it complicates things, users have to chose which flavour they want, for imho no good reason

Andarist:
Regarding - https://twitter.com/DavidKPiano/status/1098437300550336512 😉

This PR could slightly improve application bundles (when used in combination with other libraries authored in TS doing the same thing).

davidkpiano:
Closing this because XState must remain dependency-free. If you know an alternative way to do this without introducing another dependency, I would gladly take a loo.

Andarist:
What's a technical argument against this? Because as mentioned - you already depend on that code. No harsh feelings - I'm just curious.

davidkpiano:
We can make an `assignObject(...)` utility function that does the same thing without depending on another package.

PR #142, opened 2018-06-27
Fixing #131: Transition to self wrongly stops activity

### LINKED ISSUE #131: [3.3.2] Transition to self wrongly stops activity
When in a state with a current activity, and you transition to the same state (_without_ setting `internal: true`), the activity is wrongly reported as stopped.

Example machine: ([Codesandbox: see console](https://codesandbox.io/s/6lnkqxy25k))

```js
const { Machine } = require("xstate");

const machine = Machine({
    key: "test",
    initial: "test",
    states: {
        test: {
            activities: ["ACTIVITY"],
            on: {
                NEXT: [
                    {
                        // internal: true,
                        target: "test",
                    },
                ],
            },
        },
    },
});

const log = (ix, state) => {
    console.log(`${ix}\n\n${JSON.stringify(state, null, 4)}\n\n`);
};

log(1, machine.initialState);

log(2, machine.transition(machine.initialState, "NEXT"));
```

With 3.3.2, this is the output:

```

1

{
    "value": "test",
    "actions": [
        {
            "type": "xstate.start",
            "activity": "ACTIVITY",
            "data": {
                "type": "ACTIVITY"
            }
        }
    ],
    "activities": {
        "ACTIVITY": true
    },
    "data": {},
    "events": []
}

 
2

{
    "value": "test",
    "history": {
        "value": "test",
        "actions": [
            {
                "type": "xstate.start",
                "activity": "ACTIVITY",
                "data": {
                    "type": "ACTIVITY"
                }
            }
        ],
        "activities": {
            "ACTIVITY": true
        },
        "data": {},
        "events": []
    },
    "actions": [
        {
            "type": "xstate.stop",
            "activity": "ACTIVITY",
            "data": {
                "type": "ACTIVITY"
            }
        },
        {
            "type": "xstate.start",
            "activity": "ACTIVITY",
            "data": {
                "type": "ACTIVITY"
            }
        }
    ],
    "activities": {
        "ACTIVITY": false
    },
    "data": {},
    "events": []
}
```

In state 2, there are actions to stop and start ACTIVITY. But the resulting `activities` object shows `ACTIVITY: false`. I expected this to be `true`.

adiun:
Fixes #131.

StateNode should handle the exit states first before the entry states to determine activities.

Also added two tests:
(1) testing that same activities are maintained during a normal transition
(2) testing that same activities are maintained during a self transition

Btw, 2 tests are failing but that's because of commit f93cb57 which added tests that are meant to fail.

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -653,16 +653,6 @@ class StateNode {
 
     const activityMap = { ...state.activities };
 
-    Array.from(transition.entryExitStates.entry).forEach(stateNode => {
-      if (!stateNode.activities) {
-        return; // TODO: fixme
-      }
-
-      stateNode.activities.forEach(activity => {
-        activityMap[getActionType(activity)] = true;
-      });
-    });
-
     Array.from(transition.entryExitStates.exit).forEach(stateNode => {
       if (!stateNode.activities) {
         return; // TODO: fixme
@@ -670,6 +660,16 @@ class StateNode {
 
       stateNode.activities.forEach(activity => {
         activityMap[getActionType(activity)] = false;
+      });
+    });
+
+    Array.from(transition.entryExitStates.entry).forEach(stateNode => {
+      if (!stateNode.activities) {
+        return; // TODO: fixme
+      }
+
+      stateNode.activities.forEach(activity => {
+        activityMap[getActionType(activity)] = true;
       });
     });
 
--- a/test/activities.test.ts
+++ b/test/activities.test.ts
@@ -203,6 +203,21 @@ describe('transient activities', () => {
           B1: '.B1',
           B2: '.B2'
         }
+      },
+      C: {
+        initial: 'C1',
+        states: {
+          C1: {
+            activities: ['C1'],
+            on: {
+              C: 'C1',
+              C_SIMILAR: 'C2'
+            }
+          },
+          C2: {
+            activities: ['C1'],
+          },
+        }
       }
     }
   });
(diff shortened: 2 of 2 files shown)


davidkpiano:
Thank you!

mogsie:
Good catch @adiun!

karl:
I've just hit this bug in `3.3.3` while attempting to move our app from a custom state machine runner across to xstate.

@davidkpiano Is there an ETA for when this bug fix will be released? It is currently blocking our ability to use xstate, and we really want to use it and take advantage of your awesome work!

davidkpiano:
@karl Since this is in master, it will have to be in 4.0 (sorry, should have released a patch for this one), which you can use already via `xstate@next`.

karl:
No worries!

What are you feelings on the stability of `4.0.0`? 

Is it stable enough to use in production? If not would it be helpful if I back ported this fix to the `3.3.3` version of the code so it can be released as a patch?

davidkpiano:
So `4.0.0` isn't _officially_ released yet - however, everything except the Interpreter (which is optional, and you aren't using it currently anyway) is stable and well-tested, and the only changes being made right now are to:

- how activities are implemented in the Interpreter (not a change from 3.x, since activities were just strings)
- various Interpreter methods.

The biggest breaking change is in this syntax:

```js
on: {
  FOO: {
    someState: { actions: ['doThis'] }
  }
}
```

Which is now:
```js
on: {
  FOO: {
    target: 'someState',
    actions: ['doThis']
  }
}
```

karl:
Awesome, thanks for the info!

I had a quick play around with `4.0.0` and noticed the change in how activity actions are returned in the actions array.

In `3.3.3` they come back in the form:

```
{
  "type": "xstate.start",
  "activity": "activity1",
  "data": {
    "type": "activity1"
}
```

And in `4.0.0` they come back in the form:

```
{
  "type": "xstate.start",
  "activity": {
    "type": "activity1"
  }
}
```

But that is a very simple tweak for me to make in my interpreter.

Given that I can use `internal` transitions to get the behaviour I want in `3.3.3` I'm going to continue with it for now, and will swap across to `4.0.0` when it gets an official release.

davidkpiano:
> Given that I can use internal transitions to get the behaviour I want in 3.3.3 I'm going to continue with it for now, and will swap across to 4.0.0 when it gets an official release.

That's a good idea.

And yeah, the original activity syntax was redundant.

karl:
@davidkpiano Thanks for the detailed (and quick) replies. It makes it a real pleasure to work with xstate!

Issue #148, opened 2018-07-18
Upgrade to Webpack v4


davidkpiano:
This issue is to track upgrading `xstate` to Webpack v4, especially for making the package [tree-shakeable](https://webpack.js.org/guides/tree-shaking/#), which it should already be capable of doing, since all files are side-effect-free.

We will also want to set up a single configuration file so that we output four separate files:
- `xstate.js` (the core xstate Machine creator)
- `xstate.utils.js` (graph and testing utilities)
- `xstate.patterns.js` (helpful state machine patterns, such as toggles, sequences, promises, etc.)
- `xstate.interpreter.js` (the default xstate interpreter, with inspiration from @avaragado's `xstateful`)

Andarist:
I'm up for the task - but I'm not sure what we want to achieve here, so I'd like to establish it first.

In general when consuming this package bundlers reach out for files generated by TS - the ones specified in main & module entries of the package.json.

webpack is used here merely to create UMD filed & they are most commonly non-treeshakeable & also their usage is rather niche. Also - you could (I could help with that) switch to rollup for generating them as in general it does a better job when it comes to bundling libraries.

What could be done is to annotate all top-level calls with `#__PURE__` comments to let UglifyJS eliminate them as dead code (if their results stay unused ofc). This can be automated.

davidkpiano:
I would be open to using Rollup 👍

Andarist:
I'm still unsure about the goals though, should I just rollup 4 separate (self-contained) UMD bundles? Or should I flat bundle the library into 4 files but with shared (common) parts?

Andarist:
friendly 🏓

davidkpiano:
@Andarist Sorry for the late reply!

> Or should I flat bundle the library into 4 files but with shared (common) parts?

^ This - the main `xstate.js` bundled file should be self-contained, though. The others can depend on it.

davidkpiano:
Closing this - using Rollup instead.

PR #165, opened 2018-08-10
Use rollup for UMD bundles

Andarist:
related to #148

What has been done?
- removed `yarn.lock`, it seems like you are using npm rather than yarn (judging by CI config), and u had both `package-lock.json` & `yarn.lock` committed to the repo, so it was confusing
- switched to rollup for UMD builds, keeping all configs in a single file

What needs to be done?
- ~~I would like to get rid of webpack entirely, it's still used by `watch` script but I'm not sure how do you use it (or if you even use it at all?), what's your preference here? Should rollup output an unminified file on watch?~~
- Not sure how useful are those UMD bundles, how are they consumed? those created files have common code bundled in them - we could dedupe it with code splitting, but then we would force multiple scripts on the user (common chunk would have to be loaded too)

**Results**
Tested with `gzip -c filename.js | wc -c`

Before this PR
```
xstate.js 6062
xstate.utils.js 1900
xstate.interpreter.js 1542
```

Intermediate attempt bundling from `es` directory files 
```
xstate.js 5795
xstate.utils.js 1688
xstate.interpreter.js 835
```

After this PR (bundling from source files)
```
xstate.js 5792
xstate.utils.js 1688
xstate.interpreter.js 835
```

As we can see bundling from `es` & `src` has no real difference, but bundling from `src` has such advantages:
- it gets rid of a rollup warning about rewritten top level `this` (TS inserts its for inserted helper like `__assign`)
- you can run the script separately from the `build:es` script


Andarist:
Removed the webpack completely

Andarist:
@davidkpiano thanks for merging, is there any work here left to do in your opinion regarding bundling? I can help with that if there is anything you have in mind.

PR #166, opened 2018-08-15
Fixed function naming - implemented flatMap is actually just flatten

Andarist:
Additionally `flatten` could be shortened to:
```js
export const flatten = <T>(array: T[][]): T[] => {
  return [].concat(...array);
};
```
but while this has the same functionality TypeScript yells at me with:
```
src/utils.ts:196:20 - error TS2345: Argument of type 'T[]' is not assignable to parameter of type 'ConcatArray<never>'.
  Types of property 'slice' are incompatible.
    Type '(start?: number | undefined, end?: number | undefined) => T[]' is not assignable to type '(start?: number | undefined, end?: number | undefined) => never[]'.
      Type 'T[]' is not assignable to type 'never[]'.
        Type 'T' is not assignable to type 'never'.

196   return [].concat(...array);
                       ~~~~~~~~
```

And I have no idea what's the problem.

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -8,7 +8,7 @@ import {
   pathsToStateValue,
   pathToStateValue,
   getActionType,
-  flatMap,
+  flatten,
   mapFilterValues,
   nestedPath
 } from './utils';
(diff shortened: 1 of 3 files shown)


davidkpiano:
Try this and see if it makes TypeScript happier:

```ts
export const flatten = <T>(array: T[][]): T[] => {
  return ([] as T[]).concat(...array);
};
```

Andarist:
It did 😉 still slightly unsure why the force cast is needed, this should be easily inferable 🤔

PR #176, opened 2018-09-13
fix flushEventQueue by using shift instead of unshift

Calvin-Lin:
Here should be `shift()` instead of `unshift()`

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -238,7 +238,7 @@ export class Interpreter<TContext> {
   }
   private flushEventQueue() {
     if (this.eventQueue.length) {
-      this.send(this.eventQueue.unshift());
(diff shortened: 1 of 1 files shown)


Calvin-Lin:
Yes. The interpreter is really helpful. 

We are trying to use xstate as core state machine in our open source chatbot framework [Bottender](https://github.com/Yoctol/bottender). And the launch of interpreter will save us a lot of time from implementing it ourself. Thanks a lot!

davidkpiano:
Awesome, there's still some things that will be added to it before V4 is finalized, so feel free to give feedback.

E.g., I'm going to add more specialized listeners for `onAction()`, `onActivityStart()`, etc. but that can be done after the V4 release.

PR #203, opened 2018-10-17
Fix getStateNodeByPath method argument mutation

### LINKED ISSUE #202: getStateNodeByPath mutates the path argument
## Bug or feature request?
Bug
## Description:
`getStateNodeByPath` function mutates the path argument.
```javascript
import { Machine } from "xstate";

const machine = Machine({
  initial: "a",
  states: {
    a: {},
    b: {}
  }
});

const path = ["a"];
const stateNode = machine.getStateNodeByPath(path);
console.log(path); // path is: []
```

## (Bug) Expected result:
The argument shouldn't be mutated.
## (Bug) Actual result:
The argument is mutated.
## (Bug) Potential fix:
The issue is related with this line https://github.com/davidkpiano/xstate/blob/master/src/StateNode.ts#L1243. A potential fix would be to clone the array in the [toStatePath function](https://github.com/davidkpiano/xstate/blob/master/src/utils.ts#L86) or before [this loop](https://github.com/davidkpiano/xstate/blob/master/src/StateNode.ts#L1242).
Reproduction: https://codesandbox.io/s/9jy71848k4

carloslfu:
Fixes #202 .

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -1235,7 +1235,7 @@ class StateNode<
    * @param statePath The string or string array relative path to the state node.
    */
   public getStateNodeByPath(statePath: string | string[]): StateNode<TContext> {
-    const arrayStatePath = toStatePath(statePath, this.delimiter);
+    const arrayStatePath = toStatePath(statePath, this.delimiter).slice();
     let currentStateNode: StateNode<TContext> = this;
     while (arrayStatePath.length) {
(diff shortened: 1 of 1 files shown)


davidkpiano:
Thanks for this - but it's better if we fix it in the `toStatePath` function itself.

carloslfu:
Oh! my bad. Yeah, `statePath` can be a string. Fixed now.

PR #223, opened 2018-11-06
Fix hierarchical doc broken url

macs91:
The url at the bottom that references to strict mode config doc is broken ([see here](https://xstate.js.org/docs/guides/hierarchical/)). I'm not really sure the change I'm proposing solves it since I'm writing it from the browser without testing it.


davidkpiano:
Thanks! This has been fixed (the docs were redone in VuePress).

PR #257, opened 2018-11-26
Add sign in form example

bradwoods:


sbussard:
There are a lot of reasons why this is not a great way to use `xstate`. 
You shouldn't have do to this sort of thing, or you're missing the point of a state machine

```
          const disableEmail =
            state.matches('passwordErr') ||
            state.matches('awaitingResponse') ||
            state.matches('serviceErr')
          const disablePassword =
            state.matches('emailErr') ||
            state.matches('awaitingResponse') ||
            state.matches('serviceErr')
          const disableSubmit =
            state.matches('emailErr') ||
            state.matches('passwordErr') ||
            state.matches('awaitingResponse')
          const fadeHeading =
            state.matches('emailErr') ||
            state.matches('passwordErr') ||
            state.matches('awaitingResponse') ||
            state.matches('serviceErr')
```

bradwoods:
I refactored the form using React hooks & made it less _strict_ following some feedback.

tamebadger:
@bradwoods I think the link your updated codesandbox does not work, I can access [this one](https://codesandbox.io/embed/239lj2xzqp) which is the old one ? But I can't access [this one](https://codesandbox.io/embed/l3r07jkxx9) which is you new one right ?

Anyhow, let me know if you need help with this, maybe we can get it over the line together!

bradwoods:
Hey @tamebadger, thanks for the assist. Your right, the link was broken, fixed now. [This](https://codesandbox.io/s/l3r07jkxx9) has my latest attempt. 1 issue I'm having with it is when I change to an error state from the invoke, the error onEntry actions (focusEmailInput or focusPasswordInput) fire but don't actually focus the elements.

PR #265, opened 2018-11-30
Fix link to Guards page at bottom

LearningNerd:
Just fixing a broken link. I copied the convention from other pages here in the docs where you links to an internal page using `./pagename.md`


davidkpiano:
Thanks so much!

PR #278, opened 2018-12-15
Add Codesandbox integration to docs

carloslfu:
Using this integration all the examples can be easily maintained in this repo. Also, the repo is the single source of truth and there is less risk of accidentally broken examples.

So, for adding more examples we should:

- Add a folder with the sandbox to `docs/sandboxes`.
- Add it to the documentation using this syntax: `@[codesandbox](example-name)`.
- Make a PR.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -1,9 +1,18 @@
+const mdCodesandboxPlugin = require('markdown-it-codesandbox-embed')
+
 module.exports = {
   title: 'XState Docs',
   base: '/docs/',
   description:
     'Documentation for XState: State Machines and Statecharts for the Modern Web',
   ga: 'UA-129726387-1',
+  markdown: {
+    config: md => {
+      md.use(mdCodesandboxPlugin, {
+        directory: 'sandboxes',
+      })
+    }
+  },
   themeConfig: {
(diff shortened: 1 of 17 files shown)


docs/.vuepress/config.js
```diff
@@ -1,9 +1,18 @@
+const mdCodesandboxPlugin = require('markdown-it-codesandbox-embed')
```
davidkpiano: Is this the right one? I only found https://www.npmjs.com/package/markdown-it-codesandbox

davidkpiano: The issue is that this will create a new CodeSandbox every single time the docs are deployed, which is less than ideal, especially if developers are linking to the CodeSandbox.

Is there a way you can make it so that it doesn't create a new one, and instead just overwrites the existing one if it exists?

carloslfu: From CompuIves/codesandbox-client#1311, Ives mention that "the define endpoint creates a sandbox anonymously and reuses the same sandbox if the files are the same". Seems that update functionality is comming but for now I think reusing the same sandbox if files are the same is good.

davidkpiano: Sure we'll stay with the current solution.

PR #292, opened 2019-01-03
.off(listener) now removes listener from all types.

### LINKED ISSUE #291: .off(listener) should remove listener for context, send, done, event, etc listeners.
## Bug or feature request?
Bug

## Description:
The `.off(listener)` only removes transition listeners from the interpreter. I'd like a way to remove listeners for `onChange`, `onDone`, `onStop`, `onEvent`, and `onSend` without stopping the whole interpreter.

## (Bug) Expected result:
`.off(listener)` should remove the listener from any of the active listeners.

## (Bug) Actual result:
`.off(listener)` only removes from the transition listeners.

## (Bug) Potential fix:
I'm not sure what the desired API would be. `.off(listener)` could delete from all listeners, or we could change the API to `.removeListener('change', listener)` or `.removeListener('transition', listener)`. I'd like to discuss the API before creating a PR.

derek-duncan:
Closes #291

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -288,9 +288,14 @@ export class Interpreter<
    * @param listener The listener to remove
    */
   public off(
-    listener: StateListener<TContext, TEvent>
+    listener: () => void
   ): Interpreter<TContext, TStateSchema, TEvent> {
     this.listeners.delete(listener);
+    this.eventListeners.delete(listener);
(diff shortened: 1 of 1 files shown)


davidkpiano:
One error:

```
src/interpreter.ts(325,49): error TS2345: Argument of type 'StateListener<TContext, TEvent>' is not assignable to parameter of type '() => void'.
```

PR #303, opened 2019-01-16
adding TEvent to allow type inference

sangallimarco:
adding TEvent to allow type inference in state machine configuration especially for actions


sangallimarco:
@davidkpiano this would be useful to have when using typescript as we cannot infer event type in actions right now

PR #313, opened 2019-01-20
Update intenral.md

dschnare:
Rename the `key` state node setting to `id`.

Change the value of the `id` state node setting to `direction`.

Remove the section talking about what internal transitions are a shortcut for.
This was removed to avoid confusion and technically the documentation was misleading since the internal transitions were not equivalent to the examples because the examples were external transitions, which do not have the same side affects as internal transitions.


davidkpiano:
Thank you! You're absolutely right.

PR #339, opened 2019-01-31
Correct recipe example for Rxjs

RikuVan:
Copy-pasted you Rxjs recipe to play with noticed this little error.

### DIFF
--- a/docs/recipes/rxjs.md
+++ b/docs/recipes/rxjs.md
@@ -21,7 +21,7 @@ state$.subscribe(state => {
   console.log(state);
 });
 
-const event$ = event$.subscribe(event => {
+const event$ = state$.subscribe(event => {
   // a stream of event objects
   service.send(event);
 });


davidkpiano:
Good catch, thank you!

michaelsbradleyjr:
I realize this was merged two months ago, but... is it actually correct? I'm having trouble understanding how it could work as a useful pattern.

It seems like `event$` needs instead to be setup with `fromEventHandler`, similarly to `state$`:

```js
const event$ = fromEventPattern(callback => {
  someKindofEmitter.onEvent(callback)
});

event$.subscribe(event => {
  // a stream of event objects
  service.send(event);
});
```

The line `someKindofEmitter.onEvent(callback)` might in practice be replaced by `document.addEventListener('click', callback)` or node-style `emitter.on('some-event', callback)`.

See: https://rxjs.dev/api/index/function/fromEventPattern.

If I'm wrong, I'd be grateful for help in understanding how the current example is supposed to work.

davidkpiano:
You're correct, I'll update the docs.

PR #344, opened 2019-02-05
Improve Internal Transitions docs

hnordt:
`direction.left` is not a valid state reference for `wordMachine`, so I've renamed it to just `left` and highlighted the full path as `#direction.left`

### DIFF
--- a/docs/guides/internal.md
+++ b/docs/guides/internal.md
@@ -24,7 +24,7 @@ const wordMachine = Machine({
 });
 ```
 
-The above machine will start in the `'direction.left'` state, and based on what is clicked, will internally transition to its other child states. Also, since the transitions are internal, `onEntry`, `onExit` or any of the `actions` defined on the parent state node are not executed again.
+The above machine will start in the `'left'` state (for reference, the full path is `'#direction.left'`), and based on what is clicked, will internally transition to its other states. Also, since the transitions are internal, `onEntry`, `onExit` or any of the `actions` defined on the parent state node are not executed again.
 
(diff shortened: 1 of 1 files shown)


hnordt:
I've also removed the "child states" reference. `wordMachine` don't have any child state..

docs/guides/internal.md
```diff
@@ -35,10 +35,10 @@ states: {
 },
 on: {
   // internal transitions, equivalent to the first example
-  LEFT_CLICK: { target: 'direction.left', internal: true },
-  RIGHT_CLICK: { target: 'direction.right', internal: true },
-  CENTER_CLICK: { target: 'direction.center', internal: true },
-  JUSTIFY_CLICK: { target: 'direction.justify', internal: true }
```
davidkpiano: These should be `'.left'`, `'.right'`, etc.

hnordt: I tested using the Visualizer, `{ target: 'center', internal: true }` has the same effect as `{ target: '.center', internal: true }`. Is that a bug?

davidkpiano: It might be. `target: 'center'` means "the sibling state with key of `"center"`."

hnordt: I think I got it. `.someState` is actually `currentStateNode.childState`. Am I wrong?

davidkpiano: You're correct. Let's open an issue (or better yet, a PR) for adding this to the docs here: https://xstate.js.org/docs/guides/ids.html

PR #360, opened 2019-02-08
Context Guide: mention that machine.context is the same as machine.in…

hnordt:
…itialState.context

### DIFF
--- a/docs/guides/context.md
+++ b/docs/guides/context.md
@@ -130,6 +130,13 @@ dynamicCounterMachine.initialState.context;
 // => { count: 42, time: 1543687816981 }
 ```
 
+Or directly from the machine:
+
+```js
+dynamicCounterMachine.context;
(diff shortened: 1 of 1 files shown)


davidkpiano:
Not sure if this is helpful. `.initialState.context` is more explicit of what context that is.

hnordt:
@davidkpiano makes sense.

But I think it would be nice if you could mention somewhere that `machine.initialState.context` is the same as `machine.context`.

In the Machines Guide we have the following snippet:

```js
const testLightMachine = lightMachine.withContext({
  // merge with original context
  ...lightMachine.context,
  elapsed: 1000
});
```

So after reading both guides (Context and Machines), I was wondering if `machine.initialState.context` and `machine.context` they were the same thing.

davidkpiano:
Good point. I'll add it to the docs.

PR #362, opened 2019-02-12
fix comment inconsistency

kandros:
found inconsistency in comment


davidkpiano:
In this case, it'd come from `return reject('No query specified');` so it'd just be `event.data`. There is an inconsistency, so that can either be fixed or you can change the reject line to `return reject({message: 'No query specified'})`. Thanks!

PR #368, opened 2019-02-16
fix: support multiple simultaneous services

### LINKED ISSUE #367: BUG: array of invoked services overwrites itself in nested states
## Bug or feature request?

Bug

## Description:

It's not possible to provide an array to the `invoke` key in any other state than a top-level one, only the last service is kept & found.

## (Bug) Expected result:

All services are invoked.

## (Bug) Actual result:

Only the last service is invoked, all others write warnings to the console:

> No service found for invocation 'machine.one.two.invocation[0]' in machine 'machine'

Where that selector & ID string are different per machine.

## (Bug) Potential fix:

https://github.com/davidkpiano/xstate/blob/c7392dc48b11103d716348a6b1230c6507575645/src/StateNode.ts#L287-L290

I'm pretty sure `this.machine.options.services` should be included in that object spread, to avoid it being overwritten every time the array of services iterates.

https://codesandbox.io/s/zr7r6z955l?expanddevtools=1&module=%2Fsrc%2Findex.js

```js
const machine = Machine({
    id: "machine",
    initial: "one",

    states: {
        one: {
            initial: "two",

            states: {
                two: {
                    invoke: [
                        // Never invoked, throws a warning
                        {
                            id: "child",
                            src: () => () => {
                                console.log("first");

                                return () => {};
                            },
                        },
                        // Because when this one was iterated it overwrote
                        // the previous this.machine.options.services object
                        {
                            id: "child2",
                            src: () => () => {
                                console.log("second");

                                return () => {};
                            },
                        },
                    ],
                },
            },
        },
    },
});
```

tivac:
Fixes #367 

This adds test cases for the two failing examples in #367 and also adds a variation of the fix I suggested. After more staring at `StateNode.ts` I think the original object being spread there might've been a copy-paste error, I can't come up with a reason why you'd want to set `machine.options.services` to the node or parent's `options.services`. If that's wrong and that is desired behavior it's easy to add that line back.

All tests new & old are passing for me locally with the change to `StateNode.ts` though, so if it is necessary it probably needs a test for that behavior 😃

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -286,7 +286,7 @@ class StateNode<
         const invokeSrc = `${this.id}:invocation[${i}]`; // TODO: util function
         this.machine.options.services = {
           [invokeSrc]: invokeConfig.src,
-          ...(this.parent || this).options.services
+          ...(this.machine.options.services || {})
         };
 
         return {
--- a/test/invoke.test.ts
+++ b/test/invoke.test.ts
@@ -1088,4 +1088,130 @@ describe('invoke', () => {
         .start();
     });
   });
+
+  describe('multiple simultaneous services', () => {
+    // @ts-ignore
+    const multiple = Machine({
+      id: 'machine',
+      initial: 'one',
+
+      context : {},
+      
+      on : {
+        ONE : {
+          actions : assign({
+            one : 'one'
+          }),
+        },
+
+        TWO : {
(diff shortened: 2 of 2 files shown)


davidkpiano:
You're right, it should be `machine.options.services`. Thank you!

djkirby:
🙏

PR #371, opened 2019-02-21
Fixup env checks

Andarist:
Regarding - https://twitter.com/DavidKPiano/status/1098437300550336512 😉 

When performing env checks you should never reference `process` directly - replacing plugins are only concerned with replacing `process.env.NODE_ENV` lookup. Referencing `process` causes i.e. webpack to include a polyfill for this object.

Because of that developer-friendly warning which were meant to be removed in production bundles stayed in the bundles.

This has no dramatic effect but it allow bundlers to remove those unwanted strings in production, the UMD bundle went from 10.87 KB to 10.54 KB 

In general I don't recommend storing env check result in a variable (inline check is easier for minifiers to remove), and I strongly would advise against exporting it from a file - especially given a fact that you don't flat bundle and depending on the consumer's setup it might not be possible to check this value across module-boundary. 

This also provides a browser-friendly build which was actually requested originally ( https://github.com/davidkpiano/xstate/issues/227 ) but the fix was done differently. cc @TimvdLippe

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -3,7 +3,7 @@ import { terser } from 'rollup-plugin-terser';
 import rollupReplace from 'rollup-plugin-replace';
 import fileSize from 'rollup-plugin-filesize';
 
-const createConfig = ({ input, output }) => ({
+const createConfig = ({ input, output, target }) => ({
   input,
   output,
   plugins: [
@@ -14,11 +14,14 @@ const createConfig = ({ input, output }) => ({
       clean: true,
       tsconfigOverride: {
         compilerOptions: {
-          declaration: false
+          declaration: false,
+          ...(target ? { target } : {})
         }
       }
     }),
-    terser(),
+    terser({
+      toplevel: true,
+    }),
     fileSize()
   ]
 });
@@ -47,5 +50,13 @@ export default [
       format: 'umd',
       name: 'XStateInterpreter'
     }
+  }),
+  createConfig({
+    input: 'src/index.ts',
+    output: {
+      file: 'dist/xstate.web.js',
+      format: 'esm'
+    },
+    target: 'ES2015'
   })
 ];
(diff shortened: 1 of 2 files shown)


TimvdLippe:
I think this change regresses the case where `process` is undefined in the browser. E.g. in the browser, you can't have a reference to a variable called `process`. Therefore, I think this is going to break that use case.

Andarist:
That's why I'm building `./dist/xstate.esmodules-browsers.js` bundle which should be browser compatible (no references to `process`). It's the same situation as in https://github.com/redux-saga/redux-saga/pull/1717 , https://github.com/redux-saga/redux-saga/pull/1717 , etc

TimvdLippe:
Ah okay. If that works, good! I am not an expert in rollup, so I don't really understand how it is makes sure it removes the reference. Is it the `toplevel: true`?

Andarist:
`toplevel: true` is for uglify/terser - so it can mangle "top level" variables - otherwise it assumes they might be assigned as globals to window object. I think this actually should be the default for module sources, but maybe logic handling this wasnt implemented yet

> I am not an expert in rollup, so I don't really understand how it is makes sure it removes the reference

This is done by `rollup-plugin-replace`

rollup.config.js
```diff
@@ -47,5 +50,13 @@ export default [
       format: 'umd',
       name: 'XStateInterpreter'
     }
+  }),
+  createConfig({
+    input: 'src/index.ts',
+    output: {
+      file: 'dist/xstate.esmodules-browsers.js',
```
davidkpiano: Can we call this `dist/xstate.esm.js`?

Andarist: Technically we can, but `.esm.js` is often used for "module" field in package.json - which is not the same as this thing here. Didn't want to confuse anyone with this naming.

davidkpiano: Sorry to nit, I'm just not a fan of the name (not very user-friendly). What about `xstate.web.js`?

Andarist: No problem - I'll change in a moment. I've based my original name on https://babeljs.io/docs/en/babel-preset-env#targetsesmodules

Andarist:
@davidkpiano friendly 🏓

davidkpiano:
Thank you!

PR #374, opened 2019-02-22
WIP: Stronger typings for transitions

ismailhabib:
Hi, I would like to work on making improvements on the typings on XState for few things that I noticed. The MR is not complete but I would like to see if I'm on the right direction. Thanks!


henrikra:
Could you add before and after with IDE pictures? :)

ismailhabib:
Before:
screenshot_20190228_143624

After:
screenshot_20190228_152732

davidkpiano:
We will be reworking this in v5, not for v4 as it is a significant change.

PR #387, opened 2019-03-08
Fix return type of example hook

### LINKED ISSUE #375: [Bug] Hook's useEffect return type is possibly incorrect
## Description:

I expect that when I use the hook specified [in the documentation](https://xstate.js.org/docs/recipes/react.html#hooks), there is no Typescript compiler issues.

## (Bug) Actual result:

When using the hook, as published, I am receiving an error that `Type 'Interpreter<DefaultContext, any, EventObject>' is not assignable to type 'void'.  TS2345` (screenshot attached)

## (Bug) Potential fix:

This is by no means an actual fix, but I did sidetrack the issue by doing the following:

`return () => (service.stop() as unknown) as void;`

image

ksaldana1:
Fixes #375 

The compiler error was due to React's [rules](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L793) on `useEffect` cleanup function return values. Easy fix to just remove the arrow function.

```typescript
// NOTE: callbacks are _only_ allowed to return either void, or a destructor.
// The destructor is itself only allowed to return void.
type EffectCallback = () => (void | (() => void | undefined));
```

### DIFF
--- a/docs/recipes/react.md
+++ b/docs/recipes/react.md
@@ -85,9 +85,11 @@ export function useMachine(machine) {
 
   // Stop the service when the component unmounts
   useEffect(() => {
-    return () => service.stop();
+    return () => {
+      service.stop();
+    }
   }, []);
-
+ 
   return [current, service.send];
 }
 ```


davidkpiano:
Thank you! You're right.

PR #388, opened 2019-03-10
docs(hierarchical): add note about state re-entry

### LINKED ISSUE #384: Hierarchical state machines and onEntry/onExit/activities
## Bug or feature request?

Bug (or a question)

## Description:

When using nested / hierarchical state machines, transitions to nested states cause `onEntry`/`onExit` actions / restarting `activities`

## (Bug) Expected result:

```js
const XState = require("xstate")

const { Machine, interpret } = XState

const testMachine = Machine(
  {
    id: "test",
    initial: "foo",
    states: {
      foo: {
        initial: "bar",
        on: {
          RESET: `#test.foo.bar`,
        },
        onEntry: `enter`,
        onExit: `exit`,
        activities: `fooActivity`,
        states: {
          bar: {
            on: {
              TOGGLE: `baz`,
            },
          },
          baz: {
            on: {
              TOGGLE: `bar`,
            },
          },
        },
      },
    },
  },
  {
    actions: {
      enter: () => {
        console.log("enter 'foo'")
      },
      exit: () => {
        console.log("exit 'foo'")
      },
    },
    activities: {
      fooActivity: () => {
        console.log("start activity in 'foo'")
        return () => {
          console.log("stop activity in 'foo'")
        }
      },
    },
  }
)

const service = interpret(testMachine)
  .onEvent(ev => {
    console.log(`EVENT ${ev.type}`)
  })
  .onTransition(state => {
    console.log(`STATE ${JSON.stringify(state.value)}`)
  })
  .start()

service.send("RESET")
```

Should not fire `onExit` and `onEnter` actions / restart `fooActivity` activity on sending `RESET` action as `foo` state is still `active` 

## (Bug) Actual result:

```
start activity in 'foo'
enter 'foo'
EVENT xstate.init
STATE {"foo":"bar"}
exit 'foo'
stop activity in 'foo'
start activity in 'foo'
enter 'foo'
EVENT RESET
STATE {"foo":"bar"}
```

`foo` is never left, but onExit etc is called

## (Bug) Potential fix:

I didn't dig into the code yet. First, I'd like to ask for clarification if current behaviour is expected or it should be considered as bug. [Hierarchical State Machines docs](https://xstate.js.org/docs/guides/hierarchical.html) doesn't cover enter/exit actions / activities behaviour.

https://runkit.com/pieh/5c7c95a947703a00120db854 (same code as above) or copy & paste Machine to [XState Visualizer](https://statecharts.github.io/xstate-viz/), click "reset" event and check browser console

pieh:
Adding just small note to hierarchical state machines docs, documenting parent state re-entry behaviour and adding link to internal transitions.

Closes #384

### DIFF
--- a/docs/guides/hierarchical.md
+++ b/docs/guides/hierarchical.md
@@ -97,4 +97,4 @@ console.log(lightMachine.transition('green', 'UNKNOWN').value);
 
 ## Notes
 
-- Transitions to children states exit and re-enter parent state by default (triggering parent's `onExit`/`onEntry` actions and restarting parent's `activities`). To avoid that behaviour use [Internal Transitions](./internal.md).
+- Transitions from parent state to its children will exit and re-enter parent state by default (triggering parent's `onExit`/`onEntry` actions and restarting parent's `activities`). To avoid that behaviour use [Internal Transitions](./internal.md).
--- a/docs/guides/hierarchical.md
+++ b/docs/guides/hierarchical.md
@@ -94,3 +94,7 @@ If neither a state nor any of its ancestor (parent) states handle an event, no t
 console.log(lightMachine.transition('green', 'UNKNOWN').value);
 // => 'green'
 ```
+
+## Notes
+
+- Transitions to children states exit and re-enter parent state by default (triggering parent's `onExit`/`onEntry` actions and restarting parent's `activities`). To avoid that behaviour use [Internal Transitions](./internal.md).


davidkpiano:
Hmm, this might be a little misleading I believe. If you're in state `foo.bar` and you transition to `foo.baz`, the parent `foo` activities will _not_ be restart, nor will the parent be exited/re-entered.

pieh:
To be honest I lack both terminology and context here. I've made some adjustments but I'm not sure if this makes it any more clear. 

My main point here is basically so people can scan (or even ctrl+f) for keywords like `onEntry`/`onExit`/`activities` and find link to Internal transitions docs, which covers internal transitions and have example of external transition documenting parent state re-entry.

Issue #373, opened 2019-02-22
Invoke firing before transition completes?


cdomigan:
## Bug or feature request?
Bug

## Description:
On transitioning to a new state, invoke is called prior to the transition ending, leading to an incorrect final state if invoke meanwhile transitions.

## Link to reproduction or proof-of-concept:

https://codesandbox.io/s/k5opkj68vr

cdomigan:
I'm sure this isn't a bug, just a gap in my understanding. How would I go about accomplishing the use-case in the provided example? I want to make sure the states occur in serial: initial -> validateOrder -> showErrors.

davidkpiano:
Confirmed fixed with 4.4.0

PR #392, opened 2019-03-16
Handle transitions correctly

jjkola:
I had problems with callback invocations sending events back to state machine synchronously from callback invocation so I tried fixing that (transition events came in wrong order). In the process I enabled two ignored tests about error handling and noticed the problem with promise rejections. I decided to take a stab at those also.

I also tried to fix synchronously called done callback for parent/child machine (child goes directly to final state from invocation) but couldn't find solution for that easily so that is missing from this pull request. I have the added test case with me if it is needed.

If the promise changes are too controversial/risky then I can drop those and leave only the original callback invoke fix.

The transition problem is not visible if the transition is to final state as in that case the event handling is short circuited skipping the event sending.

#373 seems to be the same case as I was experiencing.


jjkola:
#373 seems to be the same case as I was experiencing.

jjkola:
Seems like the code is still not good enough, although it corrected at least some cases but when I tried to use the corrected version on my own application it was still failing. I will need to rethink the solution.

jjkola:
I was able to find out the reason for the failing code. The reason was that I had plugged the leak when using callback provided by the invoke but I was calling the normal send function -> thus circumventing the fix.

davidkpiano:
Can you merge `master` to this branch (resolve conflicts) if you think this PR is still in good shape to fix the solution?

jjkola:
Ok, I'll merge it and also add the documentation changes. I didn't yet modify the documentation as I wanted to get early feedback on the implementation.

davidkpiano:
The logic for processing events that are in c075315 has already been implemented in master, just don't want you to duplicate any work 😅

jjkola:
I read through your changes and it is mostly the same besides being embedded directly to interpreter code.

The only problems which I see with it is that it is harder to test the functionality exhaustively as there is bunch of different responsibilities and functionalities embedded into Interpreter class and the other being that your implementation does not clear the event queue in the case of errors which could lead to undefined behavior (in the worst case it could lead to security problems with the application using xstate).

I think it would be easy to add the functionality you have added to Interpreter class to the class I created with with much more clear cut functionality and easier testing. For example my current implementation I'm pretty sure I have covered all the bases but what if your code changes at some point. Can you be sure to test all the corner cases?

Also, as there is already a problem with your current implementation I think it is an indication that the functionality should not reside within Interpreter class.

But still it is your call whether you want to accept my implementation or keep the current implementation. If you give me green light I will continue with merging and finishing touches.

jjkola:
I have decided to go forward with the merge just for intellectual curiosity to see what will come out of it.

jjkola:
As I have started merging the code I have to say that I feel that the deferEvents functionality feels hacky and also a big can of worms.

The first concern is that that option can be turned on and off at any point of time which means some of the events can be processed such that deferEvents is in effect and others without it thus could lead to undefined behavior especially if there is events in the queue.

The second concern is that even if we set the option only once at the construction of Interpreter class if the instance is reused then after stop is called there can happen more events which are added to queue which again could lead to undefined behavior.

If the idea is as the documentation says then it should only affect until start then you would have to ensure that is not used after start is called effectively turning itself to another initialization flag which could be better served calling the start earlier.

All in all while it could be nice for "demo" application but for real application it could be the bane of the developer as the developer has no way of reproducing the issue, at least easily, unless exact same situation is observed (with timings and all).

davidkpiano:
> The first concern is that that option can be turned on and off at any point of time

Let's assume that it can't (because it shouldn't be able to be changed). Does that alleviate some concerns?

davidkpiano:
I also like your idea of having a separate EventProcessor class, and would like to merge that in when ready. 👍

jjkola:
I'm currently implementing it in the way that the value can be set only when constructing the interpreter and afterwards it will be ignored. And even if you try to set it afterwards (and especially after calling start-method) it will throw an error.

Also I just noticed that the functionality is by default enabled. I think this is a big no-no especially to such a breaking feature. Fortunately, at least the with the implementation I'm currently producing it should not be a too big problem but still I would like to see the default being false.

jjkola:
Is it only me with problems with current head of the master branch? I'm getting error: https://pastebin.com/uym9nS0L (link will expire in a week)

By the way, what is the preferred way for chat? I assume that here should be only chat related to this pull request.

davidkpiano:
Do you have the latest TypeScript? I remember TS not being able to infer that e.g., `'parallel'` was not just a `string` but indeed part of a `'atomic' | 'compound' | 'parallel' | ...` union type, and it seemed to work in my tests, but that issue might or might not still exist.

We can also chat on https://gitter.im/statecharts/statecharts or https://spectrum.chat/statecharts (you can one-on-one message me on either of those platforms if needed).

Also, thanks for your effort on this so far! Let's get it to the finish line. 💯

jjkola:
I was able resolve it. I think it was caused by using npm link on xstate (I think it caused some stale version). After I unlinked it the xstate tests were properly run (although grapth-tests complain about missing xstate-module but that happened also with older commits and also before unlink).

jjkola:
Hmm, for some reason Github does not show merge commits even though there is changes in some of those merge commits...

jjkola:
If there is no review comments this should be ready to be pulled.

jjkola:
Forgot to run tests for the setTimeout change, which proved to be bad idea. The change broke a test and reviewing the change and the test I realized that I shouldn't have done that change as it breaks the interface contract. I will revert the change and document the reason to code so that if somebody else has same "bright" idea then they will know what they are in for.

davidkpiano:
Merging this into a separate branch to test/refactor. Will merge that branch into `master` as soon as that's ready. Thanks!

Issue #316, opened 2019-01-22
Restored state does not populate nextEvents


kristhetraj:
## Description:

When persisting and then restoring state the nextEvents property is empty. However after doing a transition on the restored machine/state the nextEvents property is correct:

```js
const machine = newMachine();
const savedState = JSON.stringify(machine.initialState);
const restoredStateDef = JSON.parse(savedState);
const restoredState: State<{ claim: any }, any> = State.create(
    restoredStateDef
);
const restoredMachine = interpret(newMachine()).start(restoredState);
const currentState = restoredMachine.machine.transition(
    restoredState, 
    Transition.next
);
```

## (Bug) Actual result:

Here are my console logs from the various states:

```
machine.initialState.nextEvents [ 'next' ]
restoredState.nextEvents []
restoredMachine.state.nextEvents []
currentState.nextEvents [ 'previous', 'next' ]
```

davidkpiano:
Please provide a full reproducible example, including machine definition.

kristhetraj:
Hello, demo in sandbox: https://codesandbox.io/s/505rj518nl

davidkpiano:
This is working as expected - the `nextEvents` are determined by the state machine itself (through the private `state.tree`) and this has the potential to make the saved state very large, especially when there are many event and other state nodes.

We can probably have something like...

```js
// Tentative API
// instead of State.create(...)
const restoredState = machine.resolveState(restoredStateDef);
```

Which will use the machine to determine the `state.tree`, which will provide `nextEvents`, etc.

bradwoods:
This caught me out as well. I'vm working on a project that uses the nextEvents property to render buttons in the UI & makes use of save/ restore state when a user logs in/out. It would be helpful to have a method that restored a machine to exactly what it was.

geddski:
ran into this snag too. going to try the workaround.

geddski:
hmm it's easy enough to persist the `nextEvents` to localstorage etc, but there doesn't seem to be a way to set them on a state node (getter only)?

davidkpiano:
@geddski You can do this:

```js
const persistedState = Object.defineProperty(state, 'nextEvents', {
  value: state.nextEvents
});
```

as a workaround.

geddski:
@davidkpiano smart! 💯

davidkpiano:
The `machine.resolveState(state)` API is implemented in master and will be documented 🔜

davidkpiano:
[...]

PR #393, opened 2019-03-20
(docs) added the recommended way to restore states with machine.resolveState()

loque:
In [#316](https://github.com/davidkpiano/xstate/issues/316) it was identified a problem with the rehydration of `State`s. The fix is already implemented with `machine.resolveState(...)` so I modified the documentation reflecting this changes.

### DIFF
--- a/docs/guides/states.md
+++ b/docs/guides/states.md
@@ -138,7 +138,7 @@ try {
 }
 ```
 
-State can be restored (i.e., rehydrated) using the static `State.create(...)` method and resolved using the public `machine.resolveState(...)` method:
+State can be restored using the static `State.create(...)` method and resolved using the public `machine.resolveState(...)` method:
 
 ```js
 import { State, interpret } from 'xstate';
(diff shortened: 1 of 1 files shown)


davidkpiano:
This will not work exactly as you outlined - `machine.resolveState()` requires a `State` instance, so you'd still have to do this:

```js
const restoredState = State.create(restoredStateDef);
const resolvedState = machine.resolveState(restoredState);
```

If that seems a bit boilerplate-y, you're right. It's assumed that developers would restore state directly with an interpreter:

```js
const restoredState = State.create(restoredStateDef);

// ...

// This will internally call machine.resolveState(restoredState);
service.start(restoredState);
```

loque:
Oh I get it now. Based on what you explained [here](https://github.com/davidkpiano/xstate/issues/316#issuecomment-474636417) I understand that currently is necessary to use `machine.resolveState()`.
Should I document the current behavior and come back when 4.4 is released to update the docs or document both ways explaining that the automatic resolution with `service.start()` will be available in 4.4?

loque:
Let me know if you prefer to add a heads up about the automatic resolution in 4.4

Issue #397, opened 2019-03-22
state.changed remains false for internal transition with assign()


johnyanarella:
## Bug or feature request?

Bug

## Description:

The resulting state is not marked as `changed: true` following an internal transition that includes an `assign()` action that alters the context.

This poses problems when you are only triggering re-rendering logic when `changed: true` (as `@xstate/react`'s `useMachine()` does).

I think the conditions for considering the state to be "changed" (as documented [here](https://xstate.js.org/docs/guides/states.html#state-changed)) should also include the condition where the context was changed (via `assign()`).

## (Bug) Expected result:

Given this (simplified) machine:

```js
import { assign, Machine } from "xstate";

const exampleMachine = Machine(
  {
    id: "example",
    initial: "active",
    context: {
      value: 0
    },
    states: {
      active: {
        on: {
          UPDATE: {
            actions: ["update"]
          }
        }
      }
    }
  },
  {
    actions: {
      update: assign({
        value: (context, event) => event.value
      })
    }
  }
);

const initialState = exampleMachine.initialState;

console.log("Before:", initialState.context.value);
const nextState = exampleMachine.transition(initialState, {
  type: "UPDATE",
  value: 0.5
});
console.log("After:", nextState.context.value);

console.log("Changed?", nextState.changed);
```

I would expect to see:

```
Before: 0
After: 0.5
Changed? true
```

## (Bug) Actual result:

Instead, I see:

```
Before: 0
After: 0.5
Changed? false
```

## (Bug) Potential fix:

My intuition here is that the change detection logic doesn't account for the `assign()` action because it has already been consumed internally (as part of the early batch and compute phase within the machine).

For example, if I add an arbitrary new action (one that the interpreter will be responsible for executing):

```js
const exampleMachine = Machine(
  {
    id: "example",
    initial: "active",
    context: {
      value: 0
    },
    states: {
      active: {
        on: {
          UPDATE: {
            actions: ["update", "other"]
          }
        }
      }
    }
  },
  {
    actions: {
      update: assign({
        value: (context, event) => event.value
      }),
      other: (context, event) => {}
    }
  }
);
```

I get the expected result - i.e. `changed: true` - because the `other` action is present in the state's `actions`.

Given the `assign()` action is already consumed and not present in the new state's `actions`, perhaps it would make sense to compare the context objects as part of `State::changed()`?

[...]

PR #398, opened 2019-03-22
Fix to consider context when evaluating changed() for #397

johnyanarella:
* Modifies the `changed()` getter in `State` to also consider changes to the `context`.
* Adds a corresponding test.

See: #397


johnyanarella:
Per comments in #397, this is probably not the way to go - calculating `changed` in `transition()` is a more direct path to detecting the change (with full knowledge) than inspecting the result of side effects and assuming how they originated.

Issue #412, opened 2019-04-04
Problem with Vue and invocations (probably programmer error)


alfreema:
## Bug or feature request?

If there is an issue, it's a bug.

## Description:

On Firefox, I am receiving an error when trying to use invocations as described in the "Testing" section of the invocation documentation (https://xstate.js.org/docs/guides/communication.html#testing).

I DO NOT receive the service makes an actual call with axios to a a backend service -- so that tells me there is something wrong with how I am using promises, potentially.   But I am just following the example I found.

The machine/service/promise work on codepen, but I can easily reproduce it with a brand new vue project.

## (Bug) Expected result:

No error in a browser console.

## (Bug) Actual result:

I am receiving the following error in my console:

```
TypeError: "state is undefined"
    execute interpreter.js:172
    update interpreter.js:182
    start interpreter.js:283
    spawn interpreter.js:496
    exec interpreter.js:437
    execute interpreter.js:173
    execute interpreter.js:172
    update interpreter.js:182
    start interpreter.js:283
    created HelloWorld.vue:62
    VueJS 27
    <anonymous> main.js:6
    js app.js:3073
```

...

## (Bug) Potential fix:

I believe it has something to do with how the promise is working, because axios promises work, but the example promise from the Testing section causes the error to show up (but it does NOT show up in codepen for some reason).

## Link to reproduction or proof-of-concept:

To easily recreate the error with a brand new vue project, do this (assumes vue-cli 3):

1) vue create xstate-test	[I chose default (babel, eslint)]
2) cd xstate-test
3) npm install xstate@latest --save
4) Then overwrite the HelloWorld.vue component with this:

```
<template>
  <div class="home">
    Hello.
  </div>
</template>

<script>
import { Machine, interpret } from 'xstate';

// An example service that can be invoked by the state machine
const services = {
  getService: (ctx, event) => fetchApp(ctx.appId)
}

const fetchApp = async appId => {
  return {
    appId: appId,
    other: 'Other Data'
  }        
}

const machine = new Machine(
  {
    id: 'master',
    initial: 'init',
    strict: true,
    states: {
      init: {
        invoke: {
          id: 'serviceExample',
          src: 'getService',
          onDone: {
            target: 'test_state'
          }
        },
      },
      test_state: {}
    }
  },
  {
    services
  }
)

export default {
  name: 'home',

  data: () => ({
    service: null,
    current: null,
    context: {
      config: null
    }
  }),
    
  created: function() {
[...]

PR #415, opened 2019-04-07
Tweaked Promise support in interpreter to accept any thenable (aka PromiseLike)

johnyanarella:
* Added `isPromiseLike()` type guard to `utils`.
* Changed invoke/service type requirement from `Promise` to `PromiseLike`.
* Used `Promise.resolve()` to adopt untrusted thenables into trusted Promises.

Related: #412

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -27,7 +27,7 @@ import { State } from './State';
 import * as actionTypes from './actionTypes';
 import { toEventObject, doneInvoke, error } from './actions';
 import { IS_PRODUCTION } from './StateNode';
-import { mapContext, bindActionToState, warn } from './utils';
+import { isPromiseLike, mapContext, bindActionToState, warn } from './utils';
 import { Scheduler } from './scheduler';
 
 export type StateListener<TContext, TEvent extends EventObject> = (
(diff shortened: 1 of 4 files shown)


johnyanarella:
I haven't done much with Typescript - feel free to let me know if there's something I got wrong here! Ended up leveraging Typescript's built-in `PromiseLike` type.

Still pondering how best to introduce tests for this. Not keen on duplicating the entire "with promises" block, as that explodes the size of the test suite. Thoughts?

One thought I had was to wrap all the Promises in the existing tests with a wrapper that returned a PromiseLike view of them (i.e. just  exposed the `then()` function).

davidkpiano:
This looks good so far! I think your idea of wrapping existing tests would be fine, something like:

```js
describe('with promises', () => {
  // fn is (res, rej) => void
  const createRealPromise = (fn => new Promise(fn));
  const createPromiseLike = (fn => ({ then: /* ... */ })); // fake promise

  [createRealPromise, createFakePromise].forEach(createPromise => {
    // ...
    src: ctx => createPromise(resolve => { ... })

    // ...
    it('existing test etc.', () => {});
  });
});
```

johnyanarella:
Ugh. The indentation level change did a real number on the readability of the diff. 

Worth noting that there were previously a couple callback related tests that were in the promises section, moved the `describe('callbacks', ...` up to wrap those.

johnyanarella:
Also noticed an unrelated parent/child test (ping/pong machines) that looks like it snuck into the middle of the Promises tests. Moved it up and out.

PR #417, opened 2019-04-09
Improved simulated thenable in invoke test suite.

johnyanarella:
Previously, the simulated thenable would return a real Promise when `then()` was invoked. Now, `then()` returns a simulated thenable (and any subsequent `then()` will as well).


johnyanarella:
Related: #412

PR #431, opened 2019-04-20
Test created service against being a Machine before trying to spawn it

### LINKED ISSUE #430: Cannot read property initialState of undefined
## Bug or feature request?
Probably more of a feature request, but since I'm providing reproduction steps maybe it's really a bug?

## Description:

Given the following example state machine I get an error which doesn't make it obvious what the problem with my machine is. I believe it's just because I'm invoking a service that doesn't return a promise but it's unclear from the error message what is wrong.

Machine
```
const formMachine = Machine({
  id: 'form',
  initial: 'valid',

  states: {
    valid: {
      on: {
        SAVE: 'pending'
      }
    },

    pending: {
      invoke: {
        src: 'onSubmit',
        onDone: {
          target: 'success'
        }
      }
    },

    success: {
      type: 'final',
      actions: 'onDone'
    }
  }
},
{
  actions: {
    onDone: (context, event) => console.log('done', context, event)
  },

  services: {
    onSubmit: (context, event) => console.log('submit', context, event)
  }
});
```

## (Bug) Expected result:

Ideally this should return an error message like "service does not return a promise, callback function, or machine".

## (Bug) Actual result:

```
interpreter.js:354 Uncaught TypeError: Cannot read property 'initialState' of undefined
    at t.start (interpreter.js:354)
    at t.spawn (interpreter.js:622)
    at t.exec (interpreter.js:549)
    at interpreter.js:227
    at Array.forEach (<anonymous>)
    at t.execute (interpreter.js:226)
    at t.update (interpreter.js:238)
    at t.flush (interpreter.js:417)
    at StateChart.send (interpreter.js:150)
    at onClick (StateChartNode.js:146)
```

## (Bug) Potential fix:

When calling a service that we expect to return a promise, the library should throw an error when it sees that the service doesn't return a promise. It looks like the failure is around here: https://github.com/davidkpiano/xstate/blob/320166e426f549776dc7811cb6b3d11f84f202b5/src/interpreter.ts#L690 The logic there detects no promise or callback function so it seems to assume the service is a machine when in actuality the library user (me) made a stupid error from not understanding how xstate works. Instead a check should be made to determine if the service actually returned a machine and if not it should just throw an error to help the user fix the problem.

Andarist:
fixes #430

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -105,6 +105,7 @@ class StateNode<
   TStateSchema extends StateSchema = any,
   TEvent extends OmniEventObject<EventObject> = OmniEventObject<EventObject>
 > {
+  public isXStateNode: true = true;
   /**
    * The relative key of the state node, which represents its location in the overall state value.
    */
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -29,6 +29,7 @@ import { toEventObject, doneInvoke, error } from './actions';
 import { IS_PRODUCTION } from './StateNode';
 import {
   isPromiseLike,
+  isMachine,
   mapContext,
   bindActionToState,
   warn,
@@ -685,7 +686,7 @@ export class Interpreter<
             this.spawnPromise(id, Promise.resolve(source));
           } else if (isFunction(source)) {
             this.spawnCallback(id, source);
-          } else if (typeof source !== 'string') {
+          } else if (isMachine(source)) {
             // TODO: try/catch here
             this.spawn(
               data
@@ -696,6 +697,9 @@ export class Interpreter<
                 autoForward
               }
             );
+          } else if (typeof source !== 'string' && process.env.NODE_ENV !== 'production') {
+            // tslint:disable-next-line:no-console
+            warn(false, `Service does not return a promise, callback function, machine or a string.`);
           } else {
             // service is string
           }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,6 +5,7 @@ import {
   Action,
   EventObject,
   StateInterface,
+  StateMachine,
   PropertyMapper,
   Mapper,
   EventType,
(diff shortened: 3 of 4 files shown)


src/StateNode.ts
```diff
@@ -105,6 +105,7 @@ class StateNode<
   TStateSchema extends StateSchema = any,
   TEvent extends OmniEventObject<EventObject> = OmniEventObject<EventObject>
 > {
+  public isXStateNode: true = true;
```
Andarist: Why not relying on `instanceof`? Mainly to be safe when using frames, `instanceof` is not reliable in such scenarios.

As to the name - totally bikesheddable.

davidkpiano: What about using `Symbol()`?

Andarist: Symbol wouldnt work, we could use Symbol.for though. IMHO it wouldn't really change things much and it wouldn't be friendly for older environments where Symbol (and its global cross frame registry) is not available

davidkpiano: I'm wondering too if we can just do `source instanceof StateNode` - that would suffice.

Andarist: > Why not relying on instanceof? Mainly to be safe when using frames, instanceof is not reliable in such scenarios.

This is a bummer, but the scenario is possible - maybe it's not something popular to do, but there are valid use cases for this and `instanceof` wouldn't cover it.

Consider a parent page with src-less iframe holding a script tag which in turn creates a machine, but later that machine is being interpreted by a parent page (which is possible as you can synchronously access `window.parent` from such frame without being blocked).

This issue is often visible when manipulating the DOM and trying to use checks like `foo instanceof Element` which fails if an element was created in the other frame (`frame.Element !== window.Element`).

So even if it's not super popular we can just not worry about such things by using a simple property with a boolean value (something which i.e. React does too - a component's instance has `isReactComponent` property, https://overreacted.io/how-does-react-tell-a-class-from-a-function/ )

Andarist: @davidkpiano any more thoughts on this? 😃

PR #436, opened 2019-04-24
Tweak `invoke` object description with information about `data` being machine-specific.

Andarist:
…`data` being machine-specific.

### DIFF
--- a/docs/guides/communication.md
+++ b/docs/guides/communication.md
@@ -17,20 +17,20 @@ You can invoke:
 
 An invocation is defined in a state node's configuration with the `invoke` property, whose value is an object that contains:
 
-- `src` - the source of the machine to invoke, which can be:
+- `src` - the source of the service to invoke, which can be:
   - a machine
   - a string, which refers to a machine defined in this machine's `options.services`
   - a function that returns a `Promise`
   - a function that returns a "callback handler"
   - a function that returns an observable
 - `id` - the unique identifier for the invoked service
-- `autoForward` - (optional) `true` if all events sent to this machine should also be sent (or _forwarded_) to the invoked child machine (`false` by default)
-- `data` - (optional) an object that maps properties of the child machine's [context](./context.md) to a function that returns the corresponding value from the parent machine's `context`.
 - `onDone` - (optional) the [transition](./transitions.md) to be taken when:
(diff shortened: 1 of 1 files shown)


docs/guides/communication.md
```diff
@@ -14,16 +14,16 @@ You can invoke:
...
   - a string, which refers to a machine defined in this machine's `options.services`
   - a function that returns a `Promise`
   - a function that returns a callback handler, which sends events to the parent via the callback (example below)
 - `id` - the unique identifier for the invoked service
-- `forward` - (optional) `true` if all events sent to this machine should also be sent (or _forwarded_) to the invoked child machine (`false` by default)
-- `data` - (optional) an object that maps properties of the child machine's [context](./context.md) to a function that returns the corresponding value from the parent machine's `context`.
 - `onDone` - (optional) the [transition](./transitions.md) to be taken when the child machine reaches its [final state](./final.md)
-- `onError` - (optional) the transition to be taken when the child machine encounters an execution error.
+- `onError` - (optional) the transition to be taken when the child machine encounters an execution error
+- `forward` - (optional, used only for machines) `true` if all events sent to this machine should also be sent (or _forwarded_) to the invoked child machine (`false` by default)
```
davidkpiano:
```suggestion
- `forward` - (optional) `true` if all events sent to this machine should also be sent (or _forwarded_) to the invoked child (`false` by default)
```

This will also be sent to invoked callbacks (and duplex Observables in a future relase)

davidkpiano:
Looks good, thanks for doing this! Just one quick suggestion, and also we should look into `data` being used for resolved Promises as well as completed Observables in the future.

Andarist:
Thanks for the review - gonna adjust the description sometime later. 

I'm not sure what do you mean by "data being used for resolved Promises" though. Do you mean using `data` as output mechanism for resolved Promises? If so that would be somewhat cool (I've literally tried to do this today), but also kinda confusing - as in some cases data could be used as input mechanism and in others as output (truth to be told this is already the case with machines and their final states).

As to completed Observables - I'm not 100% sure about this but as far as I remember streams do not complete with a particular value, so not sure how exactly this would work.

Andarist:
@davidkpiano totally forgot about this PR, I've fixed the changes according to your suggestion now - should be OK to merge

PR #440, opened 2019-04-28
Fix State instance loosing it's context

farskid:
## The problem

Currently, since the `matches` and `toStrings` methods are added to the prototype of State, the context of `this` will be lost in case of reassignment or object destructing.

```js
const { matches } = machineState;

console.log(matches('some_state'));
```

Screenshot 2019-04-28 at 23 40 21


davidkpiano:
Thanks @farskid!! Can you write a quick test?

farskid:
@davidkpiano Tests added to the State test suite.

Issue #445, opened 2019-05-02
useMachine service not initialized correctly


GU5TAF:
## Bug or feature request?
Bug

## Description:
```
const [current, send, service] = useMachine(toggleMachine);
```
The `service` returned by `useMachine` is not initialised during the first render, triggering a render will show that `initialized` is now `true` but `service.state` still isn't equal to `current` until `send` has been called and the machine has transitioned.

## (Bug) Expected result:
For `service` to be initialised at first render and for`service.state` to have the same value as `current` when it's initialised.

## (Bug) Actual result:
`service.initialized` is `false`  and `service.state` is `undefined`.

## (Bug) Potential fix:
Fixed in https://github.com/davidkpiano/xstate/pull/446

## Link to reproduction or proof-of-concept:
[Link to reproduce bug](https://codesandbox.io/embed/mz96065x6j)
[Link to potential fix](https://codesandbox.io/embed/p5oo57nz7j)

davidkpiano:
I want to say this is expected, and you should also never rely on `service.state`, but instead listen for state changes by referencing `[ current, ... ]` or `service.onTransition(...)`.

The service should not start until the component is mounted.

GU5TAF:
The component is mounted but due to the async nature of `useEffect` it returns the service before it's had a chance to start the it.

I understand the intention but it is it really expected for the value of `current` in `useService` to always initialize as `undefined`? That's how I came across this behaviour in the first place.

```
// <Parent />
const [current, send, service] = useMachine(toggleMachine);
console.log(current) // returns expected state object

...

<Child />
const [current, send] = useService(props.service);
console.log(current) // returns undefined so always has to be guarded
```

GU5TAF:
I noticed that none of the [tests for `useService`](https://github.com/davidkpiano/xstate/blob/master/packages/xstate-react/test/useMachine.test.tsx#L137) use a service returned by `useMachine`. I might have incorrectly assumed that it should be possible to use them together in which case I think some clarification in the docs would be helpful.

davidkpiano:
The easiest workaround in your example is to check for the existence of `current` first. A service might not always be a machine; it can be something like an Observable or a pending promise, in which you're never guaranteed a value immediately.

I'll add this to the docs, but with `useService` the states you have to consider are:

- not yet started (no values yet)
- started
- stopped

GU5TAF:
Okay, thanks for the clarification 👍 

[...]

PR #446, opened 2019-05-02
fix: xstate-react useMachine service not initialized

GU5TAF:
```
const [current, send, service] = useMachine(toggleMachine);
```
The service returned by useMachine is not initialised during the first render, this PR starts the service as soon as it's created instead of doing it an effect.

### DIFF
--- a/packages/xstate-react/src/index.ts
+++ b/packages/xstate-react/src/index.ts
@@ -52,19 +52,14 @@ export function useMachine<TContext, TEvent extends EventObject>(
         setCurrent(state);
       }
     });
+    // Start the service as soon as it's created
+    serviceRef.current.start();
   }
 
   const service = serviceRef.current;
 
-  useEffect(() => {
-    // Start the service when the component mounts
-    service.start();
-
-    return () => {
-      // Stop the service when the component unmounts
-      service.stop();
-    };
-  }, []);
+  // Stop the service when the component unmounts
+  useEffect(() => () => service.stop(), []);
 
   return [current, service.send, service];
 }


davidkpiano:
This might have the unintentional side-effect of actions (and state transitions) happening before the component is mounted.

GU5TAF:
This might just be due to my understanding of hooks and xstate so feel free to call me out, but how would one go about triggering actions (and state transitions) before the component is mounted?

As far as I can tell, the only way to trigger an action/transition would be to use `send` or the `service` directly and the only way to do that before the component mounts would be to do it in the body of the functional component which is a big no-no.

davidkpiano:
@GU5TAF Is your use-case resolved? And what would be your use-case for wanting to start a service before the component is mounted? This would essentially be (almost) the same thing as just defining a running service externally, and using the `useService(...)` hook instead of the `useMachine(...)` hook.

GU5TAF:
> What would be your use-case for wanting to start a service before the component is mounted

Not quite, I want to start the service as it is mounted as opposed to the behaviour now where it starts it after it has mounted.

If I use the service returned by `useMachine` I have to add a pointless `init` transition as my root component mounts so that any children subscribing to that service will rerender and get a value for `current`. This extra transition combined with the fact that my entire app has to be rendered twice as it mounts breaks server side rendering and doesn't feel right.

As we discussed in #445, I currently have an adequate workaround where I create and start the service within the component and just avoid the `useMachine` hook.

In #445 you mentioned that a service might not always be a machine but according to the types for `useMachine(machine: StateMachine<TContext, any, TEvent>)` that doesn't appear to be the case for that hook?

PR #458, opened 2019-05-18
Add test for onEntry defined on parallel state

Andarist:
While reading SCXML spec I've found out that this is supposed to work, I then got curious if xstate implements it - and it does 😉 There was no test to cover this, so I've added one.


davidkpiano:
Much thanks!!

PR #461, opened 2019-05-19
Fix/error events

### LINKED ISSUE #427: Issue with onError in parallel states
## Bug or feature request?

Given a parallel state whose concurrent substates invoke some services, it appears that when you reject() a promise in order to fire onError in one substate, that XState is firing onError for ALL of the substates.

## Description:

I have a parallel state with three concurrent substates (Q1, Q2, Q3). Each state attempts to pop() an element off of an array. 

If the pop() is successful, the promise will resolve() and it will call the service's onDone.  onDone just transitions to itself so the substate can repeat popping elements over and over. 

When the array is empty, however, the pop() will fail and it should call onError. When onError is detected, it should transition to a 'final' state and wait there until the other Q's have also entered their 'final' state.

## (Bug) Expected result:

When ALL substates hit the 'final' state then the parallel state's onDone should fire as expected.

## (Bug) Actual result:

The parallel state's onDone is being triggered after only ONE substate has called onError.   

That tells me that onError is causing the other substates' onError to be fired as well.

## (Bug) Potential fix:

If it is intended that onError from one substate should cascade to other substates, then there is no sane way to implement retry logic.  onError should be handled by only the substate that fired it, and leave it to that substate to communicate an error that should cascade.

The behavior is reproduced in this sandbox:  [https://codesandbox.io/s/1owjq73763](https://codesandbox.io/s/1owjq73763)

All of the code is in src/components/HelloWorld.vue

The console demonstrates:  
1) When the first Q rejects() based on popping an empty array and calls onError, 
2) ALL of the other substates' onError trigger, and 
3) cause ALL substates to go to their 'final' state, which 
4) forces the final onDone to be called before the other substates are actually done.

We should not see "finish" before we see any "fetchQueue:Qn:slept=x".

### LINKED ISSUE #456: Invoked services sending wrong error type
## Description:

After reading SCXML specification (not all of it though) I believe that wrong type of error is being generated by the interpreter, i.e. [here](https://github.com/davidkpiano/xstate/blob/2877d3cb2b752c4899cb1a7302e9a37b94aa07e9/src/interpreter.ts#L830).

[This part of the spec](https://www.w3.org/TR/scxml/#errorsAndEvents) doesn't mention that `"error.execution"`  is related to "<invoke>", at least not in a direct way. From what I understand it should only get generated when evaluating all kinds of expressions/actions in a synchronous like manner.

Similarly [`<invoke>`](https://www.w3.org/TR/scxml/#invoke) doesn't mention errors at all, which is kinda given as machines/services communicate only through message passing. 

It seems to me that in fact `error.platform` should get generated (described [here](https://www.w3.org/TR/scxml/#ErrorEvents)), as the interpreter can't assume the nature of invoked service and the other possibility would be `error.communication`, but I dont think it suits all use cases.

## (Bug) Expected result:

`error.platform` type (preferrably suffixed with ID) for most of the current service-related errors instead of `error.execution`

## (Bug) Actual result:

`error.execution` being sent.

## Implementation

If you agree with me I could explore implementing a fix for this issue along with fixing https://github.com/davidkpiano/xstate/issues/427 . The whole thing would be a breaking change for xstate though.

Andarist:
This is very much still a WIP and something to be discussed, but I wanted to get the ball rolling. 

Currently `onError` get called for ANY `error.execution` - which might happen at any level of connected or parallel machines, which is IMHO very bad and unsafe because people expect it to only be called for a failure on a particular level of the machine (where defined).

Also - the PR changes `error.execution` to `error.platform` for reasons mentioned in #456 . If I understand correctly `error.execution` has defined role in SCXML and is currently used not in line with it.

fixes #427 closes #456


Andarist:
@davidkpiano any thoughts on this one?

davidkpiano:
It makes sense and I'm willing to merge it - can you resolve conflicts first? 🙏 Thanks!

Andarist:
@davidkpiano I've rebased this so it's mergeable now. Please review the logic, error event shape etc - I'm not 100% sure if I covered all intended use cases.

PR #467, opened 2019-05-28
Fix logic around actors subscriptions

Andarist:
Seems to me that this is more correct - at least in other systems. If you believe this change is wrong then feel free to close it, but I believe that **at least** next/complete/handleError shouldn't be called after subscriber unsubscribe itself - either based on newly introduced  `unsubscribed` flag or based on existing `canceled` (if you decide that subscribers should be able to stop the whole service).

If you agree with this one, I'm going to write tests & look into other service types to cover this in the same manner.

The interesting question is if subscribers should be notified after service has been stopped?

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -847,13 +847,13 @@ export class Interpreter<
   private spawnPromise(promise: Promise<any>, id: string): Actor {
     let canceled = false;
 
-    promise
-      .then(response => {
+    promise.then(
+      response => {
         if (!canceled) {
           this.send(doneInvoke(id, response));
         }
-      })
-      .catch(errorData => {
+      },
+      errorData => {
         if (!canceled) {
           const errorEvent = error(errorData, id);
           try {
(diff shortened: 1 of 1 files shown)


src/interpreter.ts
```diff
@@ -878,15 +878,21 @@ export class Interpreter<
       id,
       send: () => void 0,
       subscribe: (next, handleError, complete) => {
+        let unsubscribed = false;
         promise
           .then(response => {
+            if (unsubscribed) return;
             next && next(response);
+            if (unsubscribed) return;
```
davidkpiano: What's the use-case for this line?

Andarist: It's handling a case of consumer unsubscribing synchronously in the "next" handler - taking RxJS as example https://codesandbox.io/s/determined-rgb-bwdnn . "complete" callback should not be called in this case

davidkpiano:
This LGTM, we can merge it in when we have tests, yeah?

Andarist:
Sure, gonna add tests and apply same logic to other actor types too

Andarist:
@davidkpiano I believe I have fixed few issues around all actor types being subscribed to & unsubscribed to.

However, I'm not sure how & when `actor.subscribe` is supposed to be called, what are use cases? I would like to write tests which would present valid real-life use cases, but not sure what those are at the moment.

Is accessing a stored actor ref in context and calling .subscribe on it considered a valid use case? There are no docs nor tests around it right now, hence my question.

davidkpiano:
> Is accessing a stored actor ref in context and calling .subscribe on it considered a valid use case? There are no docs nor tests around it right now, hence my question.

It's more of a user-land thing, and actors can implement their own subscription mechanisms, which is why no tests are written yet. The primary use-case is that actors will not be able to be subscribed to (unless subscription to actor is explicitly provided).

Andarist:
> It's more of a user-land thing, and actors can implement their own subscription mechanisms, which is why no tests are written yet. The primary use-case is that actors will not be able to be subscribed to (unless subscription to actor is explicitly provided).

I don't quite understand this 😅 Would you be able to demonstrate with some pseudo-code?

PR #472, opened 2019-05-29
WIP: Separate stop and nonstop actions from exitActions

farskid:
Fixes https://github.com/davidkpiano/xstate/issues/470

## Why?

In case transition actions need to `send` events to children which are invoked services/machines.

## Solution

Since the action execution order is `exit actions -> transition actions -> entry actions`, when the interpreter tries to execute transition actions, the child is already stopped in exit actions, therefore, `sendTo` will fail.

This solution separates exit actions of type `xstate.stop` from the rest and changes the order from `rest of the exit actions -> transition actions -> exit actions of type xstate.stop -> entry actions`. This way, invoked children will be alive for transition actions and will stop after those are executed. Since stopping the invoked services is an internal act, it should not affect user-specific actions as long as the user follows the old order.

@davidkpiano Let me know what you think about this.

[ ] Write a test for the bug
[ ] Fix tests

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -1000,8 +1000,31 @@ class StateNode<
       )
     ];
 
+    // Separating actions of type xstate.stop from the rest
+    // Why? transition actions can have `send` actions to specific invoked
+    // service children, so invoked services need to stay alive until the
+    // transition.are executed
+    // https://github.com/davidkpiano/xstate/issues/470
+    const stopExitActions: Array<
+      ActionObject<any, OmniEventObject<EventObject>>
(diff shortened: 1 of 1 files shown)


farskid:
Noted that the mentioned issue is already closed as invalid. If this is an invalid usage of invoked services, we need to add a warning in the docs about this.
Although I still think this doesn't violate the spec and can be a nice to have.

davidkpiano:
Yeah, unfortunately, this PR would just "resolve" the race condition by allowing the sending of an event to a service and that service being stopped to "overlap" for a split second.

PR #473, opened 2019-05-30
Fix typo in docs

kaykayehnn:
In the states guide, the Persisting State example references a restoredStateDef variable which is not defined.
restoredStateDef used to be the result of parsing the persisted state definition but in #393 the example was updated to use State.create and the resolveState call remained referencing the old declaration. This PR fixes the inconsistency.

Something to consider is that now `restoredState` and `resolvedState` are very similar variable names and this might be a little bit confusing. Making restoredState more verbose (restoredStateDef or restoredStateDefinition) to emphasise it is the state *definition*, not the resolved state, could make it clearer IMO.


davidkpiano:
How about `previousState` instead of `restoredState` then?

PR #476, opened 2019-06-01
Adding .run() method to interpreter

gabewillen:
Adding .run() method to interpreter to keep the process alive.

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -397,7 +397,7 @@ export class Interpreter<
       if (!this.initialized) {
         this.start(initialState);
       }
-      setImmediate(this.idle.bind(this));
+      setTimeout(this.idle.bind(this), 0);
       return this;
   }
   /**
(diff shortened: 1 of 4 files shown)


gabewillen:
I'm looking into how I can add a test to determine that this is actually keeping the process alive.  If anyone has an idea let me know. Also thinking about adding a `.onIdle(...)` method similar to `.onDone()` that is called when the interpreter is idling.

src/interpreter.ts
```diff
@@ -388,6 +388,18 @@ export class Interpreter<
...
+
+  /**
+   * Starts the interpreter keeping the current process running
+   * @param initialState The state to start the statechart from
+   */
+  public run(initialState?: State<TContext, TEvent> | StateValue): Interpreter<TContext, TStateSchema, TEvent> {
+      if (!this.initialized) {
+        this.start(initialState);
+      }
+      setImmediate(this.idle.bind(this));
```
davidkpiano: This won't work for browser environments, by the way

gabewillen: Updated this to use setTimeout(..., 0) instead.

davidkpiano:
I think it would be better to provide guidance on how to do this in the docs rather than baking it into the interpreter, for now. Thank you, though!

PR #478, opened 2019-06-01
Fix potential memory leak caused by not always cleaning up children i…

Andarist:
…nside interpreter

Activities [might not have a `stop` function](https://github.com/davidkpiano/xstate/blob/709947a71b5f557399b5814e4b61ab4a8c2af60c/src/interpreter.ts#L1019) on them (EDIT:// invoked callbacks also have optional stop), so it seems to me that this has caused a minor memory leak.

There are no public APIs to query children (or are there?) so I've skipped writing a test for this (no idea how this could be tested right now).

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -783,10 +783,13 @@ export class Interpreter<
   }
   private stopChild(childId: string): void {
     const child = this.children.get(childId);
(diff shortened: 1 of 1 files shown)


davidkpiano:
I'll make children a public API, thanks!

PR #484, opened 2019-06-05
useService: explain the additional state update that might look unnecessary

hnordt:
Related to https://github.com/davidkpiano/xstate/commit/4f2701c566d4492e3e639cef1012021835c756cd#r33762426

### DIFF
--- a/packages/xstate-react/src/index.ts
+++ b/packages/xstate-react/src/index.ts
@@ -74,9 +74,6 @@ export function useService<TContext, TEvent extends EventObject>(
 
   useEffect(
     () => {
-      // Set to current service state
-      setCurrent(service.state);
-
       const listener = state => {
         if (state.changed) {
           setCurrent(state);
--- a/packages/xstate-react/src/index.ts
+++ b/packages/xstate-react/src/index.ts
@@ -74,6 +74,11 @@ export function useService<TContext, TEvent extends EventObject>(
 
   useEffect(
     () => {
+      // Set to current service state as there is a possibility
+      // of a transition occurring between the initial useState()
+      // initialization and useEffect() commit.
+      setCurrent(service.state);
+
       const listener = state => {
         if (state.changed) {
           setCurrent(state);


johnyanarella:
The current React doc examples seem to agree with your change. Always warms my heart to see lines of code *deleted*.

That said, I remember someone speaking up in some earlier thread about there being a possibility of a transition occurring between the initial `useState()` initialization and when `useEffect()` runs (when committed) and a listener is attached. (Potentially related to Concurrent mode?)

I also think I recall this coming up elsewhere with `useEffect()` when listening to browser APIs that emit events independent of the React lifecycle (ResizeObserver, IntersectionObserver, etc.).

Probably worth investigating to make sure there's not a possibility of missing an event from an independent actor. Not really sure what you're supposed to do in that situation, though. Agree that always blindly calling setState() seems redundant (maybe a comparison first?).

johnyanarella:
Will try to carve out some time tomorrow to experiment in a sandbox to see if I can replicate that concern. If you figure out the answer before me, definitely intrigued to learn the answer since it affects hooks we write outside XState, too.

hnordt:
@johnyanarella you have a good point. 

But I think a better fix for the possibility of a transition occurring between the initial `useState()` initialization and `useEffect()` itself would be taking advantage of `useEffect` deps. Maybe we should include `service.state` in `useEffect()` deps?

```js
useEffect(
    () => {
      const listener = state => {
        if (state.changed) {
          setCurrent(state);
        }
      };

      listener(service.state);

      service.onTransition(listener);

      return () => {
        service.off(listener);
      };
    },
    [service, service.state]
  );
```

hnordt:
@johnyanarella you have a good point. I've reverted my last change and added a comment based on your explanation.

davidkpiano:
I'll merge this comment 😂 thanks

johnyanarella:
Ha! Good idea on the comment, @hnordt . Otherwise, one of us was definitely going to look at this a month from now and wonder the same thing again.

hnordt:
@davidkpiano sometimes a comment worth more than 1000 console.logs 😂.

PR #490, opened 2019-06-10
InvokeCreator: Improving types

asbermudez:
After being using heavily for the last months this project, this particular type has been particulary annoying as when you're defining the initialization the service call typing gets lost.

This modification allows to keep the possibility of using the type expansively, but does set a default type that makes more sense than `any`. 

I hope you find this improvement useful.

**ATTENTION:** For this change I had to change the order of the generics, which can lead to breaking changes on the types if this is merged.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -192,7 +192,7 @@ export type InvokeCallback = ((
  * @param context The current machine `context`
  * @param event The event that invoked the service
  */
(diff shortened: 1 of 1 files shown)


davidkpiano:
This is a good improvement, thanks!

I wouldn't consider type changes like this breaking changes necessarily, since 1) the compiler will warn you if any existing code is incompatible anyway, and 2) the API doesn't change.

PR #497, opened 2019-06-16
Support autoForward option in spawned actors

### LINKED ISSUE #494: Allow autoForwarding of events to spawned actors
## Bug or feature request?
Feature
## Description:
Since `spawn` essentially wraps an `invoke` element, allow passing in the attribute autoForward so that events can be sent to spawned actors.  

Should be fairly straight forward and If I get time i'll submit a pull request for this feature.  In the mean time this will be a reminder.
```js
spawn(SomeMachine, {autoForward: true});
```
- Is this is a breaking change
No
It is part of the `invoke` specification and currently implemented  with the `invoke` property of a `state`.
You can use [CodePen](codepen.io) or [Code Sandbox](codesandbox.io) to create a reproduction/POC.

Andarist:
closes #494 

Not sure how you feel about changing public `forward` option to `autoForward`. I thought it might be a good change because:
- it aligns with SCXML
- it also aligns to what it was converted to under the hood [here](https://github.com/davidkpiano/xstate/blob/167fd5b37fe320f3101b09ac917810bc79c3c475/src/interpreter.ts#L715)

I can easily revert this back and just make `forward` supported in both APIs.

### DIFF
--- a/docs/guides/communication.md
+++ b/docs/guides/communication.md
@@ -24,7 +24,7 @@ An invocation is defined in a state node's configuration with the `invoke` prope
   - a function that returns a "callback handler"
   - a function that returns an observable
 - `id` - the unique identifier for the invoked service
-- `forward` - (optional) `true` if all events sent to this machine should also be sent (or _forwarded_) to the invoked child machine (`false` by default)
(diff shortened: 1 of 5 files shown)


gabewillen:
@Andarist Thank you!

Andarist:
@davidkpiano fixed the test description, is this otherwise ready to be merged? or is your CR not finished yet?

PR #503, opened 2019-06-22
Tweak assign typings to allow TContext inferring for parameterless assignment function

Andarist:
Previously this https://github.com/davidkpiano/xstate/blob/6eea481d82edf448cec9444ecde467e53d573be6/test/actor.test.ts#L407-L409 
was inferred to `assign<{ serverRef: Actor }>` which is not compatible with the desired inferred type of `assign<{ counter: number; serverRef?: Actor }>`.

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -24,7 +24,7 @@ import {
   SendExpr,
   SendActionObject,
   OmniEventObject,
-  PureAction
+  PureAction,
 } from './types';
 import * as actionTypes from './actionTypes';
 import { getEventType, isFunction, isString } from './utils';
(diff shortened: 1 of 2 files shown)


src/actions.ts
```diff
@@ -302,9 +302,19 @@ export function stop<TContext, TEvent extends EventObject>(
...
  */
-export const assign = <TContext, TEvent extends EventObject = EventObject>(
+export function assign<TContext, TEvent extends EventObject = EventObject>(
+  // trick to make TS inferring TContext based on return type of `assignment` instead of its arguments (microsoft/TypeScript#31933)
+  assignment: TContext extends any ?
+    (Assigner<TContext, TEvent> | PropertyAssigner<TContext, TEvent>)
+    : never
+): AssignAction<TContext, TEvent>;
+export function assign<TContext, TEvent extends EventObject = EventObject>(
+  // second overload to make this work better in generic contexts (where TS has problems with resolving conditional types)
```
Andarist: without this TS was not able to allow this:
https://github.com/davidkpiano/xstate/blob/6eea481d82edf448cec9444ecde467e53d573be6/src/scxml.ts#L77
even though it should - because TContext extends here object and object extends any, but TS can't propagate this information right now through conditional types.

Andarist:
@davidkpiano friendly 🏓

davidkpiano:
Can you merge master into this and try it? I think there's just one error left on `actor.test.ts`

Andarist:
I would swear that it has worked when I've sent this PR, but seems like I haven't tested it thoroughly in the actual codebase (made it work on TS playground - but in much more simplified scenario). 

Damn, would be sweet to have this working - gonna revisit this in the future, but for now I'm closing this.

davidkpiano:
Sounds good. The types around `assign(...)` are something that I'd like to prioritize simplifying.

PR #513, opened 2019-06-27
Implemented 3rd argument for `useMachine`: `onStop`

amitnovick:
I modelled a process for subscribing and unsubscribing from a long-running service as a state machine.
However, when integrating the machine with a React component, the unmount lifecycle was problematic with `@xstate/react` 's `useMachine`:

A warning appeared on the console notifying me that an event has been sent to a service that isn't running, and that I need to `service.start()` first before I can send the event. Basically the service was already stopped by [`useMachine`'s `React.useEffect` implementation](https://github.com/davidkpiano/xstate/blob/master/packages/xstate-react/src/index.ts#L75).

I tried to solve that in user-space by using `React.useEffect` (see [Attempt](https://github.com/amitnovick/todo-app/blob/public/src/containers/Todos/TodosContainerCloud/TodosContainerCloud.js#L178)) but I noticed that the event is not being sent to the service.

I added an example to the `README` that shows a similar scenario for unsubscribing from an interval event.

This is the easiest way I could think of to implement cleanup of the subscription.
That said, **it's not an elegant solution in my opinion**.
It's also possible to add an `onStart` argument (for completeness?), a function that would be called after the service has started.

Please share your thoughts on this! :hand:

### DIFF
--- a/packages/xstate-react/README.md
+++ b/packages/xstate-react/README.md
@@ -152,9 +152,9 @@ const Fetcher = ({ onResolve }) => {
 };
 ```
 
-## Provide a cleanup callback on Unmount lifecycle
+## Provide a Cleanup Callback to be Executed Before Service Stops
 
-The `onStop` function argument allows to perform a cleanup of subscriptions before the component unmounts.
+The `onStop` callback argument allows to perform a cleanup of subscriptions before the service is stopped.
 
 ```js
 const timerMachine = Machine({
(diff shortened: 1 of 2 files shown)


davidkpiano:
Love the documentation, but as far as implementation, we should reduce the number of arbitrary arguments we pass into `useMachine(...)` (I like 1 - 2 arguments for simplicity).

Thankfully, the unsubscribe pattern you're trying to implement is already doable with `useMachine()` by referencing the service:

```js

const [current, send, service] = useMachine(someMachine);

// Register stop listener once
useEffect(() => {
  service.onStop(() => sendFirestore('UNSUBSCRIBE'));
}, []);
```

Might be helpful to instead add the above ☝️ to the documentation.

PR #514, opened 2019-06-27
Added example of `service.onStop` for unsubscribing

amitnovick:
Reference: [unsubscribe pattern](https://github.com/davidkpiano/xstate/pull/513#issuecomment-506529280)

### DIFF
--- a/packages/xstate-react/README.md
+++ b/packages/xstate-react/README.md
@@ -151,6 +151,68 @@ const Fetcher = ({ onResolve }) => {
 };
 ```
 
+## Provide a Cleanup Callback to be Executed Before Service Stops
+
+In order to send an event to be executed before service stops, we can use the returned `service` for its' `onStop` method which allows to perform a cleanup of subscriptions before the service is stopped.
+
+```js
+const timerMachine = Machine({
+  id: 'timer',
+  initial: 'uninitialized',
+  context: {
+    intervalRef: 0, // just an initial value
+    secondsElapsed: 0
+  },
+  states: {
+    uninitialized: {
+      on: {
+        INITIALIZE: {
+          target: 'initialized',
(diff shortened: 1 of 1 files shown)


packages/xstate-react/README.md
```diff
@@ -151,6 +151,68 @@ const Fetcher = ({ onResolve }) => {
...
+
+  React.useEffect(() => {
+    const intervalRef = setInterval(() => send('TICK'), 1000);
+    send('INITIALIZE', {
+      intervalRef: intervalRef
+    });
+  }, [send]);
+
+  React.useEffect(() => {
+    service.onStop(() => send('STOP'));
```
davidkpiano: This would send a `'STOP'` event right when the service is being stopped... Might be better to tweak this example to stop an _external_ sort of service/subscription?

amitnovick: > This would send a `'STOP'` event right when the service is being stopped
Well the point of the example is that the interval should be turned off when the component unmounts.
Now obviously this could be done with a simple:
```js
const [current, send] = useMachine(timerMachine);

React.useEffect(() => {
  const intervalRef = setInterval(() => send('TICK'), 1000);
  return () => clearInterval(intervalRef);
}, [send])
```
However that doesn't capture the intention of the intended behaviour (clock ticking process should be stopped).

amitnovick: > Might be better to tweak this example to stop an _external_ sort of service/subscription?

Do you have an example for this? not sure what you mean by **external** here.

davidkpiano: > Do you have an example for this?

Yes, the original example in your issue 😉

amitnovick: Completely lost you :confused:

amitnovick:
Closed for now, might review the idea of presenting useful patterns for the `@xstate/react` package sometime in the future.

PR #516, opened 2019-07-01
Make few internal things compatible with TS strict setting

Andarist:
More to come in the future, those were just the easiest ones to fix - others need more codebase exploration on my side.


src/utils.ts
```diff
@@ -538,7 +538,7 @@ export function isObservable<T>(
 }
 
 export function isMachine(
-  value: StateMachine<any, any, any> | any
+  value: any
```
Andarist: `whatever | any === any`

PR #526, opened 2019-07-09
Fix entry/exit to onEntry/onExit

chrisdhanaraj:
Was just going through the docs (and copy pasting over to the visualizer) and noticed this small bug


davidkpiano:
Hey Chris! This is actually correct; the current visualizer has an older version of XState running, and the new one (being released this week) will have the version that uses `entry`/`exit`.

chrisdhanaraj:
Ahhhh okay! Sounds good :+1:

PR #536, opened 2019-07-12
Invoke data expressions on nested final nodes

### LINKED ISSUE #517: data property not getting invoked on final states of hierarchical machines
## Bug or feature request?

Not sure, maybe just a question.

## Description:

When a machine reaches it's final state it can pass information to the `onDone` handler via the `data` property. When we use `invoke` the `data` property of the child machine can be a mapper that gets invoked with `context` and `event` - returning a `{[key: string]:any}`, but for simple hierarchical machines this is not the case.

```
const machine = Machine({
  initial: 'secret',
  context: {
    revealedSecret: undefined
  },
  states: {
    secret: {
      initial: 'wait',
      states: {
        wait: {
          after: {
            1000: 'reveal'
          }
        },
        reveal: {
          type: 'final',
          data: {
            secret: () => 'the secret'
          }
        }
      },
      onDone: {
        target: 'success',
        actions: assign({
          revealedSecret: (context, event) => {
            // event is:
            // { type: 'done.invoke.secret', data: { secret: () => 'the secret' } }
            return event.data.secret;
          }
        })
      },
    },
    success: {}
  }
});

  .onTransition(state => console.log(state.context))
  .start();
```

## (Bug) Expected result:

`revealedSecret === 'secret'` when we're in `success`

## (Bug) Actual result:

`revealedSecret` contains the function itself instead of the result of invoking it.

## (Bug) Potential fix:

Mirror the use of the `data` property from `invoke`

[Example on codesandbox illustrating the problem](https://codesandbox.io/s/xstate-example-template-vanilla-xhhff?fontsize=14) - The example shows how this works with separate machines _and_ how this differs from passing data from hierarchical state nodes. Results are visible in the console.

Andarist:
fixes #517

⚠️ this PR has one failing test, I haven't "fixed" it or change it because I'd like to discuss it first.

In this PR I make changes to the `getDoneEvents` algorithm which I believe fixes few things (according to the SCXML spec). The previous version was traversing the whole tree looking for freshly done states (based on states which we enter) - and it was generating done events for each such state, which means:
- it was generating it for leaf final nodes (it's always "done" after all)
- it was generating it for the top level node (machine) itself
- it was all done in one pass, so when we entered a really nested final node synchronously we had ended up with many done events generates

Looking at the spec - [`<final>`](https://www.w3.org/TR/scxml/#final) and  [enterStates procedure](https://www.w3.org/TR/scxml/#enterStates) we can read that:
```
When the state machine enters the <final> child of a <state> element, the SCXML Processor must generate the event done.state.id after completion of the <onentry> elements, where id is the id of the parent state. Immediately thereafter, if the parent <state> is a child of a <parallel> element, and all of the <parallel>'s other children are also in final states, the Processor must generate the event done.state.id where id is the id of the <parallel> element.

When the state machine reaches the <final> child of an <scxml> element, it must terminate. See D Algorithm for SCXML Interpretation for details.
```
and
```
// in enterStates 

if isFinalState(s):
    if isSCXMLElement(s.parent):
        running = false
    else:
        parent = s.parent
        grandparent = parent.parent
        internalQueue.enqueue(new Event("done.state." + parent.id, s.donedata))
        if isParallelState(grandparent):
            if getChildStates(grandparent).every(isInFinalState):
                internalQueue.enqueue(new Event("done.state." + grandparent.id))
```

According to that if we enter a final state we should generate at most 2 events (one for its parent - excluding top level state - and optionally one for the grandparent if its paralellel and if all its children are done already) for each final entered.

We can ofc have to generate more done events if we enter deeply nested final state which in turn ends substantial part of its ancestors, but seems like this is done in separate passes of the internalQueue processing.

I'm having slight doubts as this can lead to a weird situation with generating a done event for parallel states multiple times in a single pass if we enter all their children immediately in their final states (each state generates its own such event according to the algorithm and I cant find anything about deduplicating those events), but this doesn't change from the previous version.
<details>
  <summary>Example</summary>

```js
const machine = Machine({
  id: "my_test",
  initial: "initial",
  states: {
    initial: {
      after: {
        10: "secret"
      }
    },
    secret: {
      type: "parallel",
      states: {
        first: {
          initial: "a",
          states: {
            a: {
              type: "final"
            }
          }
        },
        second: {
          initial: "b",
          states: {
            b: {
              type: "final"
            }
          }
        }
      },
      onDone: {
        target: "success",
        actions: () => {
          console.log("parallel on done!");
        }
      }
    },
    success: {
      type: "final"
    }
  }
});
```
</details>

with such generated events:
```json
[{"type":"done.state.my_test.secret"},{"type":"done.state.my_test.secret.first"},{"type":"done.state.my_test.secret"},{"type":"done.state.my_test.secret"},{"type":"done.state.my_test.secret.second"},{"type":"done.state.my_test.secret"}]
```

Let's go back to the broken test - https://github.com/davidkpiano/xstate/blob/8222f4bfd4cf86ebb06a6892fc35fa130694f0d0/test/final.test.ts#L109-L144

After this change actual result change from `['bazAction', 'barAction', 'fooAction']` to `['bazAction', 'fooAction', 'fooAction']`.

1. The fact that `barAction` is not executed seems like a bugfix change - as final nodes dont generate done events for themselves so `onDone` on a final state should just not be allowed.

2. The fact that we execute `fooAction` twice also seems to make sense - as we first enter a final node (generating `done.state.foo` and thus firing `onDone` there) and later we transition from this final node to a different final node of that very same state and we make `foo` done once again, so seems OK that `fooAction` fires twice here.

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -22,7 +22,8 @@ import {
   toGuard,
   isMachine,
   toSCXMLEvent,
-  toEventObject
+  toEventObject,
+  mapContext
 } from './utils';
 import {
   Event,
@@ -942,6 +943,8 @@ class StateNode<
 
   private getActions(
     transition: StateTransition<TContext, TEvent>,
+    currentContext: TContext,
+    eventObject: TEvent,
     prevState?: State<TContext>
   ): Array<ActionObject<TContext, TEvent>> {
     const prevConfig = prevState
@@ -972,15 +975,25 @@ class StateNode<
     const doneEvents = flatten(
       transition.entrySet.map(sn => {
         const events: DoneEventObject[] = [];
-        if (sn.type === 'final') {
-          events.push(
-            done(sn.id, sn.data), // TODO: deprecate - final states should not emit done events for their own state.
-            done(sn.parent!.id, sn.data)
-          );
+
+        if (sn.type !== 'final') {
+          return events;
         }
 
-        if (sn.parent && sn.parent.parent) {
-          const grandparent = sn.parent.parent;
+        const parent = sn.parent!;
+
+        events.push(
+          done(sn.id, sn.data), // TODO: deprecate - final states should not emit done events for their own state.
+          done(
+            parent.id,
+            sn.data
+              ? mapContext(sn.data, currentContext, eventObject)
+              : undefined
+          )
+        );
+
+        if (parent.parent) {
+          const grandparent = parent.parent;
 
           if (grandparent.type === 'parallel') {
             if (
@@ -992,6 +1005,7 @@ class StateNode<
             }
           }
         }
+
         return events;
       })
     );
@@ -1121,7 +1135,12 @@ class StateNode<
       : undefined;
     const currentContext = currentState ? currentState.context : context;
     const eventObject = _eventObject || ({ type: ActionTypes.Init } as TEvent);
-    const actions = this.getActions(stateTransition, currentState);
+    const actions = this.getActions(
+      stateTransition,
+      currentContext,
+      eventObject,
+      currentState
+    );
     const activities = currentState ? { ...currentState.activities } : {};
     for (const action of actions) {
       if (action.type === actionTypes.start) {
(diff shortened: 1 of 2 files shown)


Andarist:
@davidkpiano I've rebased this against master (implementation much simpler thanks to your changes around removing StateTree!) - should be ready to merge

PR #538, opened 2019-07-12
Improve Actor docs

hnordt:


docs/guides/actors.md
```diff
@@ -356,7 +356,7 @@ someService.onTransition(state => {
 });
 ```
 
-By default, `sync` is set to `false`. Even when `sync` is set to `false` (both explicitly or implicitly), you'll be able to read the actor's state, e.g. `actorRef.state`, but you should rely on that state only when `sync` is set to `true`, otherwise you might end up referencing stale state.
+By default, `sync` is set to `false`. Even when `sync` is set to `false` (both explicitly or implicitly), you'll be able to read the actor's state, e.g. `actorRef.state`. You shouldn't rely on that state when `sync` is disabled, otherwise you might end up referencing stale state.
```
Andarist: Could we forbid such reads altogether? 🤔 Making it impossible to read such state would avoid people shooting themselves in the foot

hnordt: I agree!

davidkpiano: That was the previous behavior but it caused some hard-to-track bugs. I'd like to revisit it soon, but for now, having that note in the documentation is good enough.

Let's make the messaging stronger:

```suggestion
By default, `sync` is set to `false`. Never read an actor's `.state` when `sync` is disabled; otherwise, you will end up referencing stale state.
```

PR #543, opened 2019-07-13
Upgrade to TS@3.5

Andarist:
This PR is not ideal - I had to force cast a few things to make it work, technically it doesn't regress anything though, because those places were already broken (TS-wise). At the same time I've managed to fix some minor issues.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules/typescript/lib"
+}
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -61,7 +61,8 @@ import {
   MachineConfig,
   PureAction,
   TransitionTarget,
-  InvokeCreator
+  InvokeCreator,
+  StateMachine
 } from './types';
 import { matchesState } from './utils';
 import { State, stateValuesEqual } from './State';
@@ -857,7 +858,7 @@ class StateNode<
       : [];
 
     let nextStateStrings: TransitionTarget<TContext> = [];
-    let selectedTransition: TransitionDefinition<TContext, TEvent>;
+    let selectedTransition: TransitionDefinition<TContext, TEvent> | undefined;
 
     for (const candidate of candidates) {
       const { cond, in: stateIn } = candidate;
@@ -908,7 +909,7 @@ class StateNode<
           selectedTransition! && state.value // targetless transition
             ? new StateTree(this, path(this.path)(state.value)).absolute
             : undefined,
-        transitions: [selectedTransition],
+        transitions: selectedTransition ? [selectedTransition] : [],
         entrySet:
           selectedTransition && selectedTransition.internal ? [] : [this],
         configuration: selectedTransition && state.value ? [this] : [],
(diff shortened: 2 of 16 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
Andarist: VSC is using its own copy of TS which bite me for a moment - we always want it to use the installed one. Even though it's an editor config file (which doesnt normally belong to git) I find it highly useful to include it so it doesnt trip anyone in the future.

src/StateNode.ts
```diff
@@ -908,7 +909,7 @@ class StateNode<
           selectedTransition! && state.value // targetless transition
             ? new StateTree(this, path(this.path)(state.value)).absolute
             : undefined,
-        transitions: [selectedTransition],
+        transitions: selectedTransition ? [selectedTransition] : [],
```
Andarist: is this `transitions` field supposed to be used anywhere? I've removed it entirely locally and all tests pass, so maybe we could remove it altogether?

src/StateNode.ts
```diff
@@ -2012,7 +2012,7 @@ class StateNode<
       TEvent
     > = mapValues(
       { ...onConfig, ...doneConfig, ...invokeConfig },
-      (value, event) => {
+      (value: TransitionConfig<TContext, TEvent> | string | StateMachine<any, any, any> | undefined, event) => {
```
Andarist: this is sort of a cast from unknown, unfortunately invokeConfig is typed as any atm so had to cast this somewhere as fixing invokeConfig type was not trivial in a short period of time

src/interpreter.ts
```diff
@@ -845,13 +845,13 @@ export class Interpreter<
 
     if (resolvedOptions.sync) {
       childService.onTransition(state => {
-        this.send(actionTypes.update, { state, id: childService.id });
+        this.send(actionTypes.update as any, { state, id: childService.id });
```
Andarist: seems like OmniEvent should include platform errors, done events and special update action

src/types.ts
```diff
@@ -51,9 +51,8 @@ export type Event<TEvent extends EventObject> = TEvent['type'] | TEvent;
  * as well as the built in event types and/or objects.
  */
 export type OmniEvent<TEvent extends EventObject> =
-  | TEvent['type']
-  | BuiltInEvent<TEvent>['type']
-  | OmniEventObject<TEvent>;
+  | OmniEventObject<TEvent>
```
Andarist: this has fixed toEventObject calls, both versions (before and after) should be semantically equivalent, but the previous gave TS hard time (for some reason 🤷‍♂ ) to infer TEvent inside of `toEventObject`

test/machine.test.ts
```diff
@@ -218,7 +218,7 @@ describe('machine', () => {
     });
 
     it('should resolve the state value', () => {
-      const tempState = State.from('foo');
+      const tempState = State.from<any>('foo');
```
Andarist: this is needed now as unparameterized generics are now unknown by default (instead of any)

src/StateNode.ts
davidkpiano: It will be; the idea is to eventually remove `tree`, and use `transitions`. This will also be useful for debugging.

PR #546, opened 2019-07-15
Workaround for broken TS getters (important for spreading State)

### LINKED ISSUE #511: Promise rejection issue in React Native
## Bug or feature request?

Bug in React Native, actually... sort of...

## Description:

I was getting an issue when a promise rejected, and the error I got was related to React Native's own polyfill of `Object.assign`. It fails on `hasOwnProperty`. I'm not sure if this is still an issue in the latest React Native, but I'm using expo so I'm a couple versions behind.

The offending code in RN is here: https://github.com/expo/react-native/blob/exp-latest/Libraries/polyfills/Object.es6.js#L54

The error I get is: "One of the sources for assign has an enumerable key on the prototype chain. Are you trying to assign a prototype property? We don't allow it, as this is an edge case that we do not support. This error is a performance optimization and not spec compliant."

For the record, the key it's breaking on is `inert` on a state object. Not sure what that's used for, but that's the one it doesn't like. This all takes place in the `resolveState` function.

I think this is related to the new spawning behavior in 4.6.x. I downgraded to 4.5.0, as I wasn't using the actor model stuff anyway and that fixed the issue for me. I don't really expect this library to address the weirdness in React Native, but I wanted to leave this comment here for posterity. I'm not even sure if it could be fixed from this end without breaking something else.

Maybe we can put it in a troubleshooting doc for any other poor souls that run into this?

Andarist:
fixes #511 

Not sure if we can do this in this major though. Was `.inert` getter part of the public API?

### DIFF
--- a/src/State.ts
+++ b/src/State.ts
@@ -66,6 +66,11 @@ export class State<TContext, TEvent extends EventObject = EventObject>
    */
   public tree?: StateTree;
   /**
+   * The next events that will cause a transition from the current state.
+   */
+  // @ts-ignore - getter for this gets configured in constructor so this property can stay non-enumerable
+  public nextEvents: EventType[];
+  /**
    * Creates a new State instance for the given `stateValue` and `context`.
    * @param stateValue
    * @param context
@@ -146,13 +151,6 @@ export class State<TContext, TEvent extends EventObject = EventObject>
   }
 
   /**
-   * Returns a new `State` instance that is equal to this state no actions (side-effects).
-   */
-  public get inert(): State<TContext, TEvent> {
-    return State.inert(this, this.context);
-  }
-
-  /**
    * Creates a new State instance.
    * @param value The state value
    * @param context The extended state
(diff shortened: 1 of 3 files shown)


src/State.ts
```diff
@@ -180,17 +174,8 @@ export class State<TContext, TEvent extends EventObject = EventObject>
...
-  /**
-   * The next events that will cause a transition from the current state.
-   */
-  public get nextEvents(): EventType[] {
-    if (!this.tree) {
-      return [];
-    }
 
-    return this.tree.nextEvents;
+    this.nextEvents = !config.tree ? [] : config.tree.nextEvents;
```
davidkpiano: This is a potentially expensive operation.

Andarist: Should I change it to a regular method then? Was thinking about removing it entirely, but that would force consumers to handle missing tree themselves, which is not that ergonomic.

davidkpiano: Let's change it to a method, then. It's not public API.

Andarist:
@davidkpiano changed `.nextEvents` to `.getNextEvents()` method

davidkpiano:
Actually I misspoke, `state.nextEvents` is public API.

Andarist:
Ok, I've brought back `.nextEvents` getter, but in such a form that it should fix the reported issue.

PR #550, opened 2019-07-16
Fixed transition target resolving algorithm

### LINKED ISSUE #346: Root transitions are not working properly
Consider the following snippet:

```js
const wordMachine = Machine({
  id: 'direction',
  initial: 'left',
  states: {
    left: {},
    right: {},
    center: {},
    justify: {}
  },
  on: {
    LEFT_CLICK: 'left',
    RIGHT_CLICK: 'right',
    CENTER_CLICK: 'center',
    JUSTIFY_CLICK: 'justify'
  }
});
```

If you copy and paste it in the Visualizer (https://statecharts.github.io/xstate-viz/) you'll see that you can transition to `right` by clicking on `RIGHT_CLICK` event.

It shouldn't happen. `left` means "travel to my sibling left state", `direction` has no siblings.

The expected result would be a warning: `Target 'left' not found on 'direction'` (or something like that).

Andarist:
fixes #346

This is quite a breaking change though, so maybe shouldn't get merged in right away. Not sure what are the plans for the next major version.

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -252,8 +252,13 @@ class StateNode<
     this.delimiter =
       _config.delimiter ||
       (this.parent ? this.parent.delimiter : STATE_DELIMITER);
+
+    if (!IS_PRODUCTION && _config.id && _config.id.includes(this.delimiter)) {
+      throw new Error(`Explicitly configured state node ID ("${_config.id}") cannot contain path delimiter ("${this.delimiter}").`);
+    }
+
     this.id =
-      _config.id || [this.machine.key, ...this.path].join(this.delimiter);
+        _config.id || [this.machine.key, ...this.path].join(this.delimiter);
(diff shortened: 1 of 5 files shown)


Andarist:
@davidkpiano would we like to revive this PR as part of v4? or should it wait for v5?

davidkpiano:
Let's hold off on this until V5. We need to make sure we have a good resolution algorithm, and might want to support `"#foo.bar"` resolving to:

- `id: 'foo.bar'` if it exists...
- otherwise: `id: 'foo'` and find the child `'bar'` state

Not sure yet.

davidkpiano:
Doing some cleanup, do we want to rework this for v5 instead?

davidkpiano:
Superseded by https://github.com/statelyai/xstate/pull/2881

PR #551, opened 2019-07-16
Tweak explanation around self transition target patterns

Andarist:


docs/guides/transitions.md
```diff
@@ -195,7 +195,7 @@ const buttonMachine = Machine({
 
 - `EVENT: '.foo'` - internal transition to child
 - `EVENT: { target: '.foo' }` - internal transition to child (starts with `'.'`)
-- `EVENT: { target: 'same.foo', internal: true }` - explicit internal transition to child
+- `EVENT: { target: 'same.foo', internal: true }` - explicit internal transition to a sibling's child
```
davidkpiano: It's not a sibling though, it's the original (same) state:

```suggestion
- `EVENT: { target: 'same.foo', internal: true }` - explicit internal transition to own state node's child
```

docs/guides/transitions.md
```diff
@@ -226,7 +226,7 @@ Every transition above is explicit and will have its `exit` and `entry` actions
 
 - `{ EVENT: 'foo' }` - all transitions to siblings are external transitions
 - `{ EVENT: '#someTarget' }` - all transitions to other nodes are external transitions
-- `{ EVENT: 'same.foo' }` - external transition to child node with reference to self (parent node)
+- `{ EVENT: 'same.foo' }` - external transition to a sibling's child
```
davidkpiano:
```suggestion
- `{ EVENT: 'same.foo' }` - external transition to own state node's child
```

Andarist:
Ok, I get it now - `"same.foo"` is equivalent to `{ target: ".foo", internal: false }`, right?

Looking at how this resolution algorithm behaves ( https://codesandbox.io/s/xstate-typescript-template-dlvi3 ) it's IMHO quite confusing that `"sibling"` is allowed, but `"sibling.child"` is not. 

There are quite some ways of specifying the target using "custom" resolving rules (not described in SCXML) and I kinda feel like those rules atm don't create a cohesive whole. Maybe we could discuss (in separate issue ofc) what would be desired rules for resolving (if you agree with me that current ones are slightly confusing at times)?

davidkpiano:
> Ok, I get it now - "same.foo" is equivalent to { target: ".foo", internal: false }, right?

That's correct.

Andarist:
Ok, I've adjusted the PR

PR #562, opened 2019-07-22
Bring back state node ID resolving to getRelativeStateNodes

Andarist:
Not completely sure if this is a correct fix, but `@xstate/graph` still expects this method to accept strings here:
https://github.com/davidkpiano/xstate/blob/b25ebe228411cec81067efe7ae54b79bf1a2767c/packages/xstate-graph/src/graph.ts#L94

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -1618,20 +1618,24 @@ class StateNode<
   /**
    * Returns the leaf nodes from a state path relative to this state node.
    *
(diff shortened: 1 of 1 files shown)


davidkpiano:
We should fix this in `@xstate/graph` - removing the state node ID as a resolved target improves performance and reduces a lot of previously necessary logic.

Andarist:
Sure, gonna try to look into fixing this the other way around.

PR #571, opened 2019-07-26
Migrate to Jest

Andarist:
This is a proposal to migrate to jest. While I don't have anything against used tooling jest makes few things simpler - mostly running multiple projects within a single jest instance, which comes in handy when working with monorepos.

Additional benefits here are that I've used similar thing for a TS compilation - it has "references" feature, which allows to build all projects in one go.

I've introduced a new single `dev` script (`npm run dev`). It runs everything against its respective current state - it should be way harder now to desynchronize packages etc. If i.e. we change xstate implementation and i.e. xstate-graph depends on it then xstate-graph will be compile & tested against `xstate`'s current src directory.

To make this work the repository has to be bootstrapped/linked (with lerna), while this works with npm I've found in the past that this in general works better with yarn as it's capable to install whole monorepo in one go (thanks to its workspaces feature). I recommend moving to yarn, but didn't want to do this as part of this PR. This would require moving xstate implementation to a subdirectory (preferrably into packages).

There were quite some changes, please review carefully so we can avoid any regressions and also if something is unclear then please ask about it - I'll gladly explain everything in more detail.


tivac:
I've got a lerna monorepo that installs all dependencies fine without needing to use bootstrapping. It uses `file` package references for all the cross linked packages and then has all the children defined as dependencies in the root. Works great and was really straightforward to set up, now I can `npm i` from the root and it gets everything.

Andarist:
Here we are dealing with public monorepo and no a private one, we can't quite use `file` because packages can be published and they should reference their dependencies as if they were installed from npm

tivac:
https://github.com/tivac/modular-css

Lerna replaces the file references with normal package references on publish. This workflow is what lerna uses as well because it's so low friction and straightforward.

Done with my tangent though. Jest testing the whole monorepo is a big 👍🏻 from me.

Andarist:
> Lerna replaces the file references with normal package references on publish. This workflow is what lerna uses as well because it's so low friction and straightforward.

Interesting, I had no idea lerna handles it like this. This indeed seems like a really low-friction approach for node-only packages (for those that need transpilation and stuff we still need to redirect somehow requests to respective source files). From what I see a slight downside is that you have to list manually all "linked" packages in root package.json. Also harder to make a publish without lerna, but arguably all publications should go through lerna.

It doesn't give much over using yarn though (just saying, im not rly defending yarn approach - ur approach seems really cool too). With yarn it's also easier to add dependencies to subpackages and have only a single lock file at the root (as its monorepo-aware it can do that, whereas npm cant quite do this).

Andarist:
@davidkpiano thoughts on this one? 😉

davidkpiano:
Hey, can we merge master into this and see if it works?

packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -38,65 +37,65 @@ describe('@xstate/fsm', () => {
   it('should have the correct initial state', () => {
     const { initialState } = lightFSM;
 
-    assert.deepEqual(initialState.value, 'green');
-    assert.deepEqual(initialState.actions, [{ type: 'enterGreen' }]);
+    expect(initialState.value).toEqual('green');
```
davidkpiano: Before we merge this, I strongly prefer `assert` function syntax to `expect` syntax.

Andarist: I can change this back - that's not a problem. I'm wondering if it would be good though. 

Possible drawbacks:
- lack of cohesive docs (for some things u have to look into jest, for some u have look into i.e. chai
- possibly jest error messages are designed to understand their expect API better (and thus such errors can be more readable), i.e. they provide extra hints for times when a developer uses `toBe` instead of `toEqual` by accident
- some built-in features would be inaccessible (or would lead to 2 different styles being used in the codebase which is rather ugly), such as **inline** snapshots - which are useful i.e. for expecting error messages

davidkpiano: Okay, that's fair.

PR #577, opened 2019-07-27
Add dev-only warning for misconfigured transitions (when non-last possible transition is unguarded)

### LINKED ISSUE #569: Transition with multiple targets inconsistent handling
## Description:

If a transition is in array form like `on: { DEACTIVATE: [ '.mode.inactive', '.status.disabled' ] }` the transition does not work as expected. Not until the transition is in object form like `on: { DEACTIVATE: { target: [ '.mode.inactive', '.status.disabled' ] } }` does the correct behaviour occur when sending the event.

## Expected result:

When sending an event that has multiple targets, the event should cause the transition regardless of how the transition is defined in the statechart (i.e. in array form or object form)

## Actual result:

When sending an event that has multiple targets, no transition occurs IFF in the startchart the transition is in array form like: `on: { DEACTIVATE: [...] }`

## Potential fix:

I would assume the easiest fix would be to transform transitions defined as arrays into objects. So at the lower-level you only ever deal with transitions in a normalized, object form. In short, normalize the representation of transitions from a statechart.

## Example StateChart:

Can be copy and pasted in the visualizer. Replace the `DEACTIVATE: ['.mode.inactive', '.status.disabled']` with the commented out example then you will see the event buttons enable and the logs start to log as expected.

```
  // Available variables:
  // - Machine
  // - interpret
  // - assign
  // - send
  // - sendParent
  // - spawn
  // - raise
  // - actions
  // - XState (all XState exports)

const settingsMachine = Machine({
  id: 'settings',
  type: 'parallel',
  states: {
    mode: {
      initial: 'active',
      states: {
        inactive: {},
        pending: {},
        active: {}
      }
    },
    status: {
      initial: 'enabled',
      states: {
        disabled: {},
        enabled: {}
      }
    }
  },
  on: {
    // Multiple targets
    DEACTIVATE: ['.mode.inactive', '.status.disabled']
    // Can also be coded as...
    // DEACTIVATE: {
    //   target: ['.mode.inactive', '.status.disabled']
    // }
  }
});

const i = interpret(settingsMachine)
i.start()
i.onTransition((s, e) => {
  console.log('here', e)
  console.log(s.value)
})
i.send('DEACTIVATE')
```

Andarist:
closes #569

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -295,16 +295,15 @@ export type StatesDefinition<
 };
 
 export type TransitionsConfig<TContext, TEvent extends EventObject> = {
-  [K in TEvent['type'] | BuiltInEvent<TEvent>['type']]?:
+  [K in TEvent['type'] | BuiltInEvent<TEvent>['type']]?: SingleOrArray<
     | string
     | number
     | StateNode<TContext, any, TEvent>
-    | SingleOrArray<
-        TransitionConfig<
-          TContext,
-          TEvent extends { type: K } ? TEvent : EventObject
-        >
-      >;
+    | TransitionConfig<
+      TContext,
+      TEvent extends { type: K } ? TEvent : EventObject
+    >
+  >;
 };
 
 export type TransitionsDefinition<TContext, TEvent extends EventObject> = {
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -1927,36 +1927,47 @@ class StateNode<
           return [{ target: undefined, event, actions: [], internal: true }];
         }
 
-        if (isArray(value)) {
-          return value.map(targetTransitionConfig =>
-            this.formatTransition(
-              targetTransitionConfig.target,
-              targetTransitionConfig,
-              event
-            )
+        const transitions = toArray(value)
+
+        if (!IS_PRODUCTION) {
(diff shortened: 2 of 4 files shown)


src/types.ts
```diff
@@ -295,16 +295,15 @@ export type StatesDefinition<
 };
 
 export type TransitionsConfig<TContext, TEvent extends EventObject> = {
-  [K in TEvent['type'] | BuiltInEvent<TEvent>['type']]?:
+  [K in TEvent['type'] | BuiltInEvent<TEvent>['type']]?: SingleOrArray<
```
davidkpiano: We should not have `SingleOrArray` here since the only array should be an array of `TransitionConfig` objects. In non-TS envs, `string[]` will produce the warning, but in TS envs it will cause a compilation error, which is good.

Andarist: I was thinking about this and came to the conclusion that it's better here to fail at runtime rather than when compiling. TS errors can't be customized - a runtime error here gives way more context to the user, so I would argue it's actually more helpful.

This change also allows for such config - https://github.com/davidkpiano/xstate/pull/577/files#diff-6112fd8214555eb54874e002969533ebR355 and this seems to make sense to me. We shouldn't forbid a simple form of a fallback target.

PR #578, opened 2019-07-28
Support wildcard event descriptor

### LINKED ISSUE #155: Is it possible to define default (catch all) transition for a particular state?
## Bug or feature request?

Question or feature request.

## Description:

In `switch...case`or [pattern-matching](https://github.com/tc39/proposal-pattern-matching) syntax:

```js
switch (value) {
  case 1:
    // 
  case 2: 
    //
  default:
    // .... 
}
```

There is a way to catch all of other unmatched values using `default` or `_`. Is there any approach to do this in xstate?

Use `_`  to catch all of non `SOMETHING` event:

```js
{
  states: {
    a: {
      on: {
        SOMETHING: 'b',
        _: 'c'
      }
    },
    b: {
      on: {}
    },
    c: {
      on: {}
    },
  }
}
```

or `'*'`:

```js
{
  states: {
    a: {
      on: {
        SOMETHING: 'b',
        '*': 'c'   // <- add quotes to avoid SyntaxError
      }
    },
    b: {
      on: {}
    },
    c: {
      on: {}
    },
  }
}
```

Probably No. If we do not name any events to `_` or `*`.

Probably yes? As `*` event:

```
<state id=s">
   <transition event="e" cond="x==1" target="s1"/>
   <transition event="e" target="s2"/>
   <transition event="*" target="s3"/>
</state>
```

Andarist:
closes #155

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -99,6 +99,7 @@ const STATE_DELIMITER = '.';
 const NULL_EVENT = '';
 const STATE_IDENTIFIER = '#';
 const TARGETLESS_KEY = '';
+const WILDCARD = '*'
 
 const EMPTY_OBJECT = {};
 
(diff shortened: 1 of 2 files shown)


davidkpiano:
Awesome! Before I merge this, does this type-check? Let's make sure it doesn't cause compilation errors when `TEvent` is specified in `Machine<...>`.

Andarist:
Well, I've run the whole test suite after implementing this - so it has compiled both `src` and `test` directories without any errors.

PR #586, opened 2019-08-03
Forward events before processing

### LINKED ISSUE #585: Bug: When invoking a child machine off of an onError handler, using autoForward: true causes a console error
## Bug or feature request?

Bug

## Description:

When a parent machine has a state with an onError handler for an invocation, if that onError handler invokes a child machine and uses `autoForward: true`, an error will appear in console. When removing `autoForward: true`, the error does not appear.

Error text: `Missing onError handler for invocation 'failingPromise', error was 'Error: The api call failed!'. Stacktrace was 'Error: The api call failed!`

## (Bug) Expected result:

No console error when using `autoForward: true`

## (Bug) Actual result:

Console throws error when using `autoForward: true`

https://codesandbox.io/s/silly-lehmann-86hnv

In the above example, the same child machine gets invoked two different times off of `onError` handlers. The first time is with `autoForward: true` and the second time is without it. When watching console, you'll see that a console error is thrown the first time but not the second time.

Andarist:
fixes #585

### DIFF
--- a/test/invoke.test.ts
+++ b/test/invoke.test.ts
@@ -6,7 +6,7 @@ import {
   send,
   EventObject,
   StateValue
-} from '../src/index';
+} from '../src';
 import { actionTypes, done as _done, doneInvoke } from '../src/actions';
 import { interval } from 'rxjs';
 import { map, take } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


test/invoke.test.ts
```diff
@@ -255,6 +255,107 @@ describe('invoke', () => {
     service.send('FORWARD_DEC');
   });
 
+  it.only('should forward events to services if autoForward: true before processing them', done => {
```
davidkpiano:
```suggestion
  it('should forward events to services if autoForward: true before processing them', done => {
```

Andarist: hah, this was leftover as I've started to work on this one before you've merged in jest support and forgot to remove this when i've applied my stashed changed

fixed

PR #588, opened 2019-08-04
Make context property required when TContext gets specified

### LINKED ISSUE #583: [TS] Context not required on MachineConfig
## Description:

Current typings allowed context to stay `undefined` and this might cause a runtime error - https://codesandbox.io/s/xstate-typescript-template-452k1 .

Because a MachineConfig is kinda a static template of the machine which we can later extend with `.withContext` etc this probably needs to stay optional.

I'm wondering what we can do to overcome this - would splitting the type to `MachineConfig` & `MachineConfigWithContext` help here? Assuming ofc that we would, based on this, create 2 distinct return types and make `interpret` accept only the `MachineWithContext`? 

What about "pure" Machine API? I think it might be vulnerable right now in a similar fashion.

Andarist:
Fixes #583

### DIFF
--- a/packages/xstate-graph/src/graph.ts
+++ b/packages/xstate-graph/src/graph.ts
@@ -11,18 +11,33 @@ import {
   TransitionDefinition,
   ValueAdjacencyMap
 } from 'xstate';
-import { getActionType, flatten, keys, warn, toEventObject } from 'xstate/lib/utils';
+import {
+  getActionType,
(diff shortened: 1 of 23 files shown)


davidkpiano:
This looks like a breaking change. Can you confirm?

davidkpiano:
Closing (stale) - will target v5 for this change

PR #589, opened 2019-08-04
Remove dead path in StateNode#id calculation

Andarist:


src/StateNode.ts
```diff
@@ -259,10 +259,7 @@ class StateNode<
       _config.delimiter ||
       (this.parent ? this.parent.delimiter : STATE_DELIMITER);
     this.id =
-      _config.id ||
-      (this.machine
-        ? [this.machine.key, ...this.path].join(this.delimiter)
-        : this.key);
+      _config.id || [this.machine.key, ...this.path].join(this.delimiter);
```
Andarist: this.machine can't be falsey here

PR #597, opened 2019-08-06
Tweak test scripts in package.json

Andarist:
- `test` already only runs tests - just in whole monorepo
- `test:core` will now only run tests of the core package, using the same config (root jest.config.js) for the core package which is being used by `test`
- `test:only` was obsolete (doing pretty much what `test:core` did), if I've misunderstood the  intention here - please let me now


davidkpiano:
Yeah, leave `test:only` in. The use-case is if I need to quickly only test one file, I can drag/drop (or specify) the file name and run e.g.:

```bash
npm run test:only /path/to/some.test.ts
```

davidkpiano:
Or I guess I can do this with `npm run test /path/to/some.test.ts`

PR #599, opened 2019-08-07
Fixed which _event gets passed to log

Andarist:


src/actions.ts
```diff
@@ -253,21 +256,29 @@ const defaultLogExpr = <TContext, TEvent extends EventObject>(
  *  - `event` - the event that caused this action to be executed.
  * @param label The label to give to the logged expression.
  */
-export function log<TContext, TEvent extends EventObject>(
-  expr: (
-    ctx: TContext,
-    event: TEvent,
-    meta: SCXMLEventMeta<TEvent>
-  ) => any = defaultLogExpr,
+export function log<TContext, TEvent extends EventObject, T>(
```
Andarist: I'm unsure about that `T` here, in theory it would be nice to have strong-types result of `expr`, but I'm not sure how this should be handled internally. Should we parametrize it with `any` in such case? In our generic context inside StateNode etc we have no idea what's the actual type of that T. Seems a little bit quirky? But maybe it's ok 🤷‍♂

davidkpiano: Since `log(...)` is a fire-and-forget action, we don't really care about the return value of it, if any.

Andarist: Ok, I've removed that type parameter entirely. This leads to explicit `any` being used [here](https://github.com/davidkpiano/xstate/blob/9ab85cf40305931962ad6b019b2e7135e7c45f21/src/types.ts#L750), not a big problem I guess.

PR #600, opened 2019-08-07
Fixed issue with NULL event not being processed before raised events

Andarist:


test/scxml.test.ts
```diff
@@ -101,13 +101,7 @@ const testGroups = {
...
   ],
   // 'scxml-prefix-event-name-matching': ['star0'], TODO: source ordering of wildcard appearing first
-  'w3c-ecma': [
-    'test144.txml',
-    'test158.txml',
-    'test173.txml',
-    'test174.txml',
-    'test178.txml'
-  ]
+  'w3c-ecma': ['test144.txml', 'test158.txml', 'test173.txml', 'test174.txml']
```
Andarist: I've removed this test - https://gitlab.com/scion-scxml/test-framework/blob/b46a10a1c3a3b1ca5c5cb4bb44ddb5c785611f41/test/w3c-ecma/test178.txml.scxml

It no longer has a corresponding `.json` with "expected" description. Also - it isn't meant to be tested automatically but rather verified manually. I'm not sure though what are exact assumptions around those tests, because even if verified manually I'm not sure how the "expected" can be observed - for sure we can't verify it as in XState we can't observe any difference between those 2 events: `{ type: 'foo', a: 1 }` a `{ type: 'foo', a: 2, a: 1 }`, but I'm also not sure how it can be observed in other environments with datamodel specified as ecmascript.

PR #605, opened 2019-08-08
Refactor event conversion to be more faithful

Andarist:
followup to #604 , while ofc we can convert `event` to be an `eventexpr` I've thought that it's better to be more precise while converting and thus we should stick to a simple form of `event` when possible


davidkpiano:
Makes sense.

PR #606, opened 2019-08-09
Support meta argument in delay expressions

Andarist:


src/interpreter.ts
```diff
@@ -638,40 +638,16 @@ export class Interpreter<
     }
   }
   private defer(sendAction: SendActionObject<TContext, TEvent>): void {
-    let { delay } = sendAction;
-
-    if (isString(delay)) {
```
Andarist: sends are already being resolved in StateNode so I couldn't find a reason to support this resolving in Interpreter too + no test has failed because of removing this

PR #610, opened 2019-08-09
Add support for delayexpr conversion

Andarist:


test/scxml.test.ts
```diff
@@ -305,7 +305,7 @@ const testGroups = {
     // 'test550.txml',
     // 'test551.txml',
     // 'test552.txml',
-    'test553.txml',
```
davidkpiano: Do these tests fail now?

Andarist: yeah, they were passing "by accident" - [this particular test](https://gitlab.com/scion-scxml/test-framework/blob/master/test/w3c-ecma/test553.txml.scxml#L8) was sending immediately a delayed event because delayexpr wasnt checked at all and because of that it was "passing", but the actual expectations of that test were not met

PR #614, opened 2019-08-10
Support special #_internal target for send

Andarist:


test/interpreter.test.ts
```diff
@@ -758,6 +758,38 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
         .onDone(() => done())
         .start();
     });
+
+    it('should be able to raise event using special target', () => {
+      const machine = Machine({
+        initial: 'foo',
+        states: {
+          foo: {
+            entry: [send('EVENT_2'), send('EVENT_1', { to: '#_internal' })],
```
davidkpiano: We can fix this in another PR (and this is a mistake on my part), but this should be `to: '_internal'` (no '#' prefix) for SCXML compatibility.

Andarist: I was actually trying to determine which version is correct while preparing this PR but SCXML spec isn't very clear about it.

No prefix:
- [5.10.1 The Internal Structure of Events](https://www.w3.org/TR/scxml/#InternalStructureofEvents)
- [6.2 <send>](https://www.w3.org/TR/scxml/#send)
- [G.4 Calculator Example](https://www.w3.org/TR/scxml/#N11630)

Prefix:
- [C.1 SCXML Event I/O Processor](https://www.w3.org/TR/scxml/#SCXMLEventProcessor)
- [Test 189](https://www.w3.org/Voice/2013/scxml-irp/189/test189.txml)

davidkpiano: Interesting, looks like it's also `#_internal` here: https://doc.qt.io/qt-5/qml-qtscxml-scxmlstatemachine.html

Let's keep it with the prefix; ideally, this should _never_ be manually typed anyway (as it is a special target).

PR #620, opened 2019-08-14
Cleanup/scxml event

Andarist:
2 reasons for that:

1. conversion from SCXML.Event to TEvent is always losing information, so it's better to operate internally on the former as it allows us to have better SCXML compat and only "cast" it to TEvent (by accessing `_event.data`) when we expose it to the world - so we don't lose backward compatibility
2. avoids needless conversion from TEvent to SCXML.Event all over the place

This refactor has been done hopefully without breaking changes - whole test suite passes and I haven't touched tests (besides really few lines which were testing rather private stuff.

### DIFF
--- a/src/State.ts
+++ b/src/State.ts
@@ -7,15 +7,15 @@ import {
   EventType,
   StateValueMap,
   StateConfig,
-  ActionTypes,
   SCXML,
   StateSchema,
   ExtractStateValue
 } from './types';
 import { EMPTY_ACTIVITY_MAP } from './constants';
-import { matchesState, keys, isString, toSCXMLEvent } from './utils';
+import { matchesState, keys, isString } from './utils';
 import { StateNode } from './StateNode';
 import { nextEvents } from './stateUtils';
+import { initEvent } from './actions';
 
 export function stateValuesEqual(
   a: StateValue | undefined,
@@ -63,7 +63,7 @@ export function bindActionToState<TC, TE extends EventObject>(
             exec(state.context, state.event as TE, {
               action,
               state,
-              _event: toSCXMLEvent(state.event)
+              _event: state._event
             })
         : undefined
   };
@@ -118,7 +118,6 @@ export class State<
         return new State<TC, TE>({
           value: stateValue.value,
           context: context as TC,
-          event: stateValue.event,
           _event: stateValue._event,
           historyValue: stateValue.historyValue,
           history: stateValue.history,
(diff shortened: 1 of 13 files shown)


src/State.ts
```diff
@@ -171,13 +169,12 @@ export class State<
       if (!stateValue.actions.length) {
         return stateValue as State<TC, TE>;
       }
-      const event = { type: ActionTypes.Init } as TE;
+      const _event = initEvent as SCXML.Event<TE>;
 
       return new State<TC, TE>({
         value: stateValue.value,
         context,
-        event,
```
Andarist: this is duplicated info, as `_event` always contain `event` in its `_event.data`, no need pass both here - both are assigned in State constructor though (backwar compat)

src/StateNode.ts
```diff
@@ -1276,7 +1281,7 @@ class StateNode<
         : currentState
         ? currentState.meta
         : undefined,
-      events: resolvedStateValue ? (raisedEvents as TEvent[]) : [],
```
Andarist: this has always been empty, as we empty `raisedEvents` (by mutating it) before we reach this point

src/actions.ts
```diff
@@ -143,13 +143,25 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
  */
 export function raise<TContext, TEvent extends EventObject>(
   event: Event<TEvent>
-): RaiseEvent<TContext, TEvent> {
+): RaiseAction<TEvent> | SendAction<TContext, TEvent> {
+  if (!isString(event)) {
```
Andarist: wasn't 100% sure how to approach this, weirdly SCXML allows only "simple" events in raise, so I've kept that - but at the same time we can create more complex events to be raised using `<send/>` with internal target

so I've decided to keep SCXML semantics so conversion can be accurate, but also made our `raise` "smarter" so it can handle complex events, just by using send internally instead of raise

src/interpreter.ts
```diff
@@ -561,7 +567,10 @@ export class Interpreter<
     return this.send.bind(this, event);
   }
 
-  public sendTo = (event: TEvent, to: string | number | Actor) => {
+  private sendTo = (
```
Andarist: I've changed this to private - this really has seemed like undocumented and supposed to only called by the Interpreter itself, hope this is OK

src/stateUtils.ts
```diff
@@ -190,20 +190,3 @@ export function isInFinalState(
 
   return false;
 }
-
-export function getDoneData<TContext>(
```
Andarist: was unused & using unbound `this` here, so I've just removed it for now

PR #622, opened 2019-08-14
Tweak build setup - fixes inability to bundle distributed XState with Rollup

### LINKED ISSUE #498: xstate throws errors when used with rollup
## Bug or feature request?
Bug

## Description:
Using the most basic possible rollup configuration with xstate results in several errors during the bundling process

## proof-of-concept:

rollup.config.js:

```javascript
import resolve from 'rollup-plugin-node-resolve';

export default {
  plugins: [
    resolve()
  ]
};
```

index.js:
```javascript
import { Machine, interpret } from 'xstate';


const storypointsMachine = Machine({
  id: 'storypoints',
  initial: 'inactive',
  states: {
    inactive: {
        on: {
            INIT: 'active'
        }
    },
    active: {
        on: {
            DESTROY: 'destroyed'
        }
    }
  }
});
```

build command:
```bash
rollup index.js -c rollup.config.js --format=iife
```

Output of build command:
```bash
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en#error-this-is-undefined
node_modules/xstate/es/utils.js
1: var __assign = this && this.__assign || function () {
                  ^
2:     __assign = Object.assign || function (t) {
3:         for (var s, i = 1, n = arguments.length; i < n; i++) {
...and 7 other occurrences
node_modules/xstate/es/mapState.js
1: var __values = this && this.__values || function (o) {
                  ^
2:     var m = typeof Symbol === "function" && o[Symbol.iterator],
3:         i = 0;
...and 1 other occurrence
node_modules/xstate/es/StateNode.js
1: var __assign = this && this.__assign || function () {
                  ^
2:     __assign = Object.assign || function (t) {
3:         for (var s, i = 1, n = arguments.length; i < n; i++) {
...and 9 other occurrences

...and 6 other files
created stdout in 321ms
```

Andarist:
fixes #498


rollup.config.js
```diff
@@ -1,24 +1,43 @@
...
+        declaration,
+        ...(target && { target })
+      }
+    }
+  });
+
+const createNpmConfig = ({ input, output }) => ({
+  input,
+  output,
+  preserveModules: true,
```
Andarist: This is kinda the main hero of this PR - we can use Rollup to do its thing, but also preserve the directory structure instead of its default, which is so-called "flat bundling". I would be in favor of flat bundling in the future - but that's a breaking change and also to make it possible few things would have to move places (right now some packages reach to "private" exports of XState - like [here](https://github.com/davidkpiano/xstate/blob/75f4a73de29317579b347c205abb19e1c4980a7f/packages/xstate-graph/src/graph.ts#L14)).

While we don't necessarily need rollup much here for building `lib` & `es` directories it makes one thing which fixes the original problem. Instead of putting `tslib` helpers into **each** file which is using them it tries to import them from `tslib` package. That way it tries to avoid duplicating them all over the codebase. You don't want to depend on `tslib` package though (which I'm slightly sad about 😉 ) - so we don't put `tslib` as Rollup's external and in this case it creates a "virtual" module for them, it writes it to disk and each file thanks to Rollup's compilation process can just import those helpers from that created, private file. And because it uses `tlib` version of those helpers they don't contain any attemps to read global this from within a module.

package.json
```diff
@@ -80,16 +81,17 @@
...
-    "rollup-plugin-replace": "^2.1.0",
-    "rollup-plugin-terser": "^4.0.4",
-    "rollup-plugin-typescript2": "^0.21.1",
-    "rollup-plugin-uglify": "^6.0.0",
+    "rollup": "^1.19.4",
+    "rollup-plugin-filesize": "^6.2.0",
+    "rollup-plugin-replace": "^2.2.0",
+    "rollup-plugin-terser": "^5.1.1",
+    "rollup-plugin-typescript2": "^0.22.1",
+    "rollup-plugin-uglify": "^6.0.2",
```
tivac: Is `rollup-plugin-uglify` actually used? I don't see any reference to it in `rollup.config.js` 🤔

Andarist: No, its not - I’ve actually just upgraded all rollup-related packages and ive taken them from package.json. Would you like to maybe prepare a PR with removing this?

PR #627, opened 2019-08-18
Remove index signature from EventObject to make things stricter

Andarist:


src/types.ts
```diff
@@ -15,11 +15,6 @@ export interface EventObject {
    * The type of event that is sent.
    */
   type: string;
-  /**
-   * The unique ID that identifies this specific event instance.
-   */
-  id?: string | number;
```
Andarist: don't think `id` has currently any meaning on the event object as that thing (sendid) is set on SendAction and SCXML.Event

farskid:
Does this change help TS infer event types better from the call site? some events might have a data prop and some not. can TS infer the possible types of events from the call site?

Andarist:
It doesn't help to infer anything - it just restricts places where plain EventObject is being used to only have type property. Without this in certain places unsafe access was allowed such as:
```
const ev: EventObject = { type: 'foo' }
ev.whatever.nested // unsafe
```

This has happened because `EventObject` had index signature typed to `any`.

Do you have any particular problematic cases where inferring fails? What kind of call sites do u have in mind?

PR #630, opened 2019-08-23
Fix delimiter in state.toStrings() method

alizbazar:
This fixes a bug with toStrings() method not respecting `delimiter` for deeply nested states.

```js
State.create({ value: { foo: { bar: 'poo' } } }).toStrings(undefined, ':')
```

Results in:
```
- ['foo', 'foo:bar', 'foo:bar.poo']
+ ['foo', 'foo:bar', 'foo:bar:poo']
```


Andarist:
Could u add a test for this?

alizbazar:
I can. Can you point me to instructions how to get the tests to run locally? Tried `npm install` and then `npm test` and `npm test:core`, but ended up with a bunch of errors, such as Typescript errors and `Cannot find module 'xstate'.`

Andarist:
`npm run bootstrap && npm run dev`

alizbazar:
Getting the following errors: 
image

alizbazar:
Added a test so this should now be ready to merge.

However, I still could not run the test suite, mostly failing on various Typescript errors, such as this:

```test/invoke.test.ts:1726:61 - error TS2339: Property 'value' does not exist on type 'EventObject'.```

Andarist:
Thanks for reporting those TS errors - I'm working on a fix for this.

davidkpiano:
Can you merge master into this? Types should be fixed now.

alizbazar:
Rebased from master. Still getting a few errors - this time different. `npm run dev` fails with:

image

In addition, if I run `npm test` I get the following:

image

PR #631, opened 2019-08-24
Fixed types in tests + improved inference of TEvent passed to TransitionConfig

Andarist:


packages/xstate-graph/src/graph.ts
```diff
@@ -58,7 +58,7 @@ export function deserializeEventString<TEvent extends EventObject>(
 }
 
 export interface ValueAdjMapOptions<TContext, TEvent extends EventObject> {
-  events: { [K in TEvent['type']]: Array<TEvent & { type: K }> };
+  events: { [K in TEvent['type']]?: Array<TEvent & { type: K }> };
```
Andarist: im not 100% sure about this change, please check if its correct

src/types.ts
```diff
@@ -15,7 +15,6 @@ export interface EventObject {
    * The type of event that is sent.
    */
   type: string;
-  [key: string]: any;
```
Andarist: I've removed this again because of this https://github.com/davidkpiano/xstate/pull/627#issuecomment-522462323 , this PR fixes all places in tests which are affected by removing this indexer

src/types.ts
```diff
@@ -492,7 +491,7 @@ export interface StateNodeDefinition<
   TContext,
   TStateSchema extends StateSchema,
   TEvent extends EventObject
-> extends StateNodeConfig<TContext, TStateSchema, TEvent> {
+> {
```
Andarist: This wasn't really needed, all actually set properties were defined directly on StateNodeDefinition - extending StateNodeConfig only kinda ensured that StateNodeConfig is a supertype of StateNodeDefinition, but because those types are recursive and use conditional types I couldn't satisfy TS after my changes.

After thinking about it I also came to a conclusion that those types don't have to be in super-sub relationship. Definition is a processed Config, but this fact doesn't have to restrict Definition to be a subtype of Config.

test/actor.test.ts
```diff
@@ -67,7 +67,7 @@ describe('spawning machines', () => {
       },
       SET_COMPLETE: {
         actions: send('SET_COMPLETE', {
-          to: (ctx, e) => {
+          to: (ctx, e: Extract<TodoEvent, { type: 'SET_COMPLETE' }>) => {
```
Andarist: `send` action doesn't seem to have event inferred correctly, probably something which we could improve in the future

src/types.ts
```diff
@@ -311,7 +310,7 @@ export type TransitionsConfig<TContext, TEvent extends EventObject> = {
     | StateNode<TContext, any, TEvent>
     | TransitionConfig<
         TContext,
-        TEvent extends { type: K } ? TEvent : EventObject
+        K extends TEvent['type'] ? Extract<TEvent, { type: K }> : EventObject
```
Andarist: This improves inference by propagating correctly narrowed down type.

Before:
Screenshot 2019-08-24 at 21 21 24
After:
Screenshot 2019-08-24 at 21 22 56

packages/xstate-graph/src/graph.ts
davidkpiano: Not every event needs to be represented in `events`. Previously, compilation would fail if you were missing events I believe.

test/actor.test.ts
davidkpiano: I agree, let's make an issue to track it.

src/types.ts
davidkpiano: I don't think this is a good idea, because `EventObject` should define an interface of _any_ arbitrary event, in which `[key: string]: any` allows that.

If developers want stricter event typing, they can provide it in `Machine<... MyEvent>` anyway. This change would force them to do this.

src/types.ts
```diff
@@ -15,7 +15,6 @@ export interface EventObject {
    * The type of event that is sent.
    */
   type: string;
```
davidkpiano:
```suggestion
  type: string;
  [key: string]: any;
```

Let's add this back for now. Too much code would be affected by this change, unfortunately.

src/types.ts
Andarist: > I don't think this is a good idea, because EventObject should define an interface of any arbitrary event, in which [key: string]: any allows that.

Without that indexer it still defines an interface for any event  - that indexer has other implications though, as it defines more than it should. Example of why this is bad is given in the linked comment.

This case is pretty much the same as the case of Redux and its Action type - see the definition here:
https://github.com/reduxjs/redux/blob/9c9a4d2a1c62c9dbddcbb05488f8bd77d24c81de/index.d.ts#L20-L22

> If developers want stricter event typing, they can provide it in Machine<... MyEvent> anyway. 

There are places where we fail at inferring concrete TEvent and in those places we fallback to using EventObject instead which makes all such places unsafe from TS perspective, even when a developer has opted-in for stricter typings by providing MyEvent.

> This change would force them to do this.

Is it bad though? The whole point of TS is to prevent mistakes at compile time.

> Let's add this back for now. Too much code would be affected by this change, unfortunately.

This is a valid consideration. Is it an uncommon thing to improve typings in a minor release to catch more possible errors though? I think it's not, even TS itself improves its type checker in minor releases and this always, by definition, is a breaking change for somebody as previously "valid" code might become invalid after the upgrade.

Maybe a compromise here (for a 4.x line) would be introducing `AnyEventObject` that would be a default for TEvent for Machine & Interpreter. That way it wouldn't affect places where TEvent is given by a developer and thus is would improve strictness of places where we fallback to EventObject?

Also - it potentially wouldn't have to affect a whole lot of places. The recommendation could be to just use `AnyEventObject` manually for Machine/Interpreter - this narrows down the affected places and makes the decision about having strict/non-strict types explicit - which IMHO is a good thing, because right this happens implicitly which gives a developer false feeling of being safe.

Andarist:
@davidkpiano used AnyEventObject approach, should be ready to merge now

davidkpiano:
Merged master, tested locally, good to go!

PR #647, opened 2019-09-04
Support array transition configs

Andarist:
Motivation:

In SCXML document order matters, so i.e.
```
        <transition target="pass" event="*"/>
        <transition target="fail" event="foo"/>
```
and
```
        <transition target="fail" event="foo"/>
        <transition target="pass" event="*"/>
```
are **not** equivalent. When sending `foo` event it has to be matched according to that document order.

TODO:
- [ ] review
- [x] add new tests (existing ones are passing 🎉 )
- [ ] cleanup `TransitionConfig` type drama (at different places it's intersected with `{event: TEvent["type"]}` or `{ event?: undefined }` 
- [x] consider caching possible candidates, right now every event is iterating through whole list each time
- [x] cleanup delays-related stuff (those with TODO comments)
- [ ] look into few `any`s used in this PR, try to fix them
- [x] for object configs put `*` at the end

### DIFF
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -1915,15 +1915,69 @@ class StateNode<
     };
   }
   private formatTransitions(): TransitionDefinition<TContext, TEvent>[] {
-    const onConfig = !this.config.on
-      ? []
-      : Array.isArray(this.config.on)
-      ? this.config.on
-      : flatten(
-          keys(this.config.on).map(key =>
-            toTransitionConfigArray(key, this.config.on![key as string])
-          )
+    const validateMappedTransitions = (
+      event: string,
+      transitions: Array<
+        TransitionConfig<TContext, EventObject> & {
+          event: string;
+        }
+      >
+    ) => {
+      const hasNonLastUnguardedTarget = transitions
+        .slice(0, -1)
+        .some(
+          transition =>
+            isString(transition.target) ||
+            isMachine(transition.target) ||
+            (!('cond' in transition) && !('in' in transition))
         );
+      const eventText =
+        event === NULL_EVENT ? 'the transient event' : `event '${event}'`;
+
+      warn(
+        !hasNonLastUnguardedTarget,
+        `One or more transitions for ${eventText} on state '${this.id}' are unreachable. ` +
+          `Make sure that the default transition is the last one defined.`
+      );
+    };
+
+    let onConfig: Array<
+      TransitionConfig<TContext, EventObject> & {
+        event: string;
+      }
+    >;
+
+    if (!this.config.on) {
+      onConfig = [];
+    } else if (Array.isArray(this.config.on)) {
+      onConfig = this.config.on;
+    } else {
+      const {
+        [WILDCARD]: wildcardConfigs = [],
+        ...strictOnConfigs
+      } = this.config.on;
+
+      onConfig = flatten(
+        keys(strictOnConfigs)
+          .map(key => {
+            const mapped = toTransitionConfigArray<TContext, EventObject>(
+              key,
+              strictOnConfigs![key as string]
+            );
+            if (!IS_PRODUCTION) {
+              validateMappedTransitions(key, mapped);
+            }
+            return mapped;
+          })
+          .concat(
+            toTransitionConfigArray(WILDCARD, wildcardConfigs as SingleOrArray<
+              TransitionConfig<TContext, EventObject> & {
+                event: '*';
+              }
+            >)
+          )
+      );
+    }
 
     const doneConfig = this.config.onDone
       ? toTransitionConfigArray(String(done(this.id)), this.config.onDone)
@@ -1955,7 +2009,7 @@ class StateNode<
     const delayedTransitions = this.after;
 
     let formattedTransitions = flatten(
-      [...onConfig, ...doneConfig, ...invokeConfig].map(
+      [...doneConfig, ...invokeConfig, ...onConfig].map(
         (
           transitionConfig: TransitionConfig<TContext, TEvent> & {
             event: TEvent['type'] | NullEvent['type'] | '*';
--- a/src/StateNode.ts
+++ b/src/StateNode.ts
@@ -120,6 +120,33 @@ const createDefaultOptions = <TContext>(): MachineOptions<TContext, any> => ({
   delays: {}
 });
 
+const validateArrayifiedTransitions = <TContext>(
+  stateNode: StateNode<any, any, any>,
+  event: string,
+  transitions: Array<
+    TransitionConfig<TContext, EventObject> & {
+      event: string;
+    }
+  >
+) => {
+  const hasNonLastUnguardedTarget = transitions
+    .slice(0, -1)
+    .some(
+      transition =>
+        isString(transition.target) ||
+        isMachine(transition.target) ||
+        (!('cond' in transition) && !('in' in transition))
+    );
+  const eventText =
+    event === NULL_EVENT ? 'the transient event' : `event '${event}'`;
+
+  warn(
+    !hasNonLastUnguardedTarget,
+    `One or more transitions for ${eventText} on state '${stateNode.id}' are unreachable. ` +
+      `Make sure that the default transition is the last one defined.`
+  );
+};
+
 class StateNode<
   TContext = any,
   TStateSchema extends StateSchema = any,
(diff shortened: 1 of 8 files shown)


src/StateNode.ts
```diff
@@ -500,88 +510,56 @@ class StateNode<
...
-      })
-    );
+      const eventType = after(delayRef, this.id);
 
-    allDelayedTransitions.sort((a, b) =>
-      isString(a) || isString(b) ? 0 : +a.delay - +b.delay
-    );
+      // TODO: check if using delayRef here (instead of delay) has fixed a bug
+      // check why we can't use delayRef here, how resolving of this works
+      // seems a little bit scattered across the codebase and inconsistent
```
davidkpiano: We can use `delayRef` I believe, it should work the same.

Andarist: It was what I did at first 😅 unfortunately it didn't work because we can't later (right now) retrieve original function for "inline" function delays covered by those lines:
https://github.com/davidkpiano/xstate/blob/8621b9db28bbb2ccc979b7775c1cea71ac5e022a/src/StateNode.ts#L533-L536
The entry in delays map is created by mutating `this.options.delays` (in kinda "target" state) **BUT** resolveSend gets called from "origin" (for a transition) state here: 
https://github.com/davidkpiano/xstate/blob/8621b9db28bbb2ccc979b7775c1cea71ac5e022a/src/StateNode.ts#L1160-L1164
and each~ state has its own copy of options, because they are not correctly forwarded i.e. here:
https://github.com/davidkpiano/xstate/blob/8621b9db28bbb2ccc979b7775c1cea71ac5e022a/src/StateNode.ts#L300

This might be a bug on master, gonna investigate how it behaves to see why this has slipped unnoticed. However - I'm unsure if mutating options is a good strategy for this anyway. I wouldn't expect those to be inserted there.

Andarist: After doing a quick investigation - luckily there is no bug on master but the implementatiin works sort of by accident in comparison to other similar stuff. Dont have time right now to describe in detail - gonna do that later.

Andarist:
Also, a thing to consider - after some thinking it seems that we might want to special-case `*` when converting from object config to array definitions (latter are later processed in order) and put them at the end of those definitions. So basically when using object configs (default ones) a wildcard should never take precedence over explicit~ event type. Thoughts?

davidkpiano:
> So basically when using object configs (default ones) a wildcard should never take precedence over explicit~ event type. Thoughts?

Yes - object configs should be assumed to have no order, so the safest thing to do here would be to put wildcards at the end.

For array configs, they should just maintain their position.

Andarist:
@davidkpiano this is nearly done on my side, the only things left to do are:
- review
- TS-related stuff

As to the latter - if you have any recommendations regarding `TransitionConfig` drama 😉 I'm all ears. With all of those mapped types and trying to satisfy strict requirements in type system it's quite cumbersome to handle some cases. Maybe something to think about in the future (although I'm not sure what can really be done, it is what it is).

davidkpiano:
This? 

> cleanup `TransitionConfig` type drama (at different places it's intersected with `{event: TEvent["type"]}` or `{ event?: undefined }`

What exactly are the current typing issues right now?

Andarist:
Yes, this. I refer to what I have done - not something that was here before this PR.

I've left `TransitionConfig` roughly untouched, but actually I mostly use it in combination with object type containing event property. For object configs I make this:
https://github.com/davidkpiano/xstate/blob/dcee831a084133bcc203b49c72f5f5df3a99dfee/src/types.ts#L318-L323
because of correct event coming from config map key, so using event as part of TransitionConfig is most likely a bug.

OTOH array configs needs to specify event alongside TransitionConfig (intersection):
https://github.com/davidkpiano/xstate/blob/dcee831a084133bcc203b49c72f5f5df3a99dfee/src/types.ts#L327-L336

And the latter pattern spreads through the codebase, which is not that nice as it requires being hand-written and is surprising to see - but I couldn't figure out a good name for it. Examples:
- https://github.com/davidkpiano/xstate/blob/dcee831a084133bcc203b49c72f5f5df3a99dfee/src/StateNode.ts#L126-L130
- https://github.com/davidkpiano/xstate/blob/dcee831a084133bcc203b49c72f5f5df3a99dfee/src/StateNode.ts#L1902-L1904
- https://github.com/davidkpiano/xstate/blob/dcee831a084133bcc203b49c72f5f5df3a99dfee/src/StateNode.ts#L1964-L1966
- https://github.com/davidkpiano/xstate/blob/dcee831a084133bcc203b49c72f5f5df3a99dfee/src/StateNode.ts#L1993-L1995
- https://github.com/davidkpiano/xstate/blob/dcee831a084133bcc203b49c72f5f5df3a99dfee/src/StateNode.ts#L2033-L2035

As I have gathered those examples I have also noticed right now that I've actually used different type arguments for some of those - which might indicate some inconsistency. At the same time, I'm not sure if we should even bother with typing internals so "correctly" - would be easier to just assume simple strings in place of TEvent["type"] etc. Not sure if we would lose something by doing that.

I don't know - writing generic TS with mapped types is hard 😢

davidkpiano:
> Not sure if we would lose something by doing that.

I don't think we'd lose anything _internally_ so let's make it as simple as possible.

Can we remove `{ event?: undefined }`? I don't think it's necessary to have that restriction and it'll make typing array+map config easier.

I highly doubt someone will do this:

```js
on: {
  SOMETHING: {
    event: 'SOMETHING_ELSE',
    target: 'foo'
  }
}
```

which is the only edge-case of that.

mogsie:
Forgive me for butting in, but have you explored alternative syntaxes than array syntax?  The document order in SCXML ensures that transitions are evaluated in the order specified; there are other ways of achieving this than changing to / adding an array syntax.  For example you could declare an `order` or a `priority` to transitions.

```
<transition target="pass" event="*"/>
<transition target="fail" event="foo"/>
```

could be written as

```
on: {
  '*': [
    { target: 'pass', order: 1 }
  ],
  foo: [
    { target: 'fail', order: 2 }
  ]
}
```


While:

```
<transition target="fail" event="foo"/>
<transition target="pass" event="*"/>
```

could be expressed as:

```
on: {
  '*': [
    { target: 'pass', order: 2 }
  ],
  foo: [
    { target: 'fail', order: 1 }
  ]
}
```

When an event occurs that could match multiple `on` values, the arrays would be merged, sorted by order.  I believe this could give the same "fidelity" in being able to express the document order of `<transition>` elements.

davidkpiano:
@mogsie Interesting approach; @Andarist this could simplify the typing complexity, especially since this is an edge-case that wouldn't normally happen in real library usage.

We should outline the real-life use cases that necessitate this (I realize it's important for SCXML) and see if a simplified API can cover it.

Andarist:
> Forgive me for butting in

No problem, always happy to discuss alternatives.

I'm not sure however if that would be easier. First of all, it would expose `order` on TransitionConfig type publicly and while we always want to recommend using object orderless configs this is just a recommendation.

To make it work we would have to calculate (or just assume) some default order and sort transitions according to that - this I believe would greatly increase implementation complexity.

I also think that readability would suffer, because right now with array configs the order is implied - but rather explicit, everybody knows how ordering works inside an array. Consider this alternative though:
```
on: {
  '*': [
    { target: 'pass', order: 1 }
  ],
  foo: [
    { target: 'fail' }
  ]
}
```
Which one comes first? It imposes way more cognitive load to figure this out and this is only a very simple example. It gets more complex pretty quickly (and also imagine some of those lacking explicit order):
```
on: {
  '*': [
    { target: 'pass', order: 0 },
    { target: 'other', order: 4 }
  ],
  foo: [
    { target: 'x', order: 1 },
    { target: 'y', order: 3 },
  ],
  bar: [
    { target: 'a', order: 2 },
  ]
}
```

And while we mainly want to even support order-based transitions to satisfy SCXML semantics the readability of the conversion output is very nice to have. The example above doesn't beat top-down reading order in terms of readability, IMHO it's not even close to that.

davidkpiano:
@Andarist Can you merge master into this? I'm getting some merge conflicts.

Andarist:
@davidkpiano done

Andarist:
@davidkpiano have you had time by any chance to take a look at this?

davidkpiano:
Wondering if this should be moved to 5.0 - one thing I didn't consider is that this makes transition lookup O(n) instead of O(1) (for non-conditional cases) which affects all use-cases, even though most use-cases will not depend on order. Thoughts?

Andarist:
I’m caching the candidates, so it should have the same O complexity (after first run) - it would have to be a really big statechart for this to become a problem. And even if it starts to show a perf a problem we can easily rework internals to create cache while formatting transitions (which i havent done to keep code simpler) - and then it should be totally the same for both initial and subsequent runs

Andarist:
@davidkpiano any more thoughts on this one? would love to settle it one way or another (even if it means dropping this entirely for now - although as mentioned above I think this is truly 100% compatible with 4.x line)

PR #649, opened 2019-09-04
Cleanup where created delayRef gets inserted

Andarist:
Regarding - https://github.com/davidkpiano/xstate/pull/647#discussion_r320881094

I'm not the biggest fan of spreading those things onto existing options, but this technique is used at other places, so the same should be done here.

It has worked with that "bug" because function delay got passed down and later resolveSend knew how to deal with a simple function as it has to deal with them for "regular" send actions with delay as function expression. It has never been found in machine.options as it was never added to it.

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -207,12 +207,12 @@ export function resolveSend<TContext, TEvent extends EventObject>(
       : action.event
   );
 
-  let resolvedDelay: number | string | undefined;
+  let resolvedDelay: number | undefined;
   if (isString(action.delay)) {
     const configDelay = delaysMap && delaysMap[action.delay];
     resolvedDelay = isFunction(configDelay)
       ? configDelay(ctx, _event.data, meta)
-      : configDelay || action.delay;
+      : configDelay;
   } else {
     resolvedDelay = isFunction(action.delay)
       ? action.delay(ctx, _event.data, meta)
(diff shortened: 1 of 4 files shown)


src/actions.ts
```diff
@@ -207,12 +207,12 @@ export function resolveSend<TContext, TEvent extends EventObject>(
...
   );
 
-  let resolvedDelay: number | string | undefined;
+  let resolvedDelay: number | undefined;
   if (isString(action.delay)) {
     const configDelay = delaysMap && delaysMap[action.delay];
     resolvedDelay = isFunction(configDelay)
       ? configDelay(ctx, _event.data, meta)
-      : configDelay || action.delay;
+      : configDelay;
```
Andarist: this change is causing this test to fail - https://github.com/davidkpiano/xstate/blob/5f784e1d358d00ecfd4cc6a6f7ade771f4c05b64/test/after.test.ts#L160

I've removed it as something to discuss - not sure if that's really needed. We warn about unresolved delay strings just after resolveSend gets called and if we were not able to resolve something it seems conceptually better to leave that as undefined - which makes the output SendActionObject a "correct" one, instead of preserving the string and having to deal with such possibility later on.

davidkpiano: Yeah, that's true, we should remove this test or make it so that it returns `undefined` as the delay (which should be a `number | undefined`).

Andarist: Changed the test, should be ready to merge now

PR #657, opened 2019-09-08
Rewrite interpreter's test - 'should be unsubscribable' - as it was flaky

Andarist:
The previous version of this test has managed to fail on my machine during development, because it was time-based and `.unsubscribe` call scheduled in setTimeout has been called too late to pass the assertion.


davidkpiano:
Much better, thanks!

PR #664, opened 2019-09-09
Make spawn return Interpreter if called with a machine

### LINKED ISSUE #653: typescript useService(actor)
## Bug or feature request?
Bug?
## Description:
```js
export function Todo({ todoRef }) {
  const [state, send] = useService(todoRef);
  ...
```
On the TodoMVC example, we can call `useService` and pass an actor ( `todoRef`) , but with Typescript `Argument of type 'Actor<...>' is not assignable to parameter of type 'Interpreter<...>'.`

I searched in the comments and saw a [typescript implementation of todoMVC](https://codesandbox.io/s/pmx5j5n2p0), but it ditched the whole conflict and work around the problem, using `useMachine(Machine)` instead of `useService(Actor)`
## (Bug) Expected result:
I would like to use `useService(Actor)` in ts
## (Bug) Actual result:
 `Argument of type 'Actor<...>' is not assignable to parameter of type 'Interpreter<...>'.`
## (Bug) Potential fix:

Sorry I honestly do not know how to solve it, but if there's anything i could help let me know 👍

Andarist:
fixes #653

### DIFF
--- a/packages/xstate-react/test/types.test.tsx
+++ b/packages/xstate-react/test/types.test.tsx
@@ -1,0 +1,55 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { Machine, interpret, assign, spawn, Interpreter } from 'xstate';
+import { useService } from '../src';
+
+describe('useService', () => {
+  it('should accept spawned machine', () => {
+    type TodoCtx = { completed: boolean };
+    type TodosCtx = { todos: Array<Interpreter<TodoCtx>> };
+
+    const todoMachine = Machine<TodoCtx>({
+      context: {
+        completed: false
+      },
+      initial: 'uncompleted',
+      states: {
+        uncompleted: {
+          on: {
+            COMPLETE: 'done'
+          }
+        },
+        done: {
+          entry: assign({ completed: true })
+        }
+      }
+    });
+
+    const todosMachine = Machine<TodosCtx, { type: 'CREATE' }>({
+      context: { todos: [] },
+      initial: 'working',
+      states: { working: {} },
+      on: {
+        CREATE: {
+          actions: assign(ctx => ({
+            ...ctx,
+            todos: ctx.todos.concat(spawn(todoMachine))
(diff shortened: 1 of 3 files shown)


packages/xstate-react/test/types.test.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { Machine, interpret, assign, spawn, Interpreter } from 'xstate';
+import { useService } from '../src';
+
+describe('useService', () => {
+  it('should accept spawned machine', () => {
```
Andarist: it's somewhat verbose, but I've wanted to replicate the whole situation - from spawning to using service 🤷‍♂

test/actor.test.ts
```diff
@@ -408,7 +408,7 @@ describe('actors', () => {
         initial: 'initial',
         states: {
           initial: {
-            entry: assign(() => ({
+            entry: assign((): { serverRef: Actor } => ({
```
Andarist: quite frankly I believe this shouldnt be needed but TS failed to "lose" information about the inferred type (Interpreter) when matching against expected one (Actor), even if Interpreter extends Actor

src/interpreter.ts
```diff
@@ -1176,7 +1176,7 @@ const resolveSpawnOptions = (nameOrOptions?: string | SpawnOptions) => {
 export function spawn<TContext>(
   entity: StateMachine<TContext, any, any>,
   nameOrOptions?: string | SpawnOptions
-): Actor<State<TContext>>;
+): Interpreter<TContext>;
```
davidkpiano: This should remain `Actor` because `Interpreter` is not the only thing that can be spawned.

Andarist: This is just an overload - other spawnable types return Actor

davidkpiano: Ah, I see now 😅 Thanks!

PR #670, opened 2019-09-18
Fixed callback actors subscribers being subscribed to events forwarded to that actor instead of events emitted by that actor

Andarist:
I'm not sure how to write a test for this which would use just public APIs as there is no example in tests and docs showcasing how `actor.subscribe` should be used. Should the actor ref just be grabbed from the context?

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -962,19 +962,22 @@ export class Interpreter<
   }
   private spawnCallback(callback: InvokeCallback, id: string): Actor {
     let canceled = false;
+    const receivers = new Set<(e: EventObject) => void>();
+    const listeners = new Set<(e: EventObject) => void>();
+
     const receive = (e: TEvent) => {
+      listeners.forEach(listener => listener(e));
       if (canceled) {
         return;
       }
       this.send(e);
     };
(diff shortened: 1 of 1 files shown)


davidkpiano:
Still a little confused about what broke and was fixed here. Can you explain more or show a failing test case? Ready to merge otherwise. ✅

Andarist:
The problem was that callback actor subscriber was subscribing to events sent **to** that callback actor from the service, while it should subscribe to events sent **from** that callback to the service.

This is a semantic used by promise actor:
https://github.com/davidkpiano/xstate/blob/a74af9feb796681e5dc9963ab0ba61fe3b3cff3b/src/interpreter.ts#L926-L928
and observable actors:
https://github.com/davidkpiano/xstate/blob/a74af9feb796681e5dc9963ab0ba61fe3b3cff3b/src/interpreter.ts#L1035-L1037

I would gladly add a test for this - but as mentioned in the thread I'm not completely sure what's the public-facing API for accessing actors and subscribing to them (there are no docs nor tests regarding this). Could you confirm that this should just be done by accessing an actor ref through machine's context?

davidkpiano:
Actually now I understand. LGTM

PR #677, opened 2019-09-23
Remove unused historyValue parameter from getFromRelativePath

Andarist:


src/StateNode.ts
```diff
@@ -1740,9 +1736,10 @@ class StateNode<
     const parent = this.parent!;
 
     if (!historyValue) {
-      return this.target
+      const historyTarget = this.target;
```
Andarist: `this.target` is a getter, so better to run it once here

PR #690, opened 2019-10-02
Fix small typo in transient states example

doup:
Fix misplaced parenthesis `)`.


doup:
Ups, my bad. Although the formatting makes it confusing for me.

PR #691, opened 2019-10-02
Implemented an integration for the XState DevTools browser extension

amitnovick:
My modification of the `interpreter.ts` file which works with the XState DevTools browser extension.

Reference: https://github.com/statecharts/xstate-viz/issues/30

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -164,6 +164,7 @@ export class Interpreter<
 
   // Dev Tools
   private devTools?: any;
+  private xstateDevtoolsExtension?: any;
 
   /**
    * Creates a new Interpreter instance (i.e., service) for the given machine with the provided options, if any.
@@ -249,6 +250,9 @@ export class Interpreter<
     // Dev tools
     if (this.devTools) {
       this.devTools.send(_event.data, state);
+    }
+    if (this.xstateDevtoolsExtension) {
+      this.xstateDevtoolsExtension.send(state);
     }
 
     // Execute listeners
@@ -908,6 +912,9 @@ export class Interpreter<
             if (this.devTools) {
               this.devTools.send(errorEvent, this.state);
             }
+            if (this.xstateDevtoolsExtension) {
+              this.xstateDevtoolsExtension.send(this.state);
+            }
             if (this.machine.strict) {
               // it would be better to always stop the state machine if unhandled
               // exception/promise rejection happens but because we don't want to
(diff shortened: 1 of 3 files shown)


src/interpreter.ts
```diff
@@ -250,6 +251,9 @@ export class Interpreter<
     if (this.devTools) {
       this.devTools.send(_event.data, state);
     }
+    if (this.xstateDevtoolsExtension) {
+      this.xstateDevtoolsExtension.send(state);
```
davidkpiano: How about giving this the same interface as `this.devTools` so that we don't need an extra property?

amitnovick: Well as of now the `this.devTools` property stores a reference to the Redux DevTools extension object (created using `connect(...)`).

If we reuse this property for an additional extension, then we will have a race condition where the two extension's loading onto the page compete for which reference will be written last.

This can happen when a user is alternating between the two extensions.

amitnovick: One way to solve that would be to instead let `this.devTools` point to a collection instead of a singualar extension object. Key-value or a list.

amitnovick: Want me to update this PR based on that suggestion I outlined?

davidkpiano: @amitnovick Sounds good, give it a go!

amitnovick: OK so I tried to think a little more holistically than what my initial attempt and suggestion provided.

`interpreter.ts` currently seems to be doing quite a lot of things, evidenced by its exported values:
- types (no problem here)
- class: Interpreter
- function: `spawn`
- function: `interpret`

So I thought, with some careful separation of logic away from this file we could make it just a little more about "interpreter" and less about what in this PR is about: "browser extension interoperability".

In trying to follow suit with the class-based structure of the codebase, I encapsulated this logic in a class called `BrowserExtensionsManager` with the following refactoring points:
- `this.devtools.connect(...)` -> `new BrowserExtensionsManager(...)`
- `this.devtools.send(...)` -> `this.browserExtensionsManager.update(...)`

This idea is that this class knows about the implementation details necessary for passing the data to every specific extension's interface.

As for why there are still two fields in this new class, this is because it doesn't make sense for the current XState DevTools extensions that I made to receive the `event` value (it doesn't use it), so instead I separated out the interfaces into two variables `reduxDevtoolsObject` and `xstateDevtoolsObject`. It could be in an array or in a map field, but I posit that it wouldn't be much of an improvement.

The `update(...)` method knows about each extension's implementation detail and satisfies them, without enforcing them to conform to the same interface that Redux DevTools has.

# Feedback
Let me know what you think about this... if it's naming choice, design choice or something else.

amitnovick: Made a commit today to implement the removal of a stopped service from the "service selection" dropdown UI, wrote about it on the [issue](https://github.com/statecharts/xstate-viz/issues/30#issuecomment-538691965)

amitnovick:
Closed for now in favour of [a more minimal approach](https://github.com/davidkpiano/xstate/pull/717#event-2702350812), this PR has two things that are missing in the other one:
- it works fine alongside Redux DevTools
- it reports stopped services properly

Those issues should be addressed at a later point in time.

PR #699, opened 2019-10-05
Migrate to Yarn & move core package to separate directory

Andarist:
> error brotli-size@4.0.0: The engine "node" is incompatible with this module. Expected version ">= 10.16.0". Got "8.16.1"

`brotli-size` is used by `rollup-plugin-filesize`, maybe we could drop this plugin right now and incorporate different filesize checking strategy later?


davidkpiano:
We can remove the check for `8.x`.

Andarist:
Ok, I've removed it and CI is passing 🎉 

One thing kinda missing in this PR is this https://github.com/davidkpiano/xstate/pull/699/files#diff-204539c00c95e5bca7a13872531cb52c

Not sure what we want to put there, different projects use different strategies. Most commonly they just put a placeholder/link to docs.

davidkpiano:
Let's keep the original README for now (duplicated between core and the root repo) and then I'll merge this in, yeah? I'll make a new project-wide README.md today.

Andarist:
Ok, I've duplicated the root one. Ready to merge 🚀

PR #702, opened 2019-10-06
Avoid recreating States in .transition when possible

Andarist:
Should yield some perf boost

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -200,7 +200,9 @@ export class Interpreter<
       warn(
         this.initialized,
         // tslint:disable-next-line:max-line-length
-        `Attempted to read initial state from uninitialized service '${this.id}'. Make sure the service is started first.`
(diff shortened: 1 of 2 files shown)


packages/core/src/interpreter.ts
```diff
@@ -200,7 +200,9 @@ export class Interpreter<
       warn(
         this.initialized,
         // tslint:disable-next-line:max-line-length
-        `Attempted to read initial state from uninitialized service '${this.id}'. Make sure the service is started first.`
+        `Attempted to read initial state from uninitialized service '${
+          this.id
+        }'. Make sure the service is started first.`
```
davidkpiano: I wonder why Prettier keeps going back and forth between this

Andarist: not sure, was wondering about the same thing - maybe some version mismatch? but precommit hook should use a stable version, so it's rather unlikely

if that is going to happen more often then I'm going to investigate this as it's super noisy and distracting

PR #726, opened 2019-10-11
Make @xstate/test browser-compatible leveraging package.json#browser

Andarist:
fixes https://github.com/davidkpiano/xstate/issues/725 , still need to test this manually with webpack, but in general I've been using this technique with success in the past


pangratz:
Thanks for the fix @Andarist ❤️ 

Any chance this gets released in a new version of `@xstate/test`? Would love to test this out in a project using QUnit in the browser...

Issue #721, opened 2019-10-10
Transient transition infinite loops when a default condition has no target


Ryan-McMillan:
When trying to check a condition upon entering a state and raise an action if the condition fails it infinite loops in the condition function.

Repro: Paste this into the visualizer https://xstate.js.org/viz/ and run the NEXT action with a balance <= 300 in the context.
```
  // Available variables:
  // - Machine
  // - interpret
  // - assign
  // - send
  // - sendParent
  // - spawn
  // - raise
  // - actions
  // - XState (all XState exports)
  
  const fetchMachine = Machine({
    initial: 'one',
    context: {
      balance: 0
    },
    states: {
      one: {
        on: {
          NEXT : 'two'
        }
      },
      two: {
        on: {
          "" : [
            {
              cond : ({ balance }) => {
                console.log("infinite loop")
                return balance > 300;
              },
              target : "success"
            },
            
            { actions : raise("INSUFFICIENTFUNDS") }
          ],
          INSUFFICIENTFUNDS : "bankrupt",
          SUCCESS           : "success"
        }
      },
      success: {
        type: 'final'
      },
      bankrupt: {
        type: 'final'
      }
    }
  });
  ```

I'd expect this to raise the INSUFFICIENTFUNDS action and transition to bankrupt.

davidkpiano:
This is working as expected because:

1. all transitions are _external_ by default unless otherwise specified (`internal: true`)
2. transient transitions are always taken first.

Try adding `internal: true` to get the behavior you want... or you can transition directly to `'bankrupt'`.

tivac:
In the real code it requires an id to jump to the target state, so raising an event would be way cleaner. I don't understand why the transient transition raising an event causes an infinite loop just yet. Is it possible for you to explain it in a slightly different way?

I tried adding some `internal : true` flags to the events, but I'm still getting an infinite loop.

https://xstate.js.org/viz/?gist=758f2c8511ac4b7849c708364698bbbe

Andarist:
According to the `mainEventLoop` algorithm - https://www.w3.org/TR/scxml/#mainEventLoop - there is no difference in this for external/internal transitions. Basically `selectEventlessTransitions` happens after every `microtask` where `microtask` is a processing of an event (both from external queue and internal queue, it doesn't matter).

The solution here is not to use events in eventless transitions to jump to other states - just use target, like this:
https://codesandbox.io/s/great-cartwright-5fhwq

tivac:
[...]

PR #727, opened 2019-10-11
docs: add warning from #721

tivac:
Information from https://github.com/davidkpiano/xstate/issues/721#issuecomment-540954522, I think this is worth calling out in the docs because I was really surprised at this behavior but according to @Andarist it's part of the SCML spec.

### DIFF
--- a/docs/guides/transitions.md
+++ b/docs/guides/transitions.md
@@ -323,6 +323,12 @@ Just like transitions, transient transitions can be specified as a single transi
 
 Null events are always "sent" for every transition, internal or external.
 
+::: warning
+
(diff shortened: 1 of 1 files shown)


docs/guides/transitions.md
```diff
@@ -323,6 +323,12 @@ Just like transitions, transient transitions can be specified as a single transi
 
 Null events are always "sent" for every transition, internal or external.
 
+::: warning
+
+Transient transitions will be run for every event while the statechart is in the state on which they are defined, so it is important that they use a `target` value instead of simply raising events to avoid creating infinite event loops.
```
Andarist: Hm, the issue itself is not focused on using target over events, but rather about not sending events which are in turn caught (handled) by the state which has that transient transition defined on it - because that cause that problem of transient transitions being checked again.

davidkpiano:
Let's resolve this holistically with better dev-time errors/warnings.

PR #728, opened 2019-10-11
Fix condition for hasNonLastUnguardedTarget warning

Andarist:


packages/core/src/StateNode.ts
```diff
@@ -134,16 +134,18 @@ const validateArrayifiedTransitions = <TContext>(
...
+        !('in' in transition) &&
+        (isString(transition.target) || isMachine(transition.target))
     );
   const eventText =
     event === NULL_EVENT ? 'the transient event' : `event '${event}'`;
 
   warn(
     !hasNonLastUnguardedTarget,
-    `One or more transitions for ${eventText} on state '${stateNode.id}' are unreachable. ` +
+    `One or more transitions for ${eventText} on state '${
```
Andarist: this reformatting is super annoying, tried to look briefly into fixing this - but couldn't locate the problem yet :s

davidkpiano: It's alright, we'll just deal with it 😅

PR #734, opened 2019-10-14
Remove warning about reading an initialState of a service (it's pure so it's safe)

### LINKED ISSUE #732: useMachine logs warning about reading initial state from uninitialized service
**Description**
In 4.7.0-rc4 an warning is logged to the console when calling useMachine.

**Expected Result**
No warning.

**Actual Result**
```Warning: Attempted to read initial state from uninitialized service 'toggle'. Make sure the service is started first. ```

**Reproduction**
https://codesandbox.io/s/xstate-react-template-yo5ui?expanddevtools=1&fontsize=14

**Additional context**
Occurs on 4.7.0-rc4 but did not on 4.6.7.

### PR DESCRIPTION
**Pull request by Andarist**
fixes #732

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -152,10 +152,6 @@ export class Interpreter<
    * Whether the service is started.
    */
   public initialized = false;
-  /**
-   * The initial state of the machine.
-   */
-  private _initialState?: State<TContext, TEvent>;
 
   // Actor
   public parent?: Interpreter<any>;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/core/src/interpreter.ts [hunk 1]
```diff
@@ -152,10 +152,6 @@ export class Interpreter<
    * Whether the service is started.
    */
   public initialized = false;
-  /**
-   * The initial state of the machine.
-   */
-  private _initialState?: State<TContext, TEvent>;
```
Andarist: this was never set

packages/core/src/interpreter.ts
```diff
@@ -197,24 +193,15 @@ export class Interpreter<
...
-      warn(
-        this.initialized,
-        // tslint:disable-next-line:max-line-length
-        `Attempted to read initial state from uninitialized service '${this.id}'. Make sure the service is started first.`
-      );
-    }
-
-    return (
-      this._initialState ||
-      withServiceScope(this, () => this.machine.initialState)
```
Andarist: usage of `withServiceScope` was unnecessary - reading an `initialState` can't escape a "single machine boundary" and it's pure, so it doesn't affect spawn actions etc

davidkpiano: I think there were issues with services being invoked at the initial state (e.g., in an initial `assign()` action) but those should be solved another way, so this is OK

PR #740, opened 2019-10-19
docs: make `useMemo` work in reddit tutorial

mikemoraned:
- make `useMemo` return the machine
- make `useMemo` depend on `name`
- change `data-machine` to get id from `subredditMachine`


davidkpiano:
Thanks much!

PR #745, opened 2019-10-22
Remove Lerna

Andarist:


package.json
```diff
@@ -9,8 +9,6 @@
   },
   "scripts": {
     "preinstall": "node ./scripts/ensure-yarn.js",
-    "clean:packages": "lerna run --parallel clean",
```
Andarist: Don't think those were that much needed. If any issue arise from lack of such script we can always reimplement them on our own.

packages/xstate-graph/package.json
```diff
@@ -34,7 +34,7 @@
     "url": "https://github.com/davidkpiano/xstate/issues"
   },
   "peerDependencies": {
-    "xstate": "^4.7.0"
+    "xstate": "^4.7.0-rc"
```
Andarist: there is no version 4.7.0 yet, so the previous range could not be satisfied

packages/xstate-immer/package.json
```diff
@@ -36,7 +36,7 @@
   },
   "dependencies": {},
   "peerDependencies": {
-    "immer": "^2.1.5",
+    "immer": "^3.2.0",
```
Andarist: we have 3.2.0 in devDeps, so I've aligned them - not sure if this is completely correct now though, this package is not published yet from what I remember so it doesn't matter much i think

packages/xstate-scxml/package.json
```diff
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "xml-js": "^1.6.11",
-    "xstate": "next"
```
Andarist: `next` has caused it to be fetched from npm (instead of using local monorepo version)

Issue #755, opened 2019-10-24
"setCurrent is not a function" for @xstate/react immediate: true + initial state transient transition


johnyanarella:
**Description**

Given a machine whose initial state (ex. `bootstrapping`) contains a transient transition (i.e. immediate null event transition), if the `immediate: true` option is specified for `useMachine()`, a runtime error will occur.

**Expected Result**

The machine starts without incident.

**Actual Result**

The following error is thrown when `useMachine()` attempts to start the machine:

```
TypeError: setCurrent is not a function.
```

because the `onTransition()` handler (triggered by the transient transition during startup) registered a few lines earlier references `setCurrent()` from a `useState()` call that has not yet occurred.

**Reproduction**

Admittedly a contrived example, but the goal here was strip the problem down to its essence:

- [CodeSandbox](https://codesandbox.io/s/xstatereact-initial-transient-transition-immediate-option-fr0xb)

*NOTE:*

- Commenting `immediate: true` in the `useMachine()` call's options will resolve the runtime error but reveal the warning triggered when an event is sent by a descendant component before the service starts.
- Commenting out the `useMachine` import from `@xstate/react` and uncommenting the import for the local `useMachine` implementation will allow you to try out the proposed fix in the PR linked to this issue.

**Additional context**

Encountered in: `XState 4.7.0rc5`

Why am I enabling `immediate: true` anyway? Because the real machine in question (several of which are instantiated at a time) receives events dispatched by descendent components before the `useEffect()` inside `useMachine()` starts the machine, which results in tons of warnings like this in the dev console:

```
Warning: Event "FOO" was sent to uninitialized service "foo" and is deferred. Make sure .start() is called for this service.
```

Generally, I'd probably be fine with deferring these events, but the noisy dev console is problematic.

Andarist:
fixed by https://github.com/davidkpiano/xstate/pull/756

PR #756, opened 2019-10-24
Proposed fix for #755 "setCurrent is not a function"

johnyanarella:
Prior to this change, if the `immediate: true` option was supplied for `useMachine()` and the service had a transient transition in its initial state, the `onTransition()` handler’s reference to `setCurrent` was `undefined` because `useState()` had not yet been executed to populate it, resulting in a "setCurrent is not a function" runtime error on `service.start()`.

I propose resolving this by splitting up the service creation logic such that the `onTransition()` handler is not added until after `useState()` has been executed and `setCurrent` is defined.

See: #755

### DIFF
--- a/packages/xstate-react/src/index.ts
+++ b/packages/xstate-react/src/index.ts
@@ -71,9 +71,16 @@ export function useMachine<TContext, TEvent extends EventObject>(
   const serviceRef = useRef<Interpreter<TContext, any, TEvent> | null>(null);
 
   // Create the service only once
-  const creatingService = serviceRef.current === null;
   if (serviceRef.current === null) {
-    serviceRef.current = interpret(machineRef.current, interpreterOptions);
+    serviceRef.current = interpret(
+      machineRef.current,
+      interpreterOptions
+    ).onTransition(state => {
+      // Update the current machine state when a transition occurs
+      if (state.changed) {
+        setCurrent(state);
+      }
+    });
   }
 
   const service = serviceRef.current;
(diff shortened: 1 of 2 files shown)


packages/xstate-react/src/index.ts
```diff
@@ -93,13 +86,22 @@ export function useMachine<TContext, TEvent extends EventObject>(
   }, [actions]);
 
   // Start service immediately (before mount) if specified in options
-  if (immediate) {
```
Andarist: why a simple reordering of this if statements and following `useState` wouldnt work?

Andarist:
Could you add a test for this?

packages/xstate-react/src/index.ts
johnyanarella: You're right! That was actually the first thing I tried, and at the time I thought it had introduced an issue where the action associated with the null transition was re-executed, but after trying it again I no longer see that occurring.

I'll update this PR to a simple re-ordering of those statements.

johnyanarella:
As astutely observed and suggested by @Andarist, the better solution was just to re-order the `service.start()` and `useState()` statement blocks, rather than split up the service creation logic. Revised the code to implement that approach.

Added a couple unit tests for the `immediate` option, including the scenario described in the original issue.

davidkpiano:
Looks good to me! @Andarist ?

Andarist:
👌 perfect, thanks!

PR #759, opened 2019-10-29
Fixes incorrect placement of console.log

kyleshevlin:
If one reads the source code for `@xstate/fsm`, one can see that the `service.subscribe()` method immediately calls the listener with the current state. Thus, in the example, the console.log would be called immediately after registering the listener. The `start` method, on the other hand, merely returns the service and would log nothing out. Thus, I moved the log comment up to just below the subscription.

I have made a small example showing this behavior in a sandbox: https://codesandbox.io/s/jovial-dirac-fe4fn. Without even starting the service, you can see that the `unlit` state has been logged to the console.

### DIFF
--- a/packages/xstate-fsm/README.md
+++ b/packages/xstate-fsm/README.md
@@ -386,14 +386,14 @@ const lightService = interpret(lightMachine);
 lightService.subscribe(state => {
   console.log(state);
 });
-
-lightService.start();
 // => logs {
 //   value: 'green',
 //   context: { redLights: 0 },
 //   actions: [],
 //   changed: undefined
 // }
+
+lightService.start();
 
 lightService.send('TIMER');
 // => logs {


davidkpiano:
Good catch, although I think that the listener should be called when the service is started, never before. I'll see about fixing that nuance instead.

kyleshevlin:
I think that makes sense. I've been writing an interpreter from scratch for educational purposes and referencing this source code. I made a modification in my interpreter where I fire all the listeners on calling `start` instead.

In this case, it would be removing line 181 `listener(state)`, and then making line 187:

```javascript
{
  //...
  start: () => (
    (started = true),
    listeners.forEach(listener => listener(state)),
    service
  ),
  //...
}
```

My concern was if there was a good reason I hadn't understood that we should fire them when subscribed, or if `start` should not have any other side effect like that. Sounds like this is more of a bug and not a feature.

PR #769, opened 2019-10-31
fix(@xstate/react) keep services up to date

### LINKED ISSUE #765: stale context, services, etc.
**Description**
When using `useMachine` and configuring the machine inside of a react component based on props, things get stale on subsequent renders. The xstate context for example is not updated, causing actions to use stale context.

**Expected Result**
useMachine should keep the context, services, etc. up to date when they change. Possibly by providing an array of deps option like react's `useState` does. 

**Reproduction**
https://codesandbox.io/s/xstate-react-stale-xpd17

geddski:
resolves #765

### DIFF
--- a/packages/xstate-react/src/index.ts
+++ b/packages/xstate-react/src/index.ts
@@ -90,12 +90,16 @@ export function useMachine<TContext, TEvent extends EventObject>(
 
   const service = serviceRef.current;
 
-  // Make sure actions are kept updated when they change.
+  // Make sure actions and services are kept updated when they change.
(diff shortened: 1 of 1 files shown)


Andarist:
This is OK as ad-hoc solution - but I think in the long term we should allow for swapping out the config object entirely. This could be done unconditionally in `useEffect` and would be performant as we would just reassign a single object property. For this to work we would have to stop mutating the config object internally and just use things as is. This would also fix the same issue for all other types of config - including guards etc. cc @davidkpiano

geddski:
@Andarist I'd agree with that.

PR #776, opened 2019-11-04
Use correct property name for send action options

allforabit:
Should be `to` but is `target` in the doc string.


davidkpiano:
Good catch, thank you (`target` is for SCXML)

Issue #787, opened 2019-11-07
Process references


ZempTime:
**Description**
In `packages/core/src/environment.ts` and `packages/xstate-fsm/src/index.ts` there are references to `process`. Since a web browser does not have this variable, you can not load xstate using ES modules in a browser.

**Expected Result**
I can load xstate using ES modules in a browser.

**Actual Result**
`process is not defined`

**Additional context**
Same issue: https://github.com/davidkpiano/xstate/issues/227

ZempTime:
As I was looking through `node_modules`, I noticed `dist/xstate.web.js`. Updating to:
```
import { interpret } from 'xstate/dist/xstate.web.js';
```
Also produced a working state. I'll leave this open for now because it's not totally clear to me if I'm supposed to be using this or if it's just for `unpkg`.

If this is ok, I'd consider this issue closed and #788 decline-able. I'd also like a callout in the getting started section of the docs (which I'd be happy to pr!)

Andarist:
Some context on why those are left in like this in both source files and regular dist files can be found:
- here: https://github.com/davidkpiano/xstate/pull/371
- and here: https://github.com/davidkpiano/xstate/issues/706

Could I ask what's your stack, how do you utilize xstate (and other npm dependencies), are you primary a frontend or a backend developer? I'm literally very curious about this without any bad intentions. I maintain multiple fairly popular packages, many of them using this technique and in most of them, I've witnessed near-zero reports about this causing any issues. On the other hand in XState there were already multiple reports of this being some kind of a problem - so this leads me to believe that the majority of the userbase in my other projects are webpack (or other bundlers) users, most commonly with many dependencies consumed directly from npm, but some people coming to XState have different background and are getting blind-sighted  by this.

To answer the original problem - that `xstate/dist/xstate.web.js` is supposed to be used if you want to consume this directly in the browser from CDN as an ES module. If you need similar bundle fro `@xstate/fsm` then please let me know and I'll prepare the build configuration for it.

ZempTime:
**Issue Resolution**
If using `web-` notated files inside `dist` are the supported way, we're good. Hadn't been aware of the convention but that's really straightforward. Optionally excepting a doc callout, I'd be fine to decline the pr and close this issue.

**"Could I Ask" Questions**
Absolutely.

> what's your stack?

[...]

PR #788, opened 2019-11-07
Update reference to process to be browser-friendly

ZempTime:
Addresses #787 with same approach as https://github.com/davidkpiano/xstate/commit/675ef0f06d4bfb3bbeb290e8353d63c43a24688a


Andarist:
This is unfortunately declineable. Those references are meant to be there to be later replaced by bundlers. I'll follow up on this in the issue you have created.

PR #797, opened 2019-11-13
Produce plans with multiple visits

arturohernandez10:
Here is the new feature coded. Tests are coded and running. Fixed some issues I found with running tests. Maybe you can give me some feedback, and I'll add to the documentation once we are ready.

getAlternatePathPlans('state', {maxRevisits: 1})


davidkpiano:
Looks good! Can you explain more how alternate paths are different than the other generated paths?

arturohernandez10:
We are testing our shopping cart. And we want to be able to include scenarios where a person goes back to a previous state and then continues to the checkout. But we want to get all possible combinations within reason. By using getAlternatePathPlans and specifying revisits = 2. We could cover the navigation very easily. I'm thinking the function name is not very good. But I have not found a better one.

davidkpiano:
Gotcha. I have an idea -- for maximum flexibility, do you want to create a public `.getPlans(plansGenerator)` method that allows developers to pass in any plan generator they want?

```js
import { alternatePathPlans } from '../wherever';

testModel.getPlans(alternatePathPlans);
// called internally with:
// return alternatePathPlans(this.machine, this.options)
```

Ideally, the signature should be:

```js
type PlanGenerator = (machine: StateMachine, options: TestModelOptions) => TestPlan[];
```

Or something like that.

arturohernandez10:
I have to digest that some more to be sure. One other possibility is to do both. Have a plan generator interface and add revisits to getSimplePathPlans.

For some simple websites. It may not be warranted to create a plan generator function. And plugins may be farther away.

For complex cases. Because of the possibility of state explosion, pruning early makes a lot of sense. For those a visitor pattern may be a better interface. After that visiting order/semi-order and limiting function would be icing on the cake.

davidkpiano:
Closing as this will be possible with a custom `stateSerializer` as part of the work in https://github.com/statelyai/xstate/pull/3036

PR #799, opened 2019-11-14
Use workaround to fix #744 -- documentation search

### LINKED ISSUE #744: Documentation Search issue
## Description: 
1. Search for text in the documentation - https://xstate.js.org/docs/ .
2. Click on any of the link present

## (Bug) Expected result: Link should open.
URl: https://xstate.js.org/docs/guides/parallel.html#parallel-state-nodes
## (Bug) Actual result: We get 404 because it adds the additional 'docs/' path in the url
URI : https://xstate.js.org/docs/docs/guides/parallel.html#parallel-state-nodes

## (Bug) Potential fix: Fix the path reference

**Required!** If you found a bug and can consistently reproduce it, please take a minute to think about what the potential fix might be.
Even if the fix is incorrect, this can help diagnose and find the solution to the issue as quickly as possible.

I also strongly encourage you to [create a pull request](https://github.com/davidkpiano/xstate/pulls) if you are confident in the potential fix.

You can use [CodePen](codepen.io) or [CodeSandbox template](https://codesandbox.io/s/6dsws) to create a reproduction/POC.

drmikecrowe:
See `docs/.vuepress/theme/WORKAROUND.md` for workaround details
Fixes #744

### DIFF
--- a/docs/.vuepress/theme/WORKAROUND.md
+++ b/docs/.vuepress/theme/WORKAROUND.md
@@ -1,0 +1,4 @@
+Workaround based on [this comment](https://github.com/vuejs/vuepress/issues/1802#issuecomment-533890727)
+Imlementation contained in this `.vuepress/theme` directory [details](https://github.com/rundeck/docs/commit/1e0f1cd066ee4dc1007c34a41b86929443f00fba)
+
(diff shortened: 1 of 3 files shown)


davidkpiano:
Great, thank you! Let's remember to remove this when the fix is released.

PR #801, opened 2019-11-14
feat(core): TypeScript 3.7

davidkpiano:
This PR adds TypeScript 3.7.

Please take a look and see if I forgot anything, or if we can simplify anything! 🙏


packages/core/src/State.ts
```diff
@@ -276,12 +276,9 @@ export class State<
...
     parentStateValue: TSV
-  ): this is State<
-    (TState & { value: TSV })['context'],
-    TEvent,
-    TStateSchema,
-    TState
-  > {
+  ): this is TState extends { value: TSV }
+    ? State<TState['context'], TEvent, TStateSchema, TState>
+    : never {
```
davidkpiano: Finally this works again. This didn't work in TypeScript 3.6

packages/core/test/assign.test.ts
```diff
@@ -77,7 +77,7 @@ const counterMachine = Machine<CounterContext>({
         SET_MAYBE: [
           {
             actions: [
-              assign({
+              assign<CounterContext>({
```
davidkpiano: Trying to figure out why this won't work without the generic. It did before.

package.json
```diff
@@ -54,7 +54,7 @@
     "ts-jest": "^24.1.0",
     "tslint": "^5.11.0",
     "typedoc": "^0.15.1",
-    "typescript": "^3.6.4",
+    "typescript": "^3.7.2",
```
Andarist: you should upgrade `typedoc` as well and check if it's working correctly

packages/core/src/Machine.ts
```diff
@@ -11,7 +11,7 @@ import {
 import { StateNode } from './StateNode';
 
 export function Machine<
-  TContext = any,
+  TContext = unknown,
```
Andarist: I believe the change is good, but it's also a breaking change, are we OK with that? I am 😉

packages/core/src/StateNode.ts
```diff
@@ -357,7 +357,7 @@ class StateNode<
       }
     }
 
-    dfs(this);
+    dfs(this as StateNode);
```
Andarist: shouldn't `dfs` parameter type be adjusted instead? 

Also if you change public `any` in the other file to `unknown` then I think the same thing could happen here for StateNode.

packages/core/src/interpreter.ts
```diff
@@ -442,7 +442,7 @@ export class Interpreter<
       return initialState === undefined
         ? this.machine.initialState
         : isState(initialState)
-        ? this.machine.resolveState(initialState)
+        ? this.machine.resolveState(initialState as State<TContext, TEvent>)
```
Andarist: why this requires assertion? Couldn't `isState` be improved instead?

packages/core/test/actor.test.ts
```diff
@@ -114,7 +114,7 @@ describe('spawning machines', () => {
       init: {
         entry: [
           assign({
-            server: () => spawn(serverMachine)
+            server: () => spawn(serverMachine) as Actor
```
Andarist: this should be fixable on our side, right? we shouldn't require people using `as` assertion to use `spawn`

packages/core/test/assign.test.ts
Andarist: I don't understand this either - I've tried to understand how TS infers this kind of information behavior and I've failed. There was some other case before then inference in a similar position has failed but I couldn't make sense out of it. Would be great to get help from some TS expert in explaining such things to us :<

packages/core/test/match.test.ts
```diff
@@ -338,6 +338,8 @@ describe('matches() method', () => {
 
       // never called - it's okay if the name is undefined
       expect(name).toBeTruthy();
+    } else if (init.matches('loading')) {
```
Andarist: this test is supposed to test if things compile OK - would be good to add some comment here, inside that if statement to explain what is being tested in here, what specific use case

packages/core/src/Machine.ts
davidkpiano: Hmm good catch, maybe not right now. Let's save this for V5

packages/core/test/actor.test.ts
davidkpiano: The problem is that it doesn't think an `Interpreter` is an `Actor`, even though `Interpreter implements Actor`. Any thoughts?

Andarist: Hm, seems like this should be a classic Animal vs Dog class problem and that Dog should always be assignable to Animal - such assignment should just lose extra (Dog-specific) information. If this fails for this reason - then something would be quite wrong with how TS interprets this? I might miss something about this situation of course.

Maybe you could try out if you can specify return type to both Actor & Interpreter (instead of asserting the return expression value) to see if it makes TS any happier? Maybe that would tell us something more about how it treats this? 🤷‍♂Alternatively, you could also try to specify a generic argument for `assign`. This still wouldn't improve the current situation anyhow, just maybe would help us to understand this.

The separate question is if it's a good thing that `spawn` returns different types. Maybe the could be unified? It seems though that accessing interpreter-specific stuff is valuable 🤔 So not sure how this could be simplified right now - although we maybe should assess how this can be accessed and if it's a good thing to access a child's instance? This could make also a great docs section - I think people might be a little bit confused about this (including myself 😉 )

davidkpiano: Resolved by making `state?` optional in `Actor`

packages/core/src/interpreter.ts
```diff
@@ -878,6 +880,7 @@ export class Interpreter<
     } else if (isObservable<TEvent>(entity)) {
       return this.spawnObservable(entity, name);
     } else if (isMachine(entity)) {
+      // @ts-ignore
```
Andarist: Is a reason why this is needed to be known? could we add a comment about it?

packages/core/test/actor.test.ts
Andarist: if the issue got resolved, this assertion can be removed now

packages/core/src/interpreter.ts
davidkpiano: I cannot get this working currently... this was working in the previous TS version so I still need to figure this out.

packages/core/test/assign.test.ts
davidkpiano: We'll defer this for now, there's already a TypeScript note in the docs re: this.

Andarist:
Other than the fact that CI is still failing - the changes here are good.

davidkpiano:
@Andarist I'm going to merge this in and try to figure out the test failures in another commit.

PR #803, opened 2019-11-16
DRYs up transition failure into a single object

kyleshevlin:
These two objects are the same, so I thought it might be useful to DRY them up into a single object and return that instead. Unsure of its impact on the bundle size.

### DIFF
--- a/packages/xstate-fsm/src/index.ts
+++ b/packages/xstate-fsm/src/index.ts
@@ -80,19 +80,21 @@ export function createMachine<
           );
         }
       }
+  
+      const transitionFailure = {
+        value,
+        context,
+        actions: [],
+        changed: false,
+        matches: createMatcher(value)
+      };
 
       if (stateConfig.on) {
         const transitions = toArray(stateConfig.on[eventObject.type]);
 
         for (const transition of transitions) {
           if (transition === undefined) {
-            return {
-              value,
-              context,
-              actions: [],
-              changed: false,
-              matches: createMatcher(value)
-            };
+            return transitionFailure
           }
 
           const { target, actions = [], cond = () => true } =
@@ -145,13 +147,7 @@ export function createMachine<
       }
 
       // No transitions match
-      return {
-        value,
-        context,
-        actions: [],
-        changed: false,
-        matches: createMatcher(value)
-      };
+      return transitionFailure
     }
   };
 }


packages/xstate-fsm/src/index.ts
```diff
@@ -80,19 +80,21 @@ export function createMachine<
           );
         }
       }
+  
+      const transitionFailure = {
```
Andarist: I'm a little bit concerned that this gets constructed each time - rather than just within "failing" branches of the code. Not from perf perspective - but a readability/logic one. 

I wouldn't consider creating an actual Error up front and throwing it conditionally in few failing branches helping the readability and this seems very similar to this.

davidkpiano: Would be necessary (to some extent) anyway because of `createMatcher(value)`. We can refactor in a separate PR if needed 👍

Andarist: > Would be necessary (to some extent) anyway because of createMatcher(value)

Not sure if I understand this. Could you rephrase?

Andarist: Also - the variable name is confusing. It's not that a transition has failed - there is just no matching transition to be taken (it's not a bad thing and `failure` indicated such)

kyleshevlin: Legibly, I'm not sure why it's more difficult. I think it's clearer that this is the result that happens when either a `transition` is `undefined`, or no transition is successfully taken (all conditions fail). It's why I called it `transitionFailure`, though I'm not opposed to naming it something else if you find it inaccurate.

Andarist: The comment about the chosen naming was nit-picking on my side a little bit (I still believe it's sound but it's a minor thing).

I'm still opposed to the change itself though (in its current form). I find it less readable overall because this value is branch-specific (code-wise) and it's put into a common path, I think in general such things should be avoided when possible. It's like preparing for something and then realizing it's not needed and discarding a prepared thing. For me, personally, this brings some mental overhead to this - when reading the code from top to the bottom. Here this is also a minor thing but I don't like this concept in general. 

I realize this might be my personal preference and if you both are OK with it I won't insist on this or anything. Just my 2 cents.

PR #806, opened 2019-11-19
docs: fix reddit tutorial code

dougmacklin:
The final `Subreddit.jsx` looks like:

```
const Subreddit = ({ service }) => {
  const [current, send] = useService(service);

  // ... same code as previous Subreddit component
};
```

However this is slightly misleading as the `data-machine` attribute on the `<section>` element also needs to be updated to `data-machine={service.machine.id}`

### DIFF
--- a/docs/tutorials/reddit.md
+++ b/docs/tutorials/reddit.md
@@ -511,7 +511,16 @@ In other words, spawned child actors updating will _not_ cause unnecessary reren
 const Subreddit = ({ service }) => {
   const [current, send] = useService(service);
 
-  // ... same code as previous Subreddit component
+  /* ... */
+
+  return (
+    <section
+      data-machine={service.machine.id}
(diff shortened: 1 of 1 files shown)


davidkpiano:
Thanks, but the `data-machine` property is just for displaying the state in the CodeSandbox example, and is not required (or even necessary) for this tutorial.

PerfectPixel:
@davidkpiano Just a comment on this as I just finished that tutorial (thanks again for the help on twitter).

From an outside perspective it is unclear what `data-machine` or `data-state` are used for. My first though was that there is a connection between xstate and the DOM. After digging a little in the docs and the source, I realized that is has no connection to xstate and the example and I found this PR.

Maybe it would be helpful to remove both attributes from the example. Or, if you want to keep the source the same as in the sandbox, to add a small note that those are only for the sandbox and can be ignored.

PR #808, opened 2019-11-19
Stringify condition name in guards docs

PaulRBerg:
As a new user of xstate, I ended up defining all of my guard functions as non-strings because of this inconsistency in the docs. This PR makes `searchValid` a string, as it should be, according to the definition found in the same file:

> A condition function (also known as a guard) specified on the .cond property of a transition, as a **string** or condition object with a { type: '...' } property, and takes 3 arguments...

### DIFF
--- a/docs/guides/guards.md
+++ b/docs/guides/guards.md
@@ -46,7 +46,7 @@ const searchMachine = Machine(
             {
               target: 'searching',
               // Only transition to 'searching' if the guard (cond) evaluates to true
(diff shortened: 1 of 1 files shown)


davidkpiano:
Named functions are valid, as their stringified version is derived from `function.name`. See this example: https://xstate.js.org/viz/?gist=9951c4e89122672f92a59bcf72d4ea61

I'll clarify this in the docs.

PR #811, opened 2019-11-20
@xstate/fsm Expose machine config on createMachine and state config on state

### LINKED ISSUE #810: @xstate/fsm Return config object from createMachine
I'd like to see @xstate/fsm expose the config object it was instantiated with. The most obvious way is by exposing a `config` property on the return value of `createMachine`.

ghengeveld:
Fixes #810

This adds two things:
- On the return value of `createMachine` it adds `config`, which is the `fsmConfig` you passed to `createMachine`, simply returned verbatim.
- On state objects it also adds `config`, which is the state config object for this state, as defined on `fsmConfig`.


packages/xstate-fsm/src/index.ts
```diff
@@ -141,6 +148,7 @@ export function createMachine<
 
             return {
               value: target,
+              config: nextStateConfig,
```
Andarist: shouldn't at least this be easily retrievable by `value` from the config?

Andarist:
I'm not a particular fan of this - especially I don't like exposing state configs. This accommodates a specific use case - even if the use case itself sounds cool - and the point of `@xstate/fsm` is to be a minimal version of xstate, so features should be added to it sparingly.

I could get on board with exposing just the `fsmConfig` - given that it would be a one-line change, but truth to be told I think your use case could be built without exposing this. From what I understand you still require some kind of integration from a user - so why not just require the config to be passed to you directly? Rather a created machine?

ghengeveld:
I'd be fine with having only the fsmConfig. I intentionally built the two things in separate commits.

Andarist:
What about this:
> From what I understand you still require some kind of integration from a user - so why not just require the config to be passed to you directly? Rather a created machine?

Also - will your addon work with both `xstate` and `@xstate/fsm`?

ghengeveld:
It's guesswork at this point. Perhaps we can make it work if they pass only the config. I have no intention of making it work with XState because this is for frontend components where XState is just overkill, but perhaps it'll just work.

Maybe I just got overexcited at the thought of making a simple, elegant contribution to xstate/fsm. I was expecting a more welcoming response.

Andarist:
> Maybe I just got overexcited at the thought of making a simple, elegant contribution to xstate/fsm. I know you're just guarding against feature creep. I had hoped for a welcoming response.

Please don't take me as not welcoming, I have every intention of being as welcoming as I can. We love people using XState and I love helping people with issues, PRs and all. It's just a matter of evaluating cons/pros of each thing to keep a lean API. I also haven't yet declined this altogether - just discussing and asking if alternatives have been considered and what cons are they present to you. Also waiting for @davidkpiano's opinion in this one.

ghengeveld:
Yeah I get that, no problem. Wasn't necessary to confront you with that. You have the tough job of considering the bigger picture (which I can't here) and you take it seriously. I know how it is.

In fact you have me doubting whether this change is really necessary. I think for my goals it's not, so it's about whether you think this change is useful in general. Let's see what David says.

packages/xstate-fsm/src/index.ts
davidkpiano: Agreed, especially because there will be some confusion between this `config` property, and XState (core) `state.configuration`, which is the configuration (set) of state nodes active for that state. Can we remove this?

ghengeveld: Done. I reverted my 2nd commit, so now it's just the config on the machine.

Andarist:
@ghengeveld Sorry for the initial pushback and no response in a while. We've decided to finally merge this in as it became an essential feature for `@xstate/react/fsm`.

ghengeveld:
👍

PR #817, opened 2019-11-22
docs: add warning for transient state nodes

paulrberg:
I shot myself in the foot with transient state nodes, thinking that the event that originates a transition is available in all conditions and actions defined in the null event.

This is probably obvious for some, but I think it's a good warning to have in the docs.

### DIFF
--- a/docs/guides/statenodes.md
+++ b/docs/guides/statenodes.md
@@ -213,6 +213,12 @@ const timeOfDayService = interpret(timeOfDayMachine
 
 <iframe src="https://xstate.js.org/viz/?gist=ca6a3f84f585c3e9cd6aadc3ae00b886&embed=1"></iframe>
 
+::: warning
+The event object of a transition to a transient state is not available in the conditions or actions
+defined under the null event. In the example above, `isBeforeNoon` and `isBeforeSix` would receive `{ type: '' }` as the
+second parameter, not the initial event type and data that transitioned to `unknown`.
+:::
+
 ## State Node Meta Data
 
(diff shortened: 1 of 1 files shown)


davidkpiano:
Hmm, I feel like the null event should probably include the event data. Do you have a use-case where you would expect this? Might be a bug/missing feature.

PaulRBerg:
I created a [gist](https://gist.github.com/PaulRBerg/b20c36c5a5e4d697db1a89db67c0867e#file-inputmachine-js-L107) for you. See `UPDATE_AMOUNT` on line 107, which triggers the `editing` transient state, which performs a few checks for the input value to be valid.

Andarist:
Could you fork this codesandbox - https://codesandbox.io/s/heuristic-dawn-w6fnp (created based on your gist) - and create a situation there which you observe?

An important note is also that currently you indeed might get NULL event in guards, actions, etc but in the future this won't be the case - NULL event doesn't exist (kinda) and should never be surfaced to user's code. In the future (to be compliant with SCXML) you will be given an event that leads to transient state evaluation (so in this case you would get `UPDATE_AMOUNT` event)..

PaulRBerg:
Here's the [CodeSandbox](https://codesandbox.io/s/aged-darkness-sqi1h). See line 25 in `machine.js`.

The logs for the value of the input string in both the context and event objects:

Null Event Logs

davidkpiano:
This was (is) buggy behavior in v4 and will be fixed in v5: https://github.com/statelyai/xstate/pull/2824

PR #825, opened 2019-11-26
Refactor how transitions are resolved

Andarist:
This is my refactoring I have started few weeks ago. Getting slowly back to it but thought it might be worth pushing this out (even though this is completely broken right now) as it should both motivate me to move this forward and also get initial thoughts on this. 

The point of this refactor is to fix a few problems with SCXML-compatibility and also allow implementing new, minor, features from the SCXML spec - I expect this would resolve a bunch of reported issues.

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -31,7 +31,6 @@ import {
   Event,
   StateValue,
   TransitionConfig,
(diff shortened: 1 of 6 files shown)


davidkpiano:
I believe this work has largely been done already in the `next` branch.

PR #827, opened 2019-11-26
fix(@xstate/fsm): Execute initial actions

### LINKED ISSUE #826: entry never fires on initial state
**Description**
The `entry` function never fires on the initial machine state.

**Reproduction**

`test.js`:
```javascript
import { createMachine, interpret } from '@xstate/fsm';


const beaconMachine = createMachine({
    initial: 'uninitialized',
    states: {
        uninitialized: {
        	entry: function () {
    			console.log('I never run :( ');
    		}
        }
    }
});

const beaconService = interpret(beaconMachine).start();

```

I run `node -r esm test.js` and see no console output

Andarist:
Fixes #826

- [x] write test
- [x] establish `.exec` type

### DIFF
--- a/packages/xstate-fsm/src/types.ts
+++ b/packages/xstate-fsm/src/types.ts
@@ -9,6 +9,8 @@ export interface EventObject {
   type: string;
 }
 
+export type InitEvent = { type: 'xstate.init' };
+
 export namespace StateMachine {
   export type Action<TContext, TEvent extends EventObject> =
     | string
@@ -24,7 +26,7 @@ export namespace StateMachine {
 
   export type ActionFunction<TContext, TEvent extends EventObject> = (
     context: TContext,
-    event?: TEvent
+    event: TEvent | InitEvent
   ) => void;
 
   export type AssignAction = 'xstate.assign';
--- a/packages/xstate-fsm/src/types.ts
+++ b/packages/xstate-fsm/src/types.ts
@@ -24,7 +24,7 @@ export namespace StateMachine {
 
   export type ActionFunction<TContext, TEvent extends EventObject> = (
     context: TContext,
-    event: TEvent
+    event?: TEvent
   ) => void;
 
   export type AssignAction = 'xstate.assign';
--- a/.changeset/big-jeans-wait.md
+++ b/.changeset/big-jeans-wait.md
@@ -1,0 +1,5 @@
+---
+'@xstate/fsm': patch
+---
+
+Fixed entry actions defined on an initial state not being executed.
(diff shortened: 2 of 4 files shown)


davidkpiano:
LGTM, merge conflict?

packages/xstate-fsm/src/types.ts
```diff
@@ -17,7 +17,7 @@ export namespace StateMachine {
 
   export interface ActionObject<TContext, TEvent> {
     type: string;
-    exec?: (context: TContext, event: TEvent) => void;
+    exec?: (context: TContext, event?: TEvent) => void;
```
Andarist: @davidkpiano are you ok with this change? this changes it for all actions, but at runtime, it affects only initial actions. It's unfortunate but I don't think there is an easy way to change this at TS level :s

davidkpiano: Might not be a good idea, in XState core we just provide a dummy initial action.

Andarist: > in XState core we just provide a dummy initial action.

Yeah, but we might need to remove this in v5 - if we want to be completely 100% SCXML-compatible.

Based on https://www.w3.org/TR/scxml/#SystemVariables :
> The SCXML Processor must not bind _event at initialization time until the first event is processed. Hence _event is unbound when the state machine starts up.

davidkpiano: The dummy event is more for TS; we can definitely remove it, but requiring devs to check for an undefined event will be a pretty bad developer experience, when it is almost always defined.

Andarist: I agree that this could deteriorate DX, this spreads quite badly.

It would be great though to get concrete examples of when this would become a pain - in places where a concrete event cannot be inferred a developer has to check for `.type` any way to do anything useful with contents of the event and given the fact that TS@3.7 has brought optional chaining to us `event?.type` doesn't look that bad in comparison to `event.type`. Places, where concrete event type can be inferred, wouldn't be affected.

There is also a general question here if we are willing to sacrifice SCXML compatibility in situations like this or is this a definite no go? This is a hard one, both ways have their pros & cons.

davidkpiano: Let's put a dummy event for now (re: offline discussion about SCXML compatibility)

Andarist: Ok, added a dummy event

mreinstein:
Is this PR ready to merge? I would love to see this fix as it means I could remove 1 state from each of my projects that utilzed @xstate/fsm  :-)

Issue #328, opened 2019-01-24
Detecting Final State without starting


timoxley:
# Feature request

Originally discussed here: https://spectrum.chat/statecharts/general/detecting-state-machine-is-done~56dda5bc-8031-4840-96d7-f0458cca12d6

I would like to be able to filter my persisted state machines to only those which have not reached a final state. i.e. no point starting a machine which has already reached its final state. Currently there doesn't appear to be a good way to do this without starting the machine.

Perhaps it could be a method on the `State` prototype e.g.:

```js
const state = State.create(data)
if (state.isFinal()) {
  interpret(machine).start(state)
}
```

However this would prevent filtering db records without instantiating `State` for each of them. A serialisable flag would be preferable e.g.:

```js
if (data.isFinal) {
  interpret(machine).start(data)
}

// queryable from the db level:
const datas = await db.query({ isFinal: false })
datas.forEach(data => interpret(machine).start(data))
```

davidkpiano:
Bikeshedding:

```js
state.done; // true or false

// might be confusing with type: 'final' for individual state nodes
state.final; // true or false
```

PR #832, opened 2019-11-29
fix(core): raise null events on root object, but only if machine is not "done". Fixes #754

### LINKED ISSUE #754: Top-level null events are not fired if current state has no null event transitions
Null transitions defined in the top-level of a machine are not fired if current state has no null transitions.

As far as I understand, the following examples should behave identically. Both of them should transition to 'success' state when 'ADD' event is fired.

However, the first, which has a no-op null transition defined in 'first' state, works.
Second one, which does not have a null transition defined in 'first' state, stays in 'first' state because top-level null transition is not considered.

working version: https://codesandbox.io/s/xstate-example-template-n167o
version with issue: https://codesandbox.io/s/xstate-example-template-u8v9z

davidkpiano:
This PR does 3 things:

- Detects when a machine is "done" internally (`isDone`) and prevents any events from being raised once it is done (according to SCXML spec)
- Raises a null event for a transient transition if the root machine has a transient transition (see #754)
  - Test added for this
- Adds `state.done` (enabled from first point) + tests + docs (#328)

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -1258,11 +1258,13 @@ class StateNode<
         : ({} as Record<string, Actor>)
     );
 
-    const stateNodes = resolvedStateValue
-      ? this.getStateNodes(resolvedStateValue)
+    const resolvedConfiguration = resolvedStateValue
+      ? stateTransition.configuration
+      : currentState
+      ? currentState.configuration
       : [];
 
-    const meta = [this, ...stateNodes].reduce(
+    const meta = resolvedConfiguration.reduce(
       (acc, stateNode) => {
         if (stateNode.meta !== undefined) {
           acc[stateNode.id] = stateNode.meta;
@@ -1271,12 +1273,6 @@ class StateNode<
       },
       {} as Record<string, string>
     );
-
-    const resolvedConfiguration = resolvedStateValue
-      ? stateTransition.configuration
-      : currentState
-      ? currentState.configuration
-      : [];
 
     const isDone = isInFinalState(resolvedConfiguration, this);
 
@@ -1332,7 +1328,8 @@ class StateNode<
 
     if (!isDone) {
       const isTransient =
-        this._transient || stateNodes.some(stateNode => stateNode._transient);
+        this._transient ||
+        configuration.some(stateNode => stateNode._transient);
 
       if (isTransient) {
         maybeNextState = this.resolveRaisedTransition(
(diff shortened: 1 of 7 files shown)


packages/core/test/transient.test.ts
```diff
@@ -398,4 +398,42 @@ describe('transient states (eventless transitions)', () => {
...
+          }
+        },
+        success: {
+          type: 'final'
+        }
+      },
+      on: {
+        '': [
+          {
+            target: '.success',
```
davidkpiano: Without adding the check to see if the machine was "done", this could potentially infinite loop.

The developer needs to make sure proper checks in `cond` are in place to prevent infinite loops when using transient transitions.

Andarist: We could add an explicit test to guard us against introducing that infinite loop in the future. It could be a copy of this machine here - we'd have to move moving from "first" state to "success" state to the inner level (from the root) and have a "dangerous" transient state on the root.

packages/core/src/StateNode.ts
```diff
@@ -1324,25 +1329,29 @@ class StateNode<
...
-      maybeNextState = this.resolveRaisedTransition(
-        maybeNextState,
-        {
-          type: actionTypes.nullEvent
-        },
-        _event
-      );
-    }
+    if (!isDone) {
+      const isTransient =
```
Andarist: I can't remember right now - how `stateNodes` differ from `configuration`? I know what a configuration is, although I don't remember if we keep the root node in it (SCXML doesn't - but not sure right now if we comply to it in this regard).

Previous version of this condition was not checking `this._transient` at all - so wondering if maybe this has fixed also some other bug on nested state nodes (and not only on the root one).

davidkpiano: `configuration` does not include the root node in it - it follows SCXML.

Andarist: what about `stateNodes`?

packages/core/test/transient.test.ts
davidkpiano: Yep I plan to add heuristics for this in a follow-up PR

packages/core/src/StateNode.ts
davidkpiano: We should internally deprecate `stateNodes` for `configuration`, which should be preferred to match the SCXML algorithms.

Andarist: Agreed - that was what i was going for here, wasnt sure how configuration differs from stateNodes so wasnt yet proposing it as I wasnt sure if they are roughly equivalent. Could we then use configuration + root node check for this?

PR #833, opened 2019-11-29
chore: Remove unnecessary lock files

Andarist:


packages/xstate-viz/package.json
```diff
@@ -19,11 +19,6 @@
   "peerDependencies": {
     "react": ">=16"
   },
-  "husky": {
```
Andarist: https://github.com/typicode/husky/tree/next#monorepos

davidkpiano:
Should we just gitignore `yarn.lock` files then?

Andarist:
@davidkpiano I believe we shouldn't - if any lock file pop-ups somewhere in the tree other than at the root then it's an indication that probably the root yarn.lock is corrupted (doesn't reflect the whole repo) so at least not ignoring them gives us a chance to spot the problem. That being said - in most scenarios next `yarn` install repairs the lock, so it's not a big deal. 

In here we had 2 slight problems:
- one package had package-lock.json so this would be fixed by next yarn install
- one package (example) was not included in yarn workspaces, this ain't obligatory but I think it's just nice to have a single yarn.lock at the root (unless for some reason we want to keep them separate), so I've included that example to workspaces setting. This one wouldn't get auto "fixed".

PR #843, opened 2019-12-04
chore: Fix duplicated deps, incorrect dep ranges and introduce manypkg check

Andarist:
This unifies dependencies used across the monorepo - we should always strive to have a single version of a particular direct dependency.

Additionally, this introduces a `manypkg check` as postinstall script which is going to validate this from now on.

### DIFF
--- a/packages/xstate-graph/package.json
+++ b/packages/xstate-graph/package.json
@@ -37,12 +37,11 @@
     "xstate": "^4.7.0"
   },
   "devDependencies": {
-    "@types/jest": "^24.0.11",
     "jest": "^24.8.0",
     "lerna-alias": "3.0.3-0",
-    "ts-jest": "^24.0.2",
+    "ts-jest": "^24.1.9",
     "typescript": "^3.7.2",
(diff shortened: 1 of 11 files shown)


packages/xstate-graph/package.json
```diff
@@ -37,12 +37,11 @@
...
   },
   "devDependencies": {
-    "@types/jest": "^24.0.11",
     "jest": "^24.8.0",
     "lerna-alias": "3.0.3-0",
-    "ts-jest": "^24.0.2",
+    "ts-jest": "^24.1.9",
     "typescript": "^3.7.2",
-    "xstate": "^4.7.0"
+    "xstate": "*"
```
davidkpiano: Well, I learned something new!

Andarist: 😛 this was `manypkg` recommendation - which indeed makes sense, we don't have to worry about version mismatch for local/monorepoed packages in dev dependencies as we always want to use whatever is local and there is no way (at least not currently) for this to affect installations from package manager registries

PR #844, opened 2019-12-04
fix(core): export forwardTo action in index.ts

toadslick:
* exports the `forwardTo` action in `index.ts`
* updates `actions.test.ts` to import `forwardTo` from `src/index` to verify this change


Andarist:
Thanks!

PR #845, opened 2019-12-04
Improve handling of stateMachine updates

Andarist:


packages/xstate-react/src/fsm.ts
```diff
@@ -1,35 +1,70 @@
...
-    ref.current.start();
-
+    service.subscribe(state =>
+      setState(prevState => ({ ...prevState, state }))
+    );
+    service.start();
     return () => {
-      ref.current!.stop();
-      // reset so next call re-initializes
-      ref.current = null;
```
Andarist: previously this really didn't work, because this had only a chance to be called and reset **after** the former if statement, so a new machine had no chance to actually reinitialize until next render. So this actually has been reinitializing with former stateMachine (still kept in ref) rather than with the new one.

packages/xstate-react/src/fsm.ts
```diff
@@ -1,35 +1,70 @@
...
 import { AnyEventObject } from 'xstate';
 
+// TS enums don't tree-shake well
+const NotStarted = 'not started';
+const Running = 'running';
+const Stopped = 'stopped';
+
+type InterpreterStatus = typeof NotStarted | typeof Running | typeof Stopped;
+
+const interpretDev: typeof interpret = machine => {
```
davidkpiano: Is `interpretDev` just for the console error? If so, should this be part of `@xstate/fsm` instead?

Andarist: It's for status tracking and that console.error. This, of course, can be included in `@xstate/fsm` but was not sure if we want to add size to it. Plus in here we can actually give a slightly better error message to the user (for this particular scenario).

davidkpiano: If we make status a normal enum (`0`, `1`, `2`) it should be fine to include.

Andarist: Ok, gonna split this out from this PR and create a new one for `@xstate/fsm` - after it gets merged in we will be able to iterate on this one.

Andarist: I've rebased this PR on top of the master - so I have already access to `service.status` directly, without tracking it in here. 

Question is - do we want to move this console.error to `@xstate/fsm` itself, convert it to an actual thrown error, catch it here in `interpreterDev` to add a bit specific about React & rethrow? Or does this sound too complicated?

Andarist:
Part of this PR is not good (ability to swap a machine). I'm going to close this now and just open a new one when we land or reject https://github.com/davidkpiano/xstate/pull/895 to unify the behavior of both.

PR #846, opened 2019-12-04
fix(rx_compat): Fix compatibility with RxJS interop Symbol.observable

### LINKED ISSUE #841: [4.7.2] The RxJS `from` throw an error if RxJS is imported before XState
**Description**

I see [here](https://github.com/davidkpiano/xstate/blob/285d27d33e9535d5b50f3f43cb5165259b49f488/packages/core/src/utils.ts#L530) that XState creates the `Symbol.observable` if it does not exists but [RxJS does not](https://github.com/ReactiveX/rxjs/blob/7f64088598f42d5a5ec3ebc7c3869ad3c7a74ed4/src/internal/symbol/observable.ts#L11)

This led to the possibility that the RxJS ieff is executed before the XState one and so RxJS will continually try to find `@@observable` in the object.

**Expected Result**

If the `Symbol.observable` does not exists and `@@observable` should be returned.

**Actual Result**

If the `Symbol.observable` does not exists the symbol is created.

**Reproduction**

Unfortunately i'm not able to reproduce the bug on any codesandbox because they all use a polyfill before anything else.

**Additional context**

Andarist:
fixes #841

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -520,23 +520,8 @@ export function isObservable<T>(
   }
 }
 
-export function createSymbolObservable(): symbol | string {
-  let result;
-
-  if (typeof Symbol === 'function') {
-    if (Symbol.observable) {
-      result = Symbol.observable;
-    } else {
-      result = Symbol('observable');
-      // @ts-ignore
-      Symbol.observable = result;
-    }
-  } else {
-    result = '@@observable';
-  }
-
-  return result;
-}
+export const symbolObservable = (() =>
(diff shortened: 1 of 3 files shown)


packages/core/src/utils.ts
```diff
@@ -520,23 +520,8 @@ export function isObservable<T>(
...
-      Symbol.observable = result;
-    }
-  } else {
-    result = '@@observable';
-  }
-
-  return result;
-}
+export const symbolObservable = (() =>
+  (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
```
davidkpiano: Much more succinct, thanks!

PR #847, opened 2019-12-05
fix: Add optional peer dependency for @xstate/fsm in @xstate/react

Andarist:
Strict package managers (like yarn v2 or pnpm) don't you require/import stuff unspecified in deps/peerDeps - all your dependencies have to be explicitly defined.

This change can result in false positive peer dep warnings during installations done by older package managers (or ones that just don't implement optional peer deps), but if we want to keep both of those in a single package (`@xstate/react`) then this is very much needed.

Not sure about all package managers, but yarn supports optional peer deps for quite some time (but maybe just a few months? not sure, would have to be tracked if this has been added this year or sometime before that) and npm has added this very recently.

### DIFF
--- a/packages/xstate-react/package.json
+++ b/packages/xstate-react/package.json
@@ -40,8 +40,17 @@
     "url": "https://github.com/davidkpiano/xstate/issues"
   },
   "peerDependencies": {
+    "@xstate/fsm": "^1.0.0",
     "react": "^16.8.0",
     "xstate": "^4.7.0"
+  },
+  "peerDependenciesMeta": {
+    "@xstate/fsm": {
+      "optional": true
+    },
+    "xstate": {
+      "optional": true
+    }
   },
   "devDependencies": {
     "@testing-library/react": "^8.0.9",


davidkpiano:
Would this break installations for those users who don't have `@xstate/fsm` installed?

Andarist:
> Would this break installations for those users who don't have @xstate/fsm installed?

No, with package managers not aware of optional peer dependencies this would just result in a "false positive" warning printed to the console.

davidkpiano:
@Andarist But now `@xstate/fsm` is marked as a `peerDependency` which will cause warnings.

Andarist:
Yes - when using package managers not aware of optional peer dependencies (current npm & yarn are aware of them so there wouldnt be any warnings), this is what I have called "false positive" warnings. 

Without this optional peerDep we are breaking strict package managers though (like yarn pnp & pnpm) so definitely something that should be fixed. It's not the only way to fix this problem though - we could just not use a single package (`@xstate/react`) for both libraries (`xstate` & `@xstate/fsm`), this would mean though that we'd have to figure out a name for the other package, document this etc.

After looking at this once more I have realized that `xstate` should actually be an optional peerDep as well - after all for `@xstate/fsm` users there should be no requirement of installing main `xstate` package.

Andarist:
@davidkpiano do you agree with this one as being the final solution for this problem?

davidkpiano:
@Andarist Yes, that sounds good.

PR #853, opened 2019-12-07
Add warning when attempting to update an undefined context

andrewgordstewart:
After spending a long time figuring out why my `assign` actions weren't triggered, I thought this warning might be helpful.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -405,7 +405,9 @@ export function updateContext<TContext, TEvent extends EventObject>(
   assignActions: Array<AssignAction<TContext, TEvent>>,
   state?: State<TContext, TEvent>
 ): TContext {
-  warn(!!context, 'Attempting to update undefined context');
+  if (!IS_PRODUCTION) {
+    warn(!!context, 'Attempting to update undefined context');
+  }
   const updatedContext = context
     ? assignActions.reduce((acc, assignAction) => {
         const { assignment } = assignAction as AssignAction<TContext, TEvent>;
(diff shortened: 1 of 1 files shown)


packages/core/src/utils.ts
```diff
@@ -405,6 +405,7 @@ export function updateContext<TContext, TEvent extends EventObject>(
   assignActions: Array<AssignAction<TContext, TEvent>>,
   state?: State<TContext, TEvent>
 ): TContext {
+  warn(!!context, 'Attempting to update undefined context');
```
Andarist: Could you wrap this with env check, like it's done for example here?
https://github.com/davidkpiano/xstate/blob/5eec678629074989bcc45a63ff77dbbc12126775/packages/core/src/interpreter.ts#L220-L225

andrewgordstewart: Actually, `warn` already handles this: `warn` is [initialized](https://github.com/davidkpiano/xstate/blob/master/packages/core/src/utils.ts#L439) as `() => {}`, and is only overwritten when `IS_PRODUCTION` is false.

davidkpiano: Right, but the error message is still in the source code for production builds... I'll have to look into how other libraries like React handle this.

Andarist: React has dropped (or is in the process of dropping) those invariant/warn helpers in favor of no abstraction - just do the env-guarded if statement and throw an error if needed

andrewgordstewart: > Right, but the error message is still in the source code for production builds... I'll have to look into how other libraries like React handle this.

Makes sense. Addressed in 20f5235

PR #854, opened 2019-12-07
Use changesets

Andarist:
I'd like to propose using [Changesets](https://github.com/atlassian/changesets/).

I was working lately with a monorepo that had this configured and I just fell in love with this. The main benefit of using this approach are great changelogs - but besides that, it also automates the release flow. 

**Changelogs**
I've always found using semantic commit conventions rather quirky - very often they are limiting for writing a really good changelog entry and enforce the convention on every committer and it's also not always obvious how to best write one. 

Changesets allow everyone to commit in any way they want and a changelog entry is not tied to a single commit. More than that - a single commit might produce multiple changelog entries. Adding a changeset is a manual process, BUT there is a helpful Github Bot reminding about doing this in each PR. So we can easily educate external contributors to add those, but we can also add them ourselves if needed - even after merging a PR into the master (just yet another benefit of decoupling changelog generation from commits). A changeset can also be rich (it's a markdown file - so we can put code snippets & images there) unlike commit messages.

**Releasing**
When a PR with a changeset (a file) gets merged to master then Changeset Bot creates another PR to the master. It contains the appropriate version bumps (taking care of monorepo inter-dependencies) and committed changelogs. Whenever we merge anything new to master it just force pushes to that created branch. And whenever we decide to merge that PR the Changesets Github Action configured in our repo just releases stuff. The release is always a single click away - which is great - and can be even done from a mobile if needed.


davidkpiano:
Great idea.

PR #858, opened 2019-12-08
Fix batching behaviors - forwarding before processing & make it possible to spawn actors when processing a batch

### LINKED ISSUE #856: Calling batches event changes an actor
Scenario 1: An event is called that triggers a state transition. When the new state is entered, an actor is spawned. In the console, you can see this actor is of type _Interpreter_.

Scenario 2: You call the same event as a batch. The actor is not of type _Interpreter_, rather it is an object. I'm also seeing a warning _Attempted to spawn an Actor (ID: "myActor") outside of a service. This will have no effect._ I'm not sure what is meant by this. 

See https://codesandbox.io/s/machine-spawn-playground-v1b0w

Andarist:
fixes #856

### DIFF
--- a/packages/core/test/invoke.test.ts
+++ b/packages/core/test/invoke.test.ts
@@ -363,6 +363,106 @@ describe('invoke', () => {
     service.send('INCREMENT');
   });
 
+  it('should forward events to services if autoForward: true before processing them (when sending batches)', done => {
+    const actual: string[] = [];
+
+    const childMachine = Machine<{ count: number }>({
+      id: 'child',
+      context: { count: 0 },
+      initial: 'counting',
+      states: {
+        counting: {
+          on: {
+            INCREMENT: [
+              {
+                target: 'done',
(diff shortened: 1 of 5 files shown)


packages/core/test/invoke.test.ts
```diff
@@ -363,6 +363,106 @@ describe('invoke', () => {
...
+      })
+      .onDone(() => {
+        expect(state.context).toEqual({ countedTo: 3 });
+        expect(actual).toEqual([
+          'child got INCREMENT',
+          'parent got INCREMENT',
+          'child got INCREMENT',
+          'child got INCREMENT',
+          'parent got INCREMENT',
+          'parent got INCREMENT'
```
Andarist: this illustrates the quirk of batching behavior, currently
```
send(FOO)
send(FOO)
```
is not quite equivalent of
```
send([FOO, FOO])
```
When it comes to the timing of parent & child actions.

PR #861, opened 2019-12-09
docs: change `json` to `json5`

0x8b:
I changed `json` to `json5` for better syntax highlighting.


Andarist:
Could you showcase how this provides better highlighting? It would be great to see a comparison before vs after.

0x8b:
Firefox 71 on Windows 10

before:
xstate-graph-before

after:
xstate-graph-after

Issue #807, opened 2019-11-19
[Feature] Add vue composition API functions


studnitz:
## Bug or feature request?
Feature Request
## Description:
I think with the upcoming release of Vue 3.0 and its [Composition API](https://vue-composition-api-rfc.netlify.com/) which is relatively similar in concept to React Hooks, it would be a great addition to port the current xstate-react Hooks to the Vue Composition API. At the moment the Composition API is [backported](https://github.com/vuejs/composition-api) (with a few caveats), so I could start working on a PR if it's wanted

davidkpiano:
@studnitz I would love this. What would a Vue hook look like?

Andarist:
Looking at Vue’s API - it would look exactly the same as existing useMachine 🚀. Only the implementation would differ. 

@studnitz is Vue’s `setup` called once per component’s instance?

studnitz:
@davidkpiano As @Andarist said. :)
@Andarist The [documentation](https://vue-composition-api-rfc.netlify.com/api.html#setup) says:
> `setup` is called right after the initial props resolution when a component instance is created. Lifecycle-wise, it is called before the `beforeCreate` hook.

Andarist:
Thanks for your answer! The last question - is it possible to track current props inside setup? Can't see (looking quickly through it though) any example of this. Are initial props received anyhow there? Can't see any example of receiving them as argument of the setup, nor available at `this`.

studnitz:
Yes, that's possible.

> ## Arguments
>
>The function receives the resolved props as its first argument:
>
>```js
>export default {
>  props: {
>    name: String
>  },
>  setup(props) {
>    console.log(props.name)
>  }
>}
>```
>Note this props object is reactive - i.e. it is updated when new props are passed in, and can be observed and reacted upon using the watch function:

AleksejDix:
I made a first Vue Composition API function today. Was inspired by the react useMachine
https://github.com/AleksejDix/vue-use/blob/master/src/compositions/useMachine.js

AleksejDix:
Yeah I was too late for my PR.  @studnitz  great work.

ChrisShank:
FYI the alpha for [`vue@next`](https://www.npmjs.com/package/vue/v/3.0.0-alpha.0) and  [`@vue/reactivity`](https://www.npmjs.com/package/@vue/reactivity) were released on NPM a couple of days ago.

davidkpiano:
[...]

PR #864, opened 2019-12-11
Add @xstate/vue package

studnitz:
As mentioned in #807 I started beginning implementing `useMachine` etc. using the [Vue Composition API](https://vue-composition-api-rfc.netlify.com/).

I took the liberty and named the function, that uses `@xstate/fsm` as its base `useFsm`. I hope that's okay, otherwise, we can revert that back of course.

Checklist for getting feature parity to the React Hooks:

- [x] Setup the repository so it uses @vue/composition-api, @testing-library/vue 
- [x] Add a README for the repository
- [x] Add documentation
- [x] Implement `useFsm` function
- [x] ~Copy~ Implement one test case for useFsm as a general smoke test
- [x] Add more tests for `useFsm` function
- [x] Implement `useMachine` function
- [x] Implement tests for `useMachine` function
- [x] Implement `useService` function (untested)
- [x] Implement tests for `useService` function
 
Leaving `useActor` for a later PR, having problems with getting the types right at the moment.


packages/xstate-vue/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "lib/**/*.js",
+    "lib/**/*.d.ts"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/davidkpiano/xstate.git"
+  },
+  "scripts": {
+    "clean": "rm -rf dist lib tsconfig.tsbuildinfo",
+    "prebuild": "npm run clean",
```
Andarist: Could you remove this one? It got removed in other packages - you have probably forked before it has happened.

packages/xstate-vue/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "build": "tsc",
+    "test": "jest",
+    "prepublish": "npm run build && npm run test"
+  },
+  "bugs": {
+    "url": "https://github.com/davidkpiano/xstate/issues"
+  },
+  "peerDependencies": {
+    "@vue/composition-api": "^0.3.4",
```
Andarist: is there any estimate for vue 3 being released? once it happens we'll have to change peers here, right?

packages/xstate-vue/src/useFsm.ts
```diff
@@ -0,0 +1,26 @@
+import { ref, onBeforeMount, onBeforeUnmount, Ref } from '@vue/composition-api';
+import { StateMachine, EventObject, interpret } from '@xstate/fsm';
+import { AnyEventObject } from 'xstate';
+
+export function useFsm<TC, TE extends EventObject = AnyEventObject>(
```
Andarist: IMHO it would be better to match `@xstate/react` naming here - so this would be a `useMachine`. It will be easier for people to educate people about it.

packages/xstate-vue/src/useFsm.ts
```diff
@@ -0,0 +1,26 @@
...
+): {
+  state: Ref<StateMachine.State<TC, TE, any>>;
+  send: StateMachine.Service<TC, TE>['send'];
+  service: Ref<StateMachine.Service<TC, TE>>;
+} {
+  const state = ref<StateMachine.State<TC, TE, any>>(stateMachine.initialState);
+  const service = ref<StateMachine.Service<TC, TE>>(interpret(stateMachine));
+  const send = (event: TE | TE['type']) => service.value.send(event);
+
+  onBeforeMount(() => {
```
Andarist: is unmount guaranteed to run if this gets run?

packages/xstate-vue/src/useFsm.ts
```diff
@@ -0,0 +1,26 @@
...
+import { AnyEventObject } from 'xstate';
+
+export function useFsm<TC, TE extends EventObject = AnyEventObject>(
+  stateMachine: StateMachine.Machine<TC, TE, any>
+): {
+  state: Ref<StateMachine.State<TC, TE, any>>;
+  send: StateMachine.Service<TC, TE>['send'];
+  service: Ref<StateMachine.Service<TC, TE>>;
+} {
+  const state = ref<StateMachine.State<TC, TE, any>>(stateMachine.initialState);
```
Andarist: do you need refs for those? i might miss something about vue, but looking at this code from React's perspective - it seems like this wouldn't be needed. In Vue this whole hook gets executed only once (from what I understand). Are refs different in Vue from what we have in ReacT? Could you describe how they work in Vue?

packages/xstate-vue/src/useFsm.ts
Andarist: Ok, looking at your useMachine - I could have misunderstood something. Could you describe what's the purpose of this hook? It seems like both your hooks are responsible for calling `interpret` and this confuses me.

packages/xstate-vue/src/useMachine.ts
```diff
@@ -0,0 +1,111 @@
...
+   */
+  context?: Partial<TContext>;
+  /**
+   * The state to rehydrate the machine to. The machine will
+   * start at this state instead of its `initialState`.
+   */
+  state?: StateConfig<TContext, TEvent>;
+}
+
+const defaultOptions = {
```
Andarist: I don't think it gets actually used anyhow in the code.

packages/xstate-vue/package.json
studnitz: I think it's planned for Q1 2020. It could mean changing peers if we don't want to support vue 2 anymore.

packages/xstate-vue/src/useFsm.ts
studnitz: Okay, I'll change that then.

packages/xstate-vue/src/useFsm.ts
studnitz: They correspond to the [Vue Lifecycle Hooks](https://vue-composition-api-rfc.netlify.com/api.html#lifecycle-hooks). ([See the diagram](https://vuejs.org/v2/guide/instance.html#Lifecycle-Diagram)) 

So it would live for the lifetime of the componennt, akin to the `useMachine` hook.

packages/xstate-vue/src/useMachine.ts
studnitz: Yeah, you're right. 

Should I add the same implementation like [in the react hook?](https://github.com/davidkpiano/xstate/blob/master/packages/xstate-react/src/index.ts#L110)

packages/xstate-vue/src/useFsm.ts
studnitz: In the composition API every reactive value you want to have, you wrap it with either a `const count = ref(0)` or you create a `const state = reactive({ count: 1 }`. If you want to mutate a ref's value you'd have to use its `value` property. So mutating the `count` to 1 would be `coult.value = 1`

As we want the state to be a reactive property I wrap the given `initalState` so I can later mutate it by executing `subscribe(s => (state.value = s))`. 

At the end of a `setup()` function you return those refs, so you can use them in your components as reactive values, methods etc.

Also see the [API documentation for ref](https://vue-composition-api-rfc.netlify.com/api.html#ref)

packages/xstate-vue/src/useFsm.ts
studnitz: I just copied the structure of the [React FSM useMachine hook](https://github.com/davidkpiano/xstate/blob/master/packages/xstate-react/src/fsm.ts), but called it useFsm to show the difference that when using `useFsm` you are actually relying on `@xstate/fsm` and not on `@xstate`.

But as I said in my description, I could also of course name it `useMachine` as well. It wouldn't get an export in `index.ts` though then.

packages/xstate-vue/src/useFsm.ts
davidkpiano: Right, a Vue `ref` is different than a React `ref` from my understanding.

packages/xstate-vue/src/useFsm.ts
Andarist: Ah, ok - I get it now. Was reviewing late and didn't spot this one is actually using `@xstate/fsm`. I think we need to settle on a solution for https://github.com/davidkpiano/xstate/pull/847 and then we'll be able to just follow the suite in here in regards to supporting both `xstate` & `@xstate/fsm`

packages/xstate-vue/src/useFsm.ts
Andarist: Ok - so I've read the explanation and the docs and I think I understand Vue's ref and reactive somewhat OK.

So given that - I understand why `state` is a ref here, but I don't get why `service` is one too. It's only initialized and you never write to it again, so it should be just a plain value.

packages/xstate-vue/src/useMachine.ts
```diff
@@ -0,0 +1,111 @@
...
+  );
+
+  const initialState = rehydratedState
+    ? State.create(rehydratedState)
+    : service.value.initialState;
+
+  const current = ref<State<TContext, TEvent>>(initialState);
+
+  // Make sure actions and services are kept updated when they change.
+  watch(() => {
```
Andarist: are those needed? are `actions` reactive in here?

packages/xstate-vue/test/vue-shim.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.vue' {
```
Andarist: doesn't vue come with types? If it does then we should just install & use them.

packages/xstate-vue/src/useService.ts
```diff
@@ -0,0 +1,31 @@
...
+  service: Interpreter<TContext, any, TEvent>
+): {
+  current: Ref<State<TContext, TEvent>>;
+  send: Interpreter<TContext, any, TEvent>['send'];
+  service: Ref<Interpreter<TContext, any, TEvent>>;
+} {
+  const current = ref<State<TContext, TEvent>>(service.state);
+  const serviceRef = ref<Interpreter<TContext, any, TEvent>>(service);
+
+  const { unsubscribe } = serviceRef.value.subscribe(state => {
```
Andarist: In other hooks you subscribe in `onBeforeMount` and here you subscribe directly in the setup function. Could you elaborate more on that? Is this safe?

packages/xstate-vue/src/useMachine.ts
Andarist: I don't think so - actually we need to remove `immediate` from the react hook as it's not async-react safe.

packages/xstate-vue/src/useService.ts
studnitz: You are right on that, as I'm not that familiar with React hooks I just translated `useEffect` to those Lifecycle hooks in my mind. I think generally using the `onBeforeMount` hooks are not needed, as the state machines don't interact with the DOM directly. (where `onMounted` would be needed e.g.)

packages/xstate-vue/test/vue-shim.d.ts
studnitz: It does and according to the [docs on Typescript](https://vuejs.org/v2/guide/typescript.html#Official-Declaration-in-NPM-Packages) they should work out of the box, but I couldn't get the tests working without that.

packages/xstate-vue/src/useService.ts
Andarist: Right - I'm just wondering if it's possible for a setup function to be called but a corresponding component never actually gets mounted like rendered in memory but discarded shortly after that. Do you know if it's possible in Vue?

studnitz: There are [Renderless Components](https://adamwathan.me/renderless-components-in-vuejs/) in Vue, although they are also bound to the lifecycle of the component that is using them.

AleksejDix:
i've added the doku page to the website
https://github.com/studnitz/xstate/pull/2

studnitz:
I'm having problems with implementing the `useActor` functions regarding types, I think I'll leave that to a later PR to get that sorted out. 

So, all in all, I just need to add tests for the `useService` function and add the documentation for `useFsm` and `useService` and we would be good to go. 🚀

davidkpiano:
@studnitz No problem, all we need is `useMachine` and `useService` to make this shippable I think (up to you!) - I wouldn't worry about `useActor`, it's just a general use hook.

packages/xstate-vue/src/useFsm.ts
```diff
@@ -0,0 +1,24 @@
...
+  state: Ref<StateMachine.State<TC, TE, any>>;
+  send: StateMachine.Service<TC, TE>['send'];
+  service: StateMachine.Service<TC, TE>;
+} {
+  const state = ref<StateMachine.State<TC, TE, any>>(stateMachine.initialState);
+  const service = interpret(stateMachine);
+  const send = (event: TE | TE['type']) => service.send(event);
+
+  service.subscribe(s => (state.value = s));
+  service.start();
```
Andarist: is this safe to do in here? I, unfortunately, don't have Vue expertise to judge that - but you have done the same thing for `useMachine` in `onBeforeMount`. Which one is more correct? I believe both should be started at the same time, right?

The edge-case scenario I'm concerned with is this:
1. render a component
2. while rendering change synchronously the state so the component in question gets dicarded

How far would Vue get with processing such a component? Would it be mounted (I suspect it wouldnt be)? And if it wouldnt get mounted - would `onBeforeUnmount` be called?

packages/xstate-vue/test/UseMachine.vue
```diff
@@ -0,0 +1,71 @@
...
+
+    const { current, send, service } = useMachine(fetchMachine, {
+      services: {
+        fetchData: onFetch
+      },
+      state: persistedState
+    });
+    watch(() =>
+      current.value.actions.forEach(action => {
+        if (action.type === 'load') {
```
Andarist: wouldnt this already be handled by the interpreter in `useMachine`?

studnitz: I tried to follow the equivalent [react hook test](https://github.com/davidkpiano/xstate/blob/07f390f450219f7b25a024e99e35e3d534489266/packages/xstate-react/test/useFSM.test.tsx#L46) while implementing that test, so maybe? :)

Andarist: Ok - this is actually for `useMachine` with `@xstate/fsm` so it doesn't have action execution builtin to its interpreter 👍

packages/xstate-vue/src/useMachine.ts
AleksejDix: since actions are not wrapped with ref() or reactive() they are not reactive.

Andarist: they don't need to be watched here then, right?

studnitz: If wewant to replicate the React Hooks behaviour I think they should be watched, but then we'd have to think about if the options should be provided as refs or converted to refs on the initialization.

packages/xstate-vue/src/useFsm.ts
studnitz: Okay, I think I get now what you mean. I'm not also sure what would happen here, but I can try to construct a test case and evaluate that.

But generally for now we would be on the safe side if the services get initialized in `onMounted` and stopped on `onBeforeUnmout, right?

packages/xstate-vue/src/useMachine.ts
Andarist: Do we need to replicate it? Those lines in react hooks version are only meant to fix the stale closure problem, can Vue version suffer from this as well?

packages/xstate-vue/src/useFsm.ts
Andarist: Yes, that would certainly be safe and would align the behavior to what is implemented in `@xstate/react`. In React though it's the only way of setting up a subscription safely, if it's possible in Vue to start safely as service sooner then I wouldn't mind having a difference between those packages.

AleksejDix:
I've added Doku for some other parts 
https://github.com/studnitz/xstate/pull/3

davidkpiano:
Is this about ready to be reviewed/merged?

studnitz:
I think the PR is ready to be reviewed. :) 

At the moment the `options` in `useMachine` can't be wrapped by Refs as I couldn't get the types working, which means they can't be watched upon, e.g. if they are updated by props.


```js
import useMachine from '@xstate/vue';
export default {
	setup(props) {
		// props.options is only ready once
		const { current, send } = useMachine(machine, props.options);
	}
}
```

I'd like to tackle that in a separate PR though and gather more experience how it is done in other projects, that also have implemented the composition API. (e.g. [vue-apollo](https://github.com/vuejs/vue-apollo/tree/v4/packages/vue-apollo-composable/src))

davidkpiano:
@studnitz 

> At the moment the options in useMachine can't be wrapped by Refs as I couldn't get the types working, which means they can't be watched upon, e.g. if they are updated by props.

I think that's fine. The whole premise with state machines is that once they're set, they're set (much like any electronic device you own), so we shouldn't expect `props.options` to ever change, and should provide a warning if they do.

docs/packages/xstate-vue/index.md
```diff
@@ -0,0 +1,149 @@
...
+  }
+};
+</script>
+```
+
+## API
+
+### `useMachine(machine, options?)`
+
+A [Vue comopsition function](https://vue-composition-api-rfc.netlify.com/) that interprets the given `machine` and starts a service that runs for the lifetime of the component.
```
davidkpiano:
```suggestion
A [Vue composition function](https://vue-composition-api-rfc.netlify.com/) that interprets the given `machine` and starts a service that runs for the lifetime of the component.
```

docs/packages/xstate-vue/index.md
```diff
@@ -0,0 +1,149 @@
...
+
+**Arguments**
+
+- `machine` - An [XState machine](https://xstate.js.org/docs/guides/machines.html).
+- `options` (optional) - [Interpreter options](https://xstate.js.org/docs/guides/interpretation.html#options) OR one of the following Machine Config options: `guards`, `actions`, `activities`, `services`, `delays`, `immediate`, `context`, or `state`.
+
+**Returns** `{ current, send, service}`:
```
davidkpiano:
```suggestion
**Returns** `{ current, send, service }`:
```

docs/packages/xstate-vue/index.md
```diff
@@ -0,0 +1,149 @@
...
+
+- `current` - Represents the current state of the machine as an XState `State` object.
+- `send` - A function that sends events to the running service.
+- `service` - The created service.
+
+### `useService(service)`
+
+A [Vue composition function](https://vue-composition-api-rfc.netlify.com/) that subscribes to state changes from an existing [service](TODO).
```
davidkpiano:
```suggestion
A [Vue composition function](https://vue-composition-api-rfc.netlify.com/) that subscribes to state changes from an existing [service](https://xstate.js.org/docs/guides/communication.html).
```

davidkpiano:
The API looks great, succinct, and intuitive. Looks good to me!

studnitz:
Yay! One thing I noted is that I didn't add a changeset, which is used for the package versioning, right?

Andarist:
It's not needed right now. Changesets main focus is to aid in changelog generation and bumping depending packages, keeping everything in sync. As `@xstate/vue` has not existed before, there are no changes to describe. I think we only should add a simple changeset with a link to the docs, nothing more. This can be added before publishing - from what I understand David won't publish this just yet, but rather sometime in January (hopefully). We need to introduce some minor changes to `@xstate/react` (want to promote it to v1) and would be great to keep `@xstate/vue` in sync with that.

There is one problem currently though and we'd appreciate help on that - tests are failing on master right now, the problem is about vue setup. Vue files somehow cannot be required by jest - https://github.com/davidkpiano/xstate/runs/368410854#step:6:352 . Could you take a look at it?

studnitz:
@Andarist I think the problem here is that git didn't pick up the change when I renamed `UseFsm.vue` to `UseFSM.vue`, in the filetree it's still [https://github.com/davidkpiano/xstate/blob/master/packages/xstate-vue/test/UseFsm.vue](UseFsm.vue), which fails. 

grafik

This is what I see on my machine.

Probably because I dev on Windows I guess?

RomainLanz:
Any plan to release this version/package soon?

Andarist:
@RomainLanz hopefully this month, we still need to iron some things out

Issue #820, opened 2019-11-25
Typestates not working as expected with xstate


schpet:
**Description**

not able to narrow typestate types when subscribing to a state machine service

**Expected Result**

within a `state.maches("foo") { ... }` condition, types to be narrowed to the appropriate State type

**Actual Result**

types are not narrowed, you get the Context type

![xstate-issue](https://user-images.githubusercontent.com/921605/69552997-3d663780-0f54-11ea-82e7-0d305f45f89b.png)

**Reproduction**

https://codesandbox.io/s/awesome-tree-qom29 (see `src/index.ts` L62)

**Additional context**

- https://github.com/davidkpiano/xstate/issues/819#issuecomment-558201650
- this problem is specific to xstate, works as expected with @xstate/fsm

RafalFilipek:
Example: https://github.com/RafalFilipek/xstate-typestates-bug

davidkpiano:
Fix for this coming 🔜 but for now, this should be general a workaround:

```ts
if (state.matches<any>('something')) {
  // ...
}
```

RafalFilipek:
https://github.com/RafalFilipek/xstate-typestates-bug I've just updated my example. Still no luck. I think I'll wait for fix. I can live with that since we have optional chaining and nullish coalescing :)

LookRain:
Having same issue with `useMachine` from `@xstate/react`

davidkpiano:
Fix being released soon-ish: #868

PR #868, opened 2019-12-11
fix(core): Typestates interpreter integration

davidkpiano:
This PR fixes the issue where Typestates were not propagated to interpreted services. Addresses #820

- [x] Add types
- [x] Add changeset


packages/core/src/Machine.ts
```diff
@@ -26,7 +26,7 @@ export function Machine<
   config: MachineConfig<TContext, TStateSchema, TEvent>,
   options?: Partial<MachineOptions<TContext, TEvent>>,
   initialContext?: TContext
-): StateMachine<TContext, TStateSchema, TEvent>;
+): StateMachine<TContext, TStateSchema, TEvent, any>;
```
Andarist: Do you know why this `any` is needed? I see the default value for this type parameter being `Typestate<TContext>` and I don't immediately see how this helps anything here. I still learn TS so maybe I'm missing something obvious~

packages/core/test/interpreter.test.ts
```diff
@@ -1798,7 +1798,7 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
       const service = interpret(parentMachine)
         .onTransition(state => {
           if (
-            state.matches('active') &&
+            state.matches<any>('active') &&
```
Andarist: from what I'm observing `parentMachine` is not using type state feature and thus this `state` has its typestate parameter to be `Typestate<any>` and because of that (maybe indirectly?) this `matches` call get inferred to:
```
State<any, AnyEventObject, any, Typestate<any>>.matches<"active">(parentStateValue: "active"): this is never
```
and because of that we can't read `state.children` because `state` got refined by TS to `never`. This seems like regression/bug (not sure sure if introduced here). Unless Typestate is actually provided we probably shouldn't try to utilize it because it leads to situations like this.

packages/core/src/interpreter.ts
```diff
@@ -54,10 +55,11 @@ import { isInFinalState } from './stateUtils';
...
 import { registerService } from './devTools';
 
-export type StateListener<TContext, TEvent extends EventObject> = (
-  state: State<TContext, TEvent>,
-  event: TEvent
-) => void;
+export type StateListener<
+  TContext,
+  TEvent extends EventObject,
+  TState extends Typestate<TContext> = any
```
Andarist: Could we call this here (and at other places) `TTypestate` or something? `TState` can get easily be interpreted as something different as the word state gets used quite a lot in the codebase 😉 but on the other hand, Typestate is quite a distinct name that would give a better notion of what this holds while reading usage sites of this thing

packages/core/src/Machine.ts
davidkpiano: If a Typestate isn't provided, we need to default to `any`.

Andarist: Could you elaborate? When the default applies then? Do we **always** need to supply `any`? Or is this situation here different somehow from other ones?

davidkpiano: `any` no longer needs to be supplied.

packages/core/src/types.ts
```diff
@@ -641,8 +641,8 @@ export interface StateMachine<
   TContext,
   TStateSchema extends StateSchema,
   TEvent extends EventObject,
-  TState extends Typestate<TContext> = Typestate<TContext>
-> extends StateNode<TContext, TStateSchema, TEvent, TState> {
+  TTypestate extends Typestate<TContext> = any
```
Andarist: what kind of an effect this change might have on consuming applications? default has changed after all

packages/core/test/interpreter.test.ts
Andarist: @davidkpiano this one seems like a problem - if that `<any>` is needed. We don't want people having to do that, right?

packages/core/src/types.ts
davidkpiano: None

packages/core/test/interpreter.test.ts
davidkpiano: It doesn't need that, this has been fixed.

PR #874, opened 2019-12-13
Add vue machine component

AleksejDix:
Add Vue Components which can be used without Composition API by using the 
Renderless component pattern
https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots

Can be merged after:
https://github.com/davidkpiano/xstate/pull/864

Help wanted to implement types with Vue.

### DIFF
--- a/docs/guides/context.md
+++ b/docs/guides/context.md
@@ -184,7 +184,7 @@ Both the property assigner and context assigner function signatures above are gi
 The `meta` object contains:
 
 - `state` - the current state in a normal transition (`undefined` for the initial state transition)
-- `action` - the assign action|
+- `action` - the assign action
 
 ::: warning
 The `assign(...)` function is an **action creator**; it is a pure function that only returns an action object and does _not_ imperatively make assignments to the context.


packages/xstate-vue/src/xstate-machine.ts
```diff
@@ -0,0 +1,103 @@
...
+    const { current, service, send } = this;
+    return h('div', [
+      this.$scopedSlots.default({
+        current,
+        service,
+        send
+      })
+    ]);
+  },
+  props: {
```
Andarist: is this needed? seems like a runtime type checking thing? should we just rely on static type checker with this sort of thing?

AleksejDix: this is the default way 
https://vuejs.org/v2/guide/components-props.html#Prop-Validation

I think with typescript it would be `Object as () => SomeType`

packages/xstate-vue/src/xstate-machine.ts
```diff
@@ -0,0 +1,103 @@
...
+      return this.service.initialState;
+    },
+    actions() {
+      return this.options.actions;
+    },
+    services() {
+      return this.options.services;
+    }
+  },
+  watch: {
```
studnitz: This is not needed anymore. See @davidkpiano's [comment](https://github.com/davidkpiano/xstate/pull/864#issuecomment-569761475).

PR #875, opened 2019-12-14
Drop immediate option from useMachine

Andarist:
I have some more changes planned to `@xstate/react` - don't want to make them all at once so they can be more easily reviewed. Also can't quite make parallel PRs as they would probably conflict with each other quite a bit.

There are not that many of them though - so if we review & merge them at a good pace then we would be able to release v1 of that package in the next week. That being said - it maybe would be a good thing to create a base branch from the master for those changes, so we don't block releasing of other packages while we work on `@xstate/react` (assuming we release things automatically using changesets versioning PRs - because once we merge this to master it will be immediately added to the next release batch).

### DIFF
--- a/.changeset/orange-fans-retire.md
+++ b/.changeset/orange-fans-retire.md
@@ -1,0 +1,5 @@
+---
+'@xstate/react': major
+---
+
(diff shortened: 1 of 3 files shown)


Andarist:
@davidkpiano could you take a look at this one?

davidkpiano:
@Andarist LGTM, my recent commit to `@xstate/react` addresses a related issue too, where `immediate: true` would have been a "solution" but is not needed.

PR #876, opened 2019-12-14
fix(core): cache initial state in interpreter instance. Fixes #838

### LINKED ISSUE #838: [4.7.0] assign() is calling its functions multiple times
**Description**

`assign()` is calling its functions multiple times.

**Expected Result**

`assign()` should call its functions just once.

**Actual Result**

`assign()` is calling its functions multiple times. A side effect of that is that when using `spawn()` we end up with multiple instances of the same spawned machine:

Multiple spawned machines

**Reproduction**

https://codesandbox.io/s/xstate-react-template-1u2b4

**Additional context**

Looks like this problem was introduced by `v4.7.0`, it's working perfectly in `v4.6.7`: https://codesandbox.io/s/xstate-react-template-44r3j

davidkpiano:
This PR caches `initialState` in the service instead of in the machine (which was problematic but an initial "solution").

### DIFF
--- a/.changeset/rare-pillows-fold.md
+++ b/.changeset/rare-pillows-fold.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The initial state is now cached inside of the service instance instead of the machine, which was the previous (faulty) strategy. This will prevent entry actions on initial states from being called more than once, which is important for ensuring that actors are not spawned more than once.
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -144,6 +144,7 @@ export class Interpreter<
    * The current state of the interpreted machine.
    */
   private _state?: State<TContext, TEvent>;
+  private _initialState?: State<TContext, TEvent>;
   /**
    * The clock that is responsible for setting and clearing timeouts, such as delayed events and transitions.
    */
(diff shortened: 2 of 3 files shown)


packages/core/src/interpreter.ts
```diff
@@ -913,6 +923,9 @@ export class Interpreter<
     }
   }
   public spawn(entity: Spawnable, name: string, options?: SpawnOptions): Actor {
+    if (this._status !== InterpreterStatus.Running) {
```
Andarist: this change doesn't seem to be described in the changeset nor in added tests, shouldn't we add both for this? A test would let me understand this change better - because right now I'm not sure what exact issue gets fixed by this. What does it give for us to spawn a null actor here? Wouldnt it be maybe more appropriate to throw an error instead?

.changeset/rare-pillows-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The initial state is now cached inside of the service instance instead of the machine, which was the previous (faulty) strategy. This will prevent entry actions on initial states from being called more than once, which is important for ensuring that actors are not spawned more than once.
```
Andarist: The added test doesn't quite reflect this part:
> which is important for ensuring that actors are not spawned more than once.

Could we add a test (or edit the one already added) for exactly what is described here?

packages/core/src/interpreter.ts
davidkpiano: This can be removed for now.

.changeset/rare-pillows-fold.md
davidkpiano: Executing the assignment function in the provided test is the only way an actor can potentially be spawned multiple times. It is sufficient to test that the assignment function is only called once.

Andarist: Yes - it’s sufficient but the test case doesnt make this particular intention clear. Even putting the aside the implementation of this test - it’s title only mentions caching. Im also not immediately sure how reading initialState lead to executing its actions (spawning in here) as it is supposed to be pure so I’m not sure how caching has helped here as it shouldnt matter much.

PR #882, opened 2019-12-18
fix(docs): use `onReceive` consistently as callback handler's 2nd argument

d4rekanguok:
Just a tiny error: Actor doc refers to callback actor's  2nd argument as `onReceive`, but it is `onEvent` in the linked entry about invoking callback ([link](https://xstate.js.org/docs/guides/communication.html#invoking-callbacks))

### DIFF
--- a/docs/guides/actors.md
+++ b/docs/guides/actors.md
@@ -37,7 +37,7 @@ They may have optional methods:
 All the existing invoked service patterns fit this interface:
 
 - [Invoked promises](./communication.md#invoking-promises) are actors that ignore any received events and send at most one event back to the parent
-- [Invoked callbacks](./communication.md#invoking-callbacks) are actors that can send events to the parent (first `callback` argument), receive events (second `onReceive` argument), and act on them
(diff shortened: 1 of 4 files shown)


davidkpiano:
Thanks for this, it might be better to rename `onEvent` to `onReceive` since that better reflects what's happening.

d4rekanguok:
> it might be better to rename onEvent to onReceive since that better reflects what's happening.

I'm down to help out on this one if it's a better idea, it looks like it'd only involve changing the docs + typing name here?

https://github.com/davidkpiano/xstate/blob/771becf761c14319795685feda8b2f015a75707d/packages/core/src/types.ts#L195-L198

davidkpiano:
Yes, you can change the typing name.

d4rekanguok:
Update: I've changed all references (in docs, typing names and test) to the name of callback handler's arguments to `onReceive`.

The first argument, `callback`, is also referred to under different names (`sender`, `next`) in a few places so I changed those as well.

PR #888, opened 2019-12-21
Fixed issue with delayed transitions scheduling a delayed event for each transition defined for a single delay

### LINKED ISSUE #886: (4.7.2) Conditional transitions break `after` delayed transition.
(sandbox: https://codesandbox.io/s/determined-volhard-ke3dt)

Consider this simple machine. If event `FOO` is received within 1.5 seconds state X should be re-entered and delayed `after` transition resetted. Even if state X is re-entered many times, `after` transition should be fired only once. Sure enough all works as expected here.

```javascript
const machineGood = Machine({
  id: "someMachine",
  initial: "X",
  states: {
    X: {
      on: {
        FOO: "X"
      },
      after: {
        1500: {
          target: "Y"
        }
      }
    },
    Y: {}
  }
});
```

But if conditional transitions are added to the `after` things get wrong. The following machine should exhibit the same behavior but in fact it is different. In this case delayed transition is fired multiple times. 

```javascript
const machineBug = Machine({
  id: "someMachine",
  initial: "X",
  states: {
    X: {
      on: {
        FOO: "X"
      },
      after: {
        1500: [
          {
            target: "Y",
            cond: () => 1 > 0
          },
          {
            target: "Z"
          }
        ]
      }
    },
    Y: {},
    Z: {}
  }
});
```

Here is the code to run tests:

```javascript
function sleep(period) {
  return new Promise(r => setTimeout(r, period));
}

async function demo(machine) {
  // expected outcome in the comments
  const service = interpret(machine).onTransition(state =>
    console.log(state.value)
  );
  service.start();
  //X
  await sleep(1000);
  service.send({ type: "FOO" });
  //X
  await sleep(1000);
  service.send({ type: "FOO" });
  //X
  //...after 1500 ms
  //Y
}
```

demo(machineGood);
//correct:
//X
//X
//X
//Y
demo(machineBug);
// incorrect:
//X
//X
//Y
//Y
//Y

Andarist:
regression caused by #647 
fixes #886

### DIFF
--- a/packages/core/test/after.test.ts
+++ b/packages/core/test/after.test.ts
@@ -118,6 +118,36 @@ describe('delayed transitions', () => {
       .start();
   });
 
+  it('should defer a single send event for a delayed transition with multiple conditions (#886)', () => {
+    type Events = { type: 'FOO' };
+
+    const machine = Machine<{}, Events>({
+      initial: 'X',
+      states: {
+        X: {
+          on: {
+            FOO: 'X'
+          },
+          after: {
+            1500: [
+              {
+                target: 'Y',
+                cond: () => true
+              },
+              {
+                target: 'Z'
+              }
+            ]
+          }
+        },
+        Y: {},
+        Z: {}
+      }
+    });
+
+    expect(machine.initialState.actions.length).toBe(1);
+  });
+
   describe('delay expressions', () => {
     type Events =
       | { type: 'ACTIVATE'; delay: number }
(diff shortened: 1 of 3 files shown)


packages/core/test/after.test.ts
```diff
@@ -118,6 +118,36 @@ describe('delayed transitions', () => {
       .start();
   });
 
+  it('should defer a single send event for a delayed transition with multiple conditions (#886)', () => {
```
Andarist: this is not quite a good description - there are no multiple conditions defined here, but I'm having trouble with rephrasing this

davidkpiano: I'm not sure either, this is fine for now.

PR #893, opened 2019-12-24
Fix description text if no meta description given

maoberlehner:
7f7672e8a67451b7e0c3c4285dd8061decb385e8 introduced a bug where without
a `meta.description`, the test description text is
`reaches state: "undefined"`.

This is because even if `meta.description` is `undefined`,
`"${description}"` becoming `"undefined"` is truthy and the code after
`||` can never be reached.

Also double quoting with `JSON.stringify` was fixed.

See https://github.com/davidkpiano/xstate/commit/7f7672e8a67451b7e0c3c4285dd8061decb385e8#diff-8e162acfb6f0732195f10a5f0592fdb3L325-R332

### DIFF
--- a/.changeset/moody-buttons-double.md
+++ b/.changeset/moody-buttons-double.md
@@ -1,0 +1,5 @@
+---
+'@xstate/test': patch
+---
+
+Fix description text if no meta description is given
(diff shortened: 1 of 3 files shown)


Andarist:
Thanks! Could you add a test and a changeset for this one?

maoberlehner:
Updated the PR!

PR #895, opened 2019-12-24
Reimplement useMachine to always start service immediately

Andarist:
⚠️⚠️⚠️Right now this is just an experiment I'd like to gather feedback on ⚠️⚠️⚠️ 

There is this mantra repeated by React team that subscriptions should be done in a commit phase (inside an effect) - mainly because only then we get also an "access" to a cleanup mechanism. And given the React model - this all is true.

Here, however, we don't deal with an external source - but rather we create our own and we own it. So it seems to me that rules are a little bit different for such a scenario. We could even move subscribing to the moment right after calling `.start()` (in render phase). We don't actually need a cleanup phase because we should be able to rely upon GC doing its job when our component unmounts (or doesn't ever get committed) - the source (service) should be shared **only** in its React subtree, so if this gets removed from memory then it should just free all references to created service and it should get removed from as well. `useState` could be also seen as such "internal" source and a component is "subscribed" to it right from the start after all.

### DIFF
--- a/packages/xstate-react/src/useConstant.ts
+++ b/packages/xstate-react/src/useConstant.ts
@@ -1,0 +1,13 @@
+import * as React from 'react';
+
+type ResultBox<T> = { v: T };
+
+export default function useConstant<T>(fn: () => T): T {
+  const ref = React.useRef<ResultBox<T>>();
+
+  if (!ref.current) {
+    ref.current = { v: fn() };
+  }
+
+  return ref.current.v;
+}
--- a/packages/xstate-react/test/useMachine.test.tsx
+++ b/packages/xstate-react/test/useMachine.test.tsx
@@ -137,7 +137,6 @@ describe('useMachine hook', () => {
         execute: false
       });
 
-      // expect(service.initialized).toBe(false);
       expect(service.options.execute).toBe(false);
 
       return null;
--- a/packages/xstate-react/test/useMachine.test.tsx
+++ b/packages/xstate-react/test/useMachine.test.tsx
@@ -137,7 +137,7 @@ describe('useMachine hook', () => {
         execute: false
       });
 
-      expect(service.initialized).toBe(false);
+      // expect(service.initialized).toBe(false);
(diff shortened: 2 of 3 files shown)


packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -137,7 +137,7 @@ describe('useMachine hook', () => {
         execute: false
       });
 
-      expect(service.initialized).toBe(false);
+      // expect(service.initialized).toBe(false);
```
Andarist: I think this wasnt relevant for this particular test and it just broke by starting service immediately.

packages/xstate-react/src/useConstant.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+
+type ResultBox<T> = { v: T };
+
+export default function useConstant<T>(fn: () => T): T {
```
Andarist: this just a copy of a little helper I'm using (and which you kinda had inlined here:
https://github.com/davidkpiano/xstate/blob/8545f851e0b1ef9c374e1845abb9c3a0f7735e80/packages/xstate-react/src/index.ts#L67-L70

I'm just using it a way to hide this dance with a ref and also make sure that a defined TS type is received at a call site (null can't pop up when trying to consume content of this).

I have this packaged in an npm package, but I know you don't quite like external deps 😉 

Also - can just inline this if you think this is over complicated and unnecessary here.

Andarist:
@davidkpiano I've removed the possibility to swap a machine as discussed. The code has changed quite a bit from a prior version so this is ready for a re-review now. Do you have in mind any particular tests I should add for this?

packages/xstate-react/src/useConstant.ts
davidkpiano: This is fine as-is

Andarist:
@davidkpiano WDYT about this PR as a whole? should I go for it? or do you maybe have some concerns about the used approach?

davidkpiano:
This LGTM

PR #898, opened 2019-12-26
V5 (WIP)

davidkpiano:
This PR is a placeholder PR for the ongoing V5 work.

Changes so far:
- Activities removed (in favor of `invoke`)
- `state.children` an array instead of an object


packages/core/src/types.ts
```diff
@@ -1014,6 +1014,20 @@ export interface InterpreterOptions {
 
 export type AnyInterpreter = Interpreter<any, any, any>;
 
+/**
+ * Represents the `Interpreter` type of a given `MachineNode`.
+ *
+ * @typeParam TM - the machine to infer the interpreter's types from
+ */
+export type InterpreterOf<TM extends MachineNode> = TM extends MachineNode<
```
davidkpiano, line 1022: Heads up @Andarist - added this type while refactoring to make the tests pass, and it's actually really useful!

PR #901, opened 2019-12-28
V5: invoke simplification

davidkpiano:
(Breaking change) The main goal of this PR is to simplify `invoke` to only accept an `InvokeCreator`, which is a function that takes in context, event, and meta (`parent`, `id`, etc.) and returns an `Actor`.

This means that it is no longer the responsibility of the interpreter to interpret machines, promises, observables, callbacks, etc. as actors, and this can be done functionally instead:

```js
import {
  spawnMachine,
  spawnPromise,
  spawnCallback,
  spawnObservable
} from 'xstate/invoke'; // todo: figure out how to output to this path
```

```diff
- invoke: someMachine
+ invoke: spawnMachine(someMachine)

- invoke: (ctx, e) => somePromise
+ invoke: spawnPromise((ctx, e) => somePromise)

- invoke: (ctx, e) => (cb, receive) => { ... }
+ invoke: spawnCallback((ctx, e) => (cb, receive) => { ... })

- invoke: (ctx, e) => someObservable$
+ invoke: spawnObservable((ctx, e) => someObservable$)
```

Custom actors can be made as a result:

```js
invoke: (ctx, e, { parent }) => {
  let count = 0;

  // custom actor
  return {
    send(event) {
      count++;
      parent.send({ type: 'COUNT', count });
    },
    stop() {
      console.log(`Final count: ${count}`);
    }
  };
}
```

This also includes a helper function for activities:

```diff
- activity: (ctx, e) => { ... }
+ invoke: spawnActivity((ctx, e) => { ... })
```

Still a few things to clean up, but most of the important changes are in `invoke.ts`.


Andarist:
> This means that it is no longer the responsibility of the interpreter to interpret machines, promises, observables, callbacks, etc. as actors,

I like this very much 👍 Everyone will be able to integrate anything to XState now, using the very same API we are using internally for builtin (but tree-shakeable!) helpers such as spawnCallback, spawnObservable etc.

Gonna dive into reviewing this today.

tivac:
This PR seems like it might remove support for returning one of multiple child machines from a callback function?

Also can any of these aside from spawnPromise accept an async callback so we could dynamically import() in a child machine at runtime?

Andarist:
> This PR seems like it might remove support for returning one of multiple child machines from a callback function?

Could you give an example of that?

> Also can any of these aside from spawnPromise accept an async callback so we could dynamically import() in a child machine at runtime?

Code-splitting would be an interesting use case. Gonna give this some more thought.

tivac:
@Andarist I'm not at a computer right now, but I'll try to get you one in a bit here.

And yes, I'm anyways banging that code splitting drum 😅

I forgot to say earlier that I think the extra explicitness here is excellent and I really appreciate the PR overall, this looks like great work @davidkpiano!

Andarist:
Ive reread your comment and i think you have meant an invoke callback „choosing” a machine to invoke. Could you confirm that?

tivac:
Yes, though thinking about it done we could instead do that via conditional transition and nested invoked machines. It's a little bigger but more "xstate-y".

I still want to be able to async load in machines though, especially since we tie components to machine states so being able to load those only when necessary would be great.

davidkpiano:
> I still want to be able to async load in machines though, especially since we tie components to machine states so being able to load those only when necessary would be great.

This PR gives you a truly agnostic way to do any custom "machine" loading that you want, really... here's an untested contrived example of what you're trying to do:

```js
function spawnAsyncMachine(machinePath) {
  return (context, event) => {
    const pendingMessages = [];
    let status = 0; // not started

    const actor = {
      send: event => pendingMessages.push(event)
    }

    import(machinePath).then(machine => {

      // send pending messages
      service.send(pendingMessage);

      // swap implementation
      actor.send = service.send;
      actor.stop = service.stop;
    });

    return actor;
  }
}
```

Andarist:
And similarly to that, you can implement that other pattern - invoking a concrete machine based on your conditions.

packages/core/src/interpreter.ts
```diff
@@ -681,13 +682,15 @@ export class Interpreter<
...
+      const e = {
         ...event,
         name:
           event.name === actionTypes.error ? `${error(this.id)}` : event.name,
         origin: this.sessionId
-      });
+      };
+      // console.log('..', e);
+      // Send SCXML events to machines
+      (target as Interpreter<TContext, TStateSchema, TEvent>).send(e);
```
Andarist: casting to Actor would be more appropriate here

packages/core/src/interpreter.ts
```diff
@@ -726,7 +729,7 @@ export class Interpreter<
 
       if (!child) {
         throw new Error(
-          `Unable to forward event '${event}' from interpreter '${this.id}' to nonexistant child '${id}'.`
+          `Unable to forward event '${event.name}' from interpreter '${this.id}' to nonexistant child '${id}'.`
```
Andarist: is it possible to trigger this error? forwarding happens before resolving transitions so `this.children` can't even be changed "in the middle" of the iteration as event queue is processed one by one. If my reasoning is correct then I would remove this, or leave it as dev-only with a piece of information that this is an internal error and it should be reported to us

packages/core/src/interpreter.ts
```diff
@@ -815,8 +818,9 @@ export class Interpreter<
 
         // Invoked services
         if (activity.type === ActionTypes.Invoke) {
```
Andarist: this shouldn't be named `activity` any longer

packages/core/src/interpreter.ts
```diff
@@ -838,13 +842,24 @@ export class Interpreter<
             return;
           }
 
-          const source = serviceCreator(context, _event.data);
+          const actor = serviceCreator(context, _event.data, {
+            parent: this as any,
```
Andarist: do we need to expose the whole parent to the service creator? couldnt we limit this to a very limited minimum? What kind of properties/methods can be useful for `serviceCreator`?

packages/core/src/StateNode.ts
```diff
@@ -371,7 +371,7 @@ class StateNode<
       if (!this.machine.options.services[resolvedSrc]) {
         this.machine.options.services = {
           ...this.machine.options.services,
-          [resolvedSrc]: invokeConfig.src
+          [resolvedSrc]: invokeConfig.src as any
```
Andarist: Not necessarily as part of this PR but I think we should remove overriding `this.machine.options` internally - from what I remember this is the last place where we do.

If we wouldnt override this we could more easily swap the config with a simple assignment rather than doing this:
https://github.com/davidkpiano/xstate/blob/be2bc0dd388901a894d6a2afd7eef67931194a11/packages/xstate-react/src/index.ts#L87-L93
which doesnt currently even cover all config types (guards and delays stay unhandled now).

packages/core/src/invoke.ts
```diff
@@ -177,15 +209,20 @@ export function spawnCallback<TE extends EventObject = AnyEventObject>(
       parent.send(error(id, err) as any);
     }
 
-    // if (isPromiseLike(callbackStop)) {
-    //   // it turned out to be an async function, can't reliably check this before calling `callback`
-    //   // because transpiled async functions are not recognizable
-    //   return this.spawnPromise(callbackStop as Promise<any>, id);
-    // }
+    if (isPromiseLike(callbackStop)) {
```
Andarist: the beauty of the approach in this PR is that this is not needed, the type of the spawned thing gets declared upfront, so a user really shouldn't pass an async function to `spawnCallback`

packages/core/src/invoke.ts
```diff
@@ -177,15 +209,20 @@ export function spawnCallback<TE extends EventObject = AnyEventObject>(
...
+        parent,
+        id,
+        _event
+      });
+    }
 
     const actor = {
       id,
-      send: event => receivers.forEach(receiver => receiver(event)),
+      send: receivedEvent =>
```
Andarist: I find this a little bit odd, not sure right now what would be the ideal API - just raising my concern out loud. 

Given that we assume everything is an actor it maybe would be more appropriate to allow only for a single receiver on a callback's side. Just like actors have a single mailbox. Registering multiple receivers can only make control flow convoluted and I feel that is an easy way to make a mistake along the way - so if we could figure out an API that would avoid those pitfalls then it would be great.

packages/core/src/invoke.ts
```diff
@@ -151,20 +156,47 @@ export function spawnPromise<T>(
...
+    const callback = callbackCreator(ctx, event);
     let canceled = false;
     const receivers = new Set<(e: EventObject) => void>();
     const listeners = new Set<(e: EventObject) => void>();
 
-    const receive = (e: TE) => {
-      listeners.forEach(listener => listener(e));
+    const receive = (receivedEvent: TE) => {
+      listeners.forEach(listener => listener(receivedEvent));
       if (canceled) {
```
Andarist: this is an important semantic to establish - is it OK for a stopped actor to still emit events to the outside world? shouldnt the `canceled` check be first here to prevent notifying listeners in such case? should current listeners be notified about an actor is being stopped?

packages/core/src/invoke.ts
```diff
@@ -151,20 +156,47 @@ export function spawnPromise<T>(
   };
 }
 
+export function spawnActivity<TC, TE extends EventObject>(
```
Andarist: isnt this equivalent of:
```js
export const spawnActivity = activityCreator => spawnCallback((ctx, e) => {
  const dispose = activityCreator(ctx, e)
  return () => dispose
})
```

packages/core/src/invoke.ts
```diff
@@ -151,20 +156,47 @@ export function spawnPromise<T>(
...
 
-    const receive = (e: TE) => {
-      listeners.forEach(listener => listener(e));
+    const receive = (receivedEvent: TE) => {
+      listeners.forEach(listener => listener(receivedEvent));
       if (canceled) {
         return;
       }
-      parent.send(e);
+      parent.send(receivedEvent);
```
Andarist: i've looked at this really quickly, but it seems that this is the only actor type which doesn't add `origin` to the `receivedEvent`

tivac:
Thanks for the clarification @davidkpiano, seeing it approached that way helps quite a bit!

packages/core/src/invoke.ts
Andarist: A proposal for such an API would be to expect `{send,stop}` return type here. Promises, observables and machines have onown interfaces so we know how to adapt them to the actor API but as callbacka are much more free-form we cannot do that automatically like for those other types. So maybe just require returning this clear pair instead of providing a custom way to integrate callbacks. It would nicely relate to all other places where actor interface is used

packages/core/src/invoke.ts
davidkpiano: Yes, it can be simplified

packages/core/src/invoke.ts
```diff
@@ -1,37 +1,46 @@
...
-    const childService = new Interpreter(machine, {
+): InvokeCreator<TContext, TEvent> {
+  return (ctx, event, { parent, id, data, _event }) => {
+    let resolvedMachine = isFunction(machine) ? machine(ctx, event) : machine;
+    if (data) {
+      resolvedMachine = resolvedMachine.withContext(
+        mapContext(data, ctx, _event)
+      ) as TMachine;
+    }
+    const childService = interpret(resolvedMachine, {
```
Andarist: this.options cannot be used anymore here

packages/core/src/interpreter.ts
davidkpiano: All we need is `parent.send` which is what the `Actor` interface will become (future PR)

packages/core/src/invoke.ts
davidkpiano: Keeping that here for compatibility (tests pass) for now, will later refactor when `spawnCallback` is removed.

packages/core/src/invoke.ts
davidkpiano: `spawnCallback` will be removed in a future PR. This is just to get the tests to pass. You're right, though.

packages/core/src/invoke.ts
davidkpiano: That's up to the implementor. In the real world, there's nothing stopping actors from emitting events. The best the receiver can do is ignore those events.

packages/core/src/invoke.ts
davidkpiano: To be removed

packages/core/src/invoke.ts
Andarist: well - if we plan to remove `subscribe` from actors as discussed on Slack then this comment is irrelevant, actor->parent communication is already guarded in a way I would expect it to

davidkpiano: Right, it will be

packages/core/src/interpreter.ts
davidkpiano: Yes, you can `forwardTo('nonExistantChild')` which will trigger the error.

PR #909, opened 2019-12-30
Always start service immediately for FSM-flavor as well

Andarist:
Porting logic from https://github.com/davidkpiano/xstate/pull/895 to `@xstate/fsm` flavor of `useMachine`.


packages/xstate-react/src/fsm.ts
```diff
@@ -9,27 +10,15 @@ export function useMachine<TC, TE extends EventObject = AnyEventObject>(
...
 
   useEffect(() => {
-    if (!ref.current) {
-      return;
-    }
-
-    ref.current.subscribe(setState);
-    ref.current.start();
-
+    service.subscribe(setCurrent);
```
Andarist: This has a problem now though because it's not possible to read the current state from the service - this can only be received in `onTransition` callback, but I don't want to set up a subscription (yet) in render phase. Going to discuss this with React team at the beginning of the new year as I think it might be safe-ish to subscribe in render phase to a source created by the subscribing component.

davidkpiano: `service.subscribe(...)` should immediately provide the current state.

Andarist: oh! I did not know that, cool, so this would mean that this can be merged as is now - it matches what happens in `@xstate/react` roughly 1to1

PR #918, opened 2020-01-01
Add dev-only error to `useMachine` about receiving different machines between renders

Andarist:


packages/xstate-react/src/fsm.ts
```diff
@@ -10,6 +10,17 @@ export function useMachine<TC, TE extends EventObject = AnyEventObject>(
   StateMachine.Service<TC, TE>['send'],
   StateMachine.Service<TC, TE>
 ] {
+  if (process.env.NODE_ENV !== 'production') {
+    const [initialMachine] = useState(stateMachine);
```
davidkpiano: Hmm, can we do this without using `useState`? Perhaps `useEffect`

Andarist: What problem do you see with useState here? Its either this or useConstant but as this is dev path i’ve chosen state as the simpler one. We need some “memory cell” to keep the initial value, so either state or ref.

Issue #681, opened 2019-09-26
Assign in typescript 


edgarjrg:
**Description**
Assign does not behaves as expected in typescript, it I assign the `Event` type to `assign` it becomes unassignable to the `machine`

**Expected Result**
Be able to type
```ts
const machine = Machine<Context, Schema, Event>({
  id: "machine",
  initial: "inactive",
  context: {
    count: 0
  },
  states: {
    inactive: {
      on: {
        TOGGLE: [
          {
            target: "active",
            actions: [
              assign<Context, Event>({
                //            ^ this is not possible
                count: (ctx, e) =>
                  e.type === "TOGGLE" ? ctx.count + 1 : ctx.count
              })
            ]
          }
        ]
      }
    },
    active: {
      on: {
        TOGGLE: [
          {
            target: "inactive"
          }
        ]
      }
    }
  }
});
```

**Actual Result**
`Type 'AssignAction<Context, Event>' is not assignable to type 'string'.`

**Reproduction**
https://codesandbox.io/s/xstate-typescript-error-with-assign-v18b4

**Additional context**
`"strict": true`
`xstate@4.6.7`

davidkpiano:
What happens if you try this with `xstate@next`?

edgarjrg:
could this be re-opened please 🙏 ?

I tried with `typescript@3.6.2` and `3.6.3` but nothing changed. It works ok without `strict`

edgarjrg:
it works as expected with `xstate@4.7.0-rc3`

PR #924, opened 2020-01-03
Feature: state and transition level partial context

davidkpiano:
This feature is a quick win that allows developers to define `context` partially for states and transitions in the same way that they would with the `assign()` action creator. This is equivalent to the same `assign()` action as the first action:

**State-level `context`**
```diff
// ...
context: {
  count: 0,
  message: 'hello'
},
states: {
  foo: {
-   entry: assign({ count: 10 })
+   context: { count: 10 }
  },
  bar: {
-   entry: assign({ count: ctx => ctx.count + 10 })
+   context: { count: ctx => ctx.count + 10 }
  },
  baz: {
-   entry: assign(ctx => ({...ctx, count: 42 }))
+   context: ctx => ({...ctx, count: 42 })
  }
}
```

**Transition-level `context`**
```diff
// ...
// same assign-like behavior as above
context: {
  count: 0,
  message: 'hello'
},
on: {
  SOME_EVENT: {
-   actions: assign({ count: ctx => ctx.count + 10 })
+   context: { count: ctx => ctx.count + 10 }
  }
}
```

This has a few advantages:

- No need to import `assign` action creator for most use-cases; it's built into `context`
  - Still need it for `exit` actions; `assign()` is not removed and will still work the same
- Less verbose (plain JS)
- Less painful typing (TypeScript), since the types don't need to be inferred from the `assign<TC, TE>(...)` action creator, and instead are explicit on the `context` properties.
  - This solves a huge pain point (#681)
- Not a breaking change and still compatible with SCXML since the `context` is translated to `assign` actions

- [x] State-level `context`
- [x] Transition-level `context`
- [x] Tests
- [x] Docs
- [x] Changelog

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -91,7 +91,8 @@ import {
   initEvent,
   toActionObjects,
   resolveLog,
-  resolveRaise
+  resolveRaise,
+  assign
 } from './actions';
 import { IS_PRODUCTION } from './environment';
 import { DEFAULT_GUARD_TYPE, STATE_DELIMITER } from './constants';
@@ -376,6 +377,11 @@ class StateNode<
     this.onEntry = toArray(this.config.entry || this.config.onEntry).map(
       action => toActionObject(action)
     );
+
+    if (this.parent && this.config.context) {
+      this.onEntry.unshift(assign(this.config.context));
+    }
+
     // TODO: deprecate (exit)
     this.onExit = toArray(this.config.exit || this.config.onExit).map(action =>
       toActionObject(action)
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -1885,9 +1885,15 @@ class StateNode<
 
     const target = this.resolveTarget(normalizedTarget);
 
+    const actions = toActionObjects(toArray(transitionConfig.actions));
+
+    if (transitionConfig.context) {
+      actions.unshift(assign(transitionConfig.context));
+    }
+
     return {
       ...transitionConfig,
-      actions: toActionObjects(toArray(transitionConfig.actions)),
+      actions,
       cond: toGuard(transitionConfig.cond, guards),
       target,
       source: this,
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -149,6 +149,11 @@ export type TransitionTargets<TContext> = Array<
 export interface TransitionConfig<TContext, TEvent extends EventObject> {
   cond?: Condition<TContext, TEvent>;
   actions?: Actions<TContext, TEvent>;
+  /**
+   * The (partial) context that should be assigned to the state
+   * when this transition is taken.
+   */
+  context?: Assigner<TContext, TEvent> | PropertyAssigner<TContext, TEvent>;
   in?: StateValue;
   internal?: boolean;
   target?: TransitionTarget<TContext, TEvent>;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -417,11 +417,9 @@ export interface StateNodeConfig<
    */
   type?: 'atomic' | 'compound' | 'parallel' | 'final' | 'history';
   /**
-   * The initial context (extended state) of the machine.
-   *
-   * Can be an object or a function that returns an object.
+   * The (partial) context that this state should have once entered.
    */
-  context?: TContext | (() => TContext);
+  context?: Assigner<TContext, TEvent> | PropertyAssigner<TContext, TEvent>;
   /**
    * Indicates whether the state node is a history state node, and what
    * type of history:
@@ -603,7 +601,9 @@ export interface MachineConfig<
   TEvent extends EventObject
 > extends StateNodeConfig<TContext, TStateSchema, TEvent> {
   /**
-   * The initial context (extended state)
+   * The initial context (extended state) of the machine.
+   *
+   * Can be an object or a function that returns an object.
    */
   context?: TContext | (() => TContext);
   /**
--- a/packages/core/test/context.test.ts
+++ b/packages/core/test/context.test.ts
@@ -1,4 +1,4 @@
-import { createMachine } from '../src';
+import { createMachine, sendParent, send, interpret } from '../src';
 
 describe('state context', () => {
   it('should assign context when a state is entered (property value)', () => {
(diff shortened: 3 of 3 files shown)


Andarist:
Mostly I'm going to play a devil's advocate here.

I'm not completely sold on this, especially given the v5 work that is in progress. This technically is not a breaking change, but is a binding one - we shouldn't introduce such a feature to remove it soon. So introducing it has to be carefully evaluated.

Good things about the current API are:
- it being just like every other action (as it is one), so teachability is easy 
- it's actually more explicit and thus more readable - all consequences are more visible when looking at the written code, wherein a proposed form some of them are hidden
- doesn't add any code to the core. I also recall you mentioning that XState is sort of a low-level tool by design and this goes in an opposite direction, it's a sugar syntax so it's an opposite of the low level

I personally agree with most of the things listed on [this slide](https://www.youtube.com/watch?v=4anAwXYqLG8&feature=youtu.be&t=1441), especially with:
- explicit repetitive code
- almost never add sugar
- burn as much as you add

The one that striked me most is "explicit repetitive code". I think that we can agree that in general explicit is better than implicit as the latter carries over builtin magic. Question is where this feature here lands on the implicit vs explicit spectrum.

> Not a breaking change and still compatible with SCXML since the context is translated to assign actions

This is not completely true as discussed on Slack. In SCXML `<datamodel>` works in a different way than this and while we could add in a future `datamodel` property that could work like the one in SCXML with `context` preserving the behavior from this PR it would be IMHO **really** confusing and this should rather be avoided. Those 2 potential properties are looking too much like each other.

The `datamodel` argument is in my opinion strong enough to delay the decision about introducing this till v5, because at that point in time many things should get clarified, mainly regarding our SCXML-compatibility. This might be especially important given the fact that I see people interested in XState from non-JS backgrounds, some actually coming knowing already about SCXML. So providing an API that suggests it's a different thing (it looks super close to SCXML datamodel) than it is, is dangerous - it breaks an expected, by some people, contract.

That being said - I recognize the problem with `assign` typings, but I'm not yet convinced that this is a problem that can't be solved.

davidkpiano:
This won't be 1-to-1 with the `<datamodel>` prop, since `<datamodel>` values defined within states are bound regardless of if the state is active or not:

```xml
<scxml xmlns="http://www.w3.org/2005/07/scxml" xmlns:conf="http://www.w3.org/2005/scxml-conformance" initial="s0" version="1.0" binding="early" datamodel="ecmascript">
   <state id="s0">
       <transition cond="Var1" target="pass"/>
       <transition target="fail"/>
   </state>
   
    <state id="s1">
	  <datamodel>
        <data id="Var1">
          [1,2,3]
        </data>
      </datamodel>
    </state>

   <final id="pass"><onentry><log label="Outcome" expr="'pass'"/></onentry></final>
   <final id="fail"><onentry><log label="Outcome" expr="'fail'"/></onentry></final>
</scxml>
```

Source: https://gitlab.com/scion-scxml/test-framework/blob/master/test/w3c-ecma/test551.txml.scxml

davidkpiano:
> This technically is not a breaking change, but is a binding one - we shouldn't introduce such a feature to remove it soon.

It will not be removed in V5, and is the direction it should be taken, given heavy usage of `assign()` in the wild.

> Question is where this feature here lands on the implicit vs explicit spectrum.

It is explicit and declarative (disregarding implementation details). Consider the difference:

```js
loading: {
  context: { user: null }
}
```

Reads as: "The loading state where, initially, the `user` in the context is `null`"

```js
loading: {
  entry: assign({ user: null })
}
```

Reads as: "The loading state. Also, on entry, assign the `user` in the context to `null`"

The former is more declarative, the latter imperative. We should strive towards a declarative approach rather than an imperative one, even if it's still declarative at the end of the day. Readability is a strong motivating goal here.

davidkpiano:
Alternative proposal that will hopefully mitigate your concerns:

```diff
context: {
  count: 0,
  message: 'hello'
},
states: {
  foo: {
-   entry: assign({ count: 10 })
+   assign: { count: 10 }
    on: {
      INC: {
-       actions: assign({ count: ctx => ctx.count + 1 })
+       assign: { count: ctx => ctx.count + 1 }
      }
    }
  },
}
```

tivac:
I'm a big fan of this proposal and the latest version with `assign` does seem clearer. First iteration sort of made it look like `context` could be local to the state instead of a machine-level concept. That definitely tripped me up when learning xstate!

Andarist:
> Reads as: "The loading state where, initially, the user in the context is null"

Only when you get more familiar with the API and how it works internally, the part about "initially" is hidden and isn't that obvious at first. It might pose some questions for a reader that only documentation or experimentation might answer.

Comparing to it to the current way of writing this - the information is much more visible, requiring little to no documentation around this.

> The former is more declarative, the latter imperative.

I strongly disagree with this statement. IMHO **both** are declarative - the one is just sugar over the other, it doesn't make it much more declarative. In fact, as mentioned, I think it makes it less informative because it hides some information (but that might be just my opinion).

> Alternative proposal that will hopefully mitigate your concerns, and actually more closely reflects the SCXML spec:

This would certainly address the SCXML part of my concerns and would be an improvement. There is currently a slight difference in how we treat the configuration in regards to the root node - in SCXML there is no root state, it's not a state, but we, sort of, treat it as a state. It's a little bit important here - because if we'd translate XState to SCXML just changing the configuration format then we wouldn't produce a valid SCXML. What I'm saying is that if we want to mirror SCXML closely then `context` property should be allowed currently just on the root (an `<scxml>` element) and `assign` should not be allowed there. The opposite is true for states - `context` shouldn't be allowed but `assign` should be (if we merge this PR).

> First iteration sort of made it look like context could be local to the state instead of a machine-level concept. That definitely tripped me up when learning xstate!

This is IMHO just an argument for my case - this reads **worse** than `entry: [assign()]`, even considering a context->assign change. Not knowing XState and reading through docs one could make a similar mistake, while it's way harder to make it with `entry: [assign()]`. You only need to learn a **single** powerful API - entry/exit/transition actions and you are good to go. Everything works in the very same way, it executes at an immediately known point in time and it's harder to misinterpret it.

davidkpiano:
I'm going to target V5 for this change instead, using the `assign` property instead of the `context` property (which, I agree, would be confusing).

The goal of V5 in terms of assignment will be to encourage and maintain the behavior of V4, where assignment is a "priority" action (as it is now), while providing a simpler syntax for it:

```js
// ...
context: {
  count: 0,
  message: 'hello'
},
states: {
  foo: {
-   entry: assign({ count: ctx => ctx.count + 10 })
+   assign: { count: ctx => ctx.count + 10 }
  },
  // ...
}
```

Although SCXML does not have priority actions:
- Assignment is currently treated as priority (this is a known issue)
- Assignment being priority prevents poor modeling, by encouraging explicit intermediate states
- Multiple assignment should be discouraged as well, since that results in lack of clarity
- The assign action priority issue is "fixed" by processing actions in-order instead of prioritizing; this is a breaking change (only applies to the `assign(...)` action creator).

Note that XState is not necessarily meant to be completely low-level, as the library exists for providing helpful abstractions, and others exist (`invoke` syntax, sugar around single vs. array actions, `after` delayed transitions are a pretty heavy but useful abstraction, etc.)

Additionally, this paves the way for what I propose as "virtual mode", which combines the ideas of the [P Programming Language](https://github.com/p-org/P) (a state-machine-first language) and the immutability of [immer](https://github.com/immerjs/immer) for a very natural, idiomatic way to represent updating context that also closely represents how context is updated (mutated) in SCXML and other UML/statecharts environments:

```js
states: {
  foo: {
    target: 'bar',
    assign(ctx) {
      // not actually mutated!
      ctx.count += 10;
    }
  }
}
```

This "virtual mode" idea can be expanded to states:

```js
states: {
  // This "virtual" transition definition is only run once
  // in order to determine the declarative transition definition
  foo(ctx) {
    ctx.count += 10;
    return 'bar';
  }
}
```

These are tentative ideas, and we should not be afraid to shy away from abstractions like the above that greatly improve the developer experience (especially with TypeScript). Happy to hear thoughts on this.

CodingDive:
> ``` javascript
> states: {
>   foo: {
>     target: 'bar',
>     assign(ctx) {
>       // not actually mutated!
>       ctx.count += 10;
>     }
>   }
> }
> ```

I like this idea a lot as it lowers the barrier to entry for XState as people won't need to learn about immutability before working with extended state. 

I see two problems for applying the virtual mode concept to states. 
1. Readability: On every event, I'd have to check if the return value is a state node or if `assign` has been called before and I'm dealing with context. It's nice syntactic sugar but to me, it blurs the lines between a state transition and mutating context too much. 
2. It could make visualization harder although this problem might be solved by your comment
> This "virtual" transition definition is only run once in order to determine the declarative transition definition
3. Encourage non-deterministic state transitions. 

```javascript
states: {
  temperature: {
    states: {
      cold: {},
      warm: {},
    },
    on: {
      HEAT: (ctx) => {
        ctx.temperature += 5;
        return ctx.temperature > 20 ? '.warm' : '.cold';
      }
    }
  }
}
```

Of course, XState could internally memorize the transition of each event and throw an error once it detects that a different transition is taken. I think this couldn't be determined before runtime. Nevertheless, with an error saying that the above code should be refactored to multiple transitions and guards, my concern might not be a problem.

My opinion might be an exception to your proposal as I also never use the `on: { "SOME_EVENT": "NEW_STATE" }` syntax and always refactor the code to
`on: { "SOME_EVENT": { target: "NEW_STATE" } }` for the sake of readability.

davidkpiano:
> It could make visualization harder although this problem might be solved by your comment

It would be the exact same as it is right now; everything will be visualized.

> Encourage non-deterministic state transitions.

This is impossible because the function is only ever run once - at the beginning.

But I get your concerns. For context/prior art, take a look at the P programming language: https://github.com/p-org/P/wiki/PingPong-program

```p
// PingPong.p
event PING assert 1: machine;
event PONG assert 1;
event SUCCESS;

main machine Client {
    var server: machine;

    start state Init {
        entry {
  	        server = new Server();
	        raise SUCCESS;   	   
        }
        on SUCCESS goto SendPing;
    }

    state SendPing {
        entry {
	          send server, PING, this;
	          raise SUCCESS;
	}
        on SUCCESS goto WaitPong;
    }

    state WaitPong {
        on PONG goto SendPing;
    }
}

machine Server {
    start state WaitPing {
        on PING goto SendPong;
    }

    state SendPong {
	      entry (payload: machine) {
	          send payload, PONG;
	          raise SUCCESS;		 	  
	      }
        on SUCCESS goto WaitPing;
    }
}
```

davidkpiano:
Closing this as it conflicts with in-order assigns.

Issue #718, opened 2019-10-10
@xstate/test coverage fails for states which transitions on entry


miketamis:
**Description**
If you have a transition that happens on entry like:
```js
  const TestBug = Machine({
    id: 'testbug',
    initial: 'idle',
    context: {
      retries: 0
    },
    states: {
      idle: {
        on: {
          START: 'passthrough'
        }
      },
      passthrough: {
        on: {
          '': 'end'
        }
      },
      end: {
        type: 'final'
      },
    }
  });
  ```
it will give you a coverage error for passthrough.
**Expected Result**
Coverage will not show an error if the state is passthrough
**Actual Result**
coverage error shown even though passthrough situation tested
**Reproduction**
https://codesandbox.io/s/xstate-example-template-ufm4u?fontsize=14

PR #925, opened 2020-01-03
@xstate/test: Introduce coverage options

davidkpiano:
This PR introduces options for `testModel.getCoverage(options)` and `testModel.testCoverage(options)`:

```js
// Only test coverage for state nodes with a `.meta` property defined:
 testModel.testCoverage({
  filter: stateNode => !!stateNode.meta
});
```

Related issue: #718

### DIFF
--- a/.changeset/fifty-pears-hammer.md
+++ b/.changeset/fifty-pears-hammer.md
@@ -1,0 +1,5 @@
+---
+'@xstate/test': minor
+---
+
+Options for `testModel.getCoverage()` and `testModel.testCoverage()` can now be provided to filter which state nodes should be covered by the tests.
--- a/packages/xstate-test/src/index.ts
+++ b/packages/xstate-test/src/index.ts
@@ -311,11 +311,10 @@ export class TestModel<TTestContext, TContext> {
     options?: CoverageOptions<TContext>
   ): { stateNodes: Record<string, number> } {
     const filter = options ? options.filter : undefined;
-    const stateNodes = getStateNodes(this.machine).filter(
-      filter || (() => true)
-    );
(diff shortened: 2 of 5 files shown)


packages/xstate-test/src/index.ts
```diff
@@ -306,8 +307,13 @@ export class TestModel<TTestContext, TContext> {
     }
   }
 
-  public getCoverage(): { stateNodes: Record<string, number> } {
-    const stateNodes = getStateNodes(this.machine);
+  public getCoverage(
+    options?: CoverageOptions<TContext>
+  ): { stateNodes: Record<string, number> } {
+    const filter = options ? options.filter : undefined;
+    const stateNodes = getStateNodes(this.machine).filter(
```
Andarist: maybe instead of providing a default filter in a form of `() => true` you could only call filter if needed?
```js
const allStateNodes = getStateNodes(this.machine)
// or rename the variable below to filteredStateNodes, but that would have to be propagated to the code using this
const stateNodes = options.filter ? allStateNodes.filter(options.filter) : allStateNodes
```

.changeset/fifty-pears-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/test': minor
+---
+
+Options for `testModel.getCoverage()` and `testModel.testCoverage()` can now be provided to filter which state nodes should be covered by the tests.
```
Andarist: We could mention here how actually one is supposed to filter state nodes - either by mentioning `filter` option or linking to the docs.

davidkpiano: Will do that pre-release (and once the docs are published)

PR #931, opened 2020-01-06
Add `types` field to all package.json's

Lavoaster:
IntelliJ based IDE's have difficulty trying to find the types unless you use the main path directly. I.E "@xstate/react/lib/index"

This is most likely down to the wording on the typescript publish page. (https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)

It says "Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.", so given that the types aren't at the root it isn't resolving it.

Adding this in fixes the issue.

### DIFF
--- a/.changeset/young-badgers-yawn.md
+++ b/.changeset/young-badgers-yawn.md
@@ -1,0 +1,11 @@
+---
+'@xstate/analytics': patch
+'@xstate/graph': patch
+'@xstate/immer': patch
+'@xstate/react': patch
+'@xstate/scxml': patch
+'@xstate/test': patch
+'@xstate/vue': patch
+---
+
+Added "types" to package.json
--- a/.changeset/young-badgers-yawn.md
+++ b/.changeset/young-badgers-yawn.md
@@ -1,7 +1,4 @@
 ---
-'@xstate/analytics': patch
-'@xstate/graph': patch
-'@xstate/react': patch
 '@xstate/test': patch
 ---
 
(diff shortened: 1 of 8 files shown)


.changeset/young-badgers-yawn.md
```diff
@@ -0,0 +1,11 @@
+---
+'@xstate/analytics': patch
```
Andarist: This changeset is OK - but atm we dont intend to release all packages, could you include only test package here (we publish core xstate, fsm and test packages right now)? Changes in all package.jsons should be kept.

Lavoaster: I believe I should be guarded by the `private` field in those packages?

Lavoaster: Anyway, I've just removed all packages in here that were marked as private in their package.json

Andarist: Private guards us only from publishing to npm but i want to avoid other release stuff as well - version bumping, changelog generation (covered by changesets). I plan to clean this up soon so we dont have to worry about it - but atm it is what it is.

Lavoaster: Do you just want me to remove the change set then? 😛

Andarist: Nope, right now I'd like to just keep `@xstate/test` in it - nothing more. I understand this is confusing, that's why I intend to clean things up in the near future so we can just assume that everything is publishable.

Lavoaster: 👍

Andarist:
Thanks! ❤️ Keep in mind that your problem will not be immediately fixed, because you were looking to fix primarily `@xstate/react`. We plan to release final v1 of that package in January, but until that happens there probably won't be any intermediate releases published.

PR #932, opened 2020-01-07
Vue Package xstate-machine renderless component

AleksejDix:
To follow up on the 
https://github.com/davidkpiano/xstate/pull/864 
https://github.com/davidkpiano/xstate/pull/874

Here is a clean PR with removed watchers

### DIFF
--- a/packages/xstate-vue/src/xstate-machine.ts
+++ b/packages/xstate-vue/src/xstate-machine.ts
@@ -1,0 +1,95 @@
+import { interpret, State } from 'xstate';
+import Vue, { VNode } from 'vue';
+
+export default Vue.extend({
+  render(h): VNode {
+    const { current, service, send } = this;
+    return h('div', [
+      this.$scopedSlots.default({
+        current,
+        service,
+        send
+      })
+    ]);
+  },
+  props: {
+    machine: {
+      type: Object,
+      required: true,
+      default: () => ({})
+    },
+    options: {
+      type: Object,
+      default: () => ({})
+    }
+  },
+  data() {
+    return {
+      current: null
+    };
+  },
+  computed: {
+    initialState() {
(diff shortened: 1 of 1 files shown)


packages/xstate-vue/src/xstate-machine.ts
```diff
@@ -0,0 +1,95 @@
+import { interpret, State } from 'xstate';
+import Vue, { VNode } from 'vue';
+
+export default Vue.extend({
+  render(h): VNode {
+    const { current, service, send } = this;
+    return h('div', [
```
davidkpiano: Probably a naïve question on my part, but is `'div'` necessary? Does it need to be specified, and what if the developer wants a different component for this? Or no component at all? Just thinking out loud 💭

AleksejDix: it is requeried since vue template must have 1 parent element (it was like this in react before Fragment was introduced)

studnitz: I think this may not be needed. You should be able to do this:
```js
render: VNode {
	const { current, service, send } = this;
    return this.$scopedSlots.default({
        current,
        service,
        send
      })
	},
}
```

I'm going to open a PR to fix this. :)

Andarist: @studnitz while doing this, could you also add some simple tests for this?

studnitz: Yea, was planning on that anyways.

Andarist: @studnitz do you know when you will have time to work on this? I dont want to rush you of course, just would love to know approximate timing.

studnitz: I'm trying to tackle it this evening!

PR #936, opened 2020-01-08
Escalate action with expression

davidkpiano:
This PR adds the ability for `escalate()` to take in an expression, just like `send` or `sendParent`.

### DIFF
--- a/.changeset/lazy-windows-tan.md
+++ b/.changeset/lazy-windows-tan.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/lazy-windows-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
```
Andarist: technically this is a somewhat new feature, but i guess we treat minor versions as more grandiose ones than smth like this 😅

davidkpiano: It's a missing feature that should have been included 😜

PR #940, opened 2020-01-09
Add Typestate to useMachine and useService

davidkpiano:


.changeset/rude-peaches-lick.md
```diff
@@ -0,0 +1,5 @@
+---
```
Andarist: actually - right now this one should probably be removed because it will be counted towards a release but we do not release `@xstate/react` with changesets (yet)

Issue #949, opened 2020-01-17
Type error with raise event


CodingDive:
**Description**
When using the `raise` action, I'm getting type errors.

Given an event like this:

```typescript
type IRaiseEvent = { type: 'FOO' }
```

I'm getting an unreadable TypeScript error when raising an event inside a machine.

```typescript
// error
actions: [
  raise({type: 'FOO'})
]

// error also occurs when using type arguments + object event
actions: [
  raise<IMachineContext, IRaiseEvent>({type: 'FOO'})
]

// error when using type arguments + string event
actions: [
  raise<IMachineContext, IRaiseEvent>('FOO')
]
```

The only thing that seems to work is to not pass type arguments and use a string event instead of an object.

```typescript
// no error
actions: [
  raise('FOO')
]
```

**Expected Result**
No type error

**Actual Result**
TypeScript fails on the line where I'm using the `Machine` factory function and passing the context, state schema, and discriminatory events, not on the line where an event is `raise`d. As a result, it took me a while to identify the line at fault. 
The error starts with `the types of 'some.state' are incompatible between these types` and then proceeds to list all the state nodes + events of the machine.

```
"xstate": "4.7.6"
"typescript": "^3.7.4"
```

I can't reproduce this in a CodeSandbox right now but I could refactor a test that uses raise events and add type arguments to the machine? E.g adding type arguments to the [transient.test.ts](https://github.com/davidkpiano/xstate/blob/771becf761c14319795685feda8b2f015a75707d/packages/core/test/transient.test.ts#L5) might be enough to reproduce the issue.

davidkpiano:
Sure, please make a PR that demonstrates the issue in a failing test.

PR #951, opened 2020-01-18
Removed `TContext` generic type parameter from `raise` action creator

Andarist:
relates to #949 

this doesn't fix the issue completely yet, but i believe the change is sound and makes things a little bit easier, TS shouldnt be able to assign inferred `unknown` to parametrized `TContext` after all


davidkpiano:
This would disallow the user from using event expressions.

Andarist:
Oh, right. This is not a good change.

PR #952, opened 2020-01-18
Allow any event to be raised. Fixes #949

### LINKED ISSUE #949: Type error with raise event

davidkpiano:


packages/core/test/types.test.ts
```diff
@@ -202,103 +202,95 @@ describe('Nested parallel stateSchema', () => {
...
+    >({
+      key: 'greeting',
+      context: greetingContext,
+      initial: 'pending',
+      states: {
+        pending: {
+          on: {
+            DECIDE: [
+              {
+                // This one does not work
```
Andarist: those comments are no longer valid if the issue got fixed, also - the whole test could be simplified I think, we only need 2x `raise`, one with explicit type arguments and one without any

Issue #655, opened 2019-09-06
Split StateMachine & StateNode


Andarist:
Just a place for discussing this potential but rly low priority refactor.

It's confusing at times that StateNode is also a StateMachine and I've started to believe that splitting those could result in slightly easier to follow code.

Many of the methods/properties of the StateNode seems to be exclusive for StateMachine, i.e. public `.transition` doesn't make much sense on a nested StateNode. And also vice-versa there are probably things needed only for "private" StateNode which wouldn't have to be kept on StateMachine.

Thoughts?

davidkpiano:
Definitely want to do this. Just to outline the hierarchy:

Base class: `StateNode` (would be much smaller than the current `StateNode`

- `StateMachine` extends `StateNode`
- `AtomicStateNode` extends `StateNode`
- `CompoundStateNode` extends `StateNode`
- `OrthogonalStateNode` extends `StateNode`
- `HistoryStateNode` extends `StateNode`

How's that?

Andarist:
Might be beneficial as well. My main concern is about StateMachine being combined with StateNode though - had no reasoning issues while reading/developing about different node types being the same class.

davidkpiano:
Okay, that can be a good first step - having `StateMachine` extend `StateNode`.

PR #953, opened 2020-01-19
V5: MachineNode (and StateNode refactor)

davidkpiano:
This PR splits `StateNode` into `MachineNode` and `StateNode`:

- `MachineNode` extends `StateNode`
- All machine-specific methods now only belong to `MachineNode`
- Functionality/logic for nodes is separated into functions in `nodeUtils.ts`
- `context` is removed as a 3rd parameter from `new MachineNode(config, options, context)`.

This is an intermediate (refactor) PR, and will be the basis for future PRs.

Addresses #655


packages/core/src/Machine.ts
```diff
@@ -26,26 +25,19 @@ export function Machine<
...
-  return new StateNode<TContext, TStateSchema, TEvent, any>(
+  options?: Partial<MachineOptions<TContext, TEvent>>
+): MachineNode<TContext, TStateSchema, TEvent> {
+  return new MachineNode<TContext, TStateSchema, TEvent, any>(
     config,
-    options,
-    resolvedInitialContext
-  ) as StateMachine<TContext, TStateSchema, TEvent>;
+    options
+  ) as MachineNode<TContext, TStateSchema, TEvent>;
```
Andarist: this cast most likely is not needed anymore as you instantiate `MachineNode` which is the same as the expected return type, whereas previously instantiated class was different than the return type

That being said - not sure if using "node" in the name of this public type wont be slightly confusing, after all a user doesnt call into `MachineNode` or `createMachineNode`. Seems like the previous name (StateMachine) could have been more intuitive for users. Additionally - we can treat machine as a node, but at least in the SCXML it is not. Well - it's not described in SCXML in details, but when reading through the spec one can conclude it's not a node. It has different possible descendent elements and, what's even more important, one cannot transition to it. Thinking about nodes as the stuff we can transition to simplifies the mental model - states (nodes) represent possible configuration, they are directly represented on the visualized graph, whereas a machine is more like a wrapper containing the graph.

packages/core/test/scxml.test.ts
```diff
@@ -348,7 +353,7 @@ interface SCIONTest {
 }
 
 async function runW3TestToCompletion(machine: StateNode): Promise<void> {
-  await new Promise((resolve, reject) => {
+  await new Promise((res, reject) => {
```
Andarist: why the change? :p you have shortened just `resolve` to `res` but at the same time, you have left `reject` untouched.

packages/core/test/meta.test.ts
```diff
@@ -72,8 +72,9 @@ describe('state meta data', () => {
     }
   });
 
-  it('states should aggregate meta data', () => {
+  it.only('states should aggregate meta data', () => {
     const yellowState = lightMachine.transition('green', 'TIMER');
+    console.log(yellowState);
```
Andarist: leftover console.log

packages/core/test/meta.test.ts
```diff
@@ -72,8 +72,9 @@ describe('state meta data', () => {
     }
   });
 
-  it('states should aggregate meta data', () => {
+  it.only('states should aggregate meta data', () => {
```
Andarist: this also is most likely not an intentional change 😉

packages/core/test/interpreter.test.ts
```diff
@@ -950,10 +949,9 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
         start: {
           invoke: {
             id: 'child',
-            src: spawnMachine(childMachine),
-            data: {
-              password: 'foo'
-            }
+            src: spawnMachine(() => {
```
Andarist: was the change required? `.withContext` works slightly differently than what was represented by `invoke.data` (not in this case here, but in other cases)

packages/core/src/utils.ts
```diff
@@ -593,7 +593,7 @@ export function toTransitionConfigArray<TContext, TEvent extends EventObject>(
     if (
       typeof transitionLike === 'undefined' ||
       typeof transitionLike === 'string' ||
-      isMachine(transitionLike)
+      isMachineNode(transitionLike)
```
Andarist: is this check `necessary`? a machine itself cannot be a target of a transition

packages/core/src/utils.ts
```diff
@@ -526,9 +526,9 @@ export function isObservable<T>(
 export const symbolObservable = (() =>
   (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
 
-export function isMachine(value: any): value is StateMachine<any, any, any> {
+export function isMachineNode(value: any): value is MachineNode<any, any, any> {
   try {
-    return '__xstatenode' in value;
+    return '__xstatenode' in value && value.parent === undefined;
```
Andarist: if this `isMachineNode` function is supposed to only return true for MachineNodes then we should be able to move this `__xstatenode` property onto MachineNode class and remove it from other classes (and optionally rename this property then). I dont think we really need to have this special property for state nodes, do we?

packages/core/src/nodeUtils.ts
```diff
@@ -0,0 +1,1210 @@
+import {
```
Andarist: I like this file very much 👍

packages/core/src/invoke.ts
```diff
@@ -19,18 +18,20 @@ import {
...
   TContext,
   TEvent extends EventObject,
-  TMachine extends StateMachine<any, any, any>
+  TMachine extends MachineNode<any, any, any>
 >(
   machine: TMachine | ((ctx: TContext, event: TEvent) => TMachine),
   options: { sync?: boolean } = {}
 ): InvokeCreator<TContext, TEvent> {
   return (ctx, event, { parent, id, data, _event }) => {
+    console.log(data);
```
Andarist: 2 leftover console.logs in here

packages/core/test/scxml.test.ts
davidkpiano: During development there was a `resolve()` function - avoided collision

packages/core/src/utils.ts
davidkpiano: This is meant in terms of `isStateNode(...)` but I don't have that function yet.

packages/core/test/meta.test.ts
tivac: We use [`eslint-plugin-jest`](https://www.npmjs.com/package/eslint-plugin-jest) to catch these in the linter and fail builds if any test is focused using the [`no-focused-tests`](https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/no-focused-tests.md) rule. Highly recommended, it's saved me from myself many times.

packages/core/src/Machine.ts
davidkpiano: Let's defer separating the root node to a different PR. Notice that the `<scxml>` children and `<state>` children have children in common:

- `<state>`
- `<parallel>`
- `<final>`

It might be redundant to model it differently than a compound `<state>` node, but let's explore this in a separate PR.

packages/core/test/interpreter.test.ts
davidkpiano: Reverted

packages/core/src/types.ts
```diff
@@ -571,55 +550,21 @@ export interface MachineConfig<
   /**
    * The initial context (extended state)
    */
-  context?: TContext | (() => TContext);
+  context?: TContext;
```
Andarist: Dropping support for `(() => TContext)` should also be noted down - even if it's temporary, because we can't be sure when we are going to bring it back.

packages/core/test/machine.test.ts
```diff
@@ -184,15 +184,53 @@ describe('machine', () => {
...
+      //   { foo: 'different' }
+      // );
+      // expect(differentMachine.initialState.context).toEqual({
+      //   foo: 'different'
+      // });
+    });
+  });
+
+  describe('machine.withContext', () => {
+    it('should partially override context', () => {
```
Andarist: This I believe is a new behavior, right? We should note it down 😉

packages/core/test/machine.test.ts
```diff
@@ -294,22 +332,6 @@ describe('machine', () => {
 
       expect(versionMachine.version).toEqual('1.0.4');
     });
-
-    it('should show the version on state nodes', () => {
```
Andarist: This is a breaking change, so should be communicated with changesets.

packages/core/src/MachineNode.ts
```diff
@@ -0,0 +1,351 @@
...
+  activities: {},
+  delays: {},
+  context
+});
+
+function resolveContext<TContext>(
+  context: TContext,
+  partialContext: Partial<TContext>
+): TContext {
+  if (context === undefined) {
```
Andarist: I think this might be a bug - don't we allow right now **not** defining context upfront (during "first" `Machine` call) and supplying it later for example through `withContext` or with second argument to `useMachine`? If yes - then this logic won't use the supplied context if the machine "template" has no context defined on it.

packages/core/src/MachineNode.ts
```diff
@@ -0,0 +1,351 @@
...
+> extends StateNode<TContext, TStateSchema, TEvent> {
+  public context: TContext;
+  /**
+   * The relative key of the state node, which represents its location in the overall state value.
+   */
+  public key: string;
+  /**
+   * The unique ID of the state node.
+   */
+  // public id: string;
```
Andarist: https://twitter.com/davidkpiano/status/1220739686802640896 :trollface:

packages/core/src/MachineNode.ts
```diff
@@ -0,0 +1,351 @@
...
+  TContext = any,
+  TStateSchema extends StateSchema = any,
+  TEvent extends EventObject = EventObject,
+  TTypestate extends Typestate<TContext> = any
+> extends StateNode<TContext, TStateSchema, TEvent> {
+  public context: TContext;
+  /**
+   * The relative key of the state node, which represents its location in the overall state value.
+   */
+  public key: string;
```
Andarist: this is already part of the StateNode, so we don't have to redefine it here

Andarist: And same can be said about many other properties defined here - on the MachineNode

packages/core/src/MachineNode.ts
```diff
@@ -0,0 +1,351 @@
...
+  /**
+   * The order this state node appears. Corresponds to the implicit SCXML document order.
+   */
+  public order: number = -1;
+
+  public options: MachineOptions<TContext, TEvent>;
+
+  public __xstatenode: true = true;
+
+  public __cache = {
```
Andarist: is this supposed to be public?

packages/core/src/MachineNode.ts
```diff
@@ -0,0 +1,351 @@
...
+      context = this.context
+    } = this.options;
+
+    return new MachineNode(this.config, {
+      actions: { ...actions, ...options.actions },
+      activities: { ...activities, ...options.activities },
+      guards: { ...guards, ...options.guards },
+      services: { ...services, ...options.services },
+      delays: { ...delays, ...options.delays },
+      context: { ...context, ...options.context }
```
Andarist: if we already have `resolveContext` helper then it should be used here as well

packages/core/src/MachineNode.ts
```diff
@@ -0,0 +1,351 @@
...
+  public withConfig(
+    options: Partial<MachineOptions<TContext, TEvent>>
+  ): MachineNode<TContext, TStateSchema, TEvent> {
+    const {
+      actions,
+      activities,
+      guards,
+      services,
+      delays,
+      context = this.context
```
Andarist: I believe this has a slight bug - we can have a machine config with a defined context and create a machine based on it with context option. If we call `withConfig` on it later we might end up with an incorrect value of `context` here because it will be the one from `this.options.context`, but it should actually always just point to `this.context` (which is a resolved context).

I would just recommend removing this line from here and use `this.context` directly when calling MachineNode constructor those few lines below.

packages/core/src/Machine.ts
```diff
@@ -26,26 +25,16 @@ export function Machine<
   config: MachineConfig<TContext, TStateSchema, TEvent>,
   options?: Partial<MachineOptions<TContext, TEvent>>,
   initialContext?: TContext
```
Andarist: overloads here still mention `initialContext` parameter, but it's no longer supported

packages/core/src/MachineNode.ts
```diff
@@ -0,0 +1,351 @@
...
+    if (!initialStateValue) {
+      throw new Error(
+        `Cannot retrieve initial state from simple state '${this.id}'.`
+      );
+    }
+
+    return getInitialState(this, initialStateValue);
+  }
+
+  public getStateNodeById(id: string): StateNode<TContext, any, TEvent> {
```
Andarist: Many methods have been removed in favor of state utils - should we do the same with this one?

packages/core/src/StateNode.ts
```diff
@@ -50,108 +29,43 @@ import {
...
+  getStateNodeById,
+  getRelativeStateNodes,
+  nodesFromChild,
+  evaluateGuard,
+  isStateId
+} from './nodeUtils';
+import { MachineNode } from './MachineNode';
+import { STATE_DELIMITER } from './constants';
+
+export const NULL_EVENT = '';
```
Andarist: those constants are now defined both in StateNode and in MachineNode, they should be just moved to `./constants.ts`

packages/core/src/StateNode.ts
```diff
@@ -187,7 +97,7 @@ class StateNode<
    * Whether the state node is "transient". A state node is considered transient if it has
    * an immediate transition from a "null event" (empty string), taken upon entering the state node.
    */
-  private _transient: boolean;
+  public _transient: boolean;
```
Andarist: if this has been made public I think we should drop `_` prefix

packages/core/src/StateNode.ts
```diff
@@ -358,89 +236,15 @@ class StateNode<
...
-
   /**
    * The well-structured state node definition.
    */
   public get definition(): StateNodeDefinition<TContext, TStateSchema, TEvent> {
     return {
       id: this.id,
       key: this.key,
-      version: this.version,
+      version: this.machine.version, // TODO: fix
```
Andarist: does this need any fixing? from what I see `this.machine` should always be defined and thus `this.machine.version` should just be available at all times

packages/core/src/StateNode.ts
```diff
@@ -463,6 +267,39 @@ class StateNode<
     return this.definition;
   }
 
+  /**
+   * The services invoked by this state node.
+   */
+  public get invoke(): Array<InvokeDefinition<TContext, TEvent>> {
```
Andarist: Previously this was only evaluated once - in the constructor. You have made it a getter now, but its implementation is impure so it poses a risk of introducing a bug in the future, even if there is none now. At the very least its return should be cached (like it's done for the most of other getters defined on StateNode).

Andarist: Well - it isn't pure, but it's idempotent (sort of), so I guess my remark about it posing a risk is an exaggeration, but probably it's better to cache this anyway.

packages/core/src/MachineNode.ts
davidkpiano: This is something we have to think through, but we should _not_ allow context to be `undefined` whenever context is expected to be defined. This is because it's very possible to do `const someMachine = createMachine(...)` without ever defining context, and this is a footgun.

The current pattern for this (that I will encourage in the docs, unless we design another API around it) is to provide factories for machines with dynamic initial contexts:

```js
const createUserMachine = (userId) => createMachine({
  initial: 'pending',
  context: {
    userId
  },
  // ...
});
```

davidkpiano: This code is for the specific case of when the `context` should be `undefined` (context-less machine - only finite states)

packages/core/src/MachineNode.ts
davidkpiano: Probably but not quite yet, since this one is used frequently. I want to figure out the import story first (flat imports, `xstate/utils` instead of `xstate/lib/utils`)

packages/core/src/MachineNode.ts
Andarist: > This is something we have to think through, but we should not allow context to be undefined whenever context is expected to be defined.

Agreed, I've tried to address this in the past with https://github.com/davidkpiano/xstate/pull/588 but it was a breaking change back then so we couldn't pursue this further.

It should be fairly simple to introduce this constraint on the type-level and we can even introduce some runtime dev-only validations. I've added https://github.com/davidkpiano/xstate/issues/583 to v5.0 milestone so we don't miss this.

packages/core/src/StateNode.ts
```diff
@@ -360,7 +360,11 @@ export class StateNode<
     let nextStateNodes: Array<StateNode<TContext, any, TEvent>> = [];
     let selectedTransition: TransitionDefinition<TContext, TEvent> | undefined;
 
-    for (const candidate of getCandidates(this, eventName)) {
+    const candidates =
```
Andarist: good change of moving the caching out of `getCandidates` itself 👍

packages/core/src/MachineNode.ts
Andarist: If my hunch was right about this being a slight bug then it would be good to add a test covering this.

davidkpiano: The line was removed (this code is outdated) - does this still apply?

.changeset/five-ravens-own.md
```diff
@@ -0,0 +1,15 @@
+---
```
Andarist: This doesnt matter much here, but I believe those should get their own changesets. Think about those as separate entries - I think making them separate will also generate slightly different changelog. This now is a single multi-line entry, whereas they could be punctuated (if you would use separate changesets). Using separate changesets has also other benefits - including more flexibility because separate changesets can have different "types" (major/minor/patch) and when changelog is generated those "types" are grouped together.

"Prerelease" changesets are kept when releasing prerelease versions (because when exiting the prerelease stage all of them are used to generate a single "final" changelog entry), so keeping them separate will make reviewing them easier when exiting the prerelease stage (we might want to remove no longer relevant stuff there, so they dont get listed in the final changelog entry).

packages/core/src/MachineNode.ts
Andarist: I think the code before the change had this slight bug, whereas the new version doesnt have it - so it has fixed the issue, but can potentially regress in the future if not covered by a test.

PR #954, opened 2020-01-20
Add react docs. Fixes #637

### LINKED ISSUE #637: Question: log child machine transitions
I'm new to this library so sorry if I am asking something obvious here. 

This is what I want to do:

- I want to attach a simple log function to my state transitions. So that I can follow what is happens.
- I know there is a way to do this nicely when you interpret and launch your state machine this way:
```js
const promiseService = interpret(promiseMachine).onTransition(state =>
  console.log(state.value)
);
```

But I have two complications:

1. Using react hooks:
```js
const [current, send] = useMachine(parentMachine);
```
I can just log ``current`` using  ``console.log(current);`` and I will see the state transitions. Is this correct?

2. How do I attach ``onTransition`` to child machines using invoke?
Here I have a parent machine that is invoked using react hooks. The parent machine in turn invokes child machines. How can I attach a state transition logger in the best way so that I can follow the full flow (both parent machine transitions and child machine transitions)?

```js
configuring: {
      invoke: {
        id: 'childMachine1',
        src: childMachine,
        onDone: 'render'
      }
    },
```

I explicitly do __not__ want to attach "console log" actions to each state. I just want to follow the state transitions in one place globally for a hierarchical set tree of machines.

jenny-tru:


packages/xstate-react/README.md
```diff
@@ -287,6 +287,30 @@ const App = () => {
 }
 ```
 
+## Services
```
davidkpiano: Note: this is what was added. The other additions were from copying it over (using `yarn docs:dev`)

PR #958, opened 2020-01-23
add a link to visualizer on the main page

farskid:


Andarist:
Oh, I've just noticed that this was a PR targeting `gh-pages` branch. This branch is often auto-generated and I suspect that this change will disappear once we sync master to gh-pages in the future. Could you prepare a new PR targeting master? @davidkpiano correct me if I'm wrong about the setup in this repository (I haven't checked it, just talking from experience with other repos)

farskid:
Good point. I'm not sure about the setup either but I'll open the same PR against master just in case.

PR #960, opened 2020-01-24
Machine JSON Schema definition

davidkpiano:
This PR validates the machine against a JSON schema.


packages/core/src/machine.schema.json
```diff
@@ -0,0 +1,120 @@
...
+    },
+    "actionObject": {
+      "type": "object",
+      "properties": {
+        "type": { "type": "string" },
+        "exec": {
+          "oneOf": [
+            { "type": "string" },
+            { "$ref": "#/definitions/jsonFunction" }
+          ]
```
bemayr: Thanks @davidkpiano, (not only, but especially) this really really helps when using xstate definition as a compilation target! 😃

packages/core/src/json.ts
```diff
@@ -0,0 +1,115 @@
+import { StateNode, ActionObject, Guard, InvokeDefinition } from './';
+import { mapValues, isFunction } from './utils';
+
+interface JSONFunction {
+  $function: string;
```
davidkpiano: Currently not used - will delete

packages/core/src/json.ts
```diff
@@ -0,0 +1,115 @@
+import { StateNode, ActionObject, Guard, InvokeDefinition } from './';
```
davidkpiano: This file is currently unused

davidkpiano:
@Andarist Is this good to go? We'll make refinements later but this doesn't change any behavior

packages/core/test/json.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const json = JSON.parse(JSON.stringify(machine.definition));
+
+    try {
+      validate(json);
+    } catch (err) {
+      throw new Error(JSON.stringify(JSON.parse(err.message), null, 2));
+    }
+
+    expect(validate.errors).toBeNull();
```
Andarist: does this keep the latest errors on itself? 😱 what a... quirky API 🤣

packages/core/test/json.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+      }
+    });
+
+    const json = JSON.parse(JSON.stringify(machine.definition));
+
+    try {
+      validate(json);
+    } catch (err) {
+      throw new Error(JSON.stringify(JSON.parse(err.message), null, 2));
```
Andarist: what's the purpose of this parsing back and forth? looks quite odd

packages/core/test/json.test.ts
```diff
@@ -0,0 +1,95 @@
+import { createMachine, assign } from '../src/index';
+import * as machineSchema from '../src/machine.schema.json';
+
+import * as Ajv from 'ajv';
+
+const ajv = new Ajv();
+const validate = ajv.compile(machineSchema);
+
+describe('json', () => {
+  it('should serialize the machine', () => {
```
Andarist: would be good to also add a basic "failing" test to have a confirmation that validation can actually throw

packages/core/src/types.ts
```diff
@@ -515,6 +515,7 @@ export interface StateNodeDefinition<
   id: string;
   version: string | undefined;
   key: string;
+  context: TContext;
```
Andarist: This is not quite true, but probably hard to change in v4. When merging this to v5 we should remember about moving this onto root-related types.

packages/core/src/StateNode.ts
```diff
@@ -476,9 +476,10 @@ class StateNode<
       id: this.id,
       key: this.key,
       version: this.version,
+      context: this.context!,
       type: this.type,
       initial: this.initial,
-      history: this.history,
+      history: this.history || undefined,
```
Andarist: can history be some other falsy value that `undefined` fallback is needed here?

packages/core/src/json.ts
Andarist: What problems have you encountered with this? Is the plan to use this file in the future?

Andarist:
Sorry, I've missed this being promoted to a pull request from a draft, so I've missed that I can get to reviewing it.

packages/core/test/json.test.ts
davidkpiano: Yes, its pretty bad.

packages/core/src/types.ts
davidkpiano: `.toJSON()` should only be called on the machine, anyway, where this will be defined.

packages/core/src/StateNode.ts
davidkpiano: Yes - removing this will make validation fail.

packages/core/src/json.ts
davidkpiano: Yes, I plan to use this file in the future for JSON parsing utilities and integration with other XState tools.

packages/core/src/StateNode.ts
Andarist: Right - but this is an outcome. I'm wondering about the root cause 😃 Maybe this can be avoided by handling things better "upstream"?

packages/core/src/json.ts
Andarist: I'm wondering if this will be the future way of serializing machines etc and if yes - then we could avoid adding `toJSON` to formatted transitions because we could just move its logic to a `machineToJson` or alike

davidkpiano: I still want `toJSON` to work as-is.

This file will be the basis for serializing machine definitions including non-serializable things such as functions, maps, sets, etc. and it be interpreted correctly.

Andarist: I personally would prefer to keep this out - serializing machines is an additional feature that doesn't have to belong to the "core" core, especially that most people won't really use it that way. While XState allows serialization, and this is great, it's just an added benefit - not something that is actually much needed for most runtime apps.

I understand if you feel differently and won't argue about it much. Just I prefer dedicated functions for stuff, rather than baking too much into the core.

Andarist: Especially that there is no builtin to actually rehydrate jsonified stuff in JavaScript, so this:
```js
jsonToMachine(machineToJson(machine))
```
makes the parity between those nicer than:
```js
jsonToMachine(JSON.stringify(machine))
```

davidkpiano: The assumption is that developers using `JSON.stringify/parse` should have already "stringified" their implementation details. I want to make sure that that behavior works.

In fact, `machineToJson` etc. should _not_ be the common use-case, because it will essentially be `eval()`-ing code such as functions defined as strings, which can be risky.

davidkpiano: That is, given the developer already has the implementation details of actions, service, etc., it should be possible to `JSON.parse(someMachineString)` and apply it as such:

```js
const machineConfig = JSON.parse(someMachineString);

// Implementation details are not part of the machine config!
const machine = createMachine(machineConfig, {
  actions: /* ... */,
  services: /* ... */,
  guards: /* ... */,
  // ... etc.
});
```

This is the main use-case I want to support.

Andarist: Right, this is IMHO OK - this is orthogonal to `JSON.stringify` though. Do you suspect having to call smth like `stringify(machineConfig)` over `JSON.stringify` would deteriorate DX that much? Stringifying seems like a use case for rather advanced users and as such, I wouldn't expect it to be a problem.

I know this is somewhat in opposite to what I have raised previously about not having the parity between serialize & deserialize (where `JSON.parse` serves as deserializing algorithm here), but I wasn't sure how exactly you have envisioned this to be used. 

I still think it, in general sense, would be great to have the parity between those 2. So maybe instead of adding custom `toJSON` to some internal structure, we should think of how to just represent the whole definition in a serializable way. We can split the serializable definition from the other data about the transition that we need at runtime (or maybe we wouldn't need splitting it - just a different representation).

If you don't want to introduce a separate stringifier then this is OK to be merged. We can try to pursue different definition representation in other PRs.

Also - `Object.defineProperty` is not the fastest thing in the world. Or at least it hasn't been in the past: https://humanwhocodes.com/blog/2015/11/performance-implication-object-defineproperty/ . Browsers perf characteristics are a moving target and it would have to be checked if this is still the case, but if this is uncertain - it's better to avoid it altogether (if possible ofc).

packages/core/src/StateNode.ts
```diff
@@ -1888,6 +1889,18 @@ class StateNode<
...
       eventType: transitionConfig.event
     };
+
+    Object.defineProperty(transition, 'toJSON', {
+      value: () => ({
+        ...transition,
+        target: transition.target
+          ? transition.target.map(t => `#${t.id}`)
+          : undefined,
+        source: `#{this.id}`
```
Andarist: isn't this redundant? It shouldn't be needed for rehydration as this always points to the node on which transitions are defined and this cannot be changed or manipulated anyhow.

packages/core/test/json.test.ts
davidkpiano: Making the error message readable.

packages/core/src/json.ts
aboutlo: > ```js
> const machineConfig = JSON.parse(someMachineString);
> 
> // Implementation details are not part of the machine config!
> const machine = createMachine(machineConfig, {
>   actions: /* ... */,
>   services: /* ... */,
>   guards: /* ... */,
>   // ... etc.
> });
> ```
> 
> This is the main use-case I want to support.

Just as the feedback this exactly one I'm currently doing it. 
I have a `createMyMachine` which given some business objects will return an xState plain `machineConfig` that can be used later. 
I also have the serialization scenario so as you said I need to handle `"stringified" their implementation details`

So having the JSON schema plus some utils to deal with that is pretty cool 🙏

packages/core/src/machine.schema.json
```diff
@@ -0,0 +1,316 @@
...
+        "src": {
+          "type": "string"
+        },
+        "autoForward": {
+          "type": "boolean",
+          "default": false
+        }
+      },
+      "required": ["type", "id", "src"],
+      "additionalProperties": false
```
aboutlo: I'm trying to test this JSON schema against my implementation. 
I got a validation issue because my `invoke`  contains `onDone` and `onError` as additional properties. 

Error:

```JS
{ keyword: 'additionalProperties',
        dataPath: '.states[\'f941522f-a625-4953-9472-7d3c4004d1d4\'].invoke[0]',
        schemaPath: '#/definitions/invokeObject/additionalProperties',
        params: { additionalProperty: 'onError' },
        message: 'should NOT have additional properties' },
```

Object:
```js
states: {
      [myId]: {
        id: 'doSomething',
        key: myId,
        type: 'atomic',
        invoke: [
          {
            src: 'doSomething',
            onDone: `${myId}_EXECUTED`, 
            onError: `${myId}_FAILED`,
          },
        ],
      }
```
https://xstate.js.org/docs/guides/communication.html#the-invoke-property 
Should you add `onDone` and `onError` in the schema as well or I'm missing something?

Andarist: I'm not sure - but maybe you test the wrong thing? onError & onDone are part of the config but they are converted to transitions internally and thus probably available on the definition (and not the config)

aboutlo: I see it makes sense. Indeed, I'm creating a config so I can't test against this JSON schema. I will wait for this feature to get merge then

packages/core/src/StateNode.ts
davidkpiano: It's necessary when aggregating transitions, similar to how `event` is also defined.

Andarist: Could you illustrate this with an example?

.changeset/tall-beans-suffer.md
```diff
@@ -0,0 +1,7 @@
+---
+'xstate': minor
+---
+
+The machine can now be safely JSON-serialized, using `JSON.stringify(machine)`. The shape of this serialization is defined in `machine.schema.json` and reflected in `machine.definition`.
```
Andarist: I think you have meant that `machine.defiinition` can be safely serialized - as that is what you are doing in tests:
https://github.com/davidkpiano/xstate/blob/55aa589648a9afbd153e8b8e74cbf2e0ebf573fb/packages/core/test/json.test.ts#L85

I would be in favor of keeping `machine.definition` serializable, rather than the machine itself. IMHO makes it more explicit what actually gets serialized, especially that toJSON put on a Machine would just proxy to serializing `.definition`

PR #961, opened 2020-01-25
Prepare `@xstate/react` release

Andarist:


packages/xstate-react/src/fsm.ts
```diff
@@ -28,7 +28,7 @@ export function useMachine<
     if (stateMachine !== initialMachine) {
       throw new Error(
         'Machine given to `useMachine` has changed between renders. This is not supported and might lead to unexpected results.\n' +
-          'Please make sure that you pass the same Machine as argument each time.'
+          "Please make sure that you pass the same Machine as argument each time. Keep also that in mind that `useMemo` can't be used for this kind of caching as per https://reactjs.org/docs/hooks-reference.html#usememo"
```
tamebadger: I'd recommend wording this as "Also keep in mind that `useMemo` can't be used for this kind of caching as per https://reactjs.org/docs/hooks-reference.html#usememo""

packages/xstate-react/src/index.ts
```diff
@@ -45,7 +45,7 @@ export function useMachine<
     if (machine !== initialMachine) {
       throw new Error(
         'Machine given to `useMachine` has changed between renders. This is not supported and might lead to unexpected results.\n' +
-          'Please make sure that you pass the same Machine as argument each time.'
+          "Please make sure that you pass the same Machine as argument each time. Keep also that in mind that `useMemo` can't be used for this kind of caching as per https://reactjs.org/docs/hooks-reference.html#usememo"
```
tamebadger: same comment as above 🤓

packages/xstate-react/src/fsm.ts
Andarist: lol, yea - this is some true past midnight writing 😅 my version is totally incorrect

davidkpiano:
@Andarist Is this still needed?

PR #962, opened 2020-01-26
Align implementation of @xstate/vue more closely to what is being implemented for @xstate/react

Andarist:
cc @AleksejDix @studnitz

### DIFF
--- a/packages/xstate-react/README.md
+++ b/packages/xstate-react/README.md
@@ -28,11 +28,11 @@ const toggleMachine = Machine({
 });
 
 export const Toggler = () => {
-  const [current, send] = useMachine(toggleMachine);
+  const [state, send] = useMachine(toggleMachine);
 
   return (
     <button onClick={() => send('TOGGLE')}>
-      {current.value === 'inactive'
+      {state.value === 'inactive'
         ? 'Click to activate'
         : 'Active! Click to deactivate'}
     </button>
@@ -51,9 +51,9 @@ A [React hook](https://reactjs.org/hooks) that interprets the given `machine` an
 - `machine` - An [XState machine](https://xstate.js.org/docs/guides/machines.html).
 - `options` (optional) - [Interpreter options](https://xstate.js.org/docs/guides/interpretation.html#options) OR one of the following Machine Config options: `guards`, `actions`, `activities`, `services`, `delays`, `immediate`, `context`, or `state`.
 
-**Returns** a tuple of `[current, send, service]`:
+**Returns** a tuple of `[state, send, service]`:
 
-- `current` - Represents the current state of the machine as an XState `State` object.
+- `state` - Represents the current state of the machine as an XState `State` object.
 - `send` - A function that sends events to the running service.
 - `service` - The created service.
 
@@ -65,9 +65,9 @@ A [React hook](https://reactjs.org/hooks) that subscribes to state changes from 
 
 - `service` - An [XState service](https://xstate.js.org/docs/guides/communication.html).
 
-**Returns** a tuple of `[current, send]`:
+**Returns** a tuple of `[state, send]`:
 
-- `current` - Represents the current state of the service as an XState `State` object.
+- `state` - Represents the current state of the service as an XState `State` object.
 - `send` - A function that sends events to the running service.
 
 ### `useMachine(machine)` with `@xstate/fsm` <Badge text="1.1+"/>
@@ -81,9 +81,9 @@ This special `useMachine` hook is imported from `@xstate/react/lib/fsm`
 - `machine` - An [XState finite state machine (FSM)](https://xstate.js.org/docs/packages/xstate-fsm/).
 - `options` - An optional `options` object.
 
-**Returns** a tuple of `[current, send, service]`:
+**Returns** a tuple of `[state, send, service]`:
 
-- `current` - Represents the current state of the machine as an `@xstate/fsm` `StateMachine.State` object.
+- `state` - Represents the current state of the machine as an `@xstate/fsm` `StateMachine.State` object.
 - `send` - A function that sends events to the running service.
 - `service` - The created `@xstate/fsm` service.
 
@@ -121,7 +121,7 @@ const fetchMachine = createMachine({
 });
 
 const Fetcher = ({ onFetch = () => new Promise(res => res('some data')) }) => {
-  const [current, send] = useMachine(fetchMachine, {
+  const [state, send] = useMachine(fetchMachine, {
     actions: {
       load: () => {
         onFetch().then(res => {
@@ -131,7 +131,7 @@ const Fetcher = ({ onFetch = () => new Promise(res => res('some data')) }) => {
     }
   });
 
-  switch (current.value) {
+  switch (state.value) {
     case 'idle':
       return <button onClick={_ => send('FETCH')}>Fetch</button>;
     case 'loading':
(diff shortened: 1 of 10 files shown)


packages/xstate-react/README.md
```diff
@@ -51,9 +51,9 @@ A [React hook](https://reactjs.org/hooks) that interprets the given `machine` an
 - `machine` - An [XState machine](https://xstate.js.org/docs/guides/machines.html).
 - `options` (optional) - [Interpreter options](https://xstate.js.org/docs/guides/interpretation.html#options) OR one of the following Machine Config options: `guards`, `actions`, `activities`, `services`, `delays`, `immediate`, `context`, or `state`.
 
-**Returns** a tuple of `[current, send, service]`:
+**Returns** a tuple of `[state, send, service]`:
 
-- `current` - Represents the current state of the machine as an XState `State` object.
+- `state` - Represents the current state of the machine as an XState `State` object.
```
Andarist: `state` name seems to be a little bit overloaded, might mean different things - currently it is an instance of `State` though, so `state` seems to be an appropriate name for this, but is also a little bit ambiguous, can we think of something better? maybe `State` should be renamed to something else in the next major?

packages/xstate-vue/README.md
```diff
@@ -61,9 +61,9 @@ A [Vue composition function](https://vue-composition-api-rfc.netlify.com/) that
 - `machine` - An [XState machine](https://xstate.js.org/docs/guides/machines.html).
 - `options` (optional) - [Interpreter options](https://xstate.js.org/docs/guides/interpretation.html#options) OR one of the following Machine Config options: `guards`, `actions`, `activities`, `services`, `delays`, `immediate`, `context`, or `state`.
 
-**Returns** `{ current, send, service}`:
+**Returns** `{ state, send, service}`:
```
Andarist: in `@xstate/react` we return a tuple instead of an object, was a decision about returning an object here based on Vue conventions or maybe was it done by accident? should we change this to a tuple?

packages/xstate-vue/src/fsm.ts
```diff
@@ -23,5 +37,55 @@ export function useMachine<
...
+      if (currentState.changed) {
+        state.value = currentState;
+      }
+    });
+    onCleanup(() => unsubscribe());
+  });
+
+  const send = (event: TEvent | TEvent['type']) => serviceRef.value.send(event);
+
+  return { state, send, service: serviceRef };
```
Andarist: it's a little bit weird that this return type is different from the one in `useMachine`, in here a `serviceRef` is returned and in `useMachine` it's just a `service`. It makes semantic sense, because `useMachine` can never change returned service, but this one can. It looks slightly inconsistent though, so maybe this one should return `serviceRef`? that would be inconsistent with `state` though (as it's ref as well) 

Changing the return type to a tuple would make the issue go away 😅

packages/xstate-vue/src/index.ts
```diff
@@ -1,2 +1,106 @@
...
+    ...context
+  } as TContext);
+
+  const service = interpret(createdMachine, interpreterOptions).start(
+    rehydratedState ? State.create(rehydratedState) : undefined
+  );
+
+  const state = ref<State<TContext, TEvent>>(service.state);
+
+  service.onTransition(currentState => {
```
Andarist: I still have no idea if this is OK in Vue. Can we set up a subscription in this setup function? Or should this be deferred until some lifecycle hook, like `onBeforeMount`? Do we know some Vue expert that could answer this with confidence?

packages/xstate-vue/test/UseMachine.vue
```diff
@@ -50,22 +50,13 @@ export default {
...
       new Promise(res => setTimeout(() => res('some data'), 50));
 
-    const { current, send, service } = useMachine(fetchMachine, {
+    const { state, send, service } = useMachine(fetchMachine, {
       services: {
         fetchData: onFetch
       },
       state: persistedState
     });
-    watch(() =>
```
Andarist: i think this one was unused, `load` is not referenced anywhere as far as I can tell

packages/xstate-vue/test/UseService.vue
```diff
@@ -1,26 +1,34 @@
...
+} from 'xstate';
 import { watch, ref } from '@vue/composition-api';
 
 export default {
   props: ['service'],
   setup(props): { service: Service } {
-    let { current, send } = useService(props.service);
+    let { state, send } = useService(props.service);
 
     watch(() => {
```
Andarist: I think this is completely not needed (but not sure, would love to get some validation from somebody), this is already handled inside `useService`, isn't it? is using `useService` from withitn `watch` even allowed in Vue?

packages/xstate-vue/README.md
studnitz: This is based on Vue composition API, where you always return an object, so I think no, we shouldn't change that to a tuple.

packages/xstate-vue/src/fsm.ts
studnitz: Yeah, I also spent some time thinking about how to name it. The thing is, that with the composition API you are expected to get `Ref`s, when destructuring the object (e.g. `const { foo, bar } = useBaz()`) and using them in the vue `<template>` part will also automatically unwrap them.

So I think just leaving it as `service` is more beneficial than returning `serviceRef`.

By the way, the vue-apollo (which is maintained by a vue core member) project also does it like that in their [composition API implementation.](https://github.com/vuejs/vue-apollo/blob/v4/packages/vue-apollo-composable/src/useQuery.ts#L451-L453)

packages/xstate-react/README.md
davidkpiano: I don't think it's ambiguous - it's an object that represents the overall state.

packages/xstate-react/src/fsm.ts
```diff
@@ -9,6 +9,22 @@ import {
 import { useSubscription, Subscription } from 'use-subscription';
 import useConstant from './useConstant';
 
+const getServiceState = <
+  TContext extends object,
+  TEvent extends EventObject = EventObject,
+  TState extends Typestate<TContext> = any
+>(
```
davidkpiano: Maybe we should "expose" `service.state` just like in XState core?

Andarist: Maybe, this can be easily added and would make this use case easier, just didn't want to add this on my own in the spirit of keeping `@xstate/fsm` minimal.

packages/xstate-react/README.md
Andarist: Maybe I was overthinking this, I think you are right - the name is OK. One thing that I would love exploring in the future (it's a minor pet peeve of mine) is removing `event` & `actions` from this object. I don't think those are much needed for it as they are really temporary. Removing them would allow us to only create new States when something actually changes (so it would improve perf slightly). Either way - this definitely is just an idea to explore for me right now, not something that I want to jump into in the nearest future.

davidkpiano: Those are necessary for the interpreter to properly decide what to do next - actions are necessary so the interpreter knows what to execute, and the event is necessary for action implementations that require both the `context` and `event`. I depend on them both in multiple projects.

Andarist: I understand why both of those are there, but maybe some alternative API for returning those could be explored. I hadn't given this that much thought yet, it's just something that I have noticed as potential improvement, but definitely not without exploring alternatives first. Might even in the end come to conclusion that it's fine as is

davidkpiano: Re: alternative API, it's important (IMO) to consider all the use-cases for this. For example, if this is a state you receive as a JSON payload in response to a GET/POST request to a remote statechart service, you ideally want all that information up-front in that JSON object.

PR #967, opened 2020-01-29
Merge pull request #967 from andrewgordstewart/andrewgordstewart/add-tcontext-type-to-invoke-config

andrewgordstewart:
With this change, the `invoke.src` property of a `StateNodeConfig` can perform type inference:

```
type UserContext = { foo: 'bar' };
type UserEvent = { type: 'READY' | 'SET' | 'GO' };
export const stateNode: StateNodeConfig<UserContext, any, UserEvent> = {
  invoke: {
    src: (ctx, event) => {
      switch (event.type) {
        case 'READY':
          // Property 'Foo' does not exist on type 'UserContext'. Did you mean 'foo'?ts(2551)
          return Promise.resolve(ctx.Foo);
        case 'SET':
          return Promise.resolve('SET');
        /// Type '"Go"' is not comparable to type '"READY" | "SET" | "GO"'.ts(2678)
        case 'Go':
          return Promise.resolve();
      }
    }
  }
};
```

See discussion starting [here](https://gitter.im/statecharts/statecharts?at=5e319ef63aca1e4c5f4218b3)

### DIFF
--- a/.changeset/large-elephants-smoke.md
+++ b/.changeset/large-elephants-smoke.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Add context & event types to InvokeConfig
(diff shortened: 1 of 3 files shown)


davidkpiano:
Thanks @andrewgordstewart! Can you create a quick changeset? Just run `yarn changeset` in the root `xstate` directory (this will be a `patch` change).

PR #974, opened 2020-02-02
docs: add version badge to respond section

RafaelKr:
makes it clear that responses are only available in version 4.7+


Andarist:
Thanks!

PR #976, opened 2020-02-03
Fixed issue with `useMachine` crashing without providing optional `options` parameter

### LINKED ISSUE #975: [xstate/vue] useMachine(machine) with no second parameter causes errors
**Description**

useMachine requires a second parameter to be at least an empty object, while documentation says it is optional

**Expected Result**

`useMachine(toggleMachine)` to work as documented

**Actual Result**

`useMachine` throws a type error when it tries to spread the second parameter when it's undefined.

**Reproduction**

https://codesandbox.io/s/xsharp-vue-usemachine-second-param-f5bpk?expanddevtools=1&fontsize=14&hidenavigation=1&module=%2Fsrc%2FApp.vue

Specifically App.vue:21 breaks it. Line 23 is the work around.

**Additional context**

Additionally, the documentation here: 
> https://xstate.js.org/docs/packages/xstate-vue/
does not match the documentation here: 
> https://github.com/davidkpiano/xstate/tree/master/packages/xstate-vue
specifically, the former link references `current` while the package and its documentation have moved to using `state`

```
 "dependencies": {
    "@vue/cli-plugin-babel": "4.1.1",
    "@vue/composition-api": "0.3.4",
    "@xstate/vue": "0.1.0",
    "vue": "^2.6.11",
    "xstate": "4.7.7"
  },
```

Andarist:
fixes #975

### DIFF
--- a/.changeset/beige-plants-share.md
+++ b/.changeset/beige-plants-share.md
@@ -1,0 +1,5 @@
+---
+'@xstate/vue': patch
+---
+
+Fixed issue with `useMachine` crashing without providing optional `options` parameter.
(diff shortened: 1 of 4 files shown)


Glutnix:
Awesome :)

PR #984, opened 2020-02-08
Implemented the EventsLog tab (among other things)

amitnovick:
Feel free to raise questions / report issues with anything included here :-)


packages/xstate-dev/panel/package.json
```diff
@@ -7,17 +7,29 @@
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
     "@xstate/react": "^0.8.1",
+    "date-fns": "^2.9.0",
+    "hex-rgba": "^1.0.2",
+    "immutable": "^4.0.0-rc.12",
```
davidkpiano: What is `immutable` being used for? Can we please avoid it?

packages/xstate-dev/panel/src/utils/getItemString.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { Iterable } from 'immutable';
```
davidkpiano: I don't think we should support `immutable`.

davidkpiano:
It looks good to me, except for `immutable` support. We should absolutely not support `immutable`.

packages/xstate-dev/panel/src/utils/getItemString.js
amitnovick: Agree!
I am not sure what it's being used for (extracted it from elsewhere). 
Will try to remove it and update the PR

Issue #672, opened 2019-09-19
4.7.0 Actor/Service Feedback


derek-duncan:
## Bug or feature request?
Bug

## Description:
I've been playing more with xstate@next. It seems that both conceptually and in Typescript, the Actor/Service (Interpreter) distinctions are a little blurred, so I'm trying to understand when to use either. My primary question is what are the different purposes for Actors and Services/Interpreters?

Here are some more specific notes for `v4.7.0-rc3`:

It looks like `invoke` creates an Actor behind the scenes. Accessing it through `current.children.invokeId`, you get an Actor back. Accessing it through `service.children.get(invokeId)`, you get an Interpreter. Shouldn’t `invoke` always generate services/interpreters and not an actor? If invoking a “service” is really just an Actor behind the scenes, it’s be nice to always return an Actor so you consistently know what to expect from your `current.children`, `service.children` and `spawn`.

`spawn` also returns an Interpreter type in TS, but it looks like the actual object is an Actor. Why should `spawn` ever return an Interpreter if it’s an utility for spawning actors? Looks like the TS changed in #664

Also, a minor TS bug for `spawn`. It doesn’t accept a machine with custom event types since they clash with the generic EventObject.
```
spawn(new Machine<Ctx, Evt>())
// Type 'EventObject' is not assignable to type ‘Evt’
```

Finally, most of these concepts trickle out as confusing when you start to use the `useService` and `useActor` hooks. It seems that Actors are more locked down in state access, but it’s unclear as to why? It’s much easier to just use a `useService`, but as already noted, most of the Actors coming out of `current.children` and `spawn` aren’t accepted in `useService`. It’s unclear when comparing services to actors on why you’d ever want an Actor vs an Interpreter, or vice versa.

## (Bug) Potential fix:

I need more understanding of the conceptual model to propose a fix, but I'd be more than happy to after I understand distinctions between Actors and Services.

coodoo:
I had those doubts earlier and after a quick digging up I came up with these notes, hope that helps.

- Main difference between `services` and `actors` - `services` need to be statically written in the machine definition to be `invoked`, whereas actors can be dynamically `spawned` by any action, multiple times (if you so wish).

- actors could spawn `sub-actors` and communicate with it's `parent` or other actors via event sending

- In short, `actors` are like `services` on steroid, allowing developers to more flexibly model complex application logics with multi-level of machine hierarchy interacting with each other.

[...]

Issue #424, opened 2019-04-12
[Feature] Expose Invoked Services


davidkpiano:
## Bug or feature request?
Feature

## Description:
When services are invoked in a parent service, their state is currently kept hidden to their machine. This is by design, we can traverse this invocation tree and "query" the state of each invoked service by subscribing.

## (Feature) Potential implementation:

Consider a parent `todosMachine` like this:

```js
const todosMachine = Machine({
  id: 'todos',
  initial: 'active',
  states: {
    active: {
      invoke: { src: 'todosService', id: 'todos' },
      on: { ADD_TODO, DELETE_TODO, UPDATE_TODO }
    },
    inactive: {}
  }
});
```

Then, assuming `todosService` is implemented as a `Machine`, we can subscribe to its state the same way you would do with any other service:

```js
const todosService = interpret(todosMachine)
  .onInvoke(childService => {
    childService.onTransition(state => {
      // current state of child service
    });
  })
  .start();
```

You can imagine that this would be useful in a React app:

```jsx
function TodosApp() {
  const [current, send, service] = useMachine(todosMachine);
  
  // listen to state from the todosService, whenever it's invoked
  // undefined if service is not invoked or stopped
  const [currentTodos] = useChildService(service, 'todos');

  const todos = currentTodos ? currentTodos.context.todos : [];

  return current.matches('active') ? (
    <ul>
      {todos.map(todo => <li key={todo.id}>{todo.message}</li>}
    </ul>
  ) : null;
}
```

The new APIs for `xstate` (core) would be:

- `service.onInvoke(childService => { ... })`
- `service.children` - already exists, but mapping would provide the actual services instead of thin interfaces. Only to be used for debugging.

The new APIs for `@xstate/react` would be:

- `useService(service)` - same as `useMachine()` but with a running service
- `useChildService(service, id)` - same as `useMachine()` but with a running child service whenever the child is invoked
    
**This is tentative** and I'm still researching to see how this works with the Actor model. Would love feedback!

LunarLanding:
Coming from [this conversation](https://spectrum.chat/statecharts/general/dynamical-number-of-orthogonal-substates~c421e22e-2ce3-474e-a89d-c223b15286c4), I see that this RFC helps with the issue of getting the state of invoked services out of the parent machine so that external behavior might be derived from them (such as rendering todos in the example above).
I had assumed till now that that was already possible.
However is does not address the main problem, which is managing a dynamic number of invocations,i.e. creating on demand.

[...]

PR #985, opened 2020-02-08
V5: Revert state.children mapping, represent spawned actors

davidkpiano:
This PR is a significant refactor to:

- Expose all invoked services and spawned actors on `state.children` by their ID (#424 #672)
- Introduce concept of "deferred actors", which are "placeholder" actors that can be used as actor references in machines that are replaced (in-place) with real actor implementations in invoked machines
- Remove all traces of "activities" (except for `spawnActivity`)


packages/core/src/interpreter.ts
```diff
@@ -809,73 +802,68 @@ export class Interpreter<
...
-
-          const childIndex = this.state.children.findIndex(
-            child => child.id === id
-          );
+        const actor = actorFrom(spawnable, id, this as Actor);
 
-          this.state.children[childIndex] = actor;
+        // Mutate the deferred actor reference
+        if (action.actor) {
+          Object.assign(action.actor, actor);
```
Andarist: for what part of the codebase or the consuming code this is relevant? mutating the action seems like a strong anti-pattern, so I'm wondering if we could think of a different design to avoid this. I don't understand yet the reasons why this is done though

packages/core/src/interpreter.ts
```diff
@@ -809,73 +802,68 @@ export class Interpreter<
...
+          this.forwardTo.add(id);
         }
 
+        this.children.set(id, actor);
+
+        this.state.children[id] = actor;
+
+        this.state.children[id].meta = {
+          ...this.state.children[id].meta,
+          ...actorDef
```
Andarist: is this definition useful anyhow for consumers?  the definition consists solely of `{id,src}`, consumers already have to know the ID to even access this object - so the question is, how `src` can be utilized?

packages/core/src/invoke.ts
```diff
@@ -1,163 +1,166 @@
...
+        }
+      }
+    }
+  );
 
-    const actor = childService;
+  const actor = {
+    id,
+    send: () => void 0,
+    subscribe: (next, handleError, complete) => {
```
Andarist: `subscribe` from all actor types can be dropped, because you have removed this from the `Actor` interface. Because those "converters" only return Actor type those subscribe methods are "lost" anyway - the calling/consuming code cant use them because its not defined on the Actor interface

packages/core/src/stateUtils.ts
```diff
@@ -931,15 +931,15 @@ export function getActions<TContext, TEvent extends EventObject>(
     flatten(
       Array.from(entryStates).map(entryNode => {
         return [
-          ...getActivities(entryNode).map(activity => start(activity)),
+          ...getInvocations(entryNode).map(invokeDef => start(invokeDef)),
```
Andarist: it would be better to convert those early `entryNode.invoke` could already be invoke-definitions to avoid converting them each time in `getActions`

packages/core/src/stateUtils.ts
```diff
@@ -1184,17 +1202,20 @@ export function resolveTransition<
...
-      children = children.filter(childActor => {
-        return (
-          childActor.id !==
-          (action as ActivityActionObject<TContext, TEvent>).actor.id
-        );
-      });
+      const id = (action as InvokeActionObject<TContext, TEvent>).def.id;
+
+      children = { ...children };
+      delete children[id];
```
Andarist: I wonder if we really need to represent them at `state.children` right away as "null actors". Why exactly this is important? It would be better to get back to this problem~ once we get to making the transition a microstep, I have a feeling that it might impact this thing.

packages/core/src/types.ts
```diff
@@ -405,7 +426,10 @@ export interface StateNodeConfig<
    * The services to invoke upon entering this state node. These services will be stopped upon exiting this state node.
    */
   invoke?: SingleOrArray<
-    string | InvokeCreator<TContext, TEvent> | InvokeConfig<TContext, TEvent>
+    | string
+    | ActorCreator<TContext, TEvent>
```
Andarist: I must say I'm saddened by this - I think the design with "just actors" and dedicated "converters" was superior. It had nice decoupling, composition and tree-shaking characteristics. For instance while machine & promise actors might be considered really common, observables and callbacks are much less common - and right now every consumer is "paying" the cost of built support for those "integrations". 

Having all of those types decoupled from the core was a really good way of showing people that you actually can integrate easily **any** data/source type with XState, by simply writing a small adapter - in the very same way as "core" adapters were written.

packages/core/src/interpreter.ts
davidkpiano: The reason for this is that at first, the actor is "deferred" - it's just a null actor. You can think of this as the result of:

```js
machine.transition('STATE', 'EVENT');
// => resulting state spawned an actor!
```

Obviously, we don't want an actor to actually be spawned (yet), so it's "deferred" until the interpreter decides to make it alive.

This takes inspiration from Akka:

> This particular main Actor is created using Actor.deferred, which is like a factory for a behavior. Creation of the behavior instance is deferred until the actor is started, as opposed to Actor.immutable that creates the behavior instance immediately before the actor is running.

https://doc.akka.io/docs/akka/2.5.5/scala/typed.html

And also it solves the problem of spawning actors in an initial state, since you can now do that without fear of the actor unintentionally being started before the service is started.

packages/core/src/interpreter.ts
davidkpiano: Rehydration

packages/core/src/types.ts
davidkpiano: Hah, I actually made this change because you had concerns that there would be bigger breaking changes for the majority of users 😅 

Regardless, this was added back because `spawnFrom` is needed for inline `spawn()` (although this can also be done the same way, with the extra function call...)

Let's revisit this in a future PR

packages/core/src/interpreter.ts
Andarist: Right - I understand that we want `machine.transition` to be pure and thus we cannot actually spawn an actor within it. I'm more interested in why exactly we want to represent them immediately in the returned State value over just returning "pure" actions and having Interpreter to actually execute them (and thus spawn actors).

packages/core/src/types.ts
Andarist: > Hah, I actually made this change because you had concerns that there would be bigger breaking changes for the majority of users 😅

Well, I wasn't arguing against breaking changes per se - rather that we probably shouldn't advertise v5 as "no breaking changes" because that was misleading.

> Let's revisit this in a future PR

K, would be good to discuss this later on Slack to get a final decision on this, but I agree that it doesn't have to hold off this PR.

Keep in mind that we should be able to automate migration in many cases - it should be quite possible to provide an ESLint rule that would serve as a migration codemod.

packages/core/src/interpreter.ts
Andarist: Could we add a rehydration test that would use it? Would probably be good to include something like this also in the docs.

packages/core/src/interpreter.ts
davidkpiano: Because they need to be present in the `context`.

**State-less machine:**
```js
machine.transition(someState, someEvent);
// State {
//   ...
//   context: { actor: deferredActor }
// }
```

**Interpreted machine:**
```js
service.onTransition(/* ... */)
// State {
//   ...
//   context: { actor: liveActor }
// }
```

**Persisted state:**
```js
// The information is used to restore the actor
JSON.stringify(state);
// State {
//   ...
//   context: { actor: { id, meta } }
// }
```

We do not want `actor: undefined` or `actor: null` because:

1. it's not helpful for testing
2. it loses ability to be rehydrated
3. it can't be linked to in `state.children`

Actually, that point, linking an actor to `state.children` is the most important point - it's sub-optimal (if not very difficult/impossible) to do so if the `actor` is null first.

We want the state from `machine.transition(...)` to closely reflect `service.onTransition(...)` as much as possible.

packages/core/src/types.ts
davidkpiano: One happy medium would be:

```js
// ...
invoke: spawnFrom(promise /* or observable, callback, etc. */)
// ...
```

How does that sound?

Andarist: Yeah, I think this is a good idea - we should be able to provide fine-grained spawn creators (`spawnPromise`, `spawnObservable`, etc) but they could be wrapped for convenience within a `spawnFrom` helper.

All should be importable (and not injected) to aid tree-shaking.

PR #987, opened 2020-02-09
V5: History refactor

davidkpiano:
This PR refactors the state history resolution algorithm to match the SCXML algorithm.

- `state.historyValue` eliminated, replaced with `state.historyMap`
- A lot of code removed! 🎉


packages/core/src/stateUtils.ts
```diff
@@ -856,35 +812,8 @@ export function getActions<TContext, TEvent extends EventObject>(
   stateNode: StateNode<TContext, any, TEvent>,
   transition: StateTransition<TContext, TEvent>,
   currentContext: TContext,
-  _event: SCXML.Event<TEvent>,
-  prevState?: State<TContext>
+  _event: SCXML.Event<TEvent>
 ): Array<ActionObject<TContext, TEvent>> {
-  const prevConfig = getConfiguration(
```
davidkpiano: All of this code was moved, because it didn't belong in this function (mutated `transition`)

packages/core/src/stateUtils.ts
```diff
@@ -1102,21 +1048,64 @@ export function resolveTransition<
...
-    currentContext,
-    _event,
-    currentState
+
+  // Entry and exit set
+  const prevConfig = getConfiguration(
+    [],
+    currentState ? getStateNodes(machine, currentState.value) : [machine]
   );
+  const resolvedConfig = stateTransition.configuration.length
```
davidkpiano: This is the code moved from the actions function above. Copy-paste.

packages/core/test/history.test.ts
```diff
@@ -232,66 +232,64 @@ describe('parallel history states', () => {
...
-  describe('history', () => {
-    // on.first -> on.second.A
-    const stateABKL = historyMachine.transition(
-      historyMachine.initialState,
-      'SWITCH'
-    );
-    // INNER_A twice
-    const stateACDKL = historyMachine.transition(stateABKL, 'INNER_A');
-    const stateACEKL = historyMachine.transition(stateACDKL, 'INNER_A');
+  // on.first -> on.second.A
```
davidkpiano: None of these tests were changed; they were just unnecessarily nested in a `describe` block, so I flattened them.

packages/core/src/stateUtils.ts
```diff
@@ -1102,21 +1048,64 @@ export function resolveTransition<
...
+    stateTransition.entrySet.push(machine);
+  }
+  stateTransition.exitSet.sort((a, b) => b.order - a.order);
+  stateTransition.entrySet.sort((a, b) => a.order - b.order);
+
+  // History
+  const historyMap: Record<
+    string,
+    Array<StateNode<TContext, any, TEvent>>
+  > = currentState ? currentState.historyMap : {};
```
Andarist: This map is being mutated, I don't particularly see it as a con per se - but it would be great to be consistent or at least to have some rules (documented ones) about which values are treated as immutable ones and which are mutable. This seems like an important piece of information for consumers.

packages/core/src/State.ts
```diff
@@ -55,8 +54,11 @@ export class State<
 > {
   public value: StateValue;
   public context: TContext;
-  public historyValue?: HistoryValue | undefined;
   public history?: State<TContext, TEvent, TStateSchema>;
+  public historyMap: Record<
```
Andarist: This is currently only set inside `resolveTransition`, but previously it was possible to rehydrate `historyValue` through, at least, both `State.from` and `State.create` calls. This seems like an important missing piece for successful state rehydration.

packages/core/src/stateUtils.ts
```diff
@@ -505,55 +496,20 @@ function resolveHistory<TContext, TEvent extends EventObject>(
...
-  }
-  return flatten(
-    toStatePaths(subHistoryValue!).map(subStatePath => {
-      return stateNode.history === 'deep'
-        ? getFromRelativePath(parent, subStatePath)
-        : [parent.states[subStatePath[0]]];
-    })
-  );
+
+  return state.historyMap[stateNode.id];
```
Andarist: I don't immediately see why this matters here as `state.historyMap` seems to always be checked beforehand, but running tests without has confirmed that this is very much needed. Do you maybe know, from the top of your head, why exactly this is needed?

packages/core/src/StateNode.ts
```diff
@@ -354,6 +354,17 @@ export class StateNode<
     state: State<TContext, TEvent>,
     _event: SCXML.Event<TEvent>
   ): StateTransition<TContext, TEvent> | undefined {
+    if (this.type === 'history' && state.historyMap[this.id]) {
```
Andarist: I'm not entirely sure yet, but this might introduce a regression - entering remembered history states is still entering (when we look at SCXML they are just added to `entrySet` in `computeEntrySet` & later on are treated as any other states in `enterStates` which executes executable content). And judging by this returned value & the algorithm of `getActions` I believe those entry actions wouldnt get executed.

As there were no tests for this behavior I think it's OK to let it slip for now, we should just be aware of that and fix it soon - I'm touching the whole transition resolving algorithm in https://github.com/davidkpiano/xstate/pull/825 and I think I might be able to fix this as well as part of it, or if not - I will make a new issue about it then so we can handle it in a separate PR.

packages/core/src/stateUtils.ts
Andarist: A really nice change 👍

packages/core/src/stateUtils.ts
```diff
@@ -1082,6 +1008,26 @@ export function resolveRaisedTransition<
   return state;
 }
 
+function getHistoryNodes<TContext, TEvent extends EventObject>(
+  stateNode: StateNode<TContext, any, TEvent>
+): Array<StateNode<TContext, any, TEvent>> {
+  return getChildren(stateNode).filter(sn => {
```
Andarist: this could potentially reuse introduced in this PR `isHistoryNode`

packages/core/src/stateUtils.ts
```diff
@@ -1102,21 +1048,64 @@ export function resolveTransition<
...
+  stateTransition.entrySet.sort((a, b) => a.order - b.order);
+
+  // History
+  const historyMap: Record<
+    string,
+    Array<StateNode<TContext, any, TEvent>>
+  > = currentState ? currentState.historyMap : {};
+  if (currentState && currentState.configuration) {
+    // From SCXML algorithm: https://www.w3.org/TR/scxml/#exitStates
+    for (const exitStateNode of new Set(exitSet)) {
```
Andarist: the `new Set` cast isn't needed here, we can iterate through `exitSet` as is just fine

packages/core/test/history.test.ts
```diff
@@ -354,7 +352,9 @@ describe('internal transition with history', () => {
     }
   });
 
-  it('should transition internally to the most recently visited state', () => {
+  // History is not set before the transition is taken, so this does not
```
Andarist: im not entirely sure about this:
1. microstep contains has `exitStates` as its first step: https://www.w3.org/TR/scxml/#microstepProcedure
2. one of the first things `exitStates` does is computing the exit set
3. even internal transitions exit current state according to the second example here: https://www.w3.org/TR/scxml/#N101D7 . The just don't exit its source state, but the source state in here is `second`, not `second.other`
4. so given the fact that `second.other` should be exited here then it should be reentered when this transition targets `.hist` (while it doesnt do it right now and hence fail the test)

Arguably the test description doesnt really tell this story well, so maybe you have thought that it was supposed to test something different? Or maybe it was supposed something different and just was testing the described scenario accidentally? Or maybe I have misinterpreted SCXML semantics?

packages/core/src/stateUtils.ts
davidkpiano: This is needed for determining initial state nodes (see `getFromRelativePath()` calls).

packages/core/test/history.test.ts
davidkpiano: You're right; this will be resolved in a follow-up PR for microsteps

packages/core/src/State.ts
```diff
@@ -212,7 +207,7 @@ export class State<
 
     Object.defineProperty(this, 'nextEvents', {
       get: () => {
-        return nextEvents(config.configuration);
+        return nextEvents([...new Set(config.configuration)]);
```
Andarist: is this needed? shouldnt we assume valid input (no duplicates)?

packages/core/src/StateNode.ts
```diff
@@ -136,7 +134,10 @@ export class StateNode<
 
   protected __cache = {
     events: undefined as Array<TEvent['type']> | undefined,
-    relativeValue: new Map() as Map<StateNode<TContext>, StateValue>,
+    relativeValue: new Map() as Map<
```
Andarist: I've searched through the codebase and this seems to stay unused, some artifact of the past probably.

packages/core/src/StateNode.ts
```diff
@@ -353,7 +354,11 @@ export class StateNode<
   public next(
     state: State<TContext, TEvent>,
     _event: SCXML.Event<TEvent>
-  ): StateTransition<TContext, TEvent> | undefined {
+  ): Transitions<TContext, TEvent> | undefined {
+    if (this.type === 'history' && state.historyValue[this.id]) {
```
Andarist: this is a little bit confusing to me - why do we have to have this check here? If we have stored historyValue for a particular id then we should enter those states, not return an empty array.

Seems like this is handled upstream in the `resolveHistory` already and this condition can be removed, or do I miss something?

packages/core/src/stateUtils.ts
```diff
@@ -96,6 +91,22 @@ export function getChildren<TC, TE extends EventObject>(
   return keys(stateNode.states).map(key => stateNode.states[key]);
 }
 
+export function getProperAncestors<TContext, TEvent extends EventObject>(
+  stateNode: StateNode<TContext, any, TEvent>,
+  toStateNode: StateNode<TContext, any, TEvent> | null
+): Array<typeof stateNode> {
+  const ancestors = new Set<typeof stateNode>();
```
Andarist: this is a fresh Set, since we only add ancestors recursively to it there is no chance for duplicates being added into that, so instead of creating a Set upfront and casting it to an Array at the end we should be able to just create an Array and push to it

packages/core/src/types.ts
```diff
@@ -885,7 +876,6 @@ export interface StateConfig<TContext, TEvent extends EventObject> {
   context: TContext;
   _event: SCXML.Event<TEvent>;
   _sessionid: string | null;
-  historyValue?: HistoryValue | undefined;
```
Andarist: isnt this important for rehydration? shouldnt we accept it?

packages/core/test/scxml.test.ts
```diff
@@ -59,7 +54,7 @@ const testGroups = {
     'history1',
     'history2',
     'history3',
-    // 'history4', // TODO: support history nodes on parallel states
+    'history4', // TODO: support history nodes on parallel states
```
Andarist: this got uncommented, but has a TODO comment now, is this test passing now? if yes - why there is a TODO besides it?

Andarist:
Oh - I forgot, it would be great to describe in short every user-facing changes in changesets.

packages/core/src/StateNode.ts
```diff
@@ -413,40 +418,10 @@ export class StateNode<
...
-      return {
-        transitions: [selectedTransition],
-        entrySet: [],
-        exitSet: [],
-        configuration: state.value ? [this] : [],
-        source: state,
-        actions
-      };
+      return [selectedTransition];
     }
```
khaliu: Is this being repeated? I think this extra if check is returning the same as the default return value below.

Andarist: Good catch! Yes - this same value is returned from both of those code branches, so we should be able to just return it without introducing a conditional path.

.changeset/lucky-bulldogs-march.md
```diff
@@ -0,0 +1,20 @@
...
+
+The `internal` property will no longer have effect for transitions on atomic (leaf-node) state nodes. In SCXML, `internal` only applies to complex (compound and parallel) state nodes:
+
+> Determines whether the source state is exited in transitions whose target state is a descendant of the source state. [See 3.13 Selecting and Executing Transitions for details.](https://www.w3.org/TR/scxml/#SelectingTransitions)
+
+```diff
+// ...
+green: {
+  on: {
+    NOTHING: {
```
Andarist: so this only breaks reentering using internal transitions? the description put upfront is quite abstract, would be good to mention concrete scenarios when this thing matters

davidkpiano: I'll clarify this in the docs, but an internal self-transition in a leaf node is better by not giving a target at all (which is how SCXML represents it) anyway

PR #994, opened 2020-02-11
shallow merges the existing machine context and the passed in context…

### LINKED ISSUE #993: Invoked machines with data wipe out child machine's context
**Description**
When you invoke a machine with context: e.g. parent -> child, and the parent needs to pass down a piece of context, you can use [data](https://xstate.js.org/docs/guides/communication.html#invoking-with-context) to do that. 

However, what if the child has its own context it needs to keep track of? It appears that when using `data` to invoke the child machine, it completely wipes the child machine's context and it only has the context provided by the parent.

**Expected Result**
Invoking a child machine while using `data` should add to the child machine's context, not overwrite it.

**Actual Result**
Invoking a child machine while using `data` overwrites all values of the child's context.

**Reproduction**
https://xstate.js.org/viz/?gist=2740527eeb35dd946e6b711f1c76f7e7

open the console and see the `context: {}` is logged with only the context that was assigned by the parent, and the child's other context items were wiped out.

**Additional context**
n/a

frehner:
… from the parent.

not sure what sort of effect this change will have, since it appears that when it was created it was intentionally not merging the contexts. But all the tests pass. :D 

Fixes #993

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -460,12 +460,15 @@ class StateNode<
   /**
    * Clones this state machine with custom context.
    *
-   * @param context Custom context (will override predefined context, not recursive)
+   * @param context Custom context (will shallow merge predefined context with new context)
    */
   public withContext(
     context: TContext
   ): StateNode<TContext, TStateSchema, TEvent> {
-    return new StateNode(this.config, this.options, context);
+    return new StateNode(this.config, this.options, {
+      ...this.context,
+      ...context
+    });
   }
 
   /**
--- a/packages/core/test/invoke.test.ts
+++ b/packages/core/test/invoke.test.ts
@@ -834,6 +834,68 @@ describe('invoke', () => {
         })
         .start();
     });
+
+    it('should add to the child context, not overwrite it', done => {
+      let contextFromChild;
+      const childMachine = Machine({
+        id: 'pong',
+        initial: 'active',
+        context: {
+          externalContext: null,
+          internalContext: 'child'
+        },
+        states: {
+          active: {
+            type: 'final',
+            data: { secret: 'pingpong' },
+            entry: context => {
+              contextFromChild = { ...context };
+            }
+          }
+        }
+      });
+
+      const parentMachine = Machine({
+        id: 'ping',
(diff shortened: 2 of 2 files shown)


davidkpiano:
As innocuous as this seems, this might be a breaking change... but might not be. What do you think @Andarist ?

frehner:
Yeah I agree that it has the potential to be a breaking change. I’i have workaround I can use until it’s released, so it doesn’t bother me if you merge it into the 5x branch

frehner:
Though thinking about it, because the parent context is merged second it should always win. So what would happen is that you just have additional properties on the context object that weren’t there before... hypothetically not a breaking change right?

davidkpiano:
> Though thinking about it, because the parent context is merged second it should always win. So what would happen is that you just have additional properties on the context object that weren’t there before... hypothetically not a breaking change right?

Yeah this was my thinking too. Would be good to get a second opinion too.

tivac:
Seems additive to me, and thus not technically breaking.

davidkpiano:
Sounds good to me - I myself have wished that it shallow merged, and my original thinking in the API design was that the `context` could be anything, but in V5 we will enforce that it is an object.

Can you add a changeset? `cd` into the root directory and run `yarn changeset` - this would be a `minor` change to `xstate`.

Andarist:
This technically **is a breaking change**, but I don't think we need to worry about it that much as pragmatically it probably wouldn't break anyone or just a super small fraction of users.

There is a possibility of `TContext` having an optional property that is filled in and which gets removed currently, but would be preserved with your change merged in. So if somebody would rely on it somehow it could break them.

I believe this is a change that is already merged into v5 branch, so maybe it would be better to hold off this PR and just release v5-alpha version sometime in the near future, so you could install it to get this feature sooner than others.

frehner:
If you think this change is already in v5, would it be preferable that I check that branch out and add the test there just in case?

Or what would you like me to do at this point?

davidkpiano:
@frehner It's already in the `next` branch 👍

frehner:
Ok. I'll close this PR then - and leave the issue open until there's a 5x release?

PR #1000, opened 2020-02-13
Update Multiple Guards example

mergebandit:
The multiple guards example sets the initial context of `alarm` to `false`. After calling `doorService.send('OPEN')` the example says that the state will update to `{ closed: 'error' }`, which is not true unless `alarm` is set to `true` in the original context definition.

This PR updates the documentation to include a `SET_ALARM` action. Alternatively we could just set the context to `true` 🤷‍♂

### DIFF
--- a/docs/guides/guards.md
+++ b/docs/guides/guards.md
@@ -218,6 +218,9 @@ const doorMachine = Machine(
           SET_ADMIN: {
             actions: assign({ level: 'admin' })
           },
+          SET_ALARM: {
+            actions: assign({ alert: true })
+          },
           OPEN: [
             // Transitions are tested one at a time.
             // The first valid transition will be taken.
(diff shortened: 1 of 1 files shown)


docs/guides/guards.md
```diff
@@ -247,6 +250,13 @@ const doorService = interpret(doorMachine)
   .start();
 // => { closed: 'idle' }
 
+doorService.send('OPEN');
+// => { closed: 'idle' }
+
+doorService.send('SET_ALARM');
+// => { closed: 'idle' }
```
Andarist: it would be more understandable for me if those 2 lines of comments would be reordered

Andarist: however - i think the other fix for this (updating the initial context value) would be good enough

davidkpiano: I think it's good as-is because we want to demonstrate that the door will not go to `{ closed: 'error' }` if `alert: false`.

PR #1002, opened 2020-02-14
Removed support for `service.send(type, payload)`

Andarist:
As discussed in https://github.com/davidkpiano/xstate/issues/998 and on Slack


davidkpiano:
Note to self: add `src/events.ts` and add `eventFrom()` helper to mimic this behavior (opt-in).

PR #1004, opened 2020-02-15
Types: Disallow chaining multiple withConfig/withContext calls

CodingDive:
I'm not sure if there is any merit in allowing to call `machine.withConfig({/* */}).withConfig({/* *})`.

There shouldn't be for the `withContext` as it creates an entirely new context. 
Arguably, for the `withConfig` it's a different story as it merges all the configuration options but I can't see how anybody would ever want to use it.

If `withConfig` should indeed only be called once, the PR adds a tiny improvement to the type definition as seen in the images below.

Before PR (current behavior)
image

After PR
image


CodingDive:
Before pulling from upstream, this looked like a pretty easy and straightforward change 😁 
Let me know if you want me to pursue this idea further and fix the type errors. After briefly looking at them, I believe it might complicate the type definition quite a bit and touch multiple files.

Andarist:
While this is possibly not something that is very much useful I'm not sure if doing types-shenanigans to forbid it is worth it. The main problem of this is that now anything accepting `StateNode` type **won't** be able to consume the created value and it should be possible to accept it.

CodingDive:
Right. I thought the minor type improvement for the autocomplete was cool but I also don't think it's worth complicating the type definitions. 
Thank you for the quick reply 😊

PR #1011, opened 2020-02-19
Use pretty-quick instead of lint-staged

FredyC:
A little DX improvement. The `pretty-quick` module is a faster variant of a classic `lint-staged` approach which tends to be fairly slow. 

Immer started using this recently and we switched over to it in MobX as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,10 +56,10 @@
     "jsdom": "^14.0.0",
     "jsdom-global": "^3.0.2",
     "lerna-alias": "3.0.3-0",
-    "lint-staged": "^8.2.1",
     "markdown-it-codesandbox-embed": "^0.1.0",
     "mermaid": "^8.2.3",
     "prettier": "^1.18.2",
+    "pretty-quick": "2.0.1",
     "spawn-command": "0.0.2-1",
     "ts-jest": "^24.1.9",
     "tslint": "^5.11.0",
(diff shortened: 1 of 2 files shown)


Andarist:
Thanks for the PR!

I'm not the biggest fan of this change though - I don't see anything, in particular, lacking from the lint-staged and I haven't noticed it being too slow. It gets the job done.

`pretty-quick` also seems focused on running prettier, while `lint-staged` is a more general tool and can be used for other tasks as well.

FredyC:
> while `lint-staged` is a more general tool and can be used for other tasks as well.

Yes, but is it used for other tasks in practice? I haven't seen such a case so far :) You can have both tools alongside anyway if there would be such need in the future.

Up to you. It probably depends on how fast your machine is. Especially on Windows, the difference in speed is fairly noticeable.

Andarist:
> Yes, but is it used for other tasks in practice? I haven't seen such a case so far :)

It's pretty often also used for linting. I also had to generate .md files couple of times as part of the precommit hook.

> You can have both tools alongside anyway if there would be such need in the future.

Right, but they are similar enough that's probably not worth to keep both and coordinate their usage.

davidkpiano:
We will be using `lint-staged` for building documentation as well.

FredyC:
Alright, nevermind then :) I was just excited because it made my committing so fast again :)

PR #1016, opened 2020-02-21
Use preconstruct to build dist files

Andarist:
This unifies the build setup for the whole monorepo - it's using a single tool with a single, minimal, config. `preconstruct` is opinionated about what files are being built, how they are being built, what filenames they have. Its benefits:
- monorepo support
- ease of adding custom entries (!)
- constraining some things allows for autofixing, validation and similar features (which were helpful when migrating!)
- it's built around "best practices" and will evolve with them
- preparing dist files is easy to get wrong, because for a such "minor" thing it is super complex topic, so using a wrapper tool is the only sane thing to do in a long run

Other things:
- switch to Babel for transpilation. `preconstruct` only works with it - but it also gives us more fine-grained control over the transpilation output & allows us to add custom transforms in the future. Besides added config - this is really not a big deal, changes which had to be made for this were rather minimal.
- temporarily removed typechecking from jest - I plan to bring it back, but currently, I haven't found a perfect way to do it. I have an idea how this should work - the best thing would be to use a  jest runner for this, but the existing one is **super** slow. I've reached to some people to see how the situation can be improved - however I don't want to stall this PR over this.
- centralize scripts (build, test) at the root -> way less configs, but all of them kinda have to be ran from the root (a minor downside in my opinion)


packages/core/src/index.ts
```diff
@@ -65,3 +66,13 @@ export {
 export type Actor = ActorType;
 
 export * from './types';
+
+// TODO: decide from where those should be exported
```
Andarist: Those had to be exposed because they were used by other packages. By switching to preconstruct we are effectively hiding all internal modules from the outside world, so those couldn't be accessed anymore by reaching into `xstate/lib`.

babel.config.js
```diff
@@ -0,0 +1,23 @@
+const { NODE_ENV } = process.env;
+const isTest = NODE_ENV === 'test';
+
+module.exports = {
+  presets: [
+    [
+      '@babel/preset-env',
```
Andarist: in followup PRs I'm going to optimize this Babel config to output terser code

packages/core/package.json
```diff
@@ -2,20 +2,17 @@
   "name": "xstate",
   "version": "4.7.8",
   "description": "Finite State Machines and Statecharts for the Modern Web.",
-  "main": "lib/index.js",
-  "module": "es/index.js",
-  "types": "lib/index.d.ts",
```
Andarist: you might be wondering why `"types"` have been removed here and in other packages - typings are distributed in `dist` directory and main entry point receives a sibling `.d.ts` file which allows TS to resolve distributed types, making `"types"` field obsolete

packages/core/package.json
```diff
@@ -47,26 +34,22 @@
     "url": "https://github.com/davidkpiano/xstate/issues"
   },
   "homepage": "https://github.com/davidkpiano/xstate/tree/master/packages/core#readme",
+  "dependencies": {
```
Andarist: This had to be included because `xstate` imports it from within `src/scxml.ts`. I bet you don't want to have this here - but we'd have to restructure things in some followup PR to get rid of this.

This actually has showcased a really nice validation provided by preconstruct - it has validated, during build, if your package.json declares all imported dependencies.

davidkpiano: Could this be a dev-dependency?

Andarist: Nope, it's currently part of the core package and has to declared as either dependency or peer dependency.

It's because `@xstate/scxml` imports `toMachine` here:
https://github.com/davidkpiano/xstate/blob/53bed804ec90e9c2821ed5dcd9d1b9443b8ae533/packages/xstate-scxml/test/scxml.test.ts#L7 which is depending on this `xml-js` package.

This is not a big deal though - we only need to restructure things a little bit. Most likely `core/src/scxml.ts` should just be moved to `@xstate/scxml` package and then we could remove this dependency from here (because it would be moved to the other package). This should in a followup PR - if you have any other idea on how to restructure those things then let's discuss this on Slack.

PR #1030, opened 2020-02-26
Annotate failing SCXML tests

Andarist:
Those comments **are not** exhaustive. If we resolve a particular mentioned issue a test in question might still be failing for some other reason - so the comment would have to be adjusted then.

### DIFF
--- a/packages/core/src/invoke.ts
+++ b/packages/core/src/invoke.ts
@@ -38,9 +38,10 @@ export function spawnMachine<
       ) as TMachine;
     }
     const childService = interpret(resolvedMachine, {
-      ...options, // inherit options from this interpreter
+      ...options,
       parent,
-      id: id || resolvedMachine.id
+      id: id || resolvedMachine.id,
+      clock: (parent as any).clock
     });
 
     const resolvedOptions = {
--- a/packages/core/test/scxml.test.ts
+++ b/packages/core/test/scxml.test.ts
@@ -63,7 +63,7 @@ const testGroups = {
     // 'test0', // not implemented
   ],
   in: [
-    // 'TestInPredicate', // In() conversion not implemented yet
+    // 'TestInPredicate', // conversion of In() predicate not implemented yet
   ],
   'internal-transitions': ['test0', 'test1'],
   misc: ['deep-initial'],
@@ -96,8 +96,8 @@ const testGroups = {
     'test4',
     'test5',
     'test6',
-    // 'test7',
-    // 'test7b',
+    'test7',
+    'test7b',
     'test8',
     'test9',
     'test10',
@@ -111,8 +111,9 @@ const testGroups = {
     'test18',
     'test19',
     'test20',
-    // 'test21',
-    // 'test21b',
+    'test21',
+    'test21b',
+    'test21c',
     'test22',
     'test23',
     'test24',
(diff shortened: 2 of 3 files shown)


packages/core/src/invoke.ts
```diff
@@ -38,9 +38,10 @@ export function spawnMachine<
       ) as TMachine;
     }
     const childService = interpret(resolvedMachine, {
-      ...options, // inherit options from this interpreter
+      ...options,
       parent,
-      id: id || resolvedMachine.id
+      id: id || resolvedMachine.id,
+      clock: (parent as any).clock
```
Andarist: do we need to support `clock` option or has it been introduced solely for testing purposes?

packages/core/src/invoke.ts
```diff
@@ -38,9 +38,10 @@ export function spawnMachine<
       ) as TMachine;
     }
     const childService = interpret(resolvedMachine, {
-      ...options, // inherit options from this interpreter
```
Andarist: this comment was not true, this is just spreading spawn options, not inheriting from the parent

packages/core/src/invoke.ts
davidkpiano: In V5 we won't need to; I plan on making the clock a standalone service instead of baked-in.

packages/core/test/scxml.test.ts
```diff
@@ -137,194 +138,205 @@ const testGroups = {
...
     'test174.txml',
     'test175.txml',
     'test176.txml',
-    // 'test179.txml',
-    // 'test183.txml',
+    // 'test179.txml', // conversion of <content> in <sens> not implemented yet
+    // 'test183.txml', idlocation not implemented yet
     'test185.txml',
-    // 'test186.txml',
+    // 'test186.txml', // not sure yet why
```
davidkpiano: 😆

packages/core/test/scxml.test.ts
```diff
@@ -137,194 +138,205 @@ const testGroups = {
...
-    // 'test322.txml',
-    // 'test323.txml',
-    // 'test324.txml',
-    // 'test325.txml',
-    // 'test326.txml',
-    // 'test329.txml',
-    // 'test330.txml',
-    'test331.txml',
-    // 'test332.txml',
+    // 'test319.txml', // SCXML has no init event, so _event stays unbound in onentry of initial state
```
davidkpiano: Makes me wonder if we should get rid of the `xstate.init` event (as you suggested).

I would be fine having a `null as any` as the event, and allowing it to throw if the developer happens to try to use the `event` in any entry action on the initial state (which they shouldn't do).

packages/core/test/scxml.test.ts
Andarist: :trollface:

packages/core/test/scxml.test.ts
Andarist: Definitely something to discuss. One problem with removing it is that we:
- either would have to annotate `event` as optional inside actions (especially those defined on the `options.actions`
- or annotate them as TEvent and let it crash at runtime if one access it

Introducing "XState type compiler" would just fix this issue (given people would set it up), but I doubt we are able to ship something like that together with v5 release - it's rather a bigger project & enhancement that we could explore further down the road.

I'm not opposed to removing the init event, just mentioning the downsides of doing so.

packages/core/src/invoke.ts
Andarist: k, gonna leave this here - but remove `clock` in a followup PR completely (we'll have to rewrite some tests to achieve that)

packages/core/test/scxml.test.ts
davidkpiano: I would prefer annotating it and letting it crash. We can even catch it and provide a descriptive error message, as the developer should not try to access the event initially anyway.

PR #1035, opened 2020-03-01
Fix Transition Type

kofh:
Docs previously labeled the default as "external" but I believe the default is "internal"


davidkpiano:
The default is indeed `"external"`, according to the SCXML spec (which is followed by XState): https://www.w3.org/TR/scxml/#transition

kofh:
Aha, and so it is. Sorry for the mixup!

PR #1037, opened 2020-03-02
Remove clock option

Andarist:
Followup to https://github.com/davidkpiano/xstate/pull/1030#discussion_r384432402


tivac:
We use clock heavily in our automated tests to avoid having to wait for delay transitions, is there a plan for its replacement?

I read the linked comment, it's a bit sparse on details.

Andarist:
@tivac could u use global fake timers such as https://jestjs.io/docs/en/timer-mocks.html ?

tivac:
Not easily, we have other code using timers and being able to do a targeted override for just the statechart is really valuable.

We're also running our statechart code in `puppeteer` and using `jest-puppeteer` to connect to it, so that adds another fun layer of complicated.

Andarist:
> Not easily, we have other code using timers and being able to do a targeted override for just the statechart is really valuable.

I would consider this an anti-pattern - IMHO you should only speed up execution for testing purposes, but you shouldnt do it selectively because you expose yourself to various scheduling bugs (your test environment differs quite significantly from the production one).

> We're also running our statechart code in puppeteer and using jest-puppeteer to connect to it, so that adds another fun layer of complicated.

This shouldnt be a vastly different situation - you can, for example, use https://sinonjs.org/releases/v1.17.6/fake-timers/ .

davidkpiano:
Let's hold off on this - I want to introduce a clock as a default service so that switching out clocks can be done in the same way, rather than baked into the interpreter.

So it will be removed, but I'd like to replace it as well. If it's okay @Andarist, I'll continue on this branch to add the clock service.

haywirez:
I was just about to submit a PR with adding `clock` to the interpreter options in the docs when I found this issue... I got worried for a second, but adding a replaceable clock as a default service instead seems like a reasonable change. 

For another use case to keep in mind, scheduling in media players and web audio apps often heavily relies on using service workers to prevent timer throttling in unfocused tabs. Quoting https://github.com/chrisguttandin/worker-timers:

> For scripts that rely on WindowTimers like setInterval() or setTimeout() things get confusing when the site which the script is running on loses focus. Chrome, Firefox and maybe others throttle the frequency of firing those timers to a maximum of once per second in such a situation. However this is only true for the main thread and does not affect the behavior of Web Workers. Therefore it is possible to avoid the throttling by using a worker to do the actual scheduling. This is exactly what WorkerTimers do.

Andarist:
We've agreed with @davidkpiano that this can land as is (it's targeting the next branch that is not yet being published anywhere), but most likely we are going to think about some replacement for this.

The worker timers issue is a compelling argument - I also had to deal with this in the past. That being said it's a rare-ish use case and can be handled by creating a dedicated "worker timer" actor. That would force you to ditch using `delay` property on send objects though (which is convenient).

There are many tradeoffs here and things to consider, we'll try to deliver the best solution we can think of. We believe that you shouldn't use the custom "clock" (however the API would look like) for testing purposes - global fake timers are way better suited for that and should be preferred. We are also going to document using fake timers as the preferred solution for testing time-dependent machines.

haywirez:
> That being said it's a rare-ish use case and can be handled by creating a dedicated "worker timer" actor. That would force you to ditch using `delay` property on send objects though (which is convenient).

Not just that, I use dynamic delays for state transitions that depend on rapidly changing context values (playback speed) to control the frequency of certain checks in my player. I assume having to deal with a separate actor that now also needs each of those updates would make the whole hierarchy a lot more cumbersome and "chatty" 😞

davidkpiano:
@haywirez Thanks for the concern, we're still debating the right direction to take this. 

Currently, time is an event, so the way these delayed transitions/events are triggered will remain exactly the same.

The only difference (tentative) is that the clock won't be baked into the interpreter. Neither your code nor the "chattiness" of your services will change. It's an implementation detail of XState.

tivac:
What is the impetus for removing it? It doesn't seem like a lot of code in xstate but it dramatically increases the end-user burden if they want to manipulate timers. Is there something I'm not aware of that makes it particularly burdensome to support?

Using something like `@sinonjs/fake-timers` means your tests now need to be responsible for every single timer in your app and ensuring they're firing, or giving up all control and letting time move forward on specific intervals. This is all a massive change from how our tests worked previously where we only needed to care about managing timers for the statechart (since it drives our overall app state other timers are less important when we're doing integration tests).

I strongly disagree with @Andarist's assertions around our current usage being an anti-pattern and am dreading updating our hundreds of tests one by one to account for this change.

davidkpiano:
Revert PR: https://github.com/davidkpiano/xstate/pull/1046

PR #1039, opened 2020-03-03
Make withContext argument Partial

FredyC:
I am not sure if I am overlooking something obvious, but it feels odd I would need to specify full context shape whenever I need a machine with one value changed.

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -463,7 +463,7 @@ class StateNode<
    * @param context Custom context (will override predefined context, not recursive)
    */
   public withContext(
-    context: TContext
+    context: Partial<TContext>
   ): StateNode<TContext, TStateSchema, TEvent> {
     return new StateNode(this.config, this.options, context);
   }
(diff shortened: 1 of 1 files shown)


davidkpiano:
It currently requires the entire `context`, but in V5 it will be partial. So we'll hold off on this one until then. Thank you, though!

FredyC:
Hm :/ But how far is the V5? I don't like the idea of adding hacks to my code while it can be fixed easily on XState side.

davidkpiano:
This PR doesn't necessarily fix it, though, because it's not partial - it will replace the entire context with only the partial context that you gave it.

However, I can add an updater function that is backwards- and forwards-compatible:

```js
const newMachine = machine.withContext(ctx => ({
  ...ctx,
  foo: 'bar'
});
```

How would that sound? cc. @Andarist

FredyC:
Oh, I don't know why I thought it accepts partial, it was some logical assumption I guess 😊 That updater function sounds like a great addition for sure, but if it's too complicated don't worry about it. I don't want to delay your other work considering nobody else seems to be concerned by this.

davidkpiano:
@FredyC There was another similar concern raised in another PR, so it's definitely shared.

PR #1041, opened 2020-03-03
Deep initials

Andarist:


davidkpiano:
I believe this will conflict with mine in the `v5/microstep` branch, which represents `initial` as a transition (`InitialTransitionDefinition`) rather than an array of states. 

The transition definition's `.target` property can specify 1 or more state nodes, as this PR does.

Can you compare it to that branch? I'll make a PR for that soon

packages/core/src/StateNode.ts
```diff
@@ -405,60 +407,18 @@ export class StateNode<
     return selectedTransition ? [selectedTransition] : undefined;
   }
 
-  public get initialStateValue(): StateValue | undefined {
```
Andarist: I believe this can be just retrieved by accessing `.initialState.value`, so I don't think it's worth keeping it around

packages/core/src/StateNode.ts
```diff
@@ -405,60 +407,18 @@ export class StateNode<
...
-    return this.__cache.initialStateValue;
-  }
-
   /**
    * The target state value of the history state node, if it exists. This represents the
    * default state value to transition to if no history value exists yet.
    */
-  public get target(): StateValue | undefined {
-    let target;
+  public get target(): string | undefined {
```
Andarist: this probably could be removed in the future, once we represent history state as a state with default transition etc, but for now, I've decided to keep it to reduce the scope of the PR

packages/core/src/interpreter.ts
```diff
@@ -464,13 +464,11 @@ export class Interpreter<
     const resolvedState =
       initialState === undefined
         ? this.initialState
-        : withServiceScope(this, () => {
```
Andarist: I've found it not be required as `machine.resolveState` just resolves `.configuration` and `.value`, it cant call `spawn` in the process (which is the only thing needing the "service scope" right now)

packages/core/src/stateUtils.ts
```diff
@@ -523,76 +510,28 @@ function isHistoryNode<TContext, TEvent extends EventObject>(
...
-  const initialStateNodePaths = toStatePaths(stateNode.initialStateValue!);
-  return flatten(
-    initialStateNodePaths.map(initialPath =>
-      getFromRelativePath(stateNode, initialPath)
-    )
+  ];
+  const mutStatesToEnter = new Set<StateNode<TContext, any, TEvent>>();
+  const mutStatesForDefaultEntry = new Set<StateNode<TContext, any, TEvent>>();
+
+  computeEntrySet(
```
Andarist: this is quite hacky but solves the problem of resolving initial states recursively with the existing algorithm

 i would like to refactor this in the future, but for the time being it seems OK - @davidkpiano's work on representing initial states as transitions will affect this anyway, so that would be a good time to refactor this

packages/core/src/stateUtils.ts
```diff
@@ -609,8 +548,9 @@ export function getInitialState<
  */
 export function getStateNode<TContext, TEvent extends EventObject>(
   stateNode: StateNode<TContext, any, TEvent>,
-  stateKey: string
-): StateNode<TContext, any, TEvent> {
+  stateKey: string,
+  discardInvalidNodes: boolean = false
```
Andarist: I don't like this flag at all, will look into removing it in the future. This has been introduced solely to pass this test:
https://github.com/davidkpiano/xstate/blob/797a53ad88767f649580628380bbdf86e9fc8a0c/packages/core/test/invalid.test.ts#L50
I'm not convinced if we should bother with this within core algorithms - it would be way easier if we could just assume valid values/configurations. If one migrates their schema they should be responsible for resolving conflicts - especially because we can't resolve all of them and a decision on how to migrate removed regions would be better done on the developer's side.

packages/core/src/stateUtils.ts
```diff
@@ -1035,11 +936,26 @@ function getEffectiveTargetStates<TC, TE extends EventObject>(
...
+  for (const s of transition.target) {
     if (isHistoryNode(s)) {
-      resolveHistory(s, state).forEach(t => targets.add(t));
+      if (state.historyValue[s.id]) {
+        state.historyValue[s.id].forEach(node => {
+          targets.add(node);
+        });
+      } else {
+        getEffectiveTargetStates(
+          { target: resolveHistoryTarget(s) } as TransitionDefinition<TC, TE>,
```
Andarist: this is also a hack, as the only thing actually needed by `getEffectiveTargetStates` is the target property, but this will be refactored once we add a default transition to history states, that can be done in a followup PR after both this PR and the microstep work lands on the next branch

packages/core/src/stateUtils.ts
```diff
@@ -243,7 +241,6 @@ export function getValue<TC, TE extends EventObject>(
   configuration: Configuration<TC, TE>
 ): StateValue {
   const config = getConfiguration([rootNode], configuration);
```
Andarist: im not sure why this accepts 2 arguments, i've removed the "prevStateNodes" one and all tests pass without it, should I remove it in a followup PR?

packages/core/src/StateNode.ts
```diff
@@ -336,6 +325,19 @@ export class StateNode<
     );
   }
 
+  public get initial(): StateNode<TContext, any, TEvent>[] {
```
davidkpiano: Can you please represent this as an `InitialTransitionDefinition` where that is defined as:

```ts
export interface InitialTransitionDefinition<
  TContext,
  TEvent extends EventObject
> extends TransitionDefinition<TContext, TEvent> {
  target: Array<StateNode<TContext, any, TEvent>>;
}
```

Should be a straightforward change and reduce conflicts.

packages/core/test/algorithm.test.ts
```diff
@@ -86,7 +86,6 @@ describe('algorithm', () => {
       'a',
       'b1',
       'c1',
-      'c2',
```
Andarist: I've changed the implementation of the `getConfiguration` slightly and this was the only test that has failed in the result. As mentioned in another comment - I don't know why `getStateNodes` accepts `prevNodes`. What's the exact purpose of `getConfiguration`? I use it to resolve initial nodes & ancestors of input nodes.

packages/core/src/stateUtils.ts
Andarist: @davidkpiano what is the point of `getConfiguration`? I'm not sure how why it accepts prev & current list of state nodes.

packages/core/test/algorithm.test.ts
Andarist: @davidkpiano would love your commentary on that

packages/core/src/stateUtils.ts
Andarist: @davidkpiano thoughts on this one? should we bother "fixing" invalid inputs in the core?

packages/core/test/algorithm.test.ts
davidkpiano: The previous state nodes are needed if only a partial configuration was given, e.g., if a compound node was given and not the child nodes, we can determine the child nodes from the previous config.

packages/core/src/stateUtils.ts
davidkpiano: We shouldn't, you're right. Let's assume valid values/configs.

packages/core/test/algorithm.test.ts
Andarist: So is this intended as the public API? Because as far as I can tell such a situation never happens internally.

davidkpiano: Not public API. Can safely be removed.

PR #1046, opened 2020-03-05
Revert "Remove clock option"

davidkpiano:
Reverts davidkpiano/xstate#1037


davidkpiano:
@andarist Based on comments from #1037, we should rethink this.

PR #1049, opened 2020-03-06
[Docs] Add new simple Star Wars API search example

Glutnix:


Glutnix:
I'm really excited to contribute this example to the project :)

Andarist:
This is a cool example! I'm wondering though if we shouldn't keep such examples in this repository - so we can update them when releasing new versions of XState. WDYT @davidkpiano ?

davidkpiano:
Right, this is awesome, but let's make a new `xstate-examples` repository. I'll do that today!

Andarist:
This could also be kept here - in the monorepo. I don't mind creating extra repo for this though.

Glutnix:
Let me know what I need to do :) I’ll merge my example code to wherever :)

davidkpiano:
Hey thanks again for this! We're going to create a similar example and keep it inside the repository.

PR #1057, opened 2020-03-09
Simplify getConfiguration

Andarist:
Followup to https://github.com/davidkpiano/xstate/pull/1041#discussion_r389821019


packages/core/test/algorithm.test.ts
```diff
@@ -1,95 +0,0 @@
-import { Machine } from '../src';
```
Andarist: as `getConfiguration` is not part of the public API I thought it's not really that relevant for us to test this implementation detail, can revert this deletion if you prefer to keep it

PR #1061, opened 2020-03-11
Infer Typestate in useMachine based on passed-in machine

### LINKED ISSUE #1059: @xstate/react/lib/fsm missing TypeState support
**Description**
It just sets the `TypeState` to `any`, which I didn't expect

**Expected Result**
Don't throw away the `TypeState` generic so that it can be used in the component without a typecast

**Actual Result**
`state.context` isn't narrowed by `value` because `useState` passes `any` as the typestate generic

**Reproduction**
- [CodeSandbox template (React)](https://codesandbox.io/s/competent-kirch-9vhgm)

Andarist:
fixes #1059

### DIFF
--- a/packages/xstate-react/src/fsm.ts
+++ b/packages/xstate-react/src/fsm.ts
@@ -26,17 +26,18 @@ const getServiceState = <
 };
 
 export function useMachine<
-  TC extends object,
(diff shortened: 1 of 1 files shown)


alexreardon:
Is there a release with this one? I've bumped into it

PR #1073, opened 2020-03-14
Cleanup internal structures upon receiving termination events from spawned actors

Andarist:
related to https://github.com/davidkpiano/xstate/issues/996

cc @semopz


davidkpiano:
Great cleanup!

PR #1075, opened 2020-03-15
Export pure() action creator

davidkpiano:


.changeset/old-coins-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Added support for the `pure()` action creator.
```
Andarist: Would be great to include a short description of what it is good for - it's really valuable piece of information for a changelog, shortens the learning loop (one doesnt have to jump into the docs just to get a general sense of what it is useful for).

davidkpiano:
Yes, `pure()` should be `script()`, so you're right, we can probably hold off on this, but the other major use case is "send an event to all dynamically generated actors" which is best represented as:

```js
actions: pure((ctx, e) => {
  return ctx.exampleActors.map(actor => {
    return send(e, { to: actor })
  });
})
```

Andarist:
@davidkpiano do we want to implement equivalent for SCXML's `<foreach/>`? the mentioned use case would be covered by such, so wondering if there are really any other use cases for `pure`? It's definitely easier to write actions using pure rather than with if+foreach combos, but other aspects can suffer from it, such as visualization

davidkpiano:
> @davidkpiano do we want to implement equivalent for SCXML's <foreach/>? the mentioned use case would be covered by such, so wondering if there are really any other use cases for pure? It's definitely easier to write actions using pure rather than with if+foreach combos, but other aspects can suffer from it, such as visualization

Might be a good idea, but it's not that straightforward to implement.

There are many more use-cases for `pure()`, but of course the stricter actions should be preferred.

`pure()` actually allows for better visualization than having that dynamic logic in a random action (we can visualize what it would be in various states), but of course, it'd be better through stricter actions. I'm not too concerned about visualization.

davidkpiano:
This PR is stale (`pure()` and docs already merged in)

PR #1076, opened 2020-03-16
Implement conditional actions

Andarist:
closes https://github.com/davidkpiano/xstate/issues/990

TODO
- [x] changeset
- [ ] docs
- [ ] decide (🤣) about the final name
- [x] add tests

### DIFF
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -456,7 +456,7 @@ export function error(id: string, data?: any): ErrorPlatformEvent & string {
 
   eventObject.toString = () => type;
 
-  return eventObject as (ErrorPlatformEvent & string);
+  return eventObject as ErrorPlatformEvent & string;
 }
 
 export function pure<TContext, TEvent extends EventObject>(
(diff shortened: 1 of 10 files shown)


packages/core/src/actions.ts
```diff
@@ -505,3 +517,97 @@ export function escalate<
...
+              const guard = toGuard(condition.cond, machine.options.guards)
+              return !guard || evaluateGuard(machine, guard, updatedContext, _event, currentState as any)
+            }
+          )?.actions;
+
+          if (!matchedActions) {
+            return []
+          }
+
+          const resolved = resolveActions(machine, currentState, updatedContext, _event, toActionObjects(toArray(matchedActions)))
```
Andarist: this currently has a weird side effect of assigns outside of conditional actions being prioritized even if they come **after** conditional actions in document order, this will get fixed right away on the next branch thanks to the recent microstep work, but I would prefer to keep those quirky semantics here (maybe just with a note in the README?) as it simplifies the implementation

davidkpiano: Yep, that's fine

packages/core/src/actions.ts
```diff
@@ -505,3 +517,115 @@ export function escalate<
...
+        case actionTypes.log:
+          return resolveLog(
+            actionObject as LogAction<TContext, TEvent>,
+            updatedContext,
+            _event
+          );
+        case actionTypes.decide: {
+          const decideAction = actionObject as DecideAction<TContext, TEvent>;
+          const matchedActions = decideAction.conds.find(condition => {
+            const guard = toGuard(condition.cond, machine.options.guards);
```
Andarist: is "guards" still a good name for those? 🤔 definitely made sense for guarded transitions, but feels sub-optimal for conditional actions

davidkpiano: I'd call it `predicate`

Andarist:
@davidkpiano this is ready for review now, we still need to decide on the final name of this action-wrapper and add docs for this. Would you mind taking care of the docs?

packages/core/src/actions.ts
Andarist: I was referring mainly to `machine.options.guards` as the local name doesnt matter much, the question is if we want to rename this bag of options in the future?

davidkpiano: Ah, doesn't matter. We can always refactor later.

packages/core/src/actions.ts
```diff
@@ -505,3 +517,115 @@ export function escalate<
...
+              !guard ||
+              evaluateGuard(
+                machine,
+                guard,
+                updatedContext,
+                _event,
+                currentState as any
+              )
+            );
+          })?.actions;
```
davidkpiano: Will this work? I've always had problems getting the new `foo?.bar` syntax to work with our current setup.

Andarist: Well, CI is green so it works. Maybe there was some intermediate state with some older version of prettier (or some other tool) which had no support for this syntax yet. Or maybe it's not working on next? If you encounter any problems with this on any branch then let me know and I will investigate

PR #1077, opened 2020-03-17
Update typescript.md

SaulEntersekt:
Help with action typesafety

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -132,7 +132,7 @@ const service = interpret(stateMachine);
 // This will compile
 service.send({ type: "TEST", value: "testvalue" });
 
-// This will have an compile error on the value type
+// This will have a compile error on the `value` type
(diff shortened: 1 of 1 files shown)


docs/guides/typescript.md
```diff
@@ -117,6 +117,31 @@ const myMachineConfig: MachineConfig<TContext, TSchema, TEvent> = {
...
+    type: "TEST";
+    value: string;
+}
+
+const service = interpret(stateMachine);
+
+// This will compile
+service.send({ type: "TEST", value: "testvalue" });
+
+// This will have an
```
davidkpiano: Is this missing a couple words?

SaulEntersekt: Yeah, wanted to say that it would generate an error. You've approved, though. Should I change?

davidkpiano: Yes please

PR #1082, opened 2020-03-19
[@xstate/fsm] Expose state in service. Fixes #1080

### LINKED ISSUE #1080: Suggestion: add `.state` to `@xstate/fsm` service
an intepreted service for an `xstate` machine as a `.state` property which is pretty useful to be able to grab. `@xstate/fsm` tracks this state (`let state = ..`), but does not expose it.

Context: I am looking to create an alternative `useService` for react and getting the state for initial render is tricky as you need to setup a subscriber and grab the state in the first render.

davidkpiano:


packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -258,4 +258,44 @@ describe('interpreter', () => {
...
+          on: { CHANGE: 'success' }
+        },
+        success: {}
+      }
+    });
+    const service = interpret(machine);
+
+    service.start();
+
+    service.subscribe(() => {
```
Andarist: does this have to be async? the test would be pretty much the same if you would just assert this after `service.send('CHANGE');`

PR #1084, opened 2020-03-19
[@xstate/react] useActor hook & Suspense-ready

davidkpiano:


packages/xstate-react/src/useService.ts
```diff
@@ -0,0 +1,28 @@
...
+  const actorRef = useMemo(() => {
+    return fromService<TContext, TEvent>(service);
+  }, [service]);
+
+  const [state, sendActor] = useActor(actorRef);
+
+  const send = useCallback(
+    (event: Event<TEvent>) => {
+      const eventObject = typeof event === 'string' ? { type: event } : event;
+      return void sendActor(eventObject as TEvent);
```
Andarist, line 22: this can be simplified to
```suggestion
      sendActor(eventObject as TEvent);
```

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -338,4 +339,53 @@ describe('useMachine hook', () => {
...
+      }, [state]);
+
+      return <div data-testid="count">{stateCount}</div>;
+    };
+
+    const { getByTestId } = render(<App />);
+
+    const countEl = getByTestId('count');
+
+    act(() => {
```
Andarist, line 383: `act` is only needed when you schedule some React updates and you want to have them flushed within act's callback, in here you just make assertions which is unrelated to React and thus this `act` doesn't do anything special and can be avoided

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -338,4 +339,53 @@ describe('useMachine hook', () => {
...
+      // 1 for the initial state
+      // and 1 for the two (batched) events
+      expect(countEl.textContent).toEqual('2');
+      expect(count).toEqual(2);
```
Andarist, line 387: maybe it would be good to `send('EVENT')` x3 so those numbers would be different, showcasing better the difference

packages/xstate-react/src/useMachine.ts
```diff
@@ -0,0 +1,133 @@
...
+  const [state, setState] = useState(service.state);
+
+  useEffect(() => {
+    // Whenever a new state is emitted from the service,
+    // update the state with that state, but only if
+    // that state has changed.
+    service.onTransition((currentState) => {
+      if (currentState.changed) {
+        // capture side-effects to be executed
+        actionStatesRef.current.push(currentState);
```
Andarist, line 110: im pretty sure that you miss initial actions (and any actions which happen before useEffect's callback) by starting to queue them from this point in time

packages/xstate-react/src/useMachine.ts
```diff
@@ -0,0 +1,133 @@
...
+    return () => {
+      service.stop();
+    };
+  }, []);
+
+  useEffect(() => {
+    // Flush all actions to be executed (per state)
+    actionStatesRef.current.forEach((actionState) => {
+      // Execute all actions for the captured state
+      service.execute(actionState);
```
Andarist, line 126: an additional quirk of this approach which I have thought of right now:
1. you send an event to itself, which is supposed to transition you from A to B - but it gets postponed until useEffect
2. other transitions happen in between, taking you from A to C
3. the postponed send gets executed and sends the event to a machine in a completely different state that it was supposed to, this can lead to no transition being taken or a completely different transition being taken

packages/xstate-react/src/useActor.ts
```diff
@@ -0,0 +1,22 @@
+import { ActorRef, Sender } from './ActorRef';
+import { EventObject } from 'xstate';
+import { useState, useEffect, useMemo } from 'react';
+
+export function useActor<TCurrent, TEvent extends EventObject>(
+  actorRef: ActorRef<TCurrent, TEvent>
+): [TCurrent, Sender<TEvent>] {
+  const [current, setCurrent] = useState(actorRef.current);
+
+  const send = useMemo(() => actorRef.send, [actorRef]);
```
Andarist, line 10: no need to memoize simple property access, it shouldn't be perf sensitive nor it should lead to a different reference being returned each time
```suggestion
  const send = actorRef.send;
```

packages/xstate-react/src/useMachine.ts
davidkpiano: We can test this, but the initial state/actions are sent to this `onTransition` callback.

packages/xstate-react/src/useMachine.ts
davidkpiano: That's fine; `send(...)` is not supposed to be zero-time. You should always anticipate that an event can be sent before `send(...)` is flushed.

For immediate deterministic behavior, that's what `raise(...)` is for.

packages/xstate-react/src/useMachine.ts
Andarist: How could they if you setup onTransition callback in an effect and start a machine during initial render? The initial transitions are getting resolved right after the machine gets started which is before you start listening for transitions

davidkpiano: The initial state includes actions.

Andarist: The initial state is not pushed onto this `actionStatesRef` though 😉

davidkpiano: Fixed

PR #1090, opened 2020-03-22
Include `TTypestate` for methods

rjdestigter:
This fixes issues where the state returned by `useMachine` looses the `TTypestate` information

### DIFF
--- a/.changeset/ninety-toys-smile.md
+++ b/.changeset/ninety-toys-smile.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Added TTypestate to withConfig, state, and start methods
--- a/.changeset/ninety-toys-smile.md
+++ b/.changeset/ninety-toys-smile.md
@@ -2,4 +2,8 @@
 'xstate': patch
 ---
 
-Added TTypestate to withConfig, state, and start methods
+This change carries forward the typestate type information encoded in the arguments of the following functions and assures that the return type also has the same typestate type information:
+
+- Cloned state machine returned by `.withConfig`.
+- `.state` getter defined for services.
+- `start` method of services.
--- a/packages/xstate-react/test/types.test.tsx
+++ b/packages/xstate-react/test/types.test.tsx
@@ -1,7 +1,14 @@
 import * as React from 'react';
 import { render } from '@testing-library/react';
-import { Machine, interpret, assign, Interpreter, spawn } from 'xstate';
-import { useService } from '../src';
+import {
+  Machine,
+  interpret,
+  assign,
+  Interpreter,
+  spawn,
+  createMachine
+} from '../../core/src';
+import { useService, useMachine } from '../src';
 
 describe('useService', () => {
   it('should accept spawned machine', () => {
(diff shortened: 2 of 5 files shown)


rjdestigter:
I want to add tests but I'm having trouble getting the current set of tests to run without failing. I also ran `npx lerna bootstrap` rather than `yarn` but I'm not sure if that was correct.

davidkpiano:
@rjdestigter Seems like tests passed, did you run `yarn` and then `yarn test:core` in the root dir?

rjdestigter:
Yeah, I did and that was successful.

rjdestigter:
If I want to test one or more of the other packages. Do I manually run yarn inside their respective root folder?

davidkpiano:
> If I want to test one or more of the other packages. Do I manually run yarn inside their respective root folder?

Use the same strategy as `yarn test:core`

rjdestigter:
I think this might be lerna + yarn + windows issue. _node_modules_ at the root is linking to _packages/core_ but running `npm test --prefix ./packages/xstate-react` fails due to `xstate` not being found/resolved.

Andarist:
@rjdestigter you can run `yarn dev` in the root to run all tests and tsc at the same time, this should resolve all issues regarding some packages not being found etc

.changeset/ninety-toys-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Added TTypestate to withConfig, state, and start methods
```
Andarist: could you mention what does this fix? this message will appear in our changelog and right now it might be a little bit vague for a reader

packages/xstate-react/test/types.test.tsx
```diff
@@ -57,3 +64,52 @@ describe('useService', () => {
     render(<Todo index={0} />);
   });
 });
+
+describe('useMachine', () => {
+  it('should preserve typestate information.', () => {
```
Andarist: I'm wondering - so far you have only changed `xstate`'s source files, but you have added a test to `@xstate/react`. Any particular reason behind this? Should we add similar tests to `xstate`? Were your changes essential for `@xstate/react`?

rjdestigter: The goal of the changes where to preserve typestate for components using `useMachine`.

rjdestigter:
These changes could be a breaking change for anyone using typestates to define/create their machine but then ignoring them after using any of the methods or functions using these methods, e.g., `useMachine`.

packages/xstate-react/test/types.test.tsx
rjdestigter: We should. I'm working on some additional typestate related tests for core

Andarist: > The goal of the changes where to preserve typestate for components using useMachine.

This was probably already fixed by [this commit](https://github.com/davidkpiano/xstate/commit/30c24bcc7fa7ea7127f3088fc69589d859044852#diff-dd3974b63e6fa683880195c4603244bb). Notice how the return type of `useMachine` is already typed with type states in mind, so your changes had no effect on it.

It's great to have tests to cover this now though!

rjdestigter: 👍 I've added tests for coe

PR #1103, opened 2020-03-30
Upgrade TS to 3.8.3 and simplify transition config types

davidkpiano:
This should also fix the "Type instantiation is excessively deep and possibly infinite." errors. https://github.com/microsoft/TypeScript/issues/37613


packages/core/test/actionCreators.test.ts
```diff
@@ -111,7 +111,10 @@ describe('action creators', () => {
       const resolvedAction = actions.resolveSend(
         action,
         { delay: 100 },
-        toSCXMLEvent({ type: 'EVENT', value: 50 })
+        toSCXMLEvent({ type: 'EVENT', value: 50 } as {
```
Andarist: is this assertion needed? this is not a constrained called anyhow so `toSCXMLEvent` should accept something like `EventObject` or `AnyEventObject` (dont know by heart which one) and `{type:'EVENT',value:50}` should be allowed for this

packages/core/src/types.ts
```diff
@@ -321,28 +321,34 @@ export type StatesDefinition<
...
+} & {
+  '*'?: TransitionConfigOrTarget<TContext, TEvent>;
 };
 
 type TransitionsConfigArray<TContext, TEvent extends EventObject> = Array<
   {
     [K in TEvent['type'] | NullEvent['type'] | '*']: TransitionConfig<
       TContext,
-      K extends TEvent['type'] ? Extract<TEvent, { type: K }> : EventObject
+      TEvent extends { type: K } ? TEvent : never
```
Andarist: is the following `& { event: K }` needed now? it might be obsolete, but im not 100% sure

Andarist: this probably was already the case but this type doesn't really allow for null and wildcard events to be configured here because they will be mapped to `never` (they dont extend TEvent after all)

packages/core/test/actionCreators.test.ts
davidkpiano: Yes it is. Try removing it - TS doesn't like that `type` is a string instead of `'EVENT'`.

packages/core/src/types.ts
davidkpiano: Yes, `{ event: K }` is needed to narrow array transition configs of this style, as you have added:

```js
on: [
  { event: 'something', ... }
  { event: 'another', ... }
]
```

Andarist:
Have u maybe deleted yarn.lock and regenerated it by doing a fresh install? Changes to yarn.lock seem to be too big for what you have changed in package.jsons .

.changeset/curvy-planets-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Upgraded TypeScript to 3.8.3. This also addresses #1015 and simplifies the `TransitionConfigArray` and `TransitionConfigMap` types.
```
Andarist, line 5: "Upgraded TypeScript to 3.8.3" - this part is not really relevant to our consumers. I would just focus here on simplifying those 2 types and mentioning what it is supposed to fix (excessive type instantiation reported by TS). You could also link to the mentioned issue.

```suggestion
Simplify the `TransitionConfigArray` and `TransitionConfigMap` types In order to fix excessively deep type instantiation TypeScript reports. This addresses [#1015](https://github.com/davidkpiano/xstate/issues/1015).
```

patches/@changesets+assemble-release-plan+0.3.0.patch
```diff
@@ -1,13 +0,0 @@
-diff --git a/node_modules/@changesets/assemble-release-plan/dist/assemble-release-plan.cjs.dev.js b/node_modules/@changesets/assemble-release-plan/dist/assemble-release-plan.cjs.dev.js
-index 98b84eb..3d2e5af 100644
---- a/node_modules/@changesets/assemble-release-plan/dist/assemble-release-plan.cjs.dev.js
-+++ b/node_modules/@changesets/assemble-release-plan/dist/assemble-release-plan.cjs.dev.js
-@@ -73,7 +73,7 @@ function getDependents(releases, workspaces, dependencyGraph, preInfo) {
-         versionRange
-       } = getDependencyVersionRange(dependentPkgJSON, nextRelease.name); // Firstly we check if it is a peerDependency because if it is, our dependent bump type needs to be major.
- 
--      if (depTypes.includes("peerDependencies") && nextRelease.type !== "patch" && (!releases.has(dependent) || releases.has(dependent) && releases.get(dependent).type !== "major")) {
-+      if (depTypes.includes("peerDependencies") && nextRelease.type === "major" && (!releases.has(dependent) || releases.has(dependent) && releases.get(dependent).type !== "major")) {
```
Andarist, line 10: well - this has existed for a reason :p it allows us to minor-bump XState without major-bumping dependencies its peer-dependent packages

I can bring this patch back later though

.changeset/curvy-planets-serve.md
davidkpiano: The TypeScript version is definitely relevant to many users.

patches/@changesets+assemble-release-plan+0.3.0.patch
davidkpiano: Please do - I was unable to build without following the instructions and removing this patch.

PR #1113, opened 2020-04-04
Added mentioned ‘strict’ parameter to the code

gustavjorlov:
The text below the code referred to the `strict` parameter, which wasn’t present in the code. This small fix adds it to the example.

### DIFF
--- a/docs/guides/machines.md
+++ b/docs/guides/machines.md
@@ -20,9 +20,6 @@ const lightMachine = Machine({
 
   // Initial state
   initial: 'green',
- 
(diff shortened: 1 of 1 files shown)


davidkpiano:
I think it's fine to leave this out. I don't want to encourage the use of `strict: true` since it will be removed in V5.

gustavjorlov:
Isn’t it better to remove the docs about it then?
It’s actually right at the start of the guide and I think it’s important to keep that part perfect to attract people.

PR #1114, opened 2020-04-04
Target multiple actors by providing a list of refs in the send action's .to property

rjdestigter:
This PR implements support for targeting multiple actors by allowing you to provide a list of references in a _send_ action's `.to` property. E.g.:

```ts
on : {
  TOGGLE_ALL: {
    actions: send('TOGGLE', { to: ['ref1', 'ref2'] })
  }
}
```
I'm not sure if this is SCXML compliant.

Tasks:
- [x] Update documentation
- [x] Finalize test
- [x] Add changeset
- [x] Remove logs

### DIFF
--- a/.changeset/odd-waves-laugh.md
+++ b/.changeset/odd-waves-laugh.md
@@ -1,0 +1,11 @@
+---
+'xstate': minor
+---
+
+Send an event to multiple targets by providing a list of references for a _send_ action creator's `.to` property: E.g.:
+
+```js
+send('SOME_EVENT', to: ['actor1', 'invokedMachineId2'])
+...
+send('SOME_EVENT', to: (ctx, evt) => [ctx.actor1, ctx.actor2, ...evt.addtionalTargets])
+```
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -129,14 +129,14 @@ export class Interpreter<
    * - `clock` uses the global `setTimeout` and `clearTimeout` functions
    * - `logger` uses the global `console.log()` method
    */
-  public static defaultOptions: InterpreterOptions = ((global) => ({
+  public static defaultOptions: InterpreterOptions = (global => ({
     execute: true,
     deferEvents: true,
     clock: {
       setTimeout: (fn, ms) => {
         return global.setTimeout.call(null, fn, ms);
       },
-      clearTimeout: (id) => {
+      clearTimeout: id => {
         return global.clearTimeout.call(null, id);
       }
     },
@@ -294,7 +294,7 @@ export class Interpreter<
     if (this.state.configuration && isDone) {
       // get final child state node
       const finalChildStateNode = state.configuration.find(
-        (sn) => sn.type === 'final' && sn.parent === this.machine
+        sn => sn.type === 'final' && sn.parent === this.machine
       );
 
       const doneData =
@@ -505,7 +505,7 @@ export class Interpreter<
     }
 
     // Stop all children
-    this.children.forEach((child) => {
+    this.children.forEach(child => {
       if (isFunction(child.stop)) {
         child.stop();
       }
@@ -634,7 +634,7 @@ export class Interpreter<
         });
 
         batchedActions.push(
-          ...(nextState.actions.map((a) =>
+          ...(nextState.actions.map(a =>
             bindActionToState(a, nextState)
           ) as Array<ActionObject<TContext, TEvent>>)
         );
@@ -659,8 +659,14 @@ export class Interpreter<
 
   private sendTo = (
     event: SCXML.Event<TEvent>,
-    to: string | number | Actor
+    to: string | number | Actor | Array<string | number | Actor>
   ) => {
+    if (Array.isArray(to)) {
+      console.log(to);
+      to.forEach(target => this.sendTo(event, target));
+      return;
+    }
+
     const isParent =
       this.parent && (to === SpecialTargets.Parent || this.parent.id === to);
     const target = isParent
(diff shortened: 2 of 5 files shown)


rjdestigter:
I'm also having trouble with the test. In the test I have:

```ts
TOGGLE: {
  target: 'TRUE',
  actions: [log((_, _evt) => `Toggle`), sendParent('INCREMENT')]
}
```

and the test succeeds as expected. However, originally I had a delay:

```ts
TOGGLE: {
  target: 'TRUE',
  actions: [log((_, _evt) => `Toggle`), sendParent('INCREMENT', { delay: 100 )]
}
```

and the parent machine never received the INCREMENT action. (While I'm typing this, is this because the child machine is in it's final state before the delay has passed?). I also tried to use the _respond_ action before _sendParent_ but I couldn't get it to work with that.

CodingDive:
I was definitely in need of this a few times already. Will it also work with the context callback?
`to: (context) => [context.ref1, context.ref2]`

rjdestigter:
> Will it also work with the context callback?

It should! I'll add a test for that.

CodingDive:
Nice. The only thing missing for send to handle all of my needs would be to have the event passed as a second argument of the `to: () => ` callback.

Maybe in another issue and pr as long as it's aligned with the SCXML spec.

rjdestigter:
That is already in place? I'm using it in my test atm:

```ts
RUN2: {
  actions: send<any, any>('TOGGLE', { to: (_ctx, evt) => evt.target })
}
```

where `event.target` is an array of refs _(strings)_.

CodingDive:
Wow, thank you and sorry for spamming your PR 😅

packages/core/src/interpreter.ts
```diff
@@ -659,8 +659,14 @@ export class Interpreter<
 
   private sendTo = (
     event: SCXML.Event<TEvent>,
-    to: string | number | Actor
+    to: string | number | Actor | Array<string | number | Actor>
```
davidkpiano, line 662:
```suggestion
    to: string | number | Actor | Array<string | number | Actor>
    to: SingleOrArray<string | number | Actor>
```

davidkpiano:
This looks like a great feature! @Andarist what do you think?

This is already SCXML -> XState compatible, but if we want to make it XState -> SCXML compatible, we have to keep in mind that something like:

```js
actions: send('EVENT', { to: ['actor1', 'actor2'] })
```

needs to be converted to:

```xml
<send target="actor1" event="EVENT"></send>
<send target="actor2" event="EVENT"></send>
```

rjdestigter:
Do we think _xstate_ should concern itself with empty `.to` arrays or maybe log a warning in non-prod? 

Also, you can create a infinite loop with 2 arrays referencing each other. I don't think it should be _xstate_'s responsibility to check for that but I thought I should mention it.

rjdestigter:
> This is already SCXML -> XState compatible, but if we want to make it XState -> SCXML 

Would that be handled in `actionToSCXML` David? https://github.com/rjdestigter/xstate/blob/next/packages/xstate-scxml/src/index.ts#L26

davidkpiano:
> Do we think xstate should concern itself with empty .to arrays or maybe log a warning in non-prod?

No, that should just be a no-op.

> Would that be handled in actionToSCXML David?

Yeah, but that doesn't need to be handled in this PR.

.changeset/odd-waves-laugh.md
```diff
@@ -0,0 +1,11 @@
+---
+'xstate': minor
+---
+
+Send an event to multiple targets by providing a list of references for a _send_ action creator's `.to` property: E.g.:
+
+```js
+send('SOME_EVENT', to: ['actor1', 'invokedMachineId2'])
+...
+send('SOME_EVENT', to: (ctx, evt) => [ctx.actor1, ctx.actor2, ...evt.addtionalTargets])
```
Andarist, line 10: I have a concern that this is really dynamic and we might not be able to convert this pattern to SCXML. We can't know upfront (statically) how many send actions this would return.

I'm wondering if this problem shouldn't really be solved in userland (or with a helper provided by us):
```js
const sendToMany = (event, { to: targets, ...options }) =>
  targets.map((to) => send(event, { to, ...options }));
```
This would solve the "static" list of targets, whereas the dynamic list of targets should most likely be solved by `forEach` action which we plan to introduce (SCXML has support for this). You'd need to keep all your targets in a single array-collection for `forEach` to handle this but this seems like a good constraint - forcing you to organize your context values.

I recognize the need to cover this use case, I'm just proposing alternative solutions that wouldn't suffer from the mentioned drawback

rjdestigter:
I like the idea of an action creator. I'm having trouble understanding your concerns @Andarist. Would the `forEach` action creator return a single action object? How would it be a problem converting that to SCXML?

With `sendToMany ` you would have to spread the result if you are already working with an array of actions. This isn't really a problem other than finding it visually less preferable.

Andarist:
> Would the forEach action creator return a single action object?

Yes, it would contain a description of other actions to be executed though. Kinda similar to recently developed [`choose`](https://github.com/davidkpiano/xstate/pull/1076).

> How would it be a problem converting that to SCXML?

Not sure about which one you are asking - `forEach` would be compatible with SCXML. The proposed:
```
send('EVENT', to: ctx => [/* you can pass any number of targets here *])
``` 
is not. The problem is that we can't statically know how many targets are going to returned there as it's completely dynamic expression and to convert this to SCXML we would have to know this statically to create appropriate number of `<send>` elements.

> With sendToMany you would have to spread the result if you are already working with an array of actions. This isn't really a problem other than finding it visually less preferable.

I don't find this to be much of a problem myself, but we could consider "flattening" array of actions which has been kinda proposed recently [here](https://github.com/davidkpiano/xstate/issues/1104)

rjdestigter:
I think I get it now. When `.to` is a function you can't know how many _send_ events you'll end up with, correct?

Andarist:
Yes, that's correct. Sorry if my explanations were not clear enough.

eddiewang:
@Andarist what would be the best approach to sending off actions when we have a use-case where the number of children actors is dynamic?

CodingDive:
> @Andarist what would be the best approach to sending off actions when we have a use-case where the number of children actors is dynamic?

You can use the pure action creator.

PR #1122, opened 2020-04-10
Passed state type generic to the Config interface

chancestrickland:
The `StateMachine.Config` interface should accept a third generic for our state type, which should then be provided when used in the `createMachine` function. I think. I have declared my own module and haven't had any issues, but if there are any reasons this wasn't set up this way originally let me know.

Using the state type provides helpful intellisense when defining the state chart transitions! 

**NOTE:** I ran tests locally but a bunch of them failed because local modules weren't resolved. Not sure if there is a step missing in `CONTRIBUTING.md` to link these packages.

### DIFF
--- a/.changeset/wise-lemons-travel.md
+++ b/.changeset/wise-lemons-travel.md
@@ -2,4 +2,4 @@
 '@xstate/fsm': patch
 ---
 
-Passed state type generic to the Config interface
+`StateMachine.Config` type accepts now a third type parameter - `TState` - similarly to other existing types. When provided it provides helpful intellisense when defining the state chart transitions.
--- a/.changeset/wise-lemons-travel.md
+++ b/.changeset/wise-lemons-travel.md
@@ -1,0 +1,5 @@
+---
+'@xstate/fsm': patch
+---
+
+Passed state type generic to the Config interface
--- a/packages/xstate-fsm/src/index.ts
+++ b/packages/xstate-fsm/src/index.ts
@@ -81,7 +81,7 @@ export function createMachine<
   TEvent extends EventObject = EventObject,
   TState extends Typestate<TContext> = any
 >(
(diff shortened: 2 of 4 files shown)


packages/xstate-fsm/src/index.ts
```diff
@@ -81,7 +81,7 @@ export function createMachine<
   TEvent extends EventObject = EventObject,
   TState extends Typestate<TContext> = any
 >(
-  fsmConfig: StateMachine.Config<TContext, TEvent>,
+  fsmConfig: StateMachine.Config<TContext, TEvent, TState>,
```
CodingDive, line 84: Just to open a discussion: XState core always uses `<Context, State, Event>` so I'm not sure if @davidkpiano and @Andarist would be okay with the `xstate-fsm` API to differ. In this case, it would obviously be a breaking change to decide in favor of consistency by having the type arguments in the same order as within the core.

chancestrickland: `TState` is already the third argument in the `State` and `Machine` interfaces in `fsm`, so this PR follows those patterns. Since `TState` defaults to `any` this shouldn't be a breaking change.

davidkpiano: @CodingDive Moving forward in V5, the generic arguments will always follow `<TContext, TEvent, ...>` in that order.

.changeset/wise-lemons-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/fsm': patch
+---
+
+Passed state type generic to the Config interface
```
Andarist, line 5: Changesets appear later in changelogs and this sounds really vague to a potential reader, what about something like this:
```suggestion
`StateMachine.Config` type accepts now a third type parameter - `TState` - similarly to other existing types. When provided it provides helpful intellisense when defining the state chart transitions.
```

PR #1128, opened 2020-04-15
doc: seems like this fix would clarify the example

wuweiweiwu:


davidkpiano:
This is actually a bug, since `center` is not a sibling of the root node. See #346

wuweiweiwu:
oh i see thanks for clearing that up for me

PR #1129, opened 2020-04-16
@xstate/immer initial release

davidkpiano:
This PR adds the initial implementation of `@xstate/immer`

```js
import { createMachine, interpret } from 'xstate';
import { assign, createUpdater } from '@xstate/immer';

const levelUpdater = createUpdater('UPDATE_LEVEL', (ctx, level) => {
  ctx.level = level;
});

const toggleMachine = createMachine({
  id: 'toggle',
  context: {
    count: 0,
    level: 0
  },
  initial: 'inactive',
  states: {
    inactive: {
      on: {
        TOGGLE: {
          target: 'active',
          // Immutably update context the same "mutable"
          // way as you would do with Immer!
          actions: assign((ctx) => ctx.count++)
        },
        // Use the updater for more convenience:
        [levelUpdater.type]: {
          actions: levelUpdater.action
        }
      }
    },
    active: {
      on: {
        TOGGLE: { target: 'inactive' }
      }
    }
  }
});

const toggleService = interpret(toggleMachine)
  .onTransition((state) => {
    console.log(state.context);
  })
  .start();

toggleService.send('TOGGLE');
// { count: 1, level: 0 }

toggleService.send(levelUpdater.update(9));
// { count: 1, level: 9 }

toggleService.send('TOGGLE');
// { count: 2, level: 9 }

toggleService.send(levelUpdater.update(-100));
// Notice how the level is not updated in 'inactive' state:
// { count: 2, level: 9 }
```

### DIFF
--- a/packages/xstate-immer/README.md
+++ b/packages/xstate-immer/README.md
@@ -1,3 +1,76 @@
 # @xstate/immer
 
-Coming soon!
+This package contains utilities for using [Immer](https://immerjs.github.io/immer/docs/introduction) with XState.
+
+Included in `@xstate/immer`:
+
+- `assign()` - an Immer action that allows you to immutably assign to machine `context` in a convenient way
+- `createUpdater()` - a useful function that allows you to cohesively specify a context update event, assign action, and validator, all together. (See example below)
+
+## Quick Start
+
+1. Install `xstate` and `@xstate/immer`:
+
+```bash
+npm install xstate @xstate/immer
+```
+
+2. Import the Immer utilities:
+
+```js
+import { createMachine, interpret } from 'xstate';
+import { assign, createUpdater } from '@xstate/immer';
+
+const updateLevel = createUpdater('UPDATE_LEVEL', (ctx, level) => {
+  ctx.level = level;
+});
+
+const toggleMachine = createMachine({
+  id: 'toggle',
+  context: {
+    count: 0,
+    level: 0
+  },
+  initial: 'inactive',
+  states: {
+    inactive: {
+      on: {
+        TOGGLE: {
+          target: 'active',
+          // Immutably update context the same "mutable"
+          // way as you would do with Immer!
+          actions: assign((ctx) => ctx.count++)
+        },
+        // Use the updater for more convenience:
+        [updateLevel.type]: {
+          actions: updateLevel.assign
+        }
+      }
+    },
+    active: {
+      on: {
+        TOGGLE: { target: 'inactive' }
+      }
+    }
+  }
+});
+
+const toggleService = interpret(toggleMachine)
+  .onTransition((state) => {
+    console.log(state.context);
+  })
+  .start();
+
+toggleService.send('TOGGLE');
+// { count: 1, level: 0 }
+
+toggleService.send(udpateLevel(9));
+// { count: 1, level: 9 }
+
+toggleService.send('TOGGLE');
+// { count: 2, level: 9 }
+
+toggleService.send(udpateLevel(-100));
+// Notice how the level is not updated in 'inactive' state:
+// { count: 2, level: 9 }
+```
--- a/packages/xstate-immer/README.md
+++ b/packages/xstate-immer/README.md
@@ -64,13 +64,13 @@ const toggleService = interpret(toggleMachine)
 toggleService.send('TOGGLE');
 // { count: 1, level: 0 }
 
-toggleService.send(udpateLevel(9));
+toggleService.send(updateLevel(9));
 // { count: 1, level: 9 }
 
 toggleService.send('TOGGLE');
 // { count: 2, level: 9 }
 
-toggleService.send(udpateLevel(-100));
+toggleService.send(updateLevel(-100));
 // Notice how the level is not updated in 'inactive' state:
 // { count: 2, level: 9 }
 ```
--- a/packages/xstate-immer/README.md
+++ b/packages/xstate-immer/README.md
@@ -1,13 +1,24 @@
 # @xstate/immer
 
-This package contains utilities for using [Immer](https://immerjs.github.io/immer/docs/introduction) with XState.
+This package contains utilities for using [Immer](https://immerjs.github.io/immer/docs/introduction) with [XState](https://github.com/davidkpiano/xstate).
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Quick Start](#quick-start)
+- [API](#api)
+  - [`assign(producer)`](#assignproducer)
+  - [`createUpdater(eventType, recipe)`](#createupdatereventtype-recipe)
+- [TypeScript](#typescript)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Quick Start
 
 Included in `@xstate/immer`:
 
 - `assign()` - an Immer action that allows you to immutably assign to machine `context` in a convenient way
-- `createUpdater()` - a useful function that allows you to cohesively specify a context update event, assign action, and validator, all together. (See example below)
-
-## Quick Start
+- `createUpdater()` - a useful function that allows you to cohesively define a context update event event creator and assign action, all together. ([See an example](#createUpdater-eventType-recipe) below)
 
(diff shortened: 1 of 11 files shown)


packages/xstate-immer/package.json
```diff
@@ -38,11 +38,11 @@
   },
   "dependencies": {},
   "peerDependencies": {
-    "immer": "^5.0.0",
+    "immer": "^6.0.3",
```
Andarist, line 41: is 6.0.3 required? or could we lower this to 6.0.0?

davidkpiano: Might as well use the latest 😛

Andarist: 👍

packages/xstate-immer/test/immer.test.ts
```diff
@@ -1,12 +1,13 @@
-import { Machine } from 'xstate';
-import { assign, assignPatch, patchEvent } from '../src';
+// @ts-ignore
```
Andarist, line 1: ? :D doesnt this typecheck?

packages/xstate-immer/test/immer.test.ts
```diff
@@ -100,40 +101,108 @@ describe('@xstate/immer', () => {
...
+          type: 'SUBMIT';
+        };
+
+    const formMachine = createMachine<typeof context, FormEvent>({
+      initial: 'editing',
+      context,
+      states: {
+        editing: {
+          on: {
+            [nameUpdater.type]: { actions: nameUpdater.assign },
```
Andarist, line 176: probably we could use util for this, seems repetitive - can be done later though if we ain't 100% sure how to name such a util right now

packages/xstate-immer/src/index.ts
```diff
@@ -16,37 +16,51 @@ export interface ImmerAssignAction<TContext, TEvent extends EventObject>
...
-export function patchEvent<TContext>(
-  type: string,
-  context: TContext,
-  recipe: (draftContext: TContext) => void
-): PatchEventObject {
-  return {
-    type,
-    patches: produceWithPatches(context, recipe)
-  };
+export interface ImmerUpdateEvent<TType extends string, TInput = any> {
```
Andarist, line 33: Is this used anywhere? I dont see much difference between this and the one above it. 

Also - should we default second type param to any? It makes it less type-safe by default

packages/xstate-immer/src/index.ts
```diff
@@ -16,37 +16,51 @@ export interface ImmerAssignAction<TContext, TEvent extends EventObject>
...
+  assign: ImmerAssignAction<TContext, TEvent>;
+  type: TEvent['type'];
+  validate?: (context: TContext, input: TEvent['input']) => boolean;
+}
 
-    return applyPatches(context, patches);
+export function createUpdater<TContext, TEvent extends ImmerUpdateEventObject>(
+  type: TEvent['type'],
+  producer: (ctx: Draft<TContext>, input: TEvent['input']) => void,
+  validate?: (ctx: TContext, input: TEvent['input']) => boolean
```
Andarist, line 48: Could u add an example/test for this? Not completely sure what its purpose

packages/xstate-immer/test/immer.test.ts
davidkpiano: Oops, accidental addition

packages/xstate-immer/test/immer.test.ts
davidkpiano: It's a lot less verbose than the alternative. I'm in favor of this for keeping it explicit, and for not limiting the use-cases - we might want there to be multiple actions, or custom guards, etc.

packages/xstate-immer/src/index.ts
davidkpiano: Yes, check the tests. It's meant to make defining these events simpler:

```ts
type SomeEvent = ImmerUpdateEvent<"UPDATE_NAME", string>;

// rather than...

type SomeEvent = { type: "UPDATE_NAME"; input: string }

// which is more mistake-prone
```

packages/xstate-immer/test/immer.test.ts
Andarist: the base API is good - we definitely dont want to limit use cases, I've just had in mind a helper API for producing those simple transitions which would cover most basic use cases. It's also easily implementable in the userland, so probably at the end of the day it's best to leave it to userland.

packages/xstate-immer/src/index.ts
Andarist: Ok, I see - it seems that the other one can just be removed. Take a look here: https://github.com/davidkpiano/xstate/pull/1130/files

packages/xstate-immer/src/index.ts
davidkpiano: It's been removed - it's not necessary (you can easily specify guards outside of `createUpdater`)

davidkpiano:
cc. @mweststrate Would love to get your brief thoughts on this!

packages/xstate-immer/README.md
```diff
@@ -1,3 +1,321 @@
...
+```
+
+### `createUpdater(eventType, recipe)`
+
+Returns an object that is useful for creating `context` updaters.
+
+| Argument    | Type     | Description                                                                       |
+| ----------- | -------- | --------------------------------------------------------------------------------- |
+| `eventType` | string   | The event type for the update event                                               |
+| `recipe`    | function | A function that takes in the `context` and some `input` to "mutate" the `context` |
```
Andarist, line 157: In the code this is called a `producer` - would be good to align the naming, probably requires a change in the code and not here, but it's up to you.

packages/xstate-immer/README.md
```diff
@@ -1,3 +1,321 @@
...
+// { count: 2, level: 9 }
+
+toggleService.send(levelUpdater.update(-100));
+// Notice how the level is not updated in 'inactive' state:
+// { count: 2, level: 9 }
+```
+
+## API
+
+### `assign(producer)`
```
Andarist, line 91: `producer` is described later as `recipe`

packages/xstate-immer/README.md
Andarist: Actually - I've just changed both of those names directly on your branch: https://github.com/davidkpiano/xstate/pull/1129/commits/5dc587a294260fb90fa7e17d6c2d6e62a9148baf

Feel free to adjust however you want

packages/xstate-immer/src/index.ts
```diff
@@ -2,51 +2,62 @@ import {
...
 
-export function assignPatch<
-  TContext,
-  TEvent extends PatchEventObject
->(): AssignAction<TContext, TEvent> {
-  return xstateAssign((context, event) => {
-    const [, patches] = event.patches;
+export function createUpdater<TContext, TEvent extends ImmerUpdateEvent>(
+  type: TEvent['type'],
+  recipe: (ctx: Draft<TContext>, input: TEvent['input']) => void
```
Andarist, line 47: wondering - shouldn't this have matching signature to `assign` and `immerAssign`?

davidkpiano: I thought that too but that didn't make sense initially - the created event only consists of the type and the `input` (which is essentially the event payload) itself.

Doing otherwise would force the user to always destructure:

```js
createUpdater('UPDATE_SOMETHING', (ctx, { input }) => {
 // ...
});
```

But now I see use-cases where having something besides the `input` might be beneficial.

packages/xstate-immer/src/index.ts
```diff
@@ -2,51 +2,62 @@ import {
...
 
-export function assign<TContext, TEvent extends EventObject = EventObject>(
-  assignment: ImmerAssigner<TContext, TEvent>
+function immerAssign<TContext, TEvent extends EventObject = EventObject>(
+  recipe: ImmerAssigner<TContext, TEvent>
 ): AssignAction<TContext, TEvent> {
-  // @ts-ignore (possibly infinite TS bug)
-  return xstateAssign((context, event) => {
-    return produce(context, (draft) => void assignment(draft, event));
+  return xstateAssign((context, event, meta) => {
```
mweststrate, line 24: Awesome stuff!

NIT: Pretty sure this could be written as `return xStateAssign(produce(recipe))`, as `produce` curries and passes any additional arguments in the same order

PR #1131, opened 2020-04-16
fix(core): export choose action

wKovacs64:


davidkpiano:
Sorry, this will be announced with docs tomorrow, so it'll be all fixed/documented by then. Thanks for doing this!

PR #1134, opened 2020-04-17
Add a test to prevent outer transitions from being executed when event in inner state is matched

CodingDive:
Hello beautiful people ❤️

I added a test and am opening this PR because I wanted to get some clarification around partial forbidden events.

What I'm trying to do is kind of hard to express so reading the test might be the easiest way to get a grasp on what I'm saying.

I ran into a scenario where I wanted to prevent parent state transitions with a `forbidden event`, yet handle the very same event within a child state node. I replicated my use case in the test below.

To my surprise, this is actually working. I'm not sure why or how since there are essentially two keys with the exact same name and to my knowledge, the last one always wins in JavaScript (`{ xstate: '=dank', xstate: '=love' } => { xstate: '=love' }`. In the test case below, this would mean that the `action` is executed and the `forbidden event` is simply ignored. The passing test indicates that this is not the case and the parent state transitions are successfully prevented. 

I absolutely love the fact that it's working, I'm just not sure if it's reliable and I'm supposed to use it. It is actually quite powerful in certain scenarios to forbid events only for parents up the tree while handling it in the same state node in some different way.

1. Should we use it? Is the way it works intended behavior?
2. If it is, should I move the test into a different test suite? It's touching state, context, assign and events so I figured I'll just create a new file.


packages/core/test/forbidden.test.ts
```diff
@@ -0,0 +1,77 @@
...
+          hidden: {
+            on: {
+              // We don't want the password field to be revealed based on a
+              // browser autocomplete firing an input `onChange` event.
+              [changePassword]: undefined,
+              // However, we do want to assign the value. Just prevent the
+              // parent transitions
+              [changePassword]: {
+                actions: 'assignPassword'
+              }
```
CodingDive, line 34: If we were to move the forbidden event underneath the assign action, `assignPassword` will never be called.

The ordering playing a role here makes me think that this is indeed an accidental feature and not intended. 😁

Andarist: The ordering play role here, because `{foo:1,foo:42}` is really `{foo:42}`. You can't have object with multiple properties with the same name after all.

I think the solution for your problem is just this:
```js
{
    [changePassword]: {
        // We don't want the password field to be revealed based on a
        // browser autocomplete firing an input `onChange` event.
        target: undefined,
        // However, we do want to assign the value. Just prevent the
        // parent transitions
        actions: 'assignPassword'
    },
}
```

Andarist: Ok, I've actually just re-read your issue carefully and I think I understand the confusion now. It indeed looks a little bit confusing at first and I would stick to the explicit version that I have written above. It's just not possible for multiple transitions to be taken from different nodes in the ancestor-descendant chain - only the most inner matched transition is taken. And you have matched this transition `[changePassword]: {actions: 'assignPassword'}` which doesn't have a target specified, so it's basically equivalent to what I have written above.

CodingDive: Thank you for the reply and clarification. For some reason, it didn't dawn on me that forbidden events can be achieved with `target: undefined` while still defining actions.
I must've been too caught up in my code-style of never using the `[event]: target` syntax that I couldn't make this connection. 😅

I think part of my confusion also came from the docs that say `forbidden_event: undefined` is equivalent to `actions: []`. I'll see if I can change the docs with my new learnings in the PR later.

Should I change my test to your snippet and include it? Might be useful to have a test like this as I couldn't see an existing one testing similar behavior.

If you want to include the test, should I also move it to a different file? Which test suite works best here? I'm thinking `actions` or `event` might work best.

Andarist: > I think part of my confusion also came from the docs that say forbidden_event: undefined is equivalent to actions: []. I'll see if I can change the docs with my new learnings in the PR later.

It's kinda true as all of:
- `forbidden_event: undefined`
- `forbidden_event: ''` (I think it also works, but I have not been using it)
- `forbidden_event: { actions: [] }`
- `forbidden_event: { target: undefined }`
- `forbidden_event: { target: undefined, actions: [] }`

are the same thing. After all empty actions array is like no actions at all.

> Should I change my test to your snippet and include it? Might be useful to have a test like this as I couldn't see an existing one testing similar behavior.

Probably yes, but not sure where it is the best place to put it right now. Try moving it where you think it would be best to put it and we'll iterate from that or just merge in.

CodingDive: > It's just not possible for multiple transitions to be taken from different nodes in the ancestor-descendant chain - only the most inner matched transition is taken.

This was the missing and crucial piece for my understanding. I tried to reflect this in the docs and the test too. 👍 

Thank you so much!

davidkpiano:
Yes, that is how it's supposed to work. The deeper transition is _selected_ over the parent one. For more information: https://www.w3.org/TR/scxml/#SelectingTransitions

Issue #1142, opened 2020-04-19
[TypeScript] State randomly becomes `never` type


CodingDive:
**Description**
I'm running into a weird type error with `state` randomly becoming the `never` type.

It doesn't seem to matter what I try: 

```TypeScript
const [state, send] = useMachine(someMachine); // passed the generic arguments to someMachine within the Machine() factory

// or 
const [state, send] = useMachine<Context, Event>(someMachine);

// or 
const [state, send] = useMachine<any, any>(someMachine);
```

After successfully using `state.context` or `state.matches` a couple of times, `state` suddenly gets into the `never` type and starts throwing errors randomly.

**Expected Result**
`state` to not lose its designated type.

**Actual Result**

Here you can see I can log the state while it has the correct type and also use `state.matches`. 

![image](https://user-images.githubusercontent.com/28842311/79677553-eeff5100-81f2-11ea-9f1c-c97a190bb7e2.png)

On the second `state.matches` however, the type seems to be `never`.

![image](https://user-images.githubusercontent.com/28842311/79677560-076f6b80-81f3-11ea-9d8d-ddd827b40116.png)

I then get the error `Property 'matches' does not exist on type 'never'.ts(2339)` for the rest of the file.

**Reproduction**

I couldn't reproduce this at the moment as it seems to randomly happen in only a few machines.

While I'll continue to try to reproduce the issue, does anyone know where this could be coming from? 

**Dependencies**

```
typescript: 3.8.3,
xstate: 4.9.1,
@xstate/react: 0.8.1
```

CodingDive:
Is it possible that `state.matches` somehow changes the `state` to never?
https://github.com/davidkpiano/xstate/blob/3dcf91861964f0e3c581b08fcb3357f658b75bea/packages/core/src/State.ts#L290

I seem to run into this issue right after using `state.matches`. Will try to reproduce this one more time soon :)

davidkpiano:
Can you please put this in a CodeSandbox?

CodingDive:
I was able to reproduce it in a CodeSandbox [here](https://codesandbox.io/s/never-state-xstate-i796v). Had to use `<any, any>` again to reproduce it, just like in the test. I also tried the latest `@xstate/react` rc.4 but am running into the same issue.

davidkpiano:
Workaround: use `createMachine` instead: https://codesandbox.io/s/never-state-xstate-ch5m3

That is what you should be using instead of `Machine`, moving forward.

PR #1143, opened 2020-04-19
Add failing test to reproduce never state issue of #1142

CodingDive:


packages/xstate-react/test/types.test.tsx
```diff
@@ -114,4 +110,18 @@ describe('useMachine', () => {
 
     render(<YesNo />);
   });
+
+  it('state should not become never after checking state with matches', () => {
+    const YesNo = () => {
+      const [state] = useMachine<any, any>(yesNoMachine);
```
davidkpiano, line 116: Can you remove the `<any, any>` generics? They're not necessary

CodingDive: It was the only way to reproduce the state becoming `never`. Everything seems to work fine in the test when removing the type arguments. I'm also not using `<any, any>` in my code but am running into the same problem.

packages/xstate-react/test/types.test.tsx
```diff
@@ -114,4 +110,18 @@ describe('useMachine', () => {
...
+
+      if (state.matches('no')) {
+        return <span>No</span>;
```
CodingDive, line 119: The problem seems to happen right when returning right after using `state.matches`. 

Replacing this line with `console.log(state.context.value)` works fine.

In fact, even the following code works fine. 😕 Maybe another TypeScript regression? 

```JavaScript
const isNo = state.matches('no');
if (isNo) {
  return <span>No</span>;
}
```

davidkpiano:
Can you redo this to use `createMachine` instead? `Machine` will be deprecated.

CodingDive:
This particular test is already using `createMachine`. Happened also with `Machine` in my code though.

Issue #1144, opened 2020-04-19
[4.9] stateGuard state for choose action creator


CodingDive:
Hey, I've just started playing with the new `choose` action creator. Amazing job @Andarist and @davidkpiano. I just saw the tests and was amazed to see even nested `choose` actions work. 🔥 

**Description**
It would be very nice to pass the whole `guardMeta` object, including the `state` to the conditional expressions of the `choose` action creator.

```JavaScript
entry: choose([
  {
    // state is undefined :(
    cond: (_, __, { state }) => state.matches('bar'),
    actions: assign<Ctx>({ answer: 42 })
  }
])
```

**Expected Result**

Passing the current state on top of `context` and `event`.

**Actual Result**

The third argument of the condition expression callback already exists.

When I log the stateGuard, this is what I get: 
```
cond: {
  type: "xstate.guard",
  name: "isPasswordResetMode",
  predicate: ƒ (context, event, stateGuard), // <-- how do I get to this?
}
```

However, `state` is undefined. 

**Reproduction**

I will send a PR with a failing test.

**Considerations**

I'm not sure if this is SCXML compatible but since any `choose` action can be translated to a regular state transition with normal guards, this shouldn't be a problem right?

**Additional context**
XState 4.9.1

davidkpiano:
Great that you're playing around with it! By the way, for your specific example, there might be a better way to model it. You shouldn't use `state.matches()` in a guard function - that function should be part of the matched state naturally.

CodingDive:
@davidkpiano I very rarely have the need to use `state.matches` in a guard. When I do, it's typically to check the machine state of orthogonal nodes. Is there a better way to do this?

___ 

Closing this issue. `stateGuard.state` is not available during the initialization of the machine. It should work right after. See the discussion [here](https://github.com/davidkpiano/xstate/pull/1145).

davidkpiano:
> Is there a better way to do this?

Yes, using `in`: https://xstate.js.org/docs/guides/guards.html#in-state-guards

Remodeling might also help.

FadySalama:
> > Is there a better way to do this?
> 
> Yes, using `in`: https://xstate.js.org/docs/guides/guards.html#in-state-guards
> 
> Remodeling might also help.

Is it possible to us the `in` predicate in `choose` action? If not, would it be possible to add it as a feature?

davidkpiano:
@FadySalama Can you ask this in a discussion? github.com/statelyai/xstate/discussions/new

PR #1145, opened 2020-04-19
Add test for accessing stateGuard.state in a conditional expression of the choose action creator

CodingDive:
Reproduces #1144

### DIFF
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -1237,4 +1237,29 @@ describe('choose', () => {
     service.send({ type: 'NEXT', counter: 101 });
     expect(service.state.context).toEqual({ answer: 42 });
   });
+
+  it('should provide stateGuard.state to a condition expression', () => {
+    type Ctx = { counter: number; answer?: number };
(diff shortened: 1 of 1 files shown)


Andarist:
Are you sure this doesn't just work for actions defined on the initialization path? `choose` uses the same `evaluateGuard` as conditional transitions and it seems that this should be provided:
https://github.com/davidkpiano/xstate/blob/401413e30110c73cd2b252b41c4f5488ae0f685d/packages/core/src/utils.ts#L681
unless we are talking about `choose` actions executed before the first `state` object has a chance to get created.

CodingDive:
I see. I was just testing the init path I believe. I didn't know the
machine is missing state when initizializing but it makes sense that the
first state is only available after all the transient transitions have been
resolved and the entry actions executed.

I'll take a look at this tonight and will change the test accordingly.

CodingDive:
Yes, you are right. The test is now passing after checking the non-initialization path. Thank you! 😊

davidkpiano:
Well, more tests can't hurt ➕

Issue #1155, opened 2020-04-22
Delayed transition should keep the event that originally caused the transition


CodingDive:
**Description**
I think this is more a question than a bug report. Is it possible to keep the original event that triggered the delayed transition?

```TypeScript
type Filter = { type: 'filter', input: string, nonContextVariable: 2};

type Context = { search: string };


// ....
states: {
  idle: {
    on: {
      filter: { target: 'debouncing', actions: assign<Context, Filter>({ search: (_, event) => event.input }) },
    }
  },
  debouncing: { after: { 500: 'filtering' } },
  filtering: {
    // when using a service or entry actions here, the event has nothing but the XState internal delayed event type
  },
},
```

Is there a way to possibly access the `nonContextVariable` of the event inside the `filtering` state node? 

**Expected Result**

I expected a delayed transition to behave similarly to a transient transition. Conceptually, I think about the `after` property as a transient transition with some extra `timeout`. It could possibly append some meta-information about the delay on the event, or even change the type but ideally preserve the rest of the event data. 

```JavaScript
// original event
{ type: 'filter', input: string, nonContextVariable: 2},
// should be changed to maybe 
{ type: 'filter', input: string, nonContextVariable: 2, delayed: 500},
// , or possibly 
{ type: 'xstate.after(500)#root.debouncing', input: string, nonContextVariable: 2},
```

**Actual Result**

Every value that I don't immediately assign to context is lost because the event just has the XState internal delay type:

```JavaScript
{ type: 'xstate.after(500)#root.debouncing' },
```

**Reproduction**

https://codesandbox.io/s/xstate-delayed-transition-no-event-o5hmt?file=/src/App.js

**Additional context**

I can probably work around it by defining at least two more events and refactoring the `delayed transition` into a delayed event, or by assigning more values into context than I'd normally like to. I wanted to see if there is a way to achieve this with just delayed transitions as it is by far the cleanest solution.

davidkpiano:
Yeah, it should be pretty trivial to put the original event on the special `xstate.after` event.

davidkpiano:
Marking this as "working as designed" since we shouldn't change the way that clock-based events work; that is, we shouldn't make them "magical" containers of previous events. 

The right way to do this is to `assign(...)` the data to context and access that in the `filtering` state, or to re-send the event to self with a `delay` (instead of using `after`).

PR #1156, opened 2020-04-22
Add failing test for keeping the event data after a delayed transition

CodingDive:
See #1155.

> It should be pretty trivial to put the original event on the special xstate.after event.

Based on the comment, I tried to fix the test myself for a while but I'm not yet deep enough in the codebase to understand all the things that are going on inside the `StateNode.ts` file. 😅 I'm afraid, the best thing I can do at the moment is:

image


Andarist:
Closing based on https://github.com/davidkpiano/xstate/issues/1155#issuecomment-778832957

PR #1160, opened 2020-04-26
Fix toJSON for transitions

davidkpiano:
This PR fixes `.toJSON()` for machines containing delayed transitions defined in `after` (and elsewhere).

### DIFF
--- a/.changeset/grumpy-deers-applaud.md
+++ b/.changeset/grumpy-deers-applaud.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Delayed transitions defined using `after` were previously causing a circular dependency when the machine was converted `.toJSON()`. This has now been fixed.
--- a/.changeset/grumpy-deers-applaud.md
+++ b/.changeset/grumpy-deers-applaud.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Delayed transitions defined using `after` were previously causing a circular dependency when the machine was converted `.toJSON()`. This has now been fixed.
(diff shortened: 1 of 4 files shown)


.changeset/grumpy-deers-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Delayed transitions defined using `after` were previously causing a circular dependency when the machine was converted `.toJSON()`. This has now been fixed.
```
Andarist, line 5:
```suggestion
Delayed transitions defined using `after` were previously causing a circular dependency when the machine was converted using `.toJSON()`. This has now been fixed.
```

not sure if this fixes grammar or not, feel free to ignore the suggestion - you are the native English speaker so you know best 😅

packages/core/src/StateNode.ts
```diff
@@ -1800,18 +1800,15 @@ class StateNode<
       target,
       source: this,
       internal,
-      eventType: transitionConfig.event
-    };
-
-    Object.defineProperty(transition, 'toJSON', {
-      value: () => ({
+      eventType: transitionConfig.event,
+      toJSON: () => ({
```
Andarist, line 1804: Hm, could you elaborate on what was the issue on the lower level? how changing the enumerability of this property has fixed it?

davidkpiano: Because of this:

```js
return toArray(resolvedTransition).map((transition) => ({
  ...transition,
  event: eventType,
  delay: resolvedDelay
}));
```

The `...transition` wasn't taking the `toJSON` property.

PR #1163, opened 2020-04-29
Add ActorRef

davidkpiano:
This is a substantial PR that refactors usage of the actor model to more closely mirror systems like Akka, by introducing the `ActorRef`, which is an interface to an actor that you can `.send(...)` events to.

Todo:
- [ ] Documentation
- [ ] Type docs
- [x] Changeset


Andarist:
This might take a moment to digest everything - but I will start looking into this later today.

packages/core/src/behavior.ts
```diff
@@ -0,0 +1,360 @@
...
+   */
+  initial: TEmitted;
+  subscribe?: (observer: Observer<TEmitted>) => Unsubscribable | undefined;
+}
+
+export function createCallbackBehavior<TEvent extends EventObject>(
+  callback: InvokeCallback,
+  parent?: ActorRef<any>
+): Behavior<TEvent, undefined> {
+  let canceled = false;
```
Andarist, line 61: isnt this against what behaviors should be? what i mean - from what i understand a behavior should describe how an actor should behave in a given "state" (which is not defined explicitly, but rather through behavior itself can be seen as state) and in general it should be possible to just restart actors by reusing initial behaviors, so it should merely encode the desired behavior, but should not contain a state of its own. This breaks this promise because this behavior has a mutable state in a closure

packages/core/src/behavior.ts
```diff
@@ -0,0 +1,360 @@
...
+                })
+              );
+              canceled = true;
+            },
+            (errorData) => {
+              const errorEvent = error(actorContext.name, errorData);
+              parent?.send(
+                toSCXMLEvent(errorEvent, { origin: actorContext.self })
+              );
+              // TODO: handle error
```
Andarist, line 102: what kind of error is possible here? you already notify the "caller" actor through an event, i don't think there is particularly much else we could do here

packages/core/src/interpreter.ts
```diff
@@ -931,277 +879,36 @@ export class Interpreter<
       child.stop();
     }
   }
-  public spawn(entity: Spawnable, name: string, options?: SpawnOptions): Actor {
+  public spawn(
```
Andarist, line 882: This I'm not a particular fan of. I don't think we should encode support for different Spawnables in the Interpreter. It goes against composition principles and is not tree-shakeable.

packages/core/src/types.ts
```diff
@@ -1045,26 +1052,38 @@ export declare namespace SCXML {
...
+
+  // Sends the sequence error
+  error?: (errorValue: any) => void;
+
+  // Sends the completion notification
+  complete: () => void;
+}
+
 export interface Subscribable<T> {
+  subscribe(observer: Observer<T>): Unsubscribable | undefined;
```
Andarist, line 1078: both of those subscribes define different things as required and optional, in the first one only `complete` is required, whereas in the second one only `next` is required, I believe both should be aligned to be the same

packages/core/test/actor.test.ts
```diff
@@ -766,18 +790,22 @@ describe('actors', () => {
           states: {
             same: {
               entry: assign({
-                ref: () => spawn(syncChildMachine, falseSyncOption)
+                ref: (_, __, { self, spawn }) =>
+                  spawn(
+                    createMachineBehavior(
+                      syncChildMachine,
+                      self,
```
Andarist, line 797: Having to pass `self` here seems like a problem - it should be enforced by the system, not manual because it is actually enforced by the Actor Model (everything forms a tree hierarchy). Not sure how to fix this right now though.

Andarist:
We could use a "design goals" summary here, terminology explainer (actor, actor ref, behavior) and a comparison to existing Actor Model implementations (limited to Akka for now). This would make both reviewing and discussing the whole thing easier and would be a good thing to include in the docs anyway.

packages/core/test/actor.test.ts
davidkpiano: This is fixed by using `spawn.from(syncChildMachine)`

packages/xstate-react/src/index.ts
```diff
@@ -147,3 +148,19 @@ export function useService<
...
   return [state, service.send, service];
 }
+
+export function useActor<TEvent extends EventObject, TEmitted>(
+  actorRef: ActorRef<TEvent, TEmitted>
+): [TEmitted, (event: TEvent) => void] {
+  const [state, setState] = useState(actorRef.initial);
+
+  useEffect(() => {
+    const sub = actorRef.subscribe((nextState) => {
```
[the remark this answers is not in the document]
davidkpiano: Yes, enforced now

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+export class BehaviorActorRef<TEvent extends EventObject, TEmitted>
+  implements ActorRef<TEvent, TEmitted> {
+  public ref;
+  public initial: TEmitted;
+  private context: ActorContext;
+
+  constructor(public behavior: Behavior<TEvent, TEmitted>, public id: string) {
+    this.context = {
+      self: this,
+      name: this.id
```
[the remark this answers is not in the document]
davidkpiano: In Akka, neither are public. I'll make this change.

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+export type Sender<TEvent extends EventObject> = (event: TEvent) => void;
 
-export interface Actor<
-  TContext = any,
-  TEvent extends EventObject = AnyEventObject
-> extends Subscribable<TContext> {
+export interface ActorRef<TEvent extends EventObject, TEmitted = any>
+  extends Subscribable<TEmitted> {
+  send: Sender<TEvent>;
+  ref: any;
```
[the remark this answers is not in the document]
davidkpiano: Removed

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+  createServiceBehavior,
+  createMachineBehavior,
+  createCallbackBehavior,
+  createPromiseBehavior,
+  createObservableBehavior
+} from './behavior';
+import { registry } from './registry';
+import { State } from './State';
+
+export type Sender<TEvent extends EventObject> = (event: TEvent) => void;
```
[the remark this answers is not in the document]
davidkpiano: It's also in ActorLike and InvokeCallback

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+}
+
+export function fromService<TContext, TEvent extends EventObject>(
+  service: Interpreter<TContext, any, TEvent>,
+  id: string = registry.bookId()
+): ActorRef<TEvent> {
+  return new BehaviorActorRef(createServiceBehavior(service), id);
+}
+
+export class BehaviorActorRef<TEvent extends EventObject, TEmitted>
```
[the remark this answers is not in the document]
davidkpiano: (discussed offline)

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+    return this;
+  }
+  public stop() {
+    this.behavior = this.behavior.receiveSignal(this.context, stopSignal);
+  }
+  public subscribe(observer) {
+    return this.behavior.subscribe?.(observer);
+  }
+  public send(event) {
+    this.behavior = this.behavior.receive(this.context, event);
```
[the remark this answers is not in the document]
davidkpiano: In Akka (ported to idiomatic JS) you would return a `Behaviors.Same` symbol. This is something I want to explore in a future PR.

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+    this.behavior = this.behavior.receiveSignal(this.context, startSignal);
+    return this;
+  }
+  public stop() {
+    this.behavior = this.behavior.receiveSignal(this.context, stopSignal);
+  }
+  public subscribe(observer) {
+    return this.behavior.subscribe?.(observer);
+  }
+  public send(event) {
```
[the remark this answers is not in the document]
davidkpiano: Let's add mailboxes in a future PR, it's out of scope for this one.

Right now, `Interpreter` essentially has its own "mailbox" (`Scheduler`) that avoids this issue.

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+  }
+  public start() {
+    this.behavior = this.behavior.receiveSignal(this.context, startSignal);
+    return this;
+  }
+  public stop() {
+    this.behavior = this.behavior.receiveSignal(this.context, stopSignal);
+  }
+  public subscribe(observer) {
+    return this.behavior.subscribe?.(observer);
```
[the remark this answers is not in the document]
davidkpiano: For now, yes. Will have to dive more into Akka Streams to compare how to better handle this.

packages/core/test/actor.test.ts
davidkpiano: The reason this isn't done yet is because I still need to augment `createMachineBehavior`'s signature to include options

packages/core/src/interpreter.ts
davidkpiano: Good news: we can delete the entire thing!

packages/core/src/behavior.ts
davidkpiano: Behaviors (even in Akka) can use closures. This can easily be represented as pure functions too - it's a trivial refactor.

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+  public initial: TEmitted;
+  private context: ActorContext;
+
+  constructor(public behavior: Behavior<TEvent, TEmitted>, public id: string) {
+    this.context = {
+      self: this,
+      name: this.id
+    };
+    this.ref = behavior;
+    this.initial = behavior.initial;
```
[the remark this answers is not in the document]
davidkpiano: Changed `initial` to `current`

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+import { registry } from './registry';
+import { State } from './State';
+
+export type Sender<TEvent extends EventObject> = (event: TEvent) => void;
 
-export interface Actor<
-  TContext = any,
-  TEvent extends EventObject = AnyEventObject
-> extends Subscribable<TContext> {
+export interface ActorRef<TEvent extends EventObject, TEmitted = any>
```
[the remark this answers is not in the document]
davidkpiano: Not necessarily; e.g., if a promise resolves with `number`, we might want to emit a `number` but send an `EventObject` to the parent with the number.

Still not sure if we should "convert" promises/observables or what that would look like. I think `TEmitted` should not be used as a mechanism for sending an event.

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
-
-  return tempActor;
-}
-
-export function isActor(item: any): item is Actor {
-  try {
-    return typeof item.send === 'function';
-  } catch (e) {
-    return false;
+      this.processingStatus = ProcessingStatus.Processing;
```
Andarist, line 121: nit: setting `processingStatus` could be moved outside of this loop, it would imho illustrate the intent slightly better

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+  public name: string;
+
+  constructor(behavior: Behavior<TEvent, TEmitted>, name: string) {
+    this.behavior = behavior;
+    this.name = name;
+    this.context = {
+      self: this,
+      name: this.name
+    };
+    this.current = behavior.current;
```
Andarist, line 99: this is only set in the constructor - imho we should either:
1. rename it
2. remove it
3. update after setting a new behavior

packages/core/src/Actor.ts
```diff
@@ -1,59 +1,126 @@
...
+  Processing
+}
+
+export class ObservableActorRef<TEvent extends EventObject, TEmitted>
+  implements ActorRef<TEvent, TEmitted> {
+  public current: TEmitted;
+  private context: ActorContext;
+  private behavior: Behavior<TEvent, TEmitted>;
+  private mailbox: TEvent[] = [];
+  private processingStatus: ProcessingStatus = ProcessingStatus.NotProcessing;
```
Andarist, line 89: this seems like a simple boolean, of course enum works too but it's sort of a signal to me as a reader that this might have more than 2 statuses

packages/core/src/MachineNode.ts
```diff
@@ -251,7 +259,19 @@ export class MachineNode<
    */
   public get initialState(): State<TContext, TEvent, TStateSchema, TTypestate> {
     this._init();
-    const nextState = getInitialState(this);
+    const nextState = resolveMicroTransition(this, [], undefined, undefined);
+    return macrostep(nextState, null as any, this);
+  }
+
+  public getInitialState(service?: ActorRef<TEvent>) {
```
Andarist, line 266: what's the story around this? do you plan to remove `.initialState`? how the behavior of this one is going to differ when `service` is given to it and when it is not?

packages/core/src/Actor.ts
Andarist: seems that maybe the intention was the 3rd option (?): https://github.com/davidkpiano/xstate/pull/1163/files#diff-aebc63c0de70d68a4924833fd0ea3a5fR50

davidkpiano: It shouldn't exist at all. We should move this out of the `Actor`/`ActorRef` and instead create an `ActorStream` (bikesheddable) that wraps the actor and caches the last emitted value to make available as `current`.

packages/core/src/Actor.ts
davidkpiano: And it will! In the future, we would want to handle cases where the mailbox is full/an error happened, etc.

packages/core/src/MachineNode.ts
davidkpiano: For initial states with spawned actors, `machine.initialState` will have actors in context that are not started and have no parent. Same with `machine.getInitialState()`.

With `machine.getInitialState(service)`, the actors will not be started (until the service starts them) but will have a parent. That's the distinction.

packages/core/src/behavior.ts
Andarist: what i meant is that i was under the impression that behaviors should not use mutable state (they can use closures to capture some state, but they cant write to it), but maybe I was wrong - gonna check this out later, doesn't matter much for this PR

davidkpiano: I can implement this differently, sure.

packages/core/src/behavior.ts
```diff
@@ -0,0 +1,359 @@
...
+              if (!canceled) {
+                const errorEvent = error(actorContext.name, errorData);
+
+                parent?.send(
+                  toSCXMLEvent(errorEvent, { origin: actorContext.self })
+                );
+
+                observers.forEach((observer) => {
+                  observer.error?.(errorData);
+                  observer.complete?.();
```
Andarist, line 165: this doesnt comply to common observable implementations, error and complete callbacks are mutually exclusive (only one of them can be called), see an example here: https://codesandbox.io/s/nice-water-7dfhu?file=/src/index.js

packages/core/src/interpreter.ts
```diff
@@ -357,8 +309,8 @@ export class Interpreter<
     if (typeof nextListenerOrObserver === 'function') {
       listener = nextListenerOrObserver;
     } else {
-      listener = nextListenerOrObserver.next.bind(nextListenerOrObserver);
-      resolvedCompleteListener = nextListenerOrObserver.complete.bind(
+      listener = nextListenerOrObserver.next!.bind(nextListenerOrObserver);
```
Andarist, line 312: Observer type has [`next` defined as optional](https://github.com/davidkpiano/xstate/blob/13918f4b3a49d5917fa1b682615d62f63fc57808/packages/core/src/types.ts#L1066), either it should become required or this line here should use `?` rather than `!`

packages/core/src/types.ts
```diff
@@ -543,19 +534,14 @@ export type DelayFunctionMap<TContext, TEvent extends EventObject> = Record<
...
 export type DelayConfig<TContext, TEvent extends EventObject> =
   | number
   | DelayExpr<TContext, TEvent>;
 
 export interface MachineOptions<TContext, TEvent extends EventObject> {
   guards: Record<string, ConditionPredicate<TContext, TEvent>>;
   actions: ActionFunctionMap<TContext, TEvent>;
-  activities: Record<string, ActivityConfig<TContext, TEvent>>;
-  services: Record<string, InvokeCreator<TContext, TEvent>>;
+  services: Record<string, BehaviorCreator<TContext, TEvent>>;
```
Andarist, line 544: we should probably rename this option to behaviors now

packages/core/src/Actor.ts
Andarist: ok, let's see how this plays out 👍

packages/core/src/Actor.ts
Andarist: k, definitely a potential for other PRs

packages/core/test/actor.test.ts
davidkpiano: I'm going to keep this as-is because in V5, we want to encourage developers to be explicit about events sent/received between machines. `autoForward` and `sync` are implicit, and furthermore, `autoForward` should be preferred for invoked machines instead.

However, we should keep this advanced option in the tests (not documented) if this behavior is desired.

packages/core/test/scxml.test.ts
```diff
@@ -172,7 +172,7 @@ const testGroups = {
     'test200.txml',
     'test201.txml',
     'test205.txml',
-    'test207.txml',
+    // 'test207.txml', // delayexpr
```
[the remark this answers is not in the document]
davidkpiano: This is failing for some reason, and I've nailed it down to delayexpr, spent too much time on it though. Will investigate in future.

PR #1165, opened 2020-04-29
Add warning + docs for .onDone in root nodes.

### LINKED ISSUE #1111: onDone not triggered for parallel states
**Description**
I was studying parallel states and more specifically trying out the example in the documentation (https://xstate.js.org/docs/guides/final.html#parallel-states). The `onDone` function does not get fired when children states get into the final state.

This issue might be related to https://github.com/davidkpiano/xstate/issues/335?

**Expected Result**
As the example in the documentation, `onDone` should be fired when the children get into the final state.

**Actual Result**
`onDone` is not fired when the children get into the final state.

**Reproduction**
https://codesandbox.io/s/nifty-ganguly-8bnei

**Additional context**
Latest version of xstate (4.8.0)

davidkpiano:
Closes #1111

### DIFF
--- a/.changeset/wicked-ducks-mate.md
+++ b/.changeset/wicked-ducks-mate.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-XState will now warn if you define an `.onDone` transition on the root node. Root nodes which are "done" represent the machine being in its final state, and can no longer accept any events. #1111
+XState will now warn if you define an `.onDone` transition on the root node. Root nodes which are "done" represent the machine being in its final state, and can no longer accept any events. This has been reported as confusing in [#1111](https://github.com/davidkpiano/xstate/issues/1111).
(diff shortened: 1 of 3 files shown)


.changeset/wicked-ducks-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+XState will now warn if you define an `.onDone` transition on the root node. Root nodes which are "done" represent the machine being in its final state, and can no longer accept any events. #1111
```
Andarist, line 5:
```suggestion
XState will now warn if you define an `.onDone` transition on the root node. Root nodes which are "done" represent the machine being in its final state, and can no longer accept any events. This has been reported as confusing in [#1111](https://github.com/davidkpiano/xstate/issues/1111).
```

PR #1169, opened 2020-05-01
Svelte usage recipe

sveltemaster:
Please criticize! 
I want to make sure it makes sense and is also valid. 
(It works with this simple example on my machine.. I'll be incorporating this into a bigger project soon.)

### DIFF
--- a/docs/recipes/svelte.md
+++ b/docs/recipes/svelte.md
@@ -51,6 +51,7 @@ export const toggleMachine = Machine({
     {current.matches('inactive') ? 'Off' : 'On'}
 </button>
 ```
+
 ### App.svelte - Store usage
 The toggleService has a .subscribe function that is similar to svelte stores, so it can be used as a readable store!
 ```html
--- a/docs/recipes/svelte.md
+++ b/docs/recipes/svelte.md
@@ -1,6 +1,6 @@
 # Usage with Svelte
 
-Xstate integrates great with Svelte. (Especially Svelte stores!)
+XState integrates great with Svelte, and especially [Svelte stores!](https://svelte.dev/docs#svelte_store)
 
 The only gotcha is that you'll need this extra package.
 ```js
--- a/docs/recipes/svelte.md
+++ b/docs/recipes/svelte.md
@@ -1,0 +1,68 @@
+# Usage with Svelte
+
+Xstate integrates great with Svelte. (Especially Svelte stores!)
+
+The only gotcha is that you'll need this extra package.
+```js
+npm install @rollup/plugin-replace --save-dev
+```
+Then add this to the plugins array in rollup.config.js.
+```js
+replace({ 
+    'process.browser': true, 
+    'process.env.NODE_ENV': process.env.NODE_ENV
+})
+```
+
+### machine.js
+```js
+import { Machine } from 'xstate';
+
+// This machine is completely decoupled from Svelte
+export const toggleMachine = Machine({
+  id: 'toggle',
+  initial: 'inactive',
+  states: {
+    inactive: {
+      on: { TOGGLE: 'active' }
+    },
+    active: {
+      on: { TOGGLE: 'inactive' }
+    }
+  }
+});
+```
+
+### App.svelte - Standard usage
+```html
+<script>
+    import {interpret} from 'xstate';
+    import {toggleMachine} from './machine';
+    
+    let current;
+    
+    const toggleService = interpret(toggleMachine)
+        .onTransition((state) => {
+            current = state;
+        }).start()
+</script>
+
+<button on:click={() => toggleService.send('TOGGLE')}>
+    {current.matches('inactive') ? 'Off' : 'On'}
+</button>
+```
+### App.svelte - Store usage
+The toggleService has a .subscribe function that is similar to svelte stores, so it can be used as a readable store!
+```html
+<script>
+    import {interpret} from 'xstate';
+    import {toggleMachine} from './machine';
+        
+    const toggleService = interpret(toggleMachine).start();
+</script>
+
+<button on:click={() => toggleService.send('TOGGLE')}>
+    {$toggleService.matches('inactive') ? 'Off' : 'On'}
+</button>
+```
+If you're not familiar with the '$' syntax, it basically just reads the value of a store.
--- a/docs/recipes/svelte.md
+++ b/docs/recipes/svelte.md
@@ -9,7 +9,7 @@ npm install @rollup/plugin-replace --save-dev
 Then add this to the plugins array in rollup.config.js.
 ```js
 replace({ 
-    'process.browser': true
+    'process.env.NODE_ENV': process.env.NODE_ENV
 })
 ```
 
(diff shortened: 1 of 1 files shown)


docs/recipes/svelte.md
```diff
@@ -0,0 +1,69 @@
+# Usage with Svelte
+
+Xstate integrates great with Svelte. (Especially Svelte stores!)
```
davidkpiano, line 3:
```suggestion
XState integrates great with Svelte, and especially [Svelte stores!](https://svelte.dev/docs#svelte_store)
```

docs/recipes/svelte.md
```diff
@@ -0,0 +1,69 @@
...
+```js
+replace({ 
+    'process.browser': true, 
+    'process.env.NODE_ENV': process.env.NODE_ENV
+})
+```
+
+### machine.js
+```js
+import { Machine } from 'xstate';
```
davidkpiano, line 19:
```suggestion
import { createMachine } from 'xstate';
```

docs/recipes/svelte.md
```diff
@@ -0,0 +1,69 @@
...
+
+// This machine is completely decoupled from Svelte
+export const toggleMachine = Machine({
```
davidkpiano, line 22:
```suggestion
export const toggleMachine = createMachine({
```

docs/recipes/svelte.md
```diff
@@ -0,0 +1,69 @@
...
+        }).start()
+</script>
+
+<button on:click={() => toggleService.send('TOGGLE')}>
+    {current.matches('inactive') ? 'Off' : 'On'}
+</button>
+```
+
+### App.svelte - Store usage
+The toggleService has a .subscribe function that is similar to svelte stores, so it can be used as a readable store!
```
davidkpiano, line 56:
```suggestion
The toggleService has a `.subscribe` function that is similar to Svelte stores, so it can be used as a readable store.
```

davidkpiano:
Looks like a great start! Do you have a CodeSandbox by any chance where we can demonstrate Svelte?

docs/recipes/svelte.md
```diff
@@ -0,0 +1,69 @@
...
+Xstate integrates great with Svelte. (Especially Svelte stores!)
+
+The only gotcha is that you'll need this extra package.
+```js
+npm install @rollup/plugin-replace --save-dev
+```
+Then add this to the plugins array in rollup.config.js.
+```js
+replace({ 
+    'process.browser': true,
```
Andarist, line 12: is this one needed? I don't think we use `process.browser` anywhere

tivac: https://svelte.dev/repl/e36fa239641a49509ca1d06bcebf9f36?version=3.21.0

Certainly complains about it in the svelte repl.

Andarist: It complains about `process` being undefined but this happens because `process.env.NODE_ENV` is left (on purpose) in the distributed files. My question was purely about `process.browser` though - I'm not aware of such a thing being used by us.

sveltemaster: oh yeah should add in here that i use npm i xstate to install. 
With the CDN tag, i don't need that plugin.

Hmm I will try to look into why this is happening. I was told this error has to do with a package meant for node being used on the client

sveltemaster:
> Looks like a great start! Do you have a CodeSandbox by any chance where we can demonstrate Svelte?
I was having trouble setting it up in codesandbox and the Svelte REPL.. 
https://codesandbox.io/s/nervous-yonath-62wit?file=/App.svelte
(Tivac's REPL)
https://svelte.dev/repl/e36fa239641a49509ca1d06bcebf9f36?version=3.21.0
(With CDN tag)
https://svelte.dev/repl/72980fc082be479cb257d48088b23ac3?version=3.21.0

docs/recipes/svelte.md
sveltemaster: what's the difference?

docs/recipes/svelte.md
```diff
@@ -0,0 +1,68 @@
...
+XState integrates great with Svelte, and especially [Svelte stores!](https://svelte.dev/docs#svelte_store)
+
+The only gotcha is that you'll need this extra package.
+```js
+npm install @rollup/plugin-replace --save-dev
+```
+Then add this to the plugins array in rollup.config.js.
+```js
+replace({ 
+    'process.browser': true
```
Andarist, line 12: Is this needed though? 😉 XState doesn't contain such a thing in its source code

sveltemaster: i don't *really* understand what it's doing but it fixes this error and makes it work
environment.js:1 Uncaught ReferenceError: process is not defined

Andarist: It replaces `process.browser` in the source code with `true` - it's just a simple string to string replacement. Our `environment.js` file doesn't contain this string though - it only includes `process.env.NODE_ENV`:
https://unpkg.com/xstate@4.9.1/lib/environment.js

davidkpiano: If it doesn't work without it, we can keep it here for now and investigate later.

Andarist: I'm just worried that this is ill-advised and thus will be confusing for people. To the best of my knowledge - this is not required. The problem was about `process.env.NODE_ENV` not `process.browser`, but handling the former has been removed from the suggestion altogether. I suspect the whole initial suggestion (containing both `process.env.NODE_ENV` and `process.browser`) was copied from somewhere and hasn't been adjusted to the actual needs of XState.

If this really doesn't work cause of `process.browser` being in the source code I would appreciate seeing a repro case. The provided (@tivac) repro is failing because of `process.env.NODE_ENV` being in the bundle (which is supposed to be there as we want to differentiate dev & prod code paths):
Screenshot 2020-05-03 at 00 31 48
So from my perspective, the part that has been removed (about the need to replace `process.env.NODE_ENV`) made much more sense to be included than what has been left as part of this PR.

sveltemaster: @Andarist thank you for pushing back on this, you are totally right.

docs/recipes/svelte.md
```diff
@@ -0,0 +1,68 @@
...
+XState integrates great with Svelte, and especially [Svelte stores!](https://svelte.dev/docs#svelte_store)
+
+The only gotcha is that you'll need this extra package.
+```js
+npm install @rollup/plugin-replace --save-dev
+```
+Then add this to the plugins array in rollup.config.js.
+```js
+replace({ 
+    'process.env.NODE_ENV': process.env.NODE_ENV
```
Andarist, line 12: just an additional note: this would only work if `NODE_ENV` is set for your env, is it a common practice for it to be set when working with Svelte? an alternative for this would be:
```js
replace({ 
  'process.env.NODE_ENV': JSON.stringify('production') // or JSON.stringify('development')
})
```

davidkpiano:
I'll add this in for now; we can do a follow-up PR on addressing `process.env.NODE_ENV` for now, but it's harmless to include it here.

PR #1170, opened 2020-05-01
Update rollup config for esm build

chenxsan:
I was trying to use the es module of xstate in my `index.mjs` file without any bundler:

```js
import {
  createMachine,
  interpret,
} from "https://unpkg.com/xstate@4.9.1/es/index.js";

```
And I got this error:

image

image

Because `process` does not exist in browser.

Enable `rollup-plugin-replace` for rollup would fix the `esm` build.

### DIFF
--- a/packages/core/rollup.config.js
+++ b/packages/core/rollup.config.js
@@ -18,7 +18,12 @@ const createNpmConfig = ({ input, output }) => ({
   input,
   output,
   preserveModules: true,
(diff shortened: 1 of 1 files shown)


Andarist:
The file you are trying to consume is intended for bundlers. If you want to use browser-compatible file then please use this one: https://unpkg.com/browse/xstate@4.9.1/dist/xstate.web.js

PR #1183, opened 2020-05-10
Ensure actions are executed from a restored initial state.

### LINKED ISSUE #1174: Warning "No implementation found for action type" shown when starting machine from persisted state
**Description**
The following code persists a state then recreate it. It then starts an interpreter based on the saved state. The following warning is shown in the output: 

Warning: No implementation found for action type 'report'

**Expected Result**
No warning

**Actual Result**
Warning: No implementation found for action type 'report'

**Reproduction**
```
import { Machine, interpret, State } from 'xstate'

const lightMachine = Machine({
  id: 'light',
  initial: 'green',
  states: {
    green: {
      on: {
        TIMER: {
          target: 'yellow',
          actions: 'report'
        }
      }
    },
    yellow: {
      on: {
        TIMER: {
          target: 'red',
          actions: 'report'
        }
      }
    },
    red: {
      on: {
        TIMER: 'green'
      }
    }
  }
}, {
  actions: {
    report: (context, event) => console.log('reporting function')
  }
});

const currentState = 'green'
const nextState = lightMachine.transition(currentState, 'TIMER')

// saves state and recreate it
const recreated = JSON.parse(JSON.stringify(nextState))
const previousState = State.create(recreated)
const resolvedState = lightMachine.resolveState(previousState)
// console.log(resolvedState)

const service = interpret(lightMachine)
  .onTransition(state => {
    console.log('state transition', state.value, state.changed)
  })
  .start(resolvedState)

service.send('TIMER')
```

davidkpiano:
The issue was that when restoring from an initial state, `state.actions` weren't being resolved against the specified `machine.options.actions` action mapping, so actions weren't being executed.

Fixes #1174

### DIFF
--- a/.changeset/wet-swans-deny.md
+++ b/.changeset/wet-swans-deny.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Actions from a restored state provided as a custom initial state to `interpret(machine).start(initialState)` are now executed properly. See #1174 for more information.
--- a/packages/core/test/interpreter.test.ts
+++ b/packages/core/test/interpreter.test.ts
@@ -118,6 +118,56 @@ describe('interpreter', () => {
         done();
       }, 100);
     });
+
+    // https://github.com/davidkpiano/xstate/issues/1174
+    it('executes actions from a restored initial state', (done) => {
+      const lightMachine = Machine(
+        {
+          id: 'light',
+          initial: 'green',
+          states: {
+            green: {
+              on: {
+                TIMER: {
+                  target: 'yellow',
+                  actions: 'report'
+                }
+              }
+            },
+            yellow: {
+              on: {
+                TIMER: {
+                  target: 'red'
+                }
+              }
+            },
+            red: {
+              on: {
+                TIMER: 'green'
+              }
+            }
+          }
+        },
+        {
+          actions: {
+            report: () => {
+              done();
+            }
+          }
+        }
+      );
+
+      const currentState = 'green';
+      const nextState = lightMachine.transition(currentState, 'TIMER');
+
+      // saves state and recreate it
+      const recreated = JSON.parse(JSON.stringify(nextState));
+      const previousState = State.create(recreated);
+      const resolvedState = lightMachine.resolveState(previousState);
+
+      const service = interpret(lightMachine);
+      service.start(resolvedState);
+    });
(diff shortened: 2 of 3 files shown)


.changeset/wet-swans-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Actions from a restored state provided as a custom initial state to `interpret(machine).start(initialState)` are now executed properly. See #1174 for more information.
```
Andarist, line 5:
```suggestion
Actions from a restored state provided as a custom initial state to `interpret(machine).start(initialState)` are now executed properly. See [#1174](https://github.com/davidkpiano/xstate/issues/1174) for more information.
```

packages/core/test/interpreter.test.ts
```diff
@@ -118,6 +118,56 @@ describe('interpreter', () => {
         done();
       }, 100);
     });
+
+    // https://github.com/davidkpiano/xstate/issues/1174
+    it('executes actions from a restored initial state', (done) => {
```
Andarist, line 123: nit: I wouldn't call it initial state - but rather just restored or rehydrated. The initial state (the name) is already used in other places such as here:
https://github.com/davidkpiano/xstate/blob/f685e674ac9d57f832df3f932af7de848f1e0d7f/packages/core/src/interpreter.ts#L225
The accepted parameter is called `initialState` here though:
https://github.com/davidkpiano/xstate/blob/f685e674ac9d57f832df3f932af7de848f1e0d7f/packages/core/src/interpreter.ts#L462-L466
so it follows that OK, although I find this parameter name confusing in the very same way.

Andarist:
I'm having second thoughts - actions are being considered fire and forget, so this is very likely that actions for a particular state has already been executed and only after that the state has been persisted and rehydrated later. In such a case action would be fired twice - which is not necessarily always a good thing. It should be up to a developer to decide about this because we don't know what's their strategy for persistence & rehydration.

davidkpiano:
Regardless, this PR is still correct, because there is a bug where actions aren't being resolved correctly, even though they should. The fix is general.

The behavior of whether or not those _resolved_ actions should be executed is a separate issue, but I agree, it should be addressed.

Andarist:
👌

PR #1188, opened 2020-05-16
Work around erroneously inherited execute = false option

### LINKED ISSUE #1125: [@xstate/react] rc.4 has an issue when sending events from parent to child
**Description**
xstate/react@v1.0.0-rc.4 causes an issue with sending events from a parent machine to a child machine.

**Expected Result**
Sending an event from the parent machine to the child machine causes the child to transition and fire any actions associated with the transition.

**Actual Result**
The child seems to transition to the new state but doesn't run the actions

**Reproduction**
Steps to reproduce:
1. Go to sandbox https://codesandbox.io/s/xstate-react-reddit-example-with-actors-x4nk0?file=/src/machine.js
2. Using rc.4, clicking the button doesn't run the action
3. Using rc.3 (make sure to refresh sandbox), clicking the button runs the action

**Additional context**
Happens in xstate/react@v1.0.0-rc.4
Works in xstate/react@v1.0.0-rc.3

### LINKED ISSUE #1161: Machines don't invoke service when spawned
**Description**

I followed the reddit example but with `createMachine` instead of `Machine` and found that the subreddit machine which should send an api call on `spawn` does not. This does not occur with `Machine`

**Expected Result**

expect `createMachine` and `Machine` to have the same behavior

**Actual Result**

subreddit machine is always in `loading` state and never invokes the service

**Reproduction**

https://codesandbox.io/s/awesome-tdd-37ueu?file=/src/redditMachine.js

**Additional context**

xstate: 4.9.1
@xstate/react: 1.0.0-rc4

davidkpiano:
Closes #1125, closes #1161 (issues observed with `@xstate/react@next` which has `execute: false`).

This is an alternative for #1162 that introduces no breaking changes.

The `childOptions` option will not be documented, as it is a temporary fix that will be deprecated in version 5.

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -146,7 +146,8 @@ export class Interpreter<
       }
     },
     logger: global.console.log.bind(console),
-    devTools: false
+    devTools: false,
+    childOptions: {}
   }))(typeof window === 'undefined' ? global : window);
   /**
    * The current state of the interpreted machine.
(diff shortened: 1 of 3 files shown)


Andarist:
From what I understand - this is only to handle `@xstate/react@next` case that is being caused by scheduling all actions through the React's commit phase. We've agreed that this is not needed as XState itself already guarantees actions not being re-executed multiple times so shouldn't we just revert this scheduling behavior from `@xstate/react` instead?

PR #1189, opened 2020-05-16
Fix resolution to never for state.matches(...)

### LINKED ISSUE #947: @xstate/react: typescript: state is always `never` after calling `.matches()` twice in the same expression
**Description**
@xstate/react: typescript: state is always `never` after calling `.matches()` twice (in the same expression?)

```js
function App () {
  const [state, _] = useMachine(toggleMachine)

  if (state.matches('a') || state.matches('b')) {
    //                        ^ is of type `never`
    console.log('do a thing')
  }


  /*
   * doesn't work like this either
   *
   * if (state.matches('a')) {
   *   // ...
   * } else if (state.matches('b')) {
   *   // ...
   * }
   */
}
```

Is this an antipattern or am I doing something wrong?

Sadly I don't have enough typescript experience at the moment to understand the following code in detail: 
https://github.com/davidkpiano/xstate/blob/d6eda85b28759b4cc296ca8f88633f2342045e31/packages/core/src/State.ts#L284-L291

**Expected Result**
I should be able to check if a state is either one of a state

**Actual Result**
I can't check if a type is one of either states

**Reproduction**

Simple Reproduction: https://codesandbox.io/s/xenodochial-faraday-05nmv

**Additional context**

```
{
  "@xstate/react": "^0.8.1",
  "xstate": "^4.7.6"
}
```

davidkpiano:
Closes #947

### DIFF
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -290,9 +290,12 @@ export class State<
    */
   public matches<TSV extends TTypestate['value']>(
     parentStateValue: TSV
-  ): this is TTypestate extends { value: TSV }
-    ? State<TTypestate['context'], TEvent, TStateSchema, TTypestate>
-    : never {
+  ): this is State<
+    (TTypestate extends { value: TSV } ? TTypestate : never)['context'],
+    TEvent,
+    TStateSchema,
+    TTypestate
+  > & { value: TSV } {
     return matchesState(parentStateValue as StateValue, this.value);
   }
 }
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -84,7 +84,7 @@ export class State<
   TContext,
   TEvent extends EventObject = EventObject,
   TStateSchema extends StateSchema<TContext> = any,
-  TTypestate extends Typestate<TContext> = any
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 > {
   public value: StateValue;
   public context: TContext;
--- a/.changeset/seven-gorillas-melt.md
+++ b/.changeset/seven-gorillas-melt.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Previously, `state.matches(...)` was problematic because it was casting `state` to `never` if it didn't match the state value. This is now fixed by making the `Typestate` resolution more granular.
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -51,7 +51,7 @@ export function Machine<
 export function createMachine<
   TContext,
   TEvent extends EventObject = AnyEventObject,
-  TTypestate extends Typestate<TContext> = any
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(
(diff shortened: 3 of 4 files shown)


packages/core/src/State.ts
```diff
@@ -290,9 +290,12 @@ export class State<
    */
   public matches<TSV extends TTypestate['value']>(
     parentStateValue: TSV
-  ): this is TTypestate extends { value: TSV }
-    ? State<TTypestate['context'], TEvent, TStateSchema, TTypestate>
-    : never {
+  ): this is State<
+    (TTypestate extends { value: TSV } ? TTypestate : never)['context'],
```
Andarist, line 294: this introduces a regression on `.context` property after `.matches` call:
```ts
type TContext = {foo: number}

const toggleMachine = createMachine<TContext>({
  id: 'toggle',
  initial: 'a',
  states: {
    a: { on: { TOGGLE: 'b' } },
    b: { on: { TOGGLE: 'a' } }
  }
});

const state = toggleMachine.initialState;

if (state.matches('a')) {
  // context is now `any` but was correctly preserved as TContext before this change
  const { context } = state
}
```

davidkpiano: Resolved!

packages/core/src/State.ts
```diff
@@ -290,9 +290,12 @@ export class State<
...
+    TEvent,
+    TStateSchema,
+    TTypestate
+  > & { value: TSV } {
```
Andarist, line 298: Can't say I understand this but seems to be crucial for the fix 😅

davidkpiano: Haha yeah 😅 So here's the explanation for this:

If `TTypestate` is just `{ string: TContext }`, then without that `{ value: TSV }`, this would evaluate to:

```ts
this is State<TContext, TEvent, TStateSchema, TTypestate>
```

So if we have `state.matches(...)`, that would evaluate to:

```ts
if (state.matches('foo')) {
  // state is State<TContext, TEvent, TStateSchema, TTypestate>
} else {
  // state is NOT State<TContext, TEvent, TStateSchema, TTypestate>
  // ... which is "never" (see below)
}
```

So "state is NOT `State<TContext, TEvent, TStateSchema, TTypestate>`" evaluates to `never` because TypeScript is basically doing this:

```ts
this extends State<TContext, TEvent, TStateSchema, TTypestate> ? never : State<TContext, TEvent, TStateSchema, TTypestate>;
```

And because `state` is _always_ `State<TContext, TEvent, TStateSchema, TTypestate>` and `A & !A` always evaluates to false, that expands to:

```ts
State<TContext, TEvent, TStateSchema, TTypestate> extends State<TContext, TEvent, TStateSchema, TTypestate>
  ? never
  : State<TContext, TEvent, TStateSchema, TTypestate>;
```

And it will always be `never`... which was a problem.

So we narrowed it down to `State<TContext, TEvent, TStateSchema, TTypestate> & { value: TSV }` so that `this` can still be `State<TContext, TEvent, TStateSchema, TTypestate>`, just with a different `value`.

PR #1192, opened 2020-05-19
Add tests to validate choose actions work with the config option

CodingDive:
No bug report, just more tests.

### DIFF
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -1293,7 +1293,7 @@ describe('choose', () => {
     expect(service.state.context).toEqual({ counter: 101, answer: 42 });
   });
 
-  it('should be able to use actions defined in options', () => {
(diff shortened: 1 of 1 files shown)


packages/core/test/actions.test.ts
```diff
@@ -1319,6 +1322,39 @@ describe('choose', () => {
...
+      },
+      {
+        guards: {
+          worstGuard: () => true
+        },
+        actions: {
+          revealAnswer: assign<Ctx>({ answer: 42 }),
+          conditionallyRevealAnswer: choose([
+            { cond: 'worstGuard', actions: 'revealAnswer' }
+          ])
```
CodingDive, line 1349: I have some code that does exactly that but the `revealAnswer` assign action equivalent in my code is never executed. Instead, it just logs `Warning: No implementation found for action type 'xstate.assign'` 🤔

PR #1194, opened 2020-05-19
add: Error message when a nested state is set

f1729:
I think this error message can be useful, I lost ~30 minutes trying to figure out why my SM can't did a transition.

What do you think about?

### DIFF
--- a/packages/xstate-fsm/src/index.ts
+++ b/packages/xstate-fsm/src/index.ts
@@ -91,7 +91,8 @@ export function createMachine<
       .keys(fsmConfig['states'])
       .forEach((state) => {
         if (fsmConfig['states'][state].states) {
-          throw new Error('Finite state nested is not supported in @xstate/fsm');
+          throw new Error(`Nested finite states not supported. 
+            Please check the configuration for the "${state}" state.`);
         }
       });
   }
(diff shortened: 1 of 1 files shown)


packages/xstate-fsm/src/index.ts
```diff
@@ -86,6 +86,15 @@ export function createMachine<
     actions?: StateMachine.ActionMap<TContext, TEvent>;
   } = {}
 ): StateMachine.Machine<TContext, TEvent, TState> {
+  
+  Object
```
davidkpiano, line 90: Can you wrap this in `if (!IS_PRODUCTION)`?

Also, the documentation does say that it does not support nested states.

packages/xstate-fsm/src/index.ts
```diff
@@ -86,14 +86,15 @@ export function createMachine<
...
-      if (fsmConfig['states'][state].states) {
-        throw new Error('Finite state nested is not supported in @xstate/fsm');
-      }
-    });
+  if (!IS_PRODUCTION) {
+    Object
+      .keys(fsmConfig['states'])
+      .forEach((state) => {
+        if (fsmConfig['states'][state].states) {
+          throw new Error('Finite state nested is not supported in @xstate/fsm');
```
davidkpiano, line 94:
```suggestion
          throw new Error(`Nested finite states not supported. Please check the configuration for the "${state}" state.`);
```

PR #1203, opened 2020-05-23
V5: Simplify Interpreter + Dev Tools integration

davidkpiano:
This PR incorporates some planned deprecations, and greatly simplifies the `Interpreter` class.

It also provides a simpler way of integrating custom dev tools.


davidkpiano:
@Andarist ping for review? 🙏 thanks!

.changeset/fast-tables-design.md
```diff
@@ -0,0 +1,52 @@
+---
+'xstate': major
+---
+
+- The `execute` option for an interpreted service is no longer inherited by invoked/spawned services. They default to `execute: true`.
```
Andarist, line 5: q: could we just remove `execute` altogether? imho it's pretty confusing and avoiding side-effects cant be done in a multiple different ways

packages/core/devTools/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "main": "dist/xstate.cjs.js",
+  "module": "dist/xstate.esm.js",
+  "preconstruct": {
+    "source": "../src/devTools"
```
Andarist, line 5: nit: maybe `dev-tools`? somehow this would look better for me as part of the public export, but I also don't care that much, so please chose whatever you like

packages/core/src/devTools/redux.ts
```diff
@@ -0,0 +1,39 @@
+import { State } from '../State';
```
Andarist, line 1: its pretty small but also its not part of the core, so maybe worth moving into a separate package? if you think that would be preferable I can take care of creating a new directory and wiring things up

packages/core/src/interpreter.ts
```diff
@@ -296,7 +263,7 @@ export class Interpreter<
       | ((state: State<TContext, TEvent, any, TTypestate>) => void)
       | Observer<State<TContext, TEvent, any, TTypestate>>,
     // @ts-ignore
-    errorListener?: (error: any) => void,
+    errorListener?: (error: Error) => void,
```
Andarist, line 266: this is incorrect as currently errorListeners can be called with error events, but `Error` type represents regular javascript Error objects

packages/core/src/interpreter.ts
```diff
@@ -377,12 +319,20 @@ export class Interpreter<
...
     return this;
   }
+
+  /**
+   * Adds an error listener that is notified with an `Error` whenever an
+   * error occurs during execution.
+   *
+   * @param listener The error listener
+   */
   public onError(
```
Andarist, line 329: isnt this also somewhat redundant? one could just subscribe to error listeners via subscribe

packages/core/src/interpreter.ts
```diff
@@ -497,10 +438,10 @@ export class Interpreter<
    */
   public send = (
     event: SingleOrArray<Event<TEvent>> | SCXML.Event<TEvent>
-  ): State<TContext, TEvent, TStateSchema, TTypestate> => {
+  ): void => {
```
Andarist, line 441: 👍

packages/core/src/devTools/redux.ts
davidkpiano: For now, until we have solid dev tools, let's keep this here. It won't be included in core, but should be able to be imported from here still.

packages/core/src/interpreter.ts
davidkpiano: No, this should be for any errors, even for those "uncaught" by transitions.

packages/core/devTools/package.json
davidkpiano: Changed to `dev`

PR #1204, opened 2020-05-24
added stenciljs recipe

edsilv:


docs/recipes/stencil.md
```diff
@@ -0,0 +1,85 @@
...
+
+[Stencil](https://stenciljs.com/) web components work very well with XState.
+
+### components/toggle/toggle-machine.ts
+
+```js
+import { Machine } from "xstate";
+
+// this machine is completely decoupled from stencil
+export const toggleMachine = Machine({
```
davidkpiano, line 11:
```suggestion
import { createMachine } from "xstate";

// this machine is completely decoupled from stencil
export const toggleMachine = createMachine({

```

edsilv:
@davidkpiano One concern raised about XState from the stencil people is that it's 59.1kb (vs stencil-store at 1.8kb) - is there anything on the horizon to potentially reduce this?

davidkpiano:
> One concern raised about XState from the stencil people is that it's 59.1kb (vs stencil-store at 1.8kb) - is there anything on the horizon to potentially reduce this?

It is 14-15kb minified/GZIPped, which is about the same size as MobX.

V5 will be smaller (but again, don't compare to the unminified/unGZIPped size), and also `@xstate/fsm` exists which is less than 2kb.

simonhaenisch:
To clarify, I'm the one referred to as "the stencil people" here (from a Slack conversation), but I'm just another developer using Stencil (with a couple PR contributions). I compared un-minified size for complexity reasons not for bundle size comparisons (I might not have been clear enough on that). A bigger source would somewhat mean a more complex API (i. e. more features, more explicit/verbose, etc.). I was just saying that there's projects with very simple requirements regarding state, where something slimmer might be more appropriate.

edsilv:
@simonhaenisch We could try reworking the recipe to use `@xstate/fsm` ?

simonhaenisch:
I guess a recipe for both would be good? But from what I can tell your recipe would work as-is with `@xstate/fsm`, so it's as simple as changing the import. A hint as a comment might be sufficient?

```ts
import { createMachine } from 'xstate'; // or '@xstate/fsm'
```

(That would be the same import if you had applied @davidkpiano's suggestion about using  `createMachine` instead of `Machine`... which you marked as resolved without actually applying the change?)

docs/recipes/stencil.md
```diff
@@ -0,0 +1,85 @@
+# Usage with Stencil
+
+[Stencil](https://stenciljs.com/) web components work very well with XState.
+
+### components/toggle/toggle-machine.ts
```
simonhaenisch, line 5: I'd not place this alongside the component, unless the state machine is really just relevant for this single component?

```suggestion
### `src/helpers/toggle-machine.ts`
```

docs/recipes/stencil.md
```diff
@@ -0,0 +1,85 @@
...
+      on: { toggle: "active" }
+    },
+    active: {
+      on: { toggle: "inactive" }
+    }
+  }
+});
+```
+
+### components/toggle/toggle.tsx
```
simonhaenisch, line 25:
```suggestion
### `src/components/toggle/toggle.tsx`

```

docs/recipes/stencil.md
```diff
@@ -0,0 +1,85 @@
...
+Add a `state` property to your component, decorated with `@State` so that it triggers a re-render when changed.
+
+On `componentWillLoad`, interpret the `toggleMachine` and listen for state transitions.
+
+After a transition has occured, the `state` property is set to the machine's new state, triggering a re-render.
+
+```js
+import { Component, h, State, Prop } from "@stencil/core";
+import { interpret } from "xstate";
+import { toggleMachine } from "./toggle-machine";
```
simonhaenisch, line 36:
```suggestion
import { toggleMachine } from "../helpers/toggle-machine";

```

docs/recipes/stencil.md
```diff
@@ -0,0 +1,85 @@
...
+  componentDidUnload() {
+    this._service.stop();
+  }
+
+  render() {
+    const { current } = this.state;
+    const { send } = this._service;
+
+    return (
+      <button onClick={() => send("toggle")}>
```
simonhaenisch, line 65: You're not defining `send` anywhere?

edsilv: it's being destructured from `this._service`

edsilv:
> I guess a recipe for both would be good? But from what I can tell your recipe would work as-is with `@xstate/fsm`, so it's as simple as changing the import. A hint as a comment might be sufficient?
> 
> ```ts
> import { createMachine } from 'xstate'; // or '@xstate/fsm'
> ```
> 
> (That would be the same import if you had applied @davidkpiano's suggestion about using `createMachine` instead of `Machine`... which you marked as resolved without actually applying the change?)

woops - have updated the PR

docs/recipes/stencil.md
simonhaenisch: 🤦🏻‍♂️ I'm blind.

BTW, I'd do the `send` like this:

```ts
export class Toggle {
  // ...

  toggle = () => this._service.send("toggle");

  render() {
    return (
      <button onClick={this.toggle}>...</button>
    );
  }
}
```

Because that way `onClick` always binds to the same reference, whereas binding to a new function reference (anonymous function) on each render would be detected as a prop change and re-render the element. Micro-optimization though... so anyway, whatever you find more clear.

docs/recipes/stencil.md
```diff
@@ -0,0 +1,87 @@
...
+
+Add a `state` property to your component, decorated with `@State` so that it triggers a re-render when changed.
+
+On `componentWillLoad`, interpret the `toggleMachine` and listen for state transitions.
+
+After a transition has occured, the `state` property is set to the machine's new state, triggering a re-render.
+
+```js
+import { Component, h, State, Prop } from "@stencil/core";
+import { interpret } from "xstate";
```
simonhaenisch, line 36:
```suggestion
import { interpret } from "@xstate/fsm";
```

edsilv: https://github.com/davidkpiano/xstate/pull/1204/commits/bb14464408621cc9842c275d46e1fad340552f05

docs/recipes/stencil.md
```diff
@@ -0,0 +1,87 @@
...
+import { toggleMachine } from "../helpers/toggle-machine";
+
+
+@Component({
+  tag: "my-toggle",
+  styleUrl: "toggle.css",
+  shadow: true
+})
+export class Toggle {
+  @State() state;
```
simonhaenisch, line 46: Is there a type for this? I think it's `StateMachine.Service`? Like

```ts
import { interpret, StateMachine } from '@xstate/fsm'

export class Toggle {
  @State() state: StateMachine.Service;
}
```

?

simonhaenisch: Ah no sry that's for `private _service`... there's also `StateMachine.State` though for this I think.

edsilv: I've got a much more strictly typed version here: https://github.com/edsilv/stencil-xstate-demo/blob/master/src/components/toggle/toggle.tsx
Was trying to keep it simpler for the recipe, but maybe it's better not to?

docs/recipes/stencil.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+### `src/components/toggle/toggle.tsx`
+
+
+Add a `state` property to your component, decorated with `@State` so that it triggers a re-render when changed.
+
+On `componentWillLoad`, interpret the `toggleMachine` and listen for state transitions.
+
+After a transition has occured, the `state` property is set to the machine's new state, triggering a re-render.
```
simonhaenisch, line 32: This make it sound like the state is not already set on the initial render? Would there be a way to get the initial state? Maybe sth like

```ts
export class Toggle {
  @State() state = toggleMachine.initialState;
}
```

(which would potentially even infer the correct `State` type?)

docs/recipes/stencil.md
simonhaenisch: Ah I see yeah those types are quite verbose...

docs/recipes/stencil.md
edsilv: I've made a codesandbox here btw to test this all works so far :-) 

https://codesandbox.io/s/stencil-xstate-7ll30

Great to see that stencil now works on codesandbox!

I think you may have to `interpret` first to get the initial state?

simonhaenisch: Yeah Code Sandbox is pretty cool. Have a look at this one: https://codesandbox.io/s/stencil-xstate-wz0wq?file=/src/components/toggle/toggle.tsx

It correctly infers all the types, and no need to destructure `state` anymore. Only type info I added is for creating the machine.

```ts
export const toggleMachine = createMachine<{ toggle: "active" | "inactive" }>(/* ... */);
```

```ts
export class Toggle {
  @State() state = toggleMachine.initialState;

  private _service = interpret(toggleMachine);

  componentWillLoad() {
    this.service.subscribe(state => {
      this.state = state;
    });
    // ...
  }

  render() {
    return (
      // ...
        {this.state.value === "inactive" ? "Off" : "On"}
      // ...
    }
}
```

simonhaenisch: Ok this was pretty wrong... i typed context instead of state and the initialState indeed doesn't work. But I think wrt types we can definitely still improve this recipe.

edsilv: Nice, have updated the PR to match this.

docs/recipes/stencil.md
```diff
@@ -0,0 +1,79 @@
...
+
+@Component({
+  tag: "my-toggle",
+  styleUrl: "toggle.css",
+  shadow: true
+})
+export class Toggle {
+  private _service = interpret(toggleMachine);
+
+  @State() state = toggleMachine.initialState;
```
simonhaenisch, line 41: I tested this and it doesn't actually get the initial state for me... maybe it's better to just do

```suggestion
  @State() state: typeof toggleMachine.initialState;
```

docs/recipes/stencil.md
```diff
@@ -0,0 +1,79 @@
...
+  componentDidUnload() {
+    this._service.stop();
+  }
+
+  render() {
+    const { send } = this._service;
+
+    return (
+      <button onClick={() => send("toggle")}>
+        {this.state === "inactive" ? "Off" : "On"}
```
simonhaenisch, line 60: I still had to access the state's `value` here...

```suggestion
        {this.state.value === "inactive" ? "Off" : "On"}
```

docs/recipes/stencil.md
```diff
@@ -0,0 +1,79 @@
+# Usage with Stencil
+
+[Stencil](https://stenciljs.com/) web components work very well with XState.
+
+### `src/helpers/toggle-machine.ts`
+
+```js
+import { createMachine } from "@xstate/fsm";
+
+export const toggleMachine = createMachine<{ toggle: "active" | "inactive" }>({
```
simonhaenisch, line 10: The first generic type parameter is actually for the context which is not used at all here, right?

I tried

```ts
type ToggleState = { context: {} } & ({ value: "active" } | { value: "inactive" });

export const toggleMachine = createMachine<{}, { type: "toggle" }, ToggleState>(/* ... */);
```

which I guess is correct type-wise(?), but it doesn't really give much type safety 😐 `this.state.value` is still inferred as `string` rather than the string literal type, and `send` is supposed to have type safety when you use a different syntax `send({ type: 'toggle' })` but it also just infers `type` as string for me 🤷🏻‍♂️  Maybe that only works with `xstate` but not `@xstate/fsm`?
(https://xstate.js.org/docs/guides/typescript.html#actions)

simonhaenisch: Hm never mind I think it was some caching issue with codesandbox 🙈

It does infer the `send` type but `state.value` is inferred as `string`.

docs/recipes/stencil.md
simonhaenisch: Hm that was also because of that caching issue I think, it does actually work to get the initial state!

simonhaenisch:
@edsilv I updated https://codesandbox.io/s/stencil-xstate-wz0wq?file=/src/helpers/toggle-machine.ts and it works except the string literal inference of the state value...

@davidkpiano maybe you can shed some light on this? I looked through the source and it seems like it's not possible to strongly type the state's value, at least in `@xstate/fsm`.

```ts
export const toggleMachine = createMachine<{}, { type: "toggle" }, { value: "active" | "inactive", context: {} }>({
  id: "toggle",
  initial: "active",
  states: {
    active: { on: { toggle: "inactive" } },
    inactive: { on: { toggle: "active" } },
  }
});
```

This makes sure that `states.active` and `states.inactive` are required, but the `value` type is just set to `string` here:

https://github.com/davidkpiano/xstate/blob/c2605087ce8e09ea0310eaa9fee869068e8e3bba/packages/xstate-fsm/src/types.ts#L142

It would be neat if the state's `value` type is inferred as `'active' | 'inactive'` (same for `initial`).

It would also be nice if it wasn't necessary to specify a type for `context` even though it's unused. You can use a feature called "conditional types" to improve this, I think:

https://www.typescriptlang.org/play?#code/C4TwDgpgBAKuEGdgENgQDwwMIHsB2aAHsFALxTJ4gB8ZUA2tvkcALpQTER4AmCDeAK4AbYVAA+UQbwgAzAJZ4IPdgH5Y8AMoo0AdXnAAFjkHBcBTiQBcGyNtQR9R8y0wvL1ANwAob4rQATrLIAMbQcHY6jgbGpu7EUADe3lCpUABuyMKCEDZIAYoA5j4Avr7+EEGh4VpRTobxwG7MlhxcvPyJDADWECB5wAV4haw2lCBQJbSW3Hy2EPZ6MSZmLQnJaVAha8A2TBbEpb4A9MdQAAbbB8DnUPIIeADkJAEQAI6C8q88UIaVEN4rkgoLIcDg9vAkA50Ep0pVaOQuplsrkoI9QThHpMfECSAAjZABCGQKFodDjBFJDJZHI2R4EgJYko4-DAgkAL2JiCi6GkPDkimUlKRNNR9OQ7KZLLwwKEok581JGDlwmF1JRdJVkuxJzOlx2t3uUFeHy+yl+-0BrJIGIZXKV6ESUzoIo1aNthKlQA

```ts
type Typestate<TContext = any> = [TContext] extends [null | undefined] ? TypeStateWithoutContext : TypeStateWithContext<TContext>;

interface TypeStateWithoutContext {
    value: string;
}

interface TypeStateWithContext<TContext extends { [key: string]: any }> extends TypeStateWithoutContext {
    context: TContext;
}

// `context` isn't required here
const foo: Typestate<never> = { value: 'foo' };
const bar: Typestate<any> = { value: 'bar' };
const baz: Typestate<undefined> = { value: 'baz' };
const nullz: Typestate<null> = { value: 'nullz' };

// `context` is required here
const foobar: Typestate<{}> = { value: 'foobar' };
// => Property 'context' is missing in type '{ value: string; }'
// but required in type 'TypeStateWithContext<{}>'.
```

docs/recipes/stencil.md
simonhaenisch, line 10:
```suggestion
export const toggleMachine = createMachine<
  undefined,
  { type: "toggle" },
  { value: "active" | "inactive"; context: undefined }
>({
```

The other one "works" but is incorrect.

docs/recipes/stencil.md
simonhaenisch, line 60:
```suggestion
        {this.state.value === "inactive" ? "Off" : "On"}
```

davidkpiano:
> @davidkpiano maybe you can shed some light on this? I looked through the source and it seems like it's not possible to strongly type the state's value, at least in @xstate/fsm.

No, not yet.

Issue #1195, opened 2020-05-20
state.matches does not accept valid parent value when using typestates


njdancer:
**Description**
When using typestates, the parentStateValue accepted by state.matches is constrained to only valid states of the machine(as specified in the typestate) whereas the documentation indicates this function should also be used to determine if a state is a subset of the given state.

**Expected Result**
state.matches should accept any state that is a valid parent of the states provided in typestate.

**Actual Result**
state.matches returns error. ```Argument of type '\"draft\"' is not assignable to parameter of type \"draft.editing\" | \"draft.submitted\" | ...```

**Reproduction**
https://codesandbox.io/s/xstate-typestate-matches-repro-09be7?file=/src/index.tsx

**Additional context**
Not quite sure what the solution is here. AFAIK string literals are opaque to tsc so not sure how you would be able to infer all the valid parent states when only being provided complete child states. If this means what i'm asking for is impossible then that's fine. Mostly just wanting to make sure i'm using typestates correctly. I did try explicitly adding the parent states I was using to the typestate which seems to work around this issue, however i'm unsure of the implications of this as the state machine can never really be in one of those states. Seems like the best strategy from my perspective but interested in your thoughts.

davidkpiano:
Regardless of the string literals, the behavior here is correct. In the CodeSandbox, you are trying to match `"active"` with `"active.test"`, which is not a match, since `"active.test"` is more specific than `"active"` (since `"active"` can mean `"active.anything"` which does not match `"active.test"`).

The right thing to do here is to be more explicit about your typestates:

```ts
type ToggleState =
  | { value: "inactive"; context: ToggleContext }
  | { value: 'active' | { active: 'test' }; context: ToggleContext };
```

Note to self: should add in the documentation that string representation of states for compound typestates (e.g., `"foo.bar"`) should not be allowed.

njdancer:
Ahh ok. That makes sense. Thanks for the example.

Does this mean that a typestate can only specify the context type for a top level state?

Would be happy to submit a PR to update the typestate example in the docs to include a compound state if you’d be willing to accept.

In any case, thanks for the great library.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

davidkpiano:
> Does this mean that a typestate can only specify the context type for a top level state?

Context is for the entire machine; not sure what you mean by this.

njdancer:
[...]

PR #1208, opened 2020-05-26
Add compound state warning to typescript docs

njdancer:
Following the discussion in #1195, I've updated the typescript documentation with the following:

- Compound state values inside a typestate should be specified using the object syntax rather than the string syntax.
- Added warning regarding compound states and the need for parent states to be explicitly modelled
- Added example of type state declaration using a type union for the value. This is obviously a typescript feature rather than a feature of xstate itself, however I hadn't considered this pattern prior to your example and it seems this is quite useful to avoid duplication.

I tried to avoid committing the other formatting changes however the post commit hook seemed to force it.

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -150,7 +150,7 @@ Typestates are a concept that narrow down the shape of the overall state `contex
 
 A `Typestate` is an interface consisting of two properties:
 
-- `value` - the state value of the typestate(compound states should be referenced using object syntax)
+- `value` - the state value of the typestate (compound states should be referenced using object syntax; e.g., `{ idle: 'error' }` instead of `"idle.error"`)
 - `context` - the narrowed context of the typestate when the state matches the given `value`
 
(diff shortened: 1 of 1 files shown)


docs/guides/typescript.md
```diff
@@ -122,34 +122,35 @@ const myMachineConfig: MachineConfig<TContext, TSchema, TEvent> = {
...
 ```
+
 ## Typestates <Badge text="4.7+" />
 
 Typestates are a concept that narrow down the shape of the overall state `context` based on the state `value`. This can be helpful in preventing impossible states and narrowing down what the `context` should be in a given state, without having to write excessive assertions.
 
 A `Typestate` is an interface consisting of two properties:
 
-- `value` - the state value of the typestate
+- `value` - the state value of the typestate(compound states should be referenced using object syntax)
```
davidkpiano, line 153:
```suggestion
- `value` - the state value of the typestate (compound states should be referenced using object syntax; e.g., `{ idle: 'error' }` instead of `"idle.error"`)
```

PR #1215, opened 2020-05-30
[@xstate/react] Add useActor, refactor useService

davidkpiano:
This PR adds `useActor` and refactors `useService` to use that hook.


packages/xstate-react/src/useActor.ts
```diff
@@ -0,0 +1,30 @@
...
+    const subscription = actorLike.subscribe((latest) => {
+      setCurrent(latest);
+    });
+
+    return () => {
+      subscription.unsubscribe();
+    };
+  }, [actorLike]);
+
+  return [current, actorLike.send.bind(actorLike)];
```
Andarist, line 29: imho we should either assume that `actorLike.send` is not using `this` (or it's already bound), or we should memoize this here to provide a stable reference to the consumer

packages/xstate-react/src/useActor.ts
```diff
@@ -0,0 +1,30 @@
...
+  getCurrentValue: <TEvent extends EventObject, TEmitted = any>(
+    actorLike: ActorRef<TEvent, TEmitted> | ActorRefLike<TEvent, TEmitted>
+  ) => TEmitted = defaultGetCurrentValue
+): [TEmitted, Sender<TEvent>] {
+  const [current, setCurrent] = useState(getCurrentValue(actorLike));
+
+  useEffect(() => {
+    const subscription = actorLike.subscribe((latest) => {
+      setCurrent(latest);
+    });
```
Andarist, line 22:
```suggestion
    const subscription = actorLike.subscribe(setCurrent);
```

packages/xstate-react/src/useActor.ts
```diff
@@ -0,0 +1,30 @@
...
+export function useActor<TEvent extends EventObject, TEmitted = any>(
+  actorLike: ActorRef<TEvent, TEmitted> | ActorRefLike<TEvent, TEmitted>,
+  getCurrentValue: <TEvent extends EventObject, TEmitted = any>(
+    actorLike: ActorRef<TEvent, TEmitted> | ActorRefLike<TEvent, TEmitted>
+  ) => TEmitted = defaultGetCurrentValue
+): [TEmitted, Sender<TEvent>] {
+  const [current, setCurrent] = useState(getCurrentValue(actorLike));
+
+  useEffect(() => {
+    const subscription = actorLike.subscribe((latest) => {
```
Andarist, line 20: you should also `setCurrent(getCurrentValue(actorLike))` here, within the effect because the current value could have changed between render and effect execution

packages/xstate-react/src/useActor.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const defaultGetCurrentValue = <TEvent extends EventObject, TEmitted = any>(
+  actorLike: ActorRef<TEvent, TEmitted> | ActorRefLike<TEvent, TEmitted>
+): TEmitted => {
+  return 'current' in actorLike ? actorLike.current : (undefined as any); // V4 Actors do not have 'current'
+};
+
+export function useActor<TEvent extends EventObject, TEmitted = any>(
+  actorLike: ActorRef<TEvent, TEmitted> | ActorRefLike<TEvent, TEmitted>,
+  getCurrentValue: <TEvent extends EventObject, TEmitted = any>(
```
Andarist, line 13: should this be a public parameter? we expect a particular interface of an actor to be satisfied on the position of `actorLike` so it seems that we should be able to just use this information without having to accept `getCurrentValue` from a consumer

PR #1219, opened 2020-06-06
Pass invoke.data as meta argument to service creators. Closes #529

### LINKED ISSUE #529: Add ability to pass params to service declared as property
## Description:

This exchange in the chat best summarizes the request:

Q: Is there any way to pass hardcoded values to an invocation when the service is declared in a services property? 

A: Currently no, ...  Should be just like actions and guards

// future enhancement
```js
services: {
  fetchUser: (context, event, { service }) => {
    // service.params === { role, username, password }, etc.
  }
}
```

Should not be a breaking change

davidkpiano:
The resolved value of the `invoke.data` property is now available in the "invoke meta" object, which is passed as the 3rd argument to the service creator in `options.services`. This will work for all types of invoked services now, including promises, observables, and callbacks. The behavior of `invoke.data` with invoked machines is unchanged.

```js
  initial: 'pending',
  context: {
    id: 42
  },
  states: {
    pending: {
      invoke: {
        src: 'fetchUser',
        data: {
          userId: (context) => context.id
        },
        onDone: 'success'
      }
    },
    success: {
      type: 'final'
    }
  }
},
{
  services: {
    fetchUser: (ctx, _, { data }) => {
      return fetch(`some/api/user/${data.userId}`)
        .then(response => response.json());
    }
  }
}
```

### DIFF
--- a/.changeset/thirty-tomatoes-jump.md
+++ b/.changeset/thirty-tomatoes-jump.md
@@ -1,0 +1,36 @@
+---
+'xstate': minor
+---
+
+The resolved value of the `invoke.data` property is now available in the "invoke meta" object, which is passed as the 3rd argument to the service creator in `options.services`. This will work for all types of invoked services now, including promises, observables, and callbacks. The behavior of `invoke.data` with invoked machines is unchanged.
+
+```js
+const machine = createMachine({
+  initial: 'pending',
+  context: {
+    id: 42
+  },
+  states: {
+    pending: {
+      invoke: {
+        src: 'fetchUser',
+        data: {
+          userId: (context) => context.id
+        },
+        onDone: 'success'
+      }
+    },
+    success: {
+      type: 'final'
+    }
+  }
+},
+{
+  services: {
+    fetchUser: (ctx, _, { data }) => {
+      return fetch(`some/api/user/${data.userId}`)
+        .then(response => response.json());
+    }
+  }
+}
+```
--- a/.changeset/thirty-tomatoes-jump.md
+++ b/.changeset/thirty-tomatoes-jump.md
@@ -2,7 +2,7 @@
 'xstate': minor
 ---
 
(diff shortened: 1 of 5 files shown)


docs/guides/communication.md
```diff
@@ -229,7 +230,7 @@ const pingPongMachine = Machine({
         src: (context, event) => (callback, onReceive) => {
           // Whenever parent sends 'PING',
           // send parent 'PONG' event
-          onReceive((e) => {
+          onReceive(e => {
```
Andarist: seems like maybe you have not reinstalled the project and Prettier v1 has been used to format this file instead of currently used v2

.changeset/thirty-tomatoes-jump.md
```diff
@@ -0,0 +1,36 @@
+---
+'xstate': minor
+---
+
+The resolved value of the `invoke.data` property is now available in the "invoke meta" object, which is passed as the 3rd argument to the service creator in `options.services`. This will work for all types of invoked services now, including promises, observables, and callbacks. The behavior of `invoke.data` with invoked machines is unchanged.
```
Andarist, line 5: It's a little bit nit-picky - but service creators which return machines will from now on be able to use this invoke meta as well. Including this piece of information might make this description too complicated though, so maybe not worth including it? 🤔

packages/core/src/interpreter.ts
```diff
@@ -884,9 +888,7 @@ export class Interpreter<
           } else if (isMachine(source)) {
             // TODO: try/catch here
             this.state.children[id] = this.spawnMachine(
-              data
-                ? source.withContext(mapContext(data, context, _event))
-                : source,
+              data ? source.withContext(resolvedData) : source,
```
Andarist, line 891:
```suggestion
              resolvedData ? source.withContext(resolvedData) : source,
```

packages/core/test/invoke.test.ts
```diff
@@ -2363,3 +2363,50 @@ describe('invoke', () => {
     });
   });
 });
+
+describe('services option', () => {
+  it('should resolve service from options with string src', (done) => {
```
Andarist, line 2368: this test title doesn't seem right for what we are testing here, maybe something like that?
```suggestion
  it('should provide data params to a service creator', (done) => {
```

PR #1225, opened 2020-06-09
xstate-fsm machine transition example update

slmoore:
Corrected a bug in the xstate/fsm machine.transition() example.  Transitioning from redState to greenState.

Let me know if the changeset is required for this request, thanks!


Andarist:
thanks!

PR #1236, opened 2020-06-13
Docs: clarify the Vue recipe is Vue 3 code

perry-garvin:
Some might think this is Vue 2 code since as of this pull request Vue 3 hasn't been released but I clarified that it's Vue 3 code.


davidkpiano:
Thanks @perry-garvin but the code in the recipe is indeed for Vue 2.

PR #1239, opened 2020-06-13
[@xstate/react] Fix service double initialization in React strict mode. 

### LINKED ISSUE #1237: [React] Duplicate machine execution with React.StrictMode
**Description**
The core of the problem is that with `useMachine` a machine _instance_ is created twice and then follows whatever is on its initial path (invocation, actions...). Ultimately, this causes a cascade in my code where two WebSocket connections are created and receive events. Everything then happens twice.

**Expected Result**
An only a single instance of a machine should exist.

**Actual Result**
Two machine instances are executed separately.

**Reproduction**
I tried to use CodeSandbox but due to the unreleased fix for "actions in effects by default" I could not reproduce it.

https://codesandbox.io/s/sharp-waterfall-14m3c?file=/src/App.js

I made a repo instead which contains the latest `@xstate/react` built from the `master` branch.

https://github.com/FredyC/xstate-react-multiple-services-bug

It's just CRA, so as usual `yarn install` & `yarn start`. When you run that, in the console, you can see.

image

davidkpiano:
Fixes #1237 

I was able to juggle some things around `useMachine` and get this to work. The only difference (which shouldn't affect apps) is that the first `[state, ...]` that comes from `useMachine(...)` comes from the non-interpreted created `machine`, _not_ the created `service` (but the `service` uses that `machine`).

The main issue is/was that `service.start(...)` is a side-effect that is being called in `useConstant(...)`, so I moved that to `useEffect`.


Andarist:
This IMHO creates serious usability issues around invoked/spawned machines (in the initial state) when passing them to child components: https://codesandbox.io/s/invoke-child-bug-6r5zu?file=/src/App.js . Those are not accessible right away when we postpone the start. I'm not opposed to delaying `.start()` per se, but this one seems to me like a very important use case that should just work, without anything extra on the developer's side.

This also creates a problem with effects' execution order between Child and a Parent: https://codesandbox.io/s/lingering-meadow-2wpbq?file=/src/App.js . We could technically queue sent events and replay them, but should we?

Andarist:
Just want to write down here some of my thoughts regarding this:
- I believe that not being able to read child machines synchronously during the first render is a sign of broken/incomplete solution:
  - introduces the need for defensive programming which is overall harder to read & write
  - makes SSRed content incomplete, so it might lead to FOUC effect as mount might start a machine, thus invoke children and thus render additional content
  - might be prone to layout shifting - for pretty much same reasons as mentioned in SSR argument, just in a different situation (when rendering a new machine dynamically, during app's lifetime)
- I agree though that probably starting in an effect would be overall better and, while being a step aside from a traditional XState semantics, it would make the integration with React more straightforward overall as firing side effects "on mount" is a popular thing to do. It would also fix our memory leak which I have recognized [here](https://github.com/davidkpiano/xstate/issues/956#issuecomment-633728868) - however, I don't think that this memory leak is tightly coupled to starting in a render, it's just caused by the fact that we have this global registry which maybe could be removed in the future and then starting in render would not cause a memory leak, not sure about this, just an idea based on a fact that we should push towards explicit address sharing represented by actor refs, rather than relying on a global registry of "magic strings" (session ids)
- when implementing starting when mounting we'd have to remember to implement event queuing, as it should be possible to send an event from a child machine to a parent within `useEffect` and descendant `useEffect`s are fired before parent ones

I'm not sure how to best approach this in v4 as this might require an overhaul of how Actors are initialized etc

davidkpiano:
> I'm not sure how to best approach this in v4 as this might require an overhaul of how Actors are initialized etc

Right - in V5, actors by default aren't "started" yet, and all the interpreter does is call `actor.start()` on them - so it's the exact same instance, and we can safely return the actor instances in the initial state without causing any unwanted side-effects, so V5 will "just work" with starting interpreters in `useEffect`.

I think the best strategy for now is to:
- Keep the current behavior, and perhaps _warn_ about invoking/spawning in the initial state (also add this to the docs)
- Properly "fix" this in V5, and pin `@xstate/react` version 1.0 to `xstate@5.x`.

Andarist:
Should we close this PR then?

davidkpiano:
@Andarist I revisited this:

> I'm not sure how to best approach this in v4 as this might require an overhaul of how Actors are initialized etc

Here is the new strategy:

- The initial state returned by `[state, send] = useMachine(...)` will always be the non-interpreted machine's initial state.
- This means that when `useLayoutEffect` runs, the component will be rendered again with the _interpreted_ initial state, which may be different than the first initial state.
- Namely, in the non-interpreted initial state, invoked/spawned actors will be "deferred actors", which is just the same "null-ish" actors with `.deferred = true`.
- Components will be able to interact with that "deferred actor" just the same as if it was the real actor. The `useActor(...)` hook can know if the actor is deferred (`.deferred === true`), and if it is, it will defer events by queueing them.
- When `useActor(...)` detects that the actor changed to an interpreted one, it will _replay_ those deferred events on the real actor (as the tests show).

To the developer, the way this works should be invisible. The only difference is that the component will render _two_ initial states, which should not be an issue (actions are only executed on the second initial state anyway, as that comes from the `service`).

We can probably add heuristics to prevent double-rendering the initial state if no actions were invoked, but I'm not sure this is worth doing. Thoughts?

davidkpiano:
> This IMHO creates serious usability issues around invoked/spawned machines (in the initial state) when passing them to child components: [codesandbox.io/s/invoke-child-bug-6r5zu?file=/src/App.js](https://codesandbox.io/s/invoke-child-bug-6r5zu?file=/src/App.js) . Those are not accessible right away when we postpone the start. I'm not opposed to delaying `.start()` per se, but this one seems to me like a very important use case that should just work, without anything extra on the developer's side.
> 
> This also creates a problem with effects' execution order between Child and a Parent: [codesandbox.io/s/lingering-meadow-2wpbq?file=/src/App.js](https://codesandbox.io/s/lingering-meadow-2wpbq?file=/src/App.js) . We could technically queue sent events and replay them, but should we?

As explained above, the current implementation fixes this problem.

packages/xstate-react/test/useService.test.tsx
```diff
@@ -173,4 +182,59 @@ describe('useService hook', () => {
...
+      }
+    });
+
+    const ChildTest: React.FC<{ actor: Actor<any> }> = ({ actor }) => {
+      const [state, send] = useActor(actor);
+
+      expect(state.value).toEqual('active');
+
+      React.useEffect(() => {
+        send({ type: 'FINISH' });
```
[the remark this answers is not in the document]
Andarist: Another (similar, but not involving any extra layers) test would be something like this:
```js
  it('child component should be able to send an event to a parent immediately in an effect', (done) => {
      initial: 'active',
      states: {
        active: {
          on: { FINISH: 'success' }
        },
        success: {}
      }
    });

    const ChildTest: React.FC<{ send: any }> = ({ send }) => {
      React.useEffect(() => {
        send({ type: 'FINISH' });
      }, []);

      return null;
    };

    const Test = () => {

      if (state.matches('success')) {
        done();
      }

      return <ChildTest send={send} />;
    };

    render(<Test />);
  });
```

packages/xstate-react/test/useService.test.tsx
```diff
@@ -173,4 +182,59 @@ describe('useService hook', () => {
...
+
+      if (state.matches('success')) {
+        done();
+      }
+
+      return <ChildTest actor={state.children.child} />;
+    };
+
+    render(
+      <React.StrictMode>
```
Andarist, line 235: I don't think `StrictMode` is that relevant for this test - it doesn't matter (for this test) if you wrap your root component with it or not.

packages/xstate-react/src/useMachine.ts
```diff
@@ -142,18 +142,21 @@ export function useMachine<
...
+    return machine.withConfig(machineConfig, {
       ...machine.context,
       ...context
     } as TContext);
+  });
 
-    return interpret(createdMachine, interpreterOptions).start(
-      rehydratedState ? State.create(rehydratedState) : undefined
-    );
+  const service = useConstant(() => {
```
Andarist, line 151: you could combine those 2 `useConstant` calls and just return a tuple

packages/xstate-react/src/useMachine.ts
```diff
@@ -133,7 +133,7 @@ export function useMachine<
     ...interpreterOptions
   } = options;
 
-  const service = useConstant(() => {
+  const customMachine = useConstant(() => {
```
Andarist, line 136: nit: custom sounds off here to me, maybe one of final/created/merged? or just `machine` + renaming the top-level parameter to `inputMachine`?

packages/xstate-react/src/useMachine.ts
```diff
@@ -163,44 +166,47 @@ export function useMachine<
...
-    });
-
-    // if service.state has not changed React should just bail out from this update
-    setState(service.state);
+    service
+      .onTransition((currentState) => {
+        // Only change the current state if:
+        // - the incoming state is the initial state (since it might have new actors)
+        // - OR the incoming state actually changed
+        if (currentState.changed || currentState.changed === undefined) {
```
Andarist, line 174: q: `|| currentState.changed === undefined` when this matters? if this matters in some situation then this should be covered by a test (and I don't think any of those that were added here cover this), seems like an easy thing to break by accident in the future

packages/xstate-react/src/useMachine.ts
```diff
@@ -163,44 +166,47 @@ export function useMachine<
...
+          );
+
+          layoutEffectActionsRef.current.push(
+            ...layoutEffectActions.map<ActionStateTuple<TContext, TEvent>>(
+              (layoutEffectAction) => [layoutEffectAction, currentState]
+            )
+          );
+        }
+      })
+      .start(rehydratedState ? State.create(rehydratedState) : undefined);
```
Andarist, line 209: we have already checked `rehydratedState` when initializing the state, could we just pass `state` here? 🤔 or do we have to recreate `rehydratedState` again?

packages/xstate-react/src/useMachine.ts
```diff
@@ -163,44 +166,47 @@ export function useMachine<
...
+            (action): action is ReactActionObject<TContext, TEvent> => {
+              return (
+                typeof action.exec === 'function' &&
+                '__effect' in
+                  (action as ReactActionObject<TContext, TEvent>).exec
+              );
+            }
+          );
+
+          const [effectActions, layoutEffectActions] = partition(
```
Andarist, line 189: could you add a test that those initial "react" actions are executed in the very first commit phase of a component? from what I see it works like this (which is IMHO good), but this also seems like a potentially fragile logic, so would be great to have a test for this

packages/xstate-react/src/useService.ts
```diff
@@ -1,39 +1,30 @@
...
-        return unsubscribe;
-      }
-    }),
-    [service]
-  );
-
-  const state = useSubscription(subscription);
+): [State<TContext, TEvent, any, TTypestate>, Sender<TEvent>] {
+  const serviceActor = useMemo(() => fromService(service), [service]);
+  const [state, send] = useActor<TEvent, State<TContext, TEvent>>(serviceActor);
```
Andarist, line 27: nit: we could return here directly to avoid one extra array allocation

packages/xstate-react/src/useService.ts
```diff
@@ -1,39 +1,30 @@
...
+
+export function fromService<TContext, TEvent extends EventObject>(
+  service: Interpreter<TContext, any, TEvent, any>
+): ActorRef<TEvent, State<TContext, TEvent>> {
+  const { machine } = service;
+  return {
+    send: service.send.bind(service),
+    subscribe: service.subscribe.bind(service),
+    stop: service.stop,
+    current: service.state || machine.initialState,
```
Andarist, line 14: This logs a warning:
https://github.com/davidkpiano/xstate/blob/1273d72447af45b6c92da9b002f82619f566ebb7/packages/core/src/interpreter.ts#L236-L242

You have to check the status of a service before reading this.

packages/xstate-react/src/useService.ts
```diff
@@ -1,39 +1,30 @@
...
+import { ActorRef } from './types';
+
+export function fromService<TContext, TEvent extends EventObject>(
+  service: Interpreter<TContext, any, TEvent, any>
+): ActorRef<TEvent, State<TContext, TEvent>> {
+  const { machine } = service;
+  return {
+    send: service.send.bind(service),
+    subscribe: service.subscribe.bind(service),
+    stop: service.stop,
```
Andarist, line 13: general q: should this always be exposed? should any owner of an actor ref be allowed to stop a particular actor? or should this be restricted only to the immediate parent?

packages/xstate-react/src/useService.ts
```diff
@@ -1,39 +1,30 @@
...
+import { EventObject, State, Interpreter, Typestate, Sender } from 'xstate';
+import { useActor } from './useActor';
+import { ActorRef } from './types';
+
+export function fromService<TContext, TEvent extends EventObject>(
+  service: Interpreter<TContext, any, TEvent, any>
+): ActorRef<TEvent, State<TContext, TEvent>> {
+  const { machine } = service;
+  return {
+    send: service.send.bind(service),
```
Andarist, line 11: similarly, this can trigger a warning when a child component sends to it from within first `useEffect`:
https://github.com/davidkpiano/xstate/blob/1273d72447af45b6c92da9b002f82619f566ebb7/packages/core/src/interpreter.ts#L572-L586

Maybe `@xstate/react` should always add `deferEvents: true` to the interpreter?

packages/xstate-react/src/useActor.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export function useActor<TEvent extends EventObject, TEmitted = any>(
+  actorLike: ActorRefLike<TEvent, TEmitted> | Actor
+): [TEmitted, Sender<TEvent>] {
+  const actor = useMemo(() => resolveActor(actorLike), [actorLike]);
+  const deferredEventsRef = useRef<EventObject[]>([]);
+  const [current, setCurrent] = useState(actor.current);
+
+  const send: Sender<TEvent> = useCallback(
```
Andarist, line 39: does this work right now? I can't find anywhere in the code `.deferred` being switched to false

packages/xstate-react/src/useActor.ts
```diff
@@ -0,0 +1,69 @@
...
+      subscribe: actorLike.subscribe as any, // TODO: fix (subscribe will work, just improperly typed)
+      name: actorLike.id!,
+      current: actorLike.state!
+    };
+  }
+
+  return actorLike;
+}
+
+export function useActor<TEvent extends EventObject, TEmitted = any>(
```
Andarist, line 32: some tests would be great for this, especially given that `resolveActor` accepts quite a few variations of Actors and it's quite hard to see right now how it works with all of those variations

packages/xstate-react/src/useActor.ts
```diff
@@ -0,0 +1,69 @@
...
+  actorLike: ActorRefLike<TEvent, TEmitted> | Actor
+): [TEmitted, Sender<TEvent>] {
+  const actor = useMemo(() => resolveActor(actorLike), [actorLike]);
+  const deferredEventsRef = useRef<EventObject[]>([]);
```
Andarist, line 36: right now this is owned by a hook, but not by an actor (or in other words - there is no direct mapping between this array and an actor), so there is a potential of an actor changing in between flushing or sending to it and some events might end up being sent to a new actor, whereas all scheduled events should only target the one that was here at the time when `send` got called

packages/core/src/Actor.ts
```diff
@@ -33,15 +36,22 @@ export function createNullActor(id: string): Actor {
...
  * @param invokeDefinition The meta information needed to invoke the actor.
  */
 export function createInvocableActor<TC, TE extends EventObject>(
-  invokeDefinition: InvokeDefinition<TC, TE>
+  invokeDefinition: InvokeDefinition<TC, TE>,
+  machine: StateMachine<TC, any, TE>
 ): Actor {
   const tempActor = createNullActor(invokeDefinition.id);
+  const serviceCreator = machine.options.services?.[invokeDefinition.src];
+  tempActor.deferred = true;
```
Andarist, line 50: what does this property mean for `xstate` itself? right now it is just used by `@xstate/react`

packages/xstate-react/src/useActor.ts
Andarist: oh, ok - i think I understand this now - the identity of a null actor is still replaced after we start a machine, but this time it has no deferred property at all

Andarist:
> This means that when useLayoutEffect runs, service.start() will be called and the component will be rendered again with the interpreted initial state, which may be different than the first initial state.

This IMHO would be worth optimizing in the future. When there is no work to be done, we should avoid it.

---

So a caveat of this solution right now is that a user should always interact with actors using `useActor`, just grabbing a reference to an actor and sending to it directly can be considered an anti-pattern. Did I get this right?

packages/xstate-react/test/useService.test.tsx
davidkpiano: I actually think that `StrictMode` should be used for _all_ the tests if we want to catch unforeseen issues with CM.

packages/xstate-react/src/useMachine.ts
davidkpiano: It does matter because it would represent the second "live" initial state, and it's covered by the immediately-available test.

packages/xstate-react/src/useService.ts
davidkpiano: It should in theory be "restricted" but it should be exposed to `useService`.

packages/xstate-react/src/useActor.ts
davidkpiano: Yep, that's correct.

packages/xstate-react/src/useActor.ts
davidkpiano: > so there is a potential of an actor changing in between flushing or sending to it and some events might end up being sent to a new actor

That is exactly the point of `deferredEventsRef` - we _want_ to send deferred events to the _new_ live actor.

packages/core/src/Actor.ts
davidkpiano: Doesn't mean anything for XState - it's just used in v4 as a stopgap.

davidkpiano:
> So a caveat of this solution right now is that a user should always interact with actors using useActor, just grabbing a reference to an actor and sending to it directly can be considered an anti-pattern. Did I get this right?

Yep, that's right.

packages/xstate-react/src/useActor.ts
Andarist: Right, I was commenting about a different case - but it's hard to solve with the current design and swapping actors can be also considered an anti-pattern, so nevermind for now.

packages/xstate-react/test/useService.test.tsx
Andarist: fair point

packages/xstate-react/src/useMachine.ts
Andarist: oh, ok - it wasnt that clear to me as per: https://github.com/davidkpiano/xstate/pull/1239#discussion_r445579359 , I think that split mentioned there would make this clearer

packages/xstate-react/src/useMachine.ts
davidkpiano: The logic is a little bit different (and causes tests to fail) - we need to have the logic that the initial state is `undefined` so that the service doesn't try to "reuse" the initial state.

Probably can be refactored in a later PR.

packages/xstate-react/src/useMachine.ts
davidkpiano: What would a test for this look like? Might be a bit too implementation-detail oriented. The `'layout effects should happen after normal actions'` test covers the order, ensuring that a `layout effect` happens after a `non-effect` and a normal `effect`.

packages/xstate-react/src/useService.ts
davidkpiano: Good question - currently it's part of the interface, but yes, an actor can stop at any time for any reason. Let's leave this in here for now, since we don't expose `.stop()` to the `useActor` consumer.

packages/xstate-react/src/useActor.ts
davidkpiano: I want to merge this in first and flesh out how it will work with the current V5 implementation - will write tests there.

packages/xstate-react/src/useActor.ts
davidkpiano: We'll re-examine this for V5.

packages/xstate-react/src/useMachine.ts
Andarist: > The 'layout effects should happen after normal actions' test covers the order, ensuring that a layout effect happens after a non-effect and a normal effect.

Oh, forgot about this test as it isn't part of this PR specifically. It is focused on order between normal and effect actions though - we only know from it that effect actions execute **after** regular actions, but we don't quite ensure when they execute.

You could extend that existing test or write a new one like this:

```js
  it('initial effect actions should execute during the very first commit phase', (done) => {
    let commitPhaseCounter = 0;

      {
        initial: 'active',
        states: {
          active: {
            entry: [
              asLayoutEffect(() => {
                expect(commitPhaseCounter).toBe(1);
              }),
              asEffect(() => {
                expect(commitPhaseCounter).toBe(1);
              }),
            ]
          }
        }
      },
    );

    const App = () => {
      React.useLayoutEffect(() => {
        commitPhaseCounter++
      })
      useMachine(machine);

      return <div />;
    };

    render(<App />);
    done();
  });
```

packages/xstate-react/src/useMachine.ts
```diff
@@ -142,18 +148,22 @@ export function useMachine<
...
       ...machine.context,
       ...context
     } as TContext);
 
-    return interpret(createdMachine, interpreterOptions).start(
-      rehydratedState ? State.create(rehydratedState) : undefined
-    );
+    return [
+      resolvedMachine,
+      interpret(resolvedMachine, { deferEvents: true, ...interpreterOptions })
```
Andarist, line 158: I **think** (not 100% sure) that this change has actually fixed the test mentioned here:
https://github.com/davidkpiano/xstate/pull/1239#discussion_r445583363

Would be great to check if it would indeed fail without this and add it if it does (still might be worth adding regardless as it's a slight variation over the existing test, but that would be up to you)

packages/xstate-react/test/useService.test.tsx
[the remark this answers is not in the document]
davidkpiano: Added

packages/xstate-react/src/useMachine.ts
davidkpiano: Test was added!

packages/xstate-react/src/useMachine.ts
davidkpiano: Added that test 👍

davidkpiano:
> > This means that when useLayoutEffect runs, service.start() will be called and the component will be rendered again with the interpreted initial state, which may be different than the first initial state.
> 
> This IMHO would be worth optimizing in the future. When there is no work to be done, we should avoid it.

Optimized in 40f8084

PR #1240, opened 2020-06-15
[v5] Enable additional SCXML tests

davidkpiano:


davidkpiano:
@Andarist Is this one good to go?

packages/core/src/interpreter.ts
```diff
@@ -556,9 +568,14 @@ export class Interpreter<
 
     if (!target) {
       if (!isParent) {
-        throw new Error(
+        const executionError = new Error(
           `Unable to send event to child '${to}' from service '${this.id}'.`
         );
+        this.send(
+          toSCXMLEvent<TEvent>(actionTypes.errorExecution, {
```
Andarist, line 575: Technically if we want to match SCXML this should be `error.communication`:
> If the sending SCXML session specifies a session that does not exist or is inaccessible, the SCXML Processor must place the error error.communication on the internal event queue of the sending session.

packages/core/src/scxml.ts
```diff
@@ -369,31 +370,63 @@ function toConfig(
...
-          ? {
-              cond: createCond(value.attributes!.cond as string)
+          const targets = getAttribute(value, 'target');
+          const internal = getAttribute(value, 'type') === 'internal';
+
+          let condObject = {};
+
+          if (value.attributes?.cond) {
+            const cond = value.attributes!.cond;
+            // console.log(cond);
```
Andarist, line 394:
```suggestion
```

packages/core/src/updateContext.ts
```diff
@@ -29,6 +29,12 @@ export function updateContext<TContext, TEvent extends EventObject>(
   }
   const capturedActions: InvokeActionObject[] = [];
 
+  if (!context) {
```
Andarist, line 32: dont we already warn about this 4 LOCs above?

packages/core/test/scxml.test.ts
```diff
@@ -156,19 +147,19 @@ const testGroups = {
...
     'test185.txml',
-    // 'test186.txml', // not sure yet why
+    'test186.txml',
     'test187.txml',
     'test189.txml',
-    // 'test190.txml', // _sessionid not yet available for expressions
+    'test190.txml', // note: _sessionid is undefined for expressions
     'test191.txml',
-    // 'test192.txml', // conversion of #_invokeid not implemented yet
+    // 'test192.txml', // done.invoke inexact event descriptor
```
Andarist, line 158: isnt the removed comment still relevant? i don't think we convert sends' targets yet

packages/core/test/scxml.test.ts
```diff
@@ -207,7 +198,7 @@ const testGroups = {
     // 'test278.txml', // non-root datamodel with early binding not implemented yet
     // 'test279.txml', // non-root datamodel with early binding not implemented yet
     // 'test280.txml', // non-root datamodel with late binding not implemented yet
-    // 'test286.txml', // error.execution when evaluating assign
+    'test286.txml', // error.execution when evaluating assign
```
Andarist, line 201:
```suggestion
    'test286.txml',
```

packages/core/test/scxml.test.ts
```diff
@@ -254,16 +245,16 @@ const testGroups = {
...
     'test364.txml',
-    // 'test372.txml', // microstep not implemented correctly
+    // 'test372.txml', // microstep not implemented correctly for final states
     'test375.txml',
     // 'test376.txml', // executable blocks not implemented
     'test377.txml',
     // 'test378.txml', // executable blocks not implemented
     'test387.txml',
-    // 'test388.txml', // computed historyValue not being available immediately after exiting states for the following synchronous enterStates
+    'test388.txml',
```
Andarist, line 254: hm, when this was fixed? 🤔 just wondering if maybe this doesn't pass false positively now?

packages/core/test/scxml.test.ts
```diff
@@ -295,7 +286,7 @@ const testGroups = {
     // 'test457.txml', // <foreach> not implemented yet
     // 'test459.txml', // <foreach> not implemented yet
     // 'test460.txml', // <foreach> not implemented yet
-    // 'test487.txml', // error.execution when evaluating assign
+    'test487.txml', // error.execution when evaluating assign
```
Andarist, line 289:
```suggestion
    'test487.txml',
```

packages/core/test/stateIn.test.ts
```diff
@@ -14,11 +15,11 @@ const machine = Machine({
             },
             EVENT2: {
               target: 'a2',
-              in: { b: 'b2' }
+              cond: stateIn({ b: 'b2' })
```
Andarist, line 18: we can remove or at least deprecated `in:` now, right?

packages/core/src/guards.ts
```diff
@@ -0,0 +1,35 @@
+import { EventObject, StateValue, GuardPredicate } from './types';
+import { isStateId } from './stateUtils';
+import { isString } from 'util';
+
+export function stateIn<TContext, TEvent extends EventObject>(
+  stateValue: StateValue
+): GuardPredicate<TContext, TEvent> {
+  return {
+    type: 'xstate.guard',
+    name: 'In',
```
Andarist, line 10: `In()` should be available to all conditional expressions and that includes `<if>` & `<elseif>` represented by `choose`, would be great to add support for this and some basic tests

Andarist: Note: previously `in:` was working with some relative paths, this is not. I've raised concerns about `in:` in the past though because the given string was checked "by relative grandparent" which was different than relative targets for transitions and this was IMHO quite confusing.

packages/core/test/scxml.test.ts
davidkpiano: The more pertinent issue is that `done.invoke` is an inexact event descriptor.

packages/core/test/scxml.test.ts
davidkpiano: Probably with the microstep refactor; if it is indeed a false positive, we'll see it fail with further refactoring, or see similar tests fail. Not sure if it's worth it to deeply inspect a passing test.

packages/core/test/stateIn.test.ts
davidkpiano: Almost; we have to decide what we want to do with regard to "relative-by-grandparent" state nodes.

packages/core/src/updateContext.ts
davidkpiano: Yes, we can remove the warning.

packages/core/src/interpreter.ts
davidkpiano: Not exactly; in this case, it's that the `target` is invalid. See `test194.txml.scxml`:

```xml
<?xml version="1.0" encoding="UTF-8"?><!-- we test that specifying an illegal target for <send> causes the event error.execution to be raised.  If it does,
we succeed.  Otherwise we eventually timeout and fail.  --><scxml xmlns="http://www.w3.org/2005/07/scxml" xmlns:conf="http://www.w3.org/2005/scxml-conformance" initial="s0" version="1.0" datamodel="ecmascript">

  
<state id="s0">
  <onentry>
   <!-- should cause an error -->
   <send target="baz" event="event2"/>
   <!-- this will get added to the external event queue after the error has been raised -->
   <send event="timeout"/>
    </onentry>
 
 <!-- once we've entered the state, we should check for internal events first -->   
  <transition event="error.execution" target="pass"/>
  <transition event="*" target="fail"/>
 </state>
 

<final id="pass"><onentry><log label="Outcome" expr="'pass'"/></onentry></final>
<final id="fail"><onentry><log label="Outcome" expr="'fail'"/></onentry></final>

</scxml>
```

That test will fail if I change it to `error.communication`.

packages/core/src/guards.ts
davidkpiano: Yeah, let's deprecate "by relative grandparent" in a separate PR.

davidkpiano: See card: https://github.com/davidkpiano/xstate/projects/1#card-40695477

packages/core/test/stateIn.test.ts
davidkpiano: See card; we'll deprecate `in:` in a separate PR https://github.com/davidkpiano/xstate/projects/1#card-40695477

packages/core/src/interpreter.ts
Andarist: You are right. Not sure when `error.communication` would ever be dispatched then 🤷‍♂️

davidkpiano: > You are right. Not sure when error.communication would ever be dispatched then 🤷‍♂️

I think once we get into the XState-on-the-server side of things, `error.communication` will be more relevant. E.g., a remote statechart might exist, but can return error codes if it's down. Or the `ActorRef` can indicate when communicating with it failed.

Andarist:
I forgot - would be great to create a changeset about user-facing changes.

PR #1246, opened 2020-06-17
Parent sibling state access

MysticEyeBlink:
Using YAML with anchors as a machine declaration is a pain, when you are forced to jump to state on totally different level. This MR solves problem with aliased nodes, which cannot have ID (because of duplication) and allows set state from parent states collection. 

If you have better/more descriptive naming proposition we should change it. 

In relation to: https://github.com/davidkpiano/xstate/issues/52

### DIFF
--- a/.changeset/breezy-poems-hang.md
+++ b/.changeset/breezy-poems-hang.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Transition to a sibling of a parent state
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -99,13 +99,13 @@ import { Actor, createInvocableActor } from './Actor';
 const NULL_EVENT = '';
 const STATE_IDENTIFIER = '#';
 const WILDCARD = '*';
-const SIBLING_STATE_IDENTIFIER = '^';
+const PARENT_SIBLING_STATE_IDENTIFIER = '^';
 
 const EMPTY_OBJECT = {};
 
 const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
-const isSiblingStateReference = (str: string) =>
-  str[0] === SIBLING_STATE_IDENTIFIER;
+const isParentSiblingStateReference = (str: string) =>
+  str[0] === PARENT_SIBLING_STATE_IDENTIFIER;
 const createDefaultOptions = <TContext>(): MachineOptions<TContext, any> => ({
   actions: {},
   guards: {},
@@ -1341,11 +1341,11 @@ class StateNode<
   }
 
   /**
-   * Gives a possibility to jump out from state and set one of the state siblings
+   * Gives a possibility to jump out from state and set one of the parent state siblings
    *
-   * @param statePath The string or string array sibling path to the state node. The prefix "^" is removed.
+   * @param statePath The string or string array parent sibling path to the state node. The prefix "^" is removed.
    */
-  public getSiblingStateNodeByPath(
+  public getParentSiblingStateNodeByPath(
     statePath: string | string[]
   ): StateNode<TContext, any, TEvent> {
     let stateNode: StateNode<TContext, any, TEvent> | undefined;
@@ -1356,7 +1356,7 @@ class StateNode<
 
     if (!stateNode) {
       throw new Error(
-        `Sibling state node '^${statePath}' does not exist on machine '${this.id}'`
+        `Parent sibling state node '^${statePath}' does not exist on machine '${this.id}'`
       );
     }
 
@@ -1371,9 +1371,12 @@ class StateNode<
   public getStateNodeByPath(
     statePath: string | string[]
   ): StateNode<TContext, any, TEvent> {
-    if (typeof statePath === 'string' && isSiblingStateReference(statePath)) {
+    if (
+      typeof statePath === 'string' &&
+      isParentSiblingStateReference(statePath)
+    ) {
       try {
-        return this.getSiblingStateNodeByPath(statePath.slice(1));
+        return this.getParentSiblingStateNodeByPath(statePath.slice(1));
       } catch (e) {
         // try id or individual paths
         // throw e;
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -99,10 +99,13 @@ import { Actor, createInvocableActor } from './Actor';
 const NULL_EVENT = '';
 const STATE_IDENTIFIER = '#';
 const WILDCARD = '*';
+const SIBLING_STATE_IDENTIFIER = '^';
 
 const EMPTY_OBJECT = {};
 
 const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
+const isSiblingStateReference = (str: string) =>
+  str[0] === SIBLING_STATE_IDENTIFIER;
 const createDefaultOptions = <TContext>(): MachineOptions<TContext, any> => ({
   actions: {},
   guards: {},
@@ -1338,6 +1341,29 @@ class StateNode<
   }
 
   /**
+   * Gives a possibility to jump out from state and set one of the state siblings
+   *
+   * @param statePath The string or string array sibling path to the state node. The prefix "^" is removed.
+   */
+  public getSiblingStateNodeByPath(
+    statePath: string | string[]
+  ): StateNode<TContext, any, TEvent> {
+    let stateNode: StateNode<TContext, any, TEvent> | undefined;
+
+    if (this.parent !== undefined) {
+      stateNode = this.parent.getStateNodeByPath(statePath);
+    }
+
+    if (!stateNode) {
+      throw new Error(
+        `Sibling state node '^${statePath}' does not exist on machine '${this.id}'`
+      );
+    }
+
+    return stateNode;
+  }
+
+  /**
    * Returns the relative state node from the given `statePath`, or throws.
    *
    * @param statePath The string or string array relative path to the state node.
@@ -1345,6 +1371,15 @@ class StateNode<
   public getStateNodeByPath(
     statePath: string | string[]
   ): StateNode<TContext, any, TEvent> {
+    if (typeof statePath === 'string' && isSiblingStateReference(statePath)) {
+      try {
+        return this.getSiblingStateNodeByPath(statePath.slice(1));
+      } catch (e) {
+        // try id or individual paths
+        // throw e;
+      }
+    }
+
     if (typeof statePath === 'string' && isStateId(statePath)) {
       try {
         return this.getStateNodeById(statePath.slice(1));
--- a/packages/core/test/parentSibling.test.ts
+++ b/packages/core/test/parentSibling.test.ts
@@ -47,7 +47,7 @@ const idMachine = Machine({
   }
 });
 
-describe('State node sibling state selector', () => {
+describe('Parent sibling state selector', () => {
   const expected = {
     A: {
       NEXT: 'A.dot'
--- a/packages/core/test/sibling.test.ts
+++ b/packages/core/test/sibling.test.ts
@@ -1,0 +1,67 @@
+import { testAll } from './utils';
+import { Machine } from '../src';
+
+const idMachine = Machine({
+  initial: 'A',
+  states: {
+    A: {
+      initial: 'foo',
+      states: {
+        foo: {
+          initial: 'bar',
+          states: {
+            bar: {
+              on: {
+                NEXT: '^dot'
+              }
+            }
+          }
+        },
+        dot: {
+          on: {
+            NEXT: '^B'
+          }
+        }
+      }
+    },
+    B: {
+      initial: 'foo',
+      states: {
+        foo: {
+          initial: 'baz',
+          states: {
+            baz: {
+              on: {
+                NEXT: '^dot'
+              }
+            }
+          }
+        },
+        dot: {
+          on: {
+            NEXT: '^A.foo.bar'
+          }
+        }
+      }
+    }
+  }
+});
+
+describe('State node sibling state selector', () => {
+  const expected = {
+    A: {
+      NEXT: 'A.dot'
+    },
+    'A.dot': {
+      NEXT: 'B'
+    },
+    B: {
+      NEXT: 'B.dot'
+    },
+    'B.dot': {
+      NEXT: 'A.foo.bar'
+    }
+  };
+
+  testAll(idMachine, expected);
+});


Andarist:
> Using YAML with anchors as a machine declaration is a pain, when you are forced to jump to state on totally different level.

Could you describe this use case in more detail? Preferably with some examples?

kkwiatkowski:
Hey @Andarist 
## Background
- The whole application is driven by xstate - every UI and logical interaction is handled in such a way.
- Depending on state machine organization you might either parallelize or nest states - we're mostly nesting.
- Configuration for xstate is stored in serializable format - in our case these are YAML files.
- YAML has been chosen as a way to reduce the amount of configuration repetition by using YAML anchors.
- State machine configuration can freely navigate to nested states, however it is impossible to relatively navigate back to parent states.
- Xstate already supports absolute transition given the state id, nevertheless given that our state machine configuration is generated out of YAML we're not able to name each state with its unique id.
- The above is the reason for this contribution (we also have seen that the topic was raised in the past and absolute state id was introduced back then).
- We're not decomposing machines into service state machines as it introduces lots of complexity around maintainence, context sharing and events processing, so we have a single config.

## Example
```yaml
# Given we have some generic state definitions
_shared:
  displayingMessagePrompt: &displayingMessagePrompt
    initial: rendering
    states:
      rendering:
        entry: showMessagePrompt # showMessagePrompt action uses context's property "promptMessage"
        on:
          '': awaitingUserResponse

      awaitingUserResponse:
        initial: initializing
        states:
          initializing:
            invoke: getPromptTimeout # a time after which user is able to confirm or reject prompt
            onDone: displaying
          
          displaying:
            entry:
              - savePromptTimeout # saves received prompt timeout
              - startTimer # starts timer in background
              - showTimer
            on:
              TIMER_FINISHED: awaiting
          
          awaiting:
            on:
              PROMPT_CONFIRMED: ^confirming # [1]
              PROMPT_REJECTED: ^finishing # [1]

      confirming:
        invoke:
          src: saveUserConfirmation
          onDone: finishing

      finishing:
        entry:
          type: saveToContext # equvalent of "assign" action
          data: { promptMessage: '' }
        exit: hideMessagePrompt
        type: final

# Root state machine definition
initial: idle
context: { promptMessage: '' }
states:
  idle:
    on:
      LOGOUT: loggingOut
      DISCARD_CHANGES: discardingChanges

  loggingOut:
    entry:
      type: saveToContext # equvalent of "assign" action
      data: { promptMessage: 'Do you want to logout?' }
    onDone:
      actions: logout
      target: idle
    <<: *displayingMessagePrompt

  discardingChanges:
    entry:
      type: saveToContext # equvalent of "assign" action
      data: { promptMessage: 'Your progress will be lost, do you want to continue?' }
    onDone:
      actions: resetCurrentProgress
      target: idle
    <<: *displayingMessagePrompt
```
[1] That's the case - HSMs/Statecharts gives the great modeling power and we can encapsulate states, but in this case, given the *displayingMessagePrompt* is used relatively in two places we're unable to mark absolutely specific state. That's why the relative to parent state transition would be really helpful.

davidkpiano:
@kkwiatkowski Really like the YAML format! Makes it look a lot cleaner. Anyway, for some use-cases, `onDone` and final states can/should be used instead:

```diff
"awaitingUserResponse": {
  "initial": "initializing",
  "states": {
    // ...
    "awaiting": {
      "on": {
-       "PROMPT_CONFIRMED": "^confirming",
+       "PROMPT_CONFIRMED": "confirmed",
        "PROMPT_REJECTED": "^finishing"
      }
    },
+   "confirmed": { "type": "final" }
  },
+ "onDone": "confirming"
},
"confirming": {
```

But that raises questions about `"PROMPT_REJECTED": "^finishing"`, since now this state being "done" can potentially go to a different state. You can add a guard to direct it to the right state, but hmm... that's a bit more indirect than just transitioning to the parent.

Just thinking out loud here. Still not sure if we want to encourage child -> aunt/uncle transitions directly, when the above pattern should be encouraged instead. 🤔

kkwiatkowski:
Thanks for quick reply @davidkpiano!

Yes, the indirect case with extra `final` and `onDone` combination requires to extend the definition with artificial states and apply another routing rules with guards, so we have extra as many states and conditionals as we have potential transitions. That's one of the reasons why this MR is submitted (right now we're doing similar work around).

Despite the fact it is doable, it is also bloating the definitions making the transitions more imperative - whereas at least from my standpoint it should favor declarativeness.

After all, xstate attached my attention due to its declarative modeling power. You can model your application logic on quite high level description not touching any code allowing to interact with the configs the business analysts and/or product owners (that's my little sin here :smile:).

kkwiatkowski:
I also think that the clear distinction between when to use which pattern (either `final` and `onDone` or `child -> aunt/uncle` transition) should be based on who actually has a control over the state.

This whole construct might sound a bit artificial as it is not occurring unless you do the configuration without coding.

In the given example, it is __displayingMessagePrompt__ which is owning the state, and `final` with `onDone` seems to be redundant. This state cannot/shouldn't also transition the client that attached its state fragment, therefore there should't be something like __^idle__.

On the other hand, the client code to __displayingMessagePrompt__ has no control over what's happening "inside", and the only thing it can do is hook up on the final state event (routing to specific states if necessary).

kkwiatkowski:
@davidkpiano - one more comment here - doing extra routing with states and conditionals means we're implementing manually an extra state machine to our xstate hsm definition (conditional transition based on state) - the difference is that the routing criterion has to be stored in context. Below I attach the most optimal example I could come up with with that extra routing you were mentioning:


```diff
      # ...
      awaitingUserResponse:
        initial: initializing
        states:
          # ...
          awaiting:
            on:
-              PROMPT_CONFIRMED: ^confirming
+              PROMPT_CONFIRMED:
+                actions:
+                  type: saveToContext # equivalent of "assign" action
+                  data: { childToParentRedirectionState: confirming }
+                target: childFinalState # childFinalState -> confirming
-              PROMPT_REJECTED: ^finishing
+              PROMPT_REJECTED:
+                actions:
+                  type: saveToContext # equivalent of "assign" action
+                  data: { childToParentRedirectionState: finishing }
+                target: childFinalState # childFinalState -> finishing
+
+          childFinalState:
+            type: final
+
+        onDone:
+          - cond:
+              # compares the value provided in "state" property with context's "childToParentRedirectionState" stored earlier
+              type: childToParentRedirectionStateEqualsTo 
+              state: confirming
+            actions:
+              type: saveToContext # equivalent of "assign" action
+              data: { childToParentRedirectionState: '' } # context cleanup
+            target: confirming  # actual redirection
+          - cond:
+              type: childToParentRedirectionStateEqualsTo
+              state: finishing
+            actions:
+              type: saveToContext # equivalent of "assign" action
+              data: { childToParentRedirectionState: '' }
+            target: finishing # actual redirection
+          - actions:
+              type: saveToContext # equivalent of "assign" action
+              data: { childToParentRedirectionState: '' }
+            target: parentFinalState # fallback case
+
+      parentFinalState: # final state if the routing didn't match any desired "redirection"
+        type: final

      confirming:
        invoke:
          src: saveUserConfirmation
          onDone: finishing

      finishing:
        entry:
          # ...
```

Full refined YAML code:

```yaml
# Given we have some generic state definitions
_shared:
  displayingMessagePrompt: &displayingMessagePrompt
    initial: rendering
    states:
      rendering:
        entry: showMessagePrompt # showMessagePrompt action uses context's property "promptMessage"
        on:
          '': awaitingUserResponse

      awaitingUserResponse:
        initial: initializing
        states:
          initializing:
            invoke: getPromptTimeout # a time after which user is able to confirm or reject prompt
            onDone: displaying

          displaying:
            entry:
              - savePromptTimeout # saves received prompt timeout
              - startTimer # starts timer in background
              - showTimer
            on:
              TIMER_FINISHED: awaiting

          awaiting:
            on:
              PROMPT_CONFIRMED:
                actions:
                  type: saveToContext # equivalent of "assign" action
                  data: { childToParentRedirectionState: confirming }
                target: childFinalState # childFinalState -> confirming
              PROMPT_REJECTED:
                actions:
                  type: saveToContext # equivalent of "assign" action
                  data: { childToParentRedirectionState: finishing }
                target: childFinalState # childFinalState -> finishing

          childFinalState:
            type: final

        onDone:
          - cond:
              # compares the value provided in "state" property with context's "childToParentRedirectionState" stored earlier
              type: childToParentRedirectionStateEqualsTo
              state: confirming
            actions:
              type: saveToContext # equivalent of "assign" action
              data: { childToParentRedirectionState: '' }
            target: confirming # actual redirection
          - cond:
              type: childToParentRedirectionStateEqualsTo
              state: finishing
            actions:
              type: saveToContext # equivalent of "assign" action
              data: { childToParentRedirectionState: '' }
            target: finishing # actual redirection
          - actions:
              type: saveToContext # equivalent of "assign" action
              data: { childToParentRedirectionState: '' }
            target: parentFinalState # fallback case

      parentFinalState: # final state if the routing didn't match any desired "redirection"
        type: final

      confirming:
        invoke:
          src: saveUserConfirmation
          onDone: finishing

      finishing:
        entry:
          type: saveToContext # equivalent of "assign" action
          data: { promptMessage: '' }
        exit: hideMessagePrompt
        type: final

# Root state machine definition
initial: idle
context: { promptMessage: '' }
states:
  idle:
    on:
      LOGOUT: loggingOut
      DISCARD_CHANGES: discardingChanges

  loggingOut:
    entry:
      type: saveToContext # equivalent of "assign" action
      data: { promptMessage: 'Do you want to logout?' }
    onDone:
      actions: logout
      target: idle
    <<: *displayingMessagePrompt

  discardingChanges:
    entry:
      type: saveToContext # equivalent of "assign" action
      data: { promptMessage: 'Your progress will be lost, do you want to continue?' }
    onDone:
      actions: resetCurrentProgress
      target: idle
    <<: *displayingMessagePrompt
```

Transformed YAML to JSON (you can use an online tool like https://codebeautify.org/yaml-to-json-xml-csv):
```json
{
  "initial": "idle",
  "context": {
    "promptMessage": ""
  },
  "states": {
    "idle": {
      "on": {
        "LOGOUT": "loggingOut",
        "DISCARD_CHANGES": "discardingChanges"
      }
    },
    "loggingOut": {
      "entry": {
        "type": "saveToContext",
        "data": {
          "promptMessage": "Do you want to logout?"
        }
      },
      "onDone": {
        "actions": "logout",
        "target": "idle"
      },
      "initial": "rendering",
      "states": {
        "rendering": {
          "entry": "showMessagePrompt",
          "on": {
            "": "awaitingUserResponse"
          }
        },
        "awaitingUserResponse": {
          "initial": "initializing",
          "states": {
            "initializing": {
              "invoke": "getPromptTimeout",
              "onDone": "displaying"
            },
            "displaying": {
              "entry": [
                "savePromptTimeout",
                "startTimer",
                "showTimer"
              ],
              "on": {
                "TIMER_FINISHED": "awaiting"
              }
            },
            "awaiting": {
              "on": {
                "PROMPT_CONFIRMED": {
                  "actions": {
                    "type": "saveToContext",
                    "data": {
                      "childToParentRedirectionState": "confirming"
                    }
                  },
                  "target": "childFinalState"
                },
                "PROMPT_REJECTED": {
                  "actions": {
                    "type": "saveToContext",
                    "data": {
                      "childToParentRedirectionState": "finishing"
                    }
                  },
                  "target": "childFinalState"
                }
              }
            },
            "childFinalState": {
              "type": "final"
            }
          },
          "onDone": [
            {
              "cond": {
                "type": "childToParentRedirectionStateEqualsTo",
                "state": "confirming"
              },
              "actions": {
                "type": "saveToContext",
                "data": {
                  "childToParentRedirectionState": ""
                }
              },
              "target": "confirming"
            },
            {
              "cond": {
                "type": "childToParentRedirectionStateEqualsTo",
                "state": "finishing"
              },
              "actions": {
                "type": "saveToContext",
                "data": {
                  "childToParentRedirectionState": ""
                }
              },
              "target": "finishing"
            },
            {
              "actions": {
                "type": "saveToContext",
                "data": {
                  "childToParentRedirectionState": ""
                }
              },
              "target": "parentFinalState"
            }
          ]
        },
        "parentFinalState": {
          "type": "final"
        },
        "confirming": {
          "invoke": {
            "src": "saveUserConfirmation",
            "onDone": "finishing"
          }
        },
        "finishing": {
          "entry": {
            "type": "saveToContext",
            "data": {
              "promptMessage": ""
            }
          },
          "exit": "hideMessagePrompt",
          "type": "final"
        }
      }
    },
    "discardingChanges": {
      "entry": {
        "type": "saveToContext",
        "data": {
          "promptMessage": "Your progress will be lost, do you want to continue?"
        }
      },
      "onDone": {
        "actions": "resetCurrentProgress",
        "target": "idle"
      },
      "initial": "rendering",
      "states": {
        "rendering": {
          "entry": "showMessagePrompt",
          "on": {
            "": "awaitingUserResponse"
          }
        },
        "awaitingUserResponse": {
          "initial": "initializing",
          "states": {
            "initializing": {
              "invoke": "getPromptTimeout",
              "onDone": "displaying"
            },
            "displaying": {
              "entry": [
                "savePromptTimeout",
                "startTimer",
                "showTimer"
              ],
              "on": {
                "TIMER_FINISHED": "awaiting"
              }
            },
            "awaiting": {
              "on": {
                "PROMPT_CONFIRMED": {
                  "actions": {
                    "type": "saveToContext",
                    "data": {
                      "childToParentRedirectionState": "confirming"
                    }
                  },
                  "target": "childFinalState"
                },
                "PROMPT_REJECTED": {
                  "actions": {
                    "type": "saveToContext",
                    "data": {
                      "childToParentRedirectionState": "finishing"
                    }
                  },
                  "target": "childFinalState"
                }
              }
            },
            "childFinalState": {
              "type": "final"
            }
          },
          "onDone": [
            {
              "cond": {
                "type": "childToParentRedirectionStateEqualsTo",
                "state": "confirming"
              },
              "actions": {
                "type": "saveToContext",
                "data": {
                  "childToParentRedirectionState": ""
                }
              },
              "target": "confirming"
            },
            {
              "cond": {
                "type": "childToParentRedirectionStateEqualsTo",
                "state": "finishing"
              },
              "actions": {
                "type": "saveToContext",
                "data": {
                  "childToParentRedirectionState": ""
                }
              },
              "target": "finishing"
            },
            {
              "actions": {
                "type": "saveToContext",
                "data": {
                  "childToParentRedirectionState": ""
                }
              },
              "target": "parentFinalState"
            }
          ]
        },
        "parentFinalState": {
          "type": "final"
        },
        "confirming": {
          "invoke": {
            "src": "saveUserConfirmation",
            "onDone": "finishing"
          }
        },
        "finishing": {
          "entry": {
            "type": "saveToContext",
            "data": {
              "promptMessage": ""
            }
          },
          "exit": "hideMessagePrompt",
          "type": "final"
        }
      }
    }
  }
}
```

Visualization of above configuration:
image

kkwiatkowski:
@davidkpiano, @Andarist any further thoughts on the above?

Andarist:
I kinda feel the same as @davidkpiano:
> Just thinking out loud here. Still not sure if we want to encourage child -> aunt/uncle transitions directly, when the above pattern should be encouraged instead. 🤔

As in - I'm not sure 😅 

I totally recognize your specific problem though - even when we compare the first and the second visualization then the first one wins in terms of complexity (it's less complex), and ultimately this should be favored.

When thinking about this more I don't think `onDone` in itself is that much better conceptually, as one still has to know IDs of the ancestor's siblings. I only find the proposal to be problematic because it might encourage ppl to reach for this prematurely - when there are often other, better alternatives.

Your example is quite abstract so it's hard for me to assess if this could be modeled differently as I lack your domain's knowledge, but maybe you could just move the `awaiting` node out of the `awaitingUserResponse`? That way it could respond to events by just targeting its siblings.

davidkpiano:
Closing in favor of using IDs and idiomatic use of `onDone` over introducing extra syntax

PR #1249, opened 2020-06-19
Add `always`

davidkpiano:
This PR adds support for the `always` transition, which works exactly the same as eventless transitions:

```diff
on: {
- '': 'someTarget'
},
+always: 'someTarget'
```

As the tests show, this is not a breaking change. The original way for defining eventless transitions will continue to work.

TODO: 
- [x] Docs
- [x] Tests

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -1832,20 +1832,27 @@ class StateNode<
     } else {
       const {
         [WILDCARD]: wildcardConfigs = [],
-        ...strictOnConfigs
+        ...strictTransitionConfigs
       } = this.config.on;
 
       onConfig = flatten(
-        keys(strictOnConfigs)
+        keys(strictTransitionConfigs)
           .map((key) => {
-            const arrayified = toTransitionConfigArray<TContext, EventObject>(
-              key,
-              strictOnConfigs![key as string]
-            );
-            if (!IS_PRODUCTION) {
-              validateArrayifiedTransitions(this, key, arrayified);
+            if (!IS_PRODUCTION && key === NULL_EVENT) {
+              warn(
+                false,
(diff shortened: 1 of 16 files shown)


packages/core/src/StateNode.ts
```diff
@@ -1893,7 +1901,7 @@ class StateNode<
     const delayedTransitions = this.after;
 
     const formattedTransitions = flatten(
-      [...doneConfig, ...invokeConfig, ...onConfig].map(
+      [...doneConfig, ...invokeConfig, ...onConfig, ...eventlessConfig].map(
```
Andarist, line 1904: maybe this should be put before `onConfig` to take precedence over `''`?

---

this is definitely a quick way for doing this and is good for now, but I feel that in the future we should check if maybe moving those to a separate property would be preferable (maybe not)

Andarist:
Should we introduce a deprecation warning for `''` right away?

packages/core/src/StateNode.ts
davidkpiano: There is some complicated logic that needs to be fixed in V5 if `eventlessConfig` comes before `onConfig` - if you try it, tests won't pass if you have it the other way around.

Andarist: Hm, interesting - I would not have expected it to matter 🤔 Might take a look later at this.

davidkpiano:
Going to merge this in (all comments resolved) - won't release yet, in case a critical change is needed. @Andarist

wildskyf:
we should also update the docs here:
https://xstate.js.org/docs/guides/events.html#null-events

davidkpiano:
> we should also update the docs here:
> [xstate.js.org/docs/guides/events.html#null-events](https://xstate.js.org/docs/guides/events.html#null-events)

Feel free to make a PR for this, or an issue please

PR #1253, opened 2020-06-20
chore(dep/vue) bump peerDep composition-api to 0.6

darrenjennings:
- breaking changes in composition-api 0.6.x, so I've adapted the code to conform. This is not backwards compat with previous version of @xstate/vue but peerDependencies have been updated. Luckily the changes are more in-line with Vue 3, so this will make it easier to upgrade to Vue 3 when needed.
- Uses new watcher option since it was changed to be lazy by default
- now using shallow ref since xstate objects should not be refs deeply

### DIFF
--- a/.changeset/tiny-guests-wash.md
+++ b/.changeset/tiny-guests-wash.md
@@ -1,0 +1,10 @@
+---
+'@xstate/vue': minor
+---
+
+chore(dep/vue) bump peerDep of @vue/composition-api to 0.6.x.
+
+- breaking changes in 0.6.x, adapated code. Will make it easier for
+  this library to also be updated to Vue 3.
+- Uses new watcher option since it was changed to be lazy by default
+- now using shallow ref since xstate objects should not be refs deeply.
--- a/.changeset/tiny-guests-wash.md
+++ b/.changeset/tiny-guests-wash.md
@@ -2,9 +2,4 @@
 '@xstate/vue': minor
 ---
 
-chore(dep/vue) bump peerDep of @vue/composition-api to 0.6.x.
-
-- breaking changes in 0.6.x, adapated code. Will make it easier for
-  this library to also be updated to Vue 3.
-- Uses new watcher option since it was changed to be lazy by default
-- now using shallow ref since xstate objects should not be refs deeply.
+Upgraded package for compatibility with the newest `@vue/composition-api@^0.6.0`, which means that a peer dependency requirement has changed to this version, which is a **breaking change**. The only observable behavior change is that exposed refs are now **shallow**.
--- a/packages/xstate-vue/package.json
+++ b/packages/xstate-vue/package.json
@@ -41,7 +41,7 @@
     "url": "https://github.com/davidkpiano/xstate/issues"
   },
   "peerDependencies": {
-    "@vue/composition-api": "^0.3.4",
+    "@vue/composition-api": "^0.6.5",
     "@xstate/fsm": "^1.0.0",
     "vue": "^2.6.10",
     "xstate": "^4.7.8"
@@ -57,7 +57,7 @@
   "devDependencies": {
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/vue": "^4.1.0",
-    "@vue/composition-api": "^0.3.4",
+    "@vue/composition-api": "~0.6.5",
     "@vue/test-utils": "^1.0.0-beta.30",
     "@xstate/fsm": "*",
     "babel-core": "^6.26.3",
--- a/packages/xstate-vue/package.json
+++ b/packages/xstate-vue/package.json
@@ -41,7 +41,7 @@
     "url": "https://github.com/davidkpiano/xstate/issues"
   },
   "peerDependencies": {
-    "@vue/composition-api": "^0.6.5",
+    "@vue/composition-api": "^0.6.0",
     "@xstate/fsm": "^1.0.0",
     "vue": "^2.6.10",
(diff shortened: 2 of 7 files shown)


packages/xstate-vue/package.json
```diff
@@ -57,7 +57,7 @@
   "devDependencies": {
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/vue": "^4.1.0",
-    "@vue/composition-api": "^0.3.4",
+    "@vue/composition-api": "~0.6.5",
```
Andarist, line 60: any reason why this is using `~` over `^`?

packages/xstate-vue/package.json
```diff
@@ -41,7 +41,7 @@
     "url": "https://github.com/davidkpiano/xstate/issues"
   },
   "peerDependencies": {
-    "@vue/composition-api": "^0.3.4",
+    "@vue/composition-api": "^0.6.5",
```
Andarist, line 44: do we depend on some features in `0.6.5` specifically or could we change this to `^0.6.0`?

packages/xstate-vue/test/UseService.vue
```diff
@@ -14,20 +15,24 @@ import {
   spawn,
   doneInvoke,
   State,
-  Service
 } from 'xstate';
-import { watch, ref } from '@vue/composition-api';
+import { watchEffect, ref, Ref } from '@vue/composition-api';
```
Andarist, line 19: both `ref` and `Ref` are unused in this file, could you remove them?

packages/xstate-vue/test/UseService.vue
```diff
@@ -6,6 +6,7 @@
 </template>
 
 <script lang="ts">
+import Vue, { PropType } from 'vue'
```
Andarist, line 9: Similarly, I don't see `Vue` being used here

.changeset/tiny-guests-wash.md
```diff
@@ -0,0 +1,10 @@
+---
+'@xstate/vue': minor
+---
+
+chore(dep/vue) bump peerDep of @vue/composition-api to 0.6.x.
+
+- breaking changes in 0.6.x, adapated code. Will make it easier for
+  this library to also be updated to Vue 3.
+- Uses new watcher option since it was changed to be lazy by default
+- now using shallow ref since xstate objects should not be refs deeply.
```
Andarist, line 10: Prefixes like `chore(dep/vue)` are not needed, the changelog entry created later because of this changeset file will only appear in the `./packages/xstate-vue/CHANGELOG.md` so no need to include this information here. A changeset should also only describe user-facing changes, no need to mention that, for example, a new watcher option has been used internally - it's just an implementation detail.

I propose to change this to something like this:
```suggestion
Upgraded package for compatibility with the newest `@vue/composition-api@^0.6.0`, which means that a peer dependency requirement has changed to this version, which is a **breaking change**. The only observable behavior change is that exposed refs are now **shallow**.
```

Which assumes that previously exposed refs were deep, which I don't know if it was true nor what does it actually mean in context of Vue. Would love to learn more if you have a spare moment to explain this.

packages/xstate-vue/package.json
darrenjennings: no, changing back. I was initially going to try and provide backward compat, but not anymore

packages/xstate-vue/package.json
darrenjennings: we can change it. there were bugs caught in 0.6.0->.4 within a few days so didn't want to point to versions that were knowingly buggy but I suppose that's for the consumer to figure out? Was curious why it wasn't ^0.3.0

packages/xstate-vue/test/UseService.vue
darrenjennings: yea found these test .vue files are hard to type/lint. Will clean them up a bit.

.changeset/tiny-guests-wash.md
darrenjennings: Migrating to 0.6 _seems_ to have changed how `ref` returns were typed. So actually I think it _could_ work with traditional ref's if the type error could be resolved:

```sh
Type 'State<TContext, TEvent, TState>' is not assignable to type '{ value: string; context: UnwrapRef<TContext>; actions: ActionObject<TContext, TEvent>[]; changed?: boolean | undefined; matches: <TSV extends TState["value"]>(value: TSV) => this is TState extends { ...; } ? TState : never; }'.
  Types of property 'context' are incompatible.
    Type 'TContext' is not assignable to type 'UnwrapRef<TContext>'.
```
^^not sure how to resolve the `UnwrapRef<TContext>` typing.

A traditional ref will allow for an object's properties and sub-properties to be reactive. The "gotcha" with a shallowRef is that members are not reactive. `shallowRef`'s continue to allow `@xstate/vue` to work since all `ref` assignments used are wholesale replaced e.g. `current.value = ` or `state.value = ` (vs. `state.value.value =`). However, this is a breaking change if users were wanting to mutate return ref's properties directly and still have reactivity like today, e.g. if `state.value.value` is assigned directly instead of via `send`, it would not trigger any vue reactivity (watchers, computed, template updates etc.)

Andarist: > A traditional ref will allow for an object's properties and sub-properties to be reactive.

Interesting, is it Proxy-based?

darrenjennings: the composition-api is not proxy based, though `vue-next` (v3) is

PR #1257, opened 2020-06-25
Add note about using `async` with callback service

tmikeschu:
👋 

Re: https://github.com/davidkpiano/xstate/issues/988#issuecomment-583855715

Just wanted to make this a bit more obvious, even though it can be inferred.


davidkpiano:
Makes sense, thanks!

PR #1262, opened 2020-06-30
Node data stricter types

Andarist:
Addresses https://github.com/davidkpiano/xstate/discussions/1261

I still need to write changeset for this but I would like to get feedback first.

1. I've removed all `| any` that I could find because it's like multiplying by 0 - you always get ~0~ any, all other union members are just lost
2. Had to rename `StateNode#data` to `StateNode#doneData` because TS couldn't infer things correctly due to overlapping `.data` properties in both `InvokeConfig` and `StateNode` which are both allowed for `invoke` property. A simplified version of the problem can be checked out [here](https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAsghmSAnAPAFQMIHsB2wIAewANFGgKIBuEepaACnEnALYDOUh+OAJhwN4BfAHxQAvFAAUAY2AEAXGWx4upCJUUVqeAJTjRDJqzYAoE9IA2cNhwDKwOPgByWHhHTL8ROlRrBh-FBQJkFgAK4ARhYAltJQPI5wAPyK8IgQqJi4XiRkvrRQcDggwiaCZqCQUACSOJRYANYQygBm0QDmHtmqedr+UPwhUGxI0opswEjROO3EQwkOKbAIyF0q3r1+pEUlZRXg0K0dazk+faKDQdP1TYq1N824bZ1Z67lafqIA9F9QAD5QeyOCAuNwoHbbYqlcomFphHCyaK4KDSJAQYEnHofPDCGRPDqKI4vTxY-L+HSKIHOVzuV6nTY4gZDNHAMJIHBQHBhCwWQocHZ7cxojGBaRYeH4JCKLksCIZKCCUiBSoQRQAcgAYgB5LVqhW4y5Qa6NVVMoJBEZjKBqhJsNVzc3xRKKGRyNSUPRiC7lILlQQ6IA) after uncommenting that single comment in the `Config` type

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -225,7 +225,9 @@ class StateNode<
   /**
    * The data sent with the "done.state._id_" event if this is a final state node.
    */
-  public data?: Mapper<TContext, TEvent> | PropertyMapper<TContext, TEvent>;
+  public doneData?:
+    | Mapper<TContext, TEvent, any>
+    | PropertyMapper<TContext, TEvent, any>;
   /**
    * The string delimiter for serializing the path to a string. The default is "."
    */
@@ -373,7 +375,7 @@ class StateNode<
       this.config.exit || this.config.onExit
     ).map((action) => toActionObject(action));
     this.meta = this.config.meta;
-    this.data =
+    this.doneData =
       this.type === 'final'
         ? (this.config as FinalStateNodeConfig<TContext, TEvent>).data
         : undefined;
@@ -483,7 +485,7 @@ class StateNode<
       activities: this.activities || [],
       meta: this.meta,
       order: this.order || -1,
-      data: this.data,
+      data: this.doneData,
       invoke: this.invoke
     };
   }
@@ -959,10 +961,12 @@ class StateNode<
         }
 
         events.push(
-          done(sn.id, sn.data), // TODO: deprecate - final states should not emit done events for their own state.
+          done(sn.id, sn.doneData), // TODO: deprecate - final states should not emit done events for their own state.
           done(
             parent.id,
-            sn.data ? mapContext(sn.data, currentContext, _event) : undefined
+            sn.doneData
+              ? mapContext(sn.doneData, currentContext, _event)
+              : undefined
           )
         );
 
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -974,7 +974,7 @@ class StateNode<
               isInFinalState(transition.configuration, parentNode)
             )
           ) {
-            events.push(done(grandparent.id, grandparent.data));
+            events.push(done(grandparent.id));
           }
         }
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -258,7 +258,7 @@ export interface InvokeDefinition<TContext, TEvent extends EventObject>
    *
    * Data should be mapped to match the child machine's context shape.
    */
-  data?: Mapper<TContext, TEvent> | PropertyMapper<TContext, TEvent>;
+  data?: Mapper<TContext, TEvent, any> | PropertyMapper<TContext, TEvent, any>;
 }
 
 export interface Delay {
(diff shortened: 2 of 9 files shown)


packages/core/src/types.ts
```diff
@@ -366,53 +366,51 @@ export type TransitionsConfig<TContext, TEvent extends EventObject> =
...
-        | string
-        | SingleOrArray<TransitionConfig<TContext, DoneInvokeEvent<any>>>;
-      /**
-       * The transition to take upon the invoked child machine sending an error event.
-       */
-      onError?:
-        | string
-        | SingleOrArray<TransitionConfig<TContext, DoneInvokeEvent<any>>>;
-    }
-  | StateMachine<any, any, any>;
```
Andarist, line 415: I've moved this out of `InvokeConfig` to keep `InvokeConfig` focused on the object description of the invocable actor - it's still allowed, just defined directly on the `invoke` property of `StateNodeConfig`:
https://github.com/davidkpiano/xstate/pull/1262/files#diff-2991521f751aa9cf68976ceef512e19aR463

packages/core/src/types.ts
```diff
@@ -897,14 +897,20 @@ export type PropertyAssigner<TContext, TEvent extends EventObject> = {
...
-) => any;
+) => TParams;
 
-export type PropertyMapper<TContext, TEvent extends EventObject> = Partial<{
-  [key: string]: ((context: TContext, event: TEvent) => any) | any;
-}>;
+export type PropertyMapper<
+  TContext,
+  TEvent extends EventObject,
+  TParams extends {}
```
Andarist, line 908: actually not sure how much of a value this gives - in the codebase, this is always supplied as `any`, but in theory this could be used in userland or in the codebase in the future in some specific cases to provide stricter type for a given call

packages/core/src/StateNode.ts
```diff
@@ -974,7 +978,7 @@ class StateNode<
               isInFinalState(transition.configuration, parentNode)
             )
           ) {
-            events.push(done(grandparent.id, grandparent.data));
```
Andarist, line 977: `grandparent.data` was always undefined anyway because only `.data` of final states has been copied onto state nodes:
https://github.com/davidkpiano/xstate/pull/1262/files#diff-f8963c788047c03637a3b59926c0d45dL376-R381

This also matches SCXML spec - those done events for parallel states do not hold any data.

packages/core/src/types.ts
```diff
@@ -1166,7 +1172,7 @@ export namespace SCXML {
 
 // Taken from RxJS
 export interface Unsubscribable {
-  unsubscribe(): any | void;
+  unsubscribe(): void;
```
davidkpiano, line 1175: Hmm, I have `any` in here to prevent obscure type errors when `unsubscribe` returns something accidentally, e.g.,

```js
unsubscribe = () => something.off(handler); // might return something
```

As a way of signifying that whatever value is returned should be ignored. WDYT?

packages/core/src/utils.ts
```diff
@@ -309,17 +309,17 @@ export function toArray<T>(value: T[] | T | undefined): T[] {
...
 ): any {
   if (isFunction(mapper)) {
-    return (mapper as Mapper<TContext, TEvent>)(context, _event.data);
+    return mapper(context, _event.data);
   }
 
   const result = {} as any;
 
-  for (const key of keys(mapper)) {
+  for (const key of Object.keys(mapper)) {
```
davidkpiano, line 322: Don't like the helper fn? 😛

davidkpiano:
Re: changeset, this probably fits into the "minor change" category since a "public" method is being changed from `.data` to `.doneData` (which is more accurate). We already have a pending minor change so this can just go into that.

packages/core/src/utils.ts
Andarist: the helper infers `never[]` here and then `key` cant be later used to access the mapper, I've just chosen to use the builtin instead of casting, can change it either way

Andarist:
> Re: changeset, this probably fits into the "minor change" category since a "public" method is being changed from .data to .doneData

Yeah, I was planning to make it a minor. I assume you are OK with changes here so I'm going to add changesets later and ping you for a final review.

packages/core/src/types.ts
Andarist: Gonna experiment with different things in a moment and report back - my guess is that `unknown` is best for such cases.

davidkpiano:
Yep, all good!

Andarist:
@davidkpiano ok, i've added changesets - please take a look.

PR #1267, opened 2020-07-01
[@xstate/fsm] Allow to rehydrate interpreted state machine 

nlopin:
## Intent

The main `xstate` library provides a lot of functionality, but also quite big in size. For most of the frontend projects, the minimalistic implementation is more than enough. But it is impossible to restore the state of the FSM after page reload. This PR targets to solve this issue.

It also makes it possible to improve `@xstate/react` and rehydrate the FSM state when using `@xstate/fsm`

## What

The solution is inspired by the main library:
Method `start` of the interpreted machine set machine in a state, provided by optional `initialState` parameter.

### DIFF
--- a/.changeset/light-news-add.md
+++ b/.changeset/light-news-add.md
@@ -2,4 +2,12 @@
 '@xstate/fsm': minor
 ---
 
-You can now rehydrate the state of a service by calling `start` with `initialState` argument.
+You can now rehydrate the state of a service by calling `start` with `initialState` argument:
+
+```js
+interpret(someMachine).start('active');
+
+interpret(anotherMachine).start({
+  value: 'active',
+  context: { count: 42 }
+});
--- a/.changeset/light-news-add.md
+++ b/.changeset/light-news-add.md
@@ -1,0 +1,5 @@
+---
+'@xstate/fsm': minor
+---
+
+The service can rehydrate the state on start
--- a/.changeset/light-news-add.md
+++ b/.changeset/light-news-add.md
@@ -2,4 +2,4 @@
 '@xstate/fsm': minor
 ---
 
-The service can rehydrate the state on start
+You can now rehydrate the state of a service by calling `start` with `initialState` argument.
--- a/packages/xstate-fsm/src/index.ts
+++ b/packages/xstate-fsm/src/index.ts
@@ -95,17 +95,24 @@ export function createMachine<
     });
   }
 
+  const getStateByValue = (
+    state: TState['value']
+  ): StateMachine.State<TContext, TEvent, TState> => {
+    return {
+      value: state,
+      actions: toArray(fsmConfig.states[state].entry).map((action) =>
+        toActionObject(action, options.actions)
+      ),
+      context: fsmConfig.context!,
+      matches: createMatcher(state)
+    };
+  };
+
   const machine = {
     config: fsmConfig,
     _options: options,
-    initialState: {
-      value: fsmConfig.initial,
-      actions: toArray(
-        fsmConfig.states[fsmConfig.initial].entry
-      ).map((action) => toActionObject(action, options.actions)),
-      context: fsmConfig.context!,
-      matches: createMatcher(fsmConfig.initial)
-    },
+    initialState: getStateByValue(fsmConfig.initial),
+    getStateByValue,
     transition: (
       state: string | StateMachine.State<TContext, TEvent, TState>,
       event: string | (Record<string, any> & { type: string })
@@ -229,7 +236,13 @@ export function interpret<
         unsubscribe: () => listeners.delete(listener)
       };
     },
-    start: () => {
+    start: (initialState: TState['value']) => {
+      if (
+        initialState &&
+        Array.from(Object.keys(machine.config.states)).includes(initialState)
+      ) {
+        state = machine.getStateByValue(initialState);
+      }
       status = InterpreterStatus.Running;
       executeStateActions(state, INIT_EVENT);
       return service;
--- a/packages/xstate-fsm/src/index.ts
+++ b/packages/xstate-fsm/src/index.ts
@@ -237,10 +237,7 @@ export function interpret<
       };
     },
     start: (initialState: TState['value']) => {
-      if (
-        initialState &&
-        Array.from(Object.keys(machine.config.states)).includes(initialState)
-      ) {
+      if (initialState && initialState in machine.config.states) {
         state = machine.getStateByValue(initialState);
       }
       status = InterpreterStatus.Running;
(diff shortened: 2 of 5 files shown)


packages/xstate-fsm/src/index.ts
```diff
@@ -231,7 +236,13 @@ export function interpret<
         unsubscribe: () => listeners.delete(listener)
       };
     },
-    start: () => {
+    start: (initialState: TState['value']) => {
+      if (
+        initialState &&
+        Array.from(Object.keys(machine.config.states)).includes(initialState)
```
davidkpiano, line 242: This will probably work...

```suggestion
        initialState in machine.config.states
```

packages/xstate-fsm/src/index.ts
```diff
@@ -87,27 +87,32 @@ export function createMachine<
...
+        throw new Error(`Nested finite states not supported.
             Please check the configuration for the "${state}" state.`);
-        }
-      });
+      }
+    });
   }
-  
+
+  const getStateByValue = (
```
davidkpiano, line 98: This can probably be moved to a `getStateByValue(machine, state)` function, what do you think? Will reduce output code size.

davidkpiano:
This would be a great addition! Let's make sure it only increases size minimally (see comments).

packages/xstate-fsm/src/index.ts
nlopin: This one is tricky to make. We need properties that `Machine` do not expose:
- `actions`
- `context`

I tried to use currying to workaround the issue, but it increases the size:

```
function createGetStateByValue<
  TContext extends object,
  TEvent extends EventObject,
  TState extends Typestate<TContext>
  >(
  machineConfig: StateMachine.Config<TContext, TEvent, TState>,
  actions: StateMachine.ActionMap<TContext, TEvent> = {},
): (state: TState['value']) => StateMachine.State<TContext, TEvent, TState> {
  return (state: TState['value']) => ({
    value: state,
    actions: toArray(machineConfig.states[state].entry).map((action) =>
      toActionObject(action, actions)
    ),
    context: machineConfig.context!,
    matches: createMatcher(state)
  });
}
```

packages/xstate-fsm/src/index.ts
nlopin: Fixed

.changeset/light-news-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/fsm': minor
+---
+
+The service can rehydrate the state on start
```
Andarist, line 5:
```suggestion
You can now rehydrate the state of a service by calling `start` with `initialState` argument.
```

packages/xstate-fsm/src/index.ts
```diff
@@ -231,7 +236,10 @@ export function interpret<
         unsubscribe: () => listeners.delete(listener)
       };
     },
-    start: () => {
+    start: (initialState: TState['value']) => {
+      if (initialState && initialState in machine.config.states) {
```
Andarist, line 240: Not sure if we need to second part of this condition. We either should not it at all or warn/throw (preferably only in dev) when this condition is met. Swallowing a possible mistake on the developer's side might cause discovering an error longer than needed.

packages/xstate-fsm/src/types.ts
```diff
@@ -96,6 +96,9 @@ export namespace StateMachine {
   > {
     config: StateMachine.Config<TContext, TEvent>;
     initialState: State<TContext, TEvent, TState>;
+    getStateByValue: (
```
Andarist, line 99: does this have to be a public method?

nlopin: I don't see a better approach. If you have an idea how to avoid having a private method, I'm ready to change the implementation.

The current implementation of the Service keeps track of the current state by keeping the full `State` object from the state machine. It works fine if the state machine always starts from the initial state: the state machine can expose initial state directly. 

This approach does not work, If we consider that a service can be started in any state. We need a way to receive or derive the `State` object from the state value (another approach is to ask the developer to provide the full object). It is reasonable to have a method of machine that returns a state object because the state is created from the machine configuration and a list of actions provided.

packages/xstate-fsm/src/index.ts
nlopin: I like the idea to warn! 

There will be a JS error when trying to get `fsmConfig.states[state].entry` if the provided state does not exist in the configuration

packages/xstate-fsm/src/types.ts
Andarist: Note how in the core XState we do not "resolve" entry actions when rehydrating state: https://codesandbox.io/s/competent-chaum-kf9qb?file=/src/index.js

Ideally, we would match its behavior here to avoid confusion. And to do that you would only have to recreate the State using publicly known information: value & context.

nlopin: @Andarist take a look. I rolled back things to the previous state and manually create the state object. 

The entry action of the rehydrated state will bot be triggered

Andarist: It looks good. I have only one concern - it seems to me that rehydrating context is very important for actual state persistence and this is not covered by this PR.

nlopin: I totally agree, this is another part that is not handled now. 

I'd love to implement it in a separate PR. I've checked the main library and there are a bunch of options available. We can go an easy way and allow a user to provide context as a second argument. This should not increase the size of the lib.

Another option is to create `withContext` method of the `StateMachine` that sets the context. ([Reference to the main library](https://xstate.js.org/docs/guides/context.html#initial-context))

What do you think?

Andarist: Wouldnt something like this work: https://github.com/nlopin/xstate/pull/1 ?

nlopin: Merged. I'll add the test today and we are good to go 👍

nlopin: I added the test and grouped all `start`-related tests

.changeset/light-news-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/fsm': minor
+---
+
+You can now rehydrate the state of a service by calling `start` with `initialState` argument.
```
davidkpiano, line 5:
```suggestion
You can now rehydrate the state of a service by calling `start` with `initialState` argument:

```js
interpret(someMachine).start('active');

interpret(anotherMachine).start({
  value: 'active',
  context: { count: 42 }
});
```

nlopin:
@davidkpiano when do you expect to publish the package?

davidkpiano:
> @davidkpiano when do you expect to publish the package?

This should be released already. Something went wrong with the bot that publishes releases, so I'll manually release it on GitHub, but please check the latest version on npm.

Andarist:
Yeah, I've rechecked and this got already published to npm 10 days ago 👍

PR #1273, opened 2020-07-03
Export TestPath

rpradal:
TestPath interface is not currently exported. Having this type exported is useful when writing tests.

### DIFF
--- a/.changeset/real-buses-film.md
+++ b/.changeset/real-buses-film.md
@@ -1,0 +1,5 @@
+---
+'@xstate/test': patch
+---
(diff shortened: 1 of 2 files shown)


.changeset/real-buses-film.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/test': patch
+---
+
+TestPath interface is now exported
```
Andarist, line 5:
```suggestion
`TestPath` interface got exported to be publicly available.
```

PR #1282, opened 2020-07-03
Fix warnings caused by spawn being used in initial assigns

Andarist:


davidkpiano:
Tests fail with this :(

Andarist:
@davidkpiano ye, this is the case i was describing yesterday on Slack. I've managed to dig deeper now and I've found the problem:

MUTATION 😭

`.children` are being mutated when the root service gets initialized and because the `.initialState` is cached it's the very same object so the initial `setState` here:
https://github.com/davidkpiano/xstate/blob/ca8841a5da6560f2956b0dfa08eb05252ad1eca5/packages/xstate-react/src/useMachine.ts#L189
doesn't trigger rerender at all - it just bails out.

Andarist:
This is being covered by an alternative logic based on this condition: https://github.com/davidkpiano/xstate/pull/1290/files#diff-ae24dd65ceb8410ea7eb29a00c9e7cf0R1297

PR #1284, opened 2020-07-04
[V5] Add deferred/lazy behavior

davidkpiano:


packages/core/src/Actor.ts
```diff
@@ -129,37 +133,3 @@ export class Actor<TEvent extends EventObject, TEmitted> {
     this.processingStatus = ProcessingStatus.NotProcessing;
   }
 }
-
-export class ObservableActorRef<TEvent extends EventObject, TEmitted>
```
davidkpiano, line 133: This was moved to `./ObservableActorRef.ts`

packages/core/src/behavior.ts
```diff
@@ -54,8 +57,8 @@ export interface Behavior<TReceived extends EventObject, TEmitted> {
   subscribe?: (observer: Observer<TEmitted>) => Subscription | undefined;
 }
 
-export function createCallbackBehavior<TEvent extends EventObject>(
-  callback: InvokeCallback,
+export function createDeferredBehavior<TEvent extends EventObject>(
```
davidkpiano, line 60: In the future, this could use heuristics for lazily invoked promises/observables/machines/actors. (to be discussed)

packages/core/src/interpreter.ts
```diff
@@ -707,60 +707,29 @@ export class Interpreter<
...
         try {
-          let actorRef: ActorRef<any>;
+          const actorRef = createInvocation(
+            state,
+            action as InvokeActionObject,
+            this.machine,
+            this.ref
+          );
 
-          if (isActorRef(src)) {
```
davidkpiano, line 727: This logic has been moved to `createInvocation(...)` (bikesheddable)

packages/core/src/invoke.ts
```diff
@@ -80,3 +94,53 @@ export function invokeObservable<T extends EventObject = AnyEventObject>(
     return createObservableBehavior(resolvedSource, parent);
   };
 }
+
+export function createInvocation<TContext, TEvent extends EventObject>(
```
davidkpiano, line 98: This is not new logic; it was moved from `interpreter.ts`, everything is still the same.

packages/core/src/stateUtils.ts
```diff
@@ -1547,6 +1548,21 @@ export function resolveMicroTransition<
     delete history.history;
   }
 
+  nextState.actions.forEach((action) => {
```
davidkpiano, line 1551: This introduces the previously existing behavior in V4 where invoked actors were present in `state.children` immediately (while possibly being `deferred`)

packages/core/test/actor.test.ts
```diff
@@ -173,39 +173,39 @@ describe('spawning machines', () => {
...
-      promiseRef: undefined
-    },
-    states: {
-      idle: {
-        entry: assign({
-          promiseRef: (_, __, { spawn }) => {
-            const promise = new Promise((res) => {
-              res('response');
-            });
+  it('should be able to spawn a promise', (done) => {
```
davidkpiano, line 176: Nothing changed here; just moving the machine to inside the test.

packages/xstate-react/src/useActor.ts
```diff
@@ -1,8 +1,29 @@
...
+import {
+  EventObject,
+  Interpreter,
+  AnyEventObject,
+  Subscribable,
+  InvokeDefinition
+} from 'xstate';
 import { fromService } from './useService';
 
+export interface Actor<
```
davidkpiano, line 12: This is no longer present in V5, so I moved it directly to here.

packages/core/src/invoke.ts
```diff
@@ -4,19 +4,26 @@ import {
...
 import { MachineNode } from './MachineNode';
 import {
   createMachineBehavior,
-  createCallbackBehavior,
+  createDeferredBehavior,
   Behavior,
   createObservableBehavior,
   createPromiseBehavior
 } from './behavior';
+import { InvokeActionObject } from '../dist/xstate.cjs';
```
Andarist, line 22: this should be imported from `./types` or something `/dist/` should never appear in our internal imports

packages/core/src/behavior.ts
```diff
@@ -276,19 +286,23 @@ export function createMachineBehavior<TContext, TEvent extends EventObject>(
...
     receive: (_, event) => {
-      service.send(event);
+      service?.send(event);
       return behavior;
     },
     subscribe: (observer) => {
       return service?.subscribe(observer);
     },
-    current: machine.initialState // TODO: this should get from machine.getInitialState(ref)
+    get current() {
```
Andarist, line 301: shouldnt this be read from the created `service`? right now this won't ever be updated as machines are stateless

packages/core/src/behavior.ts
```diff
@@ -135,7 +140,9 @@ export function createPromiseBehavior<T, TEvent extends EventObject>(
     receiveSignal: (actorContext: ActorContext, signal: LifecycleSignal) => {
       switch (signal) {
         case startSignal:
-          const resolvedPromise = Promise.resolve(promise);
+          const resolvedPromise = Promise.resolve(
```
Andarist, line 143: should we allow for anything other than machines to be "eager"? with machines we can accept plain machines - rather than enforcing machine creators at all times as we know that machines are stateless and thus it's safe

related question - should `createXBehavior` create the very same "template" at all times for a given set of arguments? if yes then we would have to require entity creators (lazy) as arguments to achieve this, if no then I'm not sure how useful accepting lazy creators is here - this is not the main user-facing API and those lazy creators get resolved in `invokeX` already

packages/core/src/invoke.ts
Andarist: this name is slightly confusing to me as this returns an actor ref and "invocation" sounds like a new/different concept

packages/core/src/invoke.ts
```diff
@@ -29,13 +36,19 @@ export function invokeMachine<
...
-    if (data) {
-      resolvedMachine = resolvedMachine.withContext(
-        mapContext(data, ctx, _event)
-      ) as TMachine;
-    }
-    return createMachineBehavior(resolvedMachine, parent, options);
+    const resolvedContext = data ? mapContext(data, ctx, _event) : undefined;
+    const machineOrDeferredMachine = isFunction(machine)
+      ? () => {
+          const resolvedMachine = machine(ctx, event);
```
Andarist, line 42: promise creators are called with a third argument as well, i think we should strive for unifying those signatures

packages/core/src/invoke.ts
davidkpiano: I'm not sure why VS Code sometimes auto-imports from `/dist/` - this only occurs in `next`.

packages/core/src/behavior.ts
davidkpiano: Yeah, that's a change I'm planning to make in a separate PR - we should enforce that all spawned actors be lazy, e.g.:

```diff
-spawn(promise);
+spawn(() => promise);
```

packages/core/src/invoke.ts
davidkpiano: Renamed to... something slightly more descriptive

packages/core/src/behavior.ts
davidkpiano: This is badly named; it should only represent the `initial` state

packages/core/src/behavior.ts
Andarist: 👍

packages/core/src/behavior.ts
```diff
@@ -229,26 +236,29 @@ export function createObservableBehavior<
     },
     receive: () => behavior,
     subscribe: (observer) => {
-      return observable.subscribe(observer);
+      return observable?.subscribe(observer);
```
Andarist, line 239: shouldnt `.subscribe` be always available on observables? 😄

davidkpiano: Don't worry, it is! This signifies that the _observable_ itself might not exist before `subscribe(...)` is called.

Not sure what to do here, it's an unlikely race condition that (in here) is handled by just not subscribing.

PR #1285, opened 2020-07-05
Fixed an issue with initial state of invoked machines being read without custom data

Andarist:
fixes https://github.com/davidkpiano/xstate/issues/1277

This has a caveat though - in a normal situation a service creator is called here:
https://github.com/davidkpiano/xstate/blob/8226700aab94031f66133c6518430786aab4780c/packages/core/src/interpreter.ts#L871-L877
and it is called with different values - `_event` (nor `context`) is not guaranteed to be the same between those 2 points in time. I don't feel this is fixable easily in v4 though.

### DIFF
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -53,12 +53,13 @@ export function createInvocableActor<TC, TE extends EventObject>(
   tempActor.deferred = true;
 
   if (isMachine(serviceCreator)) {
-    tempActor.state = serviceCreator.getInitialState(
-      serviceCreator.initialStateValue!,
-      invokeDefinition.data
-        ? mapContext(invokeDefinition.data, context, _event)
-        : undefined
-    );
+    const resolvedData = invokeDefinition.data
+      ? mapContext(invokeDefinition.data, context, _event)
+      : undefined;
+    tempActor.state = (resolvedData
+      ? serviceCreator.withContext(resolvedData)
+      : serviceCreator
+    ).initialState;
   }
 
   tempActor.meta = invokeDefinition;
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -2,10 +2,11 @@ import {
   EventObject,
   Subscribable,
   InvokeDefinition,
-  AnyEventObject
+  AnyEventObject,
+  SCXML
 } from './types';
 import { StateMachine } from '.';
-import { isMachine } from './utils';
+import { isMachine, mapContext } from './utils';
 
 export interface Actor<
   TContext = any,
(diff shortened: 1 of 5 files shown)


packages/core/test/transient.test.ts
```diff
@@ -629,4 +629,45 @@ describe('transient states (eventless transitions)', () => {
 
     service.send('ADD');
   });
+
+  it("shouldn't crash when invoking a machine with initial transient transition depending on custom data", () => {
```
Andarist, line 633: This is not the best test as @davidkpiano's proposed workaround [here](https://github.com/davidkpiano/xstate/issues/1277#issuecomment-653673951) is valid (to some extent). One should not rely on undefined `context` - if `TContext` parameter is configured then `.context` should satisfy that constraint at all times and thus expressions (like guards) should not fail due to some properties being missed.

However, I don't know what other observable behavior to test at XState layer. The other added test (`@xstate/react`) is much better: https://github.com/davidkpiano/xstate/pull/1285/files#diff-326fee0dc0480200270263670a5d51bfR608

packages/core/src/Actor.ts
```diff
@@ -43,14 +44,21 @@ export function createNullActor(id: string): Actor {
...
+  context: TC,
+  _event: SCXML.Event<TE>
 ): Actor {
   const tempActor = createNullActor(invokeDefinition.id);
   const serviceCreator = machine.options.services?.[invokeDefinition.src];
   tempActor.deferred = true;
 
   if (isMachine(serviceCreator)) {
-    tempActor.state = serviceCreator.initialState;
+    tempActor.state = serviceCreator.getInitialState(
```
davidkpiano, line 56: We can probably refactor this later, but:

```js
serviceCreator.withContext(mapContext(...)).initialState
```

Might be more accurate, no?

packages/core/test/transient.test.ts
```diff
@@ -629,4 +629,45 @@ describe('transient states (eventless transitions)', () => {
...
+    const timerMachine = Machine({
+      initial: 'intitial',
```
davidkpiano, line 635:
```suggestion
      initial: 'initial',
```

packages/core/test/transient.test.ts
```diff
@@ -629,4 +629,45 @@ describe('transient states (eventless transitions)', () => {
...
+      states: {
+        intitial: {
```
davidkpiano, line 637:
```suggestion
        initial: {
```

packages/core/test/transient.test.ts
Andarist: @davidkpiano do you think this test here is worth keeping?

packages/core/src/Actor.ts
Andarist: good suggestion, I've applied the change

packages/core/test/transient.test.ts
davidkpiano: I think so, yes.

PR #1290, opened 2020-07-08
Ensure that prematurely spawned actors are also deferred. Fixes #1281

### LINKED ISSUE #1281: Regression with xstate/react@1.0.0-rc.5
**Description**

userService throws an exception complaining about the fact that the service argument is undefined

`Uncaught TypeError: Cannot read property 'initialState' of undefined`

**Expected Result**
Since the service is defined. It should not throw error

This is a regression specific to xstate/react@1.0.0-rc.5.
With xstate/react@1.0.0-rc.4, everything was ok.

**Reproduction**

https://codesandbox.io/s/unruffled-wescoff-6yq1d?file=/src/index.tsx

**Additional context**
tested with 4.11.0 ans 4.10

davidkpiano:
This fixes #1281 by essentially adding `.deferred` to spawned actors, and making them behave like prematurely invoked actors would behave.


packages/xstate-react/test/useService.test.tsx
```diff
@@ -182,101 +182,4 @@ describe('useService hook', () => {
     fireEvent.click(incButton);
     expect(countEl.textContent).toBe('1');
   });
-
-  it('initial invoked actor should be immediately available', (done) => {
```
davidkpiano, line 186: These tests were moved to `useActor.test.tsx`

packages/xstate-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@xstate/react",
-  "version": "1.0.0-rc.3",
+  "version": "1.0.0-rc.6",
```
Andarist, line 3: this hasnt been published yet - I think it would be best to only bump versions when actually releasing them

sideways q: wasn't `rc.5` merged into master? i see a 3 -> 6 bump here, which looks weird and maybe we've lost some git history along the way?

davidkpiano: I think it was bumped out-of-band. Sorry about that.

davidkpiano:
> This should land together with #1282 (once failing tests there get fixed) or the warning in question should get removed entirely.

Instead of #1282, I fixed the heuristics for "lazy entities" so that we won't get those warnings.

packages/core/src/interpreter.ts
```diff
@@ -1303,19 +1293,20 @@ export function spawn(
   const resolvedOptions = resolveSpawnOptions(nameOrOptions);
 
   return withServiceScope(undefined, (service) => {
+    const isLazyEntity = isMachine(entity) || isFunction(entity);
```
Andarist, line 1296: this could be moved to the dev-only branch below so those calls would get removed in prod bundles

PR #1317, opened 2020-07-18
Unified declarations with TTypestate

Andarist:
Fixes https://github.com/davidkpiano/xstate/issues/1301

I've unified this across the codebase. I have one problem though:

The search&replace hit `@xstate/fsm` as well and while reviewing applied changes I've noticed that its `.matches` implementation is different than what is in the core, so I've decided to just unify both (it's not yet part of this PR, but can be seen [here](https://github.com/davidkpiano/xstate/commit/98bbf9b3ceb35bf04aacba97b9ee1244acf66dfa)). I have wanted to confirm that this has indeed fixed something so I could write a proper changeset so I've looked into [the latest `.matches`-related PR](https://github.com/davidkpiano/xstate/pull/1189/files) to check out tests that have been added there and... given the same test (but using `@xstate/fsm`) I can't get it working even after the very similar changes being applied to `@xstate/fsm`.

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -142,7 +142,7 @@ class StateNode<
   TContext = any,
   TStateSchema extends StateSchema = any,
   TEvent extends EventObject = EventObject,
-  TTypestate extends Typestate<TContext> = any
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 > {
   /**
    * The relative key of the state node, which represents its location in the overall state value.
@@ -614,8 +614,8 @@ class StateNode<
    * @param state The state value or State instance
    */
   public getStateNodes(
-    state: StateValue | State<TContext, TEvent>
-  ): Array<StateNode<TContext, any, TEvent>> {
+    state: StateValue | State<TContext, TEvent, any, TTypestate>
+  ): Array<StateNode<TContext, any, TEvent, TTypestate>> {
     if (!state) {
       return [];
     }
@@ -636,7 +636,8 @@ class StateNode<
     const subStateNodes: Array<StateNode<
       TContext,
       any,
-      TEvent
+      TEvent,
+      TTypestate
     >> = subStateKeys.map((subStateKey) => this.getStateNode(subStateKey));
 
     return subStateNodes.concat(
@@ -646,7 +647,7 @@ class StateNode<
         );
 
         return allSubStateNodes.concat(subStateNode);
-      }, [] as Array<StateNode<TContext, any, TEvent>>)
+      }, [] as Array<StateNode<TContext, any, TEvent, TTypestate>>)
     );
   }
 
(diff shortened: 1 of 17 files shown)


packages/core/src/StateNode.ts
```diff
@@ -614,8 +614,8 @@ class StateNode<
    * @param state The state value or State instance
    */
   public getStateNodes(
-    state: StateValue | State<TContext, TEvent>
-  ): Array<StateNode<TContext, any, TEvent>> {
+    state: StateValue | State<TContext, TEvent, any, TTypestate>
+  ): Array<StateNode<TContext, any, TEvent, TTypestate>> {
```
Andarist, line 618: actually maybe this shouldn't be parametrized with `TTypestate` as the "correct" value is only relevant for State and such as it's a top-level value and this one returns nested StateNodes as well, thoughts?

davidkpiano: Yeah, we don't need it here.

packages/core/src/actions.ts
```diff
@@ -159,7 +159,7 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
  */
 export function raise<TContext, TEvent extends EventObject>(
   event: Event<TEvent>
-): RaiseAction<TEvent> | SendAction<TContext, TEvent, TEvent> {
+): RaiseAction<TEvent> | SendAction<TContext, AnyEventObject, TEvent> {
```
davidkpiano, line 162: Confused about this - the `raise(...)` action creator should only ever be able to raise events that the machine accepts; that is, only `TEvent` events, not `AnyEventObject`.

packages/core/src/interpreter.ts
```diff
@@ -1320,7 +1320,7 @@ export function interpret<
   TContext = DefaultContext,
   TStateSchema extends StateSchema = any,
   TEvent extends EventObject = EventObject,
-  TTypestate extends Typestate<TContext> = any
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
```
davidkpiano, line 1323: Future: we should alias `{ value: any; context: TContext }` as `AnyTypestate<TContext>`.

Andarist: Yeah, was thinking about that but I have also noticed that instinctively I would declare such a type as:
```ts
type AnyTypestate<TContext> = { value: StateValue; context: TContext }
```
but from what I have checked when working on this PR this doesn't work (not sure why).

packages/core/src/actions.ts
Andarist: Right, the second type param is about the event received by send expressions - `raise` doesn't support expressions right now though, so we can just parameterized this with `AnyEventObject` so it can be passed at any position in a machine's config. The 3rd type param in `send` is `TSentEvent` and this contraints the `raise` to only raise events accepted by the machine.

packages/core/src/StateNode.ts
Andarist: k, gonna adjust later today

packages/core/src/actions.ts
davidkpiano: Ah, gotcha. Sounds good to me.

PR #1318, opened 2020-07-18
[@xstate/immer] Add createPatch()

davidkpiano:
This PR:

- Bumps the peer dependency of `immer` to 7.0.5 (we can roll this back if needed)
- Adds support for `createPatch()`, which creates an event function that can represent "mutations" to parts of the `context` as patches, which can be handled in transitions:

```js
import { createPatch } from '@xstate/immer';

// creates a "patch" event function that modifies user
const userPatch = createPatch('user');

const userMachine = createMachine({
  context: {
    user: {
      name: '',
      age: undefined,
      // ...
    }
  },
  initial: 'editing',
  states: {
    editing: {
      on: {
        [userPatch.type]: { actions: userPatch.action }
        // or, shorthand:
        // ...userPatch.transition
      }
    }
  }
});

const userService = interpret(userMachine).start();

// ...

// update the user name and age, assuming you have reference
// to the current `user` from context
userService.send(userPatch(user, u => {
  u.name = 'David';
  u.age = 30;
}));
```

### DIFF
--- a/packages/xstate-immer/package.json
+++ b/packages/xstate-immer/package.json
@@ -37,7 +37,7 @@
   },
   "dependencies": {},
   "peerDependencies": {
-    "immer": "^6.0.3",
+    "immer": "^7.0.5",
     "xstate": "^4.9.1"
   },
   "devDependencies": {
--- a/packages/xstate-immer/src/index.ts
+++ b/packages/xstate-immer/src/index.ts
@@ -5,7 +5,9 @@ import {
   assign as xstateAssign,
   AssignMeta
 } from 'xstate';
-import { produce, Draft } from 'immer';
+import { produce, Draft, Patch, enablePatches, applyPatches } from 'immer';
+
+enablePatches();
 
 export type ImmerAssigner<TContext, TEvent extends EventObject> = (
   context: Draft<TContext>,
@@ -61,3 +63,52 @@ export function createUpdater<TContext, TEvent extends ImmerUpdateEvent>(
     type
   };
 }
+
+const IMMER_PATCH_TYPE = '@immer.patch' as const;
+
+export interface ImmerPatchEvent<K> {
+  type: typeof IMMER_PATCH_TYPE;
+  key: K;
+  patches: Patch[];
+}
+
+export const createPatch = <T, K extends keyof T>(
+  key: K
+): {
+  (currentValue: T[K], fn: (draft: Draft<T[K]>) => void): ImmerPatchEvent<K>;
+  type: ImmerPatchEvent<K>['type'];
+  action: AssignAction<T, ImmerPatchEvent<K>>;
+  transition: Record<
+    ImmerPatchEvent<K>['type'],
+    { actions: Array<AssignAction<T, ImmerPatchEvent<K>>> }
+  >;
+} => {
+  const applyPatch = (
+    currentValue: T[K],
+    fn: (draft: Draft<T[K]>) => void
+  ): ImmerPatchEvent<K> => {
+    const patches: Patch[] = [];
+
+    produce(currentValue, fn, (resultPatches) => {
+      patches.push(...resultPatches);
+    });
+
+    return {
+      type: IMMER_PATCH_TYPE,
+      key,
+      patches
+    };
+  };
+
+  const action = xstateAssign({
+    [key as K]: (ctx: T, e: ImmerPatchEvent<K>) => {
+      return applyPatches(ctx[key], e.patches);
+    }
+  }) as AssignAction<T, ImmerPatchEvent<K>>;
+
+  applyPatch.type = IMMER_PATCH_TYPE;
+  applyPatch.action = action;
+  applyPatch.transition = { [IMMER_PATCH_TYPE]: { actions: [action] } };
+
+  return applyPatch;
+};
--- a/packages/xstate-immer/test/immer.test.ts
+++ b/packages/xstate-immer/test/immer.test.ts
@@ -1,5 +1,11 @@
 import { createMachine, interpret } from 'xstate';
-import { assign, createUpdater, ImmerUpdateEvent } from '../src';
+import {
+  assign,
+  createPatch,
+  createUpdater,
+  ImmerUpdateEvent,
+  ImmerPatchEvent
+} from '../src';
 
 describe('@xstate/immer', () => {
   it('should update the context without modifying previous contexts', () => {
@@ -184,12 +190,10 @@ describe('@xstate/immer', () => {
           }
         },
         submitting: {
-          on: {
-            '': {
-              target: 'success',
-              cond: (ctx) => {
-                return ctx.name === 'David' && ctx.age === 0;
-              }
+          always: {
+            target: 'success',
+            cond: (ctx) => {
+              return ctx.name === 'X' && ctx.age === 0;
             }
           }
         },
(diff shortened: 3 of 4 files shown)


packages/xstate-immer/package.json
```diff
@@ -37,7 +37,7 @@
   },
   "dependencies": {},
   "peerDependencies": {
-    "immer": "^6.0.3",
+    "immer": "^7.0.5",
```
Andarist, line 40: You should change the dev dep on immer as well

packages/xstate-immer/src/index.ts
```diff
@@ -5,7 +5,9 @@ import {
   assign as xstateAssign,
   AssignMeta
 } from 'xstate';
-import { produce, Draft } from 'immer';
+import { produce, Draft, Patch, enablePatches, applyPatches } from 'immer';
+
+enablePatches();
```
Andarist, line 10: I'm not super comfortable with a library calling something like that - it is a side effect after all. I understand why this is needed - but maybe this call should be required from a user? Is there any guidance from Immer on this for libraries integrating with Immer? cc @mweststrate

packages/xstate-immer/src/index.ts
```diff
@@ -61,3 +63,52 @@ export function createUpdater<TContext, TEvent extends ImmerUpdateEvent>(
     type
   };
 }
+
+const IMMER_PATCH_TYPE = '@immer.patch' as const;
```
Andarist, line 67:
```suggestion
const IMMER_PATCH_TYPE = '@immer.patch';
```

`as const` is redundant here, as you are assigning a primitive value within a `const` declaration

packages/xstate-immer/test/immer.test.ts
```diff
@@ -205,9 +209,63 @@ describe('@xstate/immer', () => {
       })
       .start();
 
-    service.send(nameUpdater.update('David'));
+    service.send(nameUpdater.update('X'));
```
Andarist, line 212: I have liked David much more than this X guy.

packages/xstate-immer/src/index.ts
```diff
@@ -61,3 +63,52 @@ export function createUpdater<TContext, TEvent extends ImmerUpdateEvent>(
...
+): {
+  (currentValue: T[K], fn: (draft: Draft<T[K]>) => void): ImmerPatchEvent<K>;
+  type: ImmerPatchEvent<K>['type'];
+  action: AssignAction<T, ImmerPatchEvent<K>>;
+  transition: Record<
+    ImmerPatchEvent<K>['type'],
+    { actions: Array<AssignAction<T, ImmerPatchEvent<K>>> }
+  >;
+} => {
+  const applyPatch = (
```
Andarist, line 86: In https://github.com/davidkpiano/xstate/pull/1129 you have implemented updater as a function with extra properties but after a discussion, it has been changed to be a plain object with properties. So I would say that the same should be done for this new `createPatch`, even if only for the API consistency.

packages/xstate-immer/src/index.ts
```diff
@@ -61,3 +63,52 @@ export function createUpdater<TContext, TEvent extends ImmerUpdateEvent>(
...
+    currentValue: T[K],
+    fn: (draft: Draft<T[K]>) => void
```
Andarist, line 88: this should also allow `typeof nothing` to be returned

packages/xstate-immer/src/index.ts
```diff
@@ -61,3 +63,52 @@ export function createUpdater<TContext, TEvent extends ImmerUpdateEvent>(
...
+
+  const action = xstateAssign({
+    [key as K]: (ctx: T, e: ImmerPatchEvent<K>) => {
+      return applyPatches(ctx[key], e.patches);
+    }
+  }) as AssignAction<T, ImmerPatchEvent<K>>;
+
+  applyPatch.type = IMMER_PATCH_TYPE;
+  applyPatch.action = action;
+  applyPatch.transition = { [IMMER_PATCH_TYPE]: { actions: [action] } };
```
Andarist, line 111: This is IMHO problematic for a couple of reasons:
1. returning `type` from here makes it look like this is a type exclusive for this "patcher", but it is not - it's a "global" type
2. and because it's a global type the composition suffers right now because it's not obvious (if not impossible) how to create 2 "patchers" and apply them at the same state node.

The whole thing is very similar to `createUpdater` and I believe that the API should be shared as much as possible between both of them (I'm not super familiar with Immer, so not sure how much it is possible to share the API, but at least they should resemble each other).

A global type could be used (but not returned from this factory) if only we could produce patches based on the "root" context - then all could just be applied uniformly based on a single event type. This is just a random thought - not sure how feasible this even is.

packages/xstate-immer/src/index.ts
```diff
@@ -61,3 +63,52 @@ export function createUpdater<TContext, TEvent extends ImmerUpdateEvent>(
...
+export interface ImmerPatchEvent<K> {
+  type: typeof IMMER_PATCH_TYPE;
+  key: K;
+  patches: Patch[];
+}
+
+export const createPatch = <T, K extends keyof T>(
+  key: K
+): {
+  (currentValue: T[K], fn: (draft: Draft<T[K]>) => void): ImmerPatchEvent<K>;
```
Andarist, line 78: I believe I've asked about this in the past, but I forgot so would love a reminder - why don't we get `currentValue` automatically in a callback bur rather require it to be passed in explicitly?

packages/xstate-immer/src/index.ts
davidkpiano: That's a fair assumption, we can probably construct different types per "patcher"

packages/xstate-immer/src/index.ts
davidkpiano: If we just did this:

```js
createPatch(userDraft => {
  userDraft.name = 'whatever';
  // ... etc.
});
```

I don't think `immer` would work, since it needs that _initial object_ to create into a proxy, and we don't provide that in `createPatch(...)` either (although we could...)

Unless I'm wrong, and maybe `immer` can create patches by performing a recipe on an "invisible" object... 🤔

Andarist: I would say that Immer is capable to mostly do its magic based on `{}` (havent tested it though). The problems with such approach could be that:
- produced patches could be different (maybe?), and not sure if they could be applied as is on the real object, although i would expect it to be possible
- if we supply a dummy object there then it wouldnt be possivle to read the real one there, so it wouldnt be possible to use its values to compute the new one. Or rather - i suppose it wouldnt be possible as i dont think Immer records read operations that could be "replayed". I might be wrong though.

Questions is if it would be possible to supply the real object lazily to this draft function so it wouldnt have to be done by the user? I suppose this whole thing (patching) is mainly relevant for distinct machines, right? Or just in more async env? It doesnt have much benefits to it if we deal with a single machine to which we have access as one could just use regular updater just fine in such scenario. Unless one would like to store patches for some kind of draft and commit later use cases? Anyway - in async scenarios this requires sync~ access to the data on which it can operate, right? So to really benefit from this in such scenarios one would have to implement data replication logic, do I understand this correctly?

Another question is - do you think its worth supporting lenses-like approach here? Like - patches produced from a different location in the context of a particular machine but later applied at different location of the destination machine?

davidkpiano: > so it wouldnt have to be done by the user?

I don't think this is a worthwhile goal. Think about how a user would want to update a value normally:

```js
//       vvvvv previous count!
setCount(count + 1);
```

Instead of trying to figure out the magic to make `patchUser(user => { ... })` happen, I think `patchUser(user, u => { ... })` is more explicit, unit-testable, and less magical, and we should stick with that.

davidkpiano: > Another question is - do you think its worth supporting lenses-like approach here? Like - patches produced from a different location in the context of a particular machine but later applied at different location of the destination machine?

What would this look like?

Andarist: Not sure - this was just a general wondering as we've discussed lense-like immer-related API in the past. 

One idea that comes to mind:
```ts
const patchUser = createPatch('user');

service.send(
  patchUser(context.user, (u) => {
    u.name = 'X';
    u.age = 42;
  })
);

{
  on: {
    [patchUser.type]: {
      actions: patchUser.assign((ctx, ev, { patch }) => {
        return {
          differentUser: patch(ctx.differentUser)
        }
      })
    }
  }
}
```

It isn't exactly that much lense-like but it decouples the source and destination locations of the patches.

packages/xstate-immer/src/index.ts
mweststrate: If your library is using patches, this method should indeed be called. It prevents the patch functionality from being tree-shaken out of the final bundle, which is what you want :) (you could make that conditional in your library as well if patches are not always used). It saves < 1kb gzipped, so it might not be worth the additional complexity, but was introduced to make sure that immer can be used as minimally as possible. 

Enabling this doesn't semantically affect any other consumers of Immer, and calling it multiple times is safe. So doing this here is safe and probably the correct thing to do.

PR #1320, opened 2020-07-19
Add support for invoke source object

davidkpiano:
This PR adds support for invoke source objects. This is not a breaking change; `src: 'someString'` will continue to work.

```js
invoke: {
  src: {
    type: 'search',
    endpoint: 'google.com'
  },
  // ...
}

// ...

services: {
  search: (ctx, e, { src }) => {
    console.log(src);
    // => { type: 'search', endpoint: 'google.com' }
  }
}

// ...
```

TODO: 
- [x] Documentation
- [x] Changeset

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -22,7 +22,8 @@ import {
   SingleOrArray,
   Guard,
   GuardPredicate,
-  GuardMeta
+  GuardMeta,
+  InvokeSourceDefinition
 } from './types';
 import {
   STATE_DELIMITER,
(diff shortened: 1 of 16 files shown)


davidkpiano:
@Andarist LGTY?

packages/core/src/utils.ts
```diff
@@ -678,3 +679,13 @@ export function evaluateGuard<TContext, TEvent extends EventObject>(
 
   return condFn(context, _event.data, guardMeta);
 }
+
+export function toInvokeSource(
```
Andarist, line 683: a similar one is already available in `packages/core/src/invokeUtils.ts` - any reason for having both?

davidkpiano: That file doesn't exist, link?

PR #1336, opened 2020-07-27
[@xstate/react] Rehydration should initialize machine first. Fixes #1334

### LINKED ISSUE #1334: Rehydrating from persisted state breaks delayed transitions
**Description**
According to this section https://xstate.js.org/docs/guides/states.html#persisting-state in the documentation the state of running machines can be persisted and used to rehydrate the machine to this exact state later. This is a super useful feature in many ways and it worked great so far.
However if I add delayed transitions (https://xstate.js.org/docs/guides/delays.html#delayed-transitions) to my machine I run into problems. On a fresh machine without any previous state the delayed transitions work perfectly fine but as soon as the machine gets initialised with state that was previously persisted, the delayed transitions will not work anymore and the machine gets stuck.

**Expected Result**
Delayed transitions should also work if the machine was started using a previously persisted state.

**Actual Result**
Delayed transitions only work on machines started without an initial state.

**Reproduction**
Code Sandbox which shows the problem: https://codesandbox.io/s/dry-sound-q3fxm

**Additional context**
Can be reproduced with current xstate version 4.11.0

davidkpiano:
This PR fixes #1334 (in a workaround-ish way) by always reading the `machine.initialState` first, which will trigger `machine._init()` to properly initialize all transitions, which includes resolving delayed transitions.

The real fix for this is to do the transition resolution eagerly in the machine's constructor after adding all state nodes, but this would be a breaking change in v4 for state node getters (which will be deprecated in v5), so that will wait til the next major version.

This is not a breaking change.

### DIFF
--- a/packages/core/test/types.test.ts
+++ b/packages/core/test/types.test.ts
@@ -41,6 +41,7 @@ describe('StateSchema', () => {
     context: { elapsed: 0 },
     states: {
       green: {
+        id: 'green',
         meta: { name: 'greenLight' },
         on: {
           TIMER: 'yellow',
@@ -76,7 +77,7 @@ describe('StateSchema', () => {
             }
           },
           stop: {
-            always: { target: 'green' }
+            always: { target: '#green' }
           }
         }
       }
--- a/packages/xstate-react/test/useMachine.test.tsx
+++ b/packages/xstate-react/test/useMachine.test.tsx
@@ -647,4 +647,57 @@ describe('useMachine (strict mode)', () => {
     );
     done();
   });
+
+  // https://github.com/davidkpiano/xstate/issues/1334
+  it('should work with a rehydrated state', (done) => {
+    const testMachine = Machine({
+      id: 'app',
+      initial: 'idle',
+      states: {
+        idle: {
+          on: {
+            START: 'doingStuff'
+          }
+        },
+        doingStuff: {
+          id: 'doingStuff',
+          after: {
+            100: 'idle'
+          }
+        }
+      }
+    });
+
+    const persistedState = JSON.stringify(testMachine.initialState);
+
+    let currentState = testMachine.resolveState(
+      State.create(State.create(JSON.parse(persistedState)))
+    );
+
+    const Test = () => {
+      const [state, send] = useMachine(testMachine, {
+        state: State.create(JSON.parse(persistedState))
+      });
+
+      currentState = state;
+
+      React.useEffect(() => {
+        send('START');
+      }, []);
+
+      return null;
+    };
+
+    render(
+      <React.StrictMode>
+        <Test />
+      </React.StrictMode>
+    );
+
+    setTimeout(() => {
+      if (currentState.matches('idle')) {
+        done();
+      }
+    }, 110);
+  });
 });
--- a/packages/xstate-react/test/useMachine.test.tsx
+++ b/packages/xstate-react/test/useMachine.test.tsx
@@ -670,9 +670,7 @@ describe('useMachine (strict mode)', () => {
 
     const persistedState = JSON.stringify(testMachine.initialState);
 
-    let currentState = testMachine.resolveState(
-      State.create(State.create(JSON.parse(persistedState)))
-    );
+    let currentState;
 
     const Test = () => {
       const [state, send] = useMachine(testMachine, {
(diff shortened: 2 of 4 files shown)


packages/core/test/types.test.ts
```diff
@@ -76,7 +77,7 @@ describe('StateSchema', () => {
             }
           },
           stop: {
-            always: { target: 'green' }
+            always: { target: '#green' }
```
davidkpiano, line 80: This is ~unrelated~ tangentially related to this PR - without eager transition resolution, this will not throw an error (which it should, since `'green'` is not a child state of `'red'`).

davidkpiano:
@Andarist Is this one good to go in?

packages/core/test/types.test.ts
Andarist: so what has been happening here before and how has this been changed by this PR?

davidkpiano: The target is wrong.

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -647,4 +647,57 @@ describe('useMachine (strict mode)', () => {
     );
     done();
   });
+
+  // https://github.com/davidkpiano/xstate/issues/1334
+  it('should work with a rehydrated state', (done) => {
```
Andarist, line 652: there are already 2 tests here which are supposed to handle rehydrated state:
https://github.com/davidkpiano/xstate/blob/9ecefca84bfc3db6df09c41b884ac00ef0f6efb1/packages/xstate-react/test/useMachine.test.tsx#L100
https://github.com/davidkpiano/xstate/blob/9ecefca84bfc3db6df09c41b884ac00ef0f6efb1/packages/xstate-react/test/useMachine.test.tsx#L113

How this one is different? I understand that this relates to delayed transitions somehow, but its not obvious for me why this has been failing before - what info was lacking (and where) that this has failed? I would understand more if this would be about rehydrating `doingStuff` state but this is, from what I understand, rehydrating `idle` and I don't see how this affects a delayed transition in another state.

I know that this fix is OK - it has fixed the issue after all, but would love to understand the core of the issue better.

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -647,4 +647,57 @@ describe('useMachine (strict mode)', () => {
...
+            100: 'idle'
+          }
+        }
+      }
+    });
+
+    const persistedState = JSON.stringify(testMachine.initialState);
+
+    let currentState = testMachine.resolveState(
+      State.create(State.create(JSON.parse(persistedState)))
```
Andarist, line 674: is calling `State.create` twice here intended?

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -647,4 +647,57 @@ describe('useMachine (strict mode)', () => {
...
+          after: {
+            100: 'idle'
+          }
+        }
+      }
+    });
+
+    const persistedState = JSON.stringify(testMachine.initialState);
+
+    let currentState = testMachine.resolveState(
```
Andarist, line 673: is this value used anywhere? from what i see it gets overridden pretty immediately (when rendering `Test`)

packages/core/test/types.test.ts
Andarist: Ah, ok 👌

packages/xstate-react/test/useMachine.test.tsx
davidkpiano: Nothing in the state was lacking - all the available information is available as expected.

The issue was that the machine never had a chance to "initialize" (resolve) its delayed transitions because `machine.initialState` was never read (since we were starting directly from the resolved state).

Ideally, initialization happens in the constructor, but that would currently break "target getters" (which are deprecated in v5).

davidkpiano: **Before:**

- Machine defined, delayed transitions not resolved
- Interpreter starts at restored state
- ❌ Delayed transitions still do not exist; will not be taken

**After:**

- Machine defined, delayed transitions not resolved
- `machine.initialState` read which causes the machine to initialize its transitions (resolve its delayed transitions, etc.)
  - I know this is a code smell; will be fixed in v5
- Interpreter starts at restored state
- ✅ Delayed transitions exist; will be taken

Andarist: Ok - I think I get it now. Thanks for the explanation. 👍

Andarist, line 652: Maybe something like this?

```suggestion
  it('delayed transitions should work when initializing from a rehydrated state', (done) => {
```

PR #1345, opened 2020-07-31
updated respond sample and send warning copy

EdmundsEcho:


docs/guides/actions.md
```diff
@@ -395,7 +395,7 @@ const authServerMachine = Machine({
     waitingForCode: {
       on: {
         CODE: {
-          actions: respond('TOKEN', { delay: 10 })
```
davidkpiano, line 398: The `respond(...)` action creator returns an action object.

PR #1353, opened 2020-08-03
Fix type for src in InvokeConfig

evgenyshenets91:
My project shows the following types for src:
before changes

With my changes it looks like this:
after changes

I run yarn dev and all test suites passed successfully: 
```
Test Suites: 61 passed, 61 total
Tests:       12 skipped, 893 passed, 905 total
Snapshots:   20 passed, 20 total
Time:        29.446s
Ran all test suites in 9 projects.

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -836,8 +836,9 @@ export class Interpreter<
 
         // Invoked services
         if (activity.type === ActionTypes.Invoke) {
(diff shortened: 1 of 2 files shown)


Andarist:
If I understand correctly - this is not a good change because it enforces invoked machines to have the same context, events etc types as the parent machine, whereas there is no such constraint.

evgenyshenets91:
Thanks for your point

PR #1366, opened 2020-08-08
[@xstate/graph] Add toGraph()

davidkpiano:
This PR introduces a helper function that converts machines to a directed graph structure, making it easy to use with graph-related algorithms, such as GraphViz and ELK.

```js
import { toDirectedGraph } from '@xstate/graph';

const machine = createMachine({/* ... */});

const digraph = toDirectedGraph(machine);

// returns an object with this structure:
{
  id: '...',
  stateNode: /* StateNode */,
  children: [
    { id: '...', children: [/* ... */], edges: [/* ... */] },
    { id: '...', /* ... */ },
    // ...
  ],
  edges: [
    { source: /* ... */, target: /* ... */, transition: /* ... */ }
    // ...
  ]
}
```

[See the snapshot](https://github.com/davidkpiano/xstate/compare/davidkpiano/graph-tograph?expand=1#diff-d7e2b71533634f8895d334b1611073f4) for more details.


packages/xstate-graph/src/graph.ts
```diff
@@ -319,3 +322,44 @@ export function getSimplePathsAsArray<
   const result = getSimplePaths(machine, options);
   return keys(result).map((key) => result[key]);
 }
+
+export function toGraph(stateNode: StateNode): DigraphNode {
```
Andarist, line 326: If this returns `DigraphNode` shouldnt we call it `toDigraph`?

packages/xstate-graph/src/types.ts
```diff
@@ -1,9 +1,64 @@
...
   state: StateValue | undefined;
 }
 
+export type JSONSerializable<T extends object, U> = T & {
+  toJSON: () => U;
+};
+
+export type DigraphLabel = JSONSerializable<
+  {
+    [key: string]: any;
```
Andarist, line 19: we are the only producers of those graphs so far, do we need to allow indexer properties? they make the consuming code more fragile to mistakes

davidkpiano: Yes, in some layout algorithms, label nodes can contain width/height and other meta data. This shouldn't be restricted.

Andarist: That would mean that people would have to attach properties manually after calling `toGraph` or they would have to provide compatible graph structures (containing extra metadata) to us - right now we have no tools/utils for consuming those. We only produce them - so the question is: we have to include this already? If we want to make this type-safe we could also accept extra `Extension` generic for those types, rather than allowing yolo indexers 😅

davidkpiano: That's fair, I'll remove it for now.

packages/xstate-graph/src/graph.ts
davidkpiano: I'll rename to `toDirectedGraph`

packages/xstate-graph/src/graph.ts
```diff
@@ -13,8 +13,11 @@ import {
   StatePaths,
   AdjacencyMap,
   Segments,
-  ValueAdjMapOptions
+  ValueAdjMapOptions,
+  DirectedGraphEdge,
+  DirectedGraphNode
 } from './types';
+import { getChildren } from 'xstate/lib/stateUtils';
```
Andarist, line 20: What's the exact purpose of `@xstate/graph`? Seems like such an algorithm (`getChildren`) would, maybe, be a good candidate to be included here?

Andarist (approved):
What not to like about this one? 😉 Great job!

Just please add an appropriate changeset so this will get added to the next release.

packages/xstate-graph/src/graph.ts
davidkpiano: Good call

Issue #989, opened 2020-02-09
Complete context rewrite, not assign


nikogosovd:
The currently existing assign action cannot delete properties of the context object. It only assigns new properties or changes existing ones.
Is there a way in xstate to assign a completely new object to the context in action?

If not, I see this functionality as a new function e.g. `replace` with the API similar to the existing `assign` function. It takes the context "replacer", which represents how the current context should look.
The "replacer" can be an object or a function and the main difference is that the context is completely replaced with the new object.

davidkpiano:
Yes, you can change the entire context (replace it) like this:

```js
// assume you have { one: 1, two: 2 }

actions: assign(ctx => {
  return { one: ctx.one }; // delete two
});
```

See here: https://xstate.js.org/docs/guides/context.html#assign-action

nikogosovd:
Thank you for the immediate reply! Actually I tried the mentioned approach and it doesn't work as you described. Here is the reproduction: https://codesandbox.io/s/vigorous-volhard-f0en1

Am I doing something wrong?

davidkpiano:
Would it be sufficient to just set `{ unwantedProp: undefined }`?

nikogosovd:
Actually not, because when I’m saving a context with undefined props in MongoDB, it replaces them with nulls. Yes, there is an option in mongo driver that prevents this behaviour, but MongoDB drivers must not be responsible for xstate’s unexpected behaviour.

With all respect to the effort spent on xstate and appreciation for the hard work of supporting and improving the code, the fact that xstate doesn’t work as you, the creator of xstate, described means that the issue requires a fix, not a workaround solution.

Andarist:
I agree that I would expect the same result as @nikogosovd suggests (and which @davidkpiano also thought that it would be the case), but I'm not sure if everyone would expect it the same way. Given the fact that this logic is already in place for quite some time already, I would consider changing it right now unsafe because it could break somebody badly.

That being said - I believe that this could be tackled in upcoming v5, we only would need to check how it relates to SCXML spec. From the top of my head, it seems that there is no way in it to get "rid off" some datamodel value beyond doing explicitly what @davidkpiano has suggested - assigning undefined to all other properties.

davidkpiano:
> Actually not, because when I’m saving a context with undefined props in MongoDB, it replaces them with nulls. Yes, there is an option in mongo driver that prevents this behaviour, but MongoDB drivers must not be responsible for xstate’s unexpected behaviour.

[...]

PR #1367, opened 2020-08-08
[v5] Assigner function should completely replace context

davidkpiano:
This PR addresses #989 in that `assign(fn)` should _completely_ replace context, instead of partially replacing it.

```js
// ...
{
  context: {
    name: 'David',
    temp: 'value'
  },
  // ...
  entry: assign((ctx, event) => {
    return {
      name: event.name
    }
  })
}
// The context will now be, e.g.:
// { name: 'Jenny' }
// instead of:
// { name: 'Jenny', temp: 'value' }
```

### DIFF
--- a/packages/core/test/assign.test.ts
+++ b/packages/core/test/assign.test.ts
@@ -8,7 +8,6 @@ import {
 } from '../src';
 import { invokeMachine } from '../src/invoke';
 import { ActorRef } from '../src/types';
-import { stringify } from 'querystring';
 
 interface CounterContext {
   count: number;
--- a/packages/core/test/assign.test.ts
+++ b/packages/core/test/assign.test.ts
@@ -1,6 +1,14 @@
-import { Machine, interpret, assign, send, sendParent, State } from '../src';
+import {
+  createMachine,
+  interpret,
+  assign,
+  send,
+  sendParent,
+  State
+} from '../src';
 import { invokeMachine } from '../src/invoke';
 import { ActorRef } from '../src/types';
+import { stringify } from 'querystring';
 
 interface CounterContext {
   count: number;
@@ -8,7 +16,7 @@ interface CounterContext {
   maybe?: string;
 }
 
-const counterMachine = Machine<CounterContext>({
+const counterMachine = createMachine<CounterContext>({
   initial: 'counting',
   context: { count: 0, foo: 'bar' },
   states: {
@@ -18,6 +26,7 @@ const counterMachine = Machine<CounterContext>({
           {
             target: 'counting',
             actions: assign((ctx) => ({
+              ...ctx,
               count: ctx.count + 1
             }))
           }
@@ -224,10 +233,28 @@ describe('assign', () => {
       maybe: 'defined'
     });
   });
+
+  it('does a complete context rewrite with a single function assigner', () => {
+    const machine = createMachine<{ foo: string; bar?: string }>({
+      context: {
+        foo: 'one',
+        bar: 'two'
+      },
+      initial: 'active',
+      states: {
+        active: {
+          entry: assign(() => ({ foo: 'changed' }))
+        }
+      }
+    });
+
+    expect(machine.initialState.context.bar).toBeUndefined();
+    expect(machine.initialState.context).toEqual({ foo: 'changed' });
+  });
 });
 
 describe('assign meta', () => {
-  const machine = Machine<{ count: number }>({
+  const machine = createMachine<{ count: number }>({
     id: 'assign',
     initial: 'start',
     context: { count: 0 },
@@ -308,7 +335,7 @@ describe('assign meta', () => {
       })
     }));
 
-    const childMachine = Machine({
+    const childMachine = createMachine({
       initial: 'bar',
       states: {
         bar: {}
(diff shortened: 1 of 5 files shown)


packages/core/test/assign.test.ts
```diff
@@ -18,6 +25,7 @@ const counterMachine = Machine<CounterContext>({
           {
             target: 'counting',
             actions: assign((ctx) => ({
+              ...ctx,
```
davidkpiano, line 28: This would be the simple workaround to maintain existing behavior.

davidkpiano:
@Andarist Is this good to go?

Andarist:
After some thinking - I think I'm opposed to this change. 

1. This deviates from the SCXML spec and conversion between both XState and SCXML will just get harder with this. I'm not sure if the overall plan is to make XState machines editable in SCXML-based tools but if it is then after changes in such a tool we won't be able to tell apart to what exactly should we convert it back.
2. What `assign` does right now relates 1 to 1 to the semantics of `Object.assign` that should be known by all JS developers.
3. As mentioned before [here](https://github.com/davidkpiano/xstate/issues/989#issuecomment-585466820) this kind of thing changes meaning over time, it seems too easy to introduce a new property to the context and have it removed by accident by such action. So I would say that extra consideration should be taken when introducing such "replace" action to the machine because future changes will always have to be aware of this being included in a machine. Arguably this argument could be said about other situations as well (probably) but in this very situation it seems especially prominent - it breaks "locality of changes". It's like a descendant selector in CSS - you want to change a local property but if there is a selector with higher specificity then your change won't take place. This is even more dangerous because your change might seem like taking an effect right after assigning a new property but it might get erased later - at an unknown point in time. So with CSS at least you know right from the start that you need to change your logic and here we can't do the same.

That being said - I recognize the need for smth like this. It's certainly useful from time to time, but maybe we should make a dedicated helper action for this? It should be fairly trivial to provide a new action based on `pure`+`assign`.

davidkpiano:
> That being said - I recognize the need for smth like this. It's certainly useful from time to time, but maybe we should make a dedicated helper action for this? It should be fairly trivial to provide a new action based on `pure`+`assign`.

Sure, so here's my plan B (not a breaking change):

```js
import { assign, reset } from 'xstate'; // compatible with v4

// ...
{
  context: {
    name: 'David',
    temp: 'value'
  },
  // Keeps original behavior
  // Context will still be { name: 'Jenny', temp: 'value' }
  entry: assign((ctx, event) => {
    return {
      name: event.name
    }
  })

  // Context completely reset to initial value:
  // { name: 'David', temp: 'value' }
  entry: reset(),

  // Context completely reset to custom value:
  // { name: '--' }
  entry: reset({ name: '--' }),

  // Context completely reset based on current context and event:
  // { name: 'Jenny' }
  entry: reset((ctx, event) => {
    return { name: event.value }
  })
}
```

Thoughts?

johnyanarella:
I generally find myself exclusively using the the function parameter variant of `assign()`. If I am changing more than one value in a given `assign()` action, they are typically interrelated and it makes sense to compute their values in the same `(ctx, event) => { ... }` closure. (Which is a bummer, because I know the object parameter variant provides better metadata for the visualization.)

So, my initial reaction to this PR was: "Oh, no! That's going to break *all* my assignments." That said--the migration is trivial and it is not unreasonable to have some breaking changes when moving to v5.

Still, my reservations against the `assign()` change are:

- it is a potential footgun until you really internalize the requirement to spread the original context
- it invalidates existing pre-v5 examples out there on the wild
- the friction remains between different expectations for how `assign()` should behave (what "full replacement" in the docs really means vs the `Object.assign()` precedent that @andarist mentioned)

...paired with the breaking changes, it sounds like this could potentially end up aggravating everyone (both newcomers and existing users alike).

Regarding `reset()` (prefaced with: naming things is *the worst*):

- the standalone `reset()` kinda reads like it resets the whole machine
- the non-standalone `reset((ctx, event) => { ... })` isn't really resetting anything back to something - it is replacing it
- this adds yet another idiom to the API that developers must learn - and such that there are two that are very similar but subtly different (probably unavoidable, though)

Maybe a different name?

rjdestigter:
> my initial reaction to this PR was: "Oh, no! That's going to break all my assignments." That said--the migration is trivial and it is not unreasonable to have some breaking changes when moving to v5.

Ditto for me. I do quite often just forward event payloads with partial state updates so I would have to tackle that.

> it is replacing it

Call it _replace_ instead of _reset_. Or maybe _reassign_?

johnyanarella:
Revisiting the premise--how much sense does it make to replace the whole context anyway? Why not put the context item that changes so dramatically into a property of the context? (i.e. a context property that is a union type of potentially different shapes rather than the whole context being one)

davidkpiano:
> Revisiting the premise--how much sense does it make to replace the whole context anyway? Why not put the context item that changes so dramatically into a property of the context? (i.e. a context property that is a union type of potentially different shapes rather than the whole context being one)

It doesn't make too much sense; what you're suggesting is valid. I like the `replace()` name better than `reset()`, thanks for the suggestion.

I'll close this so that nobody is afraid that we'll implement this. It is a potentially big breaking change.

johnyanarella:
Cool--seems the best of imperfect options. And now we have this to look back on when I completely forget why in about six months.

PR #1368, opened 2020-08-08
xstate-svelte

jamesopstad:
As promised, I've put together an `xstate-svelte` package. There are a few things to draw your attention to and questions to resolve.

1. In the fsm implementation, actions provided on `useMachine` are merged with existing actions. This is different to the the React and Vue versions but is, I think, more desirable.

2. I haven't created `useService` functions or returned the service from the `useMachine` functions. I feel that Svelte differs a bit from the other frameworks here because `xstate` services implement the Svelte store contract (https://svelte.dev/docs#Store_contract) and can therefore already be treated as stores. Please correct me if I'm wrong about this. I would be interested to hear your thoughts and whether there are any disadvantages to using services as stores directly. This equally applies to actors I think. Happy to implement these features if they are useful.

3. Currently, the service is started immediately when `useMachine` is called. Are there use-cases where this should instead happen when the component has mounted?

4. I used the `State` constructor rather than `State.create` for rehydrating state in order to resolve a type error that is present in the `xstate-vue` package.

5. I had to make a type change in `core/src/utils.ts` in order for the tests to run but there may be a better solution.

6. I had to use the latest version of Jest so this may need to be bumped for the project.

7. ~~The tests run fine when invoked from the `xstate-svelte` directory but currently fail if called from the root directory because it looks for a `svelte.config.js` file at the root of the project. I'm not sure if there's a way to resolve this currently so it may need to be raised as an issue with `svelte-jester`.~~

8. I haven't added a changeset yet because I wasn't sure how to answer the versioning questions.

**EDIT: The latest version of `svelte-jester` resolves the issue with locating the `svelte.config.js` file in a monorepo so I have made the necessary change.**


packages/xstate-svelte/src/fsm.ts
```diff
@@ -0,0 +1,53 @@
...
+export function useMachine<
+  TContext extends object,
+  TEvent extends EventObject,
+  TTypestate extends Typestate<TContext>
+>(
+  machine: StateMachine.Machine<TContext, TEvent, TTypestate>,
+  options: Partial<UseMachineOptions<TContext, TEvent>> = {}
+) {
+  const resolvedMachine = createMachine(machine.config, {
+    actions: {
```
davidkpiano, line 29: You can probably use `createMachine(machine.config).withOptions(...)` to handle the merging stuff for you.

jamesopstad: `withOptions` isn't available on `createMachine` in `@xstate/fsm`.

jamesopstad:
Should I look at updating `jest` and `ts-jest` project-wide (either here or in a separate PR) or are there reasons not to do this at present?

davidkpiano:
> Should I look at updating `jest` and `ts-jest` project-wide (either here or in a separate PR) or are there reasons not to do this at present?

Doesn't need to be done in this PR (unless there is a specific reason to), especially since in the v5 `next` branch, `ts-jest` isn't even used.

jamesopstad:
Sorry, I seem to have messed this up a bit. I pulled in the upstream changes but now some of those files have been added to this pull request. Any idea why the checks are failing?

Andarist:
Could you try to replace the `yarn.lock` on your branch with the one on the master branch and run `yarn` afterward? From what I see there is a high chance that the yarn.lock got corrupted/regenerated

packages/xstate-svelte/jest.config.js
```diff
@@ -0,0 +1,15 @@
+const { jest: lernaAliases } = require('lerna-alias');
+
+module.exports = {
+  moduleNameMapper: lernaAliases(),
+  transform: {
+    '^.+\\.svelte$': [
+      'svelte-jester',
+      {
+        preprocess: true,
+        rootMode: 'upward'
```
Andarist, line 10: isn't this an option related to Babel? we are not quite using it on the master branch, is it needed?

packages/xstate-svelte/src/fsm.ts
```diff
@@ -0,0 +1,53 @@
...
+      ...(machine as any)._options.actions,
+      ...options.actions
```
Andarist, line 31: Both Vue and React packages are **not** merging those:
https://github.com/davidkpiano/xstate/blob/9fe5cc5c7719e5fa245e883dc4ab60f102c2164f/packages/xstate-vue/src/fsm.ts#L42
https://github.com/davidkpiano/xstate/blob/9fe5cc5c7719e5fa245e883dc4ab60f102c2164f/packages/xstate-react/src/fsm.ts#L57

packages/xstate-svelte/jest.config.js
jamesopstad: `rootMode: 'upward'` is to allow a `svelte.config.js` inside a workspace folder rather than just at the project root. See https://github.com/mihar-22/svelte-jester/pull/15 and https://github.com/mihar-22/svelte-jester/issues/12

Andarist: TIL, quite weird - it kinda works differently than a very similar option in Babel 🤷‍♂️

packages/xstate-svelte/src/fsm.ts
jamesopstad: Yes, I saw that (mentioned in first comment). Wouldn't merging be the expected behaviour though?

packages/xstate-svelte/jest.config.js
jamesopstad: Yes, I agree the naming is confusing.

packages/xstate-svelte/src/fsm.ts
Andarist: IMHO there is more to consider here - and such a change might be introduced in the future but it should be introduced cohesively, for all integration packages. I would prefer to use the existing logic for this new package for now.

jamesopstad: Sure. I will make that change and then submit a PR in the future for changing this behaviour across the integrations.

jamesopstad: I've made this change now and also exposed the created service in both `useMachine` implementations to match the other integrations.

Andarist (approved):
LGTM but I have never used Svelte 😅

davidkpiano (approved):

davidkpiano:
> LGTM but I have never used Svelte 😅

This looks very similar to the [Svelte + XState template](https://codesandbox.io/s/xstate-svelte-template-jflv1), so this LGTM too.

Let's 🚢 !

jamesopstad:
Great. I'd be happy to create a PR for the README.

PR #1372, opened 2020-08-11
fix: use event generic in ServiceConfig type

christianchown:
Hi. Ran into a typing problem when using `MachineConfig`. The following code does not compile in TypeScript with `"strict": true`

```typescript
type TestContext = {};
type TestEvent =
  | { type: 'EVENT_WITH_PAYLOAD'; payload: string }
  | { type: 'OTHER_EVENT_WITH_PAYLOAD'; payload: number };

const services = {
  testService: (_context: TestContext, _event: TestEvent) => Promise.resolve(),
  testService2: (_context: TestContext, _event: TestEvent) => Promise.resolve(),
};

const partialMachineOptions: Partial<MachineOptions<TestContext, TestEvent>> = {
  services, 
  // Type ... is not assignable to type 'Record<string, ServiceConfig<TestContext, AnyEventObject>>'
  // ...
  // Types of parameters '_event' and 'event' are incompatible. 
  // Type 'AnyEventObject' is not assignable to type 'TestEvent'.
};

```

This is because `ServiceConfig` expects an `InvokeCreator` that uses `AnyEventObject`. 

In TypeScript's strict mode, `TestEvent` is not compatible with `AnyEventObject` as all TestEvents are inferred as having a `payload` property. 

This PR fixes the problem by having `ServiceConfig` (like the other types in `MachineOptions`) use the supplied event generic.

### DIFF
--- a/.changeset/nine-gorillas-brake.md
+++ b/.changeset/nine-gorillas-brake.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Narrowed the `ServiceConfig` type definition to use a specific event type to prevent compilation errors on strictly-typed `MachineOptions`.
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -836,8 +836,9 @@ export class Interpreter<
         // Invoked services
         if (activity.type === ActionTypes.Invoke) {
           const invokeSource = toInvokeSource(activity.src);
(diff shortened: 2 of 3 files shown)


davidkpiano:
Thanks! Would you mind adding a quick changeset?

christianchown:
thanks for reviewing so swiftly @davidkpiano, changeset added!

PR #1380, opened 2020-08-13
Minor tweak to make the behaviour explicit

Perlkonig:
For your consideration. Related to discussion #1377. Ideally we'd have a concrete example of child to parent, which I can't conceive of at the moment. I'm in the gaming headspace, and it makes sense there.


Perlkonig:
I don't think a changeset is needed, but I'm happy to generate one if you like.

davidkpiano:

PR #1385, opened 2020-08-15
Added isInline property to activities and services for easier introspection by xstate-codegen

mattpocock:
In developing xstate-codegen, I've been having a tricky time working out which activities are declared inline (i.e, in the state machine definition) and which are strings.

I don't know the difference between this:

```js
{
  invoke: {
    src: 'makeFetch'
  }
}
```

And this:

```js
{
  invoke: {
    src: () => Promise.resolve('data'),
  }
}
```

This is important because I need to figure out which services and activities are required for the MachineOptions type.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -186,6 +186,7 @@ export interface ActivityDefinition<TContext, TEvent extends EventObject>
   extends ActionObject<TContext, TEvent> {
   id: string;
   type: string;
+  isInline: boolean;
 }
 
 export type Sender<TEvent extends EventObject> = (event: Event<TEvent>) => void;
@@ -255,6 +256,10 @@ export interface InvokeDefinition<TContext, TEvent extends EventObject>
    * Data should be mapped to match the child machine's context shape.
    */
   data?: Mapper<TContext, TEvent, any> | PropertyMapper<TContext, TEvent, any>;
+  /**
+   * Whether or not the service is declared inline
+   */
+  isInline: boolean;
 }
 
 export interface Delay {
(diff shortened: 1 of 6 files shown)


packages/core/src/types.ts
```diff
@@ -186,6 +186,7 @@ export interface ActivityDefinition<TContext, TEvent extends EventObject>
   extends ActionObject<TContext, TEvent> {
   id: string;
   type: string;
+  isInline: boolean;
```
mattpocock, line 189:
```suggestion
  /**
   * Whether or not the activity is declared inline
   */
  isInline: boolean;
```

Andarist:
I'm very hesitant about merging such things to the master - especially given that right now XState codegen is just an experiment (albeit very promising one!).

Could you describe in more detail why differentiating those 2 is hard? Ir should be easy to distinguish strings from other types, shouldn't it? 🤔

mattpocock:
It is, if you're looking at the config object, but we're not. We parse the machine and look at the `StateNode` itself, which has activities and invokes described as `ActivityDefinition` and `InvokeDefinition` respectively. On those outputs, there is no way for us to tell if the service/activity was declared inline or not.

mattpocock:
Also, I'm happy for this PR to hang about unmerged until we've fully built out xstate-codegen. No rush to merge.

davidkpiano:
> It is, if you're looking at the config object, but we're not. We parse the machine and look at the `StateNode` itself, which has activities and invokes described as `ActivityDefinition` and `InvokeDefinition` respectively. On those outputs, there is no way for us to tell if the service/activity was declared inline or not.

You're in luck; each `StateNode` has a `.config` property that references the exact slice of the original `config` that was passed into `createMachine(config)` or `Machine(config)`.

That can probably be used!

mattpocock:
Ooo, if it's the exact slice then all is well. I'll check it out tomorrow and close this PR if it works. Thanks for the help.

mattpocock:
OK, the config thing didn't quite work but I found another way through.

PR #1388, opened 2020-08-18
updated State interface value type

hudecsamuel:
Fixes typing of value in State interface 
I was using an enum for the states and discovered that the `value` is still strings no matter what

```ts
export enum States {
  flow = 'flow',
  search = 'search',
  default = 'default',
}

type HeaderState =
  | {
      value: States.flow
      context:{
    }
  | {
      value: States.search
      context: {}
    }
  | {
      value: States.default
      context: {}
    }

const machine = createMachine<HeaderMachineContext, HeaderEvents, HeaderState>(...)

...

machineState.value // the type should be `States` instead of `string`
```

### DIFF
--- a/.changeset/modern-mugs-lie.md
+++ b/.changeset/modern-mugs-lie.md
@@ -2,4 +2,4 @@
 '@xstate/fsm': patch
 ---
 
-Improved typing of State interface
+`State['value']` is now correctly typed to `TState['value']`. It's important in situations when typestates are used as it now correctly is limited to values of those typestates and not widened to just `string`.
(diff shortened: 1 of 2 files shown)


Andarist:
@hudecsamuel could you include a changeset for this change?

hudecsamuel:
@Andarist Done.

.changeset/modern-mugs-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/fsm': patch
+---
+
+Improved typing of State interface
```
Andarist, line 5:
```suggestion
`State['value']` is now correctly typed to `TState['value']`. It's important in situations when typestates are used as it now correctly is limited to values of those typestates and not widened to just `string`.
```

Issue #1397, opened 2020-08-20
Multiple transitions in invoked state machines.


nathanhammond:
**Description**
Since 460e831, state transitions (when descendant from an `invoke`d `StateNode`) will occur the same number of times as the `StateNode` is nested.

**Expected Result**
The transition occurs exactly once. The `assign` function is called exactly once.

**Actual Result**
The transition occurs twice. The `assign` function is called twice. (The number of invocations is dependent upon its depth in the hierarchy.)

**Reproduction**
These two machines are 100% identical, the only change is the `xstate` dependency version:
- [Working - 4.10](https://codesandbox.io/s/optimistic-nash-tx7vj)
- [Issue - 4.11](https://codesandbox.io/s/sweet-shannon-dvzox)

nathanhammond:
### Bisecting

Bisect steps, in case you wish to reproduce. (Including a diff to make it build on one of the steps.)

```patch
commit 5813a5933a08acd54cf95b2fc1027d7ef4647135
Merge: cc65d1c7 d0aa23f5
Author: Nathan Hammond <github.com@nathanhammond.com>
Date:   Thu Aug 20 21:19:03 2020 +0800

    WIP on (no branch): cc65d1c7 Refactor useConstant calls

diff --cc packages/core/src/actions.ts
index f44e66e2,f44e66e2..12e08599
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@@ -106,7 -106,7 +106,6 @@@ export function toActionObject<TContext
        const { type, ...other } = action;
  
        actionObject = {
--        type,
          ...exec,
          ...other
        } as ActionObject<TContext, TEvent>;
diff --cc packages/core/src/interpreter.ts
index ac28865c,ac28865c..39ab9748
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@@ -259,7 -259,7 +259,7 @@@ export class Interpreter
      }
    }
    private update(
--    state: State<TContext, TEvent, TStateSchema, TTypestate>,
++    state: State<TContext, TEvent, TStateSchema, any>,
      _event: SCXML.Event<TEvent>
    ): void {
      // Attach session ID to state
```

```sh
git clone git@github.com:davidkpiano/xstate.git
cd xstate
yarn
cd packages/core && yarn link && cd ../../
git bisect start
git bisect good xstate@4.10.0
git bisect bad xstate@4.11.0
[...]

PR #1398, opened 2020-08-21
Prevent multiple `.initialState` reads.

nathanhammond:
Boolean flag proof of concept. Existing tests still pass.

***

This is a *straw man* solution for #1397. The approach appears to address the underlying issue, but to my unfamiliar eyes I'm unsure if it is a complete solution. Also, I only ran the xstate tests, it's unclear if downstream dependencies will rely on the existing behavior.

I personally hate adding boolean flags into method signatures, especially when they have to be drilled down into a much later invocation. I have no issue with trying to find a different way to accomplish this task.

Possibly the most-important step here is to land test cases for this problem to ensure correctness.

**Next Steps (unordered)**
- [ ] Identify if this (or something similar) is a palatable solution.
- [ ] Review for *actual* correctness instead of implied correctness by somebody who knows the lib.
- [ ] Port test case from [sandbox](https://codesandbox.io/s/sweet-shannon-dvzox).
- [ ] Add additional test cases for adjacent scenarios.

Thoughts?

### DIFF
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -44,6 +44,7 @@ export function createNullActor(id: string): Actor {
  * @param invokeDefinition The meta information needed to invoke the actor.
  */
 export function createInvocableActor<TC, TE extends EventObject>(
+  isInitial: Boolean,
   invokeDefinition: InvokeDefinition<TC, TE>,
   machine: StateMachine<TC, any, TE>,
   context: TC,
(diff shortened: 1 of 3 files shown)


nathanhammond:
The tests failed. Should have been foreseeable, the `tempActor`'s `state` is not set. @Andarist's caching thought was correct.

A reference which is late-bound to the `tempActor.state` might be viable.

PR #1407, opened 2020-08-24
allow running xstate in a web-worker

tomenden:
Currently, the code of the interpreter assumes that the code runs either in the browser's main thread (`window` is defined) or in node. This PR allows it to run in a web-worker, in which the _global_ is `self`

### DIFF
--- a/.changeset/nervous-shoes-retire.md
+++ b/.changeset/nervous-shoes-retire.md
@@ -1,5 +1,5 @@
 ---
-'xstate': minor
+'xstate': patch
 ---
 
-Add support for web-workers
+Fixed an issue with not being able to run XState in Web Workers due to assuming that `window` or `global` object is available in the executing environment, but none of those are actually available in the Web Workers context.
--- a/.changeset/nervous-shoes-retire.md
+++ b/.changeset/nervous-shoes-retire.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Add support for web-workers
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -124,13 +124,7 @@ export class Interpreter<
     },
     logger: global.console.log.bind(console),
     devTools: false
-  }))(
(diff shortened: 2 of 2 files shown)


tomenden:
> This PR includes changesets to release 7 packages
I think this means I probably messed up the changeset, right? I did not mean to create any major releases.

tomenden:
I could not find anywhere in xstate docs what Node.js version it is aimed for.
Assuming you are fine with supporting Node.js 12+, global/window/self can be replaced by `globalThis`
(However, this would make it a breaking change, as usage of xstate in earlier versions of node would break)

packages/core/src/interpreter.ts
```diff
@@ -124,7 +124,13 @@ export class Interpreter<
     },
     logger: global.console.log.bind(console),
     devTools: false
-  }))(typeof window === 'undefined' ? global : window);
+  }))(
+    typeof window !== 'undefined'
```
Andarist, line 128: We could just use `self` over `window` as `window.self` is supported in all popular browsers

.changeset/nervous-shoes-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Add support for web-workers
```
Andarist, line 5:
```suggestion
---
'xstate': patch
---

Fixed an issue with not being able to run XState in Web Workers due to assuming that `window` or `global` object is available in the executing environment, but none of those are actually available in the Web Workers context.
```

Andarist:
> Assuming you are fine with supporting Node.js 12+, global/window/self can be replaced by globalThis

Can't do that right now due to the fact that (like you have mentioned) this would be a breaking change.

packages/core/src/interpreter.ts
tomenden: Done

tomenden:
@Andarist Updated the PR.

Andarist (approved):

PR #1409, opened 2020-08-24
fix(core): fix type ExtractStateValue to describe real state value

jirutka:
Consider the following schema:

    type LightStateSchema = {
      states: {
        green: {},
        yellow: {},
        red: {
          states: {
            walk: {},
            wait: {},
            stop: {},
          },
        },
      },
    }

The current implementation of the ExtractStateValue type generates the
following type:

    {
      green?: {} | undefined
      yellow?: {} | undefined
      red?: "walk" | "wait" | "stop" | {
        walk?: {} | undefined
        wait?: {} | undefined
        stop?: {} | undefined
      } | undefined
    }

This does not describe the StateValue of LightStateSchema as provided
in State.value.

The new implementation generates the following type:

    "green" | "yellow" | "red" | {
      red?: "walk" | "wait" | "stop" | undefined
    }

### DIFF
--- a/.changeset/ten-seahorses-smile.md
+++ b/.changeset/ten-seahorses-smile.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Fix type `ExtractStateValue` so that it generates a type actually describing a `State.value`
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -92,16 +92,25 @@ export interface StateValueMap {
  */
 export type StateValue = string | StateValueMap;
 
+type KeysWithStates<
+  TStates extends Record<string, StateSchema> | undefined
+> = TStates extends object
+  ? {
+      [K in keyof TStates]-?: TStates[K] extends { states: object } ? K : never;
+    }[keyof TStates]
+  : never;
+
 export type ExtractStateValue<
-  TS extends StateSchema<any>,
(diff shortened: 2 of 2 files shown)


.changeset/ten-seahorses-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Fix type ExtractStateValue so that it generates a type actually describing a State.value
```
davidkpiano, line 5:
```suggestion
Fix type `ExtractStateValue` so that it generates a type actually describing a `State.value`
```

Andarist (approved):

.changeset/ten-seahorses-smile.md
jirutka: Fixed and amended to the last commit.

AliKholmatov:
image
Hi! In latest versions ExtractStateValue was deleted. What can I use instead of ExtractStateValue? Please)

Andarist:
Could you open a discussion that would describe your use case? For context, this type was removed in https://github.com/statelyai/xstate/pull/2828 as we've assumed that it was a leftover from a previous implementation and that it wasn't used externally.

AliKholmatov:
@Andarist sorry I can't answer right now, I will try figure out. I guess it was used to made deal with TS

Issue #1403, opened 2020-08-22
`initialState` lifecycle seems to miss "after" transition.


nathanhammond:
**Description**
State transitions in the `initialState` or immediately following do not seem to follow the full lifecycle. They instead appear to be possibly reentrant.

**Expected Result**
This state machine configuration should end up in an infinite loop. Each spawned `invocable-promise` should have `.stop()` called on it after 2 seconds, well before the 9 seconds that the promise takes to resolve.

**Actual Result**
The first `invocable-promise` spawned by `cancellable-promise` in `initialState` does not ever have `.stop()` called on it. Eventually this first `invocable-promise` resolves, pushing the parent state machine, `cancellable-promise`, to a `resolve`d state. All subsequent spawned machines (after the `initialState` one) do have `.stop()` successfully called on them.

**Reproduction**
Tested against 4.11, 4.12, and `HEAD` of `master`. All demonstrate the same behavior.

[4.11 test case](https://codesandbox.io/s/quizzical-greider-gf6q0)

**Additional context**
Converted to old syntax (away from `always`) to enable dropping it in the visualizer. Includes additional context from parent machines that is unused in the example code. Inline comments in the test cases.

Andarist:
Thanks for reporting the issue and providing the repro case. I will try to look into it as soon as I can.

Andarist:
This is currently broken on our side pretty badly I would say.

One problem is that we prioritize `assign` actions and they all get executed (within a single transition) before other actions - therefore when your stopping action gets called we have already assigned a new value to the `ctx.promise` (even though the order of those operations is defined to be the opposite). There is no clean workaround for this apart of splitting both things to 2 separate macro transitions (triggered by 2 external events). This has been fixed already on the `next` branch but as this behavior was documented we can't change it without releasing a new major version.

One thing that you shouldn't do is that you shouldn't call `.stop()` manually like this - we maintain parent/child relationship between those 2 machines and we keep references to each other in our internal structures. By calling `.stop()` manually you bypass our stopping logic and we have no chance to clean up after this - therefore this causes a memory leak.

How one should stop a spawned child then? That's an excellent question 😉 I believe that the best option right now is to use `stop` action. There is no documentation around this and also we don't really have good tests for this, so take this with a grain of salt.

This action accepts only string identifiers for the machine's children and to get that you have to rely on `pure` right now (as it gets access to `ctx` and can return other actions). This is a working example:
```js
[...]

PR #1434, opened 2020-09-03
Document stopping actors.

nathanhammond:
This PR includes two pieces of information that @Andarist explained to me in issue #1403, `actions.pure()` and `actions.stop()`.

However, after attempting to document (and reading the code in `interpreter.ts` for 4.X), I have additional questions that imply to me that there may be additional unaddressed issues here.
- Just sending `xstate.stop` to a `StateNode` does not seem to be complete to me as it doesn't seem like it is recursive. The receiver of `xstate.stop` may also be a machine with children which would *also* need to be stopped. The event does not appear to be propagated. [Code.](https://github.com/davidkpiano/xstate/blob/78e998674b8ac8cbcf926f43b2588787b5631953/packages/core/src/StateNode.ts#L1158-L1160)
- It seems like you should still call `.stop()` on the interpreter as well. [Code.](https://github.com/davidkpiano/xstate/blob/78e998674b8ac8cbcf926f43b2588787b5631953/packages/core/src/interpreter.ts#L476)

Assuming that my hypotheses are correct, maybe `Interpreter#stop` should also send (from leaf to root?) `xstate.stop`?

### DIFF
--- a/docs/guides/actors.md
+++ b/docs/guides/actors.md
@@ -332,6 +332,79 @@ parentService.send('LOCAL.WAKE');
 // => 'connected'
 ```
 
+### Stopping Spawned Machines
+
+The object returned from a successful `spawn` action is an `Interpreter`. It implements the `actor` interface, but has slightly different behavior. If you call `stop()` on an `Interpreter` it does not clean up any ongoing `activities` in the active `StateNode`. To properly stop spawned machines you must additionally send them an action using `actions.stop()`.
+
+Further, as `assign` actions are run in advance of custom user actions, by the time you would usually attempt to stop a spawned machine, you may need to take care that the reference to the machine in `context` is not overwritten.
+
+For actions which do not have side effects (such as mutating `context`), you can use `actions.pure()` to ensure that your custom actions is run in order of definition, and before `assign` actions' updates are applied to `context`. Depending upon why you need to stop a spawned machine, this technique may be required.
+
+The below machine demonstrates how to use `actions.pure()` and `actions.stop()` to properly stop a consistently respawned machine.
+
+```js {18-20,27-29,35}
+const childMachine = Machine({
+  id: 'child-machine',
+  initial: 'idle',
+  states: {
+    idle: {
+      /* ... */
+    }
+  }
+});
+
+const parentMachine = Machine({
+  id: 'stoppable-machine',
+  initial: 'spawn',
+  context: {},
+
+  states: {
+    reset: {
+      always: {
+        target: 'spawn',
+        actions: actions.pure((context) => {
+          context.childMachine.stop();
+          return actions.stop(context.childMachine.id);
+        })
+      }
+    },
+
+    spawn: {
+      always: {
+        target: 'waiting',
+        actions: assign({
+          childMachine: () => spawn(childMachine)
+        })
+      }
+    },
+
+    waiting: {
+      after: {
+        2000: 'reset'
+      }
+    }
+  }
+});
+```
+
+Using `actions.pure()` results in the properly interleaved ordering:
+
+1. `assign({ childMachine: () => spawn(childMachine) })` (First time through.)
+1. `actions.stop(context.childMachine.id)` (Second time through.)
+1. `assign({ childMachine: () => spawn(childMachine) })` (Second time through.)
+1. `actions.stop(context.childMachine.id)` (Third time through.)
+1. `assign({ childMachine: () => spawn(childMachine) })` (Third time through.)
+
+If you were to omit `actions.pure()` the order of actions would be:
+
+1. `assign({ childMachine: () => spawn(childMachine) })` (First time through.)
+1. `assign({ childMachine: () => spawn(childMachine) })` (Second time through.)
+1. `actions.stop(context.childMachine.id)` (Second time through.)
+1. `assign({ childMachine: () => spawn(childMachine) })` (Third time through.)
+1. `actions.stop(context.childMachine.id)` (Third time through.)
+
+This results in immediately stopping all spawned machines except for the first one, which instead continues running.
+
 ## Syncing and Reading State <Badge text="4.6.1+"/>
 
 One of the main tenets of the Actor model is that actor state is _private_ and _local_ - it is never shared unless the actor chooses to share it, via message passing. Sticking with this model, an actor can _notify_ its parent whenever its state changes by sending it a special "update" event with its latest state. In other words, parent actors can subscribe to their child actors' states.


docs/guides/actors.md
```diff
@@ -332,6 +332,79 @@ parentService.send('LOCAL.WAKE');
...
+  initial: 'spawn',
+  context: {},
+
+  states: {
+    reset: {
+      always: {
+        target: 'spawn',
+        actions: actions.pure((context) => {
+          context.childMachine.stop();
+          return actions.stop(context.childMachine.id);
```
nathanhammond, line 367: I included both versions of `stop` here because I *believe* that they are both required.

Andarist (commented):
Thank you for doing this, such docs PRs from external contributors are super rare and yet they are very, very valuable, often more valuable than code contributions. Much love for that ❤️ 

I have left some minor comments, the overall content is very good and very needed for our users. I bet others had similar problems in the past and they couldn't figure out how to do this properly with the documentation alone.

As to the whole stopping thing - yes, `pure` is required now, but it should be fairly easy to just allow for its dynamic resolution (`stop(ctx => ctx.child)`) and I think we should implement this unless @davidkpiano has a better idea about how one should manage stopping of the spawned actors. 

The PR itself doesn't have to be adjusted to account for those possible changes though - it should describe current possibilities. If we decide to move with the dynamic resolution of `stop` I will be able to implement it rather quickly and adjust the content of those paragraphs here accordingly.

Andarist:
Oh, I've missed your questions in the PR comment itself. Let's get to them:
> Just sending xstate.stop to a StateNode does not seem to be complete to me as it doesn't seem like it is recursive. The receiver of xstate.stop may also be a machine with children which would also need to be stopped. The event does not appear to be propagated. Code.

`StateNode` is a "pure" representation of a machine, it's stateless but the information in it holds a list (or a map, don't remember) of things like `activities and `children`, those shouldn't be "live" things though.

OTOH the `Interpreter` is this "live" thing that manages the pure machine that it references. It receives actions that should be executed (apart from `assign`s, those are completely evaluated & resolved within `StateNode`). In here it stops a particular child upon receiving a `stop` action:
https://github.com/davidkpiano/xstate/blob/78e998674b8ac8cbcf926f43b2588787b5631953/packages/core/src/interpreter.ts#L886
and stop should recursively stop all "nested" machines as well:
https://github.com/davidkpiano/xstate/blob/78e998674b8ac8cbcf926f43b2588787b5631953/packages/core/src/interpreter.ts#L492-L497

This should also answer the second question as those were highly interconnected.

docs/guides/actors.md
nathanhammond: I have not experienced issues, but my WIP machine is not complicated enough to trigger the thing I believe could be an issue. (And I wanted to get a first pass of the documentation together and ask before trying to create a repro of a hypothetical issue.)

My hypothesis comes out of scanning the implementations of these to get a bit more context before filing this PR. The test case that this would need is:

- Parent machine.
- Spawns a child *machine* (so that it is an `Interpreter`, not an actor).
- Child machine starts an activity (and stays in that state).
- Child machine's interpreter reference from parent machine's `context` has a listener attached to it.
- Test sending `xstate.stop`, calling `childMachine.stop()`, or doing both of those things.

Confirm:
- No activities still running. (I *think* this is what `xstate.stop` does.)
- No listeners still bound. (I *think* this is what `Interpreter#stop` does.)
- Make sure that it isn't just getting GC'd by way of clobbering all references.

I suppose that if the sent `xstate.stop` actually bubbled up into something that eventually triggered listener removal on the interpreter then it's good, but in a quick recheck I can't identify where that might happen.

docs/guides/actors.md
```diff
@@ -332,6 +332,79 @@ parentService.send('LOCAL.WAKE');
 // => 'connected'
 ```
 
+### Stopping Spawned Machines
+
+The object returned from a successful `spawn` action is an `Interpreter`. It implements the `actor` interface, but has slightly different behavior. If you call `stop()` on an `Interpreter` it does not clean up any ongoing `activities` in the active `StateNode`. To properly stop spawned machines you must additionally send them an action using `actions.stop()`.
+
+Further, as `assign` actions are run in advance of custom user actions, by the time you would usually attempt to stop a spawned machine, you may need to take care that the reference to the machine in `context` is not overwritten.
+
+For actions which do not have side effects (such as mutating `context`), you can use `actions.pure()` to ensure that your custom actions is run in order of definition, and before `assign` actions' updates are applied to `context`. Depending upon why you need to stop a spawned machine, this technique may be required.
```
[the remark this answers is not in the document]
nathanhammond: This PR is only half of the things from #1403 I assigned myself to write. I do plan to drop some more information into [Actions](https://xstate.js.org/docs/guides/actions.html#pure-action) for ordering, and maybe in [Context](https://xstate.js.org/docs/guides/context.html#action-order) too.

I went ahead and included a first pass of that information here; I think that "JIT" reinforcement of this concept is useful.

docs/guides/actors.md
```diff
@@ -332,6 +332,79 @@ parentService.send('LOCAL.WAKE');
 // => 'connected'
 ```
 
+### Stopping Spawned Machines
+
+The object returned from a successful `spawn` action is an `Interpreter`. It implements the `actor` interface, but has slightly different behavior. If you call `stop()` on an `Interpreter` it does not clean up any ongoing `activities` in the active `StateNode`. To properly stop spawned machines you must additionally send them an action using `actions.stop()`.
```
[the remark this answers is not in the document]
nathanhammond: Most of this is paragraph is based upon my hypothesis of a possible issue existing. I'll pull together that test case and get back to you on that.

Other notes:
- I will revise to add more of the conceptual model components you describe. That's actually the thing that helps the most from the conversations we've been having in different issue threads. Though I've turned it into a recipe, teaching the conceptual model is probably more important.
- Yeah, got a bit into the weeds.
- I also need to compare this to V5 so I can write this maybe generically enough to apply in both worlds.

nathanhammond:
How do y'all prefer your modifications to PRs?
A. New commits.
B. Update-in-place (continuously squash to a single commit).
C. Other

And prepping for landing:
A. Squash
B. Commit History
C. Other

And landing strategy:
A. Merge
B. Rebase
C. Other

And test strategy:
A. Failing test in a separate commit first. (red => green)
B. Tests pass at every commit.
C. Other

(Can also drop this into `CONTRIBUTING.md` if you'd like.)

docs/guides/actors.md
nathanhammond: Okay, I have created the test case:

https://codesandbox.io/s/brave-fermi-nyg8e?file=/src/index.js

My findings:
- Either one of `Interpreter#stop` or `actions.stop()` seem to result in proper teardown. (One of the `stopListeners` sends a `session.stop` via `Interpreter#stop`.) 
- What I thought was happening with `activities` doesn't appear to be relevant. I'm not sure why.

I also tried to check the behavior for sending `actions.stop()` as an event to a top-level `Interpreter` which appears to not work, but I'm not entirely sure what it's doing.

@Andarist Can you also take a look at this test case yourself to make sure that my understanding is correct and that my conclusions are valid?

If my analysis holds, I believe that `Interpreter#stop`/`<actor>.stop()` is actually a complete solution.

Andarist:
Sorry for the delay - was quite busy lately with other stuff. I haven't forgotten and will get back to you this week.

docs/guides/actors.md
Andarist: > @Andarist Can you also take a look at this test case yourself to make sure that my understanding is correct and that my conclusions are valid?

Sure thing! I've forked your demo so it doesn't restart the spawned child all over again - it was easier for me to inspect stuff this way: https://codesandbox.io/s/magical-lamarr-b09k1?file=/src/index.js 

> Either one of Interpreter#stop or actions.stop() seem to result in proper teardown. (One of the stopListeners sends a session.stop via Interpreter#stop.)

if we are talking about calling those from within that `pure` action then structures within child get cleaned up nicely but `context.childMachine.stop()` doesn't clean this child from the parent's structure, whereas `actions.pure((context) => actions.stop(context.childMachine.id) })` does both correctly.

> I also tried to check the behavior for sending actions.stop() as an event to a top-level Interpreter which appears to not work, but I'm not entirely sure what it's doing.

`actions.stop` doesn't create an event (well, actually it is compatible with an event interface so it could be abused, sort of, to be sent and processed). It creates an action object - think about actions as of instructions for the interpreter that you'd like something to happen. Those actions are received by the interpreter through other means though than by sending it events so to make `service.send(actions.stop())` work you would have to create a transition for the `xstate.stop` event and create a stop action there. This is not what you are looking for though, I'm just describing this for completeness. 

> // ALSO: Test the parent behavior.
> // service.stop();

This cleans up stuff correctly and I have confirmed that using this demo.

> // HELP: Is this correct for trying to send a stop action to an un-parented Interpreter?
> // service.send(actions.stop(service.id));

Do you want to stop the `service` here or some child service of a particular id? If the first one it should be totally valid to just call `stop` method like this `service.stop()`.

> // 2. Check to see if `window.childMachineFirst`'s StateNode has activities.

It seems that we do not clean children when stopping but they are being stopped. So references are held to them - gonna prepare a fix for this later as semantically it seems right the good thing to do. The child services are being stopped though so GC should kick in most of the time and collect the whole tree of services.

> // 3. Check to see if `window.childMachineFirst` has listeners.

This works correctly - those structures are being cleaned up just OK.

Andarist:
> How do y'all prefer your modifications to PRs?

IMHO - whatever works best for you. Ideally, we'd like to receive often contributions from the community, and requiring a particular git-flow makes it harder to achieve that, not everyone is well-versed with more advanced git techniques etc. As to the landing strategy - this one can actually be quite easily chosen from the GitHub UI so it's not much of a concern either way.

davidkpiano:
@nathanhammond Sorry for the delay in reviewing this! @Andarist and I discussed the need for a dynamic `stop(ctx => ctx.child)` and how it could be useful (and simpler) with regard to this, so I'll make those changes and then re-review this PR with the necessary changes.

Thanks so much for doing this!

davidkpiano:
Now that #1577 is in, we can simplify these docs:

-----

### Stopping Spawned Machines

To stop a spawned machine, use the `actions.stop(...)` action creator and reference the spawned machine as an expression in the first argument:

```js
on: {
  STOP_SPAWNED: {
    actions: actions.stop(context => context.someSpawnedMachine)
  }
}
```

davidkpiano:

PR #1439, opened 2020-09-05
Feature: createModel()

davidkpiano:
This PR adds support for an opt-in helper function called `createModel()`, which is meant to assist in creating the [datamodel (SCXML)](https://commons.apache.org/proper/commons-scxml/guide/datamodel.html) for the machine:

```js
import { createModel } from 'xstate/lib/model'; // opt-in, not part of main build

const userModel = createModel({
  name: 'David',
  age: 30
}).withUpdaters({
  updateName: {
    name: (_, e) => e.value // assignments for assign()
  }
});

  context: userModel.context,
  initial: 'active',
  states: {
    active: {
      on: {
        updateName: {
          actions: userModel.actions.updateName
        }
      }
    }
  }
});
```

### DIFF
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -1,0 +1,35 @@
+import { assign } from './actions';
+import { AssignAction, Assigner, EventObject, PropertyAssigner } from './types';
+import { mapValues } from './utils';
+
+export function createModel<TContext, TEvent extends EventObject>(
+  initialState: TContext
+): {
+  context: TContext;
+  withUpdaters: <
+    T extends {
+      [key: string]:
+        | Assigner<TContext, TEvent>
+        | PropertyAssigner<TContext, TEvent>;
+    }
+  >(
+    assigners: T
+  ) => {
+    context: TContext;
+    actions: {
+      [K in keyof T]: AssignAction<TContext, TEvent>;
+    };
+  };
+} {
+  return {
+    context: initialState,
+    withUpdaters: (assigners) => {
+      return {
+        context: initialState,
+        actions: mapValues(assigners, (assignment) => {
+          return assign(assignment);
+        }) as any
+      };
+    }
+  };
+}
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -343,7 +343,7 @@ export type TransitionConfigOrTarget<
   TransitionConfigTarget<TContext, TEvent> | TransitionConfig<TContext, TEvent>
 >;
 
-type TransitionsConfigMap<TContext, TEvent extends EventObject> = {
+export type TransitionsConfigMap<TContext, TEvent extends EventObject> = {
   [K in TEvent['type']]?: TransitionConfigOrTarget<
     TContext,
     TEvent extends { type: K } ? TEvent : never
--- a/packages/core/test/model.test.ts
+++ b/packages/core/test/model.test.ts
@@ -8,7 +8,12 @@ describe('createModel', () => {
       value: string;
     };
 
-    const userModel = createModel({
+    interface UserContext {
+      name: string;
+      age: number;
+    }
+
+    const userModel = createModel<UserContext, UserEvent>({
       name: 'David',
       age: 30
     }).withUpdaters({
--- a/packages/core/test/model.test.ts
+++ b/packages/core/test/model.test.ts
@@ -1,0 +1,41 @@
+import { createMachine } from '../src';
+import { createModel } from '../src/model';
+
+describe('createModel', () => {
+  it('model.machine creates a machine that is updated', () => {
+    type UserEvent = {
+      type: 'updateName';
+      value: string;
+    };
+
+    const userModel = createModel({
+      name: 'David',
+      age: 30
+    }).withUpdaters({
+      updateName: {
+        name: (_, e: UserEvent) => e.value
+      }
+    });
+
+    const machine = createMachine<typeof userModel['context'], UserEvent>({
+      context: userModel.context,
+      initial: 'active',
+      states: {
+        active: {
+          on: {
+            updateName: {
+              actions: userModel.actions.updateName
+            }
+          }
+        }
+      }
+    });
+
+    const updatedState = machine.transition(undefined, {
+      type: 'updateName',
+      value: 'Anyone'
+    });
+
+    expect(updatedState.context.name).toEqual('Anyone');
+  });
+});


johnyanarella:
I was just thinking about how the current uncertainty around #989 and #1367 confounds this effort. The API surface area gets less elegant when once you have to require the developer to differentiate between `assign` and `reset`. 

But that begs the question... does it even make sense for the entire model shape to change? Isn't that an entirely different animal--a union type where each variant might expose different available updaters...

davidkpiano:
> But that begs the question... does it even make sense for the entire model shape to change (as in the `reset` option)? Isn't that an entirely different animal--a union type where each variant might expose different available updaters...

Most of the time it doesn't, but there are usecases highlighted in those issues that demonstrate the need.

johnyanarella:
Oh, for sure - definitely didn't mean to come across as dismissive of those needs at all.

Just thinking aloud today as the 💡 slowly flickered on in my head, realizing that those concerns were really about the friction of representing a union type *using* a key/value store. (And the SCXML spec has essentially mandated a key/value store as the root data model type.) The imperfect emulation of that, and all the effort to emulate it, falls away when storing that union type as keyed value instead. Bonus: language support starts to come into play (both in TypeScript and Swift--which has enums with associated values). Obvious in retrospect, but I was just not seeing it clearly before.

I think that means this cool new `createModel()` effort gets to similarly side-step those worries, too? i.e. you could have a property of the model that is a union type property, and any related constraints become internal updater details rather than the responsibility of this new feature.

packages/core/src/model.ts
```diff
@@ -0,0 +1,35 @@
...
+    assigners: T
+  ) => {
+    context: TContext;
+    actions: {
+      [K in keyof T]: AssignAction<TContext, TEvent>;
+    };
+  };
+} {
+  return {
+    context: initialState,
```
Andarist, line 25: this is only ever be `initialContext`/`initialState`, so maybe it would be good to include that in the property name as right now one could think that the while model is impure and that it can be mutated/updates, whereas it's just a "base" for contexts of machines

packages/core/src/model.ts
```diff
@@ -0,0 +1,35 @@
...
+    withUpdaters: (assigners) => {
```
Andarist, line 26: q: why not a second parameter to the `createModel` itself?

packages/core/test/model.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      }
+    });
+
+    const machine = createMachine<typeof userModel['context'], UserEvent>({
+      context: userModel.context,
+      initial: 'active',
+      states: {
+        active: {
+          on: {
+            updateName: {
```
Andarist, line 31: There seems to be still quite a bit of a boilerplate to write using the current approach and from what I understand the main motivation for this (at least at the beginning) was lowering the barrier for newcomers. So the question is - does it deliver that?

packages/core/src/model.ts
```diff
@@ -0,0 +1,35 @@
...
+import { mapValues } from './utils';
+
+export function createModel<TContext, TEvent extends EventObject>(
+  initialState: TContext
+): {
+  context: TContext;
+  withUpdaters: <
+    T extends {
+      [key: string]:
+        | Assigner<TContext, TEvent>
```
Andarist, line 12: just a note - `TEvent` is a wide-open type, it has to accept all events accepted by the machine because we cant know where this is going to be used in the machine itself (and thus what actual events can be received here)

packages/core/src/model.ts
davidkpiano: Impossible to infer the actions in TypeScript.

packages/core/test/model.test.ts
davidkpiano: It will, but I want to save that for a follow-up PR.

packages/core/src/model.ts
davidkpiano: Yep, that's the intention.

packages/core/src/model.ts
```diff
@@ -0,0 +1,44 @@
+import { assign } from './actions';
+import { AssignAction, Assigner, EventObject, PropertyAssigner } from './types';
+import { mapValues } from './utils';
+
+export interface ContextModel<TC, TE extends EventObject> {
+  context: TC | (() => TC);
+  actions: {
+    [key: string]: Assigner<TC, TE> | PropertyAssigner<TC, TE>;
+  };
```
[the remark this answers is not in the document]
Andarist: I guess this approach might not be as convenient here as it is for `createMachine`. This is a much simpler factory function and explicit generics look more ordinary. I still don't think though that `createModel` should return `ContextModel` - it's much more like `IncompleteContextModel` or something and it's only the `.withAssigners` that actually returns a `ContextModel`.

packages/core/src/model.ts
```diff
@@ -0,0 +1,44 @@
...
+        ? AssignAction<TC, E>
+        : T[K] extends PropertyAssigner<TC, infer E>
+        ? AssignAction<TC, E>
+        : never;
+    };
+  };
+}
+
+export function createModel<TContext, TEvent extends EventObject>(
+  initialState: TContext | (() => TContext)
```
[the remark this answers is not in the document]
davidkpiano: Lazy initialization is definitely useful when you only want something to be created:
- per machine (one instance per invoked machine)
- only when the machine is created.

But also, this is future-proofing for v5, where `spawn` will be provided _internally_ (and properly typed) rather than externally, so you can imagine an API like this:

```js
initialContext: ({ spawn }) => {
  return {
    someActorRef: spawn(/* ... */)
  }
}
```

packages/core/src/model.ts
```diff
@@ -0,0 +1,44 @@
+import { assign } from './actions';
+import { AssignAction, Assigner, EventObject, PropertyAssigner } from './types';
+import { mapValues } from './utils';
+
+export interface ContextModel<TC, TE extends EventObject> {
```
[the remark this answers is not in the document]
davidkpiano: Which one do you prefer? I like the `TC` shorthand.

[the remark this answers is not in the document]
Andarist: I think I prefer the more verbose ones - but I don't feel super strongly about it.

packages/core/src/types.ts
```diff
@@ -155,9 +155,25 @@ export type TransitionTargets<TContext> = Array<
...
-  cond?: Condition<TContext, TEvent>;
-  actions?: Actions<TContext, TEvent>;
+export type RestrictedActions<
+  TContext,
+  TEvent extends EventObject,
+  TEventType extends TEvent['type']
+> = SingleOrArray<
+  | ActionType
+  | AssignAction<TContext, TEvent & { type: TEventType }>
+  | AssignAction<TContext, TEvent>
```
Andarist, line 165: Shouldn't this part be this?
```suggestion
  | AssignAction<TContext, TEvent & { type: TEventType }>
```

Also - shouldn't `ActionObject` and `ActionFunction` be restricted here as well?

packages/core/src/types.ts
```diff
@@ -936,14 +951,18 @@ export type Assigner<TContext, TEvent extends EventObject> = (
   meta: AssignMeta<TContext, TEvent>
 ) => Partial<TContext>;
 
+export type PartialAssigner<
```
Andarist, line 954: I kinda feel like it would be more appropriate to name this a `PropertyAssigner` and the other one a `PartialAssigner` 😅

packages/core/src/types.ts
```diff
@@ -155,9 +155,25 @@ export type TransitionTargets<TContext> = Array<
   string | StateNode<TContext, any>
 >;
 
-export interface TransitionConfig<TContext, TEvent extends EventObject> {
-  cond?: Condition<TContext, TEvent>;
-  actions?: Actions<TContext, TEvent>;
+export type RestrictedActions<
```
Andarist, line 158: is this used anywhere? cant find any reference to this 🤔

packages/core/test/assign.test.ts
```diff
@@ -224,6 +231,33 @@ describe('assign', () => {
       maybe: 'defined'
     });
   });
+
+  it('can assign from event', () => {
```
Andarist, line 235: q: is this here just to validate types?

packages/core/src/model.ts
```diff
@@ -0,0 +1,40 @@
...
+  AssignAction,
+  Assigner,
+  PropertyAssigner,
+  assign
+} from '.';
+import { EventObject } from './types';
+
+export interface ContextModel<TC, TE extends EventObject> {
+  context: TC;
+  actions: {
```
Andarist, line 12: is this property used anywhere?

Andarist, line 12: is this property used anywhere?

Andarist, line 12: is this property used anywhere?

packages/core/src/model.ts
```diff
@@ -0,0 +1,40 @@
...
+    [key: string]: ActionObject<TC, TE>;
+  };
+  assign: <TEventType extends TE['type'] = TE['type']>(
+    assigner:
+      | Assigner<TC, TE & { type: TEventType }>
```
Andarist, line 17: using `&` for extracting a concrete event type is quite horrible for the readability of the IDE hints etc, take a look:
Screenshot 2020-12-21 at 01 13 28

I think the better results would be achieved by using `Extract` or the introduced in this PR `ExtractEvent`

packages/core/src/model.ts
```diff
@@ -0,0 +1,40 @@
...
+  assign
+} from '.';
+import { EventObject } from './types';
+
+export interface ContextModel<TC, TE extends EventObject> {
+  context: TC;
+  actions: {
+    [key: string]: ActionObject<TC, TE>;
+  };
+  assign: <TEventType extends TE['type'] = TE['type']>(
```
Andarist, line 15: Just adding this comment for completeness.

I was experimenting with some overloads:
```diff
diff --git a/packages/core/src/model.ts b/packages/core/src/model.ts
index 71684848..ab2272a8 100644
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -7,17 +7,24 @@ import {
 } from '.';
 import { EventObject } from './types';
 
+interface ModelAssign<TC, TE extends EventObject> {
+  <TEventType extends TE['type'] = TE['type']>(
+    assigner:
+      | Assigner<TC, Extract<TE, { type: TEventType }>>
+      | PropertyAssigner<TC, Extract<TE, { type: TEventType }>>,
+    eventType?: TEventType
+  ): AssignAction<TC, Extract<TE, { type: TEventType }>>;
+  <SubEvent extends TE = TE>(
+    assigner: Assigner<TC, SubEvent> | PropertyAssigner<TC, SubEvent>
+  ): AssignAction<TC, SubEvent>;
+}
+
 export interface ContextModel<TC, TE extends EventObject> {
   context: TC;
   actions: {
     [key: string]: ActionObject<TC, TE>;
   };
-  assign: <TEventType extends TE['type'] = TE['type']>(
-    assigner:
-      | Assigner<TC, TE & { type: TEventType }>
-      | PropertyAssigner<TC, TE & { type: TEventType }>,
-    eventType?: TEventType
-  ) => AssignAction<TC, TE & { type: TEventType }>;
+  assign: ModelAssign<TC, TE>;
 }
```

but at the end of the day, I haven't really found out this to be useful. The only thing that this has allowed was this:
```ts
userModel.assign<UpdateNameEvent>({
  name: (ctx, ev) => ev.type
});
```
as event this:
```ts
userModel.assign({
  name: (ctx, ev: UpdateNameEvent) => ev.type
});
```
is matching the first overload (what is in this PR).

One additional observation is also that this is possible and this feels more natural:
```ts
userModel.assign<'updateName'>({
  name: (ctx, ev) => ev.type
});
```
but this doesn't provide any autocompletes like the version you have been using here:
```ts
userModel.assign({
  name: (ctx, ev) => ev.type
}, 'updateName');
```

Andarist (commented):
This version feels better than some previous iterations with `withAssigners` etc 👍 Good job.

packages/core/src/model.ts
davidkpiano: Renamed to `initialContext`

packages/core/src/model.ts
davidkpiano: Used `ExtractEvent` 👍

packages/core/src/model.ts
davidkpiano: Removed

packages/core/src/types.ts
davidkpiano: I know... we can change this for v5

packages/core/test/assign.test.ts
davidkpiano: Yep!

packages/core/src/model.ts
```diff
@@ -0,0 +1,38 @@
...
+  const model: ContextModel<TContext, TEvent> = {
+    initialContext,
+    assign,
+    reset: () => assign(initialContext)
+  };
+
+  return model;
+}
+
+export function assertEvent<
```
Andarist, line 33: This name is somewhat misleading because it leads the user to believe that this is an assertion function (`asserts event is TEvent`) but this is a type predicate (`event is TEvent`).

Besides that - could you add a test showcasing how this is supposed to be used?

packages/core/src/types.ts
```diff
@@ -366,16 +370,15 @@ export type TransitionConfigTarget<TContext, TEvent extends EventObject> =
...
+  | TransitionConfig<TContext, TEvent, TEventType>
 >;
 
-type TransitionsConfigMap<TContext, TEvent extends EventObject> = {
-  [K in TEvent['type']]?: TransitionConfigOrTarget<
-    TContext,
-    TEvent extends { type: K } ? TEvent : never
-  >;
+export type TransitionsConfigMap<TContext, TEvent extends EventObject> = {
+  [K in TEvent['type']]?: TransitionConfigOrTarget<TContext, TEvent, K>;
```
Andarist, line 381: Any particular reason behind this change? 🤔 seems a little bit stranger to pass `TEvent` and `K` around and compute at the end rather than pass a computed `TEvent` from here

is it about `TransitionConfigTarget` including a `StateNode` target? wouldn't it be better to allow there `any` at the `TEvent` position as we should be able to target any state node in the given machine config? As this is going to be removed in v5 anyway (targeting with getters etc) I'm not sure if it's worth complicating the codebase for its sake

Andarist (approved):
Please just remember about adding the appropriate changeset so the changelog entry gets correct attribution to PR etc :)

packages/core/src/types.ts
davidkpiano: I think this is an accidental commit, thanks for catching.

davidkpiano: Oh, I just discovered one of the reasons for this change: the `raise()` event needs to be able to infer all possible events.

packages/core/src/model.ts
davidkpiano: Removed

packages/core/test/types.test.ts
```diff
@@ -241,25 +241,25 @@ describe('Raise events', () => {
           on: {
             DECIDE: [
               {
-                actions: raise<GreetingContext, { type: 'ALOHA' }>({
+                actions: raise({
                   type: 'ALOHA'
-                }),
+                }) as any /* TODO: FIX */,
```
Andarist, line 246: Hm, I've wanted to look into it and I've checked out your branch locally but after removing those casts and comments I couldn't see any TS errors. Could you recheck on your machine? That being said - inferred generics are off and I don't understand why the inferred type is assignable here to the expected one 😭

packages/core/src/types.ts
```diff
@@ -371,10 +371,10 @@ export type TransitionConfigOrTarget<
   TransitionConfigTarget<TContext, TEvent> | TransitionConfig<TContext, TEvent>
 >;
 
-type TransitionsConfigMap<TContext, TEvent extends EventObject> = {
+export type TransitionsConfigMap<TContext, TEvent extends EventObject> = {
   [K in TEvent['type']]?: TransitionConfigOrTarget<
     TContext,
-    TEvent extends { type: K } ? TEvent : never
+    TEvent & { type: K }
```
Andarist, line 377: The previous version results in more readable IDE tooltips:

**Current**
Screenshot 2021-01-15 at 14 43 02

**Previous**
Screenshot 2021-01-15 at 14 41 21

davidkpiano: Fixed

packages/core/test/types.test.ts
davidkpiano: Hm, I'll keep the comment in here for now.

PR #1442, opened 2020-09-08
fsm minor example reformatting

tomByrer:
Only making the indentation consistent.
TY


davidkpiano:
These are auto-formatted and the formatting will likely be lost upon the next revision to the docs, unfortunately.

PR #1443, opened 2020-09-08
[v5] Deprecate `in: ...`

davidkpiano:
For https://github.com/davidkpiano/xstate/projects/1#card-40695477

This PR deprecates the `in: ...` property for transitions. It is recommended to use `stateIn()` and `stateNotIn()` guard creators instead.


.changeset/flat-clouds-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': major
+---
+
+The `in: ...` property is deprecated for transitions. It is recommended to use `stateIn()` and `stateNotIn()` guard creators instead.
```
Andarist, line 5: I wouldnt call this a deprecatiom, but rather a complete replacement of the old API with the new one

johtso:
I don't think there is a `stateNotIn` guard, instead you need to use `not(stateIn(...))`. I guess it's just the change note that's wrong?

PR #1456, opened 2020-09-12
[v5] Higher-level guards

davidkpiano:
This PR aims to simplify the implementation details and TS types for guards, as well as adds the following higher-level guards:

- `and([someGuard, anotherGuard, ...])`
- `or([someGuard, anotherGuard, ...])`
- `not(someGuard)`
- Others?

These higher-level guards can be combined at any depth:

```js
on: {
  EVENT: {
    target: 'somewhere',
    guard: and(['stringGuard', or([{ type: 'anotherGuard' }, not(() => false)])])
  }
}
```

And parameterized guards can use these higher-level guards:

```js
on: {
  EVENT: { target: 'somewhere', guard: 'everythingValid' }
},

// ...

// machine options
guards: {
  thisValid: () => true,
  thatValid: not('thatInvalid'),
  thatInvalid: (context, event) => false,
  everythingValid: and(['thisValid', 'thatValid'])
}
```

It also will rename `cond` to `guard` in transitions, as this has been a common source of confusion.

**TODO:**

- [x] Add `and([...])` + tests
- [x] Add `or([...])` + tests
- [x] Add `not(...)` + tests
- ~Add other boolean guards (if applicable)~
- ~Enable "guard definitions" in `guards` instead of just predicates~ (not in this PR)
- [x] Rename `cond` to `guards`
- [x] Simplify types


johnyanarella:
Really like this idea. Captures more interesting visualizable intent, too!

Given the variadic signature, wondering if `every()`, `some()` (or `any()`), etc. might read better left to right? That said `not()` reads better than `none()` would.

johnyanarella:
After further thought, now I've convinced myself you had it right with `and` and `or`.

🤷

packages/core/src/stateUtils.ts
```diff
@@ -1628,9 +1670,10 @@ function resolveActionsAndContext<TContext, TEvent extends EventObject>(
           );
           break;
         case actionTypes.choose: {
-          const ChooseAction = actionObject as ChooseAction<TContext, TEvent>;
-          const matchedActions = ChooseAction.conds.find((condition) => {
-            const guard = toGuard(condition.cond, machine.options.guards);
+          const chooseAction = actionObject as ChooseAction<TContext, TEvent>;
+          const matchedActions = chooseAction.conds.find((condition) => {
```
Andarist, line 1674: Maybe worth renaming `.conds` to `.cases`?

packages/core/src/stateUtils.ts
```diff
@@ -1628,9 +1670,10 @@ function resolveActionsAndContext<TContext, TEvent extends EventObject>(
...
+            const guard =
+              condition.cond && toGuard(condition.cond, machine.options.guards);
```
Andarist, line 1676: and this `.cond` to `.guard`?

packages/core/src/guards.ts
```diff
@@ -18,18 +27,41 @@ export function stateIn<TContext, TEvent extends EventObject>(
...
+    type: DEFAULT_GUARD_TYPE,
+    op: 'not',
+    children: [toGuard(condition)]
+  };
+}
 
-      return !state.matches(stateValue);
-    }
+export function and<TContext, TEvent extends EventObject>(
+  ...conditions: Array<Condition<TContext, TEvent>>
```
Andarist, line 41: wondering if this just shouldn't be an array from the start? we already accept arrays in `choose` and i don't think we use variadic parameter lists at many places (if at any), given that we treat those as arrays anyway it makes sense to me to just require it to be an array and skip the conversion step on our side

davidkpiano: Or the other way around - perhaps `choose` should just use the arguments as the array? Still trying to decide; I'll compare to some prior art.

Andarist: the best reference that i know is https://ramdajs.com/docs/#cond , other than that i don't have super strong preference for one over another

Some things that come to my mind:
- variadic functions make `.length` unusable, this is a somewhat esoteric need so kinda whatever on that
- the transpiled code is often worse as spread often creates a for loop over `arguments`

davidkpiano: Using arrays instead 👍

packages/core/src/utils.ts
```diff
@@ -328,30 +324,36 @@ export function isString(value: any): value is string {
...
-  condition?: Condition<TContext, TEvent>,
-  guardMap?: Record<string, ConditionPredicate<TContext, TEvent>>
-): Guard<TContext, TEvent> | undefined {
-  if (!condition) {
-    return undefined;
-  }
-
-  if (isString(condition)) {
+  guardConfig: GuardConfig<TContext, TEvent>,
+  guardMap?: Record<string, GuardPredicate<TContext, TEvent>>
```
Andarist, line 328: Not a specific comment for this PR so not a blocker at all - but I find it confusing for a function like this (there are other similar ones - e.g. when creating action objects) to be used kinda twice. Once when creating action objects without the context of the machine (where `guardMap` is undefined) and the second time when we resolve the previously created guard object with `guardMap` available.

IMHO the overall algorithm would be simpler if this wouldn't be aware of `guardMap` at all and later only some function like `getGuardPredicate` would get the actual predicate from the `guardMap` based on the object's type and just call it with that object - instead of going through `toGuard` again.

packages/core/test/guards.test.ts
```diff
@@ -403,3 +408,399 @@ describe('guards - other', () => {
...
+        initial: 'a',
+        states: {
+          a: {
+            on: {
+              EVENT: {
+                target: 'b',
+                guard: {
+                  type: 'testGuard',
+                  name: 'any',
+                  children: [
```
Andarist, line 425: I'm not super convinced that having a custom predicate that is responsible for resolving its children is a good idea - if u need something like this you can already use `params`, shouldn't `children` just be reserved for combinators like `not`, `and` and `or`?

In other words - could u prepare a real-life example of such a thing used in practice?

packages/core/src/utils.ts
```diff
@@ -328,30 +324,36 @@ export function isString(value: any): value is string {
...
+): GuardDefinition<TContext, TEvent> {
+  if (isString(guardConfig)) {
     return {
-      type: DEFAULT_GUARD_TYPE,
-      name: condition,
-      predicate: guardMap ? guardMap[condition] : undefined
+      type: guardConfig,
+      predicate: guardMap ? guardMap[guardConfig] : undefined,
+      params: { type: guardConfig }
```
Andarist, line 334: why the `type` is repeated in `params`? 🤔 shouldn't the `type` be enough here and `params` could just be empty?

packages/core/src/types.ts
```diff
@@ -116,36 +116,71 @@ export type ExtractStateValue<
...
 }
 
-export type Guard<TContext, TEvent extends EventObject> =
-  | GuardPredicate<TContext, TEvent>
-  | (Record<string, any> & {
-      type: string;
-    });
+export interface AnyGuardObject {
+  type: string;
+  [key: string]: any;
```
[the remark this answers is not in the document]
davidkpiano: What's the better alternative?

cbmd:
@davidkpiano `guards` not included in `package.json` in this PR. Is that intentional?

packages/core/src/utils.ts
davidkpiano: Because `type` is part of that object. There's no easy way to get the type here:

```js
guards: {
  someType: (context, event, meta) => {
    // ... how do you get the type??
  }
}
```

It's useful to know it and not beneficial to remove it.

packages/core/test/guards.test.ts
davidkpiano: > if u need something like this you can already use `params`

Not exactly, since `params` are just params - they're not resolved. `children` are resolved.

The usefulness of this is so that custom guards that might need to reference other guards can be made in userland (such as `xor(...)`, etc.).

packages/core/src/assert.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface GuardWithParams<T extends { [key: string]: any } = {}>
+  extends GuardDefinition<any, any> {
+  type: string;
+  params: {
+    type: string;
+  } & T;
+}
+
+export function assertGuard<T extends GuardDefinition<any, any>>(
```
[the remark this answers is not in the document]
davidkpiano: This was more of a stop-gap until we can strongly type guards

packages/core/test/guards.test.ts
```diff
@@ -403,3 +408,399 @@ describe('guards - other', () => {
...
+                    {
+                      type: DEFAULT_GUARD_TYPE,
+                      name: 'defaultGuard',
+                      predicate: () => true
+                    },
+                    { type: 'customGuard' }
+                  ],
+                  predicate: (_, __, { guard }) => {
+                    expect(guard.children).toHaveLength(2);
```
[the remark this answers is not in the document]
davidkpiano: Fixed with `expect.assertions(3);`

packages/core/src/utils.ts
```diff
@@ -328,30 +324,36 @@ export function isString(value: any): value is string {
...
-  if (isFunction(condition)) {
+  if (isFunction(guardConfig)) {
     return {
-      type: DEFAULT_GUARD_TYPE,
-      name: condition.name,
-      predicate: condition
+      type: guardConfig.name,
+      predicate: guardConfig,
+      params: {
+        type: guardConfig.name,
```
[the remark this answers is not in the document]
davidkpiano: We don't need both (which is why it's only `type`) - the `.name` is for the `function.name` here (scroll up - `guardConfig` is a function in this case)

packages/core/src/utils.ts
davidkpiano: You're right... let's think about how to do this in another PR for sure.

packages/core/src/assert.ts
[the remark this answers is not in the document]
davidkpiano: This has been removed

packages/core/test/actor.test.ts
```diff
@@ -361,7 +367,8 @@ describe('communicating with spawned actors', () => {
     parentService.start();
   });
 
-  it('should be able to communicate with arbitrary actors if sessionId is known', (done) => {
+  // TODO: This is an invalid use-case; consider removing
```
Andarist, line 370: I'm looking at just the title of this test:
> should be able to communicate with arbitrary actors if sessionId is known

why this would be considered invalid? Could u give more context?

Andarist (approved):

packages/core/test/actor.test.ts
davidkpiano: We talked about an interpreted machine creating an implicit "system". If we create two machines completely separately, we should create two implicit systems, instead of magically relying on some global "system" (aka registry of machines).

I can refactor this test so that the other machine is created _from_ the first machine. With the `ActorRef` refactor from master merged into this one, this test breaks though.

I don't think this is a use-case worth supporting; better to be explicit that actors can only communicate if they come from the same system.

PR #1460, opened 2020-09-16
chore: use tsdx to build @xstate/react

fubhy:
My original motivation for this pull request was to get `@xstate/react` to play nicely with Snowpack by just adding an ESM build for this particular package.

While doing so, however, I found that there is a wide range of different build setups for all the packages in this repository and was wondering if it might make sense to unify that setup in one sweeping change instead.

As `tsdx` is already used in one package and I personally also use this for my own libraries I went with that but am also open for other options.

Would you be open to accepting a pull request that reconfigures the entire repository more broadly to use tsdx throughout instead of typescript project reference builds. I'd also suggest to move the dependencies to each package individually as well as the jest runtime and then use lerna to invoke all these runtimes (build, test, lint, etc.) on a per-package basis. There are also mixtures of yarn and npm sprinkled here and there - Not sure which one you'd prefer but even with the latest improvements to npm my experience with monorepos is in favor of yarn).

Let me know if you think this would be out of scope for now or if I should proceed. For my own personal use, this pull request alone gets the job done.

@davidkpiano Btw. I don't know if you remember but we met at the first Agent Conf in Austria as speakers (I co-presented with Nik Graf) and went skiing together afterwards. We both ended up on the newbie squad and rolled down the mountain more than anything else :-D

### DIFF
--- a/packages/xstate-react/package.json
+++ b/packages/xstate-react/package.json
@@ -16,6 +16,7 @@
   "homepage": "https://github.com/davidkpiano/xstate/tree/master/packages/xstate-react#readme",
   "license": "MIT",
   "main": "lib/index.js",
+  "module": "es/index.js",
   "types": "lib/index.d.ts",
   "sideEffects": false,
   "directories": {
@@ -24,7 +25,9 @@
   },
   "files": [
     "lib/**/*.js",
-    "lib/**/*.d.ts"
+    "lib/**/*.d.ts",
+    "es/**/*.js",
+    "es/**/*.d.ts"
   ],
   "repository": {
     "type": "git",
@@ -32,7 +35,7 @@
   },
   "scripts": {
     "clean": "rm -rf dist lib tsconfig.tsbuildinfo",
-    "build": "tsc",
+    "build": "tsc && tsc --outDir es --module es2015",
     "test": "jest",
     "prepublish": "npm run build && npm run test"
   },


Andarist:
The monorepo is already configured with another build tool that is better suited for monorepos - https://github.com/preconstruct/preconstruct . It has only be done on the next branch though because introducing it to the master would technically be a breaking change.

fubhy:
Interesting. What are the benefits in comparison with tsdx? I am using tsdx hassle-free and, at least so far, didn't miss any features for monorepos.

If this is already solved in the `next` branch (unified build process and bundles), how would you like to move forward for for `master` / the current release for now if at all? For me personally a simple esm bundle for @xstate/react would get the job done. If you don't plan to include one in this release I am also happy to fork into our npm proxy for now.

Andarist:
Preconstruct:
- creates a funky, but very much working, redirecting files during development - this allows for respective **source files** to be always resolved when importing/requiring monorepoed-packages from other packages. This works **without** having to run an extra build watcher in the background, nor it requires any configuration for consuming packages (jest/node/webpack/rollup/whatever)
- it comes with the autofixer by default because it enforces certain conventions around file naming etc, so if you misconfigure something then it will both point it as being an error and very often offer fixing it automatically
- it validates some stuff - for example if we haven't forgot to put some packages that we have imported in our package.json
- it comes with the support for extra subentries for packages (so you can create things like `library/nested`)
- it also allows for easy creation for browser-specific entries - just add `"browser"` to your `package.json` (which will get fixed by preconstruct with correct paths) and start using `typeof window` / `typeof document` checks in your code to eliminate server-only code for your browser consumers
- probably more 😉 

> how would you like to move forward for for master / the current release for now if at all? 

In this case it seems that maybe the easiest way would be to invoke `tsc` with a different output dir and different target modules format?

fubhy:
Sounds good. I did actually have a need for extra subentries in the past (although arguably the need for this could be considered a bit of a code smell). I remember that I also had an issue with tsdx while trying to also compile a cli bin entry point alongside the other src. That was only possible with a separate build step. So if that's easier with preconstruct then I'll definitely give it a shot. Thanks for the intro.

> In this case it seems that maybe the easiest way would be to invoke tsc with a different output dir and different target modules format?

Yes, that was actually my first attempt until I realized that tsdx was already installed in the xstate-viz package and rollup was used elsewhere. Will update this PR to just add an additional tsc step tonight.

Andarist:
> I did actually have a need for extra subentries in the past (although arguably the need for this could be considered a bit of a code smell). 

Well, I would say - it depends 😅 It's certainly not something that one should reach very often but it's very useful for specific use cases. Like for example when you want to group a bunch of your exports - which we plan to do in v5 (we've already started).

> I remember that I also had an issue with tsdx while trying to also compile a cli bin entry point alongside the other src. That was only possible with a separate build step.

Changesets sort of do this: https://github.com/atlassian/changesets/blob/68fe90f5891e1c742dbac68ce45335a261513181/packages/cli/package.json#L18-L23 . The relation is just reversed there - it's a bin-centric package with an additional entry point (if entries share code then the output will appropriately share the code as well, thanks to Rollup, whereas with multiple builds this is impossible/much much harder to do). Leaving it just as a point of reference for you.

fubhy:
@Andarist Sorry, completely forgot that this was my weekend prolonged weekend off. Family time!

I simply added an additional tsc build for now as discussed

Andarist:
Thanks!

fubhy:
Sweet. Would you mind tagging a new release for @xstate/react? :)

davidkpiano:
> Sweet. Would you mind tagging a new release for @xstate/react? :)

Soon! I'd like this to go in for the next RC release as well: https://github.com/davidkpiano/xstate/pull/1472

PR #1462, opened 2020-09-16
Update Ember.js recipe

LevelbossMike:
This updates the Ember.js recipe to mention [ember-statecharts](https://ember-statecharts.com). The addon has existed for multiple years now and sees quite a lot of usage in the Ember.js ecosystem when looking at the stats from [emberobserver](https://emberobserver.com/addons/ember-statecharts) - it's in the 10% most downloaded Ember.js addons.

Using the addon also makes it much easier to use XState with Ember.js in general and writing your own integration layer will just be more work than installing the addon - especially if you want to come up with a nice api.

ember-statecharts mostly follows what xstate/react and xstate/vue are doing - i.e. it provides a `useMachine`-api that will create an interpreter for you that you can then use directly in your application code.

### DIFF
--- a/docs/recipes/ember.md
+++ b/docs/recipes/ember.md
@@ -1,17 +1,9 @@
 # Usage with Ember
 
-Ember follows a similar pattern to [Vue](./vue.md):
+The most straightforward way of using XState with Ember.js is through the [ember-statecharts](https://ember-statecharts.com)-addon.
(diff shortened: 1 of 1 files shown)


davidkpiano (approved):

davidkpiano:
LGTM, thanks for this!

PR #1472, opened 2020-09-21
[@xstate/react] Add lazy machine initialization, simplify useActor

davidkpiano:
This PR allows the `machine` passed into `useMachine(machine)` can now be passed in lazily:

```js
const [state, send] = useMachine(() => createMachine(/* ... */));

// ...
```

This has the benefit of avoiding unnecessary machine initializations whenever the component rerenders, without having to use `useMemo(...)`.


packages/xstate-react/src/useMachine.ts
```diff
@@ -113,6 +114,10 @@ export function useMachine<
   Interpreter<TContext, any, TEvent, TTypestate>['send'],
   Interpreter<TContext, any, TEvent, TTypestate>
 ] {
+  const machine = useMemo(() => {
+    return typeof getMachine === 'function' ? getMachine() : getMachine;
```
Andarist, line 118: Remember that `useMemo` doesn't have any semantic guarantees about the lifetime of its cache. This should be resolved within `useConstant`'s callback.

packages/xstate-react/src/useMachine.ts
```diff
@@ -113,6 +114,10 @@ export function useMachine<
...
+  }, []);
+
   if (process.env.NODE_ENV !== 'production') {
```
Andarist, line 121: Given that we won't have `machine` resolved at this point in time we should change this condition to:
```suggestion
  if (process.env.NODE_ENV !== 'production' && typeof getMachine !== 'function') {
```

packages/xstate-react/src/useActor.ts
```diff
@@ -1,69 +1,44 @@
...
-  return actorLike;
-}
+import { useState, useEffect, useRef, useCallback } from 'react';
+import { Sender, ActorRefLike } from './types';
+import { EventObject, Actor } from 'xstate';
 
 export function useActor<TEvent extends EventObject, TEmitted = any>(
-  actorLike: ActorRefLike<TEvent, TEmitted> | Actor
+  actorRef: ActorRefLike<TEvent, TEmitted> | Actor,
+  getSnapshot: (actor: typeof actorRef) => TEmitted = (a) =>
```
Andarist, line 7: This is a new thing - probably should be listed in the changelog.

packages/xstate-react/src/useActor.ts
```diff
@@ -1,69 +1,44 @@
...
+    'state' in a ? a.state : (undefined as any)
 ): [TEmitted, Sender<TEvent>] {
-  const actor = useMemo(() => resolveActor(actorLike), [actorLike]);
-  const deferredEventsRef = useRef<EventObject[]>([]);
-  const [current, setCurrent] = useState(actor.current);
+  // const actor = useMemo(() => resolveActor(actorLike), [actorLike]);
+  const deferredEventsRef = useRef<TEvent[]>([]);
+  const [current, setCurrent] = useState(() => getSnapshot(actorRef));
```
Andarist, line 12: in theory, we allow for changing `actorRef` between renders but this `current` won't sync to it so one would see stale values rendered until a new actor emits a value

davidkpiano: Good call.

davidkpiano:
@Andarist CI passes but CodeSandbox CI does not 😢

Andarist:
I've started looking into it but haven't found a quick win just yet - gonna get back to it tomorrow.

packages/xstate-react/test/useService.test.tsx
```diff
@@ -54,55 +61,6 @@ describe('useService hook', () => {
...
 
-  it('service actions should be configurable', () => {
-    const counterService = interpret(counterMachine).start();
-
-    const Counter = () => {
-      const [state, send] = useService(counterService);
-      const [otherState, setOtherState] = useState('');
-
-      useEffect(() => {
-        counterService.execute(state, {
```
Andarist, line 65: This test was IMHO incorrect - using this pattern there is a reasonably high chance that this action would **not** be executed. The only way to truly be sure that an action is going to be executed is to put it inline or in an `actions` object.

packages/xstate-react/CHANGELOG.md
```diff
@@ -2,12 +2,31 @@
...
+
+- The `machine` passed into `useMachine(machine)` can now be passed in lazily:
+
+  ```js
+  const [state, send] = useMachine(() => createMachine(/* ... */));
+
+  // ...
+  ```
+
+  This has the benefit of avoiding unnecessary machine initializations whenever the component rerenders, without having to use `useMemo(...)`.
```
Andarist, line 15: IMHO mentioning useMemo here is not good because `useMemo` shouldn't be used for this. We only care about the very first machine received by us - all others are just ignored. `useMemo` could only help with avoiding [this warning](https://github.com/davidkpiano/xstate/blob/d8f765586b82bf191727d463b6b05019716d0b55/packages/xstate-react/src/useMachine.ts#L120-L123) but even then it wouldn't avoid it entirely as `useMemo` doesn't hold a semantic guarantee. Therefore `useMemo` was never a good tool for the job here.

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -582,6 +583,7 @@ describe('useMachine (strict mode)', () => {
       // This will send an event to the parent service
       // BEFORE the service is ready.
       React.useLayoutEffect(() => {
+        debugger;
```
Andarist, line 586:
```suggestion
```

packages/xstate-react/src/useActor.ts
```diff
@@ -1,69 +1,45 @@
...
       }
     },
-    [actor]
+    [actorRef]
   );
 
   useEffect(() => {
-    const subscription = actor.subscribe(setCurrent);
+    setCurrent(getSnapshot(actorRef));
+    const subscription = actorRef.subscribe(setCurrent);
```
Andarist, line 30: shouldnt this be
```suggestion
    const subscription = actorRef.subscribe(() => setCurrent(getSnapshot(actorRef)));
```
?

packages/xstate-react/test/useMachine.test.tsx
Andarist: this could have been me 😅

packages/xstate-react/src/useActor.ts
davidkpiano: No it shouldn't; `getSnapshot` provides the last known snapshot of the last emitted value, while `subscribe` is the _actual_ latest emitted value.

In other words, `.subscribe(value => ...)` should be the ultimate source of truth. `getSnapshot(...)` is useful for the initial state.

Andarist: oh, ok - thought it was about selecting a slice of the emitted values, hence the confusion

👍

Andarist (approved):

packages/xstate-react/src/useActor.ts
davidkpiano: Selecting a slice is for the next PR 😉

PR #1483, opened 2020-09-25
Add utility types for typestates

sbking:
Three utility types to simplify working with typestates:

`CombineTypestates`: Takes a union of typestate objects and returns a new union where each context value has any missing/undefined values from other typestate contexts automatically filled in. Now each individual state only needs to define context fields that are actually relevant to that state. In my opinion, this makes it much easier at a glance to tell what the context value will actually look like. See the simplified example in [typescript.md](https://github.com/davidkpiano/xstate/compare/master...sbking:typestate-utils#diff-6890812557a98693a8ac6553e77fc608R177-R184).

`TypestateContext`: Given a union of typestates, returns a merged object type that represents the aggregate context value with all fields from all union members. Since this can now be inferred from a `CombineTypestates` result, the `createMachine` function should only need two generic parameters instead of three, but I tried to keep this changeset additive-only.

`CombineTypestateTuples`: Like `CombineTypestates`, but takes a union of `[value, context]` tuples instead of a union of `{ value, context }` objects.

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -158,15 +158,15 @@ The typestates of a machine are specified as the 3rd generic type in `createMach
 **Example:**
 
 ```ts
-import { createMachine, interpret } from 'xstate';
+import {
+  CombineTypestates,
+  TypestateContext,
+  createMachine,
+  interpret
+} from 'xstate';
 
 interface User {
   name: string;
-}
-
-interface UserContext {
-  user?: User;
-  error?: string;
 }
 
 type UserEvent =
@@ -174,26 +174,14 @@ type UserEvent =
   | { type: 'RESOLVE'; user: User }
   | { type: 'REJECT'; error: string };
 
-type UserState =
-  | {
-      value: 'idle';
-      context: UserContext & {
-        user: undefined;
-        error: undefined;
-      };
-    }
-  | {
-      value: 'loading';
-      context: UserContext;
-    }
-  | {
-      value: 'success';
-      context: UserContext & { user: User; error: undefined };
-    }
-  | {
-      value: 'failure';
-      context: UserContext & { user: undefined; error: string };
-    };
+type UserState = CombineTypestates<
+  | { value: 'idle'; context: {} }
+  | { value: 'loading'; context: { user?: User; error?: string } }
+  | { value: 'success'; context: { user: User } }
+  | { value: 'failure'; context: { error: string } }
+>;
+
+type UserContext = TypestateContext<UserState>;
 
 const userMachine = createMachine<UserContext, UserEvent, UserState>({
   id: 'user',
@@ -216,7 +204,7 @@ const userMachine = createMachine<UserContext, UserEvent, UserState>({
 
 const userService = interpret(userMachine);
 
-userService.subscribe(state => {
+userService.subscribe((state) => {
   if (state.matches('success')) {
     // from the UserState typestate, `user` will be defined
     state.context.user.name;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1052,6 +1052,43 @@ export interface Typestate<TContext> {
   context: TContext;
 }
 
+type KeyOfUnion<T> = T extends unknown ? keyof T : never;
+
+type MergeUnion<T> = {
+  [K in keyof T]: T[K];
+} &
+  {
+    [K in Exclude<KeyOfUnion<T>, keyof T>]?: T extends Record<K, any>
+      ? T[K] | undefined
+      : never;
+  };
+
+export type CombineTypestates<
+  TStates extends Typestate<any>
+> = TStates['context'] extends infer C
+  ? TStates extends Typestate<any>
+    ? {
+        value: TStates['value'];
+        context: {
+          [K in keyof (MergeUnion<C> &
+            TStates['context'])]: K extends keyof TStates['context']
+            ? TStates['context'][K]
+            : undefined;
+        };
+      }
+    : never
+  : never;
+
+export type CombineTypestateTuples<
+  TStates extends [StateValue, any]
+> = CombineTypestates<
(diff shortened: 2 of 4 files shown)


docs/guides/typescript.md
```diff
@@ -158,42 +158,30 @@ The typestates of a machine are specified as the 3rd generic type in `createMach
...
-    }
-  | {
-      value: 'success';
-      context: UserContext & { user: User; error: undefined };
-    }
-  | {
-      value: 'failure';
-      context: UserContext & { user: undefined; error: string };
-    };
+type UserState = CombineTypestates<
```
davidkpiano, line 177: Can you add a `/** typedoc */` explaining what this does (presumably combines `context` and `value` for each of the typestates)?

davidkpiano:
This will also work with non-string (object) `value`s, right?

sbking:
@davidkpiano Yes it does, I just added `{ parent: 'child' }` examples to the tests

davidkpiano (approved):

sbking:
@davidkpiano I was able to simplify `CombineTypestates` a bit by removing an unnecessary conditional type.

packages/core/src/types.ts
```diff
@@ -1052,6 +1052,43 @@ export interface Typestate<TContext> {
...
+          [K in keyof (MergeUnion<C> &
+            TStates['context'])]: K extends keyof TStates['context']
+            ? TStates['context'][K]
+            : undefined;
+        };
+      }
+    : never
+  : never;
+
+export type CombineTypestateTuples<
```
Andarist, line 1082: when one would like to use this in the real app? `CombineTypestates` seems to be sufficient, especially given that it works on data shape that we document

docs/guides/typescript.md
```diff
@@ -158,42 +158,30 @@ The typestates of a machine are specified as the 3rd generic type in `createMach
...
+  | { value: 'idle'; context: {} }
+  | { value: 'loading'; context: { user?: User; error?: string } }
+  | { value: 'success'; context: { user: User } }
+  | { value: 'failure'; context: { error: string } }
+>;
+
+type UserContext = TypestateContext<UserState>;
```
Andarist, line 184: how is this different from:
```js
type UserContext = TypestateContext<
  | { value: 'idle'; context: {} }
  | { value: 'loading'; context: { user?: User; error?: string } }
  | { value: 'success'; context: { user: User } }
  | { value: 'failure'; context: { error: string } }
>;
```

packages/core/src/types.ts
sbking: You're right :)

`CombineTypestateTuples` is the most terse way to define typestates, but I think `CombineTypestates` is more clear/obvious. I think this can be removed.

docs/guides/typescript.md
sbking: TypestateContext requires a union where each context type has every key present, including the ones that are always undefined for the state value. So it's is actually identical to:

```ts
type UserContext = TypestateContext<
  | { value: 'idle';    context: { user?: undefined; error?: undefined } }
  | { value: 'loading'; context: { user?: User;      error?: string } }
  | { value: 'success'; context: { user:  User;      error?: undefined } }
  | { value: 'failure'; context: { user?: undefined; error: string } }
>;
```

The ideal API in my mind (maybe for v5?) would require only two generic params and no helpers:

```
type UserEvent =
  | { type: 'FETCH'; id: string }
  | { type: 'RESOLVE'; user: User }
  | { type: 'REJECT'; error: string };

type UserState =
  | { value: 'idle'; context: {} }
  | { value: 'loading'; context: { user?: User; error?: string } }
  | { value: 'success'; context: { user: User } }
  | { value: 'failure'; context: { error: string } };

const userMachine = Machine<UserState, UserEvent>({
      /* ... */
});
```

Andarist: > TypestateContext requires a union where each context type has every key present, including the ones that are always undefined for the state value. So it's is actually identical to:

How useful `CombineTypestates` is then? The main benefit I'm seeing (but I, of course, might be missing something) is the ability to merge/combine typestates to create a "typestate context". If that's the main appeal - why `TypestateContext` doesn't just "call" `CombineTypestates` internally (so it could be used like on the example I've given)?

 > The ideal API in my mind (maybe for v5?) would require only two generic params, and could use TypeScript 4.1 string template types

The idea of having this as an object with paths as the keys is appealing. Not sure if you have meant that - but I don't think `TContext` should be completely gone. It's by far easier to use than typestates and actually provides type safety, whereas using typestates can lead to runtime bugs because we can't guarantee that the declared types match the reality.

sbking: Agreed, I think it should be possible to use the API with either a simple context type or a complex typestate union.

> It's by far easier to use than typestates and actually provides type safety, whereas using typestates can lead to runtime bugs because we can't guarantee that the declared types match the reality.

Maybe this can be examined for a future major version. I believe it should be possible to define the types of all typestates, events, and transitions so that an invalid state will not compile, but I'm not experienced enough with xstate yet to tell what API changes would need to be made.

sbking: > How useful CombineTypestates is then? The main benefit I'm seeing (but I, of course, might be missing something) is the ability to merge/combine typestates to create a "typestate context". If that's the main appeal - why TypestateContext doesn't just "call" CombineTypestates internally (so it could be used like on the example I've given)?

It absolutely could, but there would still need to be some utility type to extract the typestate union from the context type so it could be passed as the third generic param to `createMachine`, to avoid altering the public API of `createMachine`.

My thinking here was that, in a future version, if the state schema could be strongly typed in a way that no transition allows an invalid context for a given state value, then the typestates would be more relevant for TypeScript users than the aggregate context shape.

davidkpiano:
Closing as typestates are being deprecated (unsafe/unsound in v4), and we will try to provide a more robust solution in v5.

PR #1493, opened 2020-09-29
[@xstate/react] Migration docs + clarify useService -> useActor error

davidkpiano:


packages/xstate-react/src/useService.ts
```diff
@@ -6,11 +6,17 @@ import { ActorRef } from './types';
 export function fromService<TContext, TEvent extends EventObject>(
   service: Interpreter<TContext, any, TEvent>
 ): ActorRef<TEvent, State<TContext, TEvent>> {
-  const { machine } = service;
+  if (!('machine' in service)) {
```
Andarist, line 9:
```suggestion
  if (process.env.NODE_ENV !== 'production' && !('machine' in service)) {
```

Andarist (approved):

PR #1504, opened 2020-09-30
Fixed issue with `useService` returning an initial state for services in their final states

### LINKED ISSUE #1503: @xstate/react state is wrong in useService when the machine stops or reaches a final state
I use `useService` to get the state of an interpreted machine. All works perfectly, but when the machine reaches one of its final states and stops, `useService` returns the initial state of the machine, and not the final state. In fact, when the interpreter stops, it sets `this.initialized = false`.

I would expect the state to be the last state the stopped machine had.

https://github.com/davidkpiano/xstate/blob/37d5ed8606833f900a87f2a7de8f3f5059cf4deb/packages/xstate-react/src/useService.ts#L20

https://github.com/davidkpiano/xstate/blob/37d5ed8606833f900a87f2a7de8f3f5059cf4deb/packages/core/src/interpreter.ts#L505

Andarist:
fixes #1503

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -151,7 +151,7 @@ export class Interpreter<
    * Whether the service is started.
    */
   public initialized = false;
-  private _status: InterpreterStatus = InterpreterStatus.NotStarted;
+  public _status: InterpreterStatus = InterpreterStatus.NotStarted;
 
   // Actor
   public parent?: Interpreter<any>;
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -92,7 +92,7 @@ interface SpawnOptions {
 
 const DEFAULT_SPAWN_OPTIONS = { sync: false, autoForward: false };
 
-enum InterpreterStatus {
+export enum InterpreterStatus {
   NotStarted,
   Running,
   Stopped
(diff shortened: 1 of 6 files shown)


packages/core/src/interpreter.ts
```diff
@@ -151,7 +151,7 @@ export class Interpreter<
    * Whether the service is started.
    */
   public initialized = false;
-  private _status: InterpreterStatus = InterpreterStatus.NotStarted;
+  public _status: InterpreterStatus = InterpreterStatus.NotStarted;
```
davidkpiano, line 154: Should we make this `status` instead then?

packages/xstate-react/src/useService.ts
```diff
@@ -17,7 +17,7 @@ export function fromService<TContext, TEvent extends EventObject>(
     send: service.send.bind(service),
     subscribe: service.subscribe.bind(service),
     stop: service.stop!,
-    current: service.initialized ? service.state : machine.initialState,
+    current: service._status !== 0 ? service.state : machine.initialState,
```
davidkpiano, line 20: Should probably import `InterpreterStatus` here

packages/core/src/interpreter.ts
Andarist: done

packages/xstate-react/src/useService.ts
Andarist: couldn't do this right now, because the older XState versions don't have such export - I've exported `InterpreterStatus` from the `xstate` package though and added a comment here that we should migrate to it whenever we decide to release a new major of `@xstate/react`

davidkpiano (approved):

PR #1512, opened 2020-10-01
used always stead of ''

alirezavalizade:
I was just playing around `xstate` with examples and saw a warning which said `''` deprecated and use `always` instead. I saw also in others, If it's useful then I can change them too.

Thanks


davidkpiano:
Thank you!

PR #1514, opened 2020-10-02
[@xstate/react] Add `filter` option and `useMachineSelector()` hook

### LINKED ISSUE #1491: Reactjs doing unnecessary renders
**Description**

As discussed on [spectrum chat](https://spectrum.chat/statecharts/general/reactjs-to-many-renders~abcab5b2-ab02-44c7-a754-a7c740c5f16c).
When sending events to the machine, even when those events don't resolve to a new state transition or context change, reactjs still re-renders components.

**Expected Result**

There should be no re-rendering of the components that use the machine in question.

**Actual Result**

Components are being re-rendered.

**Reproduction**
https://codesandbox.io/s/xstate-react-render-ssnc1

davidkpiano:
- New `useMachine(...)` option: `filter(state, prevState)`, which allows you to filter the states that are provided:

```js
const [state, send] = useMachine(someMachine, {
  filter: (state, prevState) => {
    // don't rerender if only actions changed
    return (
      state.value !== prevState.value || state.context !== prevState.context
    );
  }
});
```

- New hook: `useMachineSelector(...)`, which enables you to apply a `selector` function to get part of (and/or transform) the machine's `state`:

```js
const [fullName, send] = useMachineSelector(
  userMachine,
  (state) => `${state.context.firstName} ${state.context.lastName}`
);
```

Closes #1491 cc. @alexreardon

### DIFF
--- a/packages/xstate-react/src/index.ts
+++ b/packages/xstate-react/src/index.ts
@@ -1,3 +1,4 @@
 export { useMachine, asEffect, asLayoutEffect } from './useMachine';
 export { useService } from './useService';
 export { useActor } from './useActor';
+export { useMachineSelect } from './useMachineSelect';
--- a/packages/xstate-react/src/index.ts
+++ b/packages/xstate-react/src/index.ts
@@ -1,4 +1,4 @@
 export { useMachine, asEffect, asLayoutEffect } from './useMachine';
 export { useService } from './useService';
 export { useActor } from './useActor';
-export { useMachineSelect } from './useMachineSelect';
+export { useMachineSelector as useMachineSelect } from './useMachineSelector';
--- a/packages/xstate-react/src/useMachine.ts
+++ b/packages/xstate-react/src/useMachine.ts
@@ -90,7 +90,7 @@ function executeEffect<TContext, TEvent extends EventObject>(
 
 interface UseMachineOptions<TContext, TEvent extends EventObject> {
   /**
-   * If provided, will be merged with machine's `context`.
+   * If provided, will be _merged_ with machine's `context`.
    */
   context?: Partial<TContext>;
   /**
@@ -98,6 +98,11 @@ interface UseMachineOptions<TContext, TEvent extends EventObject> {
    * start at this state instead of its `initialState`.
    */
   state?: StateConfig<TContext, TEvent>;
+
+  filter?: (
+    state: State<TContext, TEvent>,
+    prevState: State<TContext, TEvent>
+  ) => boolean;
 }
 
 export function useMachine<
@@ -140,6 +145,7 @@ export function useMachine<
     services,
     delays,
     state: rehydratedState,
+    filter,
     ...interpreterOptions
   } = options;
 
@@ -194,7 +200,11 @@ export function useMachine<
           currentState.changed === undefined &&
           Object.keys(currentState.children).length;
 
-        if (currentState.changed || initialStateChanged) {
+        if (
+          (currentState.changed || initialStateChanged) &&
+          // Only change the state if it is not filtered out
+          (!filter || filter?.(currentState, state))
+        ) {
           setState(currentState);
         }
 
--- a/packages/xstate-react/src/useMachine.ts
+++ b/packages/xstate-react/src/useMachine.ts
@@ -88,7 +88,11 @@ function executeEffect<TContext, TEvent extends EventObject>(
   originalExec();
 }
 
-interface UseMachineOptions<TContext, TEvent extends EventObject> {
+export type UseMachineOptions<
+  TContext,
+  TEvent extends EventObject,
+  TTypestate extends Typestate<TContext>
+> = {
   /**
    * If provided, will be _merged_ with machine's `context`.
    */
@@ -100,10 +104,11 @@ interface UseMachineOptions<TContext, TEvent extends EventObject> {
   state?: StateConfig<TContext, TEvent>;
 
   filter?: (
-    state: State<TContext, TEvent>,
-    prevState: State<TContext, TEvent>
+    state: State<TContext, TEvent, any, TTypestate>,
+    prevState: State<TContext, TEvent, any, TTypestate>
   ) => boolean;
-}
+} & InterpreterOptions &
+  MachineOptions<TContext, TEvent>;
 
 export function useMachine<
   TContext,
@@ -111,9 +116,7 @@ export function useMachine<
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(
   getMachine: MaybeLazy<StateMachine<TContext, any, TEvent, TTypestate>>,
-  options: Partial<InterpreterOptions> &
-    Partial<UseMachineOptions<TContext, TEvent>> &
-    Partial<MachineOptions<TContext, TEvent>> = {}
+  options: Partial<UseMachineOptions<TContext, TEvent, TTypestate>> = {}
 ): [
   State<TContext, TEvent, any, TTypestate>,
   Interpreter<TContext, any, TEvent, TTypestate>['send'],
@@ -203,7 +206,7 @@ export function useMachine<
         if (
           (currentState.changed || initialStateChanged) &&
           // Only change the state if it is not filtered out
-          (!filter || filter?.(currentState, state))
+          (!filter || filter?.(currentState, currentState.history!))
         ) {
           setState(currentState);
         }
--- a/packages/xstate-react/src/useMachineSelector.ts
+++ b/packages/xstate-react/src/useMachineSelector.ts
@@ -8,7 +8,7 @@ import {
 import { MaybeLazy } from './types';
 import { useMachine, UseMachineOptions } from './useMachine';
 
-export function useMachineSelect<
+export function useMachineSelector<
   T,
   TContext,
   TEvent extends EventObject,
--- a/.changeset/rude-ghosts-bathe.md
+++ b/.changeset/rude-ghosts-bathe.md
@@ -18,7 +18,7 @@ const [state, send] = useMachine(someMachine, {
 - New hook: `useMachineSelector(...)`, which enables you to apply a `selector` function to get part of (and/or transform) the machine's `state`:
 
 ```js
-const [fullName, send] = useMachine(
+const [fullName, send] = useMachineSelector(
   userMachine,
   (state) => `${state.context.firstName} ${state.context.lastName}`
 );
--- a/.changeset/rude-ghosts-bathe.md
+++ b/.changeset/rude-ghosts-bathe.md
@@ -1,0 +1,25 @@
+---
+'@xstate/react': minor
+---
+
+- New `useMachine(...)` option: `filter(state, prevState)`, which allows you to filter the states that are provided:
+
+```js
+const [state, send] = useMachine(someMachine, {
+  filter: (state, prevState) => {
+    // don't rerender if only actions changed
+    return (
+      state.value !== prevState.value || state.context !== prevState.context
+    );
+  }
+});
+```
+
+- New hook: `useMachineSelector(...)`, which enables you to apply a `selector` function to get part of (and/or transform) the machine's `state`:
+
+```js
+const [fullName, send] = useMachine(
+  userMachine,
+  (state) => `${state.context.firstName} ${state.context.lastName}`
+);
+```
--- a/packages/xstate-react/src/useMachineSelect.ts
+++ b/packages/xstate-react/src/useMachineSelect.ts
@@ -1,0 +1,33 @@
+import {
+  EventObject,
+  StateMachine,
+  Typestate,
+  Interpreter,
+  State
+} from 'xstate';
+import { MaybeLazy } from './types';
+import { useMachine, UseMachineOptions } from './useMachine';
+
+export function useMachineSelect<
+  T,
+  TContext,
+  TEvent extends EventObject,
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
+>(
+  getMachine: MaybeLazy<StateMachine<TContext, any, TEvent, TTypestate>>,
+  selector: (state: State<TContext, TEvent, any, TTypestate>) => T,
+  options: Partial<UseMachineOptions<TContext, TEvent, TTypestate>> = {}
+): [
+  T,
+  Interpreter<TContext, any, TEvent, TTypestate>['send'],
+  Interpreter<TContext, any, TEvent, TTypestate>
+] {
+  const [state, send, service] = useMachine(getMachine, {
+    ...options,
+    filter: (currentState, prevState) => {
+      return selector(currentState) !== selector(prevState);
+    }
+  });
+
+  return [selector(state), send, service];
+}
--- a/packages/xstate-react/test/useMachine.test.tsx
+++ b/packages/xstate-react/test/useMachine.test.tsx
@@ -724,4 +724,52 @@ describe('useMachine (strict mode)', () => {
 
     render(<App />);
   });
+
+  it('options.filter should filter states to prevent rerenders', () => {
+    let actionExecuted = false;
+    let rerenders = 0;
+
+    const App = () => {
+      const [, send] = useMachine(
+        () =>
+          createMachine({
+            initial: 'idle',
+            states: {
+              idle: {
+                on: {
+                  EXEC: {
+                    actions: () => {
+                      actionExecuted = true;
+                    }
+                  }
+                }
+              }
+            }
+          }),
+        {
+          filter: (state, prevState) =>
+            state.value !== prevState.value ||
+            state.context !== prevState.context
+        }
+      );
+
+      rerenders++;
+
+      return <button data-testid="button" onClick={() => send('EXEC')} />;
+    };
+    const { getByTestId } = render(
+      <React.StrictMode>
+        <App />
+      </React.StrictMode>
+    );
+
+    const button = getByTestId('button');
+
+    const currentRerenders = rerenders;
+
+    fireEvent.click(button);
+
+    expect(actionExecuted).toBeTruthy();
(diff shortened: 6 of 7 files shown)


cevr:
Perhaps it would be useful to set a default filter that does what you've described in the first snippet?

It would follow the semantics of things like `React.memo` or `PureComponent`

davidkpiano:
> Perhaps it would be useful to set a default filter that does what you've described in the first snippet?
> 
> It would follow the semantics of things like `React.memo` or `PureComponent`

@cevr The default behavior of `useMachine` already acts like a "filter" - it will not rerender unless the `state` has changed (either via `context` or `value` or `actions`).

cevr:
@davidkpiano  So what does the first snippet imply? That the default behavior is that it will rerender if an action is called that leads to no changes to `context` or `value`?

If so, that's what I meant by adding a default filter so that we can prevent that

davidkpiano:
There shouldn't be a default filter for that. I realized one important use-case where rerendering with actions would be necessary: persisting state.

cevr:
@davidkpiano That sounds interesting! Could you elaborate?

My understanding is that calling an action that leads to no changes in `value` or `context` is basically calling `x => x`. The only case I could see where this is incorrect is if an action has effects that affect things other than `context`. Is this true?

davidkpiano:
> @davidkpiano That sounds interesting! Could you elaborate?
> 
> My understanding is that calling an action that leads to no changes in `value` or `context` is basically calling `x => x`. The only case I could see where this is incorrect is if an action has effects that affect things other than `context`. Is this true?

Yes. For example, if you have a state where e.g., `{ actions: [{ type: 'focus' }] }` would focus an input, and you persisted/restored that state, you would expect the input to be focused (as it was when you persisted the state).

However, if the behavior was that state with only actions executed did _not_ update `state` from the `useMachine()` hook, you would never have that action, and the input wouldn't be focused.

So the default behavior has to be as it exists today. No default filters.

cevr:
@davidkpiano Last question, will the first snippet that you posted for `filter` break transitions that only execute side effects (like the focus one you mentioned)?

davidkpiano:
> @davidkpiano Last question, will the first snippet that you posted for `filter` break transitions that only execute side effects (like the focus one you mentioned)?

Good question! No, they will not break those, because those effects are executed by the interpreted machine inside the hook regardless.

The only difference is that a filtered-out state is not _provided_ through `[state, send] = useMachine(...)`.

FMcIntosh:
Would `userMachineSelector` allow a custom equality function?

It would be great to be able to do something [similar to zustand](https://github.com/pmndrs/zustand#selecting-multiple-state-slices) where you can arrange multiple selections like so :
```
import shallow from 'zustand/shallow'

// Object pick, re-renders the component when either state.nuts or state.honey change
const { nuts, honey } = useStore(state => ({ nuts: state.nuts, honey: state.honey }), shallow)

// Array pick, re-renders the component when either state.nuts or state.honey change
const [nuts, honey] = useStore(state => [state.nuts, state.honey], shallow)

// Mapped picks, re-renders the component when state.treats changes in order, count or keys
const treats = useStore(state => Object.keys(state.treats), shallow)
```

packages/xstate-react/src/useMachine.ts
```diff
@@ -194,7 +203,11 @@ export function useMachine<
           currentState.changed === undefined &&
           Object.keys(currentState.children).length;
 
-        if (currentState.changed || initialStateChanged) {
+        if (
+          (currentState.changed || initialStateChanged) &&
+          // Only change the state if it is not filtered out
+          (!filter || filter?.(currentState, currentState.history!))
```
dminkovsky, line 209: Is TS really requiring `?.` after the `!filter` check?

cbmd: It can be set but might be with diff signature, like `undefined` or something.

alexreardon:
Wonderful! For reference, here is the helper I made which has been working super well:

(this one assumes there is only a single machine on the context - which is always present)

```ts

export function useSelector<State>(selector: (machineState: MachineStateType) => State) {
  const service = useRequiredContext(ServiceContext);

  const [state, setState] = useState<State>(() => {
    const initial = service.initialized ? service.state : service.initialState;
    return selector(initial);
  });

  const currentStateRef = useRef<State>(state);

  useEffect(() => {
    const { unsubscribe } = service.subscribe((newState: MachineStateType) => {
      const result: State = selector(newState);

      const isSame: boolean = isShallowEqual(result, currentStateRef.current);

      if (isSame) {
        return;
      }

      currentStateRef.current = result;
      setState(result);
    });

    return unsubscribe;
  }, [selector, service, setState]);

  return state;
}

export function useSend(): ServiceType['send'] {
  const service = useRequiredContext(ServiceContext);
  return service.send;
}
```

packages/xstate-react/src/useMachineSelector.ts
```diff
@@ -0,0 +1,33 @@
...
+  EventObject,
+  StateMachine,
+  Typestate,
+  Interpreter,
+  State
+} from 'xstate';
+import { MaybeLazy } from './types';
+import { useMachine, UseMachineOptions } from './useMachine';
+
+export function useMachineSelector<
```
Andarist, line 11: When I've seen this API I thought this would be used for subscribing to an already live machine. I'm wondering if maybe it wouldn't be preferred to bake-in `selector` functionality into `useMachine` instead of providing a dedicated hook for this.

One of the gripes I have with this one is that I don't find the `filter` that is being here in the implementation to be particularly useful - especially that it is enforced here rather than provided as a default. The problem with this approach is that this would most likely **always** lead to rerendering because I doubt that anyone would do a smart caching within `selector`'s implementation and the simple `!==` that is used would just always return `true`.

So given that this is a very small addition to the `useMachine` itself I would rather see this as part of the main hook.

Andarist (commented):
> Yes. For example, if you have a state where e.g., { actions: [{ type: 'focus' }] } would focus an input, and you persisted/restored that state, you would expect the input to be focused (as it was when you persisted the state).
> However, if the behavior was that state with only actions executed did not update state from the useMachine() hook, you would never have that action, and the input wouldn't be focused.
> So the default behavior has to be as it exists today. No default filters.

I don't believe that this statement premise is really delivered by "no default filters" approach. There are a couple of reasons for that:
1. the expected result is situational and thus it should be up for a developer to decide about this, I could imagine a situation where it would be undesirable to have those actions re-executed. Often firing something multiple times could lead to unwanted side-effects after all
2. Actions are conceptually treated as fire and forget by us, so why this rule doesn't apply for rehydration? If we rehydrate actions then they are not really forgotten. I feel like that if something needs to be persisted then it should be somehow be embedded into the state or into the process of persisting & rehydrating - otherwise it's easy to miss something unintentionally. Consider [this example](https://codesandbox.io/s/cocky-antonelli-nykxz?file=/src/index.js) - when persisting the state the application was "focused" but when we rehydrate the persisted state this doesn't get replayed. It, therefore, means that relying on `state.actions` being re-fiered is sort-of a slippery slope and not an universal solution.
3. If we want to persist a used machine then we can freely subscribe to the created service for the purpose of persisting rather than relying on smth like `React.useEffect(persist, [state])` and if we can do that we could freely ignore states that have only caused some actions (other than `assign`) to be executed and therefore optimize the rerendering by default. What I'm saying is - we don't have to rerender to make persisting possible.

packages/xstate-react/src/useMachine.ts
```diff
@@ -88,27 +88,35 @@ function executeEffect<TContext, TEvent extends EventObject>(
...
    * The state to rehydrate the machine to. The machine will
    * start at this state instead of its `initialState`.
    */
   state?: StateConfig<TContext, TEvent>;
-}
+
+  filter?: (
+    state: State<TContext, TEvent, any, TTypestate>,
+    prevState: State<TContext, TEvent, any, TTypestate>
+  ) => boolean;
```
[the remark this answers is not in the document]
davidkpiano: The `prevState` argument might not even be necessary, since that's available on `state.history` 🤔

packages/xstate-react/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export { useMachine, asEffect, asLayoutEffect } from './useMachine';
 export { useService } from './useService';
 export { useActor } from './useActor';
+export { useMachineSelector as useMachineSelect } from './useMachineSelector';
```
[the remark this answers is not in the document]
davidkpiano: I think this was an accidental renaming

packages/xstate-react/src/useMachineSelector.ts
davidkpiano: > So given that this is a very small addition to the `useMachine` itself I would rather see this as part of the main hook.

The problem is that the return signature is different:

```js
const [state, send] = useMachine(...);
const [partOfState, send] = useMachineSelector(...);
```

Are you saying that you're okay with `state`/`partOfState` being different for the same `useMachine(...)` hook? Not entirely sure how this can be reliably typed.

Andarist: > Are you saying that you're okay with state/partOfState being different for the same useMachine(...) hook? Not entirely sure how this can be reliably typed.

Ye, this sounds OK to me, and seems that it would simplify the API (fewer hooks that do almost the same thing). It indeed raises the complexity of the types though - thanks to conditional types it's doable:
[TS playground](https://www.typescriptlang.org/play?#code/MYGwhgzhAEDKAuZ4FMCyZgAsCWA7ZAPACoDCA9rigB7wA00RAogG7KXTI1sAmMA3tHgBPAA7IAXNAjwATngDm0AL4A+aHyUAoUJBgIkhUhWp0GLNvA5dcvdYNESpshcrUbNmvChkAzDMmgAVQg0DBx8AHkReGwKCGJySk5TJlZ2ZJ5+ezFJaTlcRVV1TWhS6GBjZIB+SSMkmhKykJBkYHga6AAKaQNJfRQEypp6VIsVAEpoAF41AFdcAGtcMgB3XE0tTW5W8BkAn3m22NxoWZD0LDxCRrqTWhvzdOtbAWEcp3zC+5VOxtKAWzCVz6iBQF3ChkSdzMaXgKnu40kAG1+pChilHnCALpbHZgPbQA64I4UU7nIH4Ag3KHJe6lUZPFA2LJvRx5FxKOnQKIxOJWJm2YKhS6RaLHeK3WkwsaaH5-aCAkWOVHgq6DeoY2Hw+VkMVxDo88WaRHQADkzVa8FN-My0AWyCEZB83L1uAg8qqXXm2x8V24NuZLt5bqR5uQLTappxZRj0E9KNBaI1I0xagAPtAAErIeCzGS4IgOAgAOQoxdmIHAACMWgRDXEkQAiC1tRtYlQqaOx0rI7O5-OFsQlssV6u1+sh5vhy1tjtY8by5Go9XQhnYjyaAD0m8EOBgmDA-pEMjINeQ-2gK2w8Ew0AABmdhRDOv96BpxnetzuJzAMoGBC28CSN0iaSGAuBCJMMzQHgPjIDIcDTm0yD+lo26xvGsBISg3A4uhMbAV+3ZlD+AYvFI2EdCBvTQOBkHTGosHwYhEY4coRHEXG0AothKHQBmy6SsM0qUJ2HHEUuiYrlKa5ifhZQLtocSWBeUzQPgKxwImqqUgIPhkGQki4LM-xVsxnJ2KykimsWjAABpENaqidIpFRupYSI9CgpY-li0xkk+VwvgubnSNxXnIH5amPjpyAvm+Sghcp4W8bh-kxRScWvsUMaAcBEVQWonQCBU-wiC0KAgEI1ncNgPhwXslBOQuiUeKFHkRREMhYaxKFRQFsXxTlTTYfliaFWaVZgAAXgoYCmhstH8PKgFURFIIGAQekGUZJlmQhFmvA41m2Q5TkTAxHwKBsC5AA)

This probably has still some room for improvement - it's just a quickly prepared demo.

davidkpiano: > > Are you saying that you're okay with state/partOfState being different for the same useMachine(...) hook? Not entirely sure how this can be reliably typed.
> 
> Ye, this sounds OK to me, and seems that it would simplify the API (fewer hooks that do almost the same thing). It indeed raises the complexity of the types though - thanks to conditional types it's doable:
> [TS playground](https://www.typescriptlang.org/play?#code/MYGwhgzhAEDKAuZ4FMCyZgAsCWA7ZAPACoDCA9rigB7wA00RAogG7KXTI1sAmMA3tHgBPAA7IAXNAjwATngDm0AL4A+aHyUAoUJBgIkhUhWp0GLNvA5dcvdYNESpshcrUbNmvChkAzDMmgAVQg0DBx8AHkReGwKCGJySk5TJlZ2ZJ5+ezFJaTlcRVV1TWhS6GBjZIB+SSMkmhKykJBkYHga6AAKaQNJfRQEypp6VIsVAEpoAF41AFdcAGtcMgB3XE0tTW5W8BkAn3m22NxoWZD0LDxCRrqTWhvzdOtbAWEcp3zC+5VOxtKAWzCVz6iBQF3ChkSdzMaXgKnu40kAG1+pChilHnCALpbHZgPbQA64I4UU7nIH4Ag3KHJe6lUZPFA2LJvRx5FxKOnQKIxOJWJm2YKhS6RaLHeK3WkwsaaH5-aCAkWOVHgq6DeoY2Hw+VkMVxDo88WaRHQADkzVa8FN-My0AWyCEZB83L1uAg8qqXXm2x8V24NuZLt5bqR5uQLTappxZRj0E9KNBaI1I0xagAPtAAErIeCzGS4IgOAgAOQoxdmIHAACMWgRDXEkQAiC1tRtYlQqaOx0rI7O5-OFsQlssV6u1+sh5vhy1tjtY8by5Go9XQhnYjyaAD0m8EOBgmDA-pEMjINeQ-2gK2w8Ew0AABmdhRDOv96BpxnetzuJzAMoGBC28CSN0iaSGAuBCJMMzQHgPjIDIcDTm0yD+lo26xvGsBISg3A4uhMbAV+3ZlD+AYvFI2EdCBvTQOBkHTGosHwYhEY4coRHEXG0AothKHQBmy6SsM0qUJ2HHEUuiYrlKa5ifhZQLtocSWBeUzQPgKxwImqqUgIPhkGQki4LM-xVsxnJ2KykimsWjAABpENaqidIpFRupYSI9CgpY-li0xkk+VwvgubnSNxXnIH5amPjpyAvm+Sghcp4W8bh-kxRScWvsUMaAcBEVQWonQCBU-wiC0KAgEI1ncNgPhwXslBOQuiUeKFHkRREMhYaxKFRQFsXxTlTTYfliaFWaVZgAAXgoYCmhstH8PKgFURFIIGAQekGUZJlmQhFmvA41m2Q5TkTAxHwKBsC5AA)
> 
> This probably has still some room for improvement - it's just a quickly prepared demo.

Thanks for this; good idea, I'll implement it as a single hook.

Andarist: I've used `=> any` for the `select` and infer the type later but it could also be generic of `UseMachineOptions`. The lack of partial inference is being a real PITA for us as this would be so handy in so many situations... The best would be to write down **how** we expect people to interact with those things and just try to cover those - by using conditional types and providing different overloads we should be able to cover all of the things.

I can do this if we could only write down what are type design goals - like: people should rely on the inference that we provide, etc. Is there anything particular that you'd like to avoid in the type-API that we provide?

packages/xstate-react/src/useMachine.ts
```diff
@@ -88,27 +88,35 @@ function executeEffect<TContext, TEvent extends EventObject>(
   originalExec();
 }
 
-interface UseMachineOptions<TContext, TEvent extends EventObject> {
+export type UseMachineOptions<
```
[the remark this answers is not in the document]
davidkpiano: Because it's a union with `InterpreterOptions` and `MachineOptions`

SimeonC:
Is there a plan to have a `useServiceSelector` as well? Using the same service in multiple places would cause more rerenders as the service could be used in multiple components on the same page.

For example, if we had a multi step wizard with a Title component only based on the topmost state value (ignoring nested and context) and child states control the logic for each “page”, updating the child states or the context shouldn’t cause the title component to re-render. (I know this example is a little bit off for state.value with hierarchical state)

```
function Title() {
  const [state] = useService(...);
  return <h1>{getTitleTranslation(state.value)}</h1>;
}
```

davidkpiano:
> Is there a plan to have a `useServiceSelector` as well? Using the same service in multiple places would cause more rerenders as the service could be used in multiple components on the same page.

I might actually abandon/redo this PR with a different approach, aimed for performance:

```js
function SomeComponent() {
  // does not rerender when state changes - that's up to you!
  useMachineCallback(someMachine, {/* ... */}, callback);

  // same with service
  useServiceCallback(someService, callback);

  // ...
}
```

What you would do in `callback` would be up to you. For example, filtering state:

```js
const [state, setState] = useState(someMachine.initialState);

useMachineCallback(someMachine, null, (currentState) => {
  if (somePredicate(currentState)) {
    setState(currentState);
  }
});
```

I believe this will be a much more flexible approach.

Andarist:
Aren't both possibly orthogonal features though?

davidkpiano:
> Aren't both possibly orthogonal features though?

Which ones, filter and selector? Not really...

```js
// filter
useMachineCallback(someMachine, null, (currentState) => {
  if (somePredicate(currentState)) {
    setState(currentState);
  }
});

// selector
useMachineCallback(someMachine, null, (currentState) => {
  setState(someSelector(currentState));
});
```

Andarist:
I've meant the `useMachine` with select/filter/whatever and `useMachineCallback`

davidkpiano:
Superseded by #1915

PR #1516, opened 2020-10-03
[@xstate/react] Stabilize `send` from useActor

davidkpiano:
This PR makes `send` from the `useActor(...)` hook stable.

Reference: https://spectrum.chat/statecharts/general/xstate-react-1-0-0-breaking-changes~5ce72da1-871f-4ea4-9f70-40680bf02027

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -199,6 +199,34 @@ export interface ActivityDefinition<TContext, TEvent extends EventObject>
 
 export type Sender<TEvent extends EventObject> = (event: Event<TEvent>) => void;
 
+type ExcludeType<A> = { [K in Exclude<keyof A, 'type'>]: A[K] };
+
+type ExtractExtraParameters<A, T> = A extends { type: T }
+  ? ExcludeType<A>
+  : never;
+
+type ExtractSimple<A> = A extends any
+  ? {} extends ExcludeType<A>
+    ? A
+    : never
+  : never;
+
+type NeverIfEmpty<T> = {} extends T ? never : T;
+
+export interface PayloadSender<TEvent extends EventObject> {
+  /**
+   * Send an event object or just the event type, if the event has no other payload
+   */
+  (event: TEvent | ExtractSimple<TEvent>['type']): void;
+  /**
+   * Send an event type and its payload
+   */
+  <K extends TEvent['type']>(
+    eventType: K,
+    payload: NeverIfEmpty<ExtractExtraParameters<TEvent, K>>
+  ): void;
+}
+
 export type Receiver<TEvent extends EventObject> = (
   listener: (event: TEvent) => void
 ) => void;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -198,6 +198,7 @@ export interface ActivityDefinition<TContext, TEvent extends EventObject>
 }
 
 export type Sender<TEvent extends EventObject> = (event: Event<TEvent>) => void;
+
 export type Receiver<TEvent extends EventObject> = (
   listener: (event: TEvent) => void
 ) => void;
--- a/packages/xstate-react/src/types.ts
+++ b/packages/xstate-react/src/types.ts
@@ -46,31 +46,3 @@ export interface ActorRefLike<TEvent extends EventObject, TEmitted = any>
 }
 
 export type MaybeLazy<T> = T | (() => T);
-
-type ExcludeType<A> = { [K in Exclude<keyof A, 'type'>]: A[K] };
-
-type ExtractExtraParameters<A, T> = A extends { type: T }
-  ? ExcludeType<A>
-  : never;
-
-type ExtractSimple<A> = A extends any
-  ? {} extends ExcludeType<A>
-    ? A
-    : never
-  : never;
-
-type NeverIfEmpty<T> = {} extends T ? never : T;
-
-export interface PayloadSender<TEvent extends EventObject> {
-  /**
-   * Send an event object or just the event type, if the event has no other payload
-   */
-  (event: TEvent | ExtractSimple<TEvent>['type']): void;
-  /**
-   * Send an event type and its payload
-   */
-  <K extends TEvent['type']>(
-    eventType: K,
-    payload: NeverIfEmpty<ExtractExtraParameters<TEvent, K>>
-  ): void;
-}
--- a/packages/xstate-react/src/types.ts
+++ b/packages/xstate-react/src/types.ts
@@ -47,7 +47,14 @@ export interface ActorRefLike<TEvent extends EventObject, TEmitted = any>
 
 export type MaybeLazy<T> = T | (() => T);
 
+type OnlyIfExact<A, B, C> = A extends B ? (B extends A ? C : never) : never;
 export interface PayloadSender<TEvent extends EventObject> {
-  (event: TEvent | TEvent['type']): void;
-  (event: TEvent['type'], payload: Omit<TEvent, 'type'>): void;
+  (event: TEvent): void;
+  <K extends TEvent['type']>(
+    event: K & OnlyIfExact<{ type: K }, Extract<TEvent, { type: K }>, K>
+  ): void;
+  <K extends TEvent['type']>(
+    event: K,
+    payload: Omit<Extract<TEvent, { type: K }>, 'type'>
+  ): void;
 }
--- a/packages/xstate-react/src/types.ts
+++ b/packages/xstate-react/src/types.ts
@@ -46,3 +46,8 @@ export interface ActorRefLike<TEvent extends EventObject, TEmitted = any>
 }
 
 export type MaybeLazy<T> = T | (() => T);
+
+export interface PayloadSender<TEvent extends EventObject> {
+  (event: TEvent | TEvent['type']): void;
+  (event: TEvent['type'], payload: Omit<TEvent, 'type'>): void;
+}
--- a/packages/xstate-react/src/types.ts
+++ b/packages/xstate-react/src/types.ts
@@ -47,14 +47,30 @@ export interface ActorRefLike<TEvent extends EventObject, TEmitted = any>
 
 export type MaybeLazy<T> = T | (() => T);
 
-type OnlyIfExact<A, B, C> = A extends B ? (B extends A ? C : never) : never;
+type ExcludeType<A> = { [K in Exclude<keyof A, 'type'>]: A[K] };
+
+type ExtractExtraParameters<A, T> = A extends { type: T }
+  ? ExcludeType<A>
+  : never;
+
+type ExtractSimple<A> = A extends any
+  ? {} extends ExcludeType<A>
+    ? A
+    : never
+  : never;
+
+type NeverIfEmpty<T> = {} extends T ? never : T;
+
 export interface PayloadSender<TEvent extends EventObject> {
-  (event: TEvent): void;
+  /**
+   * Send an event object or just the event type, if the event has no other payload
+   */
+  (event: TEvent | ExtractSimple<TEvent>['type']): void;
+  /**
+   * Send an event type and its payload
+   */
   <K extends TEvent['type']>(
-    event: K & OnlyIfExact<{ type: K }, Extract<TEvent, { type: K }>, K>
-  ): void;
-  <K extends TEvent['type']>(
-    event: K,
-    payload: Omit<Extract<TEvent, { type: K }>, 'type'>
+    eventType: K,
+    payload: NeverIfEmpty<ExtractExtraParameters<TEvent, K>>
   ): void;
 }
--- a/packages/xstate-react/src/types.ts
+++ b/packages/xstate-react/src/types.ts
@@ -46,3 +46,33 @@ export interface ActorRefLike<TEvent extends EventObject, TEmitted = any>
 }
 
 export type MaybeLazy<T> = T | (() => T);
+
+// TODO: remove these types (up to PayloadSender) when
+// @xstate/react depends on xstate v5.0 (use PayloadSender from core instead)
+type ExcludeType<A> = { [K in Exclude<keyof A, 'type'>]: A[K] };
+
+type ExtractExtraParameters<A, T> = A extends { type: T }
+  ? ExcludeType<A>
+  : never;
+
+type ExtractSimple<A> = A extends any
+  ? {} extends ExcludeType<A>
+    ? A
+    : never
+  : never;
+
+type NeverIfEmpty<T> = {} extends T ? never : T;
+
+export interface PayloadSender<TEvent extends EventObject> {
+  /**
+   * Send an event object or just the event type, if the event has no other payload
+   */
+  (event: TEvent | ExtractSimple<TEvent>['type']): void;
+  /**
+   * Send an event type and its payload
+   */
+  <K extends TEvent['type']>(
+    eventType: K,
+    payload: NeverIfEmpty<ExtractExtraParameters<TEvent, K>>
+  ): void;
+}
--- a/packages/xstate-react/src/useService.ts
+++ b/packages/xstate-react/src/useService.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useRef } from 'react';
+import { useMemo, useRef } from 'react';
 import {
   EventObject,
   State,
@@ -6,6 +6,7 @@ import {
   Typestate,
   PayloadSender
 } from 'xstate';
+import useIsomorphicLayoutEffect from 'use-isomorphic-layout-effect';
 import { useActor } from './useActor';
 import { ActorRef } from './types';
 import useConstant from './useConstant';
@@ -45,7 +46,7 @@ export function useService<
   // Using a ref ensures that the constant `sender` always sends the event
   // to the latest (possibly changed) `service.send` method.
   const senderRef = useRef(service.send);
-  useEffect(() => {
+  useIsomorphicLayoutEffect(() => {
     senderRef.current = service.send;
   }, [service.send]);
   const sender = useConstant(
--- a/packages/xstate-react/src/useService.ts
+++ b/packages/xstate-react/src/useService.ts
@@ -1,13 +1,7 @@
 import { useMemo } from 'react';
-import {
-  EventObject,
-  State,
-  Interpreter,
-  Typestate,
-  PayloadSender
-} from 'xstate';
+import { EventObject, State, Interpreter, Typestate } from 'xstate';
 import { useActor } from './useActor';
-import { ActorRef } from './types';
+import { ActorRef, PayloadSender } from './types';
 
 export function fromService<TContext, TEvent extends EventObject>(
   service: Interpreter<TContext, any, TEvent>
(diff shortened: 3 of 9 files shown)


packages/xstate-react/src/types.ts
```diff
@@ -46,3 +46,8 @@ export interface ActorRefLike<TEvent extends EventObject, TEmitted = any>
 }
 
 export type MaybeLazy<T> = T | (() => T);
+
+export interface PayloadSender<TEvent extends EventObject> {
+  (event: TEvent | TEvent['type']): void;
+  (event: TEvent['type'], payload: Omit<TEvent, 'type'>): void;
```
Andarist, line 52: This is not correct for a couple of reasons.

It now kinda accepts `any` on the second position because `Omit` is just a wrapper around a combination `Pick`, `Exclude`, and `keyof`. Problem is that `keyof` used on the union (and `TEvent` is usually one) only returns common keys and usually `type` is the only known key - so in the end what we get here is `Pick<TEvent, never>` (when this gets computed based on some passed-in concrete generic parameter). See this [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5cqEixEKAApwaAGwD2cACYBlLAugAeACoYsJCAA9IIBSzSYcBImAB8ZSsgAUEc2GbaXyAD7J3ugNoA5LQMAQC6AJTM6LLACrxUTi5uOuCBwRBhADTIdFJyiszYALbAYFopYNlB9BnWkcjRsbwC5OlmuqYAvHZUVOnMAQBi2NgB9lRwICwA7tDMIACuRbjQ-F49vf3IAQBCAIIASmO9yLJgABZzrOxc-IJqCNJwUCgIslMkLKrMkjLyyqoNBUWNZBF9jA5docAtlSMhcJNESwBpA2AFkHxwoI2gBpCA0WQwbAgACqIGA72Q3QA1vjCe1wCxBEA)

To overcome this problem we could tweak this to use a distributive version of the `Omit`, which is this:
```ts
type DistributiveOmit<T, K extends keyof T> = T extends any ? Omit<T, K> : never
```
A version with this would at least accept payloads limited to a particular set but it would allow incorrect combinations because there is no relation specified between both arguments. See this [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5crQbIAIsDYdcAVzDBM2ALbAwAHgAqAGmQBpZBAAekEABMWyANYQaAexjJ1APmQBeB-qNYzyOCBrIAfmQlFQ1tHWdmEAhMKEFQSFhEFAAFOBoAGxs4EwBlL2gNDCwSQ2NvYpwCIjBnCioAChiS5nVKkgAfB3aAbQByYQg+gF0ASmZ0G2ATXkbm8Fbegfoh4e06dKyc5nFJYBk5BWU1NsxwbWWGPsdx5Enp3gEhFbQzsHM3eqpqFeY+gDFsNg+pRvr4WAB3aBRaSKXDQfjILpfKiDP4AIQAggAlEHfZA2MAAC2hrHYXH4ghMRAycCgKAQNhAbFYXmYaUy2TyBSgqnaLEcggA9EKCTYbHQWNpQIyoPTiOQWF4Gn0sbjtKQfMyoVBmAAWABMyD4o0E5CAA)

To properly type this we'd have to introduce a generic for the second overload:
```ts
<K extends TEvent["type"]>(event: K, payload: Omit<Extract<TEvent, { type: K }>, 'type'>): void;
```
See this [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5cqEixEKAApwaAGwD2cACYBlLAugAeACoYsJCAA9IIBSzSYcBImAB8ZSsgAUEc2GbaXyAD7J3ugNoA5LQMAQC6AJTM6LLACrxU6gDSyAZGJj464H4ARMEQ2aHWTi7MiQA0yHRScorM2AC2wGDqqIZQiM2+4BWk1PRMyMl81hVB-QHWkcjRsbwC5HlmuqYAvHZUVHnMAQBi2NgB9lRwICwA7tDMIACu9bjQ-F7rG1vIAQBCAIIASocbyLIwAALS6sdhcfiCNQIaRwKAoBCyU4kFiqZiSGTyZSqDSZMAsayCVHGBwfH4BHoA4GggKKALIPjhcjEhSkr6-Sknc6ggAsACYGUyWWzyZTXmSOVSQVBtnTBeQgA).

Is supporting this worth the complexity? Personally, I don't quite think it is. The gains for this API are IMHO super minimal (I actually struggle to name any technical benefit for this) - from my POV we should stick to a single API to avoid bikeshedding in teams & type complexity.

PS. The comment of mine here doesn't even touch on the fact that the first overload is broken as well because it allows silly mistakes like on this [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5cqEixEKAApwaAGwD2cACYBlLAugAeACoYsJCAA9IIBSzSYcBImAB8ZSsgAUEc2GbaXyAD7J3ugNoA5LQMAQC6AJTM6LLACrwC5MEoOuCmALx2VFRJzAEAYtjYAfZUcCAsAO7QzCAArgC2uND8XplZOcgBAEIAggBKxVnIsmAAFtWs7Fz8gmoI0nBQKAiy5SQsqsySMvLKqhopYCzWggD0p8OydCzkG8YO3f0B4eRAA). It's possible to type this correctly but it would add even more to the overall type complexity.

davidkpiano: > Is supporting this worth the complexity? Personally, I don't quite think it is.

Yes, it is. There are many developers using the two-argument variant, and I want to continue to support that.

davidkpiano: For reference, here's the strongly-typed version: [TS Playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5crQZoAHojAAeAIIAaZACF5AYQB8yALzJpyCKMggAJi0XIA-Kb0Hj288mXJmICJiiPkz171CRYiFAAKcDQANgD2cIYAyliG0JIAKhhYJFaxJsk4BERg6hRUABQuKcxJmOAAlMzoYcCGvFSSANK6+unIZSkA2gDkwhA9ALqqReVgzC0AZGISkqTU9EzILXzyqPpQs53g8vP9E8h8qvJNqlXINXUNyM2t1ibbYL39QyPF4BPydMHhkczYAFtgFJ1uwtpkwLsFgwDkd5H1Fj11OdLvV+IJ+mgxiYtPkqNClj0AGLYbA9Sj4uAgFgAd2gTgArgDcNB+MgAD5kClUfbIHoKaQAJXJ+OQYTAAAt6ax2Fw2Zy8bz+dIAFrkgTkOIIEJwKAoBBhakkFixZhBUIRaKxeIQliqQQmowFZXCirIAD07uQeoAjgzgHqTN8LZEAITkR2GZ0CtUVCOxaNCnpQ8VSqDMHqRHqHOORxPCqFU2nSgAsACYc-GnS7k2QCRmBQWxZLpZnDNm+HGgA)

Andarist: I've played a little bit with implementations from [the article](https://artsy.github.io/blog/2018/11/21/conditional-types-in-typescript/) that we've been talking about and this is what I got:
[TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5crQZoAHojAAeAIIAaZACF5AYQB8yALzJpyCKMggAJi0XIA-Kb0Hj288mXJmICJiiPkz17yH0UqUQgANgCuhhAAKr4y6lqkyADaANLIoGJBoRCSANYQNAD2MNryAETCEMWqALrM0kmVyAI+Iv7sEi1QcAAKcB0AthCQUCwy8uExtlZYNnFlzOENlHb+6WGRDNGLTi7QgmVircQAysC9dIGZ0uM6k0YmcCA0ixakfLr6UybLIatRl4tUFmk-3cnh2VC2rkEoEG8CQyG6NECeTghkOU2gknCGCwJBuNmxOAIRDA6goVAAFNtwHMCWAAJTMdB5YCGXhUAD07OQkmSeJMWMw4HiAHIysLKqpKYKwMxkgAyMQSSQzXyyhrydpKgU4+QqhhqviqeSJVQM5BMlls5BSnHMTVHE5nTLa8CqeKlXzFSpmi2sxac7m8963ZAusAisUSm3U5CJeR0OCI5GGZjYXrAKT2qRh3XUVWxhpG5Ci3zC9Q+5l+qg8t7Wfm0iOlqNUmWx+OJpEou36DrETXdPoDaDDHOx8uMyu8Rp7WkmWLA2bFgBi2GwwuB9xYAHdoE5gr1cDtXgAfMgL-PChTSABK66oVDyYAAFrvWOwuPxkKeyXn9cWrwAWuujRhEEPQoAgeQgGwrBTMwCKdqi6JQJIs6qIILBTOSl7SEBuZwHAzDCgAjMKDR0hhWE4bedLIAGUAQAAjsEwAMSYCZJiiACE5CYUY2GAcKFF8YYAk3sKuaPi+UBESiZF8MJVFXre+HQTuMnIAALAATORvFKeJuaLtREnIFJr7CnJelAA)

In addition - there is this extra requirement~ that for an event without extra params the second param should not be allowed at all. By excluding "simple" events we can easily get it:
[TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5crQZoAHojAAeAIIAaZACF5AYQB8yALzJpyCKMggAJi0XIA-Kb0Hj288mXJmICJiiPkz17yH0UqUQgANgCuhhAAKr4y6lqkyADaANLIoGJBoRCSANYQNAD2MNryAETCEMWqALrM0kmVyAI+Iv7sEi1QcAAKcB0AthCQUCwy8uExtlZYNnFlzOENlHb+6WGRDNGLTi7QgmVircQAysC9dIGZ0uM6k0YmcCA0ixakfLr6UybLIatRl4tUFmk-3cnh2VC2rkEoEG8CQyG6NECeTghkOU2gknCGCwJBuNmxOAIRDA6goVAAFNtwHMCWAAJTMdB5YCGXhUAD07OQkmSeJMWMw4HiAHIysLKqpKYKwMxkgAyMQSSQzXyyhrydpKgU4+QqhhqviqeSJVQM5BMlls5BSnHMTVHE5nTLa8CqeKlXzFSpmi2sxac7m8963ZAusAisUSm3U5CJeR0OCI5GGZjYXrAKT2qRh3XUVWxhpG5Ci3zC9Q+5l+qg8t7WT4Bb7O2kR0uVDX6DoO07nTG0t0ehhe1SSqky2PxxNIlF2jttWfdPoDaDDHOx8uMyu8Rp7WkmWLA2bFgBi2GwwuB9xYAHdoE5gr1cDtXgAfMgH-PChTSABK56oVDyMAAAtb1YdguH4ZBXzJPN9WLL8AC1z0aMIgh6FAEDyEA2FYKZmARKdUXRKBJF3VRBBYKZyU-aQkNzOA4GYYUAEZhQaOkKKomjfzpZAAygCAAEdgmAASTATJMUQAQnISijGoxDhQ4uTDAUn9hVzQCQKgJiUTYvhlK4r9f3o7Cbx05AABYACZ2Nkoz1NzQ9uI05AtNA4U9LsoA)
but I have also found a better way to do this - which also has a better IDE displays when you make an error:
[TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5crQZoAHojAAeAIIAaZACF5AYQB8yALzJpyCKMggAJi0XIA-Kb0Hj288mXJmICJiiPkz17yH0UqUQgANgCuhhAAKr4y6lqkyADaANLIoGJBoRCSANYQNAD2MNryAETCEMWqALrM0kmVyAI+Iv7sEi1QcAAKcB0AthCQUCwy8uExtlZYNnFlzOENlHb+6WGRDNGLTi7QgmVircQAysC9dIGZ0uM6k0YmcCA0ixakfLr6UybLIatRl4tUFmk-3cnh2VC2rl2vmQADltlAAJIwVCnWiSMaaMivG42eYWUFuOaCUCDeBIZDdGiBPJwQyHKbQdEYLAkHGfTA4AhEMDqChUAAU23Ac2Z4AAlMx0HlgIZeFQAPTy5CSZJs5DhUVgeIAcjK2sqqkFHLAzGSADIxBJJDNfKaGvJ2laNcb5DaGHa+Kp5IlVBLkFKZXLkEaWcxHUcTmdMs6Wap4qVfMVKn6A7LForlar3rd1ZqdXqDSHhchEvI6HAqTTDMxsL1gFJw1IY+BXdRbSWGl7kLrfNr1CnpWmqCq3tYTM2tT2GPrDUKTSWyxXqbTmHDXEiUXQ0Y3Hd0+gNoMMJ97VP3JYPeI09pqTLFgbNuwAxbDYbXA+4sADu0CcwV6uB2V4AB8yHvdttQUaQACU3yoKg8jAAALH9WHYLh+GQEC+Tbd1u0ggAtN9GjCIIehQBA8hANhWCmZhKWXOkGSgSQb1UQQWCmfkIOkQjWzgOBmG1ABGbUGjFdjOO4mCxWQDMoAgABHYJgHkkxy0rWkAEJyA4owuII7VxN0wx9Og7VWwQ5CoEE2lRL4IzJMgmC+Ko79rOQAAWAAmMSdMcszWwfKTzOQSyUO1WzfKAA)

davidkpiano: @Andarist Updated the types: 62e6b90

davidkpiano:
@Andarist is this one good to go?

packages/xstate-react/test/useActor.test.tsx
```diff
@@ -278,4 +278,64 @@ describe('useActor', () => {
     fireEvent.click(button);
     expect(div.textContent).toEqual('100');
   });
+
+  it('send() should be stable', (done) => {
```
Andarist, line 282: a similar test should be added for the `useService`

packages/xstate-react/test/useActor.test.tsx
```diff
@@ -278,4 +278,64 @@ describe('useActor', () => {
...
+    const button = getByTestId('button');
+    fireEvent.click(button);
+
+    // At this point, `send` refers to the last actor
+
+    jest.advanceTimersByTime(20);
+
+    // The effect will call the closed-in `send`, which originally
+    // was the reference to the first actor. Now that `send` is stable,
+    // it will always refer to the latest actor.
```
Andarist, line 339: I'm not super sure that this is the most correct logic - I mean, this is usually not how React works. I'm fine with it right now, but we might revisit this in the future maybe. 

I believe that the real issue reported on Spectrum was not about `send` not being stable when swapping actors but rather because currently, a single actor doesn't have a stable `send` which is related to the logic around deferred actors and this:
https://github.com/davidkpiano/xstate/blob/0165825e7e08407c9c5c8ab069fd5ffdd160242c/packages/xstate-react/src/useActor.ts#L20-L24

So while this PR solves that by just always providing the same `send` from the given `useActor` it kinda feels like a band-aid - fixing the problem a step further than it should be fixed. If we should have a stable `send` for a given `useActor` feels like a separate discussion to the original problem from Spectrum.

packages/xstate-react/src/types.ts
```diff
@@ -46,3 +46,31 @@ export interface ActorRefLike<TEvent extends EventObject, TEmitted = any>
...
+    : never
+  : never;
+
+type NeverIfEmpty<T> = {} extends T ? never : T;
+
+export interface PayloadSender<TEvent extends EventObject> {
+  /**
+   * Send an event object or just the event type, if the event has no other payload
+   */
+  (event: TEvent | ExtractSimple<TEvent>['type']): void;
```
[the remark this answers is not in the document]
davidkpiano: So what should be done here? I'm a bit confused.

packages/xstate-react/src/useService.ts
```diff
@@ -32,11 +33,26 @@ export function useService<
...
   const serviceActor = useMemo(() => fromService(service), [service]);
 
-  return useActor<TEvent, State<TContext, TEvent, any, TTypestate>>(
+  // Using a ref ensures that the constant `sender` always sends the event
+  // to the latest (possibly changed) `service.send` method.
+  const senderRef = useRef(service.send);
+  useEffect(() => {
+    senderRef.current = service.send;
+  }, [service.send]);
+  const sender = useConstant(
```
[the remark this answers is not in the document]
davidkpiano: I removed this

packages/xstate-react/src/types.ts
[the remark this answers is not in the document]
Andarist: I just meant that there is a non-zero chance that this:
```js
export interface PayloadSender<TEvent extends EventObject> {
  (event: TEvent): void;
  (event: ExtractSimple<TEvent>['type']): void;
}
```
is better than 
```ts
export interface PayloadSender<TEvent extends EventObject> {
  (event: TEvent | ExtractSimple<TEvent>['type']): void;
}
```

So the recommendation would be to just split this definition into more overloads (you currently have 2 and we can split the mentioned case into 2 to get the overall 3). So the final version would look like this:
```ts
export interface PayloadSender<TEvent extends EventObject> {
  (event: TEvent): void;
  (event: ExtractSimple<TEvent>['type']): void;
  <K extends TEvent['type']>(
    eventType: K,
    payload: NeverIfEmpty<ExtractExtraParameters<TEvent, K>>
  ): void;
}
```

I think so based on the intuition, those example diagnostics that I have pasted in and Andrew Casey (TS team) often mentioning that "type instantiation explosion" is a perf cliff so it seems to be better to usually favor less instantiation over other metrics from the diagnostics. 

There is also a non-zero chance that I have no idea what I'm talking about 😅 And probably the true answer is as usual: "it depends".

packages/core/src/types.ts
```diff
@@ -198,6 +198,35 @@ export interface ActivityDefinition<TContext, TEvent extends EventObject>
...
+
+type ExtractSimple<A> = A extends any
+  ? {} extends ExcludeType<A>
+    ? A
+    : never
+  : never;
+
+type NeverIfEmpty<T> = {} extends T ? never : T;
+
+export interface PayloadSender<TEvent extends EventObject> {
```
Andarist, line 216: Keep in mind that **moving** this type into the core results in a problem for `@xstate/react` which starts to depend on this but its peer dependency range on `xstate` allows older XState versions that do not export this type.

The correct way to do this would be to:
- keep the type duplicated in both packages for now
- or create the `@xstate/types` dependency that would be specified as a regular dependency for both packages and as the result the problem would go away once and for all

Andarist (approved):

packages/xstate-react/src/types.ts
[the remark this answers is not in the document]
davidkpiano: Combining them was an automated suggestion from TSLint:

```
This overload and the one on line 70 can be combined into one signature taking `TEvent | ExtractSimple<TEvent>['type']`. (unified-signatures)tslint(1)
```

PR #1525, opened 2020-10-05
Use on transition hook

lanmonster:
This PR adds a hook that allows you to pass a function to be run on each state transition.

example usage:

```jsx
const MyComponent = () => {
  const [state, send, service] = useMachine(someMachine);
  useOnTransition((state, event) => {
    console.log({state, event});
  }, service, { ignoreInitEvent: true });

  return (
    <div>
      {/* ... */}
    </div>
  );
}

```

### DIFF
--- a/packages/xstate-react/src/useOnTransition.ts
+++ b/packages/xstate-react/src/useOnTransition.ts
@@ -1,0 +1,25 @@
+import { useEffect } from 'react';
+import { EventObject, Typestate } from 'xstate';
+import { Interpreter, StateListener } from 'xstate/lib/interpreter';
+
+interface UseOnTransitionOptions {
+  ignoreInitEvent: boolean;
+}
+export function useOnTransition<
+  TContext,
+  TEvent extends EventObject,
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
+>(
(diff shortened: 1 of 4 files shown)


davidkpiano:
This looks useful, but also something that might better live in user-land, no?

Andarist:
Right, this is so easy to implement that we shouldn't have to provide a thing like this. Different people might have different opinions about possible options, timing of the setup (layout or passive effect), etc that it doesn't seem that useful to make those opinions baked-in.

packages/xstate-react/src/useOnTransition.ts
```diff
@@ -0,0 +1,25 @@
...
+  TContext,
+  TEvent extends EventObject,
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
+>(
+  listener: StateListener<TContext, TEvent, any, TTypestate>,
+  service: Interpreter<TContext, any, TEvent, TTypestate>,
+  options?: UseOnTransitionOptions
+) {
+  useEffect(() => {
+    service.onTransition((state, event) => {
```
Andarist, line 18: sideways observation - you forgot to implement a proper cleanup here

PR #1535, opened 2020-10-07
docs(xstate-test): add note about nested states

UberMouse:
This tripped me up before and caused me to spend a few hours debugging @xstate/test and my tests to determine why it seemed like tests were being run for the wrong state.

Turns out that @xstate/test executes not just the `meta.test` for the current state, but also all the parent states of the current state. Seems like a good thing to note in the docs as it's not immediately obvious (but makes sense now in retrospect).

---
This change is [Reviewable](https://reviewable.io/reviews/davidkpiano/xstate/1535)

### DIFF
--- a/packages/xstate-test/README.md
+++ b/packages/xstate-test/README.md
@@ -222,4 +222,4 @@ Executes each step in `testPath.segments` by:
 
 And finally, verifying that the SUT is in the target `testPath.state`.
 
-NOTE: If your model has nested states, the `meta.test` method for each parent state is also executed when veryifing the SUT is in the target `testPath.state`
+NOTE: If your model has nested states, the `meta.test` method for each parent state of that nested state is also executed when verifying that the SUT is in that nested state.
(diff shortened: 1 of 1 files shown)


davidkpiano:
Thanks! Can you do this in `packages/xstate-test/README.md` instead?

UberMouse:
Have changed the README instead of the docs page now.

packages/xstate-test/README.md
```diff
@@ -221,3 +221,5 @@ Executes each step in `testPath.segments` by:
 2. Executing the event for `segment.event`
 
 And finally, verifying that the SUT is in the target `testPath.state`.
+
+NOTE: If your model has nested states, the `meta.test` method for each parent state is also executed when veryifing the SUT is in the target `testPath.state`
```
davidkpiano, line 225:
```suggestion
NOTE: If your model has nested states, the `meta.test` method for each parent state of that nested state is also executed when verifying that the SUT is in that nested state.
```

davidkpiano (changes requested):
One quick change, and then it should be good to go!

UberMouse:
Sorry @davidkpiano, missed the notification. I've just accepted your change.

PR #1542, opened 2020-10-09
Added State and Transition constants to React example

gkiely:
- Added State and Transition constants so they can be exported and consumed.

I'm coming at this from a newcomers perspective so please excuse me if I have misunderstood the concepts, I am still reading through the docs.

Reading this page: https://xstate.js.org/docs/recipes/react.html#hooks
It seemed that re-writing the example to use state and transitions to constants would be a better pattern.

### DIFF
--- a/docs/recipes/react.md
+++ b/docs/recipes/react.md
@@ -5,16 +5,27 @@ The most straightforward way of using XState with React is through local compone
 ```js
 import { Machine } from 'xstate';
 
+// State
+export const toggleState = {
+  ACTIVE: 'active',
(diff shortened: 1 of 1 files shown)


davidkpiano:
Thanks for the PR; this would be better as a separate section. For getting started quickly, the additional boilerplate provides noise for little gain.

For maintainable projects, it's a good practice to do what you've done, though!

gkiely:
Thought that might be the case, feel free to close this, I can set up another PR/section in the future if it makes sense after I finish reading the docs.
Or if you have another section in mind happy to move it to that.

PR #1553, opened 2020-10-13
Ensure spawned actors are present in state.children. Fixes #795

### LINKED ISSUE #795: service.children empty when machine starts
**Description**
I was looking for a way to use a "main" machine to handle the whole workflow of my React application (each screen of the app has its own machine, invoked by the main one).

My `<App />` component just:
- looks at the current state of the main machine
- retrieves the corresponding invoked machine using `service.children` Map
- renders the proper screen with the invoked machine passed as props

Not sure this is the best solution (inputs are welcome!), but I came across something rather unexpected:

**Expected Result**
`service.children` should return current actors/invoked services.

**Actual Result**
`service.children` returns an empty Map on start, even if a service is invoked and runs properly.

**Reproduction**
Run the following sample and take a look at the console:
[CodeSandbox](https://codesandbox.io/s/xstate-test-7mn7y)

Starting by a transient state doesn't change anything, but delaying the first state seems to work (really ugly though).
You can test this by opening `machines/main.js` and changing `initial` state from `step1` to:
- `step0WithADelay`
- `step0WithATransientState`

**Additional context**
XState 4.6.7

davidkpiano:
This PR ensures that all actors (spawned/invoked) are represented in `state.children`.

### DIFF
--- a/.changeset/funny-spiders-push.md
+++ b/.changeset/funny-spiders-push.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The `state.children` property now properly shows all spawned and invoked actors. See #795 for more details.
--- a/.changeset/funny-spiders-push.md
+++ b/.changeset/funny-spiders-push.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-The `state.children` property now properly shows all spawned and invoked actors. See #795 for more details.
+The `state.children` property now properly shows all spawned and invoked actors. See [#795](https://github.com/davidkpiano/xstate/issues/795) for more details.
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -899,16 +899,20 @@ export class Interpreter<
     return undefined;
   }
 
+  private removeChild(childId: string): void {
+    this.children.delete(childId);
+    this.forwardTo.delete(childId);
+
+    delete this.state.children[childId];
+  }
+
   private stopChild(childId: string): void {
     const child = this.children.get(childId);
     if (!child) {
       return;
     }
 
-    this.children.delete(childId);
-    this.forwardTo.delete(childId);
-
-    delete this.state.children[childId];
+    this.removeChild(childId);
 
     if (isFunction(child.stop)) {
       child.stop();
@@ -972,7 +976,7 @@ export class Interpreter<
 
     childService
       .onDone((doneEvent) => {
-        this.stopChild(childService.id);
+        this.removeChild(childService.id);
         this.send(toSCXMLEvent(doneEvent as any, { origin: childService.id }));
       })
       .start();
@@ -985,7 +989,7 @@ export class Interpreter<
     promise.then(
       (response) => {
         if (!canceled) {
-          this.stopChild(id);
+          this.removeChild(id);
           this.send(
             toSCXMLEvent(doneInvoke(id, response) as any, { origin: id })
           );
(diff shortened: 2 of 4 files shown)


packages/core/src/interpreter.ts
```diff
@@ -844,17 +849,14 @@ export class Interpreter<
             : serviceCreator;
 
           if (isPromiseLike(source)) {
-            this.state.children[id] = this.spawnPromise(
```
Andarist, line 847: q: so those were setting the child on the not-yet-updated `this.state`?

packages/core/src/interpreter.ts
```diff
@@ -973,7 +972,7 @@ export class Interpreter<
 
     childService
       .onDone((doneEvent) => {
-        this.removeChild(childService.id);
+        this.stopChild(childService.id);
```
Andarist, line 975: in a sense, this will now send a "stop" event to the already stopped child, which doesn't sound ideal, even if a stopped child should just ignore all newly received events

packages/core/src/interpreter.ts
davidkpiano: I don't remember the specifics, but these did appear on the updated `this.state`

packages/core/src/interpreter.ts
davidkpiano: If a child is removed, it should be stopped, though.

davidkpiano: If we remove a child without sending a "stop" signal, the child will be orphaned, which is a memory leak.

Andarist: this is true for other scenarios - yes, but in here we receive a signal from the child that it has stopped, so there is no need to send a "stop" signal to it, the child is already stopped - it has just notified us about this

packages/core/src/interpreter.ts
Andarist: ok, so I've jumped into the code to understand this a little bit better and this is what I've learns - `spawn` is not handled by this part of the code at all, this is processing of `start` actions (so basically for starting invokes / activities) and that's why this has not been working. This is IMHO a very strong candidate for the refactor in v5 because invoke/spawn should IMHO behave in the very same way. The only difference should be that invokes are static and bound to state whereas spawned actors are dynamic.

However, right now they differ because invokes are more lazy and spawns are eager. For example if an invoke gets "immediately" cancelled within a single macrotask then it is not even started (according to the SCXML spec `statesToInvoke` are only processed after the `internalQueue` gets empty), but `spawn` does not behave in this way as it's eager (handled right away, rather than deferred to the end of the current macrotask).

davidkpiano: > but `spawn` does not behave in this way as it's eager (handled right away, rather than deferred to the end of the current macrotask).

This behavior is fixed in the `next` branch already. Spawned actors will behave exactly the same way.

Andarist: That's great to hear 🚀

Andarist (approved):

.changeset/funny-spiders-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The `state.children` property now properly shows all spawned and invoked actors. See #795 for more details.
```
Andarist, line 5:
```suggestion
The `state.children` property now properly shows all spawned and invoked actors. See [#795](https://github.com/davidkpiano/xstate/issues/795) for more details.
```

PR #1559, opened 2020-10-15
Ensure children are removed when stopped. Fixes #1457

### LINKED ISSUE #1457: Calling stop on a child actor does not notify its parent that it should be cleaned
This was originally discussed as a question here ( https://github.com/davidkpiano/xstate/discussions/1455 ) as I thought the problem was on my used of the library. 

This issue was created for tracking purposes ( https://github.com/davidkpiano/xstate/discussions/1455#discussioncomment-77361 )

davidkpiano:
Depends on #795

### DIFF
--- a/packages/core/test/interpreter.test.ts
+++ b/packages/core/test/interpreter.test.ts
@@ -1937,5 +1937,60 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
         })
         .start();
     });
+
+    it('stopped spawned actors should be cleaned up in parent', (done) => {
+      const childMachine = Machine({
+        initial: 'idle',
+        states: {
+          idle: {}
+        }
+      });
+
+      const parentMachine = createMachine<any>({
+        id: 'form',
+        initial: 'present',
+        context: {},
+        entry: assign({
+          machineRef: () => spawn(childMachine, 'machineChild'),
+          promiseRef: () =>
+            spawn(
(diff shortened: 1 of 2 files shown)


packages/core/test/interpreter.test.ts
```diff
@@ -1937,5 +1937,60 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
...
+          }
+        }
+      });
+
+      const service = interpret(parentMachine)
+        .onDone(() => {
+          expect(service.children.get('machineChild')).toBeUndefined();
+          expect(service.children.get('promiseChild')).toBeUndefined();
+          expect(service.children.get('observableChild')).toBeUndefined();
+          done();
```
Andarist, line 1979: I'm not sure if this really should be allowed. It's basically bypassing the machinery of state machine, allowing for arbitrary, imperative, manipulation of its internal state. This also doesn't quite deliver proper cleanup:
```js
typeof service.state.context.machineRef === 'object'
typeof service.state.context.promiseRef === 'object'
typeof service.state.context. observableRef === 'object'
```

This is the common case - spawned actors are usually assigned to the context. Clearing context values would definitely be too magical (and not trivial - with the current APIs it would require a full crawl of the current context).

davidkpiano:
Will rework this to use the `stop(...)` action creator instead.

PR #1566, opened 2020-10-18
Ensure exit actions are called for finalized invoked machines. Fixes #1109

### LINKED ISSUE #1109: exit action is not triggered on child machine
**Description**
Child machine runs using invoke on parent machine. Child machine has exit action defined. 

**Expected Result**
I expect child machine exit action to trigger after it reaches it final state

**Actual Result**
Exit action on child machine is not fired

**Reproduction**
```javascript
const childMachine = Machine({
  entry: () => { console.log('entry') },
  exit: () => { console.log('exit') },
  initial: 'a',
  states: {
    a: {
      type: 'final'
    }
  }
})

const parentMachine = Machine({
  initial: 'active',
  states: {
    active: {
      invoke: {
        src: childMachine,
        onDone: 'finished'
      }
    },
    finished: {
      type: 'final'
    }
  }
})
```
https://xstate.js.org/viz/?gist=5ebf91269382bbd3e0bc66c62643a2fa

davidkpiano:
This PR fixes #1109 by executing exit actions when an invoked machine reaches its final state.

### DIFF
--- a/.changeset/chatty-wombats-worry.md
+++ b/.changeset/chatty-wombats-worry.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Exit actions will now be properly called when an invoked machine reaches its final state. See #1109 for more details.
+Exit actions will now be properly called when an invoked machine reaches its final state. See [#1109](https://github.com/davidkpiano/xstate/issues/1109) for more details.
--- a/.changeset/chatty-wombats-worry.md
+++ b/.changeset/chatty-wombats-worry.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Exit actions will now be properly called when an invoked machine reaches its final state. See #1109 for more details.
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -275,6 +275,13 @@ export class Interpreter<
     const isDone = isInFinalState(state.configuration || [], this.machine);
 
     if (this.state.configuration && isDone) {
+      // exit interpreter procedure: https://www.w3.org/TR/scxml/#exitInterpreter
+      this.state.configuration.forEach((stateNode) => {
+        for (const action of stateNode.definition.exit) {
+          this.exec(action, state);
+        }
+      });
+
       // get final child state node
       const finalChildStateNode = state.configuration.find(
         (sn) => sn.type === 'final' && sn.parent === this.machine
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -1353,6 +1353,44 @@ describe('choose', () => {
 
     expect(service.state.context).toEqual({ answer: 42 });
   });
+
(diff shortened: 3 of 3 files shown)


.changeset/chatty-wombats-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Exit actions will now be properly called when an invoked machine reaches its final state. See #1109 for more details.
```
Andarist, line 5:
```suggestion
Exit actions will now be properly called when an invoked machine reaches its final state. See [#1109](https://github.com/davidkpiano/xstate/issues/1109) for more details.
```

packages/core/test/actions.test.ts
```diff
@@ -1353,6 +1353,44 @@ describe('choose', () => {
 
     expect(service.state.context).toEqual({ answer: 42 });
   });
+
+  // https://github.com/davidkpiano/xstate/issues/1109
+  it('exit actions should be called when invoked machine reaches final state', (done) => {
+    let exitCalled = false;
+    const childMachine = Machine({
+      exit: () => {
```
Andarist, line 1361: a root exit handler seems to be quite specific, it would be great to have an additional test that would test that exit handlers at different levels are executed in this situation

packages/core/src/interpreter.ts
```diff
@@ -275,6 +275,13 @@ export class Interpreter<
     const isDone = isInFinalState(state.configuration || [], this.machine);
 
     if (this.state.configuration && isDone) {
+      // exit interpreter procedure: https://www.w3.org/TR/scxml/#exitInterpreter
+      this.state.configuration.forEach((stateNode) => {
```
Andarist, line 279: Those exit handlers should be executed in "exit order". I'm not sure if `this.state.configuration` holds any guarantee about the order of items in it. It would actually be quite cool to rely on it without having to sort stuff here, I'm just not sure if we can safely make such an assumption now.

davidkpiano: In the `next` branch, these are sorted by exit order (I think there are some SCXML tests ensuring this). I'll see what I can do here to have the same behavior.

packages/core/test/actions.test.ts
davidkpiano: Will do.

packages/core/src/interpreter.ts
Andarist: The only SCXML I could find regarding this is: https://www.w3.org/Voice/2013/scxml-irp/250/test250.txml and it's classified as manual (it's one of the tests that requires the tester to look at the log output)

davidkpiano: This would be better suited for the `next` branch, since that more faithfully follows the SCXML algorithm. There would probably need to be substantial changes in order to ensure exit action order; let's scope this PR to just ensuring that they _are_ called.

Andarist (approved):

PR #1572, opened 2020-10-19
include license for packages

chenxsan:
closes https://github.com/davidkpiano/xstate/issues/1557

Here's the result of `npm publish --dry-run`:

image


Andarist:
q: do u expect this to be published asap or can those wait for the next releases of particular packages?

chenxsan:
I'm ok with next release.

PR #1577, opened 2020-10-21
Make stop() accept an expression. Fixes #1457

### LINKED ISSUE #1457: Calling stop on a child actor does not notify its parent that it should be cleaned

davidkpiano:
This PR allows expressions to be used in the `stop()` action creator:

```js
actions: stop(context => context.someActor)
```

This should be the way to stop spawned and invoked actors (as well as activities in v4, which are going away in v5 in favor of invoked actors). 

- Replaces https://github.com/davidkpiano/xstate/pull/1559
- Related: https://github.com/davidkpiano/xstate/pull/1434 (will need to be updated)

### DIFF
--- a/packages/core/test/interpreter.test.ts
+++ b/packages/core/test/interpreter.test.ts
@@ -13,7 +13,7 @@ import {
   createMachine
 } from '../src';
 import { State } from '../src/State';
-import { log, actionTypes, raise } from '../src/actions';
+import { log, actionTypes, raise, stop } from '../src/actions';
 import { isObservable } from '../src/utils';
 import { interval, from } from 'rxjs';
 import { map } from 'rxjs/operators';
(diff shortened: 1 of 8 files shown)


packages/core/test/interpreter.test.ts
```diff
@@ -1937,5 +1937,70 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
...
+
+      service.subscribe((state) => {
+        if (state.matches('present')) {
+          expect(state.children).toHaveProperty('machineChild');
+          expect(state.children).toHaveProperty('promiseChild');
+          expect(state.children).toHaveProperty('observableChild');
+
+          // state.children.machineChild.stop?.();
+          // state.children.promiseChild.stop?.();
+          // state.children.observableChild.stop?.();
```
Andarist, line 2000:
```suggestion
```

😅

Andarist (approved):

packages/core/test/interpreter.test.ts
davidkpiano: Much better, right?

PR #1581, opened 2020-10-22
pass context as first argument to immerAssign

nicklas-delphai:
With `produce` from immer `baseState` ( as called in the [docs](https://immerjs.github.io/immer/docs/produce)) is the scope. 

When using `immerAssign` within a `MachineConfig` the current context isn't in scope, which is why I think context should be passed as a first argument to the `immerAssigner`.


davidkpiano:
Thanks for the PR, but what problem does this solve?

If the developer wanted the original value (instead of the draft), they can use [`original(...)` from Immer](https://immerjs.github.io/immer/docs/original).

nicklas-delphai:
Well that solves the problem. I didn't know about that function.

PR #1587, opened 2020-10-24
Throw error if initial state is not specified. Fixes #674

### LINKED ISSUE #674: Always resolve state value to leaf nodes

davidkpiano:
This PR will now throw an error if an `initial` state is not specified for compound state node configs (e.g., if you specified a state node with `states` but didn't add an `initial` state). You will get a descriptive error:

```
No initial state specified for state node "#red". Try adding { initial: "walk" }
```


Andarist:
I believe that we have agreed to move this to the next branch - as this seems like a significant breaking change (even if it's an improvement). The context in the private discussion: https://statecharts.slack.com/archives/DKFRVHX3M/p1605211624000200 . The original issue also has mentioned that this should most likely only be implemented in v5 but it would be great to convert those errors into warnings in v4 so people could start "migrating" sooner.

packages/xstate-graph/test/graph.test.ts
```diff
@@ -347,6 +347,39 @@ describe('@xstate/graph', () => {
 
       expect(adj).toHaveProperty('"full" | {"count":5}');
     });
+
+    it('should get events via function', () => {
```
Andarist, line 351: Something weird has happened here - this branch contains changes from the other PR: https://github.com/davidkpiano/xstate/pull/1663

davidkpiano:
Moved to v5: #1669

PR #1594, opened 2020-10-27
Fixed an issue with internal `setState` in `useService` being called with 2 arguments instead of 1

Andarist:


davidkpiano:
LGTM but in the future, we'll have to adapt this to work with actual observers that have the form:

```js
something.subscribe({
  next: (value) => ...,
  error: (err) => ...,
  complete: () => ...
})
```

Just a note-to-self.

davidkpiano (approved):

Issue #1583, opened 2020-10-23
`this` has been rewritten to `undefined` and XState Inspect is Waiting for Connection


csell5:
**Description**
XState Inspect is Waiting for Connection on Svelte Apps maybe because of the way rollup is rewritting some typescript output to undefined.

```
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en/#error-this-is-undefined
node_modules/@xstate/inspect/es/index.js
1: var __assign = (this && this.__assign) || function () {
                   ^
2:     __assign = Object.assign || function(t) {
3:         for (var s, i = 1, n = arguments.length; i < n; i++) {
...and 1 other occurrence

```

**Expected Result**
When running locally, I would expect the Svelte SPA after going through its rollup bundle process to connect to https://statecharts.io/inspect.

**Actual Result**

Waiting for Connection is displayed with no errors in either browser. localhost:5000 -> statecharts.io/inspect. Having said that it's also been verified on a staging domain that it will not connect. So I don't think it's just localhost.

**Reproduction**

Here is a CodeSandbox where it actually works - https://codesandbox.io/s/xstate-inspect-fork-n4c92 but I think that's because Sandbox isn't going through the rollup process the same.

If you download that sandbox, `npm i`, then `npm run dev` you will be able to reproduce the problem.

**Additional context**
Please see https://github.com/distolma/xstate-svelte/issues/36 as @distolma and I have been working through this some already as I started there since I'm using that package. 

~ Clark

a-y-u-s-h:
Experiencing the same issue with locally running NextJS - `Xstate is waiting for connection`. At first I got `window` is not defined, but after putting `inspect({ iframe: false })` insde `useEffect`, and enabling pop-up, it's stuck at waiting for connection.

davidkpiano:
@a-y-u-s-h or @csell5 Do you have a reproducible repository where we can replicate this issue?

csell5:
If you just download locally to your drive that code sandbox above and run 'npm i', 'npm run dev' it should fail. 

I can throw it in a repo too but I wanted to show it working there as I don't think codesandbox is going through the same rollup process.

a-y-u-s-h:
Is this change published in the latest @xstate/inspect?

Andarist:
Not yet. @davidkpiano this release probably can be done manually unless you want to start releasing this package using changeset flow.

PR #1599, opened 2020-10-29
[@xstate/inspect] Add changeset

davidkpiano:
This should trigger it to be automatically released. Addresses #1583

### DIFF
--- a/.changeset/thin-papayas-wait.md
+++ b/.changeset/thin-papayas-wait.md
@@ -1,0 +1,5 @@
+---
+'@xstate/inspect': minor
+---
+
+The `@xstate/inspect` package is now built with Rollup.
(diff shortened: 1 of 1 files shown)


.changeset/thin-papayas-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/inspect': minor
+---
+
+The `@xstate/inspect` package is now built with Rollup.
```
Andarist, line 5:
```suggestion
The `@xstate/inspect` package is now built with Rollup which has fixed an issue with TypeScript compiler inserting references to `this` in the top-level scope of the output modules and thus making it harder for some tools (like Rollup) to re-bundle dist files as `this` in modules (as they are always in strict mode) is `undefined`.
```

Andarist (approved):

PR #1601, opened 2020-10-30
[docs] Add testing guide

davidkpiano:
This PR adds a testing guide for machines/services (different than `@xstate/test`).


Andarist (approved):
Would be cool to mention that usually, you have the machine interacting with external I/O and it might be preferable to test those rather than asserting using the actual machine API. This should kinda be given as those would actually be E2E/integration tests and those should always be possible but I feel like people forget about this all the time. Unless you don't agree that's good advice.

Other than that - LGTM.

PR #1606, opened 2020-10-31
[core] Add .onError() handler to service

davidkpiano:
This PR adds the `.onError(...)` handler to services:

```js
  .onError(err => {
    // err is instanceof Error
    console.error(err.message);
  })
  .start();
```

This was necessary to test the behavior of tweaking `reportUnhandledExceptionOnInvocation` to check that `originalError` is actually an `Error`, since `reject(anything)` might not be an `Error` object, so `err.stack` would itself throw an error.

Still deciding if we want to add `.onError((err, state) => { ... })` or other metadata 🤔


packages/core/src/interpreter.ts
```diff
@@ -417,6 +426,29 @@ export class Interpreter<
...
+   * @param errorListener The error listener
+   */
+  public onError(
+    errorListener: ErrorListener
+  ): Interpreter<TContext, TStateSchema, TEvent, TTypestate> {
+    this.errorListeners.add(errorListener);
+    return this;
+  }
+
+  private handleError(errorData: unknown): void {
```
Andarist, line 440: there are only 2 places in which this is currently called:
- when an error gets caught when we send an error event to itself after catching promise rejection
- when an action throws during `exec`

There are still a lot of uncovered situations:
- errors thrown during guards evaluation
- errors thrown during service instantiation
- errors received from machines (should we hook into their `onError`?) and from observables

packages/core/src/interpreter.ts
```diff
@@ -1010,8 +1036,12 @@ export class Interpreter<
...
             // Send "error.platform.id" to this (parent).
             this.send(toSCXMLEvent(errorEvent as any, { origin: id }));
-          } catch (error) {
-            reportUnhandledExceptionOnInvocation(errorData, error, id);
+          } catch (errorFromSend) {
+            if (this.machine.strict) {
+              this.handleError(errorFromSend);
+            }
+
+            reportUnhandledExceptionOnInvocation(errorData, errorFromSend, id);
```
Andarist, line 1044: I'm not sure what exact role does `reportUnhandledExceptionOnInvocation` serve now and what's the role of the `defaultErrorHandler`, they seem to be overlapping and most likely they can both be executed for the same error - leading to excessive logs in the console.

Andarist (commented):
It would be great to define (in the docs) what's the exact purpose of `onError`. What it should do and what it can't do. Is it useful only for reporting? Can it influence anyhow the machine's lifetime? This seems vastly important for users wanting to use this.

packages/core/src/interpreter.ts
```diff
@@ -417,6 +426,29 @@ export class Interpreter<
...
+   */
+  public onError(
+    errorListener: ErrorListener
+  ): Interpreter<TContext, TStateSchema, TEvent, TTypestate> {
+    this.errorListeners.add(errorListener);
+    return this;
+  }
+
+  private handleError(errorData: unknown): void {
+    const errorInstance =
```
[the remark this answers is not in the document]
davidkpiano: What should be the resolution here? Have error listeners treat received errors as `unknown`?

packages/core/src/interpreter.ts
davidkpiano: Right, we should handle errors there as well.

packages/core/src/interpreter.ts
davidkpiano: `defaultErrorHandler` has been removed.

packages/core/src/interpreter.ts
```diff
@@ -417,6 +426,29 @@ export class Interpreter<
...
+      errorData instanceof Error ? errorData : new Error(errorData as any);
+
+    if (this.errorListeners.size === 0) {
```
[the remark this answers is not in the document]
davidkpiano: `defaultErrorHandler` has been removed.

packages/core/src/interpreter.ts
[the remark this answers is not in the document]
davidkpiano: Added `isError`... thinking more about what we can do here, we can make a `class SCXMLError extends Error { ... }` class that has:

```js
{
  name: 'error.execution', // or 'error.platform' etc.
  message: '...',
  // non-standard properties
  data: errorData, // comes from original error data,
  origin: '...'
}
```

And throw that instead. Then, the original error data can always be determined from `error.data` and we won't force developers to determine the type of an `unknown` error instead.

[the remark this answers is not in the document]
Andarist: For reasons that I don't remember subclassing native classes ain't working properly in ES5 (and we transpile down to it). Take this simple input:
```js
class MyError extends Error {}
```
and the TS-transpiled output (same happens with Babel etc) for it:
```js
"use strict";
var __extends = (this && this.__extends) || (function () {
    var extendStatics = function (d, b) {
        extendStatics = Object.setPrototypeOf ||
            ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
            function (d, b) { for (var p in b) if (Object.prototype.hasOwnProperty.call(b, p)) d[p] = b[p]; };
        return extendStatics(d, b);
    };
    return function (d, b) {
        extendStatics(d, b);
        function __() { this.constructor = d; }
        d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
    };
})();
var MyError = /** @class */ (function (_super) {
    __extends(MyError, _super);
    function MyError() {
        return _super !== null && _super.apply(this, arguments) || this;
    }
    return MyError;
}(Error));
```
and now do the simple test:
```js
var err = new MyError()
console.log(err instanceof MyError) // oops! it's false 😬
```

---

> What should be the resolution here? Have error listeners treat received errors as unknown?

Either `unknown` or `any`. I think that error listeners should receive unwrapped errors, and not even "error events". If we want to provide extra metadata to error listeners then the second argument could contain it.

Error events should still be created so actual transitions could handle them. And I believe that `onError` shouldn't be called for handled error events. But this depends on the answers to the questions I've posted:
> It would be great to define (in the docs) what's the exact purpose of onError. What it should do and what it can't do. Is it useful only for reporting? Can it influence anyhow the machine's lifetime? This seems vastly important for users wanting to use this.

packages/core/src/interpreter.ts
```diff
@@ -432,8 +433,9 @@ export class Interpreter<
   }
 
   private handleError(errorData: unknown): void {
-    const errorInstance =
-      errorData instanceof Error ? errorData : new Error(errorData as any);
+    const errorInstance = isError(errorData)
+      ? errorData
+      : new Error(errorData as any);
```
Andarist, line 438: I believe that this still would result in a degraded DX because if we throw non-error object we'll end up with stringified messages like `[object Object]` which ain't that helpful. Whatever gets thrown should just be preserved and we shouldn't try to wrap it in any way - it mostly should be treated as opaque by us.

packages/core/src/interpreter.ts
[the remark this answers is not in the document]
davidkpiano: Okay, no longer wrapping in error 👍

packages/core/src/interpreter.ts
davidkpiano: Fixed

Andarist:
Would be great to define the role of this handler as mentioned here: https://github.com/davidkpiano/xstate/pull/1606#pullrequestreview-539334092 . We might change the semantics later but would be great to establish what this PR is currently aiming for.

davidkpiano:
> It would be great to define (in the docs) what's the exact purpose of `onError`. What it should do and what it can't do.

The purpose of `onError` is to be a mechanism for the parent to subscribe to errors, whether it's `error.*` events or thrown errors.

> Is it useful only for reporting?

It's useful for reporting and reacting to errors. 

> Can it influence anyhow the machine's lifetime?

No.

Andarist:
> The purpose of onError is to be a mechanism for the parent to subscribe to errors, whether it's error.* events or thrown errors.

Is it supposed to handle only *uncaught*/unhandled errors/events or all? I think the best one would be the former - handled stuff would lead to noise. I haven't rechecked right now which is most likely based on the provided implementation - just trying to establish key goals of this PR without looking at the implementation for now.

Is it supposed to only receive errors that originated in the top-level machine or its invoked actors? Or should it (in the future perhaps) be responsible for receiving all uncaught errors from the whole "system" that it represents?

> It's useful for reporting and reacting to errors.

How one would react to an error from it? I mean - what action could one take from there? Just kill the machine? Or maybe something else?

davidkpiano:
> > The purpose of onError is to be a mechanism for the parent to subscribe to errors, whether it's error.* events or thrown errors.
> 
> Is it supposed to handle only _uncaught_/unhandled errors/events or all? I think the best one would be the former - handled stuff would lead to noise. I haven't rechecked right now which is most likely based on the provided implementation - just trying to establish key goals of this PR without looking at the implementation for now.
> 
> Is it supposed to only receive errors that originated in the top-level machine or its invoked actors? Or should it (in the future perhaps) be responsible for receiving all uncaught errors from the whole "system" that it represents?

To make it simple, `onError` is for all errors (uncaught and `error.*` event) that originate from the machine, since (according to the SCXML spec) they are the same thing: 

> Once the SCXML processor has begun executing a well-formed SCXML document, it _must_ signal any errors that occur by raising SCXML events whose names begin with `'error.'`.
>
> https://www.w3.org/TR/scxml/#ErrorEvents

Actors don't send errors to the parent by default; that's what `escalate()` is for (or `sendParent(...)` if you want to do it manually), but these should work the same -- if an error is sent to the parent, `onError` listeners will be called with that error.

> How one would react to an error from it? I mean - what action could one take from there? Just kill the machine? Or maybe something else?

That's up to the developer. Stop/restart the machine, log the error, etc.

Andarist:
I would temporarily park this PR - until I post the error-related RFC, based on our Discord conversations. It has the potential of influencing the overall design of error handling but also this particular API.

sebinsua:
@davidkpiano I noticed that [errors thrown within `.onTransition` handlers are unhandled](https://codesandbox.io/s/crazy-goldwasser-tobfn). Would this PR make it easier to catch these `Service` errors? Has the Error RFC been worked on yet? /cc @Andarist

The company I'm currently working for was considering using `xstate` within a backend Node server and using an `onTransition` handler to persist state changes into a database. In later versions of Node, unhandled rejections could crash the server, and ideally we'd want the server to be able to respond with a non-200 status code if there is an error and to also not commit a database transaction if any part of it fails.

1. Is it safe to do database inserts within an `onTransition` handler? Do we just wrap our handler in a `try-catch`?
2. Or, should we instead move all database interaction into an invoked service?

Am I thinking about this in the wrong way?

davidkpiano:
> @davidkpiano I noticed that [errors thrown within `.onTransition` handlers are unhandled](https://codesandbox.io/s/crazy-goldwasser-tobfn). Would this PR make it easier to catch these `Service` errors? Has the Error RFC been worked on yet? /cc @Andarist

Yes, the error RFC has been recently been made internally, and we will surface it soon.

> 1. Is it safe to do database inserts within an `onTransition` handler? Do we just wrap our handler in a `try-catch`?

Yes you can; best to wrap in a try/catch 👍 

> 2. Or, should we instead move all database interactions into an invoked service that can explicitly move to a new state on success/fail?

Currently, whatever is easiest for you all. Ideally, these interactions will be in an invoked service, but the end result is the same: the machine changing state is due to an event, whether those are events sent by a service or externally. Only a minor refactor is needed in the future if you're currently sending them externally and want to move to a service.

Andarist:
@sebinsua https://github.com/statelyai/rfcs/pull/4

Andarist:
This PR targets `main` and we don't plan to introduce any new major features to that release line. The error story is going to be revamped on the `next` branch and will be released in v5.

PR #1615, opened 2020-11-05
Improve Raise event example

tomByrer:
ref: https://spectrum.chat/statecharts/general/trying-to-understand-the-semantic-difference-between-send-and-raise~54667586-2cd2-4433-908d-a65d58b0979c

I was confused also until I made a demo with different paths.


davidkpiano:
Thanks, this does look better!

PR #1622, opened 2020-11-10
[xstate] Actor typing improvements

davidkpiano:
Addresses #1534


packages/xstate-react/test/useActor.test.tsx
```diff
@@ -198,7 +214,9 @@ describe('useActor', () => {
         done();
       }
 
-      return <ChildTest actor={state.context.actorRef} />;
+      const { actorRef } = state.context;
+
+      return actorRef ? <ChildTest actor={actorRef} /> : null;
```
Andarist, line 219: This is correct from the perspective of the types but it stops fully validating what this test is about. The test's title is:
> initial spawned actor should be immediately available

```suggestion
      return <ChildTest actor={actorRef!} />;
```

packages/xstate-react/src/types.ts
```diff
@@ -45,6 +46,12 @@ export interface ActorRefLike<TEvent extends EventObject, TEmitted = any>
   [key: string]: any;
 }
 
+export type ActorRefFrom<
```
Andarist, line 49: Wouldnt this also be useful as `xstate`'s export?

packages/xstate-react/src/useActor.ts
```diff
@@ -1,12 +1,16 @@
...
-export function useActor<TEvent extends EventObject, TEmitted = any>(
-  actorRef: ActorRefLike<TEvent, TEmitted> | Actor,
-  getSnapshot: (actor: typeof actorRef) => TEmitted = (a) =>
+export function useActor<
+  TEvent extends EventObject,
+  TEmitted = unknown,
+  TActor extends ActorRefLike<TEvent, TEmitted> = ActorRefLike<TEvent, TEmitted>
+>(
+  actorRef: TActor & ActorRefLike<TEvent, TEmitted>,
+  getSnapshot: (actor: TActor) => TEmitted = (a) =>
```
Andarist, line 13: Could you describe what you have been trying to do here? It's a little bit mind-bending and I'm lost here. I don't understand why `TActor` has been introduced and why `ActorRefLike<...>` is not just used directly. When one would like to provide a subtype of that `ActorRefLike` type? I also don't understand what `TActor & ActorRefLike<TEvent, TEmitted>` is supposed to do 🤔

Andarist (commented):
I've tried to apply this patch on the original issue and either I'm doing something wrong or it's not quite fixing the reported problem:
https://codesandbox.io/s/thirsty-goldberg-z471l?file=/src/App.tsx

packages/xstate-react/src/useActor.ts
davidkpiano: Try removing `ActorRefLike<...>`. The `TEmitted` generic param becomes `unknown` which makes TS fail to properly infer the `actorRef`.

`ActorRefLike<...>` cannot be used directly because the `actor` argument in `getSnapshot` must match the inferred `actorRef` argument.

packages/xstate-react/src/types.ts
davidkpiano: Not yet; `ActorRef` doesn't exist there yet.

packages/xstate-react/test/types.test.tsx
```diff
@@ -131,4 +122,75 @@ describe('useMachine', () => {
 
     render(<YesNo />);
   });
+
+  // Example from: https://github.com/davidkpiano/xstate/discussions/1534
+  it('spawned actors should be typed correctly', () => {
```
Andarist, line 127: I have no idea why but this test doesn't compile correctly on the codesandbox I've provided: https://codesandbox.io/s/thirsty-goldberg-z471l?file=/src/App.tsx . I've also rechecked by copy-pasting this code test into the codesandbox and it suffers from the same problem. It might be depending on tsconfig settings but I've tried enabling/disabled TS strict mode in the repo and on the codesandbox and it didn't change individual results :/

davidkpiano: Not sure why, but I'm basing the validity of this on whether it passes in this repo, not in CodeSandbox.

Andarist: This repo is not perfect for replicating all kind of behaviors for two main reasons:
- it's not using TS strict mode and in general, might also use different TS options than what a user is using
- it's not using the same TS version

Arguably the second argument is not that important because we only tend to support the latest TS version but the first problem is real. If only problems would manifest with looser configs than what we have we could just advise using stricter settings - but it's the other way around here and I believe that supporting the strict mode should be our goal.

It seems that this is indeed strict mode's problem. I've created an isolated repository to test this: https://github.com/Andarist/xstate-1622-test/commits/master

1. First I've attempted to only use the new version of `@xstate/react` -> it didn't work
2. I've tried to provide an explicit type param to the used `assign` -> it still didn't work
3. I've added a new `xstate` version and reverted the explicit type param -> it still didn't work
4. I've finally re-added that explicit type param -> it did work

And later I've confirmed that removing the strict mode and removing that explicit type param (so the situation we are having here, in this repo) removes the type problem as well.

So in a sense - this test here doesn't validate in full the user's problem. I don't believe we should try to set up things to try validate those kinds of things here, but it's IMHO an important takeaway that we shouldn't take for granted that is we don't experience type problems in this repo then people won't experience them as well.

The really weird thing for me is that this explicit type param on `assign` has been changing the return type of the whole `createMachine` call. How this can even happen if supposedly providing any explicit type param turn off inference completely?

**Without explicit type param on assign - not working**
Screenshot 2020-11-17 at 14 22 19

**With explicit type param on assign - working**
Screenshot 2020-11-17 at 14 22 32

Andarist: Note: I have a potential fix for the `assign` needing an explicit type param in the pipeline but first I need to take this through the finish line: https://github.com/davidkpiano/xstate/pull/1543 , I have no idea if it's going to help with this particular case but according to my tests it has fixed other, similar situations with `assign`

packages/xstate-react/src/types.ts
davidkpiano: Now it does!

davidkpiano:
@Andarist Just made some massive improvements to the types, can you re-review?

packages/xstate-react/src/types.ts
```diff
@@ -26,25 +26,6 @@ export interface Subscribable<T> {
   ): Subscription;
 }
 
-export interface ActorRef<TEvent extends EventObject, TEmitted = any>
```
Andarist, line 29: were those considered to be public API? i don't mind breaking such things when it comes to TS in minor versions but if using those was advertised in issues etc then we should at least mention that they have been removed/moved elsewhere

packages/core/src/types.ts
```diff
@@ -1221,25 +1220,39 @@ export namespace SCXML {
...
 export type Spawnable =
   | StateMachine<any, any, any>
   | Promise<any>
   | InvokeCallback
   | Subscribable<any>;
+
+export interface ActorRef<TEvent extends EventObject, TEmitted = any>
+  extends Subscribable<TEmitted> {
+  send: Sender<TEvent>;
+  [key: string]: any;
```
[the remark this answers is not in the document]
Andarist: In general, it would be of great help to me (and probably to the users as well) if we had some kind of a page documenting our approach towards TS types. For instance, I try to stay away from `any` (as defaults for generics, object values, indexer values, etc) as much as I can. IMHO it's better to be safe by default and cast things where needed if necessary as it at least "documents" where "unsafe boundaries" lie in the code. The codebase is often not written in such a way so I naturally (to me) comment on each new call introducing such things - but if more explicit design goals would be written down then I could maybe just think more in their spirit.

packages/xstate-react/test/useActor.test.tsx
```diff
@@ -140,8 +157,9 @@ describe('useActor', () => {
 
     const Test = () => {
       const [state] = useMachine(machine);
+      const { actorRef } = state.context;
 
-      return <ChildTest actor={state.context.actorRef} />;
+      return actorRef ? <ChildTest actor={actorRef} /> : null;
```
Andarist, line 162: It seems that last time I've commented on the wrong line - although the comment was still sound for that other place - so gonna just link to that older comment:
https://github.com/davidkpiano/xstate/pull/1622#discussion_r524386007

```suggestion
      return actorRef ? <ChildTest actor={actorRef} /> : null;
```
```suggestion
      return <ChildTest actor={actorRef!} />;
```

packages/core/src/interpreter.ts
```diff
@@ -252,7 +252,8 @@ export class Interpreter<
 
     // Update children
     this.children.forEach((child) => {
-      this.state.children[child.id] = child;
+      const name = 'id' in child ? child.id : '';
+      this.state.children[name] = child;
```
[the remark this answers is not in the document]
davidkpiano: Yeah, going to refine these still - these should be some special `XStateActorRef extends ActorRef` (pseudocode) where `XStateActorRef` has the `id`

packages/core/src/interpreter.ts
```diff
@@ -1027,27 +1032,29 @@ export class Interpreter<
       }
     );
 
-    const actor = {
+    const actor: Actor<T, never> = {
```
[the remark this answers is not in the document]
davidkpiano: We should stop using `Actor<...>`

packages/core/src/types.ts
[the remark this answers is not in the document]
davidkpiano: That's true, I'll omit `[key: string]` and make it more explicit.

> but if more explicit design goals would be written down then I could maybe just think more in their spirit.

Yeah, we should do this. V5 should definitely have some design goals re: TypeScript, and I'm sure @mattpocock can help with this.

packages/xstate-react/src/types.ts
davidkpiano: Deleted here because it's available in `XState` - we can just re-export it if we don't want to break things?

mattpocock (approved):

packages/core/src/interpreter.ts
```diff
@@ -654,9 +661,11 @@ export class Interpreter<
       this.parent && (to === SpecialTargets.Parent || this.parent.id === to);
     const target = isParent
       ? this.parent
-      : isActor(to)
+      : isString(to)
+      ? this.children.get(to as string) || registry.get(to as string)
+      : isActor(to) || isSpawnedActor(to)
```
Andarist, line 666: There is no way for `isSpawnedActor` returning true without `isActor` returning true, so this check is redundant
```suggestion
: isActor(to)
```

packages/core/src/interpreter.ts
```diff
@@ -252,7 +259,8 @@ export class Interpreter<
 
     // Update children
     this.children.forEach((child) => {
-      this.state.children[child.id] = child;
+      const name = 'id' in child ? child.id : '';
```
Andarist, line 262: as a child has to be `SpawnedActorRef` now the `'id' in child` check is not required now

packages/xstate-react/src/useService.ts
```diff
@@ -33,11 +9,18 @@ export function useService<
 >(
   service: Interpreter<TContext, any, TEvent, TTypestate>
 ): [State<TContext, TEvent, any, TTypestate>, PayloadSender<TEvent>] {
-  const serviceActor = useMemo(() => fromService(service), [service]);
+  if (process.env.NODE_ENV !== 'production' && !('machine' in service)) {
+    throw new Error(
+      `Attempted to use an actor-like object instead of a service in the useService() hook. Please use the useActor() hook instead.`
```
Andarist, line 14: Please change this:
https://github.com/davidkpiano/xstate/blob/fc95a08a4c8b192a1cf947be640de7e79c540e68/packages/xstate-react/src/useActor.ts#L47
to
```js
const subscription = actorRef.subscribe(emitted => setCurrent(emitted)); 
```
Otherwise we'll end up with this problem again: https://github.com/davidkpiano/xstate/pull/1594/files

packages/xstate-react/src/useService.ts
```diff
@@ -33,11 +9,18 @@ export function useService<
...
+    );
+  }
 
   const [state] = useActor<TEvent, State<TContext, TEvent, any, TTypestate>>(
-    serviceActor,
-    (actor) => (actor as typeof serviceActor).current
+    service,
+    () =>
+      ('status' in service ? service.status : (service as any)._status) !== 0
```
Andarist, line 21: might be worth to keep this comment here:
```js
// TODO: remove compat lines in a new major, replace literal number with InterpreterStatus then as well
```

packages/xstate-react/src/types.ts
Andarist: This would still be a breaking change because `@xstate/react` can be used with versions of `xstate` that do not export those. Unfortunately, peer dependencies are PITA in that regard. I believe that the simplest way around this would be to create `@xstate/types` package that could be a regular dependency of both.

davidkpiano: > I believe that the simplest way around this would be to create `@xstate/types` package that could be a regular dependency of both.

Not a fan of adding dependencies, especially if these types should ideally live in `xstate`. I think a peer-dependency should be fine.

davidkpiano: I added the `ActorRef` back, but as the same `ActorRef` as what was added in XState core. That should be all that is necessary to prevent breaking changes.

packages/core/src/types.ts
```diff
@@ -1221,25 +1220,47 @@ export namespace SCXML {
...
 }
 
 export type Spawnable =
   | StateMachine<any, any, any>
   | Promise<any>
   | InvokeCallback
   | Subscribable<any>;
+
+export interface ActorRef<TEvent extends EventObject, TEmitted = any>
+  extends Subscribable<TEmitted> {
```
Andarist, line 1249: what is the desired interface for an actor and its `subscribe` method? Right now this extends `Subscribable` that required `subscribe` to implement **both** its overloads - it seems pretty limiting 🤔 could we change the `Subscribable` to be a union of its current overloads?

davidkpiano: It currently does need to implement both overloads, though. This is the same expected behavior as RxJS: https://github.com/ReactiveX/rxjs/blob/master/src/internal/Observable.ts#L71-L78 and the [tc39 Observable spec](https://github.com/tc39/proposal-observable#observable):

```ts
interface Observable {

    // ...

    // Subscribes to the sequence with an observer
    subscribe(observer : Observer) : Subscription;

    // Subscribes to the sequence with callbacks
    subscribe(onNext : Function,
              onError? : Function,
              onComplete? : Function) : Subscription;

    // ...
}
```

Andarist: Right, it seems a little bit limiting for users to implement both. Take a look at this simple example:
https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPICMDO0Bu0A8AKgHzIDeAUMsiBAB5gBcyAFDnADYCuEzhAlMgC8pHAHtgAEwDcVZNChiozFguZwQAT0Ejk4qbOoIxAWwAOHCJBU7REmRQC+FCqEixEKAMpcsCKMBmYMBiIORyXCCYvpj+wOgQLPzM+g7OruDQ8EjIPn4B6HDolkSklNTR+fGJYli40MwY2FB4UKXJuTFxQSEghsiVsQWJctS0DCrs3LzIAsJ2UgA0o-JQilAA-CoKSupatnr2y9RGphZWEFusB6lyHXlDgcGhsuluWZ7IAIIIYEoAShAYEQAKJ4cDyBgQECSTDkZBgTRmGaYMABEAAc2QjkWsxBJmAYEgkmEyA0mmIcnokBhcIecUKxQgoIJRIgkjKcmwMO24KYeL5N3srxcFBgkV+vTJCCQQR+fygLEQCuY8oBQPw5Nx5OIgjI6WMUTA0oVpPKJ25klUfOYZARSJmAHIAGKoVCO7F6lwnbHHC1dYYscb8ticZggLgmBJQIVSL0rZzpRCysBqxXKpT8IA

I would expect as a user this to work - XState itself will only ever call `.subscribe` using a particular signature (it doesn't matter which one it is right now). If I, as the actor author, want to support multiple signatures so it could be used by different consumers then this is fine - it's up to me, I can do that. However, right now it is required from me to support all of them which brings some boilerplate and ceremony that could be avoided.

I don't have a strong opinion about this - just something I've noticed while playing around with this branch.

packages/xstate-react/src/useActor.ts
```diff
@@ -1,13 +1,29 @@
...
+  return 'state' in actorRef;
+}
+
+function isDeferredActor<T extends ActorRef<any>>(
+  actorRef: T
+): actorRef is T & { deferred: boolean } {
+  return 'deferred' in actorRef;
+}
+
+export function useActor<
```
Andarist, line 19: I have no idea why yet but this signature is being problematic with `strictFunctionTypes` setting:
https://www.typescriptlang.org/play?noImplicitAny=false&strictNullChecks=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&alwaysStrict=false&noEmitHelpers=true#code/JYOwLgpgTgZghgYwgAgPICMDO0Bu0A8AKgHzIDeAUMsiBAB5gBcyAFDnADYCuEzhAlMgC8pHAHtgAEwDcVZNChiozFguZwQAT0Ejk4qbOoIxAWwAOHCJBU7REmRQC+FCqEixEKAMpcsCKMBmYMBiIORyXCCYvpj+wOgQLPzM+g7OruDQ8EjIPn4B6HDolkSklNTR+fGJYli40MwY2FB4UKXJuTFxQSEghsiVsQWJctS0DCrs3LzIAsJ2UgA0o-JQilAA-CoKSupatnr2y9RGphZWEFusB6lyHXlDgcGhsuluWZ7IAIIIYEoAShAYEQAKJ4cDyBgQECSTDkZBgTRmGaYMABEAAc2QjkWsxBJmAYEgkmEyA0mmIcnokBhcIecUKxQgoIJRIgkjKcmwMO24KYeL5N3srxciORyDB0LAGAAVhBfqSyGKUWjQBj0pJ5Rw4FAUDBIr9esguNgfn82nJCJKIdTobCJXzZfKwMc8azia7CGalJCafbvVBAcCrXzcVb3ezSEJvr8AUDQaG3YTiZTiCw5IhzUG+AHXRirF4QHAzJgABZiMBXFiZ3azAMHcPJ9nICgAelbJ2j1dsbY7J2QwEwAYA6oTS14wHBIN3kBsyQA6VFTlDMSKamCgdkUDoAbUbbMkuNUfL41rAQqkAF0XJqENrdchjFEwGS4HBmAGg-gyAikTMAOQwGIYj-tiuIgFwJgJFAlIUE+qLIDuS6QLi3KSJepImhAAbVm+-BAA

The only thing I've recognized so far about this problem is that introducing `TActor` generic is causing this.

packages/xstate-react/src/useActor.ts
```diff
@@ -28,7 +44,7 @@ export function useActor<TEvent extends EventObject, TEmitted = any>(
   useIsomorphicLayoutEffect(() => {
     actorRefRef.current = actorRef;
     setCurrent(getSnapshot(actorRef));
-    const subscription = actorRef.subscribe(setCurrent);
+    const subscription = actorRef.subscribe((emitted) => setCurrent(emitted));
```
Andarist, line 47: This is also a situation of how we don't utilize `error` and `complete` callbacks at all. What role do they play for us?

davidkpiano: We can/should use a full `observer` object for these. I'll make that change.

packages/xstate-react/src/useActor.ts
Andarist: Ok... I **think** I might somewhat understand this now. Please take a look at this even more simplified demo:
[TS playground](https://www.typescriptlang.org/play?noImplicitAny=false&strictNullChecks=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&alwaysStrict=false&noEmitHelpers=true#code/JYOwLgpgTgZghgYwgAgIILAeygJQjAHgChlkAVAUQDcJxkIAPSEAEwGdkBvZMATwAcIALmRswUUAHNkAXwA0RAHxcSo2ixEAKCDXAjKusAEpkAXmVVMwFgG4iMokRYQEAGzhQUMAK4gMwTBBkbzYIdCwoYlIDWjB6JnUObj5BETEJEGl5VTJw7HjmdjQMbDxCGPBFJU1VRAiy-TyoIiMRS2tHZzcPFARAsWQ67BEmsoJkgWFkAHIYTExp2SqiELCSqE0hqCMV0KbxnkmRWfnF+WL6-AOUqZOFpcVN9aMgA)

We may notice that the inferred generics for here are as follows:
```ts
function useActor<{
    type: string;
}, ActorRef<{
    type: string;
}>>(actorRef: ActorRef<{
    type: string;
}>): void
```
And this is the same with or without `strictFunctionTypes` setting. For some reason, TS infers here the least strict type for `TEvent` that satisfies this. So it's an inference limitation because if you take a look at the provided demo there is no problem is we supply generic explicitly.

Why this is a problem even without `strictFunctionTypes`? Well, the inferred and returned `Sender<TEvent>` is too loose so it allows incorrect calls:
[TS playground](https://www.typescriptlang.org/play?noImplicitAny=false&strictNullChecks=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&alwaysStrict=false&noEmitHelpers=true#code/JYOwLgpgTgZghgYwgAgIILAeygJQjAHgChlkAVAUQDcJxkIAPSEAEwGdkBvZMATwAcIALmRswUUAHNkAXwA0RAHxcSo2ixEAKCDXAjKusAEpkAXmVVMwFgG4iMokRYQEAGzhQUMAK4gMwTBBkbzYIdCwoYlIDWjB6JnUObj5BETEJEGl5VTJw7HjmdjQMbDxCGPBFJU1VRAiy-TyoIiMtHX1qWJNzZEtrR2c3DxQEQLFkOuwRJrKCZIFhZAByGExMJdkqolGQcdDWAEYzYNCmzUmoI0d9lgPNedTlgHUACVRKADUKHA2ZK+2xnEbgAmY4hMIlSIPRYrNa-OTFer4OY8BYiWHrTaKc6QoxAA)

Why that code errors with `strictFunctionTypes` setting? Because of its sole nature that changes what function types can be assigned to what function types (variance-related) - it's explained in [this SO answer](https://stackoverflow.com/a/51767568) and showcased in the corresponding demo:
[TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAEghgOwCYBsICcoF4oAp0QCOArhAM7ABcUASkaRQJTYB8t5YA9gmRAFB8AxtwpQAtiHjI06alNQZseAiXJV2q0QDIoAbygQAHsHRwAgugDm1CugCWCS1AC+zLG118o3qCobAAOiMTcysA4E4AVTBIdABhOF5cRgBuLx8CYGJ0BCgEYhQUKETihBA05yA)

I've also tried to play around with our case on some even more simplified case and to my surprise, I haven't replicated the behavior 1 to 1. We indeed get the same problem if we include the function property (and the generic isn't inferred to the most narrow type possible):
[TS playground](https://www.typescriptlang.org/play?noImplicitAny=false&strictNullChecks=false#code/C4TwDgpgBAQg9gDwDwGUoQcCA7AJgZyn2ACcBLbAcwD4oBeKAbwCgooA3ALigApiTuKAJT1a7OGVzMAvs2a4IAYwA2AQxLQAZgFdsi4GTjYo2-BBSkKlVOkw4CRS1QA0sW1jyF4yFNWo8AI0RuGCFucUk5BRV1aEUjYigghBDEJAByTTg4dOpmU3MnSkDEISA)
but if we make this a regular property a weird thing happens to inferred generics:
[TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAQg9gDwDwGUoQcCA7AJgZyn2ACcBLbAcwD4oBeKAbwCgooA3ALihWYF9mzXBADGAGwCGJaADMArthHAycbFDn4IKUhUqp0mHASI6qAGlgGseQvGQpq1ABQAjRNxgBKbuzhlcgsLiUtAiqsRQbggeiEgA5DJwcHHUzBpappSuiJ5AA)
By inspecting the `useString` call we can see this:
```ts
function useString<string, Box<"foo">>(box: Box<"foo">): void
```
which has a different type inferred at 2 different positions even though it should be the same according to the declaration 😱 

---

Either way - it seems that we just can't rely on inference here. It's just incapable of inferring this properly right now. I propose to make the second argument a regular value that has be given upfront - and if it's not then we can try to access `actorRef.state`. This should simplify the types and the inference.

davidkpiano: In that case, I think it might be better to have the full actor as the first generic type param:

```ts
declare function useActor<
  TActor extends ActorRef<any, any> = ActorRef<any, any>
>(
  actorRef: TActor,
  getSnapshot?: (actor: TActor) => TActor extends ActorRef<any, infer TEmitted> ? TEmitted : never
): [TActor extends ActorRef<any, infer TEmitted> ? TEmitted : never, (event: TActor extends ActorRef<infer TEvent, any> ? TEvent : never) => void]

declare const aaa: ActorRef<{ type: 'foo' }, number>

const [state, send] = useActor(aaa)
```

[TS Playground](https://www.typescriptlang.org/play?noImplicitAny=false&strictNullChecks=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&alwaysStrict=false&noEmitHelpers=true#code/JYOwLgpgTgZghgYwgAgPICMDO0Bu0A8AKgHzIDeAUMsiBAB5gBcyAFDnADYCuEzhAlMgC8pHAHtgAEwDcVZNChiozFguZwQAT0Ejk4qbOoIxAWwAOHCJBU7REmRQC+FCqEixEKAMpcsCKMBmYMBiIORyXCCYvpj+wOgQLPzM+g7OruDQ8EjIPn4B6HDolkSklNTR+fGJYli40MwY2FB4UKXJuTFxQSEghsiVsQWJctS0DCrs3LzIAsJ2UgA0o-JQilAA-CoKSupatnr2y9RGphZWEFusB6lyHXlDgcGhsuluWZ7IAIIIYEoAShAYEQAKJ4cDyBgQECSTDkZBgTRmGaYMABEAAc2QjkWsxBJmAYEgkmEyA0mmIcnokBhcIecUKxQgoIJRIgkjKcmwMO24KYeL5N3srxciORyDB0LAGAAVhBfqSyGKUWjQBj0pJ5Rw4FAUDBIr9esguNgfn82nJCGalJCabDvr8AUD8OTceTSEIHebAcDXWStJTiCw5IhvUC+NaoMdkBirF4QHAzJgABZiMBXFih3azSMHK2OqC26H2yM+l1aXGgGDQPGs4mkDa1wnE5DMWitCgdADa+fNRdpXqdvoryCrNcI+Ob7IbTbZJLbEFauNUfIjBf7JYLZbHhYnfLdAeQjb3UtbNEX0CFUgAui5NQhtbrkMYomAyXA4MxS86yAikTMAHIYDEMQAOxXEQC4EwEigSkKBfVFkC7VE4EgXFuUka9SRNCBI0zD9+CAA)

> I propose to make the second argument a regular value that has be given upfront - and if it's not then we can try to access `actorRef.state`.

No, we shouldn't assume that `actorRef.state` exists; remember, you didn't want `ActorRef` to be too loose re: `{ send: ..., [key: string]: any }`.

Andarist: > In that case, I think it might be better to have the full actor as the first generic type param:

You could simplify this a little bit by using a single conditional type for the return type and I don't think the default for this generic changes anything here (they usually only make sense if you are having more than one type param) 🤔 

[TS Playground](https://www.typescriptlang.org/play?noEmitHelpers=true#code/JYOwLgpgTgZghgYwgAgPICMDO0Bu0A8AKgHzIDeAUMsiBAB5gBcyAFDnADYCuEzhAlMgC8pHAHtgAEwDcVZNChiozFguZwQAT0Ejk4qbOoIxAWwAOHCJBU7REmRQC+FCqEixEKAMpcsCKMBmYMBiIORyXCCYvpj+wOgQLPzM+g7OruDQ8EjIPn4B6HDolkSklNTR+fGJYli40MwY2FB4UKXJuTFxQSEghsiVsQWJctS0DCrs3LzIAsJ2UgA0o-JQilAA-CoKSupatnr2y9RGphZWEFusB6lyHXlDgcGhsuluWZ7IAIIIYEoAShAYEQAKJ4cDyBgQECSTDkZBgTRmGaYMABEAAc2QjkWsxBJmAYEgkmEyA0mmIcnokBhcIecUKxQgoIJRIgkjKcmwMO24KYeL5N3srxciORyDB0LAGAAVhBfqSyGKUWjQBj0pJ5Rw4FAUDBIr9esguNgfn82nJCGalJCabDvr8AUD8OTceTKcQWHJEObATA+NaoMdkBirF4QHAzJgABZiMBXFg+3azQMHK2OqC26H2wN+l1aXGgGDQPGs4mkDalwnE5DMWitCgddPmrO0h2+51FkuESXgQsgYuZntl9mU6iVgDaw+r7Nxqj5fF7YCFUgAunI6xAGxRNQhtbrkMYomAyXA4Mxc86yAikTMAOQwMRiO-Y3EgLgmBJQSkUI+o5ATqicCQLi3KSKupImhAgaJme-BAA)

> No, we shouldn't assume that actorRef.state exists; remember, you didn't want ActorRef to be too loose re: { send: ..., [key: string]: any }.

But we already assume that `.state` might exist at runtime - this ain't a completely type-safe assumption but it seems like a fair tradeoff here. I don't think the proposed solution would really be any different from what we have now:

```js
// current
function useActor(
  actorRef,
  getSnapshot = (a) => isActorWithState(a) ? a.state : undefined
) {
  // just for demo purposes
  const snapshot = getSnapshot(actorRef)
  // ...
}

// proposed
function useActor(
  actorRef,
  snapshot
) {
  // just for demo purposes
  snapshot = snapshot === undefined && isActorWithState(a) ? a.state  : undefined
  // ...
}
```

This would improve the TS inference while using the "standard" generics that we use across the codebase:
[TS playground](https://www.typescriptlang.org/play?noEmitHelpers=true#code/JYOwLgpgTgZghgYwgAgPICMDO0Bu0A8AKgHzIDeAUMsiBAB5gBcyAFDnADYCuEzhAlMgC8pHAHtgAEwDcVZNChiozFguZwQAT0Ejk4qbOoIxAWwAOHCJBU7REmRQC+FCqEixEKAMpcsCKMBmYMBiIORyXCCYvpj+wOgQLPzM+g7OruDQ8EjIPn4B6HDolkSklNTR+fGJYli40MwY2FB4UKXJuTFxQSEghsiVsQWJctS0DCrs3LzIAsJ2UgA0o-JQilAA-CoKSupatnr2y9RGphZWEFusB6lyHXlDgcGhsuluWZ7IAIIIYEoAShAYEQAKJ4cDyBgQECSTDkZBgTRmGaYMABEAAc2QjkWsxBJmAYEgkmEyA0mmIcnokBhcIecUKxQgoIJRIgkjKcmwMO24KYeL5N3srxciORyDB0LAGAAVhBfqSyGKUWjQBj0pJ5Rw4FAUDBIr9esguNgfn82nJCJKIdTobCJXzZfKwMc8azictiCw5IhzYCYMwzQCgaC+bire72cRXZgQHAzJgABZiMBXCOE4nICgdADa6bZklxqj5fGtYCFUgAui5NQhtbrkMYomAyb9dt821B-fgyAikTMAOQwMRiAfY3EgLgmBJQSkUJuo5A51FwSC47mSSukk0QINQFi+pT8IA)

There is ofc a little bit of ambiguity in this version regarding calls like `useActor(actor, undefined)` but it also seems like a fair tradeoff as this shouldn't affect any real-life code. I don't see any difference when it comes to runtime semantics between the current and proposed solutions - the only difference is how the snapshot value gets provided (getter vs explicitly).

I don't mind your solution much - just exploring alternatives here now that I believe are just somewhat simpler conceptually (the need to use 2 conditional types raises the complexity quite a bit when using a single `TActor` approach).

davidkpiano: I disagree with getting the snapshot value "explicitly" as a mutable value, as was done before. That exposes implementation details, and makes a wrongful assumption (that the snapshot is always available as a mutated property of the actor), and also limits flexibility with the `useActor` hook. Basically, this should behave like `getSnapshot` in the [`useMutableSource`](https://github.com/reactjs/rfcs/blob/master/text/0147-use-mutable-source.md) hook and Redux's `store.getState()`.

packages/core/src/types.ts
davidkpiano: > However, right now it is required from me to support all of them which brings some boilerplate and ceremony that could be avoided.

Can you show me some example code where that boilerplate is occurring? The main use-cases regarding this are `.subscribe(...)`, as in consuming a subscribable object, which does not require anything of the developer.

Right now, the actors are assumed to come from `invoke` or `spawn()`, which will already do the necessary work for a compatible `.subscribe(...)`. The use-case of creating an actor externally and using that within `useActor(...)` seem to be very edge-case.

We can make `.subscribe()` an optional property, though 🤔

davidkpiano: Also, we can make the internal `toObserver(...)` function a helper function:

```js
const subscribe = (a, b, c) => {
  const observer = toObserver(a, b, c);

  // observer is an object with { next, error, complete }
}
```

Andarist: The boilerplate/ceremony part is showcased in the playground in my comment ([TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPICMDO0Bu0A8AKgHzIDeAUMsiBAB5gBcyAFDnADYCuEzhAlMgC8pHAHtgAEwDcVZNChiozFguZwQAT0Ejk4qbOoIxAWwAOHCJBU7REmRQC+FCqEixEKAMpcsCKMBmYMBiIORyXCCYvpj+wOgQLPzM+g7OruDQ8EjIPn4B6HDolkSklNTR+fGJYli40MwY2FB4UKXJuTFxQSEghsiVsQWJctS0DCrs3LzIAsJ2UgA0o-JQilAA-CoKSupatnr2y9RGphZWEFusB6lyHXlDgcGhsuluWZ7IAIIIYEoAShAYEQAKJ4cDyBgQECSTDkZBgTRmGaYMABEAAc2QjkWsxBJmAYEgkmEyA0mmIcnokBhcIecUKxQgoIJRIgkjKcmwMO24KYeL5N3srxcFBgkV+vTJCCQQR+fygLEQCuY8oBQPw5Nx5OIgjI6WMUTA0oVpPKJ25klUfOYZARSJmAHIAGKoVCO7F6lwnbHHC1dYYscb8ticZggLgmBJQIVSL0rZzpRCysBqxXKpT8IA)).

It's not possible to create a very simple and lightweight actor that would match our expectations. The `toObserver` util is just what would make it slightly easier for developers but is a part of the ceremony I'm talking about. It wouldn't be sufficient though because one has to still type arguments but maybe something like this would be enough to make work:
```ts
const subscribe: Subscribable<T>["subscribe"] = (a, b, c) => {
  const observer = toObserver(a, b, c);

  // observer is an object with { next, error, complete }
}
```

It ain't super straightforward to always type this using the provided `Subscribable` - it depends on how do u write your code, for example, an object's property is not that easy (intuitive) to type using this.

But I get that usually people are not supposed to write actors by hand, although it would be nice if this would be supported out of the box.

> We can make .subscribe() an optional property, though 🤔

It's not really about it being optional or not - I would like to actually provide it, just with simpler types.

--- 

This is not a blocking comment by any means - just something that I've observed while reviewing this. Food for thought and maybe not even smth that should be changed - I still think it was OK to raise this concern, after all, we want to primarily focus on actors and their interfaces, compatibility with different data types (like observables) should come second. I don't believe that consuming actors as observables is really that popular of a use case 🤔

packages/xstate-graph/test/graph.test.ts
```diff
@@ -347,6 +347,39 @@ describe('@xstate/graph', () => {
 
       expect(adj).toHaveProperty('"full" | {"count":5}');
     });
+
+    it('should get events via function', () => {
```
Andarist, line 351: Something weird has happened here as well - this branch contains changes from the other PR: https://github.com/davidkpiano/xstate/pull/1663

I know it's somewhat nitpicky but such things truly make it harder to focus on the given PR, maybe not in here because the changes are nicely isolated to `@xstate/graph` but in general this can be truly problematic if changes affect the same package.

A side question: do you know how this has happened? 🤔 I'm curious from the technical PoV - the commit that introduces this (https://github.com/davidkpiano/xstate/commit/d2a59a92cce84dba5c212f654e30de2a6d85a9ea) is the same commit on all 3 branches but there are no merges between those branches as far as I can tell so how did it even get onto them as the same commit? Git is still a mystery to me sometimes 😅

Andarist: Oh, my hypothesis is that you have merged in the #1663 **locally** and merged the local version of `master` into 2 other branches. That would kinda explain it (maybe?).

packages/xstate-react/src/useActor.ts
Andarist: Well, our usage of `getSnapshot` is different from the `useMutableSource` - they **always** read state through it, even when they are notified about a change through a subscription (Redux does the same as `useMutableSource`). And we only use it when initializing the state and in the subscribing effect.

I agree though that given the second way in which we are using it this has to be a getter function and can't be a plain value.

packages/xstate-react/src/useActor.ts
```diff
@@ -1,24 +1,49 @@
...
+  TActor extends ActorRef<any, any>
+> = TActor extends ActorRef<any, infer TEmitted> ? TEmitted : never;
+
+const noop = () => {
+  /* ... */
+};
+
+export function useActor<
+  TActor extends ActorRef<any, any> = ActorRef<any, any>
+>(
```
Andarist, line 32: Maybe worth adding overloads to support `<TContext, TEvent>`? We could do this:
1. the first overload - the complex one with conditional types
2. the second one - `<TContext, TEvent, TActorWithoutDefault>`, so people could pass a custom `ActorRef` if needed
3. the third one - `<TContext, TEvent>`, what would probably be used most of the time when typed explicitly, has to come last to act as a fallback

packages/xstate-react/src/useActor.ts
Andarist: > EDIT: also, the implementation might/will probably change in v5, re: getting the snapshot value. We should not hardcode reading from actor.state.

Agreed, although I'm not sure how to do that without requiring `getSnapshot` at all times and from what I remember (but maybe I'm wrong) you were opposed to "standardizing" a common property in the Actor interface that could be read for those purposes.

packages/core/src/types.ts
davidkpiano: > It's not really about it being optional or not - I would like to actually provide it, just with simpler types.

We can either provide it with simpler types and lose compatibility with RxJS and other libraries that implement the tc39 spec (proposal), or we can maintain full compatibility. Let's aim for the latter.

For the developer wishing to make a compatible actor, the "boilerplate" types should be invisible to them (not directly imported). At most, the only thing they would need to import is the `Observer` type:

```ts
import { Observer } from 'xstate';

const actor = {
    send(event: { type: 'FOO' }) {

    },
    subscribe(next: Observer<number> | ((val: number) => void)) {
      return { unsubscribe: () => {} }
    }
}

acceptActor(actor)
```
[TS Playground](https://www.typescriptlang.org/play?ssl=33&ssc=40&pln=33&pc=21#code/JYOwLgpgTgZghgYwgAgPICMDO0Bu0A8AKgHzIDeAUMsiBAB5gBcyAFDnADYCuEzhAlMgC8pHAHtgAEwDcVZNChiozFguZwQAT0Ejk4qbOoIxAWwAOHCJBU7REmRQC+FCqEixEKAMpcsCKMBmYMBiIORyXCCYvpj+wOgQLPzM+g7OruDQ8EjIPn4B6HDolkSklNTR+fGJYli40MwY2FB4UKXJuTFxQSEghsiVsQWJctS0DCrs3LzIAsJ2UgA0o-JQilAA-CoKSupatnr2y9RGphZWEFusB6lyHXlDgcGhsuluWZ7IAIIIYEoAShAYEQAKJ4cDyBgQECSTDkZBgTRmGaYMABEAAc2QjkWsxBJmAYEgkmEyA0mmIcnokBhcIecUKxQgoIJRIgkjKcmwMO24KYeL5N3srxcMEiv16ZIQSCCPz+UBYiHlzDlAKB+HJuPJxEEZHSAHp9cgAHKhFBiGAIgAWKCKYjwyGMkhQwDhtCQmEwcCgmmQMEUJmtKGdeA4YmRUBcxiiYCl8tJ5RO3Mkqj5zDICKRMwA5AAxVCobPY3UuE7Y45JrrDFjjflNeptEBcEwJKCkAA+rDYnGYTZb0CFUn4JbL1CgVi4UDCGcigwZMyS83IjmxK2c6UQMrAqoVSqU-CAA)

packages/xstate-react/src/useActor.ts
davidkpiano: > you were opposed to "standardizing" a common property in the Actor interface that could be read for those purposes.

Yes, I'm opposed to it because we should make actors a minimal `{ send, subscribe }` interface. The idea of having a property that "holds" a snapshot value is conceptually the same as having something that previously subscribed (or asked for via sending an event with a self-reference) to an actor and locally held (cached) the last-known-emitted value. But it shouldn't be a built-in part of actors.

packages/xstate-react/src/useActor.ts
davidkpiano: I don't think we can add overloads just for generics - not sure if that's possible in TS.

Andarist: A lot of TS "stdlib" is using this pattern - totally possible to provide overloads with totally different generic types, different amounts of them etc. An adjusted demo from the other post:
[TS playground](https://www.typescriptlang.org/play?noEmitHelpers=true#code/JYOwLgpgTgZghgYwgAgPICMDO0Bu0A8AKgHzIDeAUMsiBAB5gBcyAFDnADYCuEzhAlMgC8pHAHtgAEwDcVZNChiozFguZwQAT0Ejk4qbOoIxAWwAOHCJBU7REmRQC+FCqEixEKAMpcsCKMBmYMBiIORyXCCYvpj+wOgQLPzM+g7OruDQ8EjIPn4B6HDolkSklNTR+fGJYli40MwY2FB4UKXJuTFxQSEghsiVsQWJctS0DCrs3LzIAsJ2UgA0o-JQilAA-CoKSupatnr2y9RGphZWEFusB6lyHXlDgcGhsuluWZ7IAIIIYEoAShAYEQAKJ4cDyBgQECSTDkZBgTRmGaYMABEAAc2QjkWsxBJmAYEgkmEyA0mmIcnokBhcIecUKxQgoIJRIgkjKcmwMO24KYeL5N3srxciORyDB0LAGAAVhBfqSyGKUWjQBj0pJ5Rw4FAUDBIr9esguNgfn82nJCGalJCabDvr8AUD8OTceTSEIHebAcDXWStJTiCw5IhvUC+NaoMdkBirF4QHAzJgABZiMBXFih3azSMHK2OqC26H2yM+l1aXGgGDQPGs4mkDa1wnE5DMWitCgdADa+fNRdpXqdvoryCrNcI+Ob7IbTbZJLbEFauNUfIjBf7JYLZbHhYnfLdAeQjb3UtbNEX0CFUgAuhRNQhtbrkPqQIbQsbTQX8JbJRDqcW4TIBEkRVdEsRxH863ZCggxDLdw0HKAyxPcBcQnKCOWjWMwHjRMUzTDMs2URDkN-MA0MnOdiDzSjiU7Zge1o9llwvcA+DIq9JFvO8tR1FBjCiMAyQLZhS2dIDlWYAByGAxDEKTsVxEAuBMBIoEpCgBNRZAu1ROBIFxbkuNJE0IEjTMC34TTQm03SwH0iAACZDOLRzrxMz9zXwCSQOk2T5MUmgVLUoMiKsrShLshyAGYXJhaL3M9UzI3wMTgR85E-LkhScSC1ToGIULLKAA)

davidkpiano: Oh cool! Didn't know you can do that 👍

davidkpiano: Upon trying this, adding the overload doesn't work well with inference.

We want `useActor(someActor)` to properly infer to `useActor<typeof someActor>(someActor)` but because of the overload, this will not work as expected.

Let's tackle in a separate PR if you think this is important; I am fine with having only `useActor<TActor>` supported for now; it doesn't make sense to separate `TEvent, TEmitted` when both of those are inherent parts of the same `TActor`.

Andarist: Could you take a look at this one: https://github.com/davidkpiano/xstate/pull/1668 ? Overloads are always matched from the top to the bottom and the matching stop on the first match so in a sense, it really shouldn't matter what overloads follow the signature with only `TActor` generic, and also the following overloads have no impact on the preceding ones.

Not a big deal but this would avoid a breaking change for the types - the second overload I've put there is matching what we have on master.

Andarist (approved):
I've retried this on the original case reported by the user ([see](https://github.com/Andarist/xstate-1622-test/commit/2dbe62c07cc683a4a4a09dadaad00eb8bc4b9346)) and everything looks great, great job 👍 

Just dont forget to add changesets for this. Docs regarding stuff like `ActorRefFrom` would also be helpful.

Issue #1112, opened 2020-04-03
Docs vs implementation: should `autoForward` also forward events to other services than just machines?


phryneas:
**Description**

The documentation states:

> autoForward - (optional) true if all events sent to this machine should also be sent (or forwarded) to the invoked child (false by default)

This wording implies that all children, also callbacks, should get those forwarded events. In reality, those are only forwarded to machines as the actor is only added to `forwardTo` in `spawnMachine`:

https://github.com/davidkpiano/xstate/blob/master/packages/core/src/interpreter.ts#L986-L988

This has cost me about 6 hours of wondering why my code doesn't work, pulling my hair and finally stepping through the code & reading the xstate source.

I'm not sure what the right behaviour would be, so this is either a bug report against the code or the documentation.

Personally, I'd love this to be possible.

sunndeath:
Is it relevant?

davidkpiano:
> This has cost me about 6 hours of wondering why my code doesn't work

Sorry about that; in the future, don't hesitate to go in the chat https://gitter.im/statecharts/statecharts and ask for help.

Yes, this currently only works for machines, but we can make an enhancement to make it work for callbacks too.

phryneas:
Yeah... sorry for my tone. I was quite frustrated at that point. 
Oh, and I was in the chat, but as it is with chats: some time everyone is asleep ;)

Personally, I'm not using this at the moment, but it would still very cool if you could add it :)

PR #1631, opened 2020-11-13
autoForward for machines only (see issue: #1112)

evangeloszotos:
I spend a couple of hours trying to get autoForwarding to work on a callback.
It works with forwardTo, though when just quickly trying to prototype a behavior, autoForward comes in handy.

So I thought I am doing something wrong, and keept trying, until I found issue #1112, where it is clearly stated by @davidkpiano that indeed, at the moment it works only for machines. 

So I thought it might be helpful to state it in the docs.
It might save some time of other developers.

Best
Evangelos

### DIFF
--- a/docs/guides/communication.md
+++ b/docs/guides/communication.md
@@ -31,7 +31,7 @@ An invocation is defined in a state node's configuration with the `invoke` prope
   - the invoked observable completes
 - `onError` - (optional) the transition to be taken when the invoked service encounters an execution error.
 - `autoForward` - (optional) `true` if all events sent to this machine should also be sent (or _forwarded_) to the invoked child (`false` by default)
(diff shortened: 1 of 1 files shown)


docs/guides/communication.md
```diff
@@ -31,7 +31,7 @@ An invocation is defined in a state node's configuration with the `invoke` prope
   - the invoked observable completes
 - `onError` - (optional) the transition to be taken when the invoked service encounters an execution error.
 - `autoForward` - (optional) `true` if all events sent to this machine should also be sent (or _forwarded_) to the invoked child (`false` by default)
```
davidkpiano, line 33:
```suggestion
- `autoForward` - (optional) `true` if all events sent to this machine should also be sent (or _forwarded_) to the invoked child machine (`false` by default)
```

evangeloszotos (commented):
Makes sense

PR #1643, opened 2020-11-20
[core] Ensure that imperatively stopped actors are cleaned up in parent.

### LINKED ISSUE #1642: Memory leak with stopped actor ?
**Description**
after stopping a spawned actor and clearing its reference in context, `state.children` still retains the actor

**Expected Result**
the actor should be cleared from `state.children`

**Reproduction**
https://codesandbox.io/s/priceless-allen-ezjnd

davidkpiano:
Fixes #1642


davidkpiano:
@Andarist Is this okay to merge?

Andarist:
We have already once declined a very similar change: https://github.com/davidkpiano/xstate/pull/1559 . Some reasons behind not implementing this were outlined in the comment here:
https://github.com/davidkpiano/xstate/pull/1559#discussion_r507977849

The additional concern would be that this doesn't trigger listeners so any code that aims to "react" to a child's existence would not get triggered (an example of this would be React's rendering).

PR #1644, opened 2020-11-20
Update TypeScript to 4.1

davidkpiano:
This PR updates TypeScript to the latest version, 4.1.2

For reference, this was done by doing the following in the root directory:
- `yarn add typescript -W`
- `yarn manypkg fix`

And then build and test:
- `yarn build`
- `yarn test`


.changeset/poor-pianos-marry.md
```diff
@@ -0,0 +1,16 @@
...
+'@xstate/inspect': patch
+'@xstate/react': patch
+'@xstate/scxml': patch
+'@xstate/test': patch
+'@xstate/viz-example': patch
+'@xstate/viz': patch
+'@xstate/vue': patch
+---
+
+TypeScript version updated to 4.1
```
Andarist, line 16: This changeset should rather be removed - the TS upgrade in our repo has no effect on our consumers (just yet).

Andarist (commented):
The upgrade itself is good and fine but one thing to consider is if we should start using new syntaxes in this repo. My take on this one would be that we shouldn't - at least for some time. By using new TS features we'll make new versions of XState unusable by many consumers that haven't had a chance to upgrade to TS@4.1 just yet.

davidkpiano:
> The upgrade itself is good and fine but one thing to consider is if we should start using new syntaxes in this repo. My take on this one would be that we shouldn't - at least for some time. By using new TS features we'll make new versions of XState unusable by many consumers that haven't had a chance to upgrade to TS@4.1 just yet.

Yeah, I'll wait a couple of minor versions to introduce TS 4.1 features.

PR #1663, opened 2020-11-26
[@xstate/graph] Add ability to resolve events from state

davidkpiano:
Options passed into graph functions (e.g., `getShortestPaths(machine, options)`) can now resolve `.events` based on the `state`:

```js
const countMachine = createMachine({
  initial: 'active',
  context: {
    count: 0
  },
  states: {
    active: {
      on: {
        ADD: {
          actions: assign({
            count: (context, event) => {
              return context.count + event.value;
            }
          })
        }
      }
    }
  }
});

const shortestPaths = getShortestPaths(countMachine, {
  events: {
    ADD: (state) => {
      // contrived example: if `context.count` is >= 10, increment by 10
      return state.context.count >= 10
        ? [{ type: 'ADD', value: 10 }]
        : [{ type: 'ADD', value: 1 }];
    }
  }
});

// The keys to the shortest paths will look like:
// "active" | { count: 0 }
// "active" | { count: 1 }
// "active" | { count: 2 }
// ...
// "active" | { count: 10 }
// "active" | { count: 20 }
// "active" | { count: 30 }
```

### DIFF
--- a/.changeset/poor-bugs-reflect.md
+++ b/.changeset/poor-bugs-reflect.md
@@ -1,0 +1,47 @@
+---
+'@xstate/graph': minor
+---
+
+Options passed into graph functions (e.g., `getShortestPaths(machine, options)`) can now resolve `.events` based on the `state`:
+
+```js
+const countMachine = createMachine({
+  initial: 'active',
+  context: {
+    count: 0
+  },
+  states: {
+    active: {
+      on: {
+        ADD: {
+          actions: assign({
+            count: (context, event) => {
+              return context.count + event.value;
+            }
+          })
+        }
+      }
+    }
+  }
+});
+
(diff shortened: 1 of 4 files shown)


.changeset/poor-bugs-reflect.md
```diff
@@ -0,0 +1,47 @@
+---
+'@xstate/graph': minor
+---
+
+Options passed into graph functions (e.g., `getShortestPaths(machine, options)`) can now resolve `.events` based on the `state`:
```
Andarist, line 5: Just a praise comment for Changesets idea 🙇 

Super useful for readers with a descriptive comment and example!

Andarist (approved):
Great addition 👍

PR #1664, opened 2020-11-27
Increased width for visualizer img in docs

sidiousvic:
Not sure if targeting the right base repo.

Small change. The visualizer image was hard to see at such a small resolution. 

(It's also outdated—tried to make a new one, but realized it might have to be hosted on designated Imgur account)

Great work with XState!


davidkpiano:
Closing as the visualizer images have been updated.

PR #1668, opened 2020-11-28
Provide a second overload for useActor

Andarist:


packages/xstate-react/test/types.test.tsx
```diff
@@ -171,10 +171,14 @@ describe('useMachine', () => {
 
     function Element({ myActor }: Props) {
       const [current, send] = useActor(myActor);
+      const bar: number = current.context.bar;
```
davidkpiano, line 174: Why is the `: number` assertion needed?

Andarist: Oh, its not really needed - i have wanted to check assignability of this (so kinda inference by extension) but it could always to be any so this check is not that great :/

PR #1669, opened 2020-11-29
[v5] Throw error if initial state is not specified. Fixes #674

### LINKED ISSUE #674: Always resolve state value to leaf nodes

davidkpiano:
Moving this from #1587

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -193,6 +193,16 @@ export class StateNode<
         )
       : EMPTY_OBJECT) as StateNodesConfig<TContext, TEvent, TStateSchema>;
 
+    if (this.type === 'compound' && !this.config.initial) {
+      throw new Error(
+        `No initial state specified for compound state node "#${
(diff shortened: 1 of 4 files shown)


packages/core/src/StateNode.ts
```diff
@@ -193,6 +193,16 @@ export class StateNode<
         )
       : EMPTY_OBJECT) as StateNodesConfig<TContext, TEvent, TStateSchema>;
 
+    if (this.type === 'compound' && !this.config.initial) {
```
Andarist, line 196: just a general q: what's your stance regarding dev-only **errors**? Would it be maybe worth it to remove this branch of code in production?

Andarist (approved):

packages/core/src/StateNode.ts
davidkpiano: Errors are something I'd love to have extracted instead, using a similar solution to: https://github.com/formium/tsdx#error-extraction

PR #1670, opened 2020-11-29
WIP: Add schema to machine config

davidkpiano:
This PR will allow you to define schema types inline (in the future, with additional metadata like JSON Schema that can be read at runtime):

```ts

interface LightStateSchema {
  states: {
    green: any;
    yellow: any;
    red: any;
  };
}

type LightStateEvents =
  | {
      type: 'TIMER';
    }
  | { type: 'POWER_OUTAGE' }
  | { type: 'FORBIDDEN_EVENT' }
  | { type: 'PED_COUNTDOWN' };

const lightMachine = createMachine({
  key: 'light',
  initial: 'green',
  schema: {
    states: createSchema<LightStateSchema>(),
    events: createSchema<LightStateEvents>()
  },
  states: {/* ... */}
});
```

This mitigates the need to declare the types as generics.


packages/core/src/schema.ts
```diff
@@ -0,0 +1,3 @@
+export function createSchema<TSchema>(): TSchema {
+  return (null as any) as TSchema;
```
Andarist, line 2: It would be more true if this would return `undefined` as this lands, for example, on the `.schema.states` which has a type of `T | undefined`. 

It still would not be type-safe at runtime though as this function returns `TSchema` and not `TSchema | undefined`. I've implemented the same thing as part of the https://github.com/davidkpiano/xstate/pull/1543 and I have used a slightly other direction there. "Schemas" are functions accepting a particular argument, see here: https://github.com/davidkpiano/xstate/blob/ab4d9c165d84de7d0669a012950dfe0acbe2dad5/packages/core/src/schemas.ts which makes it possible to leverage inference in a type-safe manner, see the types: https://github.com/davidkpiano/xstate/blob/ab4d9c165d84de7d0669a012950dfe0acbe2dad5/packages/core/src/types.ts#L761-L764

My approach doesn't allow for accessing the schema at runtime though and I suppose that we are aiming for that to facilitate better visualization~. But this one doesn't really allow for that either as this always returns `null`. Could we try to account for this use case as it seems to be the one that you very much like to see land in the end?

Maybe something like this:
```ts
export function createSchema<TSchema>(jsonSchema: any): (item: TSchema) => any  {
  const validator = (item: TSchema) => {}
  validator.$schema = jsonSchema
  return validator
}
```

From what I understand there is also a plan to opt-in to some validations in the future. Do you have an idea how this would work? How one would opt-into them and what would happen on a failed validation?

davidkpiano: Added an example to c2b7e15 - in my view, a schema is not a validation function, but rather one that can be used for validation. It should be completely serializable:

```js
// ...
  schema: {
    events: createSchema<SomeEvents>({
      anyOf: [
        { type: 'object', properties: {/* ... */} },
        { type: 'object', properties: {/* ... */} },
        // ...
      ]
    })
  }
});
```

In the above example, JSON Schema for the events is passed in (this doesn't have to be JSON Schema, it can be anything, including a function), and this is saved on `machine.schema.events` as-is. (Ideally for TS, we'd use opaque/nominal/??? types for this, since what's passed in and returned is obviously not `SomeEvents` but it _represents_ `SomeEvents`)

Then, it can be used in whichever way the developer wants, for example:

```js
// using json-schema-faker
function sendSampleEvent(service) {
  const sampleEvent = jsf.generate(service.machine.schemas.events);

  service.send(sampleEvent);
}
```

Andarist: Ok, let's roll with that - if you think it's worth it then we could use opaque types for this but I also don't care that much in the end.

packages/core/src/types.ts
```diff
@@ -705,6 +713,8 @@ export interface HistoryStateNode<TContext> extends StateNode<TContext> {
   target: StateValue | undefined;
 }
 
+export type SchemaValidator<T> = (value: any) => value is T;
```
Andarist, line 716: is this used anywhere?

packages/core/src/types.ts
```diff
@@ -674,6 +674,14 @@ export interface MachineConfig<
    * The initial context (extended state)
    */
   context?: TContext | (() => TContext);
+  /**
+   * Describes the events that this machine accepts.
+   *
+   * **NOTE:** This is TypeScript-only.;
+   */
+  schema?: {
```
Andarist, line 682: wasn't `states` also allowed here before?

packages/core/src/schema.ts
```diff
@@ -0,0 +1,3 @@
+export function createSchema<TSchema>(value?: any): TSchema {
+  return (value as any) as TSchema;
```
Andarist, line 2: `value` is already `any`, so you can just return it here

```suggestion
  return value;
```

packages/core/test/schema.test.ts
```diff
@@ -0,0 +1,76 @@
+import { createMachine, EventObject } from '../src';
+import * as Ajv from 'ajv';
+
+describe('event schema', () => {
+  it('an event schema can validate incoming events', () => {
```
Andarist, line 5: 👍

packages/core/test/types.test.ts
```diff
@@ -34,59 +35,122 @@ describe('StateSchema', () => {
...
       }
-    }
+    });
+
+    noop(lightMachine);
+    expect(true).toBeTruthy();
   });
 
-  noop(lightMachine);
+  it('should work with a StateSchema defined inline', () => {
```
Andarist, line 96: This test only tests the event schema so maybe a more precise title would be better here?
```suggestion
  it('should work with an event schema defined inline', () => {
```

packages/core/src/types.ts
davidkpiano: It isn't; I used it when experimenting with the other (validation function) approach 😅 

We can probably make use of this in future work if we do some of the "design by contract" stuff, but I'll delete it

packages/core/src/types.ts
davidkpiano: Yes, but because of the overloads for the `Machine()` function, it didn't make a difference - the `TStateSchema` was still being resolved (inferred?) as `any` and the only way to "fix" that was to change this:

```diff
export function Machine<
  TContext = DefaultContext,
- TStateSchema extends StateSchema = any,
+ TStateSchema extends StateSchema = unknown,
  TEvent extends EventObject = AnyEventObject
>(
  config: MachineConfig<TContext, TStateSchema, TEvent>,
  options?: Partial<MachineOptions<TContext, TEvent>>,
  initialContext?: TContext
): StateMachine<TContext, TStateSchema, TEvent>;
```

...which caused a lot of other TypeScript issues.

I want to move away from `TStateSchema` in general as well (bigger topic), as I feel like this is something that should be either handled by better TS inference of the config object, or by something like `xstate-codegen` 🤔

PR #1676, opened 2020-12-02
Use always keyword instead of empty string

wildskyf:
ref #1249


davidkpiano:
Can you change this in `events.md` instead? This is auto-generated.

PR #1692, opened 2020-12-05
Fix history state entering after transient transitions

Andarist:
fixes #1691 

Updating `historyValue` is done within [`exitStates`](https://www.w3.org/TR/scxml/#exitStates) procedure which is part of the [`microstep`](https://www.w3.org/TR/scxml/#microstep) procedure so this should always keep the very latest resolved values and nothing should be "preserved" from older states.

### DIFF
--- a/packages/core/test/history.test.ts
+++ b/packages/core/test/history.test.ts
@@ -1,4 +1,4 @@
-import { Machine, createMachine } from '../src/index';
+import { Machine, createMachine, interpret } from '../src/index';
 
 describe('history states', () => {
   const historyMachine = createMachine({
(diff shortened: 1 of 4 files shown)


packages/core/test/history.test.ts
```diff
@@ -71,6 +71,55 @@ describe('history states', () => {
     const nextState = historyMachine.transition(onState, 'H_POWER');
     expect(nextState.history!.history).not.toBeDefined();
   });
+
+  it.only('should go to the most recently visited state by a transient transition', () => {
```
davidkpiano, line 75:
```suggestion
  it('should go to the most recently visited state by a transient transition', () => {
```

davidkpiano (changes requested):

packages/core/test/history.test.ts
Andarist: 🤦 fixed!

PR #1699, opened 2020-12-06
[@xstate/inspect] Handle circular references gracefully in inspector.

### LINKED ISSUE #1497: The visualizer breaks on JSON.stringify calls when there are circular references in machine definition
**Description**
Quite often we have a hierarchy of state machines (e.g. parent-child relationship) where by the nature of definition there are circular references. In such cases, the call to `JSON.stringify` throws an error. 

**Expected Result**
The visualizer is able to visualize parent-child state machine relationship.

**Actual Result**
An error

**Suggestion**
Replace the `JSON.stringify` calls with a safe alternative e.g. [fast-safe-stringify](https://www.npmjs.com/package/fast-safe-stringify)

davidkpiano:
This PR fixes stringifying circular reference in `@xstate/inspect` by using [fast-safe-stringify](https://www.npmjs.com/package/fast-safe-stringify) when normal `JSON.stringify(...)` throws.

Currently, this will result in the string `"[Circular]"` in the inspect UI, which is much better than completely failing (previous behavior):

CleanShot 2020-12-06 at 17 52 45@2x

Fixes #1497

### DIFF
--- a/packages/xstate-inspect/src/index.ts
+++ b/packages/xstate-inspect/src/index.ts
@@ -87,7 +87,7 @@ export const createInspectMachine = (
               const { event } = e;
               const scxmlEventObject = JSON.parse(event) as SCXML.Event<any>;
               const service = serviceMap.get(scxmlEventObject.origin!);
-              service?.send(JSON.parse(event));
+              service?.send(scxmlEventObject);
             }
           },
           unload: {
--- a/packages/xstate-inspect/src/index.ts
+++ b/packages/xstate-inspect/src/index.ts
@@ -7,7 +7,9 @@ import {
   EventObject,
   EventData
 } from 'xstate';
+import { XStateDevInterface } from 'xstate/lib/devTools';
 import { toSCXMLEvent, toEventObject } from 'xstate/lib/utils';
+import safeStringify from 'fast-safe-stringify';
 
 export type ServiceListener = (service: Interpreter<any>) => void;
 
@@ -16,30 +18,16 @@ type MaybeLazy<T> = T | (() => T);
 export interface InspectorOptions {
   url: string;
   iframe: MaybeLazy<HTMLIFrameElement | null | false>;
-}
-
-declare global {
-  interface Window {
-    __xstate__: {
-      register: (service: Interpreter<any>) => void;
-      unregister: (service: Interpreter<any>) => void;
-      onRegister: (
-        listener: ServiceListener
-      ) => {
-        unsubscribe: () => void;
-      };
-      services: Set<Interpreter<any>>;
-    };
-  }
+  devTools: MaybeLazy<XStateDevInterface>;
 }
 
 const serviceMap = new Map<string, Interpreter<any>>();
 
-export function createDevTools() {
+export function createDevTools(): XStateDevInterface {
   const services = new Set<Interpreter<any>>();
   const serviceListeners = new Set<ServiceListener>();
 
-  window.__xstate__ = {
+  return {
     services,
     register: (service) => {
       services.add(service);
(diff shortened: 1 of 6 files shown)


packages/xstate-inspect/src/index.ts
```diff
@@ -68,80 +56,109 @@ export function createDevTools() {
...
-            service?.send(JSON.parse(event));
-          }
-        },
-        unload: {
-          actions: (ctx) => {
-            ctx.client!.send({ type: 'xstate.disconnect' });
-          }
-        },
-        disconnect: 'pendingConnection'
+export const createInspectMachine = (
```
Andarist, line 59: q: should this be a public export?

packages/xstate-inspect/src/index.ts
```diff
@@ -68,80 +56,109 @@ export function createDevTools() {
...
+  devTools: XStateDevInterface = globalThis.__xstate__
```
Andarist, line 60: note: `globalThis` might not be available in all envs but I guess for the inspect package this might be OK

packages/xstate-inspect/src/index.ts
```diff
@@ -68,80 +56,109 @@ export function createDevTools() {
...
+          },
+          'service.stop': {
+            actions: (ctx, e) => ctx.client!.send(e)
+          },
+          'xstate.event': {
+            actions: (_, e) => {
+              const { event } = e;
+              const scxmlEventObject = JSON.parse(event) as SCXML.Event<any>;
+              const service = serviceMap.get(scxmlEventObject.origin!);
+              service?.send(JSON.parse(event));
```
Andarist, line 90:
```suggestion
              service?.send(scxmlEventObject);
```

packages/xstate-inspect/src/index.ts
```diff
@@ -68,80 +56,109 @@ export function createDevTools() {
...
+    globalThis.__xstate__ = devTools;
+    return devTools;
+  }
 };
 
+function getLazy<T>(value: MaybeLazy<T>): T {
+  return typeof value === 'function' ? (value as () => T)() : value;
+}
+
+function stringify(value: any): string {
```
Andarist, line 140: Couldn't we just use `safeStringify` directly?

Andarist (approved):

packages/xstate-inspect/test/inspect.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      context: circularStructure,
+      states: {
+        active: {}
+      }
+    });
+
+    const devTools = createDevTools();
+
+    devTools.onRegister(() => {
+      done();
```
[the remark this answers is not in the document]
davidkpiano: In the future, register/onRegister may be async, and if that happens, this test will fail.

packages/xstate-inspect/src/index.ts
davidkpiano: I don't think it should; it's an internal implementation.

packages/xstate-inspect/src/index.ts
davidkpiano: Good call; for those other environments, the expectation is for developers to pass in a custom reference to `devTools` instead of this one (I'll add this to the docs once I get the Node + inspect implementation working)

packages/xstate-inspect/src/index.ts
davidkpiano: This is a [recommendation from the `fast-safe-stringify`](https://www.npmjs.com/package/fast-safe-stringify#protip) library itself:

1. Try normal `JSON.stringify`
2. Only if that fails, try `safeStringify`

Andarist: Ok, havent realized that 👍

PR #1704, opened 2020-12-07
Preserve global context with `clock` methods (`setTimeout`/`clearTimeout`)

### LINKED ISSUE #1703: Interpretter clock.setTimeout method does not work in Firefox WebExtension content scripts by default
**Description**

When an XState machine with [delayed events/transitions](https://xstate.js.org/docs/guides/delays.html#delayed-transitions) is used in a [Firefox WebExtension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) [content script](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts), the default implementation of `clock.setTimeout` and `clock.clearTimeout` throws this exception:

```console
TypeError: 'setTimeout' called on an object that does not implement interface Window.
```

**Expected Result**

I expected the machine, as defined, to work without throwing an exception.

**Actual Result**

An exception is thrown, and the state machine does not work properly.

**Reproduction**

Because this happens in a Firefox WebExtension, I couldn't use any of the recommended reproduction templates. Instead, I created a sample repository to show the issue: https://github.com/blimmer/xstate-webext-setTimeout-issue. There is information in the [Quick Start section of the README](https://github.com/blimmer/xstate-webext-setTimeout-issue#quick-start) that explains how to run the example.

**Additional context**

This issue seems to be caused by the default implementation of `setTimeout`/`clearTimeout` defined here: https://github.com/davidkpiano/xstate/blob/844c2c30d6160763ab164dc0592e33a942746361/packages/core/src/interpreter.ts#L125-L132

I believe the issue is with `thisArg` being `null`. I locally changed the default implementation to look like this (note the `global` `thisArg`) and the state machine started working with the default implementation.

```js
clock: {
  setTimeout: (fn, ms) => {
    return global.setTimeout.call(global, fn, ms);
  },
  clearTimeout: (id) => {
    return global.clearTimeout.call(global, id);
  }
},
```

I was looking at the history and noticed that @davidkpiano change the default `clock` behavior in [this commit](https://github.com/davidkpiano/xstate/commit/66c8cf2b20c11863fd388295a8ab382615d1fb4b) to use the `.call(null, ...args)` pattern. This makes sense since you want `setTimeout` bound to the `global` instance passed in, but I think Firefox is doing some additional checks on the `thisArg` which causes this error.

blimmer:
Fixes #1703

### DIFF
--- a/.changeset/tall-apes-tickle.md
+++ b/.changeset/tall-apes-tickle.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The `clock` methods (`setTimeout`, `clearTimeout`) are now properly bound to the `global` scope internally (fixes #1703)
--- a/.changeset/tall-apes-tickle.md
+++ b/.changeset/tall-apes-tickle.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-The `clock` methods (`setTimeout`, `clearTimeout`) are now properly bound to the `global` scope internally (fixes #1703)
+The default `clock` methods (`setTimeout` and `clearTimeout`) are now invoked properly with the global context preserved for those invocations which matter for some JS environments. More details can be found in the corresponding issue: [#1703](https://github.com/davidkpiano/xstate/issues/1703).
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -124,10 +124,10 @@ export class Interpreter<
     deferEvents: true,
     clock: {
       setTimeout: (fn, ms) => {
-        return global.setTimeout.call(null, fn, ms);
+        return global.setTimeout.call(global, fn, ms);
       },
       clearTimeout: (id) => {
-        return global.clearTimeout.call(null, id);
+        return global.clearTimeout.call(global, id);
       }
     },
     logger: global.console.log.bind(console),
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -124,7 +124,7 @@ export class Interpreter<
     deferEvents: true,
     clock: {
       setTimeout: (fn, ms) => {
-        return global.setTimeout.call(global, fn, ms);
+        return setTimeout(fn, ms);
       },
       clearTimeout: (id) => {
         return global.clearTimeout.call(global, id);
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -127,7 +127,7 @@ export class Interpreter<
         return setTimeout(fn, ms);
       },
       clearTimeout: (id) => {
-        return global.clearTimeout.call(global, id);
+        return clearTimeout(id);
       }
     },
     logger: global.console.log.bind(console),


blimmer:
Whoops, adding changeset now.

.changeset/tall-apes-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+@blimmer Bind `clock` methods to `global` `thisArg`. Fixes #1703.
```
davidkpiano, line 5:
```suggestion
The `clock` methods (`setTimeout`, `clearTimeout`) are now properly bound to the `global` scope internally (fixes #1703)
```

blimmer:
By the way, I just confirmed this fix by using [`patch-package`](https://github.com/ds300/patch-package) with the following patchfile for my WebExtension project:

```patch
diff --git a/node_modules/xstate/es/interpreter.js b/node_modules/xstate/es/interpreter.js
index a35ec06..8c7fedf 100644
--- a/node_modules/xstate/es/interpreter.js
+++ b/node_modules/xstate/es/interpreter.js
@@ -1237,10 +1237,10 @@ function () {
       deferEvents: true,
       clock: {
         setTimeout: function (fn, ms) {
-          return global.setTimeout.call(null, fn, ms);
+          return global.setTimeout.call(global, fn, ms);
         },
         clearTimeout: function (id) {
-          return global.clearTimeout.call(null, id);
+          return global.clearTimeout.call(global, id);
         }
       },
       logger: global.console.log.bind(console),
diff --git a/node_modules/xstate/lib/interpreter.js b/node_modules/xstate/lib/interpreter.js
index 8216933..64213c7 100644
--- a/node_modules/xstate/lib/interpreter.js
+++ b/node_modules/xstate/lib/interpreter.js
@@ -1039,10 +1039,10 @@ var Interpreter = /** @class */ (function () {
         deferEvents: true,
         clock: {
             setTimeout: function (fn, ms) {
-                return global.setTimeout.call(null, fn, ms);
+                return global.setTimeout.call(global, fn, ms);
             },
             clearTimeout: function (id) {
-                return global.clearTimeout.call(null, id);
+                return global.clearTimeout.call(global, id);
             }
         },
         logger: global.console.log.bind(console),

```

Firefox now works and Chrome still works with the change.

packages/core/src/interpreter.ts
```diff
@@ -124,10 +124,10 @@ export class Interpreter<
     deferEvents: true,
     clock: {
       setTimeout: (fn, ms) => {
-        return global.setTimeout.call(null, fn, ms);
+        return global.setTimeout.call(global, fn, ms);
```
Andarist, line 127:
```suggestion
        return setTimeout(fn, ms);
```

packages/core/src/interpreter.ts
```diff
@@ -124,10 +124,10 @@ export class Interpreter<
...
       },
       clearTimeout: (id) => {
-        return global.clearTimeout.call(null, id);
+        return global.clearTimeout.call(global, id);
```
Andarist, line 130:
```suggestion
        return clearTimeout(id);
```

Andarist (commented):
The problem fixed by using `global.` etc was probably very similar to what you have experienced - before the fix the `this` has been changed because `setTimeout` and `clearTimeout` were attached to another object and they were used from there (oh, our beautiful JS). See the previous version here: https://github.com/davidkpiano/xstate/commit/66c8cf2b20c11863fd388295a8ab382615d1fb4b#diff-5b8880ef88d507d13e4d97365bcfbafe3a9d36a16655ef655f51c1c38ab01bb7L108

We don't do any hoops though with global, call etc if we just invoke them using the global context (like in my suggested changes). This way they will be always looked up on a global object and the invocation context will be correct.

If RxJS is using those timers as is (without hoops) I bet that we can do that as well 😉 : 
https://github.com/ReactiveX/rxjs/blob/cfa267bc0916ede09c8b14aedcdb69a791055fb6/src/internal/scheduler/timeoutProvider.ts#L21

blimmer:
I'm happy to do this whichever way is most preferable. @davidkpiano, what do you think?

Andarist:
Could you try the proposed changes with `patch-package` on your project?

davidkpiano:
> I'm happy to do this whichever way is most preferable. @davidkpiano, what do you think?

Let's do the changes and try them out. If your use-case still passes, it's probably fine 👍

.changeset/tall-apes-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The `clock` methods (`setTimeout`, `clearTimeout`) are now properly bound to the `global` scope internally (fixes #1703)
```
Andarist, line 5:
```suggestion
The default `clock` methods (`setTimeout` and `clearTimeout`) are now invoked properly with the global context preserved for those invocations which matter for some JS environments. More details can be found in the corresponding issue: [#1703](https://github.com/davidkpiano/xstate/issues/1703).
```

blimmer:
I tested with @Andarist 's patch as well (below), and both work. I don't feel strongly which one gets merged, so I'll defer to you 👍 

```
diff --git a/node_modules/xstate/es/interpreter.js b/node_modules/xstate/es/interpreter.js
index a35ec06..d2e2abb 100644
--- a/node_modules/xstate/es/interpreter.js
+++ b/node_modules/xstate/es/interpreter.js
@@ -1237,10 +1237,10 @@ function () {
       deferEvents: true,
       clock: {
         setTimeout: function (fn, ms) {
-          return global.setTimeout.call(null, fn, ms);
+          return setTimeout(fn, ms);
         },
         clearTimeout: function (id) {
-          return global.clearTimeout.call(null, id);
+          return clearTimeout(id);
         }
       },
       logger: global.console.log.bind(console),
diff --git a/node_modules/xstate/lib/interpreter.js b/node_modules/xstate/lib/interpreter.js
index 8216933..5f8be29 100644
--- a/node_modules/xstate/lib/interpreter.js
+++ b/node_modules/xstate/lib/interpreter.js
@@ -1039,10 +1039,10 @@ var Interpreter = /** @class */ (function () {
         deferEvents: true,
         clock: {
             setTimeout: function (fn, ms) {
-                return global.setTimeout.call(null, fn, ms);
+                return setTimeout(fn, ms);
             },
             clearTimeout: function (id) {
-                return global.clearTimeout.call(null, id);
+                return clearTimeout(id);
             }
         },
         logger: global.console.log.bind(console),

```

davidkpiano (approved):

PR #1710, opened 2020-12-09
Do not crash when stopping already stopped interpreter. 

### LINKED ISSUE #1697: TypeError: Cannot read property 'configuration of undefined' when calling `stop` on an interpreter that was not initialized
**Description**
Recently, my tests are failing due to a crash in xstate. It seems that this happens because the interpreter was not initialized (see repro below).
```
Summary of all failing tests
 FAIL  packages/zwave-js/src/lib/driver/Driver.test.ts (6.342 s)
  ● lib/driver/Driver =>  › sending messages =>  › should not be possible if the driver failed to start

    TypeError: Cannot read property 'configuration' of undefined

      1222 | 
      1223 |            // First stop the send thread machine and close the serial port, so nothing happens anymore
    > 1224 |            this.sendThread.stop();
           |                            ^
      1225 |            if (this.serial != undefined) {
      1226 |                    if (this.serial.isOpen) await this.serial.close();
      1227 |                    this.serial = undefined;

      at Interpreter.stop (node_modules/xstate/lib/interpreter.js:487:20)
      at Driver.destroy (packages/zwave-js/src/lib/driver/Driver.ts:1224:19)
      at Object.<anonymous> (packages/zwave-js/src/lib/driver/Driver.test.ts:140:11)
```

**Expected Result**
No crash

**Actual Result**
crash

**Reproduction**
1. clone `https://github.com/zwave-js/node-zwave-js/commit/18a000888b12143409569241e0128aac9f476499`
2. run `npm run test:ts -- --testPathPattern=driver/Driver.test.ts`

```
$ npm run test:ts -- --testPathPattern=driver/Driver.test.ts

> test:ts
> jest "--testPathPattern=driver/Driver.test.ts"

  console.warn
    Warning: Attempted to read state from uninitialized service 'SendThread'. Make sure the service is started first.

      1222 | 
      1223 |            // First stop the send thread machine and close the serial port, so nothing happens anymore
    > 1224 |            this.sendThread.stop();
           |                            ^
      1225 |            if (this.serial != undefined) {
      1226 |                    if (this.serial.isOpen) await this.serial.close();
      1227 |                    this.serial = undefined;

[...]

  ● lib/driver/Driver =>  › sending messages =>  › should not be possible if the driver failed to start

    TypeError: Cannot read property 'configuration' of undefined

      1222 | 
      1223 |            // First stop the send thread machine and close the serial port, so nothing happens anymore
    > 1224 |            this.sendThread.stop();
           |                            ^
      1225 |            if (this.serial != undefined) {
      1226 |                    if (this.serial.isOpen) await this.serial.close();
      1227 |                    this.serial = undefined;

      at Interpreter.stop (node_modules/xstate/lib/interpreter.js:487:20)
      at Driver.destroy (packages/zwave-js/src/lib/driver/Driver.ts:1224:19)
      at Object.<anonymous> (packages/zwave-js/src/lib/driver/Driver.test.ts:140:11)

Test Suites: 1 failed, 1 total
```

**Additional context**

davidkpiano:
Fixes #1697

### DIFF
--- a/.changeset/clean-walls-breathe.md
+++ b/.changeset/clean-walls-breathe.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Stopping an already stopped interpreter will no longer crash. See #1697 for details.
+Stopping an already stopped interpreter will no longer crash. See [#1697](https://github.com/davidkpiano/xstate/issues/1697) for details.
--- a/.changeset/clean-walls-breathe.md
+++ b/.changeset/clean-walls-breathe.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Stopping an already stopped interpreter will no longer crash. See #1697 for details.
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -503,6 +503,11 @@ export class Interpreter<
       this.doneListeners.delete(listener);
     }
 
+    if (!this.initialized) {
+      // Interpreter already stopped; do nothing
+      return this;
+    }
+
     this.state.configuration.forEach((stateNode) => {
       for (const action of stateNode.definition.exit) {
         this.exec(action, this.state);
--- a/packages/core/test/interpreter.test.ts
+++ b/packages/core/test/interpreter.test.ts
@@ -1388,6 +1388,19 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
         done();
       }, 10);
     });
+
+    it('stopping a not-started interpreter should not crash', () => {
+      const service = interpret(
+        createMachine({
+          initial: 'a',
+          states: { a: {} }
+        })
+      );
+
+      expect(() => {
+        service.stop();
+      }).not.toThrow();
+    });
   });
 
   describe('off()', () => {


.changeset/clean-walls-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Stopping an already stopped interpreter will no longer crash. See #1697 for details.
```
Andarist, line 5:
```suggestion
Stopping an already stopped interpreter will no longer crash. See [#1697](https://github.com/davidkpiano/xstate/issues/1697) for details.
```

packages/core/src/interpreter.ts
```diff
@@ -503,6 +503,11 @@ export class Interpreter<
       this.doneListeners.delete(listener);
     }
 
+    if (!this.initialized) {
+      // Interpreter already stopped; do nothing
+      return this;
```
Andarist, line 508: Shouldn't we throw an error that just would lead users to fix their mistake? isn't stopping twice breaking an invariant rule of XState? I see arguments both ways (throwing vs making this noop) and I'm somewhat undecided what's better myself - just raising this as a general question.

davidkpiano: There can exist use-cases where the not-started interpreter is stopped by no fault of the developer; throwing would be counter-productive and unexpected.

Contrived example: in React, a component that is rendered but immediately unmounted (for some reason) would need to `.stop()` all services, even if they haven't started yet.

I think the best thing to do is to no-op, since the intention is to stop the service, and if the service is already stopped, mission accomplished 👍

Andarist: > Contrived example: in React, a component that is rendered but immediately unmounted (for some reason) would need to .stop() all services, even if they haven't started yet.

This actually can't happen - if we don't reach the commit phase we won't `start` anything **and** unmounting just wont happen so `stop` won't be called either.

Andarist (approved):

PR #1715, opened 2020-12-10
StateNode config as HistoryStateNodeConfig too

Denys-Bushulyak:
And ignore jetBrains IDE files


davidkpiano:
Can you share the code that was breaking (not compiling with TS) without this change?

PR #1731, opened 2020-12-16
Finite State Transducers

davidkpiano:
This PR introduces a generalized [finite-state transducer](https://en.wikipedia.org/wiki/Finite-state_transducer) interface.

Despite the scary-sounding name, a finite-state transducer (FST) is just a finite-state machine/automaton (FSA) with "output". The statecharts created with XState are already FSTs, as the `state.actions` are considered its "output". In general, FSAs are useful for "recognizing", whereas FSTs are useful for "parsing", and we can consider machines/statecharts used in applications to "parse" a sequence of events (the input) to actions that should be executed as a result (the output).

The reason for introducing FST types and utility functions in this PR is to make XState and its tooling more widely applicable with other libraries. For example, this PR demonstrates creating an FST to be used with `@xstate/graph` to generate the shortest paths without using a machine created from XState.

Other applications include converting normal reducers (e.g., from Redux) to FSTs that can be used with XState tools:

```js
import { fromReducer } from 'xstate/lib/fst';
import { getShortestPaths } from '@xstate/graph';

function someReducer(state, event) {
  // ...
}

const initialState = {/* ... */};

const someFST = fromReducer(someReducer, initialState);

const shortestPaths = getShortestPaths(someFST);
```


packages/core/src/fst/actor.ts
```diff
@@ -0,0 +1,36 @@
+import { FST } from '.';
+import { ActorRef, EventObject, Observer } from '..';
+import { toObserver } from '../utils';
+
+export function toActorFST<TState, TEvent extends EventObject, TOutput = any>(
```
Andarist, line 5: `toActorFST` sounds like a function to convert to some kind of an FST actor, or converting the actor to an FST whereas this converts FST to an actor, maybe `fstToActor` would fit this better? or just `toActor` in the FST-related package/directory? it could be used like this:
```
import * as FST from '...'
FST.toActor(/**/)
```

davidkpiano: Good call, will rename.

tomByrer:
I'm trying to wrap my head around what is a 'finite-state transducer'.

Basically a way to wrap 'normal' JS, to transform the output into something that is XState compatible?

Andarist:
> Basically a way to wrap 'normal' JS, to transform the output into something that is XState compatible?

Somewhat - but not into something that is fully compatible with XState (if we think about statecharts etc). 

A transducer is a function that determines next state and effects for a given state and an event. So you can wrap "normal" JS (that is pure) into it so another helper can "query" it and based on that it kinda can build a graph of possible transitions - because he can query it with what it has received from it and known events until it reaches a particular state that we are interested in.

packages/core/src/fst/actor.ts
```diff
@@ -0,0 +1,38 @@
+import { FST } from '.';
+import { EventObject, Observer, SpawnedActorRef } from '..';
+import { toObserver } from '../utils';
+
+export function toActor<TState, TEvent extends EventObject, TOutput = any>(
+  fst: FST<TState, TEvent, TOutput>,
+  handleOutput?: (output: TOutput) => void
+): SpawnedActorRef<TEvent, TState> {
+  let currentState = fst.initialState;
+  const observers = new Set<Observer<TState>>();
```
Andarist, line 10: Using Set here is slightly incorrect because each subscription should be properly isolated. In here, when using `subscribe` with **the same** object observer, we won't notify that observer twice when an event is sent to this actor, and also unsubscribing once (for that object observer) will remove that single subscription:
```js
let called = 0
const observer = { next() { called++ } }
const actor = toActor(() => {})
actor.subscribe(observer)
const { unsubscribe } = actor.subscribe(observer)
actor.send({ type: 'FOO' })
expect(called).toBe(2) // actual: 1
unsubscribe()
actor.send({ type: 'FOO' })
expect(called).toBe(3) // actual: 1
```

packages/core/src/fst/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { State, StateMachine } from '..';
+
+export { toActor as fromActor } from './actor';
+export { fromMachine } from './machine';
+export { fromReducer } from './reducer';
+
+export type FSTTransition<TState, TEvent, TOutput> = (
+  state: TState,
+  input: TEvent
+) => [TState] | [TState, TOutput];
```
Andarist, line 12: This return type won't play nicely with the consuming side of things. This code snippet should be valid:
```js
type FSTTransition<TState, TEvent, TOutput> = (
  state: TState,
  input: TEvent
) => [TState] | [TState, TOutput];

type MyState = { count: number }
type MyEvent = { type: 'INC' }
type MyOutput = { type: 'alert' }

declare const transition: FSTTransition<{ count: number }, MyEvent, MyOutput>
declare const handleOutput: (output: MyOutput) => void

let [state, output] = transition({ count: 0 }, { type: "INC" })
handleOutput(output) // oops, error since output is of type `MyOutput | undefined`
```

Unless this is the intention but then the return type of `[TState, TOutput | undefined]` would be more explicit about it.

packages/core/test/fst.test.ts
```diff
@@ -0,0 +1,13 @@
+import { fromReducer } from '../src/fst';
+
+describe('fst', () => {
```
Andarist, line 3: could u also add some tests for barebone fst interface? as well some example that would utilize `handleOutput`?

packages/core/src/fst/actor.ts
davidkpiano: Is there prior art to this? I thought that observers need to be unique.

Andarist: This is how RxJS (among others) handle this sort of a situation: https://codesandbox.io/s/sharp-https-wzk76?file=/src/index.js:191-238

Obervers~ are unique internally so the same external observer creates separate subscriptions. Internally, RxJS converts subscribers to "safe subscribers": https://github.com/ReactiveX/rxjs/blob/5de9e1b355234158ea4320eb4ffd08cbee988c32/src/internal/Observable.ts#L217 and in there all original functions from the observer object are bound https://github.com/ReactiveX/rxjs/blob/5de9e1b355234158ea4320eb4ffd08cbee988c32/src/internal/Subscriber.ts#L168-L170 which returns new references.

To fix that you could just replace this line:
https://github.com/davidkpiano/xstate/blob/a4256dcdbb79a6c19412f01991bff49751d661ca/packages/core/src/utils.ts#L694
with:
```js
    return { next: nextHandler.next, error: nextHandler.error, complete: nextHandler.complete };
```

PR #1756, opened 2020-12-20
Patch: ReferenceError: process is not defined

dimitardanailov:
package: `@xstate/react`
Parent pull request: https://github.com/davidkpiano/xstate/pull/1738

Issue: 
If someone import: https://unpkg.com/@xstate/react@1.2.0/dist/xstate-react.umd.min.js. The import throws the following error: `ReferenceError: process is not defined`

The current patch uses `rollup-replace` to remove the unnecessary variable: `process`.

### DIFF
--- a/.changeset/pretty-hotels-rush.md
+++ b/.changeset/pretty-hotels-rush.md
@@ -1,0 +1,5 @@
+---
+'@xstate/react': minor
+---
+
+'rollup-replace' replaces process.env.NODE_ENV declarations.
--- a/.changeset/pretty-hotels-rush.md
+++ b/.changeset/pretty-hotels-rush.md
@@ -1,5 +1,5 @@
 ---
-'@xstate/react': minor
+'@xstate/react': patch
 ---
 
-'rollup-replace' replaces process.env.NODE_ENV declarations.
+Fixed an issue with `process` references not being removed correctly from the UMD bundles.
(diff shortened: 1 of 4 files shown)


davidkpiano (approved):

Andarist:
@dimitardanailov thanks for the PR! Please note that I've force-pushed to your branch so it would be best if you just reset to our current `master`

dimitardanailov:
@Andarist 

I wish you a happy holiday period. 
I'd like to share with you a new issue: The issue is: 

`xstate-react.umd.min.js:16 Uncaught TypeError: n is not a function`

Screen Shot 2020-12-26 at 9 46 28 PM

1) 
What's your opinion if library supports `Unminified` version ?

2)
I've been working since 21:30 local time to understand why the issue occurs.

dimitardanailov:
I'd like to share my personal thoughts: 

The error behind error is a missing dependency. The missing dependency is: `use-isomorphic-layout-effect`(https://github.com/Andarist/use-isomorphic-layout-effect). I can't find a suitable browser version.

Andarist:
> What's your opinion if library supports Unminified version ? My personal opinion is: The full version or source map version will be useful if umd.min.js has an issue.

source maps could be generated for `.min.js` files

> The error behind error is a missing dependency. The missing dependency is: use-isomorphic-layout-effect(https://github.com/Andarist/use-isomorphic-layout-effect). I can't find a suitable browser version.

From what I've checked the culprit is not the `use-isomorphic-layout-effect` but the `use-subscription`. I've fixed the problem locally but it's late already and gonna push out the fix tomorrow.

PR #1762, opened 2020-12-21
Update description in svelte.md

jonasholbech:
Add import satement for 'rollup-plugin-replace';


jonasholbech:
Not the correct version of 'rollup-plugin-replace'; closing

PR #1782, opened 2020-12-27
[v5] More SCXML tests

davidkpiano:
- Added `forEach()` action creator and enabled relevant tests


packages/core/src/actionTypes.ts
```diff
@@ -18,3 +18,4 @@ export const error = ActionTypes.ErrorCustom;
 export const update = ActionTypes.Update;
 export const choose = ActionTypes.Choose;
 export const pure = ActionTypes.Pure;
+export const forEach = ActionTypes.ForEach;
```
Andarist, line 21: if the actions creator is named `each` it would probably be best to reflect that here as well so we use the unified terminology in the codebase, docs, etc

packages/core/src/actions/resolveActionsAndContext.ts
```diff
@@ -0,0 +1,210 @@
...
+
+          if (isArray(array)) {
+            array.forEach((item, index) => {
+              const [nextContext] = updateContext(
+                context,
+                _event,
+                [
+                  assign<TContext, TEvent>({
+                    [action.item]: item as any,
+                    [action.index]: index as any // TODO: fix
```
Andarist, line 176: Index is optional in SCXML and in here you would add an extra property under the undefined key which would raise eyebrows when people would see that while debugging

Andarist (commented):
Im not super fond of some things in this PR - but its really more of a problem with SCXML spec than with the provided implementation.

IMHO assigning item and index to context is weird and surprising for a modern application. The problem is that those variables outlive the `each` execution and this is rather leaky (it literally might lead to retaining unwanted objects). I guess I would much better prefer providing special set of variables under a reserved keys and discard them immediately after processing `each`.

.changeset/strong-colts-dance.md
```diff
@@ -0,0 +1,28 @@
...
+
+```js
+// ...
+context: {
+  todos: [/* ... */],
+  todo: null, // will be populated by `each` action
+  currentTodo: null // will be populated by `each` action
+}
+actions: each([
+  (context, event) => {
```
Andarist, line 15: This looks a little bit different from other actions - it's actually very different (it's an aggregate action~) from most so this might very well be justified. I'm wondering what are your thoughts about either:
- flipping args - a short `{array, item}` would be easier to spot before potentially lengthy actions
- or going full object style - `each({ array, item, actions })`

.changeset/strong-colts-dance.md
```diff
@@ -0,0 +1,28 @@
...
+    console.log({
+      todo,
+      currentTodo
+    });
+  },
+  // ... action objects
+], {
+  array: 'todos',
```
Andarist, line 23: we could potentially handle other iterables as well, maybe worth renaming this to a `collection`? OTOH not sure if we actually want to handle non-arrays from the start and another name would definitely indicate that this can handle more than arrays 🤔

davidkpiano:
> I guess I would much better prefer providing special set of variables under a reserved keys and discard them immediately after processing `each`.

By default? Sure, I agree.

packages/core/src/scxml.ts
```diff
@@ -172,41 +175,83 @@ function mapAction<
...
+                    `;
 
-          return evaluateExecutableContent(context, _ev, meta, fnBody);
-        };
+            return evaluateExecutableContent(context, _ev, meta, fnBody);
+          };
+        }
+      }
+
+      if ('idlocation' in element.attributes!) {
```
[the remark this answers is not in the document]
davidkpiano: It's awkward but ultimately a good idea.

packages/core/src/scxml.ts
```diff
@@ -263,6 +308,28 @@ function mapAction<
       conds.push(current);
       return actions.choose(conds);
     }
+    case 'foreach':
+      const { array, item, index } = element.attributes!;
+
+      // check if expected attributes are plain variable identifiers
+      for (const value of [array, item, index]) {
+        if (value && !/^\w/.test(value as string)) {
```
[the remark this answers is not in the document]
davidkpiano: Probably, yes.

.changeset/strong-colts-dance.md
davidkpiano: I'd probably lean towards the second suggestion (full object style) 🤔

.changeset/strong-colts-dance.md
davidkpiano: The name `array` is meant to match SCXML, as the only intention for the action creator is to support SCXML. Should I revise the changeset?

PR #1793, opened 2020-12-30
 add "exports" top level field to package.json

virtuoushub:
## why

> Export Maps help package authors accomplish two things: transition from Common.js (CJS)  to ESM & make a package's internal files private.
>
> Historically, it has been impossible for some packages to reliably support multiple module systems (UMD, CJS, ESM) and multiple environments (Web, Node.js, Deno). Past efforts to standardize have relied solely on community buy-in, with little official support from Node.js or npm. Export Maps aim to solve this problem as the official Node.js standard to define package entrypoints per-environment.

source: https://docs.skypack.dev/package-authors/package-checks#export-map

see also:
https://docs.skypack.dev/package-authors/package-checks#esm
https://nodejs.org/api/packages.html#packages_subpath_exports

Might help w/ #1625

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -17,6 +17,10 @@
     "es/**/*.js",
     "es/**/*.d.ts"
   ],
+  "exports": {
+    "require": "./lib/index.js",
+    "import": "./es/index.js"
+  },
   "keywords": [
     "statechart",
     "state machine",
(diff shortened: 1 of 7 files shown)


Andarist:
This is a breaking change - we'll consider adding this in v5 but there are some subtle details that we need to consider.

PR #1808, opened 2021-01-03
[v5] Rename machine.withConfig(...) to machine.provide(...)

davidkpiano:
Reference: https://github.com/davidkpiano/xstate/projects/1#card-40598347

This PR also renames `options.behaviors` to `options.actors`, as previously discussed.

### DIFF
--- a/.changeset/eighty-chefs-return.md
+++ b/.changeset/eighty-chefs-return.md
@@ -1,0 +1,5 @@
+---
+'xstate': major
+---
+
+Renamed `machine.withConfig(...)` to `machine.withOptions(...)`. https://github.com/davidkpiano/xstate/projects/1#card-40598347
--- a/.changeset/eighty-chefs-return.md
+++ b/.changeset/eighty-chefs-return.md
@@ -2,4 +2,4 @@
 'xstate': major
 ---
 
-Renamed `machine.withConfig(...)` to `machine.withOptions(...)`. https://github.com/davidkpiano/xstate/projects/1#card-40598347
+Renamed `machine.withConfig(...)` to `machine.provide(...)`.
--- a/.changeset/rotten-rivers-wonder.md
+++ b/.changeset/rotten-rivers-wonder.md
@@ -18,7 +18,7 @@ const machine = createMachine(
   },
   {
 -   activities: {
-+   behaviors: {
++   actors: {
 -     someActivity: ((context, event) => {
 +     someActivity: invokeActivity((context, event) => {
         // ... some continuous activity
@@ -32,7 +32,7 @@ const machine = createMachine(
 );
 ```
 
-**Breaking:** The `services` option passed as the second argument to `createMachine(config, options)` is renamed to `behaviors`. Each value in `behaviors`should be a function that takes in `context` and `event` and returns a [behavior](TODO: link). The provided invoke creators are:
+**Breaking:** The `services` option passed as the second argument to `createMachine(config, options)` is renamed to `actors`. Each value in `actors`should be a function that takes in `context` and `event` and returns a [behavior](TODO: link) for an actor. The provided invoke creators are:
 
 - `invokeActivity`
 - `invokePromise`
@@ -53,7 +53,7 @@ const machine = createMachine(
   },
   {
 -   services: {
-+   behaviors: {
++   actors: {
 -     fetchFromAPI: ((context, event) => {
 +     fetchFromAPI: invokePromise((context, event) => {
         // ... (return a promise)
@@ -86,7 +86,7 @@ const machine = createMachine({
 
 ```
 
-**Breaking:** The `src` of an `invoke` config is now either a string that references the machine's `options.behaviors`, or a `BehaviorCreator`, which is a function that takes in `context` and `event` and returns a `Behavior`:
+**Breaking:** The `src` of an `invoke` config is now either a string that references the machine's `options.actors`, or a `BehaviorCreator`, which is a function that takes in `context` and `event` and returns a `Behavior`:
 
 ```diff
 import { createMachine } from 'xstate';
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -1,5 +1,5 @@
 import {
-  MachineOptions,
+  MachineImplementations,
   DefaultContext,
   MachineConfig,
   StateSchema,
@@ -14,7 +14,7 @@ export function Machine<
   TEvent extends EventObject = AnyEventObject
 >(
   config: MachineConfig<TContext, TEvent, any>,
-  options?: Partial<MachineOptions<TContext, TEvent>>
+  options?: Partial<MachineImplementations<TContext, TEvent>>
 ): MachineNode<TContext, TEvent>;
 export function Machine<
   TContext = DefaultContext,
@@ -22,7 +22,7 @@ export function Machine<
   TStateSchema extends StateSchema = any
 >(
   config: MachineConfig<TContext, TEvent, TStateSchema>,
-  options?: Partial<MachineOptions<TContext, TEvent>>
+  options?: Partial<MachineImplementations<TContext, TEvent>>
 ): MachineNode<TContext, TEvent, TStateSchema>;
 export function Machine<
   TContext = DefaultContext,
@@ -30,7 +30,7 @@ export function Machine<
   TStateSchema extends StateSchema = any
 >(
   config: MachineConfig<TContext, TEvent, TStateSchema>,
-  options?: Partial<MachineOptions<TContext, TEvent>>
+  options?: Partial<MachineImplementations<TContext, TEvent>>
 ): MachineNode<TContext, TEvent, TStateSchema> {
   return new MachineNode<TContext, TEvent, TStateSchema, any>(config, options);
 }
@@ -41,7 +41,7 @@ export function createMachine<
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(
   config: MachineConfig<TContext, TEvent, any>,
-  options?: Partial<MachineOptions<TContext, TEvent>>
+  options?: Partial<MachineImplementations<TContext, TEvent>>
 ): MachineNode<TContext, TEvent, any, TTypestate> {
   return new MachineNode<TContext, TEvent, TTypestate>(config, options);
 }
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -13,26 +13,29 @@ export function Machine<
   TContext = any,
   TEvent extends EventObject = AnyEventObject
 >(
-  config: MachineConfig<TContext, TEvent, any>,
-  options?: Partial<MachineImplementations<TContext, TEvent>>
+  definition: MachineConfig<TContext, TEvent, any>,
+  implementations?: Partial<MachineImplementations<TContext, TEvent>>
 ): MachineNode<TContext, TEvent>;
 export function Machine<
   TContext = DefaultContext,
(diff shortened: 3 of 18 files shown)


.changeset/rotten-rivers-wonder.md
```diff
@@ -18,7 +18,7 @@ const machine = createMachine(
   },
   {
 -   activities: {
-+   behaviors: {
++   actors: {
```
Andarist, line 21: a good change 👍 you still mention behaviors in a lot of places, I wonder if we want to stick with that terminology? I see it as potentially very confusing for people, at the very least maybe we should bury it deep down and only mention it in like "the very advanced" section of the docs? 😅

packages/core/src/MachineNode.ts
```diff
@@ -142,22 +142,21 @@ export class MachineNode<
   }
 
   /**
-   * Clones this state machine with custom options and context.
```
Andarist, line 145: any reason behind the removal of this? you still support passing `context` in this method

.changeset/eighty-chefs-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': major
+---
+
+Renamed `machine.withConfig(...)` to `machine.withOptions(...)`. https://github.com/davidkpiano/xstate/projects/1#card-40598347
```
Andarist, line 5: This card doesn't seem to contain a lot of information that would be helpful for readers - I think its mention can be safely removed from here.
```suggestion
Renamed `machine.withConfig(...)` to `machine.withOptions(...)`.
```

Andarist (commented):
I'm not suuuuper-stoked about this change but on the other hand, both versions are so generic that it's hard to assess which one is better. Since we provide lookups for actions/actors/guards/etc in this, maybe it would be worth renaming this to `withImplementations`? It's somewhat mouthful though.

Don't take me wrong - I don't really mind the change, so do whatever you think is best for this one. The @johnyanarella's input could be valuable here as well.

davidkpiano:
> I'm not suuuuper-stoked about this change but on the other hand, both versions are so generic that it's hard to assess which one is better. Since we provide lookups for actions/actors/guards/etc in this, maybe it would be worth renaming this to `withImplementations`? It's somewhat mouthful though.
> 
> Don't take me wrong - I don't really mind the change, so do whatever you think is best for this one. The @johnyanarella's input could be valuable here as well.

It's important for consistency in terminology, and was mentioned externally which is why I made a note of it:

```js
const machine = createMachine(config, options);
//        ______________________⬆        ⬆
//        |                               |
machine.withConfig(/* ... */) //          |
//        ________________________________|
//        |
machine.withOptions(/* ... */)
```

Andarist:
Right - the consistency is definitely an improvement but it could be approached from the other end (I'm not saying that it should be). You could rename the `createMachine` parameters to smth like:
```js
createMachine(schema, implementations)
```

I didn't even realize that those 2 were currently named like this and it seems to me that it can still be quite confusing because, as mentioned, both config & options are quite generic so if I say "Change your config" one still can be confused on which thing they should adjust.

Not a big issue though - naming is hard 🤷‍♂️ I don't really have any strong contenders for the chosen names.

davidkpiano:
Implementation is the "correct" answer but you're right, it's not the most dev-friendly word to type out 🤔

Silverwolf90:
Apologies, but I couldn't resist some drive-by bikeshedding... but if open to dropping `withXXX` convention, what about `use` (or `using`)

`machine.use({ ... })`
`machine.using({ ... })`

anyway, sorry again, feel free to ignore 😄

davidkpiano:
@Silverwolf90 Thanks for the suggestion - still bikeshedding this!

What do you think about:
```diff
machine.withConfig(/* ... */)
machine.provide(/* ... */)
```

As in, "provide these implementations"?

johnyanarella:
I think it will be worth all the bikeshedding in the end - I'm excited about the potential that tweaking the names here has for reducing the learning curve for newcomers.

Moving away from "configuration" is great news, given SCXML already uses it in this problem space—adhering to the classical definition of an arrangement or combination of things—to describe the set of currently active states. The interpreter "options" conform to the common expectation of a *fixed set* of knobs and toggles with defaults. Using that same term to describe passing a machine an arbitrary and unbounded set of methods doesn't feel like a good fit.

At one point I think I halfheartedly suggested the idea of calling the thing passed in here a "wiring harness". Not keen on that one, but I do still wonder if we could mine the terminology associated with machines in the physical world to find something familiar (and less abstract) that intuitively and unambiguously conveys the idea here.

Determining the right terminology to describe this pattern—referencing actions/actors/guards by name and plugging in language native implementations—will make a huge difference in highlighting this capability, its purpose and value, as the XState community expands beyond JavaScript.

I gotta concede that `withImplementations()` is precise, but a real pain to type. `provide()` is compelling. That method name and behavior will be familiar to anyone who has used an IoC container API, which is essentially what's happening here...

Andarist:
> "options" conform to the common expectation of a fixed set of knobs and toggles with defaults

This might be the best description for `options` that I have ever seen! Gonna probably think about it in those exact terms since now - I would only say that not all options have to provide defaults, your statement kinda reads as if that would be a requirement.

Do you have any good definition for `config`? 😉

johnyanarella:
Fair point!

You're absolutely right about an `option` choice among `options` sometimes not even being optional ("I'll give you two options...!"). And then there's the perspective on optionality where `nil` is one of the options—and the default one, at that. (Or in JavaScript, the nullish `undefined` or `null` 🤣).

In practice, we see `configuration`, `preferences`, `properties` and `settings` used interchangeably in software.

Configuration is arguably distinct among those as precisely describing a specific arrangement or combination among possibilities. When you configure your MacBook Pro M1 for purchase, you pick a configuration of cpu speed, ram and ssd size options (another example where the options aren't really optional!).

Given their mathematical underpinnings, it's not surprising that the SCXML specification and even David Harel's original statecharts paper use `configuration` when referring to a distinct set of states among possible states. Harel is even more precise in referring to it as a "state-configuration".

Anyway:

* I still think the option (😮) of using `config` for anything other than a state configuration here has been precluded by the pre-existing use of `configuration` in this particular problem space; and
* `provide()` makes more and more sense the longer I've had to think about it, as this is performing DI, and that's a pretty conventional API for one; and
* I'm not sure what you name the type of thing thing you would pass to `provide()`.

johnyanarella:
Regarding this suggestion above:

```js
createMachine(schema, implementations)
```

Over in `xstate-swift`, I currently have:

```swift
// an instance of MachineSchematic created either via the Swift DSL or parsed from SCXML
let machineSchematic = MachineSchematic.fromSCXML(contentsOf: scxmlURL)

// an instance of MachineOptions mapping named references to native implementations
let machineOptions = MachineOptions(
    actions: ...,
    actors: ...,
    guards: ...
)

let machine = Machine(schematic: machineSchematic, options: machineOptions)
```

The term `schema` comes with almost as much baggage as `state`.

I gravitated toward `MachineSchematic` and `schematic` there as it has some familiar association with the term `Machine` and helps convey their relationship—the schematic is inert on its own and is a blueprint used later to construct a machine.

As we've discussed above, I'm no longer convinced `MachineOptions` or `options` is ideal here. `implementations` is compelling, but mouthful, and `MachineImplementations` leaves a lot to be desired as a type name.

davidkpiano:
@johnyanarella Super helpful thoughts, thanks. So just to digest this, the `createMachine()` factory takes two arguments:

1. The machine _schema_, which is its _partial definition_
2. The implementation details _provided_ to the machine

For 1., the full definition can be provided as well, and providing a partial definition is just for improved DX, since manually typing out a full definition is tedious, and a lot of things can be inferred (I'm saying this to justify `definition` as a term).

So, for tentative naming, I propose this:

```js
createMachine(definition, implementations);
```

And for the original PR, this:

```js
someMachine.provide(implementations);
```

Of course, the developers would not need to type out `"implementations"` - that would just be what's present in the documentation and types.

Andarist:
I very much like where this bike-shedding has brought us to 👍 

one important realization is that with this change we effectively seal what the second parameter is used for. We should be cautious in the future not to introduce there anything that is not an „implementation”

johnyanarella:
Nice!

I'll mirror this terminology choice over into `xstate-swift`. Everything there that comprises the `MachineDefinition` (resulting from parsing SCXML or the Swift DSL) is named with a similar suffix already (`StateDefinition`, `TransitionDefinition`, etc.), so that lines up nicely.

With Swift, I've got to contend with labelled parameters. So, I'll probably sidestep `implementations` in externally facing APIs where I can with a `provided` label:

```swift
let machine = Machine(
    definition: MachineDefinition.fromSCXML(data: scxmlData),
    provided: ImplementationDetails(
        actions: [
            "foo": { context, event, meta in
              // ...
            }
        ],
        // ...
    )
)
```

(barring anyone coming up with a better type name for `ImplementationDetails`).

johnyanarella:
...or did I misinterpret your comment above, and `MachineSchema` is the type name for your `definition` parameter?

johnyanarella:
(Worth noting, that the same possibility exists in `xstate-swift` of passing a full definition and omitting the second parameter, if you've 1) used the Swift DSL and don't reference actions/actors/guards/etc. by name; or 2) parsed an SCXML document that needs no native implementation details.)

packages/core/src/MachineNode.ts
davidkpiano: Reverted

packages/core/src/Machine.ts
```diff
@@ -14,23 +14,23 @@ export function Machine<
   TEvent extends EventObject = AnyEventObject
 >(
   config: MachineConfig<TContext, TEvent, any>,
-  options?: Partial<MachineOptions<TContext, TEvent>>
+  options?: Partial<MachineImplementations<TContext, TEvent>>
```
Andarist, line 17: The name of the second argument should be adjusted as well if we are following the discussed naming convention

packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,54 @@
 # xstate
 
+## 4.16.0
```
Andarist, line 3: Please do not merge master to feature branches targeting next before you merge master to next itself 🙏  it makes reviewing much harder

packages/core/src/MachineNode.ts
```diff
@@ -38,10 +38,10 @@ export const WILDCARD = '*';
...
 const createDefaultOptions = <TContext>(
   context: TContext
-): MachineOptions<TContext, any> => ({
+): MachineImplementations<TContext, any> => ({
   actions: {},
   guards: {},
-  behaviors: {},
+  actors: {},
   delays: {},
   context
```
Andarist, line 46: q: if we are naming those as implementations, should `context` be part of it?

Andarist (approved):

packages/core/CHANGELOG.md
davidkpiano: Ah, oops, sorry!

packages/core/src/MachineNode.ts
davidkpiano: Not sure, still thinking about whether this is a good idea.

Andarist: Created a card so we can decide about this later - https://github.com/davidkpiano/xstate/projects/1#card-53045090

devanshj:
> So, for tentative naming, I propose this:
> 
> ```js
> createMachine(definition, implementations);
> ```

Stumbled upon this, any chance txstate influenced this naming or just a coincidence? :P

PR #1811, opened 2021-01-04
[v5] Add support for partial wildcards (prefix matching)

davidkpiano:
This PR adds support for prefix matching with wildcards (and without wildcards for SCXML machines only):

```js
  // ...
  on: {
    // This transition accepts event types like:
    // "mouse", "mouse.click", "mouse.move.out", etc.
    // but not events like "mouseclick"
    'mouse.*': {/* ... */}
  }
});
```

https://github.com/davidkpiano/xstate/projects/1#card-52117183

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -554,14 +554,14 @@ export class Interpreter<
 
   private sendTo(
     event: SCXML.Event<AnyEventObject>,
-    to: string | number | ActorRef<any>
+    to: string | ActorRef<any>
   ) {
     const isParent = this.parent && to === SpecialTargets.Parent;
     const target = isParent
       ? this.parent
       : isActorRef(to)
       ? to
-      : this.children.get(to);
+      : this.children.get(to.startsWith('#_') ? to.slice(2) : to);
 
     if (!target) {
       if (!isParent) {
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -563,7 +563,7 @@ export class Interpreter<
       ? this.parent
       : isActorRef(to)
       ? to
-      : this.children.get(to.startsWith('#_') ? to.slice(2) : to);
+      : this.children.get(to);
 
     if (!target) {
       if (!isParent) {
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -229,7 +229,9 @@ export class Interpreter<
           : undefined;
 
       for (const listener of this.doneListeners) {
-        listener(doneInvoke(this.id, doneData));
+        listener(
+          toSCXMLEvent(doneInvoke(this.id, doneData), { invokeid: this.id })
+        );
       }
       this.stop();
     }
--- a/packages/core/src/scxml.ts
+++ b/packages/core/src/scxml.ts
@@ -318,12 +318,6 @@ function toConfig(
         target: target ? `#${target}` : undefined
       };
     }
-    case 'final': {
-      return {
-        ...nodeJson.attributes,
-        type: 'final'
-      };
-    }
     default:
       break;
   }
@@ -377,13 +371,6 @@ function toConfig(
         );
 
         return events.map((event) => {
-          if (event === 'done.invoke') {
-            throw new Error(
-              'done.invoke gets often used in SCXML tests as inexact event descriptor.' +
-                " As long as this stay unimplemented or done.invoke doesn't get a specialcased while converting throw when seeing it to avoid tests using this to pass by accident."
-            );
-          }
-
           const targets = getAttribute(value, 'target');
           const internal = getAttribute(value, 'type') === 'internal';
 
@@ -472,6 +459,7 @@ function toConfig(
           }
         : undefined),
       ...(parallel ? { type: 'parallel' } : undefined),
+      ...(nodeJson.name === 'final' ? { type: 'final' } : undefined),
       ...(stateElements.length
         ? {
             states: mapValues(states, (state, key) =>
@@ -497,7 +485,18 @@ function scxmlToMachine(
   scxmlJson: XMLElement,
   options: ScxmlToMachineOptions
 ): MachineNode {
-  const machineElement = scxmlJson.elements!.find(
+  if (!scxmlJson.elements) {
+    // If there is no SCXML element, this should call 'done.invoke' immediately.
+    // TODO: there's probably a better way to do this
+    return Machine({
+      initial: 'final',
+      states: {
+        final: { type: 'final' }
+      }
+    });
+  }
+
+  const machineElement = scxmlJson.elements.find(
     (element) => element.name === 'scxml'
   ) as XMLElement;
 
@@ -529,7 +528,8 @@ function scxmlToMachine(
   return Machine({
     ...toConfig(machineElement, '(machine)', options),
     context,
-    delimiter: options.delimiter
+    delimiter: options.delimiter,
+    scxml: true
   });
 }
 
--- a/packages/core/src/scxml.ts
+++ b/packages/core/src/scxml.ts
@@ -318,6 +318,12 @@ function toConfig(
         target: target ? `#${target}` : undefined
       };
     }
+    // case 'final': {
+    //   return {
+    //     ...nodeJson.attributes,
+    //     type: 'final'
+    //   };
+    // }
     default:
       break;
   }
@@ -474,7 +480,7 @@ function toConfig(
     };
   }
 
-  return { id };
+  return { id, ...(nodeJson.name === 'final' ? { type: 'final' } : undefined) };
 }
 
 export interface ScxmlToMachineOptions {
--- a/packages/core/src/scxml.ts
+++ b/packages/core/src/scxml.ts
@@ -318,12 +318,6 @@ function toConfig(
         target: target ? `#${target}` : undefined
       };
     }
-    // case 'final': {
-    //   return {
-    //     ...nodeJson.attributes,
-    //     type: 'final'
-    //   };
-    // }
     default:
       break;
   }
--- a/packages/core/src/scxml.ts
+++ b/packages/core/src/scxml.ts
@@ -485,18 +485,7 @@ function scxmlToMachine(
   scxmlJson: XMLElement,
   options: ScxmlToMachineOptions
 ): MachineNode {
-  if (!scxmlJson.elements) {
-    // If there is no SCXML element, this should call 'done.invoke' immediately.
-    // TODO: there's probably a better way to do this
-    return Machine({
-      initial: 'final',
-      states: {
-        final: { type: 'final' }
-      }
-    });
-  }
-
-  const machineElement = scxmlJson.elements.find(
+  const machineElement = scxmlJson.elements!.find(
     (element) => element.name === 'scxml'
   ) as XMLElement;
 
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -261,9 +261,10 @@ export function getCandidates<TEvent extends EventObject>(
   stateNode: StateNode<any, TEvent>,
   eventName: TEvent['type'] | NullEvent['type'],
   /**
-   * If `true`, will use SCXML event token matching semantics
+   * If `true`, will use SCXML event partial token matching semantics
+   * without the need for the ".*" suffix
    */
-  matchTokens: boolean = false
+  partialMatch: boolean = false
 ): Array<TransitionDefinition<any, TEvent>> {
   const transient = eventName === NULL_EVENT;
   const candidates = stateNode.transitions.filter((transition) => {
@@ -274,7 +275,7 @@ export function getCandidates<TEvent extends EventObject>(
 
     // transient transitions can't match non-transient events
     if (transient) {
-        return false;
+      return false;
     }
     // Then, check if transition is a wildcard transition,
     // which matches any non-transient events
@@ -282,20 +283,32 @@ export function getCandidates<TEvent extends EventObject>(
       return true;
     }
 
-    if (!matchTokens && !transition.eventType.endsWith('.*')) {
+    if (!partialMatch && !transition.eventType.endsWith('.*')) {
       return false;
     }
 
     const partialEventTokens = transition.eventType.split('.');
     const eventTokens = eventName.split('.');
 
-    // tslint:disable-next-line: forin
-    for (const tokenIndex in partialEventTokens) {
+    for (
+      let tokenIndex = 0;
+      tokenIndex < partialEventTokens.length;
+      tokenIndex++
+    ) {
       const partialEventToken = partialEventTokens[tokenIndex];
       const eventToken = eventTokens[tokenIndex];
 
       if (partialEventToken === '*') {
-        return true;
+        const isLastToken = tokenIndex === partialEventTokens.length - 1;
+
+        if (!IS_PRODUCTION) {
+          warn(
+            isLastToken,
+            `Infix wildcards in transition events are not allowed. Check the "${transition.eventType}" event.`
+          );
+        }
+
+        return isLastToken;
       }
 
       if (partialEventToken !== eventToken) {
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -272,10 +272,14 @@ export function getCandidates<TEvent extends EventObject>(
       return true;
     }
 
+    // transient transitions can't match non-transient events
+    if (transient) {
+        return false;
+    }
     // Then, check if transition is a wildcard transition,
-    // which matches all non-transient events
+    // which matches any non-transient events
     if (transition.eventType === WILDCARD) {
-      return !transient;
+      return true;
     }
 
     if (!matchTokens && !transition.eventType.endsWith('.*')) {
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -259,36 +259,45 @@ export const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
 
 export function getCandidates<TEvent extends EventObject>(
   stateNode: StateNode<any, TEvent>,
-  eventName: TEvent['type'] | NullEvent['type'],
+  receivedEventType: TEvent['type'] | NullEvent['type'],
   /**
    * If `true`, will use SCXML event partial token matching semantics
    * without the need for the ".*" suffix
    */
   partialMatch: boolean = false
 ): Array<TransitionDefinition<any, TEvent>> {
-  const transient = eventName === NULL_EVENT;
+  const transient = receivedEventType === NULL_EVENT;
   const candidates = stateNode.transitions.filter((transition) => {
+    const { eventType } = transition;
     // First, check the trivial case: event names are exactly equal
-    if (transition.eventType === eventName) {
+    if (eventType === receivedEventType) {
       return true;
     }
 
-    // transient transitions can't match non-transient events
+    // Transient transitions can't match non-transient events
     if (transient) {
       return false;
     }
+
     // Then, check if transition is a wildcard transition,
     // which matches any non-transient events
-    if (transition.eventType === WILDCARD) {
+    if (eventType === WILDCARD) {
       return true;
     }
 
-    if (!partialMatch && !transition.eventType.endsWith('.*')) {
+    if (!partialMatch && !eventType.endsWith('.*')) {
       return false;
     }
 
-    const partialEventTokens = transition.eventType.split('.');
-    const eventTokens = eventName.split('.');
+    if (!IS_PRODUCTION) {
+      warn(
+        !/.*\*.+/.test(eventType),
+        `Wildcards can only be the last token of an event descriptor (e.g., "event.*") or the entire event descriptor ("*"). Check the "${eventType}" event.`
+      );
+    }
+
+    const partialEventTokens = eventType.split('.');
+    const eventTokens = receivedEventType.split('.');
 
     for (
       let tokenIndex = 0;
@@ -304,7 +313,7 @@ export function getCandidates<TEvent extends EventObject>(
         if (!IS_PRODUCTION) {
           warn(
             isLastToken,
-            `Infix wildcards in transition events are not allowed. Check the "${transition.eventType}" event.`
+            `Infix wildcards in transition events are not allowed. Check the "${eventType}" event.`
           );
         }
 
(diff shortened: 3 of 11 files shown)


packages/core/src/stateUtils.ts
```diff
@@ -259,15 +259,49 @@ export const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
...
+    // First, check the trivial case: event names are exactly equal
+    if (transition.eventType === eventName) {
+      return true;
+    }
+
+    // Then, check if transition is a wildcard transition,
+    // which matches all non-transient events
+    if (transition.eventType === WILDCARD) {
+      return !transient;
+    }
```
Andarist, line 279: This works OK now but I had to spend a little bit of time to see if everything is OK here - especially given the fact that the `eventType: '.*'` gets processed by the rest of this function. I think it's an incorrect descriptor and it's correctly not matched here but I think that this would be slightly more obvious to analyze:
```suggestion
    // transient transitions can't match non-transient events
    if (transient) {
        return false;
    }
    // Then, check if transition is a wildcard transition,
    // which matches any non-transient events
    if (transition.eventType === WILDCARD) {
      return true;
    }
```

packages/core/src/stateUtils.ts
```diff
@@ -259,15 +259,49 @@ export const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
 
 export function getCandidates<TEvent extends EventObject>(
   stateNode: StateNode<any, TEvent>,
-  eventName: TEvent['type'] | NullEvent['type'] | '*'
+  eventName: TEvent['type'] | NullEvent['type'],
+  /**
+   * If `true`, will use SCXML event token matching semantics
+   */
+  matchTokens: boolean = false
```
Andarist, line 266: General q: is it even worth supporting this? I mean - my bet is that this will only ever be used in our tests and we could deal with that differently. For example by forking those tests that require all event descriptors to be treated as inexact.

I don't have a super-strong desire to remove this - but usually, I try to advocate for not including things like this.

davidkpiano: It's not public API and this is the way I solved dealing with it in our tests.

I'd rather not change the tests and make matching tokens opt-in (internal) behavior than forking the tests.

It'll also be important for anyone using XState as an SCXML processor in the future.

packages/core/src/scxml.ts
```diff
@@ -318,12 +318,12 @@ function toConfig(
...
-        ...nodeJson.attributes,
-        type: 'final'
-      };
-    }
+    // case 'final': {
+    //   return {
+    //     ...nodeJson.attributes,
+    //     type: 'final'
+    //   };
+    // }
```
davidkpiano, line 326:
```suggestion
```

packages/core/src/scxml.ts
```diff
@@ -497,7 +491,18 @@ function scxmlToMachine(
   scxmlJson: XMLElement,
   options: ScxmlToMachineOptions
 ): MachineNode {
-  const machineElement = scxmlJson.elements!.find(
+  if (!scxmlJson.elements) {
+    // If there is no SCXML element, this should call 'done.invoke' immediately.
+    // TODO: there's probably a better way to do this
```
davidkpiano, line 490: Note: I think a couple SCXML tests literally have this same "finish immediately" machine  invoked in the tests 😝

packages/core/test/interpreter.test.ts
```diff
@@ -1742,6 +1742,7 @@ describe('interpreter', () => {
 
       const service = interpret(parentMachine)
         .onTransition((state) => {
+          console.log(state.value, state._event);
```
davidkpiano, line 1745:
```suggestion
```

packages/core/src/stateUtils.ts
```diff
@@ -259,15 +259,49 @@ export const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
...
+      }
+
+      if (partialEventToken !== eventToken) {
+        return false;
+      }
+
+      continue;
+    }
+
+    return true;
```
Andarist, line 308: Please recheck what I'm proposing here because I haven't tested this but I think that this whole chunk of code could be replaced with:
```suggestion
    const matcher = new RegExp(`^${escapeDots(transition.eventType.replace(/\.*/, ''))}(\\.|$)`)
    return matcher.test(eventName)
```
where `escapeDots` would be defined as:
```js
const escapeDots = (str: string): string => str.replace(/\./g, '\\.')
```

--- 

After actually writing this I'm not super sure if this is that much better - but still takes a little bit less of a focus (for me) to understand how this matching works. It's subjective though 🤷‍♂️

packages/core/src/stateUtils.ts
```diff
@@ -259,15 +259,49 @@ export const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
...
+      return true;
+    }
+
+    // Then, check if transition is a wildcard transition,
+    // which matches all non-transient events
+    if (transition.eventType === WILDCARD) {
+      return !transient;
+    }
+
+    if (!matchTokens && !transition.eventType.endsWith('.*')) {
```
Andarist, line 285: According to event descriptors section in the SCXML: https://www.w3.org/TR/scxml/#EventDescriptors it seems that `foo`, `foo.*` and `foo.` are all equivalent o.O which I have just learned a second ago.

packages/core/src/stateUtils.ts
Andarist: Actually, according to the SCXML, event names (and descriptors) can only contain alphanumeric characters delimited with dots - not mentioning this as the argument for using regexes here (as I've mentioned - in the end this didn't turn out as clean as I have wanted) but rather to mention that your example is not valid (according to the SCXML).

The only thing I would change in that tokens loop would be to actually iterate over `partialEventToken.length` to obtain indices - you wouldn't have to tell TSLint to shut up then and it would kinda be more idiomatic. Knowing that those are arrays I've assumed that this is for-of loop and was scratching my head how the hell you got access to the current index this way 😅

packages/core/src/stateUtils.ts
```diff
@@ -259,15 +259,53 @@ export const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
...
+
+      if (partialEventToken === '*') {
+        return true;
+      }
+
+      if (partialEventToken !== eventToken) {
+        return false;
+      }
+
+      continue;
```
Andarist, line 305:
```suggestion
```

packages/core/src/stateUtils.ts
Andarist: I'm not really sure if we need to support this - it's obscure and I can't find any test that would actually check this, nor I can't find this being explicitly mentioned in the CCXML spec. My vote would be against implementing support for this ;p

packages/core/src/scxml.ts
```diff
@@ -318,12 +318,6 @@ function toConfig(
         target: target ? `#${target}` : undefined
       };
     }
-    case 'final': {
```
Andarist, line 321: Hm, *I think* this change will allow (and create) more invalid configs than previously. `<final>` accepts way fewer children element types than regular `<state>`. Also - from what I see (but I could have missed something) we don't yet support `<donedata>` conversion - which is `<final>`-specific, so it would make sense to me to keep a dedicated `<final>` converter (instead of reusing the looser~ converter)

packages/core/src/stateUtils.ts
Andarist: could we rename this option to something else even if it's just `scxml`? even for `matchTokens === false` we are still matching tokens which makes it confusing (only the precondition for matching is different)

packages/core/test/scxml.test.ts
```diff
@@ -358,14 +358,15 @@ async function runW3TestToCompletion(machine: MachineNode): Promise<void> {
         nextState = state;
       })
       .onDone(() => {
-        if (nextState.value === 'pass') {
+        // Add 'final' for test230.txml which does not have a 'pass' state
```
Andarist, line 361: eh, it's classified as a manual test - not sure rly why as this one would be totally possible to be rewritten as an automated one (it would grow a little bit due to added messages passing and context assignments but whatever)

🤷‍♂️

packages/core/src/stateUtils.ts
```diff
@@ -259,15 +259,51 @@ export const isStateId = (str: string) => str[0] === STATE_IDENTIFIER;
...
+
+    const partialEventTokens = transition.eventType.split('.');
+    const eventTokens = eventName.split('.');
+
+    // tslint:disable-next-line: forin
+    for (const tokenIndex in partialEventTokens) {
+      const partialEventToken = partialEventTokens[tokenIndex];
+      const eventToken = eventTokens[tokenIndex];
+
+      if (partialEventToken === '*') {
```
Andarist, line 297: According to SCXML `.*` is only allowed at the end of the event descriptor and this would return `true` for a wildcard at any position in the delimited descriptor. Even if we don't validate the inputs and assume they are valid it still would be, IMHO, preferable not to trip on edge cases like this.

You can test this with:
```js
var matchTokens = true // just for simplicity of event descriptor etc, a similar effect can be achieved with false
var eventName = 'foo.xyz.baz'
var stateNode = { transitions: [{ eventType: 'foo.*.bar' }] }
```

Even if this would be a valid descriptor this `eventName` should not be matched by it.

packages/core/test/eventDescriptors.test.ts
```diff
@@ -60,4 +60,157 @@ describe('event descriptors', () => {
...
     expect(service.state.value).toBe('pass');
   });
+
+  it('should NOT support non-tokenized wildcards', () => {
+    const machine = Machine({
+      initial: 'start',
+      states: {
+        start: {
+          on: {
+            'event*': 'success'
```
Andarist, line 70: this seems like a clearly incorrect descriptor (SCXML defines it as invalid, interestingly this works in CCXML though and would match `eventwhatever`), should we maybe add dev-only warnings for this?

packages/core/test/eventDescriptors.test.ts
```diff
@@ -60,4 +60,157 @@ describe('event descriptors', () => {
...
+
+    expect(
+      machine.transition(undefined, 'event').matches('success')
+    ).toBeTruthy();
+    expect(
+      machine.transition(undefined, 'eventually').matches('success')
+    ).toBeFalsy();
+  });
+
+  it('should support prefix matching with wildcards (+1)', () => {
```
Andarist, line 110: the machine from this test is the same as in the previous one - was maybe something else the intention of this test?

packages/core/test/eventDescriptors.test.ts
```diff
@@ -60,4 +60,157 @@ describe('event descriptors', () => {
...
+            event: 'success'
+          }
+        },
+        success: {
+          type: 'final'
+        }
+      }
+    });
+
+    expect(
```
Andarist, line 208: would be great to add this as well here:
```
    expect(
      SCXMLMachine.transition(undefined, 'eventually').matches('start')
    ).toBeTruthy();
```

packages/core/test/eventDescriptors.test.ts
```diff
@@ -60,4 +60,157 @@ describe('event descriptors', () => {
...
+      nonSCXMLMachine.transition(undefined, 'event.whatever').matches('start')
+    ).toBeTruthy();
+
+    expect(
+      SCXMLMachine.transition(undefined, 'event.whatever').matches('success')
+    ).toBeTruthy();
```
Andarist, line 214:
```suggestion
    expect(
      SCXMLMachine.transition(undefined, 'event.whatever').matches('success')
    ).toBeTruthy();
    
    expect(
      SCXMLMachine.transition(undefined, 'eventually').matches('start')
    ).toBeTruthy();
```

packages/core/src/stateUtils.ts
davidkpiano: The `transition.eventType.endsWith('.*')` clause takes care of this.

packages/core/src/stateUtils.ts
davidkpiano: Yeah, I agree with not supporting `foo.`

packages/core/src/stateUtils.ts
davidkpiano: > even for `matchTokens === false` we are still matching tokens

We're not matching tokens; we're either matching string equivalence (`transition.eventType === eventName`) or any string with a wildcard.

davidkpiano: Changed this to `partialMatch`

packages/core/src/stateUtils.ts
davidkpiano: Makes sense - I switched this to a simple `for` loop with indexes.

packages/core/src/stateUtils.ts
davidkpiano: Added tests for this and disallowed the invalid event with a warning.

packages/core/test/eventDescriptors.test.ts
davidkpiano: SCXML doesn't define it as invalid directly; it only says that the event name is alphanumeric. However, it only makes this "restriction" (assumedly) for event names, but not for state names or generated IDs.

This restriction can be problematic because a state might nave an ID like `"fetching-user"` and generated event might look like `"done.state.fetching-user"`, which would be invalid if SCXML disallowed it (because of the `"-"` character.

However, this would be contradictory since it's only restrictive for event names.

IMO I would rather be more liberal - event names can consist of any character, but `"."` is the only special character used for tokenization.

packages/core/src/scxml.ts
davidkpiano: I'm going to refactor this in a separate PR - this file needs to be cleaned up in general. Main goal is getting the tests to pass for now.

packages/core/test/eventDescriptors.test.ts
Andarist: There are also restrictions for state IDs - they need to be of ID type (defined in the XML Schema spec). But quite frankly I don't care about it, reading that spec would take me more time than i'd like it too and from what I can quickly assess - you are right in a sense that state IDs can have more than alphanumeric characters (for instance I'm pretty sure that `_` is allowed) and that kinda contradicts that statement about alphanumeric restriction for the event names (like you have mentioned).

I'm OK with being more liberal here for other purposes but I think that at least `*` should be considered illegal in the middle of a string. It doesn't do much good and given it has a special meaning when it is on the end it might only end up being confusing if we allow it at any position but without special meaning.

packages/core/src/interpreter.ts
```diff
@@ -554,14 +556,14 @@ export class Interpreter<
...
-    to: string | number | ActorRef<any>
+    to: string | ActorRef<any>
   ) {
     const isParent = this.parent && to === SpecialTargets.Parent;
     const target = isParent
       ? this.parent
       : isActorRef(to)
       ? to
-      : this.children.get(to);
+      : this.children.get(to.startsWith('#_') ? to.slice(2) : to);
```
Andarist, line 566: Just a general wondering - should this be handled here at the end or maybe normalized when resolving `send()`? It has implications on what will be displayed in the visualizer etc. 

I don't have an opinion about this one - just something I'm raising for additional consideration.

packages/core/test/scxml.test.ts
```diff
@@ -168,30 +168,30 @@ const testGroups = {
...
+    // 'test224.txml', // <invoke idlocation="...">
     // 'test225.txml', // unique invokeids generated at invoke time
     // 'test226.txml', // <invoke src="...">
-    // 'test228.txml', // done.invoke used as inexact event descriptor + _event.invokeid not implemented yet
+    'test228.txml',
     'test229.txml',
-    // 'test230.txml', // done.invoke used as inexact event descriptor + SCXML _event properties not implemented yet
-    // 'test232.txml', // done.invoke used as inexact event descriptor
+    'test230.txml',
+    // 'test232.txml', // TODO: not sure
```
Andarist, line 179: This test is passing now and I don't see anything much special about it so let's assume that it has not been passing for you in the middle of development due to some minor bug
```suggestion
    'test232.txml',
```

packages/core/test/scxml.test.ts
```diff
@@ -168,30 +168,30 @@ const testGroups = {
...
+    'test220.txml',
     // 'test223.txml', // idlocation not implemented yet
-    // 'test224.txml', // done.invoke used as inexact event descriptor
+    // 'test224.txml', // <invoke idlocation="...">
     // 'test225.txml', // unique invokeids generated at invoke time
     // 'test226.txml', // <invoke src="...">
-    // 'test228.txml', // done.invoke used as inexact event descriptor + _event.invokeid not implemented yet
+    'test228.txml',
     'test229.txml',
-    // 'test230.txml', // done.invoke used as inexact event descriptor + SCXML _event properties not implemented yet
```
Andarist, line 178: I don't think this should be uncommented - just yet. This is a "manual" test. Those manual tests usually just require observing the logged output so it's quite easy for us to actually automate them. We can spy `console.log` and snapshot results - but even if you'd end up doing that as part of this PR this test still shouldn't be accounted as "passing" because the snapshotted logs would not match expectations.

packages/core/test/scxml.test.ts
```diff
@@ -168,30 +168,30 @@ const testGroups = {
...
-    // 'test232.txml', // done.invoke used as inexact event descriptor
+    'test230.txml',
+    // 'test232.txml', // TODO: not sure
     // 'test233.txml', // <finalize> not implemented yet
     // 'test234.txml', // <finalize> not implemented yet
     'test235.txml',
-    // 'test236.txml', // done.invoke used as inexact event descriptor
-    // 'test237.txml', // done.invoke used as inexact event descriptor
-    // 'test239.txml', // done.invoke used as inexact event descriptor
+    // 'test236.txml', // TODO: not sure
```
Andarist, line 183: this block: 
https://github.com/davidkpiano/xstate/blob/8d328e92f74d0a428aa66a2395f9aa85bbba3223/packages/core/src/interpreter.ts#L219
should contain a similar logic as this:
https://github.com/davidkpiano/xstate/blob/8d328e92f74d0a428aa66a2395f9aa85bbba3223/packages/core/src/interpreter.ts#L426-L430
```suggestion
    // 'test236.txml', // reaching a final state should execute all onexit handlers
```

packages/core/test/scxml.test.ts
```diff
@@ -230,13 +230,13 @@ const testGroups = {
     'test335.txml',
     'test336.txml',
     'test337.txml',
-    // 'test338.txml', // _event.invokeid not implemented yet
+    'test338.txml',
```
Andarist, line 233: This is passing only by accident - because both variables that are being compared yield `undefined` value but to make this test actually working both should be assigned with correct values
```suggestion
    // 'test338.txml', // <invoke idlocation="..."> + _event.invokeid available on <send> events received from the invoked child
```

packages/core/test/scxml.test.ts
```diff
@@ -253,9 +253,9 @@ const testGroups = {
     'test387.txml',
     'test388.txml',
     'test396.txml',
-    // 'test399.txml', // inexact prefix event matching not implemented
-    // 'test401.txml', // inexact "error" event (should be "error.execution")
-    // 'test402.txml', // error.execution when evaluating assign + inexact prefix event matching not implemented
+    'test399.txml',
+    'test401.txml',
+    'test402.txml',
```
Andarist, line 258: I believe that this might be passing accidentally and this needs further investigation. It currently fails because context is undefined - I'm not 100% sure if that's correct. Maybe it should fail on assigning to non-existent context property (it tries to assign to `location="foo.bar.baz "`). Maybe failing  with:
```
Cannot assign to undefined `context`. Ensure that `context` is defined in the machine config.
```
is expected here but it's worth ensuring that - can be done later though.

If I unblock this by providing an empty context I hit another set of problems - this is why I believe it would be best to leave it as commented out for now.

1. Our script for [`exposing datamodel`](https://github.com/davidkpiano/xstate/blob/8d328e92f74d0a428aa66a2395f9aa85bbba3223/packages/core/src/scxml.ts#L107-L111) to expressions is naive - it results in things like: `const foo.bar.baz  = context['foo.bar.baz '];` which are syntax errors)
2. raised event is executed **before** executable content (`assign` in this test). This is not correct - `raise` should add to the internal queue from which the event should be taken off after executing whole executable content caused by a transition

```suggestion
    // 'test402.txml', // TODO: investigate more, it expects error.execution when evaluating assign, check if assigning to a deep location is even allowed, check if assigning to an initialized datamodel is allowed, improve how datamodel is exposed to constructed functions
```

Andarist:
I've reviewed some SCXML tests that got enabled to check if they were enabled correctly - I haven't done that for all of them. I might do it but not sure if you want to stall this PR waiting for that. I might compile a list of enabled tests later for myself to re-check them.

packages/core/src/scxml.ts
```diff
@@ -497,7 +485,18 @@ function scxmlToMachine(
   scxmlJson: XMLElement,
   options: ScxmlToMachineOptions
 ): MachineNode {
-  const machineElement = scxmlJson.elements!.find(
+  if (!scxmlJson.elements) {
+    // If there is no SCXML element, this should call 'done.invoke' immediately.
```
Andarist, line 489: I'm not entirely sure if this is correct. According to SCXML:
> A conformant SCXML document must have at least one <state>, <parallel> or <final> child

I've found out that this conditional block is only executed for [the test 530](https://gitlab.com/scion-scxml/test-framework/-/blob/b46a10a1c3a3b1ca5c5cb4bb44ddb5c785611f41/test/w3c-ecma/test530.txml.scxml). And for it we get `scxmlJson` equal to:
```json
{
  "type": "element",
  "name": "content",
  "attributes": {
    "expr": "Var1"
  }
}
```
This test actually assigns the SCXML document to context (! 😱 ) and later uses it in `<invoke>` - I'm pretty sure our SCXML converted doesn't support this yet so this test should get disabled for now and those lines here might not be needed (or rather they should be replaced with other logic)

packages/core/test/eventDescriptors.test.ts
davidkpiano: > I'm OK with being more liberal here for other purposes but I think that at least * should be considered illegal in the middle of a string.

Agreed, we can add a warning for that.

packages/core/src/interpreter.ts
davidkpiano: You're right, this should probably be resolved in `send()`.

IIRC the reason for `#_` is because the `src` can be a URL instead of an ID, and this is how it's distinguished.

In the future, in XState, a URL `src` can be converted to an actor with its URL as an ID (XState would not care that it's a URL).

packages/core/test/eventDescriptors.test.ts
davidkpiano: Warning added 👍

PR #1855, opened 2021-01-16
[@xstate/react] useInterpret

davidkpiano:
Supersedes #1514 

Usage:

```js
const App = () => {
  const service = useInterpret(someMachine);

  // ...
}
```

### DIFF
--- a/.changeset/breezy-beds-trade.md
+++ b/.changeset/breezy-beds-trade.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Improved `.matches(value)` inference for typestates containing union types as values.
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,10 +12,6 @@
 [![npm version](https://badge.fury.io/js/xstate.svg)](https://badge.fury.io/js/xstate)
 <img src="https://opencollective.com/xstate/tiers/backer/badge.svg?label=sponsors&color=brightgreen" />
 
-<div class="blm-callout">
-Black lives matter. <a href="https://support.eji.org/give/153413/#!/donation/checkout" target="_blank">Support the Equal Justice Initiative.</a> ✊🏽✊🏾✊🏿 
-</div>
-
 JavaScript and TypeScript [finite state machines](https://en.wikipedia.org/wiki/Finite-state_machine) and [statecharts](https://www.sciencedirect.com/science/article/pii/0167642387900359/pdf) for the modern web.
 
 📖 [Read the documentation](https://xstate.js.org/docs)
@@ -168,9 +164,9 @@ Read [📽 the slides](http://slides.com/davidkhourshid/finite-state-machines) (
 <img src="https://imgur.com/rqqmkJh.png" alt="Light Machine" width="300" />
 
 ```js
-import { Machine } from 'xstate';
+import { createMachine } from 'xstate';
 
-const lightMachine = Machine({
+const lightMachine = createMachine({
   id: 'light',
   initial: 'green',
   states: {
@@ -204,7 +200,7 @@ const nextState = lightMachine.transition(currentState, 'TIMER').value;
 <img src="https://imgur.com/GDZAeB9.png" alt="Hierarchical Light Machine" width="300" />
 
 ```js
-import { Machine } from 'xstate';
+import { createMachine } from 'xstate';
 
 const pedestrianStates = {
   initial: 'walk',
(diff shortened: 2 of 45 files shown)


Andarist (commented):
I still believe that it should be `useMachine`'s core responsibility to aid with selective rendering. 

```js
// we can't even easily initialize this!
// the machine can be lazy (a factory function) so we won't even be able to read `.initialState`
// the logic in useMachine that initialized the state is also a little bit more complex than the property read
// as it also handles rehydration
//
// EDIT: I've missed that `useMachineObserver` returns a service and we can access service.machine
// but using a closure like this is not a common pattern and might be frowned upon by some
const [state, setState] = React.useState(machine.initialState) 

useMachineObserver(machine, {}, (state) => {
  // didn't give us anything
  setState(state)
});
```

This doesn't help us at all because object references will always be new and thus always leading to a rerender:
```js
const [state, setState] = React.useState({
  foo: machine.initialState.context.foo,
  bar: machine.initialState.context.bar,
});

useMachineObserver(machine, {}, (state) => {
  setState({
    foo: state.context.foo,
    bar: state.context.bar
  });
});

```

pseudo-OK, but far from ergonomic (and this only has 2 state slots):
```js

const [foo, setFoo] = React.useState(machine.initialState.context.foo);
const [bar, setBar] = React.useState(machine.initialState.context.bar);

useMachineObserver(machine, {}, (state) => {
  setFoo(state.context.foo)
  setBar(state.context.bar)
});
```
we can dirty-check manually:
```js

const [state, setState] = React.useState({
  foo: machine.initialState.context.foo,
  bar: machine.initialState.context.bar,
});

useMachineObserver(machine, {}, (state) => {
  if (
    state.history.context.foo !== state.context.foo ||
    state.history.context.bar !== state.context.bar
  ) {
    setState({
      foo: state.context.foo,
      bar: state.context.bar,
    });
  }
});
```
or introduce `shallowEqual`:
```js
const [state, setState] = React.useState({
  foo: machine.initialState.context.foo,
  bar: machine.initialState.context.bar,
});

useMachineObserver(machine, {}, (state) => {
  if (
    !shallowEqual(
      {
        foo: state.history.context.foo,
        bar: state.history.context.bar,
      },
      {
        foo: state.context.foo,
        bar: state.context.bar,
      }
    )
  ) {
    setState({
      foo: state.context.foo,
      bar: state.context.bar,
    });
  }
});

```

yes - one could introduce helpers to aid a little bit with this, but at the end of the day this is still quite a bit of boilerplate to be written and one can **easily** mess this up. Given that we should be aiming to help to solve day-to-day problems I don't think this approach is delivering that.

davidkpiano:
> I still believe that it should be useMachine's core responsibility to aid with selective rendering.

Strongly disagree here. The `useMachine` hook should be made simpler and have the same responsibility: provide the `state` (and `send` and `service`) whenever it changes. We want to avoid adding multiple responsibilities to the `useMachine` hook IMO.

Regardless, `useMachineObserver` does not preclude us from adding a `filter` option in `useMachine` - in fact, it makes it easier to implement, if we wanted to.

Andarist:
> We want to avoid adding multiple responsibilities to the useMachine hook IMO.

I think its main purpose is to allow integrating React with XState as seamlessly as possible. Unfortunately, React doesn't have any builtin mechanism for granular updates and people have started to complain about perf problems when using `useMachine` (which, of course, doesn't affect all usages of `useMachine`). 

I wish React would come with better ways to notify "subscribing" components about changes but that's not a reality we are facing and I highly doubt this will change any time soon. A lot of other libraries (like `react-redux`) have established "selectors" as a common pattern for manual optimizations - even React itself is embracing similar patterns with things like `useMemo`, `memo`, `shouldComponentUpdate`, etc. 

It's fair to say that we are XState-centric and this is "just" the integration package with React but I think that we should balance the needs of both sides of the story, ty to treat both XState and React as equals in this package but, of course, without sacrificing core principles behind XState. And React's part clearly needs some way to optimize the perf/update frequency which doesn't strictly go against any core principles of XState. Trying to push it away from `@xstate/react` will only hurt our consumers and make them come up with custom solutions. It seems that it's our responsibility to guide them to what we believe is best in perf-constrained situations.

From what I know - this work here has been prompted by this exact problem (perf) and even if this refactoring is not solely about it I think it should enable us to tackle the perf issue. Or to rephrase - it might not be the enabler for that as the problem is somewhat orthogonal but I think it's best to discuss both things together to land on a solution that takes both things into consideration - otherwise, we risk providing non-ideal APIs.

PR #1865, opened 2021-01-20
Improved `.matches(value)` inference for typestates containing union types as values

Andarist:
reported on Discord: https://discord.com/channels/795785288994652170/799416943324823592/801198150894223411

### DIFF
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -291,7 +291,11 @@ export class State<
   public matches<TSV extends TTypestate['value']>(
     parentStateValue: TSV
   ): this is State<
-    (TTypestate extends { value: TSV } ? TTypestate : never)['context'],
+    (TTypestate extends any
+      ? { value: TSV; context: any } extends TTypestate
+        ? TTypestate
+        : never
+      : never)['context'],
     TEvent,
     TStateSchema,
     TTypestate
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -61,7 +61,7 @@ export function isState<
 
 export function bindActionToState<TC, TE extends EventObject>(
   action: ActionObject<TC, TE>,
-  state: State<TC, TE>
+  state: State<TC, TE, any, any>
 ): ActionObject<TC, TE> {
   const { exec } = action;
   const boundAction: ActionObject<TC, TE> = {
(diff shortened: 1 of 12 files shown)


packages/core/src/State.ts
```diff
@@ -291,7 +291,11 @@ export class State<
   public matches<TSV extends TTypestate['value']>(
     parentStateValue: TSV
   ): this is State<
-    (TTypestate extends { value: TSV } ? TTypestate : never)['context'],
```
Andarist, line 294: this didn't work for this particular case because:
```ts
type T = { value: "active" | { other: "one" } };
type Result = T extends { value: "active" } ? true : false // results in false
```
We should actually flip this test and test something like this:
```ts
type T = { value: "active" | { other: "one" } };
type Result = { value: "active" } extends T ? true : false // results in true
```
However, this doesn't solve the problem because:
```ts
type T = { value: "active" | { other: "one" }, context: { count: number } };
type Result = { value: "active" } extends T ? true : false // results in false
```
This one is fairly easy to fix, just add `context: any` to the left side:
```ts
type T = { value: "active" | { other: "one" }, context: { count: number } };
type Result = { value: "active", context: any } extends T ? true : false // results in true
```
Ok, we are close - we now need to create a utility type that is able to filter a union:
```ts
type T = 
  | { value: "active" | { other: "one" }, context: { count: number } }
  | { value: "inactive" | { other: "one" }, context: undefined };

type ComputeResult<Typestate, TSV> = { value: TSV, context: any } extends Typestate ? Typestate : never;

type Result = ComputeResult<T, "active">
```
This doesn't work - it doesn't filter the input union type but it preserves it which means that the truthy branch of the conditional type has been chosen which is a good thing.

Why is that? Well, this doesn't distribute the union - it turns out that unions are only distributed by conditional types if the union type is "naked" in the test, and to achieve that here we must wrap this in a `T extends any`:
```ts
type T = 
  | { value: "active" | { other: "one" }, context: { count: number } }
  | { value: "inactive" | { other: "one" }, context: undefined };

type ComputeResult<Typestate, TSV> = Typestate extends any ? { value: TSV, context: any } extends Typestate ? Typestate : never : never;

type Result = ComputeResult<T, "active">
```

This works 🎉

davidkpiano:
Some type errors, otherwise LGTM

Andarist:
Huh, gonna take a look - was running the full test suite locally and everything has passed 🤔

davidkpiano (approved):

PR #1892, opened 2021-01-29
@xstate/fsm: Add internal transition

cblanc:
Adds internal state transitions to @xstate/fsm via `internal` attribute

Excludes internal transition definition via `{target: undefined}` or omitting `target` to avoid a backward breaking change. Currently fsm produces an external transition if undefined target

### DIFF
--- a/packages/xstate-fsm/src/index.ts
+++ b/packages/xstate-fsm/src/index.ts
@@ -177,7 +177,12 @@ export function createMachine<
             return createUnchangedState(value, context);
           }
 
-          const { target = value, actions = [], cond = () => true } =
+          const {
+            target = value,
+            actions = [],
+            cond = () => true,
+            internal = false
+          } =
             typeof transition === 'string'
               ? { target: transition }
               : transition;
@@ -185,7 +190,11 @@ export function createMachine<
           if (cond(context, eventObject)) {
             const nextStateConfig = fsmConfig.states[target];
             const allActions = ([] as any[])
-              .concat(stateConfig.exit, actions, nextStateConfig.entry)
+              .concat(
+                internal ? [] : stateConfig.exit,
+                actions,
+                internal ? [] : nextStateConfig.entry
+              )
               .filter((a) => a)
               .map<StateMachine.ActionObject<TContext, TEvent>>((action) =>
                 toActionObject(action, (machine as any)._options.actions)
--- a/packages/xstate-fsm/test/fsm.test.ts
+++ b/packages/xstate-fsm/test/fsm.test.ts
@@ -193,6 +193,64 @@ describe('@xstate/fsm', () => {
       expect(nextState.context.go).toBeFalsy();
     }
   });
+
+  describe('internal transitions', () => {
+    interface InternalContext {
+      didEntry: boolean;
+      didExit: boolean;
+      didAction: boolean;
+    }
+    type InternalEvents = { type: 'EXTERNAL' } | { type: 'INTERNAL' };
+    interface InternalStates {
+      value: 'initial';
+      context: InternalContext;
+    }
+
+    const definition = {
+      initial: 'initial',
+      context: {
+        didEntry: false,
+        didExit: false,
+        didAction: false
+      },
+      states: {
+        initial: {
+          entry: assign({ didEntry: true }),
+          exit: assign({ didExit: true }),
+          on: {
+            EXTERNAL: {
+              target: 'initial',
+              actions: assign({ didAction: true })
+            },
+            INTERNAL: {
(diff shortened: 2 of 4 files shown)


packages/xstate-fsm/src/index.ts
```diff
@@ -177,15 +177,22 @@ export function createMachine<
...
               ? { target: transition }
               : transition;
 
           if (cond(context, eventObject)) {
             const nextStateConfig = fsmConfig.states[target];
             const allActions = ([] as any[])
-              .concat(stateConfig.exit, actions, nextStateConfig.entry)
+              .concat(internal ? [] : stateConfig.exit)
+              .concat(actions)
+              .concat(internal ? [] : nextStateConfig.entry)
```
davidkpiano, line 195:
```suggestion
              .concat(
                internal ? [] : stateConfig.exit,
                actions,
                internal ? [] : nextStateConfig.entry)
```

cblanc: Updated PR

jamesopstad:
Should this become the default behaviour in the next major version release?

cblanc:
Updated the branch a couple times for @davidkpiano change request and fix test description

packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -193,6 +193,64 @@ describe('@xstate/fsm', () => {
...
+          entry: assign({ didEntry: true }),
+          exit: assign({ didExit: true }),
+          on: {
+            EXTERNAL: {
+              target: 'initial',
+              actions: assign({ didAction: true })
+            },
+            INTERNAL: {
+              actions: assign({ didAction: true }),
+              internal: true
```
Andarist, line 227: Actually - if anything, this is a targetless transition. And for those being internal or external doesn't matter much - no states should be exited/entered for those. Changing this for `@xstate/fsm` would be IMHO a breaking change at this point in time though.

cblanc: Thanks. I incorrectly assumed this shorthand of specifying "target: *self*". I'll need to update my fsm project 😅

I've updated the test. Will become problematic if targetless transitions do make its way into `@xstate/fsm`.

Andarist: `@xstate/fsm` is meant to be a minimal implementation and I'm a little bit hesitant about introducing internal/external transitions concept here - I think we should aim to keep a single type supported. I wouldn't necessarily change the logic to use internal transitions semantic at all times here but there is a lot to consider for that and I would expect this to be decided after we settle this discussion: https://github.com/davidkpiano/xstate/discussions/1829

OTOH - adding proper support for targetless transitions seems like something we should definitely do.

Those are obviously just my 2 cents - not a final decision of any sort

cblanc: Thanks for the perspective. I appreciate the need to keep this lib conceptually minimal. However I do wish xstate hadn't filled my mind with all these useful concepts in the excellent guides before I laid my eyes on fsm 😀

Just to add a single uninformed pov in favour of this. This ~50 byte inclusion allows us to delete code and simplify our projects. Currently need to pull the full ~50kb xstate for internal transitions.

It looks as though targetless transitions will be a backwards incompatible change. I can PR that if it can get merged?

Andarist: I would consider support for targetless transitions an addition so it should be safe to add this in the current version. If those would get merged in & released - would you still need internal transitions?

Also an additional note: the current implementation here is not quite correct for internal transitions because it doesn't check the relation between current & next states - `internal: true` is ignored (acting like external transition) if you target a state that doesn't overlap with the source state.

cblanc: Either is fine. I'm going by the xstate docs and they appear to serve the same purpose. The finer conceptual differences are very much lost on me

Also, thanks for spotting the issue. I had a feeling I may not have fully grasped the nuances. Do you have any reading suggestions to get a firm understanding of targetless transitions before I attempt a PR?

In the meantime, I'll close this PR in favour of tackling this via targetless transitions

Andarist: As far as I understand targetless transitions are only supposed to execute their actions and nothing more really.

As to internal transitions - if the target state is within the source state (the one that has a particular internal transition definition on it) then internal transitions do not reenter their source state whereas external transitions reenter that state. If the mentioned precondition is not satisfied (target state is not contained within the source state) then the external and internal transitions behave exactly the same. Reentering means exiting and entering (so both exit and entry actions for that state are executed)

PR #1897, opened 2021-01-29
Remove React double import

hosein2398:


davidkpiano:
Thanks, can you add a quick changeset (patch)?

Run `yarn changeset` at the root. Should be easy from there!

hosein2398:
IDK if I'm right about this. If there's any problem, let me know. 
Thanks

Andarist:
@davidkpiano this is not something that has any impact on end-users so I don't believe this deserves a changeset - it can be freely removed from this PR. It's like a stylistic change - which we wouldn't like to list in our changelogs, right? 😉

davidkpiano:
> @davidkpiano this is not something that has any impact on end-users so I don't believe this deserves a changeset - it can be freely removed from this PR. It's like a stylistic change - which we wouldn't like to list in our changelogs, right? 😉

Sure, that makes sense.

PR #1908, opened 2021-02-03
Update immer

davidkpiano:
Will merge once tests pass


packages/xstate-immer/package.json
```diff
@@ -37,11 +37,11 @@
   },
   "dependencies": {},
   "peerDependencies": {
-    "immer": "^6.0.3",
+    "immer": "^8.0.1",
```
Andarist, line 40: This will require a changeset for `@xstate/immer` with at least a `"minor"` bump (can be minor since the package is still not on `>=1`)

PR #1911, opened 2021-02-05
[@xstate/fsm] Add support for targetless transitions

davidkpiano:
This PR adds support for transitions with `undefined` targets, which will not exit/re-enter the state.

```js
// ...
someState: {
  entry: [/* ... */],
  exit: [/* ... */],
  on: {
    SOME_EVENT: {
      // undefined target - will not exit/re-enter
      actions: [/* ... */]
    }
  }
}
// ...
```

### DIFF
--- a/.changeset/afraid-eggs-wonder.md
+++ b/.changeset/afraid-eggs-wonder.md
@@ -1,5 +1,5 @@
 ---
-'@xstate/fsm': patch
+'@xstate/fsm': minor
 ---
 
 Transitions with `undefined` targets will no longer exit and re-enter the state, which means that `exit` and `entry` actions will not be executed:
--- a/.changeset/afraid-eggs-wonder.md
+++ b/.changeset/afraid-eggs-wonder.md
@@ -1,0 +1,20 @@
+---
+'@xstate/fsm': patch
+---
+
+Transitions with `undefined` targets will no longer exit and re-enter the state, which means that `exit` and `entry` actions will not be executed:
+
+```js
+// ...
+someState: {
+  entry: [/* ... */],
+  exit: [/* ... */],
+  on: {
+    SOME_EVENT: {
+      // undefined target - will not exit/re-enter
+      actions: [/* ... */]
+    }
+  }
+}
+// ...
+```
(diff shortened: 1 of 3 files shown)


.changeset/afraid-eggs-wonder.md
```diff
@@ -0,0 +1,20 @@
+---
+'@xstate/fsm': patch
```
Andarist, line 2: IMHO this is sort-of a new feature. Personally, I would use a minor bump here

```suggestion
'@xstate/fsm': minor
```

packages/xstate-fsm/src/index.ts
```diff
@@ -170,22 +170,31 @@ export function createMachine<
...
             return createUnchangedState(value, context);
           }
 
-          const { target = value, actions = [], cond = () => true } =
+          const { target, actions = [], cond = () => true } =
             typeof transition === 'string'
               ? { target: transition }
               : transition;
 
+          const isInternal = target === undefined;
```
Andarist, line 187: Those shouldn't be confused with internal transitions - internal transitions in SCXML and XState can have targets

```suggestion
          const isTargetless = target === undefined;
```

packages/xstate-fsm/src/index.ts
```diff
@@ -170,22 +170,31 @@ export function createMachine<
...
+
           if (cond(context, eventObject)) {
-            const nextStateConfig = fsmConfig.states[target];
+            const nextStateConfig = fsmConfig.states[target ?? value];
```
Andarist, line 190: this is somewhat a dead code - the resulting variable won't ever be used if we have a target (I suspect that you have used `??` fallback only to satisfy types as this is now declared before it's actually used), so maybe to avoid this you could do this:
```ts
if (cond(context, eventObject)) {
  const nextStateConfig = fsmConfig.states[target ?? value];
  const allActions =
    target === undefined
      ? actions
      : ([] as any[])
          // Only add exit and entry actions
          .concat(stateConfig.exit, actions, fsmConfig.states[target].entry)
          .filter((a) => a)
          .map<StateMachine.ActionObject<TContext, TEvent>>((action) =>
            toActionObject(action, (machine as any)._options.actions)
          );
}
```

which would also differentiate `target === undefined` branching a little bit more explicitly

Andarist (commented):
As I don't believe we are dealing with internal transitions here I would also adjust the PR title to avoid that term

Andarist (approved):

cblanc:
Thanks! 🙏

PR #1915, opened 2021-02-06
[@xstate/react] Add useSelector() and useInterpret()

davidkpiano:
```js
import { useInterpret, useSelector } from 'xstate';

const App = () => {
  const service = useInterpret(machine);
  const count = useSelector(service, state => state.context.count);

  return <div>{count}</div>
}
```

Also works for actors:

```js
const count = useSelector(anyActor, snapshot => snapshot.count);
```


packages/xstate-react/src/useMachine.ts
```diff
@@ -114,141 +96,86 @@ export function useMachine<
...
-      ...machine.context,
-      ...context
-    } as TContext);
+  const service = useInterpret(getMachine, options);
 
-    return [
-      machineWithConfig,
-      interpret(machineWithConfig, { deferEvents: true, ...interpreterOptions })
-    ];
+  useEffect(() => {
```
Andarist, line 101: Reposting from https://github.com/davidkpiano/xstate/pull/1855#discussion_r559147349

The subscription has been happening in the layout effect previously - not anymore though. After more thinking about this in recent months I'm not totally sure which one is best, but I'm inclined to say that probably - for our use case - the layout effect is better. At least for the time being.

packages/xstate-react/src/useSelector.ts
```diff
@@ -0,0 +1,35 @@
...
+export function useSelector<
+  TActor extends ActorRef<any, any>,
+  T,
+  TEmitted = TActor extends Subscribable<infer Emitted> ? Emitted : never
+>(
+  actor: TActor,
+  selector: (emitted: TEmitted) => T,
+  compare: (a: T, b: T) => boolean = (a, b) => a === b,
+  getSnapshot: (a: TActor) => TEmitted = (a) =>
+    a instanceof Interpreter
```
Andarist, line 15: I would prefer if we could not rely on the `instanceof` - it might create very hard to debug problems for users. That's why we've created a special property on the StateNode in the past that has allowed us to differentiate those objects. 

A recent Twitter thread mentioning this problem: https://twitter.com/_jayphelps/status/1356992397243256837

packages/xstate-react/src/useMachine.ts
```diff
@@ -114,141 +96,86 @@ export function useMachine<
...
-  const [state, setState] = useState<State<TContext, TEvent, any, TTypestate>>(
-    () => {
-      // Always read the initial state to properly initialize the machine
-      // https://github.com/davidkpiano/xstate/issues/1334
-      const { initialState } = resolvedMachine;
-      return rehydratedState ? State.create(rehydratedState) : initialState;
-    }
-  );
-
   const effectActionsRef = useRef<
```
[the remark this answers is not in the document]
davidkpiano: That's fine, I can make that change.

packages/xstate-react/src/useMachine.ts
```diff
@@ -114,141 +96,86 @@ export function useMachine<
...
+    service.subscribe((s) => {
+      // Only change the current state if:
+      // - the incoming state is the "live" initial state (since it might have new actors)
+      // - OR the incoming state actually changed.
+      //
+      // The "live" initial state will have .changed === undefined.
+      const initialStateChanged =
+        s.changed === undefined && Object.keys(s.children).length;
+
+      if (s.changed || initialStateChanged) {
```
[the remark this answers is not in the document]
davidkpiano: This actually isn't an issue because the first `state` from `[state, send] = useMachine(...)` is always the initial/restored state, and _not_ the subscribed state. Added a test to show this.

packages/xstate-react/src/useMachine.ts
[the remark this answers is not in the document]
davidkpiano: Made as custom hook: `useReactEffectActions`

packages/xstate-react/src/useMachine.ts
davidkpiano: Define "better"?

davidkpiano:
@Andarist All comments resolved 😬 🙏

packages/xstate-react/src/useInterpret.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    return interpret(machineWithConfig, {
+      deferEvents: true,
+      ...interpreterOptions
+    });
+  });
+
+  useIsomorphicLayoutEffect(() => {
+    if (observerOrListener) {
+      service.subscribe(toObserver(observerOrListener));
```
Andarist, line 84: This subscription will suffer from a stale closure problem - we probably should avoid that. Maybe the easiest way to handle this would be to:
1. only allow function listener right now - they are the most useful anyway
2. use a pattern like this:
```js
const listenerRef = React.useRef(listener)
useIsomorphicLayoutEffect(() => (listenerRef.current = listener))

  service.subscribe(state => {
    const currentListener = listenerRef.current
    currentListener && currentListener(state)
  })
  // ...
})
```

The first part of this React dance (😰) is what I've encapsulated in the [`useLatest`](https://github.com/Andarist/use-latest) helper as I had already handled this just too many times.

There is also a quite good, recent, read touching the callbacks+refs subject: https://dev.to/yelouafi/a-simple-rule-for-using-callbacks-in-react-4jah

packages/xstate-react/src/useInterpret.ts
```diff
@@ -0,0 +1,110 @@
...
+    return () => {
+      service.stop();
+    };
+  }, []);
+
+  // Make sure machine options are kept updated when they change.
+  // This mutation assignment is safe because the service instance is only used
+  // in one place -- this hook's caller.
+  useEffect(() => {
+    Object.assign(service.machine.options, {
```
Andarist, line 99: This introduces a small behavioral change - previously some of the things were updated using separate `Object.assign` calls, so a particular implementation type was **merged** with what was already held in the machine whereas now things are **swapped out**. 

I'm actually not sure which way is better and I think we already have some related issues to that (albeit for raw Machine APIs and not the React integration). We should establish how things should be handled in v5 - there are also some typing issues related to what we allow right now. I would actually love to discuss this during the upcoming call - let's add this to the agenda. 

While this stays somewhat undecided - or not fully speced~ - I would say that it's better to preserve the current semantics for the time being and possibly change this in the future. So this would become:
```js
Object.assign(service.machine.options.guards, guards)
Object.assign(service.machine.options.actions, actions)
Object.assign(service.machine.options.services, services)
Object.assign(service.machine.options.delays, delays)
```

Oh! And this is especially important because the version in this PR actually **removes** implementation types:
```js
var options = { actions: { doSmth: () => {} } }
Object.assign(options, { actions: undefined })
// oops, options don't have any `actions` now.
```
This definitely deserves to get a regression test as this is a pretty common and important scenario - to not provide any implementations using `useInterpret`/`useMachine` options and only provide them using a fully static definition.

packages/xstate-react/src/useInterpret.ts
```diff
@@ -0,0 +1,110 @@
...
+      guards,
```
Andarist, line 100: Looking at the implementation of XState, I don't think updating `guards` will actually work - guards are currently destructured in `formatTransition`:
https://github.com/davidkpiano/xstate/blob/a9ab426b53e416c776af514b7b3a4f6ef3f0d855/packages/core/src/StateNode.ts#L1826
and "embedded" in the transition definition here:
https://github.com/davidkpiano/xstate/blob/a9ab426b53e416c776af514b7b3a4f6ef3f0d855/packages/core/src/StateNode.ts#L1833
https://github.com/davidkpiano/xstate/blob/a9ab426b53e416c776af514b7b3a4f6ef3f0d855/packages/core/src/utils.ts#L495-L501

and the unpackaged `.predicate` is used here in the `evaluateGuard`:
https://github.com/davidkpiano/xstate/blob/a9ab426b53e416c776af514b7b3a4f6ef3f0d855/packages/core/src/utils.ts#L659-L665

packages/xstate-react/src/useReactEffectActions.ts
```diff
@@ -0,0 +1,81 @@
...
+    >
+  >([]);
+  const layoutEffectActionsRef = useRef<
+    Array<
+      [ReactActionObject<TContext, TEvent>, State<TContext, TEvent, any, any>]
+    >
+  >([]);
+
+  useIsomorphicLayoutEffect(() => {
+    const sub = service.subscribe((currentState) => {
```
Andarist, line 23: Because of the implemented split of things, some initial React actions will be missed - you could end up with such a situation even more easily than in the test case provided here:
https://github.com/davidkpiano/xstate/pull/1922/files#diff-753cb411217f73964488b6bea771171aaef83d7e7f66c3717d5ae3103f90bdb3R733

It would only require:
- having an initial React action
- sending an event within initial actions

This is an edge case so I don't think there is much need to fix this right now but we probably should think about how we could fix this in the future.

Random thought: after more thinking about "React actions" I'm starting to think that this is really quirky and maybe we should revisit this in the future altogether. React's model is "synchronize to state" and maybe this is just what we should stick to in this regard.

packages/xstate-react/src/useSelector.ts
```diff
@@ -0,0 +1,39 @@
...
+  useEffect(() => {
+    const sub = actor.subscribe((emitted) => {
+      const nextSelected = selector(emitted);
+      if (!compare(selected, nextSelected)) {
+        setSelected(nextSelected);
+      }
+    });
+
+    return () => sub.unsubscribe();
+  });
```
Andarist, line 36: Is the intention here to resubscribe on each React's commit? I'm not saying it's incorrect - just wondering if it was a deliberate decision.

davidkpiano: It is not; will change.

packages/xstate-react/src/useInterpret.ts
davidkpiano: Ref might not be needed:

```jsx
    let sub;
    if (observerOrListener) {
      sub = service.subscribe(toObserver(observerOrListener));
    }

    return () => {
      sub?.unsubscribe();
    };
  }, [observerOrListener]);
```

packages/xstate-react/src/useInterpret.ts
```diff
@@ -0,0 +1,117 @@
...
+  useIsomorphicLayoutEffect(() => {
+    let sub;
+    if (observerOrListener) {
+      sub = service.subscribe(toObserver(observerOrListener));
+    }
+
+    return () => {
+      sub?.unsubscribe();
+    };
+  }, [observerOrListener]);
```
Andarist, line 91: Given that `useMachine` uses an inline subscriber right now this effect will be refired for each commit of components using `useMachine`. I don't see any functional difference - we'll only be doing some minor extra work that shouldn't matter much and memoizing the observer also has some minor cost, so... Just something to keep in mind if it raises any problems after the release.

packages/xstate-react/src/useSelector.ts
```diff
@@ -0,0 +1,41 @@
...
+  useEffect(() => {
+    const sub = actor.subscribe((emitted) => {
+      const nextSelected = selector(emitted);
+      if (!compare(selected, nextSelected)) {
+        setSelected(nextSelected);
+      }
+    });
+
+    return () => sub.unsubscribe();
+  }, [selector]);
```
Andarist, line 38: This dependency array suffers from 2 problems.

A minor thing is that this doesn't depend on `compare` and people can provide custom ones. In most real-life scenarios comparator should be pure, but it still might be worth putting it in this dependency array.

The second thing is much worse and defeats the main purpose of this PR (in non that unlikely scenarios) which is being able to optimize rerenders using `useSelector`. The created subscriber doesn't depend on `selected` and thus it suffers from the stale closure problem.

davidkpiano: Fixed this, I believe - please see a4e0c0a

packages/xstate-react/src/useInterpret.ts
davidkpiano: Took care of this 👍 7f4caf3

.changeset/polite-turkeys-sin.md
```diff
@@ -0,0 +1,15 @@
+---
+'@xstate/react': minor
+---
+
+New hook: `useSelector(actor, selector)`, which subscribes to `actor` and returns the selected state derived from `selector(snapshot)`:
```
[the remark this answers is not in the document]
davidkpiano: There is a difference though:

- **emitted** = most recent subscribed-to value
- **snapshot** = last known value

A snapshot might not necessarily be the same as an emitted value. Emitted values come from subscriptions, whereas snapshots come from some persistence layer.

packages/xstate-react/src/useSelector.ts
```diff
@@ -0,0 +1,50 @@
...
+  useEffect(() => {
+    const updateSelectedIfChanged = (nextSelected: T) => {
+      if (!compare(selectedRef.current, nextSelected)) {
+        setSelected(nextSelected);
+        selectedRef.current = nextSelected;
+      }
+    };
+
+    const initialSelected = selector(getSnapshot(actor));
+    updateSelectedIfChanged(initialSelected);
```
Andarist, line 39: good thinking - I've missed this 👍

.changeset/polite-turkeys-sin.md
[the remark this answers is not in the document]
Andarist: I see the difference when it comes to the source of both values but in general, we assume that they are the same, don't we? At least - they are of the same type:
- [`getSnapshot` has to return `TEmitted`](https://github.com/davidkpiano/xstate/blob/7f4caf3276f2dcba4f91ae53c9c9a21cb8c58705/packages/xstate-react/src/useSelector.ts#L20)
- `selector` receives both a [snapshot](https://github.com/davidkpiano/xstate/blob/7f4caf3276f2dcba4f91ae53c9c9a21cb8c58705/packages/xstate-react/src/useSelector.ts#L27) and an [emitted](https://github.com/davidkpiano/xstate/blob/7f4caf3276f2dcba4f91ae53c9c9a21cb8c58705/packages/xstate-react/src/useSelector.ts#L42)

This is just a subtle difference, of course, but using 2 terms for somewhat the same thing might be confusing for people. Maybe the `emitted` should be renamed to `emittedSnapshot`? Unless you believe that they are completely not the same - I would love to discuss this more to align our mental models.

The main reason why I've brought this up here is that we can read in this changeset that:
> subscribes to `actor` and returns the selected state derived from `selector(snapshot)`

And if `snapshot` & `emitted` are not the same then this description is slightly incorrect, isn't it?

Either way - this shouldn't block this PR but you might want to adjust the changeset's content if you consider my point valid(-ish).

Andarist (approved):

PR #1922, opened 2021-02-10
Fixed an issue with synchronous initial updates not updating useMachine

Andarist:


packages/xstate-react/src/useMachine.ts
```diff
@@ -95,23 +94,19 @@ export function useMachine<
...
-      // The "live" initial state will have .changed === undefined.
-      const initialStateChanged =
-        s.changed === undefined && Object.keys(s.children).length;
-
-      if (s.changed || initialStateChanged) {
-        setState(s);
-      }
-    });
-  }, [service]);
+    if (s.changed || initialStateChanged) {
```
Andarist, line 106: it would be great if we could kill the `.changed` in v5

davidkpiano: Why?

davidkpiano (approved):

PR #1930, opened 2021-02-12
Fixes 'final' state node names not showing in test plan description

jimwheaton:
Fixes an issue where `'final'` state nodes were not showing up in the test plan description. For example, given the state node `success: { type: 'final' }`, the description was

```
 reaches states:  ({ some_context_val: 0 })
 ```
 
The description is now:

```
 reaches state: "success"  ({ some_context_val: 0 })
 ```


davidkpiano:
Beautiful, thanks!

PR #1947, opened 2021-02-18
add typestate type to service returned from react/fsm useMachine

lukekarrys:
This PR stems from the discussion I started here: https://github.com/davidkpiano/xstate/discussions/1944.

I was able to create a failing test and I confirmed that passing the Typestate to the returned service from `useMachine` fixes the issue.

### DIFF
--- a/packages/xstate-react/src/fsm.ts
+++ b/packages/xstate-react/src/fsm.ts
@@ -36,8 +36,8 @@ export function useMachine<
   }
 ): [
   StateMachine.State<TContext, TEvent, TState>,
-  StateMachine.Service<TContext, TEvent>['send'],
-  StateMachine.Service<TContext, TEvent>
+  StateMachine.Service<TContext, TEvent, TState>['send'],
+  StateMachine.Service<TContext, TEvent, TState>
 ] {
   if (process.env.NODE_ENV !== 'production') {
     const [initialMachine] = useState(stateMachine);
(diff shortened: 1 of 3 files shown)


packages/xstate-react/src/fsm.ts
```diff
@@ -88,11 +88,9 @@ export function useService<
...
   StateMachine.Service<TContext, TEvent, TState>
 ] {
-  const subscription: Subscription<StateMachine.State<
-    TContext,
-    TEvent,
-    TState
-  >> = useMemo(() => {
+  const subscription: Subscription<
+    StateMachine.State<TContext, TEvent, TState>
+  > = useMemo(() => {
```
lukekarrys, line 93: I thought about excluding this change but it was generated by the pre-commit hook, so I wasn't sure what I should do. Happy to push a commit reverting this line and not run the pre-commit hook.

packages/xstate-react/src/fsm.ts
```diff
@@ -36,8 +36,8 @@ export function useMachine<
   }
 ): [
   StateMachine.State<TContext, TEvent, TState>,
-  StateMachine.Service<TContext, TEvent>['send'],
-  StateMachine.Service<TContext, TEvent>
+  StateMachine.Service<TContext, TEvent, TState>['send'],
```
lukekarrys, line 39: I added `TState` to this line too to be consistent with the types I saw elsewhere in the codebase, but I couldn't see any difference with it included or not.

davidkpiano (approved):

PR #1955, opened 2021-02-23
[core] Modeling events with createModel()

davidkpiano:
Making this PR to discuss and see how we can improve this. This PR provides type inference for **events** in `createModel()` that are specified as "event creators":

```ts
const userModel = createModel(
  {
    name: 'David',
    age: 30
  },
  {
    updateName: (value: string) => ({ value }),
    updateAge: (value: number) => ({ value }),
    anotherEvent: () => ({})
  }
);

// Example of an externally-defined assign action
const assignName = userModel.assign(
  {
    name: (_, event) => {
      return event.value;
    }
  },
  'updateName'
);

const machine = createMachine<
  typeof userModel['types']['context'],
  typeof userModel['types']['events']
>({
  context: userModel.initialContext,
  initial: 'active',
  states: {
    active: {
      on: {
        updateName: {
          actions: assignName
        },
        updateAge: {
          actions: userModel.assign((_, e) => {
            return {
              age: e.value
            };
          })
        }
      }
    }
  }
});

let updatedState = machine.transition(
  undefined,
  userModel.events.updateName('Anyone')
);

expect(updatedState.context.name).toEqual('Anyone');

updatedState = machine.transition(
  updatedState,
  userModel.events.updateAge(42)
);

expect(updatedState.context.age).toEqual(42);
```

This is a little awkward, sure:

```ts
const machine = createMachine<
  typeof userModel['types']['context'],
  typeof userModel['types']['events']
>({/* ... */})
```

But the important part is this:

```ts
let updatedState = machine.transition(
  undefined,
  userModel.events.updateName('Anyone')
);

updatedState = machine.transition(
  updatedState,
  userModel.events.updateAge(42)
);
```

The second argument to `createModel(initialCtx, eventCreators)` enables two things:

- Events to be easily created and strongly typed using these convenient event creators, and...
- Event types to be **inferred** automatically from these event creators! 🎉 

Let me know your thoughts!

### DIFF
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -69,6 +69,14 @@ type GetEventsFromEventCreatorMap<T extends EventCreatorMap<any>> = Distribute<
   T
 >;
 
+interface ModelCreators<
+  _TContext,
+  TEvent extends EventObject,
+  TEM extends EventCreatorMap<TEvent>
+> {
+  events: TEM;
+}
+
 export function createModel<TContext, TEvent extends EventObject>(
   initialContext: TContext
 ): Model<TContext, TEvent, never>;
@@ -76,12 +84,17 @@ export function createModel<
   TContext,
   TEM extends EventCreatorMap<any>,
   TEvent extends EventObject = GetEventsFromEventCreatorMap<TEM>
->(initialContext: TContext, eventCreators: TEM): Model<TContext, TEvent, TEM>;
+>(
+  initialContext: TContext,
+  creators: ModelCreators<TContext, TEvent, TEM>
+): Model<TContext, TEvent, TEM>;
 export function createModel<
   TContext,
   TEM extends EventCreatorMap<any>,
   TEvent extends EventObject = GetEventsFromEventCreatorMap<TEM>
->(initialContext: TContext, eventCreators?) {
+>(initialContext: TContext, creators?) {
+  const eventCreators = creators?.events;
+
   const model: Model<TContext, TEvent, TEM> = {
     initialContext,
     assign,
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -8,10 +8,16 @@ import type {
 } from './types';
 import { mapValues } from './utils';
 
+type AnyFunction = (...args: any[]) => any;
+
+type Cast<A1 extends any, A2 extends any> = A1 extends A2 ? A1 : A2;
+type Compute<A extends any> = { [K in keyof A]: A[K] } & unknown;
+type Prop<T, K> = K extends keyof T ? T[K] : never;
+
 export interface Model<
   TContext,
   TEvent extends EventObject,
-  TEM extends EventCreatorMap<TEvent> = EventCreatorMap<TEvent>
+  TModelCreators = never
 > {
   initialContext: TContext;
   assign: <TEventType extends TEvent['type'] = TEvent['type']>(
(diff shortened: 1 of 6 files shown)


HipsterBrown:
Instead of needing to do:

```ts
const machine = createMachine<typeof myModel['types']['context'], typeof myModel['types']['events']>({});
```

Could there be a function definition for `createMachine` to accept a `ContextModel` type argument that uses those `types` fields?

```ts
export function createMachine<ContextModel, TTypeStates>(config: MachineConfig): StateMachine<ContextModel['types']['context'], any, ContextModel['types']['events'], TTypeStates>;
```

davidkpiano:
> Instead of needing to do:
> 
> ```ts
> const machine = createMachine<typeof myModel['types']['context'], typeof myModel['types']['events']>({});
> ```
> 
> Could there be a function definition for `createMachine` to accept a `ContextModel` type argument that uses those `types` fields?
> 
> ```ts
> export function createMachine<ContextModel, TTypeStates>(config: MachineConfig): StateMachine<ContextModel['types']['context'], any, ContextModel['types']['events'], TTypeStates>;
> ```

Ideally yes, but I don't know if it's possible to have function overloading with different generic types... @Andarist @mattpocock ?

HipsterBrown:
> I don't know if it's possible to have function overloading with different generic types

It looks like the `Machine` function does this currently: https://github.com/davidkpiano/xstate/blob/master/packages/core/src/Machine.ts#L13-L29

Andarist:
> Ideally yes, but I don't know if it's possible to have function overloading with different generic types... @Andarist @mattpocock ?

Yep, totally possible.

eddiewang:
looks great! I have been using createModel in my machines lately to make typing more consistent and it's been working well. would this be compatible with the `createUpdater` fns from Immer?

davidkpiano:
> looks great! I have been using createModel in my machines lately to make typing more consistent and it's been working well. would this be compatible with the `createUpdater` fns from Immer?

It should be! I haven't tested it out, might need to update the types or create an Immer-specific `createModel`.

cbmd:
@davidkpiano how about to implement an array of events acceptable?
Something like....
```ts
import isType from '@sindresorhus/is';
import type { EventObject } from 'xstate';
import type { AssertedTypesUnion } from './types';

export function assertEventType<
  TEvent extends EventObject,
  TExpectedType extends TEventTypeArray | TEventType,
  TEventType extends TEvent['type'],
  TEventTypeArray extends TEventType[]
>(
  event: TEvent,
  expected: TExpectedType,
): asserts event is TEvent & {
  type: AssertedTypesUnion<TExpectedType>;
} {
  const events = isType.array<TExpectedType>(expected) ? expected : [expected];
  const { type } = event;
  if (!events.includes(type as TExpectedType)) {
    throw new Error(
      `Invalid event: expected one of "${events.join(
        ',',
      )}" but got "${type}" instead.`,
    );
  }
}
```
wherever `isType` from `@sindresorhus/is` surely could be eliminated, just copy/pasted from existing project to make it simple...
Any objections against adding that @davidkpiano / @Andarist?

packages/core/src/model.ts
```diff
@@ -6,24 +6,82 @@ import type {
...
+    }
+  : never;
+
+type Expand<T> = T extends infer O ? { [K in keyof O]: O[K] } : never;
+type Distribute<U, V extends Record<string, () => object>> = U extends keyof V &
+  string
+  ? Expand<{ type: U } & ReturnType<V[U]>>
+  : never;
+
+type GetEvents<T extends EventCreatorMap<any>> = Distribute<keyof T, T>;
```
Andarist, line 54: This is quite convoluted right now 🤔 but I also couldn't make it simpler so far in a way that would satisfy the `EventObject` constraints in places where this type is used. A simpler version (which only works for `GetEvents` used on its own):
```ts
type GetEvents<T extends EventCreatorMap<any>> = {
    [K in keyof T]: Expand<{ type: K } & ReturnType<T[K]>> 
}[keyof T]
```
[TS playground](https://www.typescriptlang.org/play?ssl=15&ssc=11&pln=13&pc=1#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDJXJgCeADhAFzIDOYUoA5uQL7nk6jNJnABhKBDhgA9lACycegB4AKhiwkIAD0ggAJqxGa8hYgD5kAXjKVkAbQDWEWslDJ1osPYDkQiD4AuoEsABR2VAB00XBQXKwscCC09oF2AJTWltgAtsBgahrgyABkZDQMzMjOrrwANMh+lT7mANx8Av5o2vRJ+mqWNqrIOnqGbiAw0MjYyAD85fYA0hPVLjIwMyEzy4HIvMgsIBCYUO2ClcgA4hBgRWCsaiO6WOP3ElKyCkrKSbTmg1s1AcK3cNQ2Hm2qB6fWUpAqjBYKwOZQASrcAK5QECqSpqXYA5B8JzrTaqNIXYSqCDsazXW73R4UYEwGQyMJwFjsTggLiZKyWeGctgcbgNXAsXBsgA2UhA+zqEWQuFiYQlyplcoaCCOGJyuGg-MFyG0tAAXiwfCqzdw4D59nxzOQgA)

packages/core/src/model.ts
```diff
@@ -6,24 +6,82 @@ import type {
...
+
 export function createModel<TContext, TEvent extends EventObject>(
```
Andarist, line 56: q: what is the intention here - is the model supposed to restrict accepted `TEvent` for assigners when `eventCreators` are used? previously there were no such restrictions (as obviously `eventCreators` didn't exist) so assigners could be used with "arbitrary" `TEvent` that could just be configured through the generic. So this part is being "lost" - which isn't necessarily a bad thing though.

packages/core/src/model.ts
```diff
@@ -6,24 +6,82 @@ import type {
   ExtractEvent,
   EventObject
 } from './types';
+import { mapValues } from './utils';
 
-export interface ContextModel<TContext, TEvent extends EventObject> {
+export interface ContextModel<
+  TContext,
+  TEvent extends EventObject,
+  TEM extends EventCreatorMap<TEvent> = EventCreatorMap<TEvent>
```
Andarist, line 14: I'm unsure if this should have a default here - if one currently has smth like this used anywhere:
```ts
type A = ContextModel<{ count: number }, { type: 'inc' } | { type: 'dec' }>
```
then they would suddenly get a guarantee that there is an `events` property available on `A` - whereas this is not true. It's only defined if the `ContextModel` comes from the `createModel` call with `eventCreators` provided.

To make this more correct I would try to restructure this to accept already computed `FullEventCreatorMap` and default this to `never`.  This way lone `ContextModel` usages wouldn't get this property but `createModel` could enhance the model with it.

packages/core/src/model.ts
```diff
@@ -6,24 +6,82 @@ import type {
...
   assign: <TEventType extends TEvent['type'] = TEvent['type']>(
     assigner:
       | Assigner<TContext, ExtractEvent<TEvent, TEventType>>
       | PropertyAssigner<TContext, ExtractEvent<TEvent, TEventType>>,
     eventType?: TEventType
   ) => AssignAction<TContext, ExtractEvent<TEvent, TEventType>>;
+  events: FullEventCreatorMap<TEM>;
+  types: {
+    context: TContext;
+    events: TEvent;
```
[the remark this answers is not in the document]
davidkpiano: Is there prior art on this? I would like to avoid the confusion... maybe have these under `types`?

Thinking about it, these might not be necessary. We can make a helper, e.g.,:

```ts
type EventsOf<TModel> = TModel extends Model<any, infer Event> ? Event : never;
```

packages/core/test/model.test.ts
```diff
@@ -130,4 +130,68 @@ describe('createModel', () => {
...
+      {
+        name: 'David',
+        age: 30
+      },
+      {
+        updateName: (value: string) => ({ value }),
+        updateAge: (value: number) => ({ value }),
+        anotherEvent: () => ({})
+      }
+    );
```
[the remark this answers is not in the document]
davidkpiano: What would that look like? These are event creators; not sure how to avoid the confusion.

packages/core/src/model.ts
davidkpiano: The intention is to not provide `TEvent` (or any generics) at all when specifying event creators.

packages/core/src/model.ts
davidkpiano: Maybe, but the main use-case is a model that is _not_ specified with generics (inferred from arguments instead).

Andarist: Yes, I understand that but up until now providing generics was the preferred way of using the model and thus this is a conceptually perfect code:
https://codesandbox.io/s/objective-ellis-oy5jq?file=/src/index.ts:526-601

With changes from this PR, the new property `.events` would slip into the model there and that would be incorrect as it wouldn't actually be available at runtime

packages/core/test/model.test.ts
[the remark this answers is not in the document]
Andarist: Not super great but maybe smth like:
```js
createModel(ctx, { events })
```

packages/core/src/model.ts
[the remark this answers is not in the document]
davidkpiano: Omitted `types` - using `ModelContextFrom<Model>` and `ModelEventsFrom<Model>` instead

packages/core/src/model.ts
davidkpiano: This is not a concern, because the inferred type becomes `never`:

CleanShot 2021-03-13 at 20 44 33@2x
CleanShot 2021-03-13 at 20 45 41@2x

packages/core/src/model.ts
davidkpiano: This doesn't exactly work unless you add `EventObject & ...`:

```ts
type GetEvents<T extends EventCreatorMap<any>> = EventObject &
  {
    [K in keyof T]: Expand<{ type: K } & ReturnType<T[K]>>;
  }[keyof T];
```

But then it shows up weird:

CleanShot 2021-03-13 at 20 59 40@2x

We can work around this by adding some `Assert<TAsserted, TExpected>` type so that we have `Assert<GetEvents<...>, EventObject>` to clean this up, but at this point, it's almost as complex as the original types.

Let's have this as a refactor target if you want.

packages/core/src/model.ts
```diff
@@ -6,24 +6,82 @@ import type {
...
+  TEM extends EventCreatorMap<any>,
+  TEvent extends EventObject = GetEvents<TEM>
+>(initialContext: TContext, eventCreators?) {
+  const model: ContextModel<TContext, TEvent, TEM> = {
     initialContext,
     assign,
+    events: (eventCreators
+      ? mapValues(
+          eventCreators,
+          (fn, eventType) => (...args: Parameters<typeof fn>) => ({
```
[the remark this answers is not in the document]
davidkpiano: I made this change for runtime only: even if you return a `type`, it will not override the original `type`.

Not sure how to do this at the type level - everything I tried made TS angry 😡

packages/core/test/model.test.ts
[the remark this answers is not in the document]
davidkpiano: Done

packages/core/test/model.test.ts
```diff
@@ -130,4 +130,70 @@ describe('createModel', () => {
...
+      {
+        name: 'David',
+        age: 30
+      },
+      {
+        events: {
+          updateName: (value: string) => ({ value }),
+          updateAge: (value: number) => ({
+            value,
+            type: 'this should be overwritten'
```
Andarist, line 145: this is a very specific case - maybe worth moving to a dedicated test?

packages/core/src/model.ts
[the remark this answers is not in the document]
Andarist: I didn't really mean that we should validate this at type-level - it would be quite complicated for the benefit it would bring us (poor ROI in my opinion). Just meant a runtime validation - but switching the order of the property and the spread is OK as well.

packages/core/src/model.ts
Andarist: Yep - the inferred type is correct. The demo showcases how the explicit usage of the Model type can go wrong though.

davidkpiano: Let's tackle that in a follow-up PR.

Andarist: The change is really simple ( https://github.com/davidkpiano/xstate/pull/1995 ) if we'd like to do it. Note also that creating a default `EventCreatorMap` has also produced unsafe code even for models with appropriate creators - since all arguments were typed as `any` in such a case.

Andarist (approved):

PR #1966, opened 2021-03-01
useSelector() should not cause rerenders on non-memoized selectors

### LINKED ISSUE #1965: useSelector() needlessly requires a memoized selector
**Description**
`useSelector` requires the `selector` parameter to be wrapped in `React.useCallback()`. The [release notes of @xstate/react@1.3.0](https://github.com/davidkpiano/xstate/releases/tag/%40xstate%2Freact%401.3.0) do not stipulate such a necessity.

The selector itself will probably be a pure function in 99% of cases, so it should be OK to not update the `selected` data when the `selector` function changes.

**Expected Result**
Expected the following code to not rerender needlessly:
```typescript
const count = useSelector(someActor, state => state.context.count);
```

**Actual Result**
The selector `useEffect` is triggered constantly because the identity of the selector function changes on each rerender.

**Additional context**
```
xstate@4.16.2
@xstate/react@1.3.1
```

VanTanev:
Closes #1965

Something else to consider, the `useSelector()`  `useEffect()` should probably have `actor` as a dependency.

### DIFF
--- a/packages/xstate-react/src/useSelector.ts
+++ b/packages/xstate-react/src/useSelector.ts
@@ -25,7 +25,11 @@ export function useSelector<
   compare: (a: T, b: T) => boolean = defaultCompare,
   getSnapshot: (a: TActor) => TEmitted = defaultGetSnapshot
 ) {
-  const [selected, setSelected] = useState(() => selector(getSnapshot(actor)));
+  const selectorRef = useRef(selector);
+  selectorRef.current = selector;
+  const [selected, setSelected] = useState(() =>
+    selectorRef.current(getSnapshot(actor))
+  );
   const selectedRef = useRef<T>(selected);
 
   useEffect(() => {
@@ -36,16 +40,16 @@ export function useSelector<
       }
     };
 
-    const initialSelected = selector(getSnapshot(actor));
+    const initialSelected = selectorRef.current(getSnapshot(actor));
     updateSelectedIfChanged(initialSelected);
 
     const sub = actor.subscribe((emitted) => {
-      const nextSelected = selector(emitted);
+      const nextSelected = selectorRef.current(emitted);
       updateSelectedIfChanged(nextSelected);
     });
 
     return () => sub.unsubscribe();
-  }, [selector, compare]);
+  }, [compare]);
 
   return selected;
 }


davidkpiano:
Not sure if this is a good idea - if the selector changes from e.g., `(state) => state.context.foo` to `(state) => state.context.bar`, then the selected value should be recalculated.

VanTanev:
@davidkpiano My feeling is that such changes would almost never happen, while people accidentally causing a huge number of rerenders because they didn't wrap their selector in `React.useCallback()` will happen all the time.

Indeed, your very example when the feature was released missed `useCallback()`.

davidkpiano:
> @davidkpiano My feeling is that such changes would almost never happen, while people accidentally causing a huge number of rerenders because they didn't wrap their selector in `React.useCallback()` will happen all the time.
> 
> Indeed, your very example when the feature was released missed `useCallback()`.
> 
> PS For what it's worth, I caught this because I have [npmjs.com/package/stop-runaway-react-effects](https://www.npmjs.com/package/stop-runaway-react-effects) - not sure I would've noticed it otherwise.

There is already a test (with an inline selector!) that shows that rerenders don't occur if the selector changes (which, in the tests case, it always does) but the selected _value_ doesn't change: https://github.com/davidkpiano/xstate/blob/master/packages/xstate-react/test/useSelector.test.tsx#L8-L75

davidkpiano:
> @davidkpiano My feeling is that such changes would almost never happen, while people accidentally causing a huge number of rerenders because they didn't wrap their selector in `React.useCallback()` will happen all the time.

Unfortunately, React doesn't care about "would almost never happen" - it needs to assume that changes to values will either definitely happen or never will, and it does this by referential comparison. All React developers must be aware of this, and we can't just allow for the "95% use cases" for minor developer experience improvements.

In fact, you do not have to `useCallback` to memoize the function! Instead, define it outside the component:

```js

const selector = (state) => state.context.count;

const App = () => {
  // ...
  const count = useSelector(service, selector);

  // ...
}
```

VanTanev:
@davidkpiano I am unable to run the tests locally (guidance on that would be apprecaited.)

However, what I think triggers the bug is something like this:
```typescript
const { count } = useSelector(service, (state) => ({ count: state.context.count }));
```

If the selector itself returns a value whose identity changes on each selector execution, then we get into an infinite loop. If selectors only ran on state changes, the above would be perfectly fine.

As for defining the selector outside the function body:
1. I think it's harder to read.
2. We still lose TypeScript type inference, and would need to manually define types for the selector.

davidkpiano:
> However, what I think triggers the bug is something like this:
> 
> ```ts
> const { count } = useSelector(service, (state) => ({ count: state.context.count }));
> ```

The bigger issue here is the comparison function always thinking that the object is changing. You're going to have the same type of issue here:

```js
const Thing = ({ someObjectProp }) => {
  useEffect(() => {
    // will always run, UNLESS someObjectProp is memoized/stable
  }, [someObjectProp])
}
```

So whether you memoize the selector or not, you will have the same rerender issue if the selector returns an object.

That's what the comparison function is for:
```ts
useSelector(service,
  (state) => ({ count: state.context.count }),
  (a, b) => a.count === b.count
);
```

You will encounter this issue when using any sort of selector mechanism in any other library - for objects/arrays/etc, you must either provide stable/memoized references or smart comparison functions.

> As for defining the selector outside the function body:
> - I think it's harder to read.
> - We still lose TypeScript type inference, and would need to manually define types for the selector.

You can define the selector inline, which will give you type inference.

VanTanev:
@davidkpiano OK, I think I broadly agree with you. But a couple of notes

1. >  You can define the selector inline, which will give you type inference.

Not if you try to memoize it with `React.useCallback()`, as I showed [here](https://user-images.githubusercontent.com/202527/109529817-bc82dd80-7abe-11eb-8ac8-ea7b2818e94c.png) 

2. Then why aren't `actor`, `compare`, and `getSnapshot` part of the `useEffect()` deps array?

davidkpiano:
> 1. Then why aren't `actor` and `getSnapshot` part of the `useEffect()` deps array?

That's a good point; they probably should be.

> Not if you try to memoize it with React.useCallback(), as I showed here

Right, I don't recommend that.

Andarist:
Note that this also follows conventions used by the React team: https://github.com/facebook/react/tree/d857f9e4d057a3e039c0804c6ae087d348f97a06/packages/use-subscription#usage

> In order to avoid removing and re-adding subscriptions each time this hook is called, the parameters passed to this hook should be memoized. This can be done by wrapping the entire subscription with useMemo(), or by wrapping the individual callbacks with useCallback().

davidkpiano:
> @davidkpiano @Andarist Thank you for taking the time, and apologies for my wrong-headedness.
> 
> I hope you have a great day!

No worries, this is all great feedback that we can address in the documentation.

PR #1976, opened 2021-03-10
docs: fix links

lekterable:


Andarist:
From what I can see the current version on master works:
https://github.com/davidkpiano/xstate/blob/master/docs/packages/xstate-react/index.md#useserviceservice
while the proposed one doesn't:
https://github.com/davidkpiano/xstate/blob/db4a503a6c509e813b59eb1d52fa5af9695ace99/docs/packages/xstate-react/index.md#useservice-service

lekterable:
@Andarist hmm interesting, looks like links that work in README don't work in the online docs, because the issue I found comes from here:
https://xstate.js.org/docs/packages/xstate-react/

any clue why this might be? if not I'll try to look for the problem on my own later
CC: @davidkpiano

davidkpiano:
Closing as the docs have moved to https://github.com/statelyai/docs

PR #1977, opened 2021-03-11
[core] Add schema

davidkpiano:
This PR adds the `schema` property to the machine config and definition.

```ts
  // ...
  schema: {
    // These can be arbitrary meta data, accessible at runtime
    // Typecast these (something as SpecificType) for inferred types!
    context: someContextSchema,
    events: someEventSchemaMap,
    actions: someActionSchemaMap,
    guards: someGuardSchemaMap,
    services: someServiceSchemaMap
  },
  // ...
});
```

Purpose of this PR:

- Provides a place to specify important metadata for runtime types and validation (e.g., with JSON Schema) as part of the machine definition
- Allows for easier type inference without having to specify generics. Notice how the `context` and `events` are strongly typed, even though their generic types aren't specified:

Screen Shot 2021-03-10 at 10 10 00 PM

### DIFF
--- a/.changeset/chilled-windows-destroy.md
+++ b/.changeset/chilled-windows-destroy.md
@@ -46,9 +46,7 @@ import { createSchema, createMachine } from 'xstate';
 // Both `context` and `events` are inferred in the rest of the machine!
 const machine = createMachine({
   schema: {
-    context: createSchema<{ count: number }>({
-      // ...
-    }),
+    context: createSchema<{ count: number }>(),
     // No arguments necessary
     events: createSchema<{ type: 'FOO' } | { type: 'BAR' }>()
   }
--- a/.changeset/chilled-windows-destroy.md
+++ b/.changeset/chilled-windows-destroy.md
@@ -1,0 +1,57 @@
+---
+'xstate': minor
+---
+
+The `schema` property has been introduced to the machine config passed into `createMachine(machineConfig)`, which allows you to provide metadata for the following:
+
+- Context
+- Events
+- Actions
+- Guards
+- Services
+
+This metadata can be accessed as-is from `machine.schema`:
+
+```js
+const machine = createMachine({
+  schema: {
+    // Example in JSON Schema (anything can be used)
+    context: {
+      type: 'object',
+      properties: {
+        foo: { type: 'string' },
+        bar: { type: 'number' },
+        baz: {
+          type: 'object',
+          properties: {
+            one: { type: 'string' }
+          }
+        }
+      }
+    },
+    events: {
+      FOO: { type: 'object' },
+      BAR: { type: 'object' }
+    }
+  }
+  // ...
+});
+```
+
+Additionally, the new `createSchema()` identity function allows any schema "metadata" to be represented by a specific type, which makes type inference easier without having to specify generic types:
+
+```ts
+import { createSchema, createMachine } from 'xstate';
+
+// Both `context` and `events` are inferred in the rest of the machine!
+const machine = createMachine({
+  schema: {
+    context: createSchema<{ count: number }>({
+      // ...
+    }),
+    // No arguments necessary
+    events: createSchema<{ type: 'FOO' } | { type: 'BAR' }>()
+  }
+  // ...
+});
+```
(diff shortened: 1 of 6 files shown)


packages/core/test/schema.test.ts
```diff
@@ -0,0 +1,97 @@
+import { createMachine } from '../src';
+
+namespace JSONSchema {
```
davidkpiano, line 3: This is just for example/experimentation purposes; please do not scrutinize!

packages/core/src/types.ts
```diff
@@ -678,6 +678,15 @@ export interface MachineConfig<
    * The machine's own version.
    */
   version?: string;
+  schema?: MachineSchema<TContext, TEvent>;
+}
+
+export interface MachineSchema<TContext, TEvent extends EventObject> {
+  context?: TContext;
+  events?: TEvent;
+  actions?: ActionObject<TContext, TEvent>;
```
Andarist, line 687: I'm not sure this is a proper type for this - this should focus on the schema, so on a serializable `{ type: string, [key: string]: any }`. And this type includes an `exec` property and stuff.

I imagine that in the future we plan to put a schema for additional action parameters here.

packages/core/src/types.ts
```diff
@@ -678,6 +678,15 @@ export interface MachineConfig<
    * The machine's own version.
    */
   version?: string;
+  schema?: MachineSchema<TContext, TEvent>;
+}
+
+export interface MachineSchema<TContext, TEvent extends EventObject> {
+  context?: TContext;
```
Andarist, line 685: I was wondering how exactly this will interact with the top-level context property and according to my [simple experiments](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgLKIBaggZQRiAWzgB4AVAYQHtwIAPMAPmQG8AoZZBGyBgLmSUe9MABoOyAM74icAPwD2nTt1oMFg6mrASAvm31sYAVxAIwwGlygQ4kPAWLlGACmmP5AuCACeASgEyVgkbMGMoECkZYmQ4SUEAbgM2I1NzS0iEGzsIdHxsci1eJhc4ATysEFxo0iFtRj9WZFDwyLhkQyzbSArsFyUuYX5g5UHTSCgBABYAJj1xTndZRQkVIbABLpyHWRIWZBgqKgFJMChQAHMO1z89A1u2LZ7MPoHVYpXR7nHoabnOXQLKIeT7Kd4iTbZew1PYHI4nM6XURjWiTZAgYyEABG0GuLluAPuKSeuReVX6EnBwwGax+aNm8wkS2IoLW2kh3WqHlhhyoGlO5xAF2R31RAgx2Nxuhud10fiAA) this will work perfectly (like expected). 

In all cases TS favours the `schema.context` type which I suspect is done because:
- an explicit~ type is provided there through `createSchema`, so this is preferred over inferring the type of the top-level context
- inference for inner call (like the presented `identity`) has to happen first and by then the type is "explicit" when TS starts inferring for the whole `createMachine` call which leads us to the same situation as in the previous point

packages/core/src/schema.ts
```diff
@@ -0,0 +1,3 @@
+export function createSchema<T>(schema?: any): T {
```
Andarist, line 1: it would be great to export this from the main entry point as well

packages/core/src/types.ts
davidkpiano: Yep, I was pleasantly surprised by this too. Good to know!

packages/core/src/schema.ts
davidkpiano: Alright, I was afraid about increasing file size, but this is basically an identity function.

.changeset/chilled-windows-destroy.md
```diff
@@ -0,0 +1,57 @@
...
+
+```ts
+import { createSchema, createMachine } from 'xstate';
+
+// Both `context` and `events` are inferred in the rest of the machine!
+const machine = createMachine({
+  schema: {
+    context: createSchema<{ count: number }>({
+      // ...
+    }),
```
Andarist, line 51: For this example, I think it's better to just showcase the `createSchema` with provided generic (so no value actually passed to the `createSchema` call)

```suggestion
    context: createSchema<{ count: number }>(),
```

Andarist (approved):

PR #1986, opened 2021-03-12
[v5] Bring spawn API closer to v4

Andarist:
This brings the implementation closer to what we had in v4. However, this time instead of having a [`serviceScope`](https://github.com/davidkpiano/xstate/blob/7ae8f8e573f51e6222750afcda573cfabfdd0ad0/packages/core/src/serviceScope.ts) API we just set module-level "variable" that can be read when spawning etc to retrieve the current parent.

This is safe to do because `.transition` should be pure and thus it shouldn't really trigger any nested services in any way - beyond reading their initial state (a fresh initial state is later read when `start()` gets called and it's the one which actually ends up being used). So this time we don't need to maintain the stack of services (to the best of my knowledge).

This internal API ain't super pretty and has some potential for improvement but I'm also unsure right now how I would change it. The great benefit of being sneaky behind scenes is that the user is not responsible for passing the correct ref as `parent` - which IMHO is something what we should strive for. This creates a pit of success for the user as they just can't mess this up.

We could pass the actorRef through all layers (as it was done previously) but we still need to "capture" spawns within the scope of our module so at the end of the day we'd still do this sneaky stuff somewhere, just further down the road and closer to the moment where `assign` gets called.

We might actually want to revert to this later - it will be an easy refactor. Let's maybe see how this one plays out (unless you have some immediate suggestions on how this can be improved) and change this when it will be required or when doing some other polish/cleanup PRs. Ultimately, this doesn't really matter that much as what matters is the public API.

Right now `spawn` doesn't support referenced actors (the one defined in `implementations.actors`) - this IMHO should be changed as we really need that for things like "system" hydration.

There are reasonable concerns that a "detached" `spawn` like this is hard to type - it has no relation to the consuming machine so there is no contextual typing that could be utilized by this. OTOH - we don't really provide strict types for this kind of thing for `invoke` either so it's not like it's a new problem or anything like that. `invoke` has a potential of being improved though as it's contextual and we could use the `machine.schema.actors` to utilize types for this. I'm not super sure about this and it would require further experimentation but using this API we could enforce types when assigning to the context (as only valid actors should be assignable), but ye - definitely not possible to type-check the `spawn` call site.

Generally speaking, I'm  proponent of making our types stricter and limiting the unsafe things that people can do before the runtime (and right now there are quite a bit of those). However, there are some things that we won't ever~ be able type-check using TS alone and we desperately need an accompanying compiler/linter tool for XState. This will just have so many positive implications that it's just a necessity to create one - but, of course, this won't happen over night and we have a long road to take before we get there.

Given that this API feels the closest to what provide in all other places (to me) and as such - even considering the mentioned drawbacks - it's the one worth pursuing. 

To make this happen (`spawn` accepting referenced actors) we'll have to somehow refactor the types (accepted data types really) as `spawn` has to be able to consume what is assigned to `implementations.actors[key]` (or at least that is the most straightforward way of thinking about this imho). Right now `spawn` accept a behavior while `invoke` accepts a behavior factory. This whole thing can/should happen in the followup PR when we agree on the details.

I think this PR might fix an issue with some actors being started incorrectly when a macrostep has multiple microsteps but I need to verify this with the `next` branch and if my suspicion is correct then I'll add a test for this here.

`invokeAction` gets mutated within the pure `.transition` call now (so we don't call `createActorRefFromInvokeAction` from both the machine internals and the interpreter - so that's a plus). I think resolving this there is the correct approach - it can be just treated like resolving assigns, pures, etc (which already happen within the machine internals) - but maybe we should avoid mutating the object and just rather "resolve" it (like we do with other action types).

Actor refs are now stopped within the Interpreter and not within the Machine internals - which makes the machine pure again (profit 🚀).

We need to establish from what entry point the `spawn*` stuff should be imported. Maybe we should consolidate both `invoke*` and `spawn*` into a single thing and provide `xstate/spawn`? Having both, very similar, factory sets will be confusing for people.


packages/core/src/invoke.ts
```diff
@@ -135,21 +122,21 @@ export function createActorRefFromInvokeAction<
       if (!IS_PRODUCTION) {
         warn(
           false,
-          `No behavior found for invocation '${src.type}' in machine '${machine.id}'.`
+          `No actor found for invocation '${src.type}' in machine '${machine.id}'.`
```
davidkpiano, line 125:
```suggestion
          `Actor type '${src.type}' not found in machine '${machine.id}'.`
```

packages/core/src/interpreter.ts
```diff
@@ -453,6 +464,24 @@ export class Interpreter<
     return this;
   }
 
+  private _transition(
+    state: State<TContext, TEvent>,
+    event: SCXML.Event<TEvent>
+  ) {
+    try {
```
davidkpiano, line 471: What kind of errors can occur here?

Andarist: Caused by invalid arguments, problems in our implementation and probably some errors thrown when we call to the user’s code as I think we dont convert everything correctly to error.execution events.

This is just a cleanup so we dont land with some corrupted internal state - if we did it would mean that some other error has happened before (here) and it should be handled/fixed. In ideal world - this shouldnt be needed. OTOH this cleans avoids an extremely small memory leak in extremely unlikely situation.

davidkpiano (commented):
Tentatively approved, let's get the tests passing

davidkpiano:
Regarding the `spawn()` issues, here's a few options for strongly typing and being able to reference parameterized actor behaviors:

```ts
const machine = createMachine((self) => ({
  // ...
  entry: assign({
    // `self.spawn` strongly typed
    foo: (ctx, e) => self.spawn(/* ... */)
  })
}));
```

-----

```ts
const model = createModel({/* ... */}, {
  actors: {
    fetchUser: {/* ... */}
    // ... actors
  }
});

  model, // inferred types!
  // ...
  entry: assign({
    // `spawn()` not strongly typed, safety is in `model` instead
    foo: (ctx, e) => spawn(model.actors.fetchUser, 'user'),
    bar: (ctx, e) => spawn('fetchUser', 'user'),
  })
}));
```

-----
```ts
  // ...
  entry: assign({
    // `spawn()` not strongly typed, safety is in `meta.actors` instead
    foo: (ctx, e, { actors }) => spawn(actors.fetchUser, 'user')
  })
});
```

Let me know if any of these options is appealing.

Andarist:
1. This, while solving the problem, deviates from the XState's API too much (IMHO). Having a plain object as the config/states schema should IMHO be the goal - it composes well, people can declare objects at the top-level of their modules and adding `spawn` to a machine doesn't require wrapping the whole schema in a factory function~.

2. This one seems the closest to what we currently have in v4 and is what I would go with right now. This influences the API the least of all 3 and can be added, somewhat, at any moment. It requires some further exploration - for instance how a model should interact with the `implementations` parameter.

3. I like it aesthetically quite a bit but it also deviates from the current APIs a little bit and I'm not sure if that's a good thing. All the other stuff - actions, delays, guards, etc are referenced using simple strings. I feel like keeping consistency in that matter makes learning XState easier.

packages/core/src/interpreter.ts
```diff
@@ -171,11 +171,22 @@ export class Interpreter<
...
+        this._initialState ||
+        ((this._initialState = this.machine.getInitialState()),
+        this._initialState)
+      );
+    } finally {
+      CapturedState.current = {
+        actorRef: undefined,
+        spawns: []
+      };
+    }
```
davidkpiano, line 189: Should we (re)introduce a `withCurrentRef(this.ref, c => ... )` helper function to DRY this up?

packages/core/src/invoke.ts
```diff
@@ -135,21 +122,21 @@ export function createActorRefFromInvokeAction<
...
     const behavior = behaviorCreator(context, _event.data, {
-      parent: parentRef as any, // TODO: fix
       id,
       data,
       src,
       _event
     });
 
     actorRef = new ObservableActorRef(behavior, id);
+    invokeAction.src = actorRef;
```
davidkpiano, line 139: Can you explain the mutation here?

packages/core/test/invoke.test.ts
```diff
@@ -1735,6 +1735,7 @@ describe('invoke', () => {
         JSON.stringify(waitingState);
       }).not.toThrow();
 
+      // TODO: the `src` is resolved now to ObservableActorRef, figure out what's the proper behavior
```
davidkpiano, line 1738: The `src` should be a description of the source of the actor ref, not the actual actor ref

davidkpiano (commented):
One (big) remaining question on mutating `invokeAction.src` (probably shouldn't be done)

packages/core/src/interpreter.ts
Andarist: Maybe, I've mainly wanted to avoid a closure here (which probably ultimately doesn't matter) - gonna also look later if it's possible to somehow drop this when reading the initial state, or just remodel how the initial state is instantiated. I'm not yet sure how to do that and which path to explore but there are some subtle things in terms of how machine API and the interpreter interact with each other - need to start writing things down to open a discussion about this later.

davidkpiano: Yeah, initial thoughts:

- If interpreter started without context (`service.start()`), then the "cached" initial state can be used
- If interpreter started with context/different state (`service.start(someState)`), then the initial state needs to be reconstructed

packages/core/src/invoke.ts
Andarist: This was mentioned in the PR description as far from ideal - it was just something that I've hacked around to get me going and to get the final API shape (for this step of the refactor) passing tests.

I've refactored this now and this mutation doesn't happen anymore 🎉 Could you take another look? :)

davidkpiano (approved):

PR #1991, opened 2021-03-13
[@xstate/vue] Update hooks: New useActor and useInterpret + update useService and useMachine

santicros:
Hii!

I've tried to update the @xstate/vue composables to port the newest @xstate/react useInterpret + useActor and updated useMachine + useService.

I'm very very new to Typescript, so any feedback will be appreciated.
For now, I haven't created new tests as I'm very newby... But I've checked that all the useMachine & useService are passing. I could try to add some later or would welcome any contributions...

I've also found that the React package has ES Modules and the Vue doesn't, so I've changed that too. Hope it's ok :)

Thanks!!


packages/xstate-vue/package.json
```diff
@@ -25,15 +25,17 @@
   "files": [
     "lib/**/*.js",
     "lib/**/*.d.ts",
-    "dist/**/*.js"
+    "dist/**/*.js",
+    "es/**/*.js",
```
Andarist, line 29: the ESM build should be added correctly using `"module"` package.json key so people don't have to worry about picking it manually

.changeset/shy-moons-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/vue': minor
+---
+
+New useInterpret & useActor composables + ES Module exports for Vue
```
Andarist, line 5: We are not bound to create a single changeset for a package - those are two separate concerns and should be just put into 2 separate changesets

Andarist:
It would be great if we could add some basic tests as part of this PR.

packages/xstate-vue/package.json
santicros: True! Forgot to add it. Added:
`"module": "es/index.js"` 👍

.changeset/shy-moons-hear.md
santicros: Sorry! Haven't used changeset before. Added new changeset now :)

santicros (commented):
Great! Added "module": "es/index.js" & changeset. I'll try to do some tests later 👍

santicros:
I've added some tests for the new useInterpret & useActor! @Andarist 

One thing I don't really understand is how the deferred useActor events works event on React. I've tried to mimic the test on Vue, but I don't think is the defer is doing anything on it... Would love to know more about it :)

packages/xstate-vue/test/UseInterpret.vue
```diff
@@ -0,0 +1,38 @@
...
+
+export default defineComponent({
+  setup() {
+    const service = useInterpret(machine);
+    const state = ref();
+    service.subscribe((nextState) => {
+      state.value = nextState.value;
+    });
+    return { service, state };
+  }
```
Andarist, line 36: To kinda utilize the `useInterpret`'s API better you could this:
```suggestion
    const state = ref();
    const service = useInterpret(machine, {}, (nextState) => {
      state.value = nextState.value;
    });
    return { service, state };
  }
```

packages/xstate-vue/test/UseInterpret.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <button @click="service?.send('TOGGLE')" data-testid="button">
```
Andarist, line 2: is the `?.` required here? It seems to me that `service` should just always be available here 🤔

packages/xstate-vue/test/UseActorDeferred.vue
```diff
@@ -0,0 +1,28 @@
...
+import { useActor } from '../src';
+
+export default defineComponent({
+  props: {
+    actor: {
+      type: Object
+    }
+  },
+  setup(props) {
+    const { state: actorState, send: actorSend } = useActor(props.actor as any);
```
Andarist, line 19: why do have to use `as any` here? how one could avoid it in the real life project?

Andarist:
> One thing I don't really understand is how the deferred useActor events works event on the React tests. I've tried to mimic the test on Vue, but I don't think the defer is doing anything on it... Would love to know more about it :)

I don't think this is needed at all here - unless in Vue it's also possible for a component to *render* without being actually ever mounted. Is it? 

In Vue we call `.start()` within the setup function and thus you don't really deal with deferred events there. In React `.start()` gets called from within `useEffect` which when communicating the parent and the child through events might require the logic of the deferred events to function properly to handle events sent from children to a parent before it gets a chance to get started.

packages/xstate-vue/test/UseInterpret.vue
santicros: True!

packages/xstate-vue/test/UseActorDeferred.vue
santicros: Tried to improve it with:

``` js
 const { state: actorState, send: actorSend } = useActor(
      props.actor as ActorRefFrom<any>
);
```
If we had the actual machine on the component we could also put like on the React test:
``` js
const { state: actorState, send: actorSend } = useActor(
      props.actor as ActorRefFrom<typeof childMachine>
 );
```

santicros:
> > I don't think this is needed at all here - unless in Vue it's also possible for a component to _render_ without being actually ever mounted. Is it?
> 
> Not sure... The lifecycle for the composition API is: **setup()** (previously was beforeCreate + created) > **onBeforeMount** > **onMounted** > ... 
> 
> 

I think now I understand the meaning better, and I also believe in Vue it's not possible to render something before it's mounted, so then it wouldn't make sense to have that as you were saying..

packages/xstate-vue/test/UseActorDeferred.vue
Andarist: Well, the ideal situation would be if could just have `props` typed correctly to avoid `as` assertions altogether (as they are not 100% type-safe). Is there a way to do this?

santicros: Sorry! I'm very new to Typescript.. I tried with this:
``` vue
<script lang="ts">
import { defineComponent, onMounted, PropType } from 'vue';
import { useActor } from '../src';
import { ActorRefFrom } from 'xstate';

export default defineComponent({
  props: {
    actor: {
      type: Object as PropType<ActorRefFrom<any>>
    }
  },
  setup(props) {
    const { state: actorState, send: actorSend } = useActor(props.actor);

    onMounted(() => {
      actorSend({ type: 'FINISH' });
    });

    return { actorState, actorSend };
  }
});
```

But I was getting an error on props.actor:
```
No overload matches this call.
  Overload 1 of 2, '(actorRef: ActorRef<any, any> | Ref<ActorRef<any, any>>, getSnapshot?: ((actor: ActorRef<any, any>) => any) | undefined): { state: Ref<any>; send: Sender<...>; }', gave the following error.
    Argument of type '(SpawnedActorRef<EventObject, State<unknown, EventObject, any, Typestate<unknown>>> & { state: State<unknown, EventObject, any, Typestate<...>>; }) | undefined' is not assignable to parameter of type 'ActorRef<any, any> | Ref<ActorRef<any, any>>'.
      Type 'undefined' is not assignable to type 'ActorRef<any, any> | Ref<ActorRef<any, any>>'.
  Overload 2 of 2, '(actorRef: ActorRef<EventObject, State<unknown, EventObject, any, Typestate<unknown>>> | Ref<ActorRef<EventObject, State<unknown, EventObject, any, Typestate<...>>>>, getSnapshot?: ((actor: ActorRef<...>) => State<...>) | undefined): { ...; }', gave the following error.
    Argument of type '(SpawnedActorRef<EventObject, State<unknown, EventObject, any, Typestate<unknown>>> & { state: State<unknown, EventObject, any, Typestate<...>>; }) | undefined' is not assignable to parameter of type 'ActorRef<EventObject, State<unknown, EventObject, any, Typestate<unknown>>> | Ref<ActorRef<EventObject, State<unknown, EventObject, any, Typestate<...>>>>'.
      Type 'undefined' is not assignable to type 'ActorRef<EventObject, State<unknown, EventObject, any, Typestate<unknown>>> | Ref<ActorRef<EventObject, State<unknown, EventObject, any, Typestate<...>>>>'.Vetur(2769)
```

Do you have any suggestion?

Andarist: I've pushed out some commits into your branch - so please pull them before you continue working on this branch.

I've removed this as assertion from another file (since this one was removed) and it seems to work. Please recheck if it's OK on your machine now (if not maybe try in a fresh copy of the repo? or rebuild the monorepo?).

packages/xstate-vue/test/UseService.vue
```diff
@@ -17,8 +17,7 @@ export default defineComponent({
     }
   },
   setup(props) {
-    const { service } = toRefs(props);
```
Andarist, line 21: was this needed? I've figured out that maybe not and using it here has been creating a type problem

packages/xstate-vue/test/UseActorDeferred.vue
santicros: Thanks!! It's interesting...
If done a fresh copy of the repo but had the same errors. But have found that:
- if I open the repo on VS Code from the xstate-vue package folder, where there's a tsconfig.json, Vetur gives this Typescript errors.
- If i open the folder from the root xstate folder, Vetur gives an initial startup error that can't find a tsconfig, but then I don't have this Typescript errors... So I'll continue working this way :)

.changeset/nasty-pigs-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/vue': minor
+---
+
+Declare external package peerDependencies in UMD rollup build
```
Andarist, line 5:
```suggestion
Fixed the UMD build by externalizing XState & Vue correctly. 
```

.changeset/shy-moons-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/vue': minor
+---
+
+New useInterpret & useActor + updated useMachine & useService composables ported from @xstate/react.
```
Andarist, line 5: Added new `useInterpret` which is a low-level composable that interprets the `machine` and returns the `service`:

```js
import { useInterpret } from '@xstate/vue';
import { someMachine } from '../path/to/someMachine';

export default defineComponent({
  setup() {
    const state = ref();
    const service = useInterpret(machine, {}, (nextState) => {
      state.value = nextState.value;
    });
    return { service, state };
  }
});
```

.changeset/spotty-mayflies-care.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/vue': minor
+---
+
+Add ES Module builds & exports
```
Andarist, line 5:
```suggestion
Added a proper ESM file using the`"module"` field in the `package.json`. It helps bundlers to automatically pick this over a file authored using CommonJS and allows them to apply some optimizations easier.
```

santicros:
Awesome, will do now!

santicros:
Everything ready for my part!

.changeset/late-wasps-nail.md
```diff
@@ -0,0 +1,17 @@
...
+'@xstate/vue': minor
+---
+
+Added new `useActor`, which is a composable that subscribes to emitted changes from an existing `actor`:
+
+```js
+import { useActor } from '@xstate/vue';
+
+export default defineComponent({
+  props: ['someSpawnedActor'],
```
Andarist, line 11: is this correct? how would this work? 🤔

santicros: It's the most basic way to pass a prop to a component! Not recommended with Typescript but was to simplify and focus on the hook itself :)
https://v3.vuejs.org/guide/component-props.html#prop-types

But could change it to a more detailed prop!

Andarist: It's OK then - quick followup question: is this *required*? or does it only enable some static/runtime checks? I guess this might be required by the compiler, but I'm wondering if that's the case.

.changeset/late-wasps-nail.md
```diff
@@ -0,0 +1,17 @@
...
+  setup() {
+    const { state, send } = useActor(someSpawnedActor);
```
Andarist, line 13: Not 100% sure but I would expect this to come from props:

```suggestion
  setup({ someSpawnedActor }) {
    const { state, send } = useActor(someSpawnedActor);
```

.changeset/late-wasps-nail.md
santicros: If you want to use the prop in the component it is required. You can choose to do it in an array format or object.

Andarist (approved):
Great work 🎉 I've never written any Vue code myself but I've tried my best here and reviewed this with extra. Looks legit, no scam 😉

.changeset/late-wasps-nail.md
santicros: It's the same idea as a prop in React that you have to put in the function argument ;)

.changeset/late-wasps-nail.md
santicros: Oops. Sorry, you are totally correct. But better not to destructure props because you'll loose Vue's reactiviy. Should be:
``` js
setup(props) {
    const { state, send } = useActor(props.someSpawnedActor);
    ...
}
```

PR #1994, opened 2021-03-14
Reorder createMachine overloads

Andarist:
I would also intuitively put the Model-related overload first but it seems that this was problematic in here - all the other generics have defaults so using this:
```ts
createMachine<any>({})
```
has been caught by the first overload (the Model-related one) and resulted in this:
Screenshot 2021-03-14 at 21 59 02
which in turn has caused some TS error down the road when accessing the `context` which got inferred to unknown: https://github.com/davidkpiano/xstate/pull/1955/checks?check_run_id=2104765685#step:6:1150

And it really seems that up until now the user expectation was different - this meant `TContext` being `any` so this change makes this behave closer to what we have on the master branch.

### DIFF
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -50,17 +50,17 @@ export function Machine<
 }
 
 export function createMachine<
-  TModel extends Model<any, any>,
-  TContext = ModelContextFrom<TModel>,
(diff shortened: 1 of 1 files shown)


Andarist:
Oh, well - this won't work as Model matches the TContext and thus the second overload is never reached.

davidkpiano:
Possible solution for this in #2060

PR #1995, opened 2021-03-14
Change the default type for Model's EventCreatorMap to never

Andarist:


Andarist:
This is being superseded by https://github.com/davidkpiano/xstate/pull/1999

PR #1999, opened 2021-03-15
Self based types for model

Andarist:


packages/core/src/model.ts
```diff
@@ -8,10 +8,16 @@ import type {
 } from './types';
 import { mapValues } from './utils';
 
+type AnyFunction = (...args: any[]) => any;
+
+type Cast<A1 extends any, A2 extends any> = A1 extends A2 ? A1 : A2;
+type Compute<A extends any> = { [K in keyof A]: A[K] } & unknown;
```
davidkpiano, line 14: What does this do?

davidkpiano (approved):

packages/core/src/model.ts
Andarist: It's the equivalent of your [`Expand`](https://github.com/davidkpiano/xstate/blob/5bc2ea6702683e1de15c3ec4b0bf10ebb838d3e4/packages/core/src/model.ts#L52). I've just been using this on TS playground when I've been experimenting and later copied it from here - can revert back to your `Expand` if you like that version better.

PR #2007, opened 2021-03-18
Added a runtime warning for "unreachable" `onDone` for parallel states

Andarist:
⚠️ I haven't really tested this yet - just created this in a moment so we don't forget about it. Gonna add a test for this when I get a chance to get back to it. 

Created "in response" to: https://github.com/davidkpiano/xstate/discussions/2003

### DIFF
--- a/.changeset/olive-tools-greet.md
+++ b/.changeset/olive-tools-greet.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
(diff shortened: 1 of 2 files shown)


davidkpiano:
Closing as stale, and can probably be a studio-side validation https://linear.app/statelyai/issue/STA-5024

PR #2021, opened 2021-03-21
Add additional environment variable for devTools

SimeonC:
When attempting to use devtools to “preview” in a storybook story, this works in development but not in production build due to the IS_PRODUCTION variable. I need to bypass this variable when running `build-storybook` or the devTools do not run at all.

Found this little problem here when building my storybook addon. Works great in local dev, but not in build mode. https://github.com/SimeonC/storybook-xstate-addon/issues/6

### DIFF
--- a/.changeset/nervous-toes-exercise.md
+++ b/.changeset/nervous-toes-exercise.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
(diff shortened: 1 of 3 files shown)


Andarist:
Unfortunately, I don't think this is the best solution that we can come up with for the long run. This issue has actually been raised yesterday as well and I would recommend watching it and commenting there with your own thoughts on the matter: https://github.com/davidkpiano/xstate/issues/2020

SimeonC:
Oh, haha. Was thoroughly in code mode and didn't think to check the issues. Thanks for pointing this out.

PR #2026, opened 2021-03-23
[v5] Merge vue changes

davidkpiano:
Merging `master` (with latest Vue changes) into `next`


packages/core/src/StateNode.ts
```diff
@@ -120,6 +120,8 @@ export class StateNode<
    */
   public order: number = -1;
 
+  public __xstatenode: true = true;
```
Andarist, line 123: just wondering - any particular reason why this has been added to StateNode? The main reason for this property was to check if something is a machine or not, do we need to check anywhere if something is a StateNode?

davidkpiano: Accidental addition. Will remove

davidkpiano:
```suggestion
```

PR #2029, opened 2021-03-23
[@xstate/vue] Update Docs: Add new useActor & useInterpret Hooks + Fixes

santicros:
Hi again!
I've updated the Vue docs, so they have the latest hooks from #1991.

I've also added:
- A table of contents with doctoc at the top as @xstate/react has.
- Added the missing Configuring Machines from #1973 PR that was done only to the docs folder and is deleted when building the project.
- Fixed the Vue links so they point to the latest Vue v3 docs.

### DIFF
--- a/packages/xstate-vue/README.md
+++ b/packages/xstate-vue/README.md
@@ -116,13 +116,103 @@ This special `useMachine` hook is imported from `@xstate/vue/lib/fsm`
 
 **Example** (TODO)
 
-## Configuring Machines (TODO)
+## Configuring Machines
 
 Existing machines can be configured by passing the machine options as the 2nd argument of `useMachine(machine, options)`.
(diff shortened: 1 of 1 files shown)


santicros:
My only question is: should I commit the changes that the npm run build or npm run dev does locally? So it copies the packages files to the docs folder, or will this done automatically on merge?
Thanks!!

davidkpiano:
This is wonderful, thank you!

davidkpiano:
> My only question is: should I commit the changes that the `yarn docs:dev` or `yarn docs:build` does locally? So it copies the package README to the docs folder, or will this done automatically on merge?
> Thanks!!

Still need to write a script for this. Don't worry about it, I'll do it manually.

PR #2035, opened 2021-03-25
[docs] Isolate documentation dependencies

davidkpiano:


docs/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@xstate/docs",
+  "version": "1.0.0",
```
Andarist, line 3: As this is not publishable to the registry we don't have to track the version. At the same time it's preferable to add a private flag so we don't accidentally publish this.
```suggestion
  "private": true,
```

docs/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@xstate/docs",
```
Andarist, line 2: Given our packages use the `@xstate` scope and this is not a package per se we might want to avoid using this scope like that to avoid confusion. It's a really minor thing though so if you don't agree - just disregard the comment.

```suggestion
  "name": "xstate-docs",
```

docs/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "docs:dev": "vuepress dev .",
+    "docs:readme": "cpy ../README.md .",
+    "docs:packages": "cpy '../packages/*/README.md' docs --parents --rename=index.md && cpy '../packages/*/docs/*.md' docs --parents",
+    "docs:build": "npm run docs:readme && npm run docs:packages && vuepress build .",
+    "docs:pdf": "npm run docs:build && vuepress export ."
+  },
+  "dependencies": {
+    "@vuepress/plugin-google-analytics": "^1.8.2",
+    "cpy-cli": "^3.1.1",
+    "gh-pages": "^3.1.0",
```
Andarist, line 19: this was probably supposed to be used by the publish script but so far that script is missing

davidkpiano: Yeah, not sure how to test this... gotta be careful with this.

Andarist: Well, you'll have to do a test run on production :trollface: Unless you want to go through the trouble of testing this on a fork.

PR #2039, opened 2021-03-26
React Suspense Support via useSuspensefulGetter Hook

andria-dev:
This PR adds the feature discussed in #1746 with type definitions and documentation.

## Changes

* Added `useSuspensefulGetter.ts`
* Exported useSuspensefulGetter and Resource out of `index.js`
* Added documentation for `useSuspensefulGetter`

## To-Do

- [ ] Adding tests
- [ ] Add changeset

Hopefully, Suspense tests aren't too different from normal React tests. I'm going to try to write the tests without updating the React dev dependency but if they simply don't work in React 16, I will probably need to update the React dev dependency if that's okay

### DIFF
--- a/docs/packages/xstate-react/index.md
+++ b/docs/packages/xstate-react/index.md
@@ -11,6 +11,7 @@
   - [`useActor(actor, getSnapshot)`](#useactoractor-getsnapshot)
   - [`useInterpret(machine, options?, observer?)`](#useinterpretmachine-options-observer)
   - [`useSelector(actor, selector, compare?, getSnapshot?)`](#useselectoractor-selector-compare-getsnapshot)
+  - [`useSuspensfulGetter(state, condition, getter)`](#usesuspensfulgetterstate-condition-getter)
   - [`asEffect(action)`](#aseffect-action)
   - [`asLayoutEffect(action)`](#aslayouteffectaction)
   - [`useMachine(machine)` with `@xstate/fsm`](#usemachinemachine-with-xstatefsm)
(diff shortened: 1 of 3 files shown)


docs/packages/xstate-react/index.md
```diff
@@ -252,6 +253,59 @@ const App = ({ service }) => {
...
+
+In other words, once your `condition` is satisified, `read()` will stop suspending and simply give you your value that you asked for with `getter`.
+
+#### Example
+
+In this example, `Outer` is the resource creator and `Inner` is the resource consumer. In `Outer`, we use our machine `userMachine` to determine if the user is logged in and fetch their information. In most cases, getting a user's data is asynchronous and you don't often know how long it will take. Therefore, we will use `useSuspensefulGetter` to create a `resource` to suspend the section of the application that needs to know something about the user until we have access to the user — represented in code, this condition is `user !== undefined`.
+
+`useSuspensefulGetter` does not do the suspending all on its own. After the creation of this `resource`, **we must pass it down at least one level before reading it to avoid an unending loop**. Now that we've passed it down to `Inner`, we can call `resource.read()`. This will either give us access to `state.context.user` or it will suspend all elements up to the nearest `<Suspense>` element and display its `fallback` attribute instead.
```
Andarist, line 276: This is the main worrying factor for me - the limitation is real and seems like a sign that maybe there are other ways to do this.

Andarist (commented):
First of all - thank you for such a good PR (in terms of the content, docs included and everything).

Personally, I would wait with it though. The mentioned limitation imposed on how one is supposed to use this is a red flag for me in terms of how well this integrates with Suspense etc. I understand why the limitation is there but I'm also wondering if maybe some other things could be figured out to remove this limitation.

The Suspense still feels quite new and I have not yet fully explored its potential or patterns around it (nor React team provides much guidance for that at the moment).

Introducing a new API has to always be considered carefully because it's sort of a commitment to it - we don't want to drop introduced APIs to avoid churn for our consumers.

docs/packages/xstate-react/index.md
ChrisBrownie55: I see where you're coming from about this limitation. I too wonder if there is a better way to do this but according to [the examples in React's documentation on "Suspense for Data Fetching"](https://reactjs.org/docs/concurrent-mode-suspense.html#what-is-suspense-exactly), they always create the resource outside of the component where the resource is used. In our case, we need to create the resource dynamically so we have to do this within another React component.

You can also see [more examples of how this pattern works in the React documentation](https://reactjs.org/docs/concurrent-mode-patterns.html) when they demonstrate using `useState` to create the `resource` and `setResource` variables. They then follow this by passing the `resource` down at least one level because if they were to call `resource.read()` where `useState` was called, it would cause the infinite rerendering issue — a.k.a. throwing a promise causes "suspending" to occur but also blows away the state holding the resource and then rerenders. I believe that this is just a fundamental part of how Suspense and "suspending" was designed.

With all of that said, I completely understand waiting until this feature is officially no longer listed as experimental and the ideas behind Concurrent Mode and Suspense are declared solidified and ready.

ChrisBrownie55: Although to be clear, while this feature, Suspense and the ability to suspend by throwing a promise, is "experimental" it has been in stable releases for quite some time now from what I can tell. It seems to be that the Concurrent Mode features as a whole are considered experimental though.

benbender: Any energy to take a new look at this PR? Now while concurrent-mode/suspense are in alpha and announced for react 18, it may be a good point in time! :)

davidkpiano:
Closing as we're going to re-examine how (and if) we will support experimental Suspense within `@xstate/react`.

PR #2046, opened 2021-03-29
Allow devTools usage in production

### LINKED ISSUE #2020: @xstate/inspect should not bail in production
`@xstate/inspect` works when `process.env.NODE_ENV === 'development'`, but fails when `NODE_ENV === 'production'`. This is due to this line of code:

https://github.com/davidkpiano/xstate/blob/f3870f837e7d86bd4e3a1d28485e76b026579291/packages/core/src/devTools.ts#L43

Reproduction:

https://github.com/mattpocock/xstate-inspect-repro

Instead of bailing here, we should leave it up to the user whether or not they choose to have `devTools: true` in production

SimeonC:
Fixes #2020

### DIFF
--- a/.changeset/hot-pumas-chew.md
+++ b/.changeset/hot-pumas-chew.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Removed IS_PRODUCTION check for initialising devTools in intepreter.
--- a/.changeset/hot-pumas-chew.md
+++ b/.changeset/hot-pumas-chew.md
@@ -2,4 +2,4 @@
 'xstate': minor
 ---
 
-Removed IS_PRODUCTION check for initialising devTools in intepreter.
+Allow machines to communicate with the inspector even in production builds.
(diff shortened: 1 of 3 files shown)


with-heart (approved):
Would it make sense to specifically call out the inspect-as-documentation use case that initiated the issue discussion somewhere?

Otherwise this lgtm!

davidkpiano:
@Andarist Would you consider this a breaking change?

.changeset/hot-pumas-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Removed IS_PRODUCTION check for initialising devTools in intepreter.
```
Andarist, line 5:
```suggestion
Allow machines to communicate with the inspector even in production builds.
```

Andarist (approved):

Andarist:
@davidkpiano no, I don't think so - as @SimeonC has pointed out in the issue: this doesn't change the behavior that much as it's the `inspect` that creates the inspector, and not this code here. So it's hard for me to imagine how this could break anyone - it should only allow to setup the inspector in production builds at the expense of few bytes being added to the `xstate`'s bundlesize (but just a few as this tries to call into the inspector API which gets provided through a global).

PR #2050, opened 2021-03-31
exposing react/fsm as a separate entrypoint

theKashey:
Thank you for the library and thank you again for providing a bare minimum implementation for it.

However, the recommended way to use `react/lib/fsm` is something I would like to improve, and make it just `react/fsm`, with autopicking the correct implementation for the correct environment.

Not using fancy `package.exports` - adding separate entrypoint in an old fashion way, compatible with completely everything.

Long story short -
```diff
- import { useMachine } from `@xstate/react/lib/fsm`;
- import { useMachine } from `@xstate/react/es/fsm`;
+ import { useMachine } from `@xstate/react/fsm`;
```

### DIFF
--- a/.changeset/healthy-rice-shout.md
+++ b/.changeset/healthy-rice-shout.md
@@ -7,3 +7,4 @@ Added an explicit entrypoint for `@xstate/react/fsm` which you can use instead o
 ```diff
 -import { useMachine } from '@xstate/react/lib/fsm'
 +import { useMachine } from '@xstate/react/fsm'
+```
--- a/.changeset/healthy-rice-shout.md
+++ b/.changeset/healthy-rice-shout.md
@@ -1,0 +1,5 @@
+---
+'@xstate/react': patch
+---
+
+Exposing a separate entrypoint for @xstate/react/fsm
--- a/.changeset/healthy-rice-shout.md
+++ b/.changeset/healthy-rice-shout.md
@@ -2,4 +2,8 @@
 '@xstate/react': patch
 ---
 
-Exposing a separate entrypoint for @xstate/react/fsm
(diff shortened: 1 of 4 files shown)


Andarist (approved):
This is already provided on the `next` branch - but still a good change for the current release line, so definitely smth that we can incorporate. 

Could you just add a changeset describing the change (from the user perspective)?

theKashey:
Not sure about release processes here, and how release notes are generated (are those _examples_ in releases coming from changeset as well?).
Provided a minimal message about the change, without any migration guides. Please double check if this is acceptable/matching your expectations.

.changeset/healthy-rice-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/react': patch
+---
+
+Exposing a separate entrypoint for @xstate/react/fsm
```
Andarist, line 5:
```suggestion
Added an explicit entrypoint for `@xstate/react/fsm` which you can use instead of `@xstate/react/lib/fsm`. This is the only specifier that will be supported in the future - the other one will be dropped in the next major version.

```diff
-import { useMachine } from '@xstate/react/lib/fsm'
+import { useMachine } from '@xstate/react/fsm'
```
```

File: .changeset/healthy-rice-shout.md [hunk 2]
```diff
@@ -0,0 +1,9 @@
+---
+'@xstate/react': patch
+---
+
+Added an explicit entrypoint for `@xstate/react/fsm` which you can use instead of `@xstate/react/lib/fsm`. This is the only specifier that will be supported in the future - the other one will be dropped in the next major version.
+
+```diff
+-import { useMachine } from '@xstate/react/lib/fsm'
++import { useMachine } from '@xstate/react/fsm'
```
Andarist, line 9: @theKashey could you add missing \``` after this block? GitHub code suggestions don't handle this well 😬

theKashey: Applied

davidkpiano (approved):

PR #2055, opened 2021-04-01
Add exception when next state config is undefined

f-elix:
I added an error throw in case the `nextStateConfig` is undefined, so we don't get a runtime error later in the code (when accessing `nextStateConfig.entry`, specifically). 

I also tidied up another `if` statement earlier in the same function, if that's all right with you.

I added a test to verify the behavior. It's find it a bit convoluted, as I had to modify the test machine config, so I'm open to better ways of testing this. In any case, all tests are passing.

### DIFF
--- a/packages/xstate-fsm/src/index.ts
+++ b/packages/xstate-fsm/src/index.ts
@@ -159,14 +159,10 @@ export function createMachine<
       const eventObject = toEventObject<TEvent>(event);
       const stateConfig = fsmConfig.states[value];
 
-      if (!IS_PRODUCTION) {
-        if (!stateConfig) {
-          throw new Error(
-            `State '${value}' not found on machine${
-              fsmConfig.id ? ` '${fsmConfig.id}'` : ''
-            }.`
-          );
-        }
+      if (!IS_PRODUCTION && !stateConfig) {
+        throw new Error(
+          `State '${value}' not found on machine ${fsmConfig.id ?? ''}`
+        );
       }
 
       if (stateConfig.on) {
@@ -186,8 +182,18 @@ export function createMachine<
 
           const isTargetless = target === undefined;
 
+          const nextStateValue = target ?? value;
+          const nextStateConfig = fsmConfig.states[nextStateValue];
+
+          if (!IS_PRODUCTION && !nextStateConfig) {
+            throw new Error(
+              `State '${nextStateValue}' not found on machine ${
+                fsmConfig.id ?? ''
+              }`
+            );
+          }
+
           if (cond(context, eventObject)) {
-            const nextStateConfig = fsmConfig.states[target ?? value];
             const allActions = (isTargetless
               ? toArray(actions)
               : ([] as any[])
--- a/packages/xstate-fsm/test/fsm.test.ts
+++ b/packages/xstate-fsm/test/fsm.test.ts
@@ -10,7 +10,8 @@ describe('@xstate/fsm', () => {
   type LightEvent =
     | { type: 'TIMER' }
     | { type: 'INC' }
-    | { type: 'EMERGENCY'; value: number };
+    | { type: 'EMERGENCY'; value: number }
+    | { type: 'TARGET_INVALID' };
 
   type LightState =
     | {
@@ -25,6 +26,8 @@ describe('@xstate/fsm', () => {
         value: 'red';
         context: LightContext & { go: false };
       };
+
+  const invalidState = 'invalid';
 
   const lightConfig: StateMachine.Config<
     LightContext,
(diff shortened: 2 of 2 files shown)


packages/xstate-fsm/src/index.ts
```diff
@@ -186,8 +182,18 @@ export function createMachine<
...
 
+          const nextStateValue = target ?? value;
+          const nextStateConfig = fsmConfig.states[nextStateValue];
+
+          if (!IS_PRODUCTION && !nextStateConfig) {
+            throw new Error(
+              `State '${nextStateValue}' not found on machine ${
+                fsmConfig.id ?? ''
+              }`
+            );
```
davidkpiano, line 193: Can we DRY this up to reduce code size, since this is the same as the other error?

Andarist: just a note: the code size won't really be reduced since this is a dev-only branch of code so overall DRYing might decrease the readability a little bit for kinda no gains, I'm ok with either way though

packages/xstate-fsm/src/index.ts
```diff
@@ -186,8 +182,18 @@ export function createMachine<
 
           const isTargetless = target === undefined;
 
+          const nextStateValue = target ?? value;
+          const nextStateConfig = fsmConfig.states[nextStateValue];
```
Andarist, line 186: q: why the code has been moved? this has previously been resolved after passing a guard

packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -140,6 +146,14 @@ describe('@xstate/fsm', () => {
     }).toThrow();
   });
 
+  it('should throw an error for undefined next state config', () => {
+    expect(() => {
+      lightFSM.transition('green', 'TARGET_INVALID');
```
Andarist, line 151: could you create a separate machine for this test? we try to not reuse machines between tests right now - the rest of the tests has been written when there was no such decision yet and they just haven't been rewritten yet

packages/xstate-fsm/src/index.ts
f-elix: I wanted to at first, but I realized you would have to resolve the `stateConfig` and the `nextStateConfig` at the same time to throw the same error. I don't see how this is possible with the way the code is written are right now. 

I could also write a function that returns the error message string with the value passed in to avoid re-writing the same line twice, but I don't think that's what you had in mind, correct?

packages/xstate-fsm/src/index.ts
f-elix: My thinking was that the result of the guard doesn't matter much if the `nextStateConfig` is invalid, so I figured I would throw the error early.

packages/xstate-fsm/test/fsm.test.ts
f-elix: Sure, I'd be happy to! Shouldn't that be in a separate PR though?

packages/xstate-fsm/src/index.ts
davidkpiano: > just a note: the code size won't really be reduced since this is a dev-only branch of code so overall DRYing might decrease the readability a little bit for kinda no gains, I'm ok with either way though

Ah, missed that. No worries then!

packages/xstate-fsm/src/index.ts
f-elix: Should I change anything here?

packages/xstate-fsm/test/fsm.test.ts
Andarist: It's a new test - not asking about refactoring the old one, just using a separate machine for this new one.

packages/xstate-fsm/src/index.ts
Andarist: no, it's fine

packages/xstate-fsm/test/fsm.test.ts
davidkpiano: Can even have a machine like this:

```js
const testMachine = createMachine({
  initial: 'green',
  states: {
    green: {
      on: { TARGET_INVALID: 'blue' }
    },
    yellow: {}
  }
});
```

If you want to copy-paste the above.

f-elix: Thanks! I just pushed the changes.

davidkpiano (approved):

davidkpiano:
@f-elix Can you add a changeset to this?

davidkpiano:
@f-elix Can you add a changeset to this?

.changeset/grumpy-dingos-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@xstate/fsm": patch
+---
+
+Added a more helpful error, in the development build, for a situation where the next state (the target of a transition) doesn't exist in the config.
```
Andarist, line 5: @davidkpiano I've added a changeset here - please recheck my grammar/wording before merging :p

Andarist (approved):

PR #2058, opened 2021-04-03
Resolve actions with xstate.stop event 

davidkpiano:


packages/core/src/interpreter.ts
```diff
@@ -510,7 +516,15 @@ export class Interpreter<
     }
 
     this.state.configuration.forEach((stateNode) => {
-      for (const action of stateNode.definition.exit) {
+      const exitActions = stateNode.definition.exit;
+      const [resolvedExitActions] = resolveActions(
+        this.machine,
+        this.state,
+        this.state.context,
+        toSCXMLEvent('xstate.stop'), // dummy event, for now
```
Andarist, line 524: according to SCXML, this should be `this.state._event` but I actually don't like this part of the SCXML spec 😬

davidkpiano: Yeah, I had that at first but realized this isn't accurate in real-world: the actions are caused by _stopping_ the state machine (hence `xstate.stop`) rather than by the previous event.

Andarist:
This got superseded by https://github.com/statelyai/xstate/pull/3126

PR #2060, opened 2021-04-03
[v5] Remove Machine()

davidkpiano:
This PR is a substantial refactor that replaces Machine() with createMachine().

### DIFF
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -1,8 +1,6 @@
 import {
   MachineImplementations,
-  DefaultContext,
   MachineConfig,
-  StateSchema,
   EventObject,
   AnyEventObject,
   Typestate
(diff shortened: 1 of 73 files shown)


packages/core/src/Machine.ts
```diff
@@ -1,44 +1,21 @@
...
-  TEvent extends EventObject = AnyEventObject,
-  TStateSchema extends StateSchema = any
->(
-  definition: MachineConfig<TContext, TEvent, TStateSchema>,
-  implementations?: Partial<MachineImplementations<TContext, TEvent>>
-): MachineNode<TContext, TEvent, TStateSchema>;
-export function Machine<
-  TContext = DefaultContext,
+export function createMachine<
+  TContext extends never,
```
Andarist, line 12: q: what's the purpose of this overload?

davidkpiano: I should add a comment... `any` extends `never`, so if the developer specifies `createMachine<any, any>`, it will correctly choose the context-event overload instead of the model overload.

PR #2065, opened 2021-04-05
DOCS - starting a service from nested state

bennyKY:
Add example of how to start a service from nested state


davidkpiano:
Thanks for the PR! However, this is already covered here: https://xstate.js.org/docs/guides/states.html#persisting-state

The main use-case for `.start(someState)` is for restoring persisted state.

PR #2076, opened 2021-04-09
Clarify that tests will fail unless "yarn build" has been ran

VanTanev:
This is something small, but it bit me when I first tried to do local changes to xstate.

`yarn dev` also will perform a build and make standalone `yarn test` work, but stating it explicitly like this seems most straightforward.


VanTanev:
Not sure if this needs a changeset

davidkpiano:
Thanks @VanTanev!

PR #2077, opened 2021-04-09
Make useSelector() watch for actor changes

VanTanev:
As discussed towards the end of #1966, useSelector() should recompute the value if the actor changes.

### DIFF
--- a/packages/xstate-react/src/useSelector.ts
+++ b/packages/xstate-react/src/useSelector.ts
@@ -45,7 +45,7 @@ export function useSelector<
     });
 
     return () => sub.unsubscribe();
-  }, [selector, compare]);
+  }, [actor, selector, compare]);
 
   return selected;
 }
--- a/packages/xstate-react/test/useSelector.test.tsx
+++ b/packages/xstate-react/test/useSelector.test.tsx
@@ -137,4 +137,95 @@ describe('useSelector', () => {
 
     expect(nameEl.textContent).toEqual('DAVID');
   });
+
+  it('should catch actor changes', () => {
+    const machine = createMachine<{ count: number; other: number }>({
+      initial: 'active',
+      context: {
+        other: 0,
+        count: 0
+      },
+      states: {
+        active: {}
+      },
+      on: {
+        OTHER: {
+          actions: assign({ other: (ctx) => ctx.other + 1 })
+        },
+        INCREMENT: {
+          actions: assign({ count: (ctx) => ctx.count + 1 })
+        }
+      }
+    });
+
+    let rerenders = 0;
+
+    const App = () => {
+      const service1 = useInterpret(machine);
+      const service2 = useInterpret(machine);
+      const [service, setService] = React.useState(service1);
+
+      const count = useSelector(
+        service,
+        React.useCallback(
+          (state: typeof service['state']) => state.context.count,
+          []
+        )
+      );
+
+      rerenders++;
+
+      return (
+        <>
+          <div data-testid="count">{count}</div>
+          <button
+            data-testid="toggle-service"
+            onClick={() => {
+              setService((service) =>
+                service === service1 ? service2 : service1
+              );
+            }}
+          ></button>
+          <button
+            data-testid="other"
+            onClick={() => service.send('OTHER')}
+          ></button>
+          <button
+            data-testid="increment"
+            onClick={() => service.send('INCREMENT')}
+          ></button>
+        </>
+      );
+    };
+
+    const { getByTestId } = render(
+      <React.StrictMode>
+        <App />
+      </React.StrictMode>
+    );
+    const countButton = getByTestId('count');
+    const otherButton = getByTestId('other');
+    const incrementEl = getByTestId('increment');
+    const toggleServiceButton = getByTestId('toggle-service');
+
+    fireEvent.click(incrementEl);
+
+    rerenders = 0;
+
+    fireEvent.click(otherButton);
+    fireEvent.click(otherButton);
+    fireEvent.click(otherButton);
+    fireEvent.click(otherButton);
+
+    expect(rerenders).toEqual(0);
+
+    fireEvent.click(incrementEl);
+
+    expect(countButton.textContent).toBe('2');
+
+    fireEvent.click(toggleServiceButton);
+
+    expect(rerenders).not.toEqual(0);
+    expect(countButton.textContent).toBe('0');
+  });
 });


VanTanev:
In 
https://github.com/davidkpiano/xstate/pull/2077/commits/5bf0fd06560be0ce9d4de3c8144ca95bccfa85ea I fixed a bug where `getSnapshot()` was not used inside `actor.subscribe()`.

It also points to the types needing amendment, but I don't know exactly how to do that -  the value that is given to the selector is not really:
```typescript
TEmitted = TActor extends Subscribable<infer Emitted> ? Emitted : never
```
It is rather:
```typescript
ReturnType<getSnapshot>
```

packages/xstate-react/src/useSelector.ts
```diff
@@ -39,13 +39,13 @@ export function useSelector<
     const initialSelected = selector(getSnapshot(actor));
     updateSelectedIfChanged(initialSelected);
 
-    const sub = actor.subscribe((emitted) => {
-      const nextSelected = selector(emitted);
+    const sub = actor.subscribe(() => {
+      const nextSelected = selector(getSnapshot(actor));
```
Andarist, line 43: this is something that @davidkpiano has wanted as `getSnapshot` is treated as reading the latest known value which could be out of date. In reality, I don't think this matters that much but this was not using this pattern intentionally. We could maybe reconsider this though.

Actor here is close to things like [`useSubscription`](https://github.com/facebook/react/blob/343710c9230d643ec442acef4634e380c1f0bce9/packages/use-subscription/src/useSubscription.js) and [`useMutableSource`](https://github.com/facebook/react/blob/343710c9230d643ec442acef4634e380c1f0bce9/packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js). And as such I think we should try to stick to semantics coined by the React team for those as that looks like a "blessed" way of doing this kind of thing.

The first notable difference between their implementation and this one is that they would keep the `actor` in the internal state. Based on that they would `setState` with the new actor in the render - using "the derived state from props" pattern.

Why this matter? When the `actor` changes, then with this the next **committed** selected state to the screen will be stale (from the previous actor) and the update with the new state will be scheduled right away. So there are 2 takeaways from this:
1. we go through the render->commit cycle twice
2. the user sees the stale value for longer than they should and if the same source state is also displayed somewhere else on the page then what will be displayed here and what will be displayed there won't match. This is known as "tearing"

With the algorithm used by the React team, this doesn't happen as the new committed value will be the correct one right away - at the cost of calling the render function twice but that should be cheaper.

VanTanev: I would like to pull the discussion back a bit though:
1) Previously the hook did not even register actor changes, so we're not really breaking on the render -> commited  commit cycle.
2) `getSnapshot()` is defined in the docs as:
> a function that should return the latest emitted value from the actor.
Defaults to attempting to get the actor.state, or returning undefined if that does not exist. Will automatically pull the state from services.

All that this PR attempts to do is bring the code in line with the documentation.

Tangentially: I now noticed that the `getSnapshot()` change needs to happen here too:
https://github.com/davidkpiano/xstate/blob/7d07bfb1ade3137acb1bb6fff33b23d32bfa91a6/packages/xstate-react/src/useActor.ts#L64

--- 

To bring it back to this specific discussion: 
> When the `actor` changes, then with this the next **committed** selected state to the screen will be stale (from the previous actor) and the update with the new state will be scheduled right away. 

If I understand it correctly, to fix this we would need to adapt this code: https://github.com/facebook/react/blob/343710c9230d643ec442acef4634e380c1f0bce9/packages/use-subscription/src/useSubscription.js#L41-L56

and, in our case, instead of `getCurrentValue`, it would be `getSnapshot`, correct?

Andarist: > so we're not really breaking on the render -> commited commit cycle.

Could you rephrase? I'm not sure if I got what you have meant here.

> All that this PR attempts to do is bring the code in line with the documentation.

Can you get any more recent value by calling `getSnapshot` than the one you have just received in the `subscribe` callback? Those values should be really the same so there should be no need to call `getSnapshot` in the callback. Some actors might not expose a property for reading the "snapshot" but they can still emit a value and this is something that won't be handled by the proposed changes.

> If I understand it correctly, to fix this we would need to adapt this code:

Yes. However, I would start with keeping the `actor` and the `value` in the state to see how it goes. I'm not yet sure if we need to keep the `getSnapshot` there.

VanTanev: > Could you rephrase? I'm not sure if I got what you have meant here.
I meant, changing the actor flat out did not work before. Introducing a working version with tearing is a big step in the right direction, even if not a perfect solution.

> Can you get any more recent value by calling getSnapshot than the one you have just received in the subscribe callback?
It's not about being more recent. It's about being able to specify a custom `getSnapshot()` implementation, and that implementation being honored. Currently, overwriting `getSnapshot()` in a way that produces different result from the value emitted on the subscription does not work. Now, a reasonable question is, "Should that even work?" I think that it should, but I'm not married to that position.

> Yes. However, I would start with keeping the actor and the value in the state to see how it goes. I'm not yet sure if we need to keep the getSnapshot there.
Should I try to write a `useSubscription()`-style implementation in this PR?

Andarist: > Introducing a working version with tearing is a big step in the right direction, even if not a perfect solution.

Sure, but since we know that the issue is there we can try to fix avoid it altogether - especially that steps towards avoiding it are clear.

> Still, it seems weird to only use getSnapshot() on the initial value, and not on emit events.

I agree and I had similar concerns when this was introduced. But `TEmitted` is used for both the emitted value and the return type of the `getSnapshot` so both really should be the same and I think this is a reasonable assumption. We don't want to support "arbitrary" snapshots.

VanTanev:
@Andarist @davidkpiano I adopted the style of [useSubscription](https://github.com/facebook/react/blob/343710c9230d643ec442acef4634e380c1f0bce9/packages/use-subscription/src/useSubscription.js), with some caveats:

1) React's `useSubscription()` expects the `getCurrentValue()` callback to be memoized, but the current `useSelector()` API does not expect `selector` to be memoized. To get similar behavior, the value is recalculated on every render, and if different from the previous render, it's immediately returned. This fixes the tearing issue without requiring a breaking API change.
2) Similarly, `compare` and `getSnapshot` are not required to be memoized. However, if any of the params to the hook is not memoized, and the hook is rerendered because of a parent rerender, we will be doing a resubscribe to the actor. I'm not sure if that's an expensive operation, but there is no way around it.

We should consider a memoized API for V5, maybe?
```typescript
const service = useInterpret(machine)
const value = useSelector(useMemo(() => ({
  actor: service,
  selector: state => state.context.count,
}), [service]))
```

VanTanev:
I tried to write a object param version of use selector: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgVwM4FMCiAzb6DGMANChgMowCGM6Ja6AIugEbIDmAapQDbLpwBfONigQQcAORR0lQhIDcAKFCRYiOAEFC0AErpsJAJIA7GlDDSzJMsmap8UYM0rNu-ISLGSAHqio0FZXBoeCRgVC0YaAB1YBgACwpqd2FRcQkAAV9-dAB6aVkYXO4nXPpI6ECVEPU2dBgydCgAN2B8dDJjSjBUeIh4DzTJLL9k-JlCYtL6Rpa29EDFbGRjQmAIYzhw2db2gApC6AAuOEpjAE8AShPDqC3UOBMzC3qmgB4z85JP74vf84AfIhFHA4JZkFBNhJRgEtptbnAAGSIyQgWTxYDGBZw07aKBKASKRT4DZ+OAAE30lGQ3BgAGExGBKNI4ABeOAHEjMS5soGUNms9nMJQk4xkynYam0gDi9U63V6-TZHMoPNZAJB9x28wOl01oIA-HA6g0mrsOl0en0YLr9XATuEKlBYgkkjRbaDDacAHQw9B2k4rCWY9DkpSKdDeVTwGDnMD8RpuPEM4zYYBsN4AFSdJEzJEwIDiNHJ1nQSaiUCB7IQmtuJ2zeKUoIw5eOcETBArTbgJPAzPQBpOB3rXPraqBzAgEDcZ27JvlVv6g5V9ad47gBaLoYJSkj0eEKzWG1IHTLnegb01eavTrgkZoxnJDydemwHz+pwuVc0eNf76+n6AkQV6bjAxbKg2FZ3t4D5Pu2tj2I4ziuOgbyYngdygcWQJGlhob2igxgANbGBAADuxiKACeyimmbAnDWoJ1nAkHQN2LbnlAQ7oIWYGhvWeHkuumbdr2TLSMuw4saOLHrpO04yMYc5ypaiowJJzGsVAwmCQS1wsbuUY1MsqwwOsmwzGeeKXqCWnQbBz6-vo-7-ACwG2e5LGCRBt73ugj4PDYdgOE4LhuGhqZNBuvHYXAuExfhJxYs0TRUXstZ4queKeRxmUcjxW7kgJCVCby0mamJ-YaSOcDMGOZXyTOxiefOqnWtVLFrmVmaCYo+kiRGRlqCZR4WeQVkVns3rTcybCoDcFwANoALo8oxcBuPAzEvs5PyAQC3abXAuUVkOlALVcZWfId9Q9oy-ZnRdMmfHJU5NTd8CtQq1pnXlL1XRc3bANgHISLcEg4rNqCLQADMtSIotCE2VJDUBzbDq3Ap6HLrdjtyedjJ3QATnqVSy7ISlK9L3dIJOgl9i7wBTVI0jAsoNG1-R2kI7JQxjlzdkIZYYFjnqLXaTHZRLx3I1AdN3X25MUiztIMortDSwzanKpTrPswual2vDvNo6ggtEqCopkotfokBgDQ5MbJ5uugex7Ouey47iFYk80PB8CcRNQHsWvWgceKXHqoICJH4agkdfu8OgmYQHoMAQps7J+t6id8B9cC5-wWeyyHKnff04cVrHmrAxy2cIgAhIK3vQHAAA+bdwA3ZOu4XKdpxnJCF5Hougn3qf1BnyqF92zZyjknvS-j0uFyc48D5C3MC5qhKavQTCsJw-u98f-eT5C297xgOB4IQbvrl7R3ksA5IAKpighIXMEXwg8Bgs93TFPAfA8QCBEQAGLQFfmAckyQHjsj2AVPiRVAIP2lrXPYz834f2CkhdAa1pagnBJCX+3B-7S13tjS2pJ4CF2VEHRBJVt5UJlg7ZIbsXjNBdmglhoIMGcJdt6RuzdbgEN4Z6YhmwBE5AAdjShvCME9z2NI5IOdj5D2PiPL2LDJFwBUTQWRnp5E6PPpsJA01vT6NoAXY+ghDExwAQIABVt4CoFsMqW4vpP54JoqA-AECoEwLgcw0mfiAlQGgbAmgqBS4c3LjaURl9sa6PdmVbRFIX7vzcbgpwP8YBQDztLbJ3oVjZMQrk92jjBYkEWvjGWrY5YK3EtY0O-RVpxzBKYmxScz7p0hASIkiggA
```typescript
export function useSelector<
  T,
  TActor extends ActorRef<any, any> = ActorRef<any, any>,
  TEmitted = TActor extends Subscribable<infer Emitted> ? Emitted : unknown
>(config: {
  actor: TActor;
  selector: (emitted: TEmitted) => T;
  compare?: (a: T, b: T) => boolean;
  getSnapshot?: (actor: TActor) => TEmitted;
}): T;
```
When used with `useMemo()` it does require specifying the output `T`, but still seems overall better than having to do up to 3 `React.useCallback()` calls (for `selector`, `compare` and `getSnapshot`)
```typescript
        const count = useSelector<number>(
          React.useMemo(
            () => ({
              actor: service,
              selector: (state) => state.context.count
            }),
            []
          )
        );
```

VanTanev:
What are the next steps here?

davidkpiano:
> What are the next steps here?

I'll take a deeper look at this soon! (Sorry, extremely busy week)

VanTanev:
I want to resurrect this PR when you guys have the time.

davidkpiano (commented):
> I want to resurrect this PR when you guys have the time.

Will get to it today!

packages/xstate-react/src/useSelector.ts
```diff
@@ -24,28 +24,52 @@ export function useSelector<
...
+  let value = selector(getSnapshot(actor));
+  if (state.actor !== actor || !compare(valueToReturn, value)) {
+    valueToReturn = value;
+    setState({
+      actor,
+      value: valueToReturn
+    });
+  }
+
+  useDebugValue(valueToReturn);
```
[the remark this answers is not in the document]
VanTanev: Not strictly - it's a dev-only helper that shows you the value of the selector in React Developer Tools. It seems OK to have it, but we can easily remove it.

VanTanev:
Ping. This should be considered before the getSnapshot() changes are released, I think.

packages/xstate-react/README.md
```diff
@@ -221,6 +221,21 @@ const App = ({ service }) => {
...
+
+const App = ({ service, offset }) => {
+  const count = useSelector(
+    service,
+    React.useCallback((state) => state.context.count + offset, [offset])
+  );
+
+  // ...
+};
+```
```
davidkpiano, line 237: Wouldn't this be better done as this?

```js
const firstcount = useSelector(/* ... */);
const count = useMemo(() => firstCount + offset, [firstCount, offset]);
```

packages/xstate-react/src/useSelector.ts
davidkpiano: @Andarist @VanTanev Is this resolved?

davidkpiano:
Before this goes in, let's consider assuming that `compare` and `getSnapshot` are memoized/stable, if that's the same behavior in React's built-in hooks. Thoughts @Andarist ?

packages/xstate-react/README.md
VanTanev: Either variation is valid. This however exemplifies that if the selector needs to change for some reason, it needs to be wrapped in a `React.useCallback()`. We can't know for sure which approach will be chosen by users, so it makes sense to call attention to the fact that functions that change their return value must be memorized with `React.useCallback()`.

VanTanev:
> Before this goes in, let's consider assuming that `compare` and `getSnapshot` are memoized/stable, if that's the same behavior in React's built-in hooks. Thoughts @Andarist ?

Where does React assume parameters are stable? The `useSubscription()` hook [explicitly states that its params must be memoized by the user](https://github.com/facebook/react/tree/master/packages/use-subscription#subscribing-to-event-dispatchers).

davidkpiano:
> > Before this goes in, let's consider assuming that `compare` and `getSnapshot` are memoized/stable, if that's the same behavior in React's built-in hooks. Thoughts @Andarist ?
> 
> Where does React assume parameters are stable? The `useSubscription()` hook [explicitly states that its params must be memoized by the user](https://github.com/facebook/react/tree/master/packages/use-subscription#subscribing-to-event-dispatchers).

Ah, nevermind, misread previous comments.

packages/xstate-react/README.md
```diff
@@ -252,6 +267,22 @@ const App = ({ service }) => {
 };
 ```
 
+With Typescript, you might want to define the selector and comparator inline for better type inference. In that case, you still need to use `React.useCallback()` for both. To get proper types you need to provide the 2 generic arguments to `useSelector()` - the type of the service we're selecting from, and the output type of our selector.
```
Andarist, line 270: when using inline functions, like here - shouldn't types be just inferred properly by TS?

Andarist:
@VanTanev sorry for not taking care of this PR sooner. Could you refresh my memory a little bit on it? I'm wondering if it wouldn't be easier to just use `useSubscription` as we try to follow its "style" in the implementation. If that's not possible - why? what are the notable differences between that hook and the implementation here?

VanTanev:
I'll try to answer soon. Am away from computer the last few days.

PR #2079, opened 2021-04-11
Add support for single-state (combinatorial) machines

davidkpiano:
This PR adds support for combinatorial machines:

- No `initial` property
- No `state` property
- Root-level transitions

```js
const testMachine = createMachine({
  context: { value: 42 },
  on: {
    INC: {
      actions: assign({ value: (ctx) => ctx.value + 1 })
    }
  }
});
```

The finite `state.value` of the combinatorial machine is an empty object: `{}`

### DIFF
--- a/.changeset/six-chicken-call.md
+++ b/.changeset/six-chicken-call.md
@@ -1,0 +1,18 @@
+---
+'xstate': minor
+---
+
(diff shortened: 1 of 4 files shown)


mattpocock (approved):

Andarist (approved):
This almost feels like a bug fix to me 😅

PR #2081, opened 2021-04-12
Speedup execution of "yarn dev" tests

VanTanev:
While working on #2077, I found that running `jest` with `--runInBand` (which means, run only one process for Jest without spawning a separate worker for each test) was noticeably faster.

Changing `packages/xstate-react/src/useSelector.ts`:

Before:
```
 PASS  packages/xstate-react/test/types.test.tsx
 PASS  packages/xstate-react/test/useInterpret.test.tsx
 PASS  packages/xstate-react/test/useService.test.tsx
 PASS  packages/xstate-react/test/useActor.test.tsx
 PASS  packages/xstate-react/test/useSelector.test.tsx
 PASS  packages/xstate-react/test/useMachine.test.tsx

Test Suites: 6 passed, 6 total
Tests:       42 passed, 42 total
Snapshots:   1 passed, 1 total
Time:        1.32 s
Ran all test suites related to changed files.

Watch Usage: Press w to show more.
```

After:
```
 PASS  packages/xstate-react/test/useSelector.test.tsx
 PASS  packages/xstate-react/test/useMachine.test.tsx
 PASS  packages/xstate-react/test/useActor.test.tsx
 PASS  packages/xstate-react/test/useService.test.tsx
 PASS  packages/xstate-react/test/useInterpret.test.tsx
 PASS  packages/xstate-react/test/types.test.tsx

Test Suites: 6 passed, 6 total
Tests:       42 passed, 42 total
Snapshots:   1 passed, 1 total
Time:        0.83 s, estimated 1 s
Ran all test suites related to changed files.

Watch Usage: Press w to show more.
```

```
Ran on Intel 9900KS @ 5GHz, Ubuntu 16.04
```


Andarist:
What's the general guidance from the Jest team around this? Why parallelism affect us that much and where is usually the tipping point where it starts to pay off?

VanTanev:
@Andarist I wrote a blog post about this 😅 https://dev.to/vantanev/make-your-jest-tests-up-to-20-faster-by-changing-a-single-setting-i36

Basically: It's both system and testsuite dependent, but there are some general trends that I gleaned from talking to a lot of people who ran tests on their machines:

1) Changing the default behavior results in a performance degradation only on some old non-hyperthreaded Intel CPUS.
2) The sweet spot for running all tests is `--maxWorkers=50%` - which runs on the half available CPU threads. The best performance would be to run on all available CPU cores (and ignore hyperthreading), but NodeJS does not expose true core count vs thread count
3) For watch mode, it's test dependent, but in my testing `--maxWorkers=25%` or `--runInBand` perform the best. 

Why changing the defaults works at all?

There is a startup penalty that needs to be paid for each worker process started. If the tests themselves are fast, the startup penalty ends up larger than the saving from executing the tests in parallel.

davidkpiano:
I'd be willing to adopt it. Tests are pretty slow (not painfully, but it's noticeable).

VanTanev:
@davidkpiano Running all tests on a 2016 macbook pro (2 core / 4 threads):

```
➜  xstate git:(master) hyperfine 'yarn test' 'yarn test --maxWorkers=50%' -w 1 -m 3
Benchmark #1: yarn test
  Time (mean ± σ):     13.763 s ±  0.415 s    [User: 20.601 s, System: 3.114 s]
  Range (min … max):   13.494 s … 14.241 s    3 runs

Benchmark #2: yarn test --maxWorkers=50%
  Time (mean ± σ):     12.520 s ±  0.243 s    [User: 17.809 s, System: 2.674 s]
  Range (min … max):   12.324 s … 12.791 s    3 runs

Summary
  'yarn test --maxWorkers=50%' ran
    1.10 ± 0.04 times faster than 'yarn test'
```

It's not a lot, but it's something.

package.json
```diff
@@ -13,9 +13,10 @@
     "postinstall": "manypkg check",
     "clean:packages": "yarn workspaces run clean",
     "build": "tsc -b ./tsconfig.monorepo.json",
-    "test": "jest",
-    "test:core": "npm test --prefix packages/core",
-    "dev": "node ./scripts/dev.js",
+    "build:dev": "tsc -b ./tsconfig.monorepo.json -w",
```
Andarist: IMHO this would be better as `dev` is already "reserved" for the "dev" script and this has no direct relation to it (even though the dev script is kinda using it already)

```suggestion
    "build:watch": "tsc -b ./tsconfig.monorepo.json -w",
```

VanTanev:
@Andarist @davidkpiano 
I changed direction and instead focused on the double/triple running of tests.

Previously, when `build -w` and `test --watch` commands were run simultaneously, each change in sources leading to a rebuild of one (or several packages). Then Jest detecting both the initial change, as well as the changes of all build artefacts, and ran the tests 2-3 times for each change (from my experience) instead of one time.

I updated jest to be able to run even without build, you can check with:
```
$ yarn clearn:packages
$ yarn test
```

And made it so it that it ignores build artifacts in watch mode.

package.json
```diff
@@ -13,9 +13,9 @@
     "postinstall": "manypkg check",
     "clean:packages": "yarn workspaces run clean",
     "build": "tsc -b ./tsconfig.monorepo.json",
-    "test": "jest",
-    "test:core": "npm test --prefix packages/core",
-    "dev": "node ./scripts/dev.js",
```
Andarist, line 18: this is equivalent and purely stylistic (up to personal preferences) - I would in general try to avoid such changes in PRs as they introduce some noise and don't bring any value

tsconfig.base.json
```diff
@@ -14,6 +14,12 @@
...
     "jsx": "react",
     "lib": ["es6", "dom"],
-    "strictPropertyInitialization": true
+    "strictPropertyInitialization": true,
+    "paths": {
+      "xstate": ["./packages/core/src"],
+      "xstate/lib/*": ["./packages/core/src/*"],
+      "@xstate/graph": ["./packages/xstate-graph/src"],
+      "@xstate/fsm": ["./packages/xstate-fsm/src"]
+    }
```
Andarist, line 23: what's the practical outcome for this for us? given we use composite projects here this all gets correctly re-mapped already (apart from the `xstate/lib/*`)

packages/core/jest.config.js
```diff
@@ -2,5 +2,6 @@ const { jest: lernaAliases } = require('lerna-alias');
 
 module.exports = {
   preset: 'ts-jest',
-  moduleNameMapper: lernaAliases()
+  moduleNameMapper: lernaAliases(),
+  watchPathIgnorePatterns: ['/lib/', '/dist/', '/es/']
```
Andarist, line 6: does this bring any noticeable gains? we already have a very different setup on the next branch (which simplifies stuff quite a bit automatically~) so if those changes to `jest.config.js` files are not necessarily then I would like to avoid those to avoid merge conflicts etc

package.json
VanTanev: This was reverted

packages/core/jest.config.js
VanTanev: In `yarn dev`, changing a file used to cause the jest tests to run two+ times for each single change. This fixes it.

tsconfig.base.json
VanTanev: This is so that tests do not require build to be ran beforehand - so jest no longer depends on build artifacts, which makes it safe to ignore build artifacts in jest's watcher.

package.json
VanTanev: Done 👍

VanTanev:
Is there anything else I can prepare for this, or should we close it in favor of whatever setup v5 has?

Andarist:
@VanTanev this week has been very busy for me and I didn't have time to dig into this yet - gonna do that as soon as possible.

davidkpiano:
Doing some cleanup - we might move over to vitest, but for now, test performance is acceptable. Thank you!

PR #2100, opened 2021-04-15
Fixes #2084: `interpert` vs. `useInterpret`

### LINKED ISSUE #2084: Inconsistent behavior for initial state using `interpret` vs. `useInterpret`
**Description**
Using `interpret` and `useInterpret` results in different runtime behavior.
If a machine **without initial context** is started using `interpret` the `context` resolves to `undefined`, if it is started with `useInterpret` the context resolves to `{}`, this in turn affects the behavior of `assign(...)`.
Because of the huge advantage of xstate being framework independent, I think that's unwanted or even potential risky behavior.
This inconsistency can easily be solved by simply providing an initial context, but **it currently fails silently if one switches from `useInterpret` to `interpret`**.

The reason for this is: (@VanTanev 🙏)
https://github.com/davidkpiano/xstate/blob/14cdef612df29b9e067a34ca96b0750e729c9aff/packages/xstate-react/src/useInterpret.ts#L90-L93

> [@VanTanev on discord](https://discord.com/channels/795785288994652170/799416943324823592/831444212322533418): `useInterpret()` calls `machine.withConfig()`. The second argument to `withConfig()` is a new `context` object, that overwrites `context: undefined` in the original machine definition. Here it's kind of accidental, and possibly **a more consistent way to write this would be something like**:

```ts
const machineWithConfig = machine.withConfig(machineConfig, (machine.context || context ? {
  ...machine.context,
  ...context
} : undefined) as TContext);
```

**Expected Result**
Consistent runtime behavior using either `interpret` or `useInterpret`. Preferably the behavior of `interpret`, because it is the underlying framework-agnostic version of `useInterpret`. _The power of writing framework-agnostic logic code is just too valuable to be given up 😍._

**Actual Result**
Different runtime behavior (see **Reproduction**).

**Reproduction**
_xstate 4.17.1; @xstate/inspect 0.4.1; @xstate/react 1.3.1_
`interpret`-Example: https://codesandbox.io/s/xstateinitial-context-interpret-24bbz ⮕ this results in `context: undefined`
`useInterpret`-Example: https://codesandbox.io/s/xstateinitial-context-useinterpret-33950 ⮕ this results in `context: {}`

**Additional context**
- _There already was a little bit of discussion on discord about this:_
  - https://discord.com/channels/795785288994652170/799416943324823592/831217793479082024
  - https://discord.com/channels/795785288994652170/799416943324823592/831264646274482210
  - https://discord.com/channels/795785288994652170/799416943324823592/831444212322533418
  - https://discord.com/channels/795785288994652170/799416943324823592/831444929062240266
- Changing this behavior might be a breaking change ⚡.

bemayr:
While working on #2084 I realized that the `useSelector` hook in @xstate/vue faces the same little inconsistency, so I decided to fix both.

The according tests are written and fail without the patch. Thanks again to @VanTanev for suggesting the fix itself! 👏

I am not sure about the changeset though, whether this is a breaking change or not. I decided to go for a *patch* version because according to the docs an initial context property should be provided anyway, so I classified it as a fix.

---

_BTW_: Thanks for the awesome project setup @davidkpiano and @Andarist (and of course everybody who contributed as well), it is such a joy that everything works out of the box (even on windows :o)! 🤩

### DIFF
--- a/packages/xstate-react/test/useInterpret.test.tsx
+++ b/packages/xstate-react/test/useInterpret.test.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { createMachine } from 'xstate';
+import { createMachine, interpret } from 'xstate';
 import { render, cleanup, fireEvent } from '@testing-library/react';
 import { useInterpret } from '../src';
 
@@ -110,5 +110,29 @@ describe('useInterpret', () => {
     rerender(<App value={42} />);
 
     expect(actual).toEqual([1, 42]);
+  });
+
+  it('should behave the same as `interpret` when initial context is not defined', (done) => {
+    const machine = createMachine({
+      initial: 'foo',
+      states: {
+        foo: {}
+      }
+    });
+
+    const interpretService = interpret(machine);
+
+    const App = () => {
+      const useInterpretService = useInterpret(machine);
+
+      expect(useInterpretService.initialState.context).toEqual(
+        interpretService.initialState.context
+      );
+      done();
+
+      return null;
+    };
+
+    render(<App />);
   });
 });
--- a/packages/xstate-react/test/useInterpret.test.tsx
+++ b/packages/xstate-react/test/useInterpret.test.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { createMachine, interpret } from 'xstate';
 import { render, cleanup, fireEvent } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
 import { useInterpret } from '../src';
 
 afterEach(cleanup);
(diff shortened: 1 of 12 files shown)


VanTanev:
... oops :D

bemayr:
> Even if this is fixing an inconsistency I would rather say that this should stay this way until the next major release. Changing an object to undefined might blow somebody's app 😬

Yes, this was what I immediately thought as well 😕.
Then we should also think of **@xstate/svelte** where `useMachine(...)` has the same inconsistency: https://github.com/davidkpiano/xstate/blob/14cdef612df29b9e067a34ca96b0750e729c9aff/packages/xstate-svelte/src/index.ts#L58-L61

So this should be postponed until new major releases of _@xstate/react_, _@xstate/vue_ and _@xstate/svelte_?

And regarding #2099 (@VanTanev, awesome that we even chose the same test 😂), what's the favored way of implementing the tests? (maybe @Andarist?) Checking whether the implementation is consistent with _xstate-core's:`interpret`_ or asserting the behavior individually, like in [@VanTanev's PR](https://github.com/davidkpiano/xstate/pull/2099/files#diff-8111112444e8a54087f52d1dd28838b4393b3bae28da23461502492cee5103bdR124-R126)?

packages/xstate-react/test/useInterpret.test.tsx
```diff
@@ -111,4 +111,28 @@ describe('useInterpret', () => {
...
+        foo: {}
+      }
+    });
+
+    const interpretService = interpret(machine);
+
+    const App = () => {
+      const useInterpretService = useInterpret(machine);
+
+      expect(useInterpretService.initialState.context).toEqual(
```
Andarist, line 128: does the same issue happen for `useInterpretService.state.context`? If yes - then I would probably prefer to test that as using `initialState` is probably way less popular. Not even sure if we should expose `initialState` publicly on services, they are mostly useful for machines 🤔

Andarist:
> So this should be postponed until new major releases of @xstate/react, @xstate/vue and @xstate/svelte?

that would be my preferred approach here - u could target the `next` branch with this

> And regarding #2099 (@VanTanev, awesome that we even chose the same test 😂) what's the favored way of implementing the tests?

Since you have both gone through the trouble of handling this - maybe just merge both PRs? both assertions make sense, we need to test individual behavior and we can also add an extra assertion that would ensure this matches the core's `interpret`. And both could be just in a single test - probably not worth splitting into separate ones unless you think otherwise.

packages/xstate-react/test/useInterpret.test.tsx
bemayr: Thanks for this, I now know what you mean with `initialState` vs `state` vs `machine` and rewrote the tests for consistency in 29c2e33.

bemayr:
Closing this, because @VanTanev incorporated the changes in #2099. Thanks for doing that! 🙏
I am gonna open a PR to your branch with my latest changes, as this makes more sense. 👌

PR #2102, opened 2021-04-15
Provide convenience type InterpreterFrom<typeof machine>

VanTanev:
Add the convenience type
```typescript
export type InterpreterFrom<
  T extends StateMachine<any, any, any, any>
> = T extends StateMachine<
  infer TContext,
  infer TStateSchema,
  infer TEvent,
  infer TTypestate
>
  ? Interpreter<TContext, TStateSchema, TEvent, TTypestate>
  : never;
```

This is useful in the following pattern:
```typescript
import React from 'react'
import { createMachine, InterpreterFrom } from 'xstate'
import { useService } from '@xstate/react'

const machine = createMachine(...)
type MachineService = InterpreterFrom<typeof machine>

const MachineServiceContext = React.createContext<MachineService>(null!)
const useMachineService = (): MachineService  => {
  const service = React.useContext(MachineServiceContext)
  if (!service) throw new Error(`You must call useMachineService in a child component of <MachineProvider>`)
  return service
}
const MachineProvider: React.FC = ({ children }) => {
  const service = useInterpret(machine)
  return <MachineServiceContext.Provider value={service}>{children}</MachineServiceContext.Provider>
}

const App = () => {
  return <MachineProvider><Compontent /></MachineProvider>
}
const Component = () => {
  const service = useMachineService()
  const [state, send] = useService(service)
  // ...
}
```

This is a companion type to `ActorRefFrom<T>`.

### DIFF
--- a/.changeset/fifty-owls-check.md
+++ b/.changeset/fifty-owls-check.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Provide convenience type InterpreterFrom<typeof machine>
--- a/.changeset/fifty-owls-check.md
+++ b/.changeset/fifty-owls-check.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Provide convenience type InterpreterFrom<typeof machine>
+Provide a convenience type for getting the `Interpreter` type based on the `StateMachine` type by transferring all generic parameters onto it. It can be used like this: `InterpreterFrom<typeof machine>`
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1285,3 +1285,14 @@ export type ActorRefFrom<
   : never;
 
 export type AnyInterpreter = Interpreter<any, any, any, any>;
+
+export type InterpreterFrom<
+  T extends StateMachine<any, any, any, any>
+> = T extends StateMachine<
+  infer TContext,
+  infer TStateSchema,
+  infer TEvent,
+  infer TTypestate
+>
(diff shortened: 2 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -1285,3 +1285,14 @@ export type ActorRefFrom<
   : never;
 
 export type AnyInterpreter = Interpreter<any, any, any, any>;
+
+export type InterpreterFrom<
```
Andarist, line 1289: Could you also reexport this from the root file? While accessing `/types` is currently possible it is not recommended and will probably break in the future.

.changeset/fifty-owls-check.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Provide convenience type InterpreterFrom<typeof machine>
```
Andarist, line 5:
```suggestion
Provide a convenience type for getting the `Interpreter` type based on the `StateMachine` type by transferring all generic parameters onto it. It can be used like this: `InterpreterFrom<typeof machine>`
```

packages/core/src/types.ts
Andarist: as @davidkpiano has pointed out to me, this is already handled: https://github.com/davidkpiano/xstate/blob/14cdef612df29b9e067a34ca96b0750e729c9aff/packages/core/src/index.ts#L77

Andarist (approved):

PR #2107, opened 2021-04-15
Feat: restrict invoked Subscribable to EventObject

woutermont:
Adds feature from #2106.

Signed-off-by: Wouter Termont <woutermont@gmail.com>


packages/core/src/types.ts
```diff
@@ -264,7 +264,7 @@ export type InvokeCreator<
 ) =>
   | PromiseLike<TFinalContext>
   | StateMachine<TFinalContext, any, any>
-  | Subscribable<any>
+  | Subscribable<EventObject>
```
Andarist, line 267: This makes sense since we expect that emitted items are of this type: https://github.com/davidkpiano/xstate/blob/14cdef612df29b9e067a34ca96b0750e729c9aff/packages/core/src/interpreter.ts#L1156

Technically we also support strings and SCXML events here at runtime but I don't think that any of those are worth supporting here.

.changeset/gorgeous-kiwis-drive.md
```diff
@@ -0,0 +1,13 @@
+---
+'xstate': major
```
Andarist, line 2:
```suggestion
'xstate': patch
```

.changeset/gorgeous-kiwis-drive.md
```diff
@@ -0,0 +1,13 @@
...
+
+This update restricts invoked `Subscribable`s to `EventObject`s,
+so that type inference can be done on which `Subscribable`s are
+allowed to be invoked.
+
+Existing `MachineConfig`s that invoke `Subscribable`s that are not
+`Subscribable<EventObject>`s should be updated accordingly:
+either their emissions are desired, and should thus be acted on
+as events, or they are not, in which case an action would
+probably be a better choice than a service.
```
Andarist, line 13: I don't quite understand this advice - could you rephrase this somehow or explain this in more detail in an answer to me?

woutermont: Sure, but it depends on whether you see this as a breaking change ore not. If current implementations follow the documentation, no machine configs will exist that have subscribable services that emit non-events. However, strictly following the current typings, it might be possible that such machines _do_ exist, even though they are quiete useless since the emitted values of those subscribables would not end up anywhere. Therefore my advice was to either emit events ifthe output of the subscribable is relevant, or use an action instead of a service if not. Not sure it is the perfect choice in every case though.

.changeset/gorgeous-kiwis-drive.md
woutermont: Fine by me, but it COULD break compilation of machine configs that currently invoke subscribables that output non-events.

packages/core/src/types.ts
woutermont: I would also be happy with a union type `Subscribable<string | EventObject | SCXML.Event<EventObject>` (not entirely sure of that last part, since I haven't worked with SCXML yet).

.changeset/gorgeous-kiwis-drive.md
Andarist: Sure, but they probably don't work at runtime either. Quite frankly - every change is potentially a breaking change for somebody. Semver must be applied with some dose of pragmatism ;p

packages/core/src/types.ts
Andarist: I think this can be skipped for now and we can just reconsider doing this in the future

woutermont:
Voila, some pragmatism and an updated description ... ready for merging.

VanTanev:
This should be coordinated with #2125. It seems to me that `Subscribable` should use `TEvent`, which as defined in that PR is `TEvent extends EventObject = AnyEventObject`

VanTanev:
@Andarist Should I incorporate this PR into #2125 ?

davidkpiano (approved):

PR #2118, opened 2021-04-19
Add assertEvent() helper

VanTanev:
Add `assertEvent()` helper.

Example:

```typescript
import { assign, createMachine, ExtractEvent } from 'xstate';
import { assertEvent } from 'xstate/lib/assertEvent';

type Context = {
  date: Date;
};

type Event = 
 | { type: 'SHOW_DATE' } 
 | { type: 'UPDATE_DATE'; date: Date };

const machine = createMachine<Context, Event>(
  {
    on: {
      SHOW_DATE: { actions: 'showDate' },
      UPDATE_DATE: { actions: 'updateDate' }
    }
  },
  {
    actions: {
      showDate: (ctx, e) => {
        assertEvent(e, 'SHOW_DATE');
        alert(ctx.date.toISOString());
      },
      updateDate: assign({
        date: (_, e) => {
          assertEvent(e, 'UPDATE_DATE');
          return e.date; // has type "Date" !
        }
      })
    }
  }
);
```

### DIFF
--- a/.changeset/ten-buckets-exist.md
+++ b/.changeset/ten-buckets-exist.md
@@ -2,7 +2,9 @@
 'xstate': patch
 ---
 
-Add `assertEvent()` helper.
+Add `assertEvent()` helper that throws an exception when the given event is not of the expected type.
+
+This is usefull to type-narrow events inside actions, and make sure that expected data exists on an event object.
 
 Example:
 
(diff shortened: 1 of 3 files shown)


.changeset/ten-buckets-exist.md
```diff
@@ -0,0 +1,41 @@
+---
+'xstate': patch
+---
+
+Add `assertEvent()` helper.
```
with-heart, line 5: I feel like we might want to add a note that this throws if the assertion fails. That's not entirely clear from the example alone imo, and that's all most users will see in the release notes.

VanTanev: Roger, will update the changeset

packages/core/src/assertEvent.ts
```diff
@@ -0,0 +1,23 @@
+import { EventObject, ExtractEvent } from './types';
+
+export function assertEvent<
+  TEventType extends TEvent['type'],
+  TEvent extends EventObject = EventObject
+>(e: TEvent, type: TEventType): asserts e is ExtractEvent<TEvent, TEventType>;
+export function assertEvent<
+  TEventType extends TEvent['type'],
+  TEvent extends EventObject = EventObject
+>(e: TEvent, type: TEventType[]): asserts e is ExtractEvent<TEvent, TEventType>;
```
Andarist, line 10: 1. Would using a single overload with `e: TEvent | TEvent[]` change something here? for now I'm just curious, not requesting any changes
2. I would prefer replacing `e` with `event` - this stuff pops up in the IDE tooltips and it would be best for this to avoid abbreviations
3. I would, personally, reorder those generics. I feel like it's easier to make a mistake if we use the proposed order because `assertEvent<'FOO'>` will use the `EventObject` since partial inference is not supported by TS. I understand that the optional type param can't be used first but I think it should be OK to just drop that default here.

VanTanev: > Would using a single overload with e: TEvent | TEvent[] change something here? for now I'm just curious, not requesting any changes

I thought to try this too. For some mystic eldritch reasons, it does not work with that definition.

> I would prefer replacing e with event - this stuff pops up in the IDE tooltips and it would be best for this to avoid abbreviations

Done

> I would, personally, reorder those generics. I feel like it's easier to make a mistake if we use the proposed order because assertEvent<'FOO'> will use the EventObject since partial inference is not supported by TS. I understand that the optional type param can't be used first but I think it should be OK to just drop that default here.

Done

.changeset/ten-buckets-exist.md
VanTanev: Expanded the changeset docs

VanTanev:
Closing in favor of https://github.com/vantanev/xstate-helpers

PR #2125, opened 2021-04-21
Improve the type of callback invoke params

VanTanev:
```typescript
type Event = { type: 'START' } | { type: 'STOP' }

const machine = createMachine<{}, Event>({
  invoke: {
    id: 'callback-service',
    src: () => (callback, onReceive) => {
      onReceive(event => {
        // event is { type: 'START' } | { type: 'STOP' }
      })
      callback({ type: 'I_DONT_EXIST' }) // type errors
    }
  }
})
```

### DIFF
--- a/.changeset/stupid-windows-sniff.md
+++ b/.changeset/stupid-windows-sniff.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+In callback invokes, the types of `callback` and `onReceive` are properly scoped to the machine TEvent.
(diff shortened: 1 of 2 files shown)


VanTanev:
@Andarist @mattpocock this should be a pretty straightforward type improvement. It didn't break any of my projects, but it would be nice if you could test with yours too.

davidkpiano (approved):
This makes sense to me. I think an initial motivation for having it as `any` was that in some situations, it's not easy for the child to know what event types it could send the parent. But there's ways around that.

mattpocock:
@davidkpiano I 100% agree with @VanTanev that this should be typed as a `Sender<TEvent>`. I really like using the invoked callback pattern but having to manually type this is not particularly fun. I also haven't found a case where I want it typed as anything else.

PR #2143, opened 2021-04-24
[core] Tagged states

davidkpiano:
This PR introduces the `tags` property for state nodes. The tags for states in the state configuration are merged into a `state.tags: Set<string>` property:

```js
  initial: 'green',
  states: {
    green: {
      tags: ['go']
    },
    yellow: {
      tags: ['go'],
      on: {
        TIMER: 'red'
      }
    },
    red: {
      tags: ['stop']
    }
  }
});

machine.initialState.tags.has('go'); // true
```

cc. @mogsie


with-heart (approved):
This is so cool! Such a small change to add such potentially useful functionality.

mogsie (approved):
I approve of this. 😁

packages/core/src/State.ts
```diff
@@ -127,6 +127,7 @@ export class State<
    * An object mapping actor IDs to spawned actors/invoked services.
    */
   public children: Record<string, ActorRef<any>>;
+  public tags: Set<string>;
```
Andarist, line 130: `toJSON` should convert this to an array and an accompanying `.from` method (I think it's the one - but I might be wrong) should convert an array to a Set

packages/core/src/StateNode.ts
```diff
@@ -278,6 +278,7 @@ class StateNode<
   };
 
   private idMap: Record<string, StateNode<TContext, any, TEvent>> = {};
+  public tags?: string[];
```
Andarist, line 281: Usually, we accept `SingleOrArray<T>` for this kind of stuff - should we keep this consistent for this?

packages/core/src/StateNode.ts
```diff
@@ -1322,6 +1324,10 @@ class StateNode<
     // Preserve original history after raised events
     maybeNextState.history = history;
 
+    maybeNextState.tags = new Set(
+      flatten(maybeNextState.configuration.map((sn) => sn.tags ?? []))
```
Andarist, line 1328: I think that it would be preferable to normalize `tags` to an empty array in the constructor, so we wouldn't have to deal with this here.

packages/core/test/tags.test.ts
```diff
@@ -0,0 +1,54 @@
+import { createMachine } from '../src';
+
+describe('tags', () => {
```
Andarist, line 3: we need a test like this:
1. two parallel regions with the same tag
2. exit one of them
3. the expected result is that we should still have that tag available to us

packages/core/test/tags.test.ts
```diff
@@ -0,0 +1,54 @@
...
+            },
+            wait: {
+              tags: ['crosswalkLight']
+            }
+          }
+        }
+      }
+    });
+
+    expect(machine.initialState.tags.has('go')).toBeFalsy();
```
Andarist, line 50: I'm wondering if we shouldn't provide a specialized `.has` on the state itself so we could support "typed tags" in the future. See my comment here: https://github.com/davidkpiano/xstate/discussions/2124#discussioncomment-644216 . The idea is pretty much the same as with typestates.

packages/core/src/State.ts
davidkpiano: Done and test added

packages/core/src/StateNode.ts
davidkpiano: Yeah, that's a good idea

davidkpiano: Done + test added

packages/core/test/tags.test.ts
davidkpiano: I think that's a good idea; let's save it for a separate PR

packages/core/test/tags.test.ts
davidkpiano: Done

packages/core/test/tags.test.ts
Andarist: I agree that we don't have to introduce "typed tags" right from the start but if that's a direction that we are going to then we could already introduce the `has` method to `State` (it would basically just call `this.tags.has(tag)` and nothing more than that.

Why introducing it already? To keep the documented interface more future-proof - using the underlying set directly limits our possibilities and when we finally introduce `has` then people will have to refactor their existing code to switch to it.

Andarist (approved):

.changeset/cyan-dolphins-change.md
```diff
@@ -0,0 +1,29 @@
...
+      tags: 'go'
+    },
+    red: {
+      tags: ['stop', 'other'] // multiple tags
+    }
+  }
+});
+```
+
+You can query whether a state has a tag via `state.hasTag(tag)`:
```
tivac, line 24: Should this also mention `state.tags` or is that not something you want advertised?

davidkpiano: Not right now, I think that'll be an implementation detail for now.

packages/core/src/StateNode.ts
davidkpiano: Done

.changeset/cyan-dolphins-change.md
```diff
@@ -0,0 +1,29 @@
...
+
+```js
+const canGo = state.hasTag('go');
+// => `true` if in 'green' or 'red' state
```
singuerinc, line 28: Hi!, maybe I don't understand how tags works, but shouldn't this example be:

```diff
-// => `true` if in 'green' or 'red' state
+// => `true` if in 'green' or 'yellow' state
```

davidkpiano: Oops, just updated it!

PR #2148, opened 2021-04-26
add simple failing test for parallel machine state value

ahdinosaur:
hi,

first of all: thanks for the work put into `xstate`, i really appreciate the concept and implementation. :purple_heart: 

i'm starting to use `xstate` in a project to replace other state management (redux or react hooks), but am currently stuck on a failing TypeScript type mismatch for my parallel machine's state value.

i'd expect (and at runtime this is the case) that the state value is a nested object with keys per parallel state, however the TypeScript compiler expects that the state value is a single string.

i tried to add a failing test to exemplify my current situation, however it's very possible i'm doing something wrong, so am very happy for any feedback. :smiley_cat: 

image

cheers! :seedling:

### DIFF
--- a/packages/core/test/types.test.ts
+++ b/packages/core/test/types.test.ts
@@ -139,6 +139,12 @@ describe('Parallel StateSchema', () => {
   it('should work with a parallel StateSchema defined', () => {
     expect(true).toBeTruthy();
   });
+
+  it('should infer correct state value', () => {
+    const service = interpret(parallelMachine);
+    const startedService = service.start();
+    expect(startedService.state.value.baz).toEqual('one');
+  });
 });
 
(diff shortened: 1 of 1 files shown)


davidkpiano:
Thanks for the PR! This is actually a lot tricker than it might seem, because there can be many possible state values, and you can't just take the union of all of them (some might be strings, some might be objects that are incompatible with other objects, etc.) and it gets really messy, if not impossible to discern (right now).

Type-wise, you should use `state.matches(...)` to assert the correct value:

```ts
  it('should infer correct state value', () => {
    const service = interpret(parallelMachine);
    const startedService = service.start();

    if (startedService.state.matches({ baz: {} })) {
      expect(startedService.state.value.baz).toEqual('one');
    }
  });
```

PR #2149, opened 2021-04-26
[core] Ensure root node gets included in configuration for entry/exit actions

### LINKED ISSUE #2147: Sending events from invoked callback service in machine root causes service to be invoked again
**Description**
Sending an event from a callback service invoked in the root of the machine, appears to cause the service to be invoked again.

Given
```ts
const m = Machine(
  {
    invoke: {
      src: () => (callback) => {
        const interval = setTimeout(() => callback({ type: "UPDATE" }), 1000);

        return () => clearInterval(interval);
      }
    },
    on: {
      UPDATE: {
        actions: "log"
      }
    }
  },
  {
    actions: {
      log: (ctx, e) => console.log(e)
    }
  }
);

const service = interpret(m);
service.start();
```

**Expected Result**
Given the above machine, I would expect the `UPDATE` event to be dispatched and logged to the console *once* after 1000ms.

**Actual Result**
Given the above machine, the `UPDATE` event is dispatched and logged console infinitely, with 1000ms delays.

**Reproduction**
[CodeSandbox](https://codesandbox.io/s/elated-noether-qhqyz?file=/src/index.ts)

**Additional context**
Invoking the service inside a "proper" state works as expected.
Tested with XState@4.18.0.

First time submitting an issue, feel free to let me know if I messed anything up.

davidkpiano:
Fixes #2147

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -637,10 +637,11 @@ class StateNode<
    */
   public getStateNodes(
     state: StateValue | State<TContext, TEvent, any, TTypestate>
-  ): Array<StateNode<TContext, any, TEvent, any>> {
+  ): Array<StateNode<TContext, any, TEvent, TTypestate>> {
     if (!state) {
       return [];
     }
+
     const stateValue =
       state instanceof State
         ? state.value
@@ -651,13 +652,15 @@ class StateNode<
 
       return initialStateValue !== undefined
         ? this.getStateNodes({ [stateValue]: initialStateValue } as StateValue)
-        : [this.states[stateValue]];
+        : [this, this.states[stateValue]];
     }
 
     const subStateKeys = keys(stateValue);
     const subStateNodes: Array<
-      StateNode<TContext, any, TEvent>
+      StateNode<TContext, any, TEvent, TTypestate>
     > = subStateKeys.map((subStateKey) => this.getStateNode(subStateKey));
+
+    subStateNodes.push(this);
 
     return subStateNodes.concat(
       subStateKeys.reduce((allSubStateNodes, subStateKey) => {
--- a/packages/core/test/invoke.test.ts
+++ b/packages/core/test/invoke.test.ts
@@ -836,6 +836,50 @@ describe('invoke', () => {
         })
         .start();
     });
+
+    it('should not reinvoke root-level invocations', (done) => {
+      // https://github.com/davidkpiano/xstate/issues/2147
+
+      let invokeCount = 0;
+      let invokeDisposeCount = 0;
+      let actionsCount = 0;
+
+      const machine = createMachine({
+        invoke: {
+          src: () => () => {
+            invokeCount++;
+
+            return () => {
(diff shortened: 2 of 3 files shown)


packages/core/src/StateNode.ts
```diff
@@ -651,14 +652,16 @@ class StateNode<
...
+        : [this, this.states[stateValue]];
     }
 
     const subStateKeys = keys(stateValue);
     const subStateNodes: Array<
-      StateNode<TContext, any, TEvent>
+      StateNode<TContext, any, TEvent, TTypestate>
     > = subStateKeys.map((subStateKey) => this.getStateNode(subStateKey));
 
+    subStateNodes.push(this);
```
davidkpiano, line 663: This is probably already fixed in v5 since the algorithm is different.

packages/core/test/invoke.test.ts
```diff
@@ -836,6 +836,50 @@ describe('invoke', () => {
         })
         .start();
     });
+
+    it('should not reinvoke root-level invocations', (done) => {
```
Andarist, line 840: a similar test could be written for entry actions - they are handled almost the same internally, but externally they look very different

it would be also great if the test title would include some more detail about this specific test because as it is it's a little bit vague

davidkpiano:
@Andarist Ready for final review

.changeset/nasty-trainers-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Invocations and entry actions for _combinatorial_ machines (machines with only a single root state) now behave predictably and will not re-execute upon transitions.
```
Andarist, line 5:
```suggestion
Invocations and entry actions for _combinatorial_ machines (machines with only a single root state) now behave predictably and will not re-execute upon targetless transitions.
```

Andarist:
Observation: before investigating #2176 I had a mental model of us adding an "implied" root node but I've realized here that it cannot be seen as such without overcomplicating abstract logic for final states. What we do (in contrast to SCXML) is that we allow `<scxml>` to work more closely to `<state>`/`<parallel>` - by allowing, for example, transitions to be defined for it. If we plan to support XState -> SCXML conversion then this should be accounted for as wrapping the whole thing in that extra "implied" root node would just not cut it.

Question: what should be the value of a combinatorial machine?

davidkpiano:
> Question: what should be the value of a combinatorial machine?

Just an object: `{}`

Andarist:
Isn't this looking slighty weird?
```js
// value: {}
createmachine({})

// value: 'a'
createMachine({
  initial: 'a',
  states: {
    a: {}
  }
})

// value: { a: 'b' }
createMachine({
  initial: 'a',
  states: {
    a: {
      initial: 'b',
      states: {
        b: {}
      }
    }
  }
})
```

davidkpiano:
> Do you think the second one should be: `{ a: {} }`?

Maybe in version 5 🤔 that's not a bad idea.

PR #2173, opened 2021-05-04
Fixed an issue with tags not being set correctly after sending an event to a machine that didn't result in selecting any transitions

Andarist:
Fixes https://github.com/davidkpiano/xstate/issues/2172

When we don't select any transitions then we just return here:
https://github.com/davidkpiano/xstate/blob/64ab1150e0a383202f4af1d586b28e081009c929/packages/core/src/StateNode.ts#L1280-L1282
but tags are set later on, here:
https://github.com/davidkpiano/xstate/blob/64ab1150e0a383202f4af1d586b28e081009c929/packages/core/src/StateNode.ts#L1328-L1330

So this didn't have a chance to be set correctly for the fixed test case.

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -1263,7 +1263,8 @@ class StateNode<
       configuration: resolvedConfiguration,
       transitions: stateTransition.transitions,
       children,
-      done: isDone
+      done: isDone,
+      tags: currentState?.tags
     });
 
     const didUpdateContext = currentContext !== updatedContext;
--- a/.changeset/soft-cats-trade.md
+++ b/.changeset/soft-cats-trade.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/core/src/StateNode.ts
```diff
@@ -1263,7 +1263,8 @@ class StateNode<
       configuration: resolvedConfiguration,
       transitions: stateTransition.transitions,
       children,
-      done: isDone
+      done: isDone,
+      tags: currentState?.tags
```
Andarist, line 1267: this is kinda funky since we'll override this later on when we match some transition and this is outdated at the time of creating this `nextState` but it was an easy-to-do thing now - this happens in our internals, so hopefully this shouldn't be that observable.

We might call some expressions with that outdated info when processing the micro queue but that might be partially expected (this part already works in a different way than in SCXML since it updates the configuration as it goes and we don't do that).

davidkpiano (approved):

PR #2184, opened 2021-05-09
[core] Add test for "finalized" parallel states

davidkpiano:


packages/core/test/parallel.test.ts
```diff
@@ -922,4 +922,69 @@ describe('parallel states', () => {
       expect(run2.context.log.length).toBe(2);
     });
   });
+
+  it('should raise a "done.state.*" event when all child states reach final state', (done) => {
```
Andarist, line 926: This was already sort-of tested here: https://github.com/davidkpiano/xstate/blob/430da6ec5744f5508f2b35ab556492bb79a632f8/packages/core/test/final.test.ts#L69 . More tests the better though :)

I would probably reword the title of this test - since `done.state.*` doesn't appear anywhere in it and might be confusing for people not familiar with SCXML and our internals. It's not a big deal though so I won't insist on it. Additional consideration might be that `done.state.*` is actually a regular event and `onDone` is just a sugar syntax for it - so we could add a test that would transition explicitly based on this event.

Andarist (approved):

PR #2187, opened 2021-05-10
[docs] Run Grammarly on docs

VanTanev:


VanTanev:
@davidkpiano I incorporated the 2 comments from the previous PR.

docs/guides/actors.md
```diff
@@ -196,12 +196,12 @@ const fetchData = (query) => {
 ```
 
 ::: warning
-It is not recommended to spawn promise actors, as [invoking promises](./communication.md#invoking-promises) is a better pattern for this, since they are dependent on state (self-cancelling) and have more predictable behavior.
+It is not recommended to spawn promise actors, as [invoking promises](./communication.md#invoking-promises) is a better pattern for this since they are dependent on the state (self-canceling) and have more predictable behavior.
```
with-heart, line 199:
```suggestion
It is not recommended to spawn promise actors, as [invoking promises](./communication.md#invoking-promises) is a better pattern for this since they are dependent on state (self-canceling) and have more predictable behavior.
```

I think "the" is not a good change here. **The** state? Which state is that?

docs/guides/communication.md
```diff
@@ -11,7 +11,7 @@ You can invoke:
 - [Promises](#invoking-promises), which will take the `onDone` transition on `resolve`, or the `onError` transition on `reject`
 - [Callbacks](#invoking-callbacks), which can send events to and receive events from the parent machine
 - [Observables](#invoking-observables), which can send events to the parent machine, as well as a signal when it is completed
-- [Machines](#invoking-machines), which can also send/receive events, and also notify the parent machine when it reaches its [final state](./final.md)
+- [Machines](#invoking-machines), which can also send/receive events and also notify the parent machine when it reaches its [final state](./final.md)
```
with-heart, line 14:
```suggestion
- [Machines](#invoking-machines), which can send/receive events and also notify the parent machine when it reaches its [final state](./final.md)
```

docs/guides/effects.md
```diff
@@ -2,7 +2,7 @@
 
 In statecharts, "side-effects" can be grouped into two categories:
 
-**"Fire-and-forget" effects**, which execute a synchronous side-effect with no events sent back to the statechart, or _send an event synchronously_ back to the statechart:
+**"Fire-and-forget" effects**, which execute asynchronous side-effect with no events sent back to the statechart, or _send an event synchronously_ back to the statechart:
```
with-heart, line 5:
```suggestion
**"Fire-and-forget" effects**, which execute a synchronous side-effect with no events sent back to the statechart, or _send an event synchronously_ back to the statechart:
```

This would change the meaning of the sentence.

docs/guides/events.md
```diff
@@ -1,6 +1,6 @@
 # Events
 
-An event is what causes a state machine to [transition](./transitions.md) from its current [state](./states.md) to its next state. All state transitions in a state machine are due to these events; state cannot change unless some stimulus (the event) causes it to change.
+An event is what causes a state machine to [transition](./transitions.md) from its current [state](./states.md) to its next state. All state transitions in a state machine are due to these events; a state cannot change unless some stimulus (the event) causes it to change.
```
with-heart, line 3:
```suggestion
An event is what causes a state machine to [transition](./transitions.md) from its current [state](./states.md) to its next state. All state transitions in a state machine are due to these events; state cannot change unless some stimulus (the event) causes it to change.
```

This makes more sense in the original form imo

docs/guides/guards.md
```diff
@@ -319,7 +319,7 @@ const lightMachine = createMachine({
 When an `in`-state guard is present with other `cond` guards in the same transition, _all_ guards must evaluate to `true` for the transition to be taken.
 
 ::: tip
-Using "in state" guards is usually a sign that the machine can be refactored in a way that makes their usage unnecessary. Prefer avoiding "in state" guards when possible.
+Using "`in`-state" guards is usually a sign that the machine can be refactored in a way that makes their usage unnecessary. Prefer avoiding "`in`-state" guards when possible.
```
with-heart, line 322:
```suggestion
Using `in` state guards is usually a sign that the machine can be refactored in a way that makes their usage unnecessary. Avoid `in` state guards when possible.
```

docs/guides/ids.md
```diff
@@ -185,7 +185,7 @@ IDs correspond to the definition of IDs in the SCXML spec:
 
 ## Quick Reference
 
-**Default, automatically generated ID:**
+**Default ID, automatically generated**
```
with-heart, line 188: This feels wrong 🤷

docs/guides/actions.md
```diff
@@ -103,7 +103,7 @@ It depends! They mean different things:
 // ...
 ```
 
-- A transition actions mean "execute this action **only on this transition**". Use transition actions when the action is dependent on the event and the state node that it is currently in.
+- A transition actions mean "execute this action **only on this transition**." Use transition actions when the action is dependent on the event and the state node that it is currently in.
```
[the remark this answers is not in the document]
davidkpiano: Yeah, even though I know it's correct to put the punctuation on the inside, we'll likely run into counterexamples... I would disable this rule.

docs/guides/events.md
VanTanev: I prefer the version with "a," but am not a native speaker, so 🤷

davidkpiano:

PR #2191, opened 2021-05-12
[v5] Remove `StateSchema`

davidkpiano:
This PR removes `StateSchema` from all types in v5. We'll figure out a better way to have strong typing of state keys, but this is the first step.


davidkpiano:
@Andarist The only tests failing here are the ones around JSX, which I'm ignoring for now.

Andarist (approved):
Apart from the missing changeset it's all good and... hooray for the change 🥳

PR #2195, opened 2021-05-12
[docs] Replace createMachineConfig with MachineConfig

SergeAstapov:
Does not seem like `createMachineConfig` exists anywhere


davidkpiano:
Whoops! Find/replace typo 🙈

PR #2196, opened 2021-05-13
Update resources.md

justinphilpott:
Adding in the excellent xstate-catalogue.com website


davidkpiano:
Good call – I think it's better if we add it as a top-level link. I can go ahead and do that (also adding examples directly to the repo)

justinphilpott:
Great, it's an awesome resource for an awesome lib!

Sent from ProtonMail mobile

-------- Original Message --------

davidkpiano:
This has been added, thanks!

PR #2198, opened 2021-05-15
[examples] Fetch example

davidkpiano:
This PR adds a basic `fetch` example


examples/fetch/README.md
```diff
@@ -0,0 +1,9 @@
+# Fetch example
+
+[Open in CodeSandbox](https://githubbox.com/davidkpiano/xstate/tree/master/examples/fetch)
```
Andarist, line 3: would be cool to make this a nice button in the future, might be worth checking if CodeSandbox already provides a nice image for this

PR #2200, opened 2021-05-16
[examples] Update fetch example

davidkpiano:
This PR updates the fetch example.


examples/fetch/src/index.ts
```diff
@@ -8,7 +8,31 @@ const fetchMachine = createMachine({
...
+        src: (context, event) =>
+          fetch('https://dog.ceo/api/breeds/image/random').then((response) =>
+            response.json()
+          ),
+        onDone: {
+          target: 'success'
+        },
+        onError: {
+          target: 'failure'
+        }
```
Andarist, line 23: Other transitions here are using shortcut transition configs, so maybe good to unify those with that style?
```suggestion
        onDone: 'success',
        onError: 'failure'
```

davidkpiano: I'm thinking about saving it to `context`, so we'll keep it as an object

Andarist (approved):

PR #2211, opened 2021-05-18
[v5] Separate StateMachine and StateNode

davidkpiano:
This PR adds `machine.root` to the `StateMachine` class, which represents the root `StateNode` for the machine. Previously, the `StateMachine` (previously named `MachineNode`) _extended_ `StateNode`, but ideally these should be separate.

CleanShot 2021-05-18 at 15 23 54@2x

### DIFF
--- a/packages/core/src/MachineNode.ts
+++ b/packages/core/src/MachineNode.ts
@@ -17,19 +17,13 @@ import { IS_PRODUCTION } from './environment';
 import { STATE_DELIMITER } from './constants';
 import {
   getConfiguration,
-  getChildren,
   getAllStateNodes,
   resolveMicroTransition,
   macrostep,
   toState,
   isStateId
 } from './stateUtils';
-import {
-  getStateNodeById,
-  getStateNodes,
-  transitionNode,
-  resolveStateValue
-} from './stateUtils';
+import { getStateNodes, transitionNode, resolveStateValue } from './stateUtils';
 import { StateNode } from './StateNode';
 
 export const NULL_EVENT = '';
--- a/packages/core/src/MachineNode.ts
+++ b/packages/core/src/MachineNode.ts
@@ -169,7 +169,7 @@ export class MachineNode<
    *
    * @param state The state to resolve
    */
-  public resolveState(state: State<TContext, TEvent>): State<TContext, TEvent> {
+  public resolveState(state: State<TContext, TEvent, any>): typeof state {
     const configuration = Array.from(
       getConfiguration(getStateNodes(this.root, state.value))
     );
--- a/packages/core/src/MachineNode.ts
+++ b/packages/core/src/MachineNode.ts
@@ -1,4 +1,4 @@
-import { toArray, isBuiltInEvent, toSCXMLEvent } from './utils';
+import { isBuiltInEvent, toSCXMLEvent } from './utils';
 import {
   Event,
   StateValue,
@@ -8,11 +8,11 @@ import {
   SCXML,
   Typestate,
   Transitions,
-  MachineSchema
+  MachineSchema,
+  StateNodeDefinition
 } from './types';
 import { State } from './State';
 
-import { toActionObject } from './actions';
 import { IS_PRODUCTION } from './environment';
 import { STATE_DELIMITER } from './constants';
 import {
@@ -21,7 +21,8 @@ import {
   getAllStateNodes,
   resolveMicroTransition,
   macrostep,
-  toState
+  toState,
+  isStateId
 } from './stateUtils';
 import {
   getStateNodeById,
@@ -63,7 +64,7 @@ export class MachineNode<
   TContext = any,
   TEvent extends EventObject = EventObject,
   TTypestate extends Typestate<TContext> = any
-> extends StateNode<TContext, TEvent> {
+> {
   public context: TContext;
   /**
    * The machine's own version.
@@ -84,6 +85,15 @@ export class MachineNode<
 
   public __xstatenode: true = true;
 
+  public idMap: Map<string, StateNode<TContext, TEvent>> = new Map();
+
+  public root: StateNode<TContext, TEvent>;
+
+  public key: string;
+
+  public states: StateNode<TContext, TEvent>['states'];
+  public events: Array<TEvent['type']>;
+
   constructor(
     /**
      * The raw config used to create the machine.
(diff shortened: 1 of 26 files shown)


packages/core/src/MachineNode.ts
```diff
@@ -63,7 +58,7 @@ export class MachineNode<
   TContext = any,
   TEvent extends EventObject = EventObject,
   TTypestate extends Typestate<TContext> = any
-> extends StateNode<TContext, TEvent> {
```
Andarist, line 66: so maybe it's not a node in the end and we should rename `MachineNode` to a `Machine`?

packages/core/src/MachineNode.ts
```diff
@@ -187,13 +169,13 @@ export class MachineNode<
    *
    * @param state The state to resolve
    */
-  public resolveState(state: State<TContext, TEvent>): State<TContext, TEvent> {
+  public resolveState(state: State<TContext, TEvent, any>): typeof state {
```
Andarist, line 172: using `typeof` to type the return value - nice 👍 I don't think I've ever used it like this and it's quite nice

packages/core/src/MachineNode.ts
```diff
@@ -273,7 +255,25 @@ export class MachineNode<
     >;
   }
 
-  public getStateNodeById(id: string): StateNode<TContext, TEvent> {
-    return getStateNodeById(this, id);
+  public getStateNodeById(stateId: string): StateNode<TContext, TEvent> {
```
Andarist, line 258: does this new implementation of this method result in a different behavior somehow?

packages/core/src/StateNode.ts
```diff
@@ -178,17 +176,17 @@ export class StateNode<
         ? 'history'
         : 'atomic');
 
+    this.order = this.machine.idMap.size;
```
Andarist, line 179: quite an improvement over `dfs` algorithm 👍

packages/core/src/stateUtils.ts
```diff
@@ -1816,24 +1814,22 @@ export function resolveStateValue<TContext, TEvent extends EventObject>(
...
 }
 
-export function toState<TContext, TEvent extends EventObject>(
-  state: StateValue | State<TContext, TEvent>,
-  machine: MachineNode<TContext, TEvent>
-): State<TContext, TEvent> {
+export function toState<TMachine extends MachineNode<any, any, any>>(
+  state: StateValue | TMachine['initialState'],
+  machine: TMachine
+): TMachine['initialState'] {
```
Andarist, line 1824:
```suggestion
): StateFromMachine<TMachine> {
```

packages/core/test/fixtures/id.ts
```diff
@@ -42,28 +42,6 @@ export const machine = createMachine({
           id: 'B.dot'
         }
       }
-    },
-    getter: {
```
Andarist, line 46: was the support for getters dropped anywhere in this PR?

packages/xstate-graph/src/graph.ts
```diff
@@ -342,7 +342,12 @@ export function getSimplePathsAsArray<
   return keys(result).map((key) => result[key]);
 }
 
-export function toDirectedGraph(stateNode: StateNode): DirectedGraphNode {
+export function toDirectedGraph(
+  machineNode: MachineNode | StateNode
+): DirectedGraphNode {
+  const stateNode =
+    machineNode instanceof StateNode ? machineNode : machineNode.root;
```
Andarist, line 349: In general, I like to avoid relying on `instanceof` checks for any code that escapes a boundary of a single package - this would fail with 2 copies of XState in `node_modules`.

I don't think that's something that should be fixed/handled as part of this PR though but perhaps, we should add a TODO comment here?

packages/core/src/MachineNode.ts
davidkpiano: Or `StateMachine` to prevent confusion (since v4 used `Machine` for something slightly different)

packages/core/src/MachineNode.ts
davidkpiano: It does not. `getStateNodeById` should only be used for the entire machine anyway.

packages/core/src/StateNode.ts
davidkpiano: Yeah, that was a bit ugly 🙈

packages/core/test/fixtures/id.ts
davidkpiano: It was dropped in a previous PR. Forgot to remove this.

packages/xstate-graph/src/graph.ts
davidkpiano:
```suggestion
    machineNode instanceof StateNode ? machineNode : machineNode.root; // TODO: accept only machines
```

packages/core/src/MachineNode.ts
Andarist: That works too 👍

packages/core/src/MachineNode.ts
Andarist: Hm, I feel like I'm missing something - there is still a standalone `getStateNodeById` util that is used in some places in the code. If this method here is the equivalent of that util (which was used here before this PR) then shouldn't that util be removed and calls to it should be replaced with calls to this method?

davidkpiano: Removed in ba8ea1a

packages/core/src/MachineNode.ts
davidkpiano: Done in 9f554cd

with-heart:
@davidkpiano Could you expand a bit on the _why_? What makes the separation of `MachineNode` and `StateNode` ideal?

Andarist:
@with-heart in this instance we might even want to rename `MachineNode` to `StateMachine`. It becomes less concerned about being a particular state (the root state) and more about handling things like top-level resolving of stuff, transitions, lookups, and more. In the past everything was put on a `StateNode` so deep state nodes had access to those top-level concerns while they never have used any of that stuff - so this PR is yet another step of cleaning this up.

Andarist (approved):

Andarist:
I don't think there is much to document here publicly in a changeset but if you feel otherwise - feel free to add a changelog entry for this.

davidkpiano:
> @davidkpiano Could you expand a bit on the _why_? What makes the separation of `MachineNode` and `StateNode` ideal?

Additionally, this allows someone to make something like `SCXMLStateMachine` which might have multiple `StateNode` instances on the root instead of the default one. In general, it's a better abstraction/separation of responsibilities.

with-heart:
Thanks yall for the explanations!

PR #2223, opened 2021-05-20
Consider how TEmitted and getSnapshot() return values relate

davidkpiano:
https://github.com/davidkpiano/xstate/pull/2223#discussion_r637368750

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1275,10 +1275,13 @@ export interface ActorRef<TEvent extends EventObject, TEmitted = any>
   send: Sender<TEvent>;
 }
 
-export interface SpawnedActorRef<TEvent extends EventObject, TEmitted = any>
-  extends ActorRef<TEvent, TEmitted> {
+export interface SpawnedActorRef<
+  TEvent extends EventObject,
+  TEmitted = any,
+  TSnapshot = TEmitted | undefined
+> extends ActorRef<TEvent, TEmitted> {
   id: string;
-  getSnapshot: () => TEmitted | undefined;
+  getSnapshot: () => TSnapshot;
   stop?: () => void;
   toJSON?: () => any;
 }
@@ -1286,7 +1289,11 @@ export interface SpawnedActorRef<TEvent extends EventObject, TEmitted = any>
 export type ActorRefFrom<
   T extends StateMachine<any, any, any> | Promise<any>
 > = T extends StateMachine<infer TContext, any, infer TEvent, infer TTypestate>
-  ? SpawnedActorRef<TEvent, State<TContext, TEvent, any, TTypestate>> & {
+  ? SpawnedActorRef<
+      TEvent,
+      State<TContext, TEvent, any, TTypestate>,
+      State<TContext, TEvent, any, TTypestate>
+    > & {
       state: State<TContext, TEvent, any, TTypestate>;
     }
   : T extends Promise<infer U>
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1294,6 +1294,9 @@ export type ActorRefFrom<
       State<TContext, TEvent, any, TTypestate>,
       State<TContext, TEvent, any, TTypestate>
     > & {
+      /**
+       * @deprecated
+       */
       state: State<TContext, TEvent, any, TTypestate>;
     }
   : T extends Promise<infer U>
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -1275,18 +1275,25 @@ export interface ActorRef<TEvent extends EventObject, TEmitted = any>
   send: Sender<TEvent>;
 }
 
-export interface SpawnedActorRef<TEvent extends EventObject, TEmitted = any>
-  extends ActorRef<TEvent, TEmitted> {
+export interface SpawnedActorRef<
```
Andarist, line 1278: writing down somewhere what is considered to be an Actor, ActorRef and SpawnedActorRef would be very helpful - I often have to recheck this on my own in the codebase as it's not entirely intuitive for me (especially the distinction between `ActorRef` and `SpawnedActorRef`

packages/core/src/types.ts
```diff
@@ -1275,18 +1275,25 @@ export interface ActorRef<TEvent extends EventObject, TEmitted = any>
...
+  TEvent extends EventObject,
+  TEmitted = any,
```
Andarist, line 1280: What is `TEmitted` conceptually? I thought it was kinda a snapshot - are those output events? If yes - shouldn't this extend `EventObject`?

packages/core/src/types.ts
```diff
@@ -1275,18 +1275,25 @@ export interface ActorRef<TEvent extends EventObject, TEmitted = any>
...
 export type ActorRefFrom<
   T extends StateMachine<any, any, any> | Promise<any>
 > = T extends StateMachine<infer TContext, any, infer TEvent, infer TTypestate>
-  ? SpawnedActorRef<TEvent, State<TContext, TEvent, any, TTypestate>> & {
+  ? SpawnedActorRef<
+      TEvent,
+      State<TContext, TEvent, any, TTypestate>,
+      State<TContext, TEvent, any, TTypestate>
+    > & {
       state: State<TContext, TEvent, any, TTypestate>;
```
Andarist, line 1297: should maybe this property be deprecated in favour of `getSnapshot`?

packages/core/src/types.ts
davidkpiano: `TEmitted` represents whatever is emitted to observers in `actor.subscribe(...)`. It doesn't necessarily extend `EventObject` since e.g., a Promise can emit non-event values.

packages/core/src/types.ts
davidkpiano: Yes

packages/core/src/types.ts
Andarist: Right, but for promises that value that is emitted is the snapshot so I wonder in what situations the emitted value and snapshot would be different.

davidkpiano: The snapshot is the emitted value _at any point in time_. For observables and promises, that may be `undefined` if nothing was emitted yet. For machines, that is always defined (`State`) since the initial state is known immediately. That's why `TEmitted` and `TSnapshot` can be different.

davidkpiano: Removed the 3rd generic param in 1321351

packages/core/src/types.ts
with-heart: Could write that down right here to start 😁 Pretty please!

packages/core/src/types.ts
Andarist: > That's why TEmitted and TSnapshot can be different.

That would mean that `getSnapshot` would not simply return `TSnapshot` for promises etc - since if your snapshot is a `number` then `getSnapshot` would have to return `number | undefined` (resolved from `TSnapshot | undefined`). That was my main concern about this.

Since you have unified them now - we could unify the naming further in the future and rename most `TEmitted` to `TSnapshot`, right? This would be mostly an internal change.

Andarist (approved):

packages/core/src/types.ts
davidkpiano: > > That's why TEmitted and TSnapshot can be different.
> 
> That would mean that `getSnapshot` would not simply return `TSnapshot` for promises etc - since if your snapshot is a `number` then `getSnapshot` would have to return `number | undefined` (resolved from `TSnapshot | undefined`). That was my main concern about this.
> 
> Since you have unified them now - we could unify the naming further in the future and rename most `TEmitted` to `TSnapshot`, right? This would be mostly an internal change.

Added here: https://github.com/davidkpiano/xstate/projects/1#card-61626771

packages/core/src/types.ts
davidkpiano: Added as todo here: https://github.com/davidkpiano/xstate/projects/1#card-61626855

PR #2227, opened 2021-05-21
Fixed build of the core by enforcing compatible tslib version

Andarist:
This should fix the problem that has manifested here: https://github.com/davidkpiano/xstate/runs/2600705137

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,10 +54,14 @@
     "prettier": "^2.0.2",
     "spawn-command": "0.0.2-1",
     "ts-jest": "^26.5.6",
+    "tslib": "^2.2.0",
     "tslint": "^5.11.0",
     "typedoc": "^0.17.4",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -54,11 +54,15 @@
...
     "spawn-command": "0.0.2-1",
     "ts-jest": "^26.5.6",
+    "tslib": "^2.2.0",
     "tslint": "^5.11.0",
     "typedoc": "^0.17.4",
     "typescript": "^4.2.4",
     "webpack-dev-middleware": "^3.6.0"
   },
+  "resolutions": {
+    "**/tslib": "^2.2.0"
```
Andarist, line 64: this ain't ideal but I couldn't make a more granular resolution (`rollup-plugin-typescript2/tslib`) to work, even though it should work according to the RFC for this feature

PR #2230, opened 2021-05-21
[@xstate/vue] Add useSelector composable

santicros:
This PR adds `useSelector` hook to `@xstate/vue`.
Added tests & docs 💖

### DIFF
--- a/.changeset/nasty-mirrors-hope.md
+++ b/.changeset/nasty-mirrors-hope.md
@@ -1,0 +1,18 @@
+---
+'@xstate/vue': minor
+---
+
+Added new composable `useSelector(actor, selector)`, which subscribes to actor and returns the selected state derived from selector(snapshot):
+
+```js
+import { useSelector } from '@xstate/vue';
+
+export default {
+  props: ['someActor'],
+  setup(props) {
+    const count = useSelector(props.someActor, (state) => state.context.count);
+    // ...
+    return { count };
+  }
+};
+```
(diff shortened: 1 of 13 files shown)


davidkpiano (approved):

ChrisShank:
Just a thought, not particularly about adding `useSelector`, but discussing how we can potentially automate this composable so that its not need in the future!

If `xstate` could provide us with a list of assigned/changed properties on each transition and `state.value.context` was a [`shallowReactive`](https://v3.vuejs.org/api/basic-reactivity.html#shallowreactive) we should be able to only loop through the `context` properties that recently changed and update them through assignment, meaning `useSelector` *might* not be needed any more! I think we would also have to change `state` to not be a `shallowRef` (maybe it would also have to become a `shallowReactive`? 🤔)

davidkpiano:
> First is there a way that`xstate` could provide us with a list of assigned/changed properties on each transition? Sort of similar to `lit’s` [willUpdate](https://lit.dev/docs/components/lifecycle/) lifecycle hook.

That would be interesting, and I believe it is possible, since the only way `context` can be updated is through `assign(...)` actions (by design) 🤔❕  There's 3 considerations though:

- This would only allow change detection at a single level; i.e., no deep update detection
- `assign(fn)` assumes that _all_ of `context` has been changed, so we'd have to shallow compare for uses of that.
- Change detection is not the same as computed property change detection, and `useSelector(...)` deals with both.

> Even the maybe there would still be cases when `useSelector` would be useful?

I believe that it would still be useful, given the above. For example, if you have a selector that computes `state.context.someProp.length`, it might be unchanged even if the actual `someProp` changes.

PR #2234, opened 2021-05-23
[DOCS] Typescript example are type correct

niallpaterson:
This PR goes some way to updating the typescript documentation. Its main purpose is to make the typescript examples type-correct.

It makes the following changes:

1. The first light machine example has now had its state schema stripped out. It now gives an example of simply providing the TEvent and TContext generic parameters. The reasoning behind removing as opposed to replacing the state type with the newer syntax is that there already exists a section detailing how to type the state (see 2. below), and so removing it at this point avoids repitition.
2. The section at the end which details TypeStates has been moved below the lightMachine example. It seemed a more appropriate place for it. It should now be easy to see how to fully type createMachine(...). Having this section at the end also meant full typing information for createMachine was fragmented throughout the guide. Now all the typing info is side-by-side.
3. I have updated the description in the Config Object section detailing the type relationship between MachineConfig and createMachine(...). It states that they accept the same types. Although this is true for Machine and MachineConfig, it is not true for createMachine and MachineConfig. I've also added an example of how to type the state for MachineConfig (see myStateSchema).
4. I replaced 'key' with 'id' in the lightMachine example since I assume that's what was intended.
5. I tidied up a little bit: there was a trailing comma and a rogue triple colon. Both have been removed.

IMO it would be useful if there was an example of how to type the older Machine syntax. If I have time next week I'll put in a PR for that, but I guess it depends on the extend to which the old syntax is being phased out. 

Also, I wondered whether there are plans for the MachineConfig type to be updated so it is the same as the type for createMachine(...)? That seems to make sense to me. At least for now the typing info in the examples is correct.

Let me know if there's any issues and ill get it fixed ASAP 👍.


davidkpiano (approved):

davidkpiano:
> Also, I wondered whether there are plans for the MachineConfig type to be updated so it is the same as the type for createMachine(...)? That seems to make sense to me. At least for now the typing info in the examples is correct.

You mean if the return type of `createMachine(...)` will be `MachineConfig`? It will not be, since those can be different.

VanTanev:
If you want, wait on merging this until I get my typescript verification script integrated into xstate?

docs/guides/typescript.md
```diff
@@ -253,4 +188,58 @@ type State =
...
+
+The `send` action on the interpreted machine `interpret(stateMachine)` isn't always type safe. To get typechecking for this function signature, use the following pattern:
+
+```ts
+type UserEvents = {
+  type: 'TEST';
+  value: string;
+};
+
+const service = interpret(stateMachine);
```
with-heart, line 230: I know this isn't new, but I think we should note how `UserEvents` is used. It's only clear if you already know how types work with XState.

```suggestion
// type is attached to `stateMachine` (not shown)
type UserEvents = {
  type: 'TEST';
  value: string;
};

// `service` is aware that `stateMachine` events are of type `UserEvents`
const service = interpret(stateMachine);
```

VanTanev: Good point, I'll update the example

VanTanev: Wait, I have. Check out https://github.com/davidkpiano/xstate/pull/2235 that builds on this PR

VanTanev:
https://github.com/davidkpiano/xstate/pull/2235 builds on top of this PR. I suggest we do all further changes there?

with-heart:
@VanTanev That's fair!

docs/guides/typescript.md
```diff
@@ -253,4 +188,58 @@ type State =
...
+
+// This will compile
+service.send({ type: 'TEST', value: 'testvalue' });
+
+// This will have a compile error on the `value` type
+service.send({ type: 'TEST', value: 1 });
+```
+
+If you use the following pattern, you'll lose type safety, so both of these will compile:
```
davidkpiano, line 239: Not sure if this is accurate anymore (if it is, it will soon not be)

VanTanev: In #2235 I verified that it's currently true.

niallpaterson:
> > Also, I wondered whether there are plans for the MachineConfig type to be updated so it is the same as the type for createMachine(...)? That seems to make sense to me. At least for now the typing info in the examples is correct.
> 
> You mean if the return type of `createMachine(...)` will be `MachineConfig`? It will not be, since those can be different.

Soz for the late reply! Looks like @VanTanev has been doing some good work over on #2235  :) 

Apologies for being unclear, I meant the generic types. The docs state that:

The generic types for MachineConfig<TContext, TSchema, TEvent> are the same as those for createMachine<TContext, TSchema, TEvent>.

Which is nice for when you are defining your config external to Machine. You just don't get that anymore with the new typing of createMachine (which is a little less nice).

davidkpiano:

PR #2242, opened 2021-05-27
[v5] Simplify Behavior interface

davidkpiano:
This PR simplifies the `Behavior` interface by consolidating `receive` and `receiveSignal` under the same reducer (as in XActor)

### DIFF
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -145,7 +145,7 @@ export class Actor<TEvent extends EventObject, TEmitted> {
   public current: TEmitted;
   private context: ActorContext;
   private behavior: Behavior<TEvent, TEmitted>;
-  private mailbox: Array<TEvent | LifecycleSignal> = [];
+  private mailbox: Array<TEvent> = [];
   private processingStatus: ProcessingStatus = ProcessingStatus.NotProcessing;
   public name: string;
 
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -145,7 +145,7 @@ export class Actor<TEvent extends EventObject, TEmitted> {
   public current: TEmitted;
   private context: ActorContext;
   private behavior: Behavior<TEvent, TEmitted>;
-  private mailbox: Array<TEvent> = [];
+  private mailbox: Array<TEvent | LifecycleSignal> = [];
   private processingStatus: ProcessingStatus = ProcessingStatus.NotProcessing;
   public name: string;
 
@@ -178,7 +178,7 @@ export class Actor<TEvent extends EventObject, TEmitted> {
   public subscribe(observer) {
     return this.behavior.subscribe?.(observer) || nullSubscription;
   }
-  public receive(event: TEvent) {
+  public receive(event: TEvent | LifecycleSignal) {
     this.mailbox.push(event);
     if (this.processingStatus === ProcessingStatus.NotProcessing) {
       this.flush();
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -142,7 +142,7 @@ export function spawnFrom<TEvent extends EventObject>(
   name?: string
 ): ObservableActorRef<TEvent, undefined>;
 export function spawnFrom(entity: any): ObservableActorRef<any, any> {
-  return spawn(createBehaviorFrom(entity));
+  return spawn(createBehaviorFrom(entity)) as ObservableActorRef<any, any>; // TODO: fix
 }
 
 enum ProcessingStatus {
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -40,8 +40,8 @@ export function isSpawnedActorRef(item: any): item is SpawnedActorRef<any> {
   return isActorRef(item) && 'name' in item;
 }
 
-export function fromObservable<T extends EventObject>(
-  observable: Subscribable<T>,
+export function fromObservable<TEvent extends EventObject>(
+  observable: Subscribable<TEvent>,
   name: string
 ): ActorRef<never> {
   return new ObservableActorRef(
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -177,6 +177,7 @@ export class Actor<TEvent extends EventObject, TEmitted> {
     return this;
   }
   public stop() {
+    this.mailbox.length = 0; // TODO: test this behavior
     this.current = this.behavior.receive(
       this.current,
       stopSignal,
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -10,11 +10,9 @@ import {
 } from './types';
 import { StateMachine } from './StateMachine';
 import { State } from './State';
-import { Interpreter } from './interpreter';
 import {
   Behavior,
   ActorContext,
-  createServiceBehavior,
   createMachineBehavior,
   createDeferredBehavior,
   createPromiseBehavior,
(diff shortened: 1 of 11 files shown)


packages/core/src/Actor.ts
```diff
@@ -185,15 +193,15 @@ export class Actor<TEvent extends EventObject, TEmitted> {
     }
   }
   public receiveSignal(signal: LifecycleSignal) {
```
Andarist, line 195: shouldn't this be unified with `receive` as well?

packages/core/src/Actor.ts
```diff
@@ -169,31 +169,35 @@ export class Actor<TEvent extends EventObject, TEmitted> {
...
+    this.current = this.behavior.receive(
+      this.current,
+      startSignal,
+      this.context
+    );
     return this;
   }
   public stop() {
-    this.behavior = this.behavior.receiveSignal(this.context, stopSignal);
+    this.current = this.behavior.receive(
```
Andarist, line 180: this should also call `this.mailbox.length = 0` before calling `this.behavior.receive`. `stop()` can be called while flushing and we can't continue processing events after `.stop` has been called

packages/core/src/Actor.ts
```diff
@@ -154,7 +154,7 @@ export class Actor<TEvent extends EventObject, TEmitted> {
   public current: TEmitted;
```
Andarist, line 154: should we rename all `TEmitter` to `TSnapshot`? 🤔

packages/core/src/behavior.ts
```diff
@@ -42,22 +42,21 @@ export type LifecycleSignal = typeof startSignal | typeof stopSignal;
  * @template TReceived The received event
  * @template TEmitted The emitted value
  */
-export interface Behavior<TReceived extends EventObject, TEmitted> {
+
+export interface Behavior<T extends EventObject, TEmitted = any> {
```
Andarist, line 46: to align with the rest of the codebase I think that we should use `TEvent` over `T` here

packages/core/src/behavior.ts
```diff
@@ -42,22 +42,21 @@ export type LifecycleSignal = typeof startSignal | typeof stopSignal;
...
-    event: TReceived
-  ) => Behavior<TReceived, TEmitted>;
-  receiveSignal: (
-    actorContext: ActorContext,
-    signal: LifecycleSignal
-  ) => Behavior<TReceived, TEmitted>;
-  /**
-   * The initial emitted value
-   */
+    state: TEmitted,
```
Andarist, line 48: I like the new signature 👍

packages/core/src/Actor.ts
davidkpiano: Done

packages/core/src/Actor.ts
davidkpiano: No because emitted !== snapshot (sometimes, but not always)

packages/core/src/behavior.ts
davidkpiano: Done

packages/core/src/behavior.ts
```diff
@@ -67,59 +66,59 @@ export function createDeferredBehavior<TEvent extends EventObject>(
...
-      return behavior;
-    },
-    receiveSignal: (actorContext, signal) => {
-      if (signal === startSignal) {
-        const sender: Sender<TEvent> = (e) => {
-          if (canceled) {
-            return;
-          }
+    receive: (_, event, actorContext) => {
+      if (isSignal(event)) {
```
Andarist, line 70: in the body of this block `event` is still being checked against concrete signals (start & stop) so this `isSignal` check seems somewhat redundant here

packages/core/src/behavior.ts
```diff
@@ -313,12 +316,12 @@ export function createServiceBehavior<TContext, TEvent extends EventObject>(
   service: Interpreter<TContext, TEvent>
 ): Behavior<TEvent, State<TContext, TEvent>> {
   const behavior: Behavior<TEvent, State<TContext, TEvent>> = {
-    receive: (actorContext, event) => {
+    receive: (state, event, actorContext) => {
+      if (isSignal(event)) {
```
Andarist, line 320: I wonder now - when does `createServiceBehavior` come useful? it is used to convert a living `service` to the actor ref~ but a service should already be compatible with ActorRef so we should be able to avoid rewrapping it to a similar thing.

I know that service is subtype of the Actor/ActorRef interface but that shouldn't matter much - people should be able to just use it in places where that supertype is expected

packages/core/src/behavior.ts
```diff
@@ -255,46 +258,46 @@ export function createMachineBehavior<TContext, TEvent extends EventObject>(
...
-                },
-                { origin: actorContext.self }
-              )
+              toSCXMLEvent(doneEvent, {
+                origin: actorContext.self
+              })
             );
           });
+
+          if (options?.sync) {
```
Andarist, line 279: can we kill `sync` since we have `getSnapshot`?

packages/core/src/behavior.ts
```diff
@@ -135,51 +134,54 @@ export function createPromiseBehavior<T, TEvent extends EventObject>(
   const observers: Set<Observer<T>> = new Set();
```
Andarist, line 134: if I recall this correctly - behaviors in Akka are pure, right? is the goal to make them pure in XState? if they are not pure - is calling them behaviors appropriate?

packages/core/src/behavior.ts
davidkpiano: You're right, it might not need to be wrapped. I'll test it.

packages/core/src/behavior.ts
davidkpiano: Yes, in a separate PR though because we want to be sure that we can support the pattern of receiving events via a subscription from an actor (which is basically what `sync` does, implicitly)

packages/core/src/behavior.ts
davidkpiano: Yes, they should be pure. To make them pure, they would look more like reducers than closures (which is what's happening in some of the behavior implementations)... if we want to make them pure, we have to ensure that internal state (such as the `canceled` variable here) isn't exposed as an emitted value. This might mean changing the typings.

packages/core/src/Actor.ts
davidkpiano: Done in 16a398a

packages/core/src/behavior.ts
davidkpiano: We need to check if it's a signal; otherwise TypeScript complains later. This has been refactored.

packages/core/src/behavior.ts
davidkpiano: You were right! 🎉

packages/core/src/Actor.ts
```diff
@@ -154,7 +145,7 @@ export class Actor<TEvent extends EventObject, TEmitted> {
...
   private context: ActorContext;
   private behavior: Behavior<TEvent, TEmitted>;
-  private mailbox: TEvent[] = [];
+  private mailbox: Array<TEvent | LifecycleSignal> = [];
```
Andarist, line 148: I don't think that `LifecycleSignal` can ever hit the mailbox - those are always directly passed to `receive` in start/stop methods

```suggestion
  private mailbox: Array<TEvent> = [];
```

packages/core/src/Actor.ts
```diff
@@ -169,31 +160,36 @@ export class Actor<TEvent extends EventObject, TEmitted> {
...
+      this.current,
+      stopSignal,
+      this.context
+    );
   }
   public subscribe(observer) {
     return this.behavior.subscribe?.(observer) || nullSubscription;
   }
-  public receive(event) {
+  public receive(event: TEvent | LifecycleSignal) {
```
Andarist, line 181: similarly here - `LifecycleSignal` shouldn't be passed manually here, it should always come from start/stop methods
```suggestion
  public receive(event: TEvent) {
```

packages/core/src/behavior.ts
Andarist: A somewhat bigger challenge here, than typings, would be how we should keep around and expose the public and internal "snapshots" - having access to the internal state is crucial for persisting & rehydrating

Andarist (approved):
Apart from 2 small nit picks and failing CI - LGTM

davidkpiano:
@Andarist Removing `LifecycleSignal` doesn't actually work

packages/core/src/behavior.ts
davidkpiano: That is why there is the potential for `.getSnapshot()` to be different than what `TEmitted` represents. This is why we need two types.

PR #2255, opened 2021-05-29
Add tests for action order for parallel transitions

Andarist:
~The second test is not passing since `a1` is never pushed to the array - it seems like something that should be fixed. Good news is that both pass on the `next` branch.~

For some reason the second test was failing for me locally, maybe some cache issue - nvmd though as it's OK now and CI passes as well


Andarist:
Lol, maybe it’s passing after all 🤷 it was failing for me locally, i will have to retest this when i get back home

PR #2256, opened 2021-05-29
Added plopfile for generating examples in example directory

mattpocock:
https://www.loom.com/share/0a3064907006444b97e969069eafbd00


examples/readme.md
```diff
@@ -0,0 +1,6 @@
+# Contributing an example
+
+1. Run `yarn generate` from the repo's root
```
davidkpiano, line 3: Should we have this instead be from the `/examples` directory, not the root (as discussed in Discord)?

mattpocock: It's pretty comfortable to have the generate function in the root - i.e. no need to cd around to run your command.

If we move it out of the root, we'll need to have another `package.json` in the `examples` dir to run this command only, which is less concise and more setup than just having it in root. So I prefer having the `yarn generate` and the `plop` installation in the root.

davidkpiano (approved):

davidkpiano:
Finally got around to testing this out; LGTM

PR #2261, opened 2021-05-31
[core] Add event schema map

davidkpiano:
This PR does two things:

1. Changes `config.schema.events` to be a mapping of events to their individual schemas instead of a unified schema

2. Adds an overload to `createMachine` that special-cases a `config` with a `.schema` property present for better inference

### DIFF
--- a/packages/core/test/schema.test.ts
+++ b/packages/core/test/schema.test.ts
@@ -53,7 +53,10 @@ describe('schema', () => {
             }
           }
         }),
-        events: createSchema<{ type: 'FOO' } | { type: 'BAR' }>()
+        events: {
+          FOO: {},
+          BAR: createSchema<{ value: number }>()
+        }
       },
       context: { foo: '', bar: 0, baz: { one: '' } },
       initial: 'active',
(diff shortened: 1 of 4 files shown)


packages/core/test/schema.test.ts
```diff
@@ -53,7 +53,10 @@ describe('schema', () => {
             }
           }
         }),
-        events: createSchema<{ type: 'FOO' } | { type: 'BAR' }>()
```
Andarist, line 56: To work around the lack of partial inference I've imagined that we are going to use schema to provide that. This change deviates slightly from it and I'm slightly worried that this will hurt the adoption of this pattern - people are already defining unions for `TEvent`, they just provide that explicitly as a generic to the `createMachine`. I've hoped that they could easily migrate to the schema-based generics and this won't allow that because each event has to be declared separately.

The proposed change also requires `EventFromEventSchemaMap` trickery - which works and is doable but I'm also worried that this will lead to less readable TS errors etc.

davidkpiano: Yeah, that makes sense.

PR #2266, opened 2021-05-31
[core] Add actions to model

davidkpiano:


packages/core/src/Machine.ts
```diff
@@ -50,12 +55,12 @@ export function Machine<
...
 
 export function createMachine<
-  TModel extends Model<any, any, any>,
+  TModel extends Model<any, any, any, any>,
   TContext = ModelContextFrom<TModel>,
   TEvent extends EventObject = ModelEventsFrom<TModel>,
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(
-  config: MachineConfig<TContext, any, TEvent>,
+  config: MachineConfig<TContext, any, TEvent, ModelActionsFrom<TModel>>,
```
Andarist, line 63: I would move this to the list of generics - so all the "computed" stuff from the model would live there

packages/core/src/model.ts
```diff
@@ -56,7 +63,8 @@ type EventCreators<Self> = {
 };
 
 type ModelCreators<Self> = {
-  events: EventCreators<Prop<Self, 'events'>>;
+  events?: EventCreators<Prop<Self, 'events'>>;
+  actions?: EventCreators<Prop<Self, 'actions'>>;
```
Andarist, line 67: They are both using `EventCreators` - I guess their behavior is pretty much the same, right? We could rename the `EventCreators` to smth like `ModelCreators` though

packages/core/src/model.ts
```diff
@@ -89,16 +98,27 @@ export function createModel<
   creators: TModelCreators
 ): Model<
   TContext,
-  Cast<
-    EventFromEventCreators<Prop<TFinalModelCreators, 'events'>>,
-    EventObject
-  >,
+  Prop<TFinalModelCreators, 'events'> extends never
```
Andarist, line 101: this could potentially read a little bit better, wdyt?

```ts
'events' extends keyof TFinalModelCreators
```

packages/core/src/model.ts
```diff
@@ -107,6 +127,7 @@ export function createModel(initialContext: object, creators?): unknown {
           type: eventType
         }))
       : undefined) as any,
+    actions: null as any,
```
Andarist, line 130: this ain't finished just yet, right? this can somewhat easily support custom actions - what about built-in ones? how to model, for example, `send` with it?

packages/core/test/model.test.ts
```diff
@@ -200,6 +200,31 @@ describe('createModel', () => {
...
+        actions: {
+          greet: (message: string) => ({ message })
+        }
+      }
+    );
+
+    createMachine<typeof userModel>({
+      context: userModel.initialContext,
+      initial: 'active',
+      entry: { type: 'greet', message: 'hello' },
```
Andarist, line 219: this won't be executed here correctly, should we provide the model's implementations automatically to the machine? this would require supporting `model` top-level property as this would have to be accessible at runtime

packages/core/src/types.ts
```diff
@@ -673,7 +673,11 @@ export interface MachineOptions<TContext, TEvent extends EventObject> {
 export interface MachineConfig<
   TContext,
   TStateSchema extends StateSchema,
-  TEvent extends EventObject
+  TEvent extends EventObject,
+  TActions extends ActionObject<TContext, TEvent> = ActionObject<
+    TContext,
+    TEvent
+  >
```
Andarist, line 680: I think this should basically be bound to the action equivalent of `EventObject`. I mean:
```ts
TAction extends { type: string }
```

This would allow for the stuff like this:
https://github.com/davidkpiano/xstate/pull/2266/files#diff-8637c68497a6e188837ab695801648e2f6f72a9831f7420c80409cc354fd1a45R219

but we don't have to embed `TContext` and `TEvent` within that type (or at least I don't see why this would be beneficial right now). Those types are still useful to type actions but it feels like those slots could be just provided as arguments etc, not necessarily as part of the action object~

packages/core/test/model.test.ts
```diff
@@ -200,6 +200,31 @@ describe('createModel', () => {
...
+      states: {
+        active: {
+          entry: [userModel.actions.greet('hello')]
```
Andarist, line 222: There is IMHO a conceptual clash between this and the other `entry` that is above here. This one is somewhat an equivalent of:
```js
entry: () => greet({ type: 'hello' })
```
while the other one is closer to:
```js
entry: (ctx, ev, { action }) => greet(action)
```

packages/core/src/model.ts
davidkpiano: This didn't work when I tried it

packages/core/src/model.ts
davidkpiano: Yes, we should ideally support built-in ones too. Not sure if we should do this via actually baking them in (wanting to be cognizant of bundle size) or something like:

```ts
actions: someModel.action(send('TYPE_SAFE_EVENT'))
```

So that we can pass in "bare" actions but they'll be type-safe

packages/core/test/model.test.ts
davidkpiano: Those aren't equivalent; actions that are like `() => something` are not pre-defined actions; they're arbitrary actions.

This is exactly equivalent to the direct object: `{ type: 'greet', message: 'hello' }`

packages/core/src/types.ts
davidkpiano: Added `BaseActionObject` for this

packages/core/test/model.test.ts
Andarist: What I've meant is that `model.actions.*` are opaque~ to the config etc, it never sees those parameters, while the other one (`{ type: 'greet', message: 'hello' }`) can be introspected. At least in the current `model.events.*`-like form.

It's not obvious how to translate `{ type: 'greet', message: 'hello' }` to `model.actions.greet` because the latter has only positional arguments etc. This is partially what I've meant by those two forms clashing with each other.

packages/core/src/model.ts
Andarist: > actions: someModel.action(send('TYPE_SAFE_EVENT'))

What kind of type-safety would this give to the user? I don't immediately see a benefit from this PoV. Could you elaborate?

A follow-up question would be - do you intend to support context/event-based payloads with this API?

packages/core/test/model.test.ts
davidkpiano: I'm still confused; this is no different than writing this:

```ts
// same as:
// entry: [model.actions.greet('hello')]
entry: [{ type: 'greet', name: 'hello' }]
```

Andarist: Disregard this whole thread - I've somehow assumed that `model.actions` would actually contain action implementations but they are only used to define action "shaped", the implementations still have to be delivered using regular means (using implementations~ argument).

packages/core/src/types.ts
```diff
@@ -73,6 +77,23 @@ export type Action<TContext, TEvent extends EventObject> =
...
+export type BaseAction<
+  TContext,
+  TEvent extends EventObject,
+  TAction extends BaseActionObject
+> =
+  | TAction['type']
+  | TAction
+  | SendAction<TContext, TEvent, any>
+  | AssignAction<TContext, TEvent>
+  | ActionFunction<TContext, TEvent>;
```
davidkpiano, line 89: @Andarist This makes types pass but I feel like we've had it before and removed it for some reason (I could be wrong), do you think this is fine?

packages/core/src/model.ts
davidkpiano: I added types for actions instead for now, even if they're the same. We can consolidate later, but I feel that they will be different, especially when we add guards, etc.

packages/core/src/model.ts
```diff
@@ -55,50 +62,91 @@ type EventCreators<Self> = {
...
+  ? EventObject
+  : {
+      [K in keyof TEventCreators]: TEventCreators[K] extends AnyFunction
+        ? ReturnType<TEventCreators[K]>
+        : never;
+    }[keyof TEventCreators];
 
 export function createModel<TContext, TEvent extends EventObject>(
   initialContext: TContext
-): Model<TContext, TEvent, void>;
```
Andarist, line 82: I think there was a reason behind this `void` here - will have to take this branch for a spin locally to try few things out. I'll try my best to add some additional type tests for the desired behavior.

packages/core/src/model.ts
```diff
@@ -107,6 +155,12 @@ export function createModel(initialContext: object, creators?): unknown {
           type: eventType
         }))
       : undefined) as any,
+    actions: actionCreators
+      ? mapValues(actionCreators, (fn, eventType) => (...args: any[]) => ({
+          ...fn(...args),
+          type: eventType
+        }))
```
Andarist, line 162:
```suggestion
      ? mapValues(actionCreators, (fn, actionType) => (...args: any[]) => ({
          ...fn(...args),
          type: actionType
        }))
```

packages/core/src/types.ts
```diff
@@ -73,6 +77,23 @@ export type Action<TContext, TEvent extends EventObject> =
   | ActionObject<TContext, TEvent>
   | ActionFunction<TContext, TEvent>;
 
+export type BaseAction<
+  TContext,
+  TEvent extends EventObject,
+  TAction extends BaseActionObject
+> =
+  | TAction['type']
+  | TAction
```
Andarist, line 86: This won't be type-safe - we'd have to filter out type-only actions and allow `actions: 'foo'` only for those. If an action type is defined as `{ type: 'bar', required: string }` we cannot allow `actions: 'bar'`.

packages/core/src/types.ts
Andarist: Hard to tell without giving this a spin locally - I think though that the list is missing some other builtin actions like cancel, log etc.

packages/core/src/types.ts
davidkpiano: Finally made this type-safe! https://github.com/statelyai/xstate/pull/2266/files#diff-8637c68497a6e188837ab695801648e2f6f72a9831f7420c80409cc354fd1a45R407-R436

davidkpiano:
@Andarist I think this is due for a re-review 🚀

packages/core/src/model.types.ts
```diff
@@ -28,20 +30,25 @@ export interface Model<
...
   events: Prop<TModelCreators, 'events'>;
+  actions: Prop<TModelCreators, 'actions'>;
   reset: () => AssignAction<TContext, any>;
   createMachine: (
-    config: MachineConfig<TContext, any, TEvent>,
-    implementations?: Partial<MachineOptions<TContext, TEvent>>
-  ) => StateMachine<TContext, any, TEvent>;
+    config: MachineConfig<TContext, any, TEvent, TActions>,
+    implementations?: Partial<MachineOptions<TContext, TEvent, TActions>>
+  ) => StateMachine<TContext, any, TEvent, any>;
```
Andarist, line 38: This has brought back the problem that got fixed here: https://github.com/statelyai/xstate/pull/2436

Andarist: I've pushed out a fix (and a test) for this: https://github.com/statelyai/xstate/pull/2266/commits/2f2895b03ac0871b6053cd4ffc5f6d2c872f98e4

packages/core/src/types.ts
```diff
@@ -735,7 +811,11 @@ export interface StateMachine<
   TContext,
   TStateSchema extends StateSchema,
   TEvent extends EventObject,
-  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext },
+  _TAction extends ActionObject<TContext, TEvent> = ActionObject<
```
Andarist, line 815: Is there any value in adding this here now if it isn't used anywhere?

packages/core/src/types.ts
```diff
@@ -74,6 +91,42 @@ export type Action<TContext, TEvent extends EventObject> =
...
+  | RaiseAction<any>
+  | SendAction<TContext, TEvent, any>
+  | AssignAction<TContext, TEvent>
+  | LogAction<TContext, TEvent>
+  | CancelAction
+  | StopAction<TContext, TEvent>
+  | ChooseAction<TContext, TEvent>
+  | ActionFunction<TContext, TEvent>;
+
+export type BaseActions<
```
Andarist, line 124: "Base" doesn't tell us much - what do you think about changing that to "Inline"? I'm not entirely sure if that was the intention but it seems to hold right now. And there might be some benefit in having `InlineAction` and `ImplementationAction` split anyway.

packages/core/test/model.test.ts
```diff
@@ -488,4 +488,26 @@ describe('createModel', () => {
     // @ts-expect-error
     model.actions.test();
   });
+
+  it('should allow for the action type to be explicitly given when creators have not been configured', () => {
```
davidkpiano, line 492: @Andarist I don't think we should support this; one of the biggest purposes of `createModel` is to _avoid_ specifying generic types. (Also, the tests fail)

packages/core/src/types.ts
davidkpiano: The intention is for there to be an action object without `exec`; I typically use `Base*` as a type name to signify a type that other types build on. But this isn't for a specific type of action; it's for all actions.

Andarist: This would still, potentially, leave `BaseActionObject` untouched.

packages/core/src/model.ts
```diff
@@ -14,7 +14,7 @@ import {
 
 export function createModel<TContext, TEvent extends EventObject>(
   initialContext: TContext
-): Model<TContext, TEvent, any, any>;
+): Model<TContext, TEvent, void, void>;
```
davidkpiano: We shouldn't use `void` here; the behavior should be the same as `createMachine` without the model (any actions are supported)

packages/core/src/Machine.ts
```diff
@@ -49,13 +50,16 @@ export function Machine<
 }
 
 export function createMachine<
-  TModel extends Model<any, any, any>,
+  TModel extends Model<any, any, any, any>,
   TContext = ModelContextFrom<TModel>,
   TEvent extends EventObject = EventFrom<TModel>,
-  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext },
+  TAction extends ActionObject<TContext, TEvent> = ModelActionsFrom<TModel>
```
Andarist, line 57: `ActionObject` has an `exec` property - it seems to me that this wouldn't be desired here, as this type is meant to only represent "action descriptions". What do you think about changing the constraint to the `BaseActionObject`? Where maybe the name of this type could be improved in the process (not sure if using `BaseActionObject` would fit this location here best)

packages/core/src/types.ts
davidkpiano: Ah I thought you were talking about `BaseActionObject` - maybe naming this `PlainActions` would be better

packages/core/src/Machine.ts
davidkpiano: `BaseActionObject` makes sense as the base for other action objects; let's keep the name for now. It's not meant to be used externally anyway.

packages/core/src/types.ts
davidkpiano: Not at the moment; this will become useful when we want to strongly type actions in `State` (separate PR)

packages/core/src/types.ts
```diff
@@ -74,6 +91,42 @@ export type Action<TContext, TEvent extends EventObject> =
...
+ */
+type SimpleActionsFrom<T extends BaseActionObject> = ActionObject<
+  any,
+  any
+> extends T
+  ? T // If actions are unspecified, all action types are allowed (unsafe)
+  : T extends {
+      [K in keyof Omit<T, 'type'>]: never;
+    }
+  ? T
```
Andarist, line 105: This condition is slightly indirect. It maps all other properties to `never` and then checks if `T` is a subtype of that, where it won't be if any key has been successfully remapped. The truthy branch is chosen when we end up with `T extends {}`.

A more straightforward approach to this problem could like this:
```ts
type ExtractSimple<T> = T extends any
  ? Exclude<keyof T, "type"> extends never
    ? T
    : never
  : never;

/**
 * Extracts action objects that have no extra properties.
 */
type SimpleActionsFrom<T extends BaseActionObject> = ActionObject<
  any,
  any
> extends T
  ? T // If actions are unspecified, all action types are allowed (unsafe)
  : ExtractSimple<T>;
```

This seems to test if the union member has only the `type` property more explicitly.

Note that we already have a slight variation of this in this very file:
https://github.com/statelyai/xstate/blob/58fd61d7c4f752dd517a25616b9a47dc39323fe2/packages/core/src/types.ts#L261-L265

davidkpiano: Sure, feel free to refactor this if the outcome is the same.

packages/core/test/model.test.ts
Andarist: I don't think this is super useful either, I've just added this since we already support the same for events:
https://github.com/statelyai/xstate/blob/095e40652c3e83b440d1c0b546df64bafc263dd0/packages/core/test/model.test.ts#L28-L31

As to the tests - silly mistake. I forgot that our type tests are also executed at runtime and didn't account for that. I've already fixed that.

packages/core/src/model.ts
Andarist: THis already got changed:
https://github.com/statelyai/xstate/blob/095e40652c3e83b440d1c0b546df64bafc263dd0/packages/core/src/model.ts#L21

Note that the last argument has to be `void` in this case. Otherwise those tests would not pass:
https://github.com/statelyai/xstate/blob/095e40652c3e83b440d1c0b546df64bafc263dd0/packages/core/test/model.test.ts#L478-L498

packages/core/src/Machine.ts
Andarist: What about changing the `extends` constraint here?

davidkpiano: To `BaseActionObject`? Works for me

davidkpiano:
@Andarist What are the remaining tasks here?

packages/core/src/types.ts
devanshj: (Just for the sake of bike-shedding & perhaps micro-blogging :P)

The most principally correct way of writing `ExtractSimple` is this...
```ts
type ExtractSimple<T> = T extends any ? { type: Prop<T, "type"> } extends T ? T : never : never

// for posterity
type Prop<T, K> = K extends keyof T ? T[K] : never
```
What we ultimately want to check is -- is `{ type: action.type }` assignable to `action` and that's exactly and literally what we're doing when we write `Pick<T, "type"> extends T ? T`.

All other ways of writing it are, if I may, incorrect. Even if they produce same results. And fun fact -- they don't. For example in Matuesz's `ExtractSimple` what he's doing is "check if there are no keys other than `type`" which is not what we want to check -- for `ExtractSimple<{ type: "foo", bar?: number }>` his version would produce `never` and my version would produce `"foo"` and mine is correct (unless the original requirement is not about assignability)

The reason I often champion to be "principally correct" rather than say "utilitarian correct" is that you don't have to think about edge cases. For example if didn't write my version thinking about the optional edge case, I just wrote it the way it was supposed to be written and then I just happen to realize this downside in Matuesz's version.

But why do I say "Even if they produce same results"? For being future proof. What if tomorrow typescript introduces a modifier say "unqiue" so that `keyof { unqiue type: "foo" }` is `"type"` but `{ type: "foo" }` is not assignable to `{ unqiue type: "foo" }` (unqiue essentially meaning that property value can't and shouldn't be produced alone) -- in that case my version would still work and some other version might fail.

Another good take away is that we aren't used to leveraging the fact that `extends` checks assignability and come up with really obscure ways to check it, like [Steve was doing and I suggested a refactored](https://twitter.com/devanshj__/status/1416468275408478209) and even the versions here are somewhat indirections.

My sincere apologies to use this space as my blog (high time I make one) and please ignore this comment -- use any version it doesn't matter much :P

Andarist: I actually thought about the optional properties case myself - just didn't figure out how to support it and put it aside as potential future improvement.

Thanks for providing the variant handling this!

Andarist (approved):
Just don't forget to add a changeset! 😉

packages/core/src/types.ts
devanshj: Ah gotcha. You're welcome!

devanshj:
Also the major concern I have about these action factories is that they only partially enable the user to provide the schema, the schema also should include implementation types. So I'd suggest that we first concretely think about the schema and address it's current shortcomings before adding features to `createModel` and publishing them.

davidkpiano:
> Also the major concern I have about these action factories is that they only partially enable the user to provide the schema, the schema also should include implementation types. So I'd suggest that we first concretely think about the schema and address it's current shortcomings before adding features to `createModel` and publishing them.

Can you show an example of the shortcoming here?

devanshj:
Sure. Take this example.

```ts
// Person A wrote the machine
let loginMachine = createMachine({
  schema: {
    context: createSchema<{ user: null } | { user: User }>(),
    events: createSchema<{ type: "LOGIN", credentials: Credentials }>(),
    actions: createSchema<{ type: "assignUser" }>,
  },
  context: { user: null },
  initial: "idle",
  states: {
    idle: { on: { LOGIN: "loggingIn" } },
    loggingIn: {
      invoke: "logInWithCredentials", // let's assume txstate knows this type via schema
      onDone: "loggedIn"
    },
    loggedIn: {
      entry: "assignUser"
    }
  }
})

// Person B is using it
useMachine(loginMachine, {
  logInWithCredentials: () => /* whatever */,
  assignUser: assign({ user: (_, e) => e.data })
})
```

All good now let's say that the Person A makes the following change...
```ts
- initial: "idle",
+ initial: "loggedIn"
```

Now Person B would get an error that `data does not exsist on e` but it's the Person A's fault, we need to show the error while Person A is authoring a the machine. This could be fixed with my version of the schema.

```ts
  schema: {
    context: t<{ user: null } | { user: User }>(),
    events: { LOGIN: t<{ credentials: Credentials }>() },
    actions: { assignUser: t<(c: never, e: { data: User }) => Assign<{ user: User }>>(),
    behaviors: { logInWithCredentials: t<(c: never, e: { credentials: Credentials }) => PromiseBehavior<User> }
  },
```

Now when Person A makes that change they'd get an error like so...

```ts
      entry: "assignUser"
//           ~~~~~~~~~~~~
// "assignUser" can't be assigned to "Error: `assignUser`'s event requirements are not matched`"
// (we'll could make the error as specific as possible just an example here)
```

So the schema indeed needs to provide implementation types for a good experience. Action factories will only merely provide the names and shape of the actions but not their requirements.

davidkpiano:
@Andarist Note that in `model.test.ts` there are some type errors; I narrowed the issue to `typeof userModel` not being assignable to `Model<any, any, any, any>` because the events are now `never`:

CleanShot 2021-07-22 at 16 19 19

Can you please take a look? We should fix these before merging.

Andarist:
@davidkpiano sure, I have not noticed them when I was working on this branch - nor CI has failed 🤔 Gonna give it a spin first time in the morning tomorrow

davidkpiano:
> @davidkpiano sure, I have not noticed them when I was working on this branch - nor CI has failed 🤔 Gonna give it a spin first time in the morning tomorrow

Strangely enough, CI will pass even with these errors present (they show in VS Code), maybe this has something to do with strict mode not being on? 

I fear that these errors _will_ appear where `strict: true` in other codebases.

devanshj:
Nitpick: the correct way to do `Foo<any, any>` is to actually `Foo<FirstParamConstraint, SecondParamConstraint>` like I do it [here](https://github.com/devanshj/txstate/blob/6620c824a77edce714f77b5e082a4df097624fdf/src/Behavior/index.ts#L3-L17). My suspicion is bringing `any` to picture is always dubious^1, when merging txstate I'll fix this everywhere :P Would also bring the naming convention, it's based of the fact that the implicit constraint of a type parameter is `unknown` hence when something is used as a constraint it needs to be `UnknownSomething`. 

And not only that it's also important to test if the concrete version is assignable to the unknown version (like I test if `Behavior<never, "foo">` which is basically `PromiseBehavior<"foo">` is assignable to `UnknownBehavior` or not), hence one has to get the variance of the type parameters right. I first got it wrong for `Behavior` so I changed all methods from `x: (a: A) => B` to `x(a: A): B` because the latter is [bivariant which is a less strict](https://twitter.com/karoljmajewski/status/1290805632602447874).

Saying in case it helps resolving the problem here I haven't taken a look at it tho :P

^1: For example in `type T = any extends "foo" ? "a" : "b"`, `T` is `"a" | "b"` which you most probably don't want. You likely need `unknown` instead of `any` which would give `"b"` instead (afk haven't double checked).

packages/core/src/model.types.ts
```diff
@@ -12,7 +12,18 @@ import {
...
 
-export type Cast<A1 extends any, A2 extends any> = A1 extends A2 ? A1 : A2;
+// https://github.com/microsoft/TypeScript/issues/23182#issuecomment-379091887
+export type IsNever<T> = [T] extends [never] ? true : false;
+
+/**
+ * Casts T to TCastType if T is assignable to TCastType.
+ * When `never` extends `TCastType`, `TCastType` will be chosen to prevent `never`.
+ */
+export type Cast<T extends any, TCastType extends any> = IsNever<T> extends true
```
davidkpiano, line 22: @Andarist @devanshj Can you please double check this?

My VS Code was showing errors (though mysteriously, CI was passing) because `Cast<never, EventObject>` resulted in `never`, when I would expect it to result in `EventObject`. This is because `never extends EventObject` (which still confuses me) but I added a check to make sure this doesn't happen.

devanshj: The short answer is I'd very strongly suggest to not change `Cast` because that's not the problem here and it's behaving perfectly fine. I'd suggest you to check for `never` at the place of issue and then swap it with `EventObject`.

But the fact that you need a check like this is weird can you tell me the original problem? Where exactly `Cast<never, EventObject>` being `never` was a problem? I'll clone the repo

I'm sure you know some of this but we use Cast in places like...
```ts
type Foo<T extends string> = {}
type Bar<T> = Foo<Cast<T, string>>
// Foo<T> doesn't match the constraint
```

But if you have...
```ts
type Foo<T extends string> = {}
type Bar<T extends never> = Foo<T>
// no problemo no point of using Cast and even if you use it Cast<never, string> will be never
```

So the fact you need your version of `Cast` itself is fishy

davidkpiano: For reference, the problem happens here: https://github.com/statelyai/xstate/pull/2266#issuecomment-885206683

davidkpiano: @devanshj The reason we need `Cast` here is because of this:

CleanShot 2021-07-23 at 18 15 09

> Type 'TComputedEvent' does not satisfy the constraint 'EventObject'.ts(2344)

Even though _we_ know that `TComputedEvent` will always be an `EventObject`, TS doesn't seem to know this.

devanshj: Yep I understand the original requirement of cast, that's valid. But wanting `Cast<never, EventObject>` to be `EventObject` (which is not "casting" behavior as `never` already suffices) is invalid. As I suspected the problem is something else. It's this...

TypeScript is picking the second overload, but why? Let's remove the second overload and let's keep only the model overload then this happens...

image

What typescript is saying is `typeof userModel` does not extend `Model<any, any, any, any>`, we can even check it...

image

Which basically means `userModel` can't be assigned to `Model<any, any, any, any>`...

image

And why is that? Because...

image

So perhaps what you were trying to do is make the `never` to be `EventObject` so that it works...

image

But that could be the incorrect way of solving it. Here are two possible reasons of the original problem -
1. `createModel` definition is bad, it should never return `Model<X, never>`
2. `Model` type definition is bad, `Model<X, never>` should extend `Model<any, any>` (or `UnknownModel` to be precise)

Now if the reason is former than your fix is correct to never return `Model<X, never>` but I'm suspicious the problem is latter. Why? I had a very similar problem with `Behavior`...

`createBehaviorFromPromise` returns `Behavior<never, T>` but it wasn't assignable to `UnknownBehavior`. Here too there could be the same two reasons - either `createBehaviorFromPromise` is bad or `Behavior` is bad. It was very apparent from the definition of `Behavior` that a promise behavior should be `Behavior<never, T>` - so `createBehaviorFromPromise` was right and the problem was with `Behavior`. So I made relax all the call signatures in `Behavior` so that the it's less strict and [assignability works](https://github.com/devanshj/txstate/commit/f92b56be918e2fd62db4cdbc11f2ec2692511164#diff-cd8075578a54f2b8997de6a2c06bccf76356e9c93bb5c153cefbd91fb33a1cefR17) (the changes are in the same commit you can take a quicklook)

Now in our case even if let's say the reason is latter, xstate is way too filled with "type-debt" to even fix it (I did attempt to change the variance of things all the way down but no luck). So I'd say let's just pretend the problem is former and never produce `Model<X, never>`.

But your definition of cast will make cast no longer cast (I'll explain later why `never extends X` holds), cast isn't the problem here and changing it will result in bigger problems so that's an absolute no-go. What you should do is this....

```ts
IsNever<TComputedEvent> extends true ? EventObject : Cast<TComputedEvent, EventObject>
```

I hope it helps! And sorry/welcome for the long explanation :P

devanshj: Okay so why `never extends X` holds? (X can be anything it'd always hold).

A type is basically a set of values assignable to it. So the set equivalent of `"a" | "b" | "c"` is `{ "a", "b", "c" }`. So when we want to check if A extends B what we're asking is - is a value of type B assignable to a value of type A (this is by the definition of `extends`). Which would be asking is the set equivalent of type B a subset of set equivalent of type A.

So when you look at the following code...
```ts
let a: "a" = "a";
let aOrBOrC: "a" | "b" | "c" = a;
```
it makes sense because `{ "a" }` is a subset of `{ "a", "b", "c" }`. Now `Ø` (null set) is also a subset of `{ "a", "b", "c" }` so the type equivalent of `Ø` should also be assignable `"a" | "b" | "c"`. Now the type equivalent of `Ø` in TypeScript is `never` which is called the "bottom type". Hence just like how `Ø ∪ X` is `X` similarly `never | X` is `X`.

Okay now fuck all this xD let me aslo explain you intuitively.

```ts
let x: "a" | "b" | "c" = {} as "a" | "b" | "c"   // as it is
let x: "a" | "b" | "c" = {} as "a" | "b"         // pop
let x: "a" | "b" | "c" = {} as "a"               // pop pop
let x: "a" | "b" | "c" = {} as never | "a"       // same as above
let x: "a" | "b" | "c" = {} as never             // pop pop pop
````

This is also the reason the lower bound in `type Extract<T, U> = T extends U ? T : never` is `never` so that `Extract<"a", "a">` gives `never`.

You can even think of it in terms of control flow analysis.
```
let x: "a" | "b" | "c" = {} as "a" | "b" | "c"
if (x === "c") return;
x // "a" | "b"
if (x === "b") return;
x // "a"
if (x === "a") return;
x // never;

let event: { type: string };
event = x; // now it seems should give an compile error because of the following line
event.type.toUpperCase();
// but this would never execute as the control flow never reaches here
// so it's okay to assign `never` to `{ type: string }`
```

What's why when you return `never` from a function typescript understands that the control flow will [never go past that](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3iAHgSwwC54AKASngF4A+eFEANxBgG4AoUSWBZdLKYuSq0A3gF82bXAXLso5IA). So essentially a function that returns `never` never actually returns. And hence again it's safe to write `let x: { type: string } = exit()` because exit would never return so who cares about `x`

Again sorry/welcome for the long explanation xD

Comment (Andarist) [reply packages/core/src/model.types.ts]: I would very much like to look at this before we merge this (even if only to level up my TS)... but it doesn't seem to be a lightweight material for a weekend 😬 I promise to take a look at this on Monday 🙏

Comment (davidkpiano) [reply packages/core/src/model.types.ts]: @devanshj I'm pretty sure we want the events for `Model` to never be `never`, because a model without events specified should result in a machine that can accept an `EventObject`. 

This PR implements your suggestion: https://github.com/statelyai/xstate/pull/2266/commits/458abd5750b0f6858c48d3fe779f987e2c64e2a0

Comment (devanshj) [reply packages/core/src/model.types.ts]: > because a model without events specified should result in a machine that can accept an EventObject

Why would you say that? A machine based on `Model<{}, { type: "FOO" } | { type: "BAR" }>` only takes events that are assignable to `{ type: "FOO" } | { type: "BAR" }`. Then in the same way a machine based on `Model<{}, never>` should take event assignable to `never` which basically means it should take no events.
Ah wait gotcha, you mean that if `createModel` has no event creators then we should fallback to `{ type: string }`, well if that's a conscious choice then it's okay.

But I still think `Model<{ foo: string }, never>` should extend `Model<unknown, { type: string }>` the root problem is that [`Action<{ foo: string }, never>` is not assignable to `UnknownAction` which is `Action<unknown, UnknownEvent>`](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggxsAlgewHYB4DCAaKBRKCAD2AlQBMBnKAVVQGtVkB3VPANzOAD4oBeKAAo4aUiQBcUHIU6pgkvAEp+vdskTkA3AChQkWgyasOXflADeUPREmVgAJ0SoA5lAC+u8NDqMWqeEhoZgEoGACuhn64PkZssjza2gA2EMBQRJIxfiFBAuZuUACG1DkYlgBmyMi2Dk6ubrioEJz23EA) which makes me think [`UnknownAction` should be `Action<never, never>` instead](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggxsAlgewHYB4DCAaKBRKCAD2AlQBMBnKAVVQGtVkB3VPANzOAD4oBeKAAo4aUiQBcUHIU6pgkvAEp+vdskTkA3AChQkWgyasOXflADeUPREmVgAJ0SoA5lAC+u8NDqMWqeEhoZgEoGKgQnPa44ZHc2toANhDAUESSPkb+CKFm5m5QAIbUIWjolgBmyMi2Dk6ubtEREPbcQA). So maybe instead of writing `Model<any, any>` one should write `Model<any, never>` idk. I'm not sure what's the "principally correct" way to solve it I haven't thought deeply about this.

But all this aside I given you're consciously making that choice then it's okay to do what you're doing with `CastIfNever`. Though to nitpick I'd suggest you to inline the `IsNever` check and remove the `CastIfNever`. Having that type would encourage it's usage where it not might be apt like a future person would go "Hmm Cast is not working, I see there is a CastIfNever" \*slaps that\* "Ha now it's working cool" and calls it a day xD instead of looking into the problem. But feel free to take your call, if you'd like to keep `CastIfNever` do that.

Comment (Andarist) [reply packages/core/src/model.types.ts]: Ok, I think I have finally digested this thread. 

Would it be correct to say that David's problem was caused by the variance issues? `Model["assign"]` is a function so its parameter types are contravariant. This should not result in an error under `strictFunctionTypes: false` which we currently have implicitly configured - we plan to change this and enable strict mode in the whole repo but this has not been done yet. I don't get why both you and David have seen this error and why I can't see it on my machine, nor CI fails. That being said - it's a good thing that this has been caught now but I don't understand why it was caught 😬 

I also quite don't understand the `UnknownBehavior` fix - I've tried to replicate that [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAqgdgazgewO5wKIDcJ2FAXigG8pRIAuKAZ2ACcBLOAcygF8AoD86AWWQAmEADYAeACoBhZHggAPYABoo47LnzzguAdViIU6NXgB8JDlAtQAhtWoNmcKgAoAxsDlUpMrQoCUhUyxkBgFzS2ptZwgsTyNgfwJA4NDObnBoeCQ0OH4hYUIoXJFRAFd9bOVMg0wcEy4XGVooAFtBESoq7KL8omI2a11u0VIGsq06KjgS5oAjCDp2ZTho+eMuDiA) and it doesn't type check (which makes sense to me with `strictFunctionTypes: true`). How your Behavior/UnknownBehavior is different from this?

---

In general, I agree with @devanshj - this statement seems like an off choice:
> model without events specified should result in a machine that can accept an EventObject.

If the whole point of types is to make our apps safer then we shouldn't produce unsafe variants by default. If the consumer doesn't want to type something, they are in a rush or having trouble doing so, then they should **explicitly** annotate stuff as being unsafe (with `any`). This IMHO should always be a conscious decision made by a user and not something that we promote by effectively making that a default that can be produced unintentionally.

Comment (davidkpiano) [reply packages/core/src/model.types.ts]: > If the whole point of types is to make our apps safer then we shouldn't produce unsafe variants by default. If the consumer doesn't want to type something, they are in a rush or having trouble doing so, then they should **explicitly** annotate stuff as being unsafe (with `any`). This IMHO should always be a conscious decision made by a user and not something that we promote by effectively making that a default that can be produced unintentionally.

Okay, that makes sense. I'll make the appropriate changes.

File: packages/core/test/model.test.ts [hunk 20]
```diff
@@ -199,6 +208,233 @@ describe('createModel', () => {
...
+        age: 30
+      },
+      {
+        actions: {
+          greet: (message: string) => ({ message })
+        }
+      }
+    );
+
+    userModel.createMachine({
```
Andarist, line 224: So I guess that the `createMachine<typeof userModel>` should work as well but fixing that can wait for @devanshj's feedback because I'm unsure how this should be refactored to allow that in the strict mode.

Andarist (approved):

packages/core/test/model.test.ts
davidkpiano: Right. `model.createMachine` does not give me type errors.

packages/core/src/model.types.ts
devanshj: > Would it be correct to say that David's problem was caused by the variance issues? Model["assign"] is a function so its parameter types are contravariant.

Right. Idr if it was the `assign` that caused the issue or not but you're probably right.

> I don't understand why it was caught 😬

Maybe our vscode was buggy and didn't pick the tsconfig? xD

>  I've tried to replicate that here

Not sure by "replicate" you meant replicate the bug or replicate the fix. If you meant the latter then nope xD Look closely the call signatures are different from `Behavior`, [here's how you fix it](https://www.typescriptlang.org/play?#code/C4TwDgpgBAqgdgazgewO5wKIDcJ2FAXigG8pRIAuKAZ2ACcBLOAcygF8AoDp4COgMwCGAY2gBZZABMIAGwA8AFQDCyPBAAewADRQF2XPg284k6rEQp0+vAD4SHKI6iDq1BszgAKYcHVVlqryaAJRUWMgMkg5O1LiSnhBY-tbAoVDhkRycHOTQ8EhocBLSMoRQxbJyAK4WhTr5lpg4tlzCqrRQALZSslQNhRWlRMRszmaDcqRtNbx0VHBVnQBGfOw6cIl8NlwcQA) (`type` also works but using `interface` just to be on a safer side). I tried applying this fix that day, either I applied it incorrectly or your repro is too minimal.

PR #2274, opened 2021-06-03
Fix useService hook API docs

loklaan:
The `service` given to `useService` should be started with `.start()`, or else an error occurs.

This may be an obvious step but it took me awhile to realise (after looking at the test suites for react & vue) 😅


davidkpiano:
Good catch!

Issue #2271, opened 2021-06-02
getSnapshot() should be part of the ActorRef type


VanTanev:
**Description**
Currently `getSnapshot()` is on the `SpawnedActorRef` type. This seems incorrect.
`SpawnedActorRef` seems to represent actors that XState controls in some fashion, while `ActorRef` is the more generic actor object.

It seems like `getSnapshot()` belongs on the generic `ActorRef` type, as all actors should provide this interface - including user-defined ones. Was this not done because of BC reasons?

I think `getSnapshot()` should be moved to the `ActorRef` type.

PR #2275, opened 2021-06-03
[core] Unify ActorRef type

davidkpiano:
This PR unifies the `ActorRef` type to contain all methods previously only present on the `SpawnedActorRef` type. This was found to be redundant in #2271, so now the types are as follows:

- `ActorRef` contains all methods (required)
- `BaseActorRef` only requires `send`; all other methods are optional. This should seldom be used.
- `SpawnedActorRef` is now an alias to `ActorRef`; it is deprecated and will be removed in v5.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -196,7 +196,7 @@ export interface ActivityDefinition<TContext, TEvent extends EventObject>
   type: string;
 }
 
-export type Sender<TEvent extends EventObject> = (event: Event<TEvent>) => void;
+export type Sender<TEvent extends EventObject> = (event: TEvent) => void;
 
 type ExcludeType<A> = { [K in Exclude<keyof A, 'type'>]: A[K] };
 
@@ -1280,12 +1280,12 @@ export type ExtractEvent<
 > = TEvent extends { type: TEventType } ? TEvent : never;
 
 export interface BaseActorRef<TEvent extends EventObject> {
-  send: Sender<TEvent>;
+  send: (event: TEvent) => void;
 }
 
 export interface ActorRef<TEvent extends EventObject, TEmitted = any>
   extends Subscribable<TEmitted> {
-  send: Sender<TEvent>;
+  send: Sender<TEvent>; // TODO: this should just be TEvent
   id: string;
   getSnapshot: () => TEmitted | undefined;
   stop?: () => void;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1279,30 +1279,38 @@ export type ExtractEvent<
   TEventType extends TEvent['type']
 > = TEvent extends { type: TEventType } ? TEvent : never;
 
+export interface BaseActorRef<TEvent extends EventObject, TEmitted = any> {
+  send: Sender<TEvent>;
+  id?: string;
+  getSnapshot?: () => TEmitted | undefined;
+  stop?: () => void;
+  toJSON?: () => any;
+  subscribe?: Subscribable<TEmitted>['subscribe'];
+}
+
 export interface ActorRef<TEvent extends EventObject, TEmitted = any>
   extends Subscribable<TEmitted> {
   send: Sender<TEvent>;
-}
-
-export interface SpawnedActorRef<TEvent extends EventObject, TEmitted = any>
-  extends ActorRef<TEvent, TEmitted> {
   id: string;
   getSnapshot: () => TEmitted | undefined;
   stop?: () => void;
   toJSON?: () => any;
 }
 
+/**
+ * @deprecated Use `ActorRef` instead.
+ */
+export type SpawnedActorRef<
+  TEvent extends EventObject,
+  TEmitted = any
+> = ActorRef<TEvent, TEmitted>;
+
 export type ActorRefFrom<
   T extends StateMachine<any, any, any> | Promise<any>
 > = T extends StateMachine<infer TContext, any, infer TEvent, infer TTypestate>
-  ? SpawnedActorRef<TEvent, State<TContext, TEvent, any, TTypestate>> & {
-      /**
-       * @deprecated
-       */
-      state: State<TContext, TEvent, any, TTypestate>;
-    }
+  ? ActorRef<TEvent, State<TContext, TEvent, any, TTypestate>>
   : T extends Promise<infer U>
-  ? SpawnedActorRef<never, U>
+  ? ActorRef<never, U>
   : never;
 
 export type AnyInterpreter = Interpreter<any, any, any, any>;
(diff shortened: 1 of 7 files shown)


packages/core/src/types.ts
```diff
@@ -1279,30 +1279,38 @@ export type ExtractEvent<
...
 
-export interface ActorRef<TEvent extends EventObject, TEmitted = any>
-  extends Subscribable<TEmitted> {
+export interface BaseActorRef<TEvent extends EventObject, TEmitted = any> {
   send: Sender<TEvent>;
+  id?: string;
+  getSnapshot?: () => TEmitted | undefined;
+  stop?: () => void;
+  toJSON?: () => any;
+  subscribe?: Subscribable<TEmitted>['subscribe'];
```
Andarist, line 1288: Shouldn't this `BaseActorRef` be only focused on the `send`? Isn't specifying more than that here "too much"? 🤔

packages/core/src/types.ts
```diff
@@ -1279,30 +1279,38 @@ export type ExtractEvent<
...
+> = ActorRef<TEvent, TEmitted>;
+
 export type ActorRefFrom<
   T extends StateMachine<any, any, any> | Promise<any>
 > = T extends StateMachine<infer TContext, any, infer TEvent, infer TTypestate>
-  ? SpawnedActorRef<TEvent, State<TContext, TEvent, any, TTypestate>> & {
-      /**
-       * @deprecated
-       */
-      state: State<TContext, TEvent, any, TTypestate>;
```
Andarist, line 1302: isn't removing this sort of a breaking change?

VanTanev:
What is the guidance about `actorRef.id`? Is it supposed to be globally-unique, or machine unique, or...?

It seems to be somewhat XState specific, rather than actor-specific.

VanTanev:
The `toActorRef()` functon seems to not infer well:
```ts
type TEvent = { type: 'one' } | { type: 'two' }

type TActor = BaseActorRef<TEvent>
const baseActor: TActor = null!

// Argument of type 'TActor' is not assignable to parameter of type 'BaseActorRef<EventObject, any>'.
//  Type 'EventObject' is not assignable to type 'TEvent'.
//    Type 'EventObject' is not assignable to type '{ type: "two"; }'.
//      Types of property 'type' are incompatible.
//        Type 'string' is not assignable to type '"two"'.(2345)
const a = toActorRef(baseActor)
```

packages/core/src/types.ts
davidkpiano: Not necessarily (it's type-only) but I'll add a deprecation notice.

davidkpiano: 73255c5

packages/core/src/types.ts
davidkpiano: e4bc10e

packages/core/src/types.ts
```diff
@@ -1279,30 +1279,39 @@ export type ExtractEvent<
   TEventType extends TEvent['type']
 > = TEvent extends { type: TEventType } ? TEvent : never;
 
-export interface ActorRef<TEvent extends EventObject, TEmitted = any>
-  extends Subscribable<TEmitted> {
+export interface BaseActorRef<TEvent extends EventObject, TEmitted = any> {
   send: Sender<TEvent>;
+  subscribe?: Subscribable<TEmitted>['subscribe'];
```
Andarist, line 1284: I've seen that you have removed this first but then added it back. Any particular reason? What makes `subscribe` special to be part of this interface (even if it's optional)?

packages/core/src/types.ts
```diff
@@ -1279,30 +1279,39 @@ export type ExtractEvent<
...
 }
 
-export interface SpawnedActorRef<TEvent extends EventObject, TEmitted = any>
-  extends ActorRef<TEvent, TEmitted> {
+export interface ActorRef<TEvent extends EventObject, TEmitted = any>
+  extends Subscribable<TEmitted> {
+  send: Sender<TEvent>;
   id: string;
   getSnapshot: () => TEmitted | undefined;
   stop?: () => void;
```
Andarist, line 1292: q for v5: should we kill `.stop` on ActorRefs? should such direct call only be allowed for parents? being able to stop an arbitrary actor when the ref gets shared seems to be dangerous

packages/core/src/types.ts
davidkpiano: It's what makes `TEmitted` relevant; I guess I can just remove it from the type.

packages/core/src/types.ts
davidkpiano: You should be able to stop an actor from any other actor that has a ref to it. This matches the behavior in Akka; I don't see how this can be dangerous (obviously, it's the responsibility of the developer to stop it properly): https://alvinalexander.com/scala/how-to-stop-akka-actors-in-scala-gracefulstop-poisonpill/

davidkpiano:
> What is the guidance about `actorRef.id`? Is it supposed to be globally-unique, or machine unique, or...?
> 
> It seems to be somewhat XState specific, rather than actor-specific.

Right; we can probably rename this to `name` (as it is in v5)

davidkpiano:
> The `toActorRef()` functon seems to not infer well:
> 
> ```ts
> type TEvent = { type: 'one' } | { type: 'two' }
> 
> type TActor = BaseActorRef<TEvent>
> const baseActor: TActor = null!
> 
> // Argument of type 'TActor' is not assignable to parameter of type 'BaseActorRef<EventObject, any>'.
> //  Type 'EventObject' is not assignable to type 'TEvent'.
> //    Type 'EventObject' is not assignable to type '{ type: "two"; }'.
> //      Types of property 'type' are incompatible.
> //        Type 'string' is not assignable to type '"two"'.(2345)
> const a = toActorRef(baseActor)
> ```

Fixed in 15c6e7d

packages/core/src/types.ts
Andarist: IMHO there are two dangers here:
1. it's an imperative/eager call, we should prefer pure actions 
2. that actor may live both in contexts of different actors and also in internal structures of its parent, especially for the latter we can't just shift the responsibility of handling that to the user as they don't control that at all

According to the Akka docs we can find [this](https://doc.akka.io/docs/akka/current/typed/from-classic.html#stopping):
> Classic actors can be stopped with the stop method of ActorContext or ActorSystem. [...] There is also a stop method in the ActorContext but it can only be used for stopping direct child actors and not any arbitrary actor.
and [this](https://doc.akka.io/docs/akka/current/actors.html#stopping-actors):
> Typically the context is used for stopping the actor itself or child actors and the system for stopping top level actors

So it seems to me that stopping the arbitrary actor might not be a recommended pattern in Akka, even if possible.

Andarist (approved):
This change affects public interfaces from the types perspective so adding some changeset about this would still be a good thing to do so people get aware of the change if it affects them anyhow

packages/core/src/types.ts
davidkpiano: You're right, but let's discuss in a separate PR, as this might be a breaking change.

davidkpiano: Card added: https://github.com/davidkpiano/xstate/projects/1#card-62575983

PR #2280, opened 2021-06-05
[core] Spawn/invoke behaviors

davidkpiano:
This PR allows you to spawn and/or invoke _behaviors_, which is the common interface for XState actors. Especially in v5, behaviors are not a new/special thing; promises/observables/callbacks/etc. in v5 all have specific behaviors that have the same interface, and in v5, _only behaviors_ are spawned/invoked. This is a step towards that migration.


Andarist:
IMHO behaviors are still somewhat a subject for refinement (maybe they will not need further refinement - but this has not been decided yet), see for example this thread: https://github.com/davidkpiano/xstate/pull/2242#discussion_r642108302 . I think it would be better to skip introducing this concept into v5 to avoid risking breaking changes further down the road for something that has just been introduced recently.

davidkpiano:
> IMHO behaviors are still somewhat a subject for refinement (maybe they will not need further refinement - but this has not been decided yet), see for example this thread: [#2242 (comment)](https://github.com/davidkpiano/xstate/pull/2242#discussion_r642108302) . I think it would be better to skip introducing this concept into v5 to avoid risking breaking changes further down the road for something that has just been introduced recently.

We don't have to expose/document it; behaviors can just be a refactor abstraction if you feel strongly we shouldn't be introducing this into v4. This has no breaking changes, and actually can clean up the code somewhat.

Andarist:
I'm just somewhat hesitant to expose this publicly at the moment - and not sure if there is much value in cleaning up the v4 codebase (we'll have to continuously resolve conflicts with the next branch and all)

packages/core/src/State.ts
```diff
@@ -126,7 +126,7 @@ export class State<
   /**
    * An object mapping actor IDs to spawned actors/invoked services.
    */
-  public children: Record<string, ActorRef<any>>;
+  public children: Record<string, SpawnedActorRef<any>>;
```
Andarist, line 129: since [this](https://github.com/davidkpiano/xstate/pull/2275) got merged in we should revert back to using `ActorRef` (there are multiple places in this PR doing the opposite switch)

packages/core/src/types.ts
```diff
@@ -1325,3 +1327,12 @@ export type InterpreterFrom<
 >
   ? Interpreter<TContext, TStateSchema, TEvent, TTypestate>
   : never;
+
+export interface ActorContext {
+  parent?: SpawnedActorRef<any, any>;
+}
+
+export interface Behavior<TEvent extends EventObject, TEmitted = any> {
```
Andarist, line 1335: Are we changing `TEmitted` to `TSnapshot` in v4? Or is it more of a v5 change?

```suggestion
export interface Behavior<TEvent extends EventObject, TSnapshot = any> {
```

Andarist (commented):
As mentioned privately - it would be really helpful to me if we could introduce `fromReducer` here, or at least propose the API and provide pseudocode for its usage. It would paint a better picture for me of how being able to spawn behaviors is an enabler for that feature and, in general, how things fit together.

packages/xstate-react/test/useSpawn.test.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import { useActor } from '../src';
+import { render, cleanup, fireEvent } from '@testing-library/react';
+import { useSpawn } from '../src/useSpawn';
+import { fromReducer } from 'xstate/src/behavior';
```
Andarist, line 5: I would export this from the core's index for now - so people won't have to reach for `/behavior`-named file.

packages/xstate-react/test/useSpawn.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+      }
+
+      return state;
+    };
+
+    const behavior = fromReducer(reducer, 0);
+
+    const Test = () => {
+      const actorRef = useSpawn(behavior);
+      const [count, send] = useActor(actorRef);
```
Andarist, line 23: Quite frankly - this doesn't sell me on those APIs. Why would I like to refactor the existing reducer to that? It would be great to list the benefits and possibilities of this API. The only thing for this particular code that comes to my mind right now is that we'd be able to hook this into our inspector - only to display the current state and stuff (not to actually draw a chart).

I'm trying to understand what are compelling reasons for users to refactor their code to this, arguably this is more boilerplate-y. This wouldn't be that much of a problem if people would have to write stuff like this in the beginning but if they must refactor less code lines to more code lines then we better give them something in return :p

packages/xstate-react/src/useSpawn.ts
```diff
@@ -0,0 +1,20 @@
...
+ * React hook that spawns an `ActorRef` with the specified `behavior`.
+ * The returned `ActorRef` can be used with the `useActor(actorRef)` hook.
+ *
+ * @param behavior The actor behavior to spawn
+ * @returns An ActorRef with the specified `behavior`
+ */
+export function useSpawn<TState, TEvent extends EventObject>(
+  behavior: Behavior<TEvent, TState>
+): ActorRef<TEvent, TState> {
+  const actorRef = useMemo(() => {
```
Andarist, line 15: This should `useConstant` instead of `useMemo` because the created `actorRef` is stateful. Unfortunately, `useMemo` is an optimization "hint" and not a semantic guarantee - and we don't want to lose state when fast refreshing and stuff.

`useConstant` also has its cons - the stored value won't react to input changes (`behavior` changes in this code) but it's not a must for us to support this (and also there is a high chance that we even shouldn't support it). Given cons of both solutions I would definitely should `useConstant` over `useMemo` here

packages/xstate-react/test/useSpawn.test.tsx
davidkpiano: This is a contrived example; one obvious benefit over `useReducer` is that the actor ref can be shared:

```jsx
const Parent = () => {
  const actorRef = useSpawn(behavior);

  return <div>
    <Child actorRef={actorRef} />
    <OtherChild />
  </div>
}
```

And now `<OtherChild />` won't rerender when the state in `actorRef` changes.

-----

Another benefit is that behaviors aren't just for reducers. Future example:

```jsx
const behavior = fromPromise(() => fetch('something').then(...));

const Test = () => {
  const actorRef = useSpawn(behavior);
  const [state, _] = useActor(actorRef);

  // ...
}
```

Now you have the exact same interface for behaviors other than reducers!

packages/core/src/types.ts
```diff
@@ -1325,3 +1328,23 @@ export type InterpreterFrom<
 >
   ? Interpreter<TContext, TStateSchema, TEvent, TTypestate>
   : never;
+
+export interface ActorContext {
+  parent?: ActorRef<any, any>;
+  self: ActorRef<any, any>;
```
Andarist, line 1334: it should be easy to type `self` more strictly - just add existing~ generics to `ActorContext`

packages/core/src/types.ts
```diff
@@ -1288,6 +1290,7 @@ export interface ActorRef<TEvent extends EventObject, TEmitted = any>
   send: Sender<TEvent>; // TODO: this should just be TEvent
   id: string;
   getSnapshot: () => TEmitted | undefined;
+  start?: () => void;
```
Andarist, line 1293: is this used anywhere? the `ActorRef` type represents already live actor - i don't think it should have a start method 🤔

packages/core/test/invoke.test.ts
```diff
@@ -2050,6 +2053,116 @@ describe('invoke', () => {
...
+          return count + 1;
+        } else {
+          return count - 1;
+        }
+      };
+
+      const countMachine = createMachine({
+        invoke: {
+          id: 'count',
+          src: () => fromReducer(countReducer, 0)
```
Andarist, line 2144: this feels like a new thing/feature - is the appeal to be able to embed reducer in a machine and use `getSnapshot` on its evolving state?

packages/core/test/invoke.test.ts
```diff
@@ -2050,6 +2053,116 @@ describe('invoke', () => {
...
+      };
+
+      const pingMachine = createMachine({
+        initial: 'waiting',
+        states: {
+          waiting: {
+            entry: send('PING', { to: 'ponger' }),
+            invoke: {
+              id: 'ponger',
+              src: () => pongBehavior
```
mattpocock, line 2112: Do behaviours passed into invoke.src always need to be like this? Or can they be declared like this:

`src: pongBehavior`

packages/core/test/invoke.test.ts
davidkpiano: Yes, and it shouldn't feel like a completely new thing. It should feel like invoking a callback, except using a reducer instead.

packages/core/test/invoke.test.ts
davidkpiano: That's a good question - sometimes, behaviors need to be dynamically instantiated with data from context (i.e., actors can be "born with innate knowledge"). So then the question becomes whether we want to support `src: () => behavior` as well as `src: behavior`, or avoid confusion and say that the signature must be:

```js
src: (context, event) => behavior
```

This is how it is in V5, btw ☝️

packages/core/src/types.ts
davidkpiano: I thought it was used in v5, will double-check. One use-case that came to mind was for _restarting_ actors:

```js
someActor.start(snapshot);
```

But there might be a better way to do this.

packages/xstate-react/test/useSpawn.test.tsx
davidkpiano: Ideally we would have `import { fromReducer } from 'xstate/behaviors'`, can we do this in v4?

Andarist: It's possible by adding an additional file (`packages/core/behaviors/package.json`) but I thought that we don't want to publicly call those behaviors in v4 and keep that as an implementation detail.

davidkpiano: We have to call this something:

```js
const actorRef = useSpawn(/* ??? */);
```

packages/core/src/types.ts
davidkpiano: Let's do this in v5

packages/xstate-react/test/useSpawn.test.tsx
davidkpiano: Let's call this behaviors, I think it's useful to introduce this as a term.

packages/xstate-react/test/useSpawn.test.tsx
davidkpiano: @Andarist Are you sold yet? 😜

packages/xstate-react/src/useSpawn.ts
davidkpiano: Do we want the actor to be restarted (with a new actor) when the `behavior` changes?

packages/core/src/StateNode.ts
```diff
@@ -1216,7 +1216,7 @@ class StateNode<
       },
       currentState
         ? { ...currentState.children }
-        : ({} as Record<string, ActorRef<any>>)
+        : ({} as Record<string, SpawnedActorRef<any>>)
```
Andarist, line 1219: This one should be reverted to the `ActorRef` as well

packages/core/src/behavior.ts
```diff
@@ -0,0 +1,59 @@
...
+    initialState
+  };
+}
+
+interface SpawnBehaviorOptions {
+  id?: string;
+  parent?: ActorRef<any>;
+}
+
+export function spawnBehavior<TActorEvent extends EventObject, TEmitted>(
```
Andarist, line 27: nit pick: I would rename `TActorEvent` to just `TEvent`. 

Right now the signature of this function is using `SpawnBehaviorOptions` and `TActorEvent` - so it IMHO is a little bit confusing if the main thing here is a behavior or an actor. Making this generic less specific~ would avoid this mismatch in naming entirely.

packages/core/src/behavior.ts
```diff
@@ -0,0 +1,59 @@
...
+  behavior: Behavior<TActorEvent, TEmitted>,
+  options: SpawnBehaviorOptions = {}
+): ActorRef<TActorEvent, TEmitted> {
+  let state = behavior.initialState;
+  const observers = new Set<Observer<TEmitted>>();
+
+  const actor = toActorRef({
+    id: options.id,
+    send: (event: TActorEvent) => {
```
Andarist, line 36: this currently lacks the concept of mailbox/scheduling. It doesn't matter how we achieve this right now but the point is that with this implementation it's possible to start processing a new event in the middle of processing a previous event

packages/core/src/behavior.ts
```diff
@@ -0,0 +1,59 @@
...
+      const eventObject = toEventObject(event);
```
Andarist, line 37: according to the signature of `send` this is redundant and it should always return the input object

packages/core/src/types.ts
```diff
@@ -1142,7 +1143,7 @@ export interface StateConfig<TContext, TEvent extends EventObject> {
   events?: TEvent[];
   configuration: Array<StateNode<TContext, any, TEvent>>;
   transitions: Array<TransitionDefinition<TContext, TEvent>>;
-  children: Record<string, ActorRef<any>>;
+  children: Record<string, SpawnedActorRef<any>>;
```
Andarist, line 1146: This one should also be reverted to the `ActorRef`

packages/core/src/types.ts
Andarist: Since this is not used anywhere and there is no clear use case defined for it - I would just drop this from this type for the time being. We can always revisit.

packages/core/src/types.ts
```diff
@@ -1325,3 +1328,23 @@ export type InterpreterFrom<
 >
   ? Interpreter<TContext, TStateSchema, TEvent, TTypestate>
   : never;
+
+export interface ActorContext<TEvent extends EventObject, TEmitted> {
```
Andarist, line 1332: mentioning "actor context" will become ambiguous now 😬 before this PR I would totally think of a particular machine-actor's context value.

packages/core/src/types.ts
```diff
@@ -1325,3 +1328,23 @@ export type InterpreterFrom<
...
+      type: 'xstate.start';
+    }
+  | { type: 'xstate.stop' };
+
+export interface Behavior<TEvent extends EventObject, TEmitted = any> {
+  transition: (
+    state: TEmitted,
+    event: TEvent | LifecycleSignal,
+    actorCtx: ActorContext<TEvent, TEmitted>
+  ) => TEmitted;
```
Andarist, line 1348: Q: do u still plan to introduce transducers anyhow? if yes - how they will relate to behaviors? should behavior's `transition` be pure?

packages/core/test/actor.test.ts
```diff
@@ -26,7 +33,7 @@ describe('spawning machines', () => {
   });
 
   const context = {
-    todoRefs: {} as Record<string, ActorRef<any>>
+    todoRefs: {} as Record<string, SpawnedActorRef<any>>
```
Andarist, line 36: And this one should be reverted too 😉

packages/xstate-react/src/useSpawn.ts
Andarist: I don't think so - we currently do not restart with a new machine if the config changes between `useMachine`/`useInterpret` re-renders.

packages/core/test/invoke.test.ts
```diff
@@ -2050,6 +2053,116 @@ describe('invoke', () => {
     });
   });
 
+  describe('with behaviors', () => {
```
Andarist, line 2056: we could use at last some basic test for spawning behaviors - unless I've missed it only tests using `invoke` have been introduced.

packages/core/src/types.ts
davidkpiano: I can't think of a better name - `context` is what's used by Akka

packages/core/src/types.ts
davidkpiano: Will have to think about this. The simplest approach is that the actions _are_ in the state (just like in a machine), but then we need to tell whatever is interpreting the behavior that those actions should be executed 🤔

packages/core/test/invoke.test.ts
davidkpiano: Done here: c30c0386cd87aa9ff7d73cbe2da703a9b3d5b5cd

packages/core/src/behavior.ts
davidkpiano: bf1e33a11a19871286f2a937c0028d33e3e5801d

packages/core/src/types.ts
Andarist: @mattpocock @farskid maybe you would have some names for this?

davidkpiano: Reference: https://doc.akka.io/japi/akka/current/akka/actor/ActorContext.html

My opinion: I'm fine with calling it `ActorContext` in the code but referring to it in the documentation just as an object, e.g.:

> You can access the parent from the 3rd argument:
> ```js
> const transition = (state, event, { parent }) => {
>   // ...
> };
> ```

packages/xstate-react/src/useSpawn.ts
davidkpiano: 2628fad6344437376d219489ea55278af76f2ce7

davidkpiano: 2628fad6344437376d219489ea55278af76f2ce7

packages/core/src/behaviors.ts
```diff
@@ -55,12 +59,31 @@ export function spawnBehavior<TEvent extends EventObject, TEmitted>(
...
+    }
+    flushing = false;
+  };
 
   const actor = toActorRef({
     id: options.id,
     send: (event: TEvent) => {
-      state = behavior.transition(state, event, actorCtx);
-      observers.forEach((observer) => observer.next(state));
+      enqueue(event);
```
Andarist, line 86: `enqueue` is a super short function - I would probably just inline it here to decrease the number of layers one has to go through to understand how processing works

packages/core/src/behaviors.ts
```diff
@@ -0,0 +1,111 @@
...
+          return event.data;
+        default:
+          return state;
+      }
+    },
+    initialState: undefined,
+    start: ({ self }) => {
+      promiseFn().then((data) => {
+        self.send({ type: 'resolve', data });
+      });
```
Andarist, line 44: We should also handle errors here and add handling of that event
```suggestion
      promiseFn().then(
        (data) => {
          self.send({ type: 'resolve', data });
        },
        (error) => {
          self.send({ type: 'reject', error });
        }
      );
```

packages/core/src/behaviors.ts
davidkpiano: 7a91088ba9f53aae846118e214b3091cca842719

packages/core/src/behaviors.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+  });
+
+  const actorCtx = {
+    parent: options.parent,
+    self: actor,
+    id: options.id || 'anonymous'
+  };
+
+  state = behavior.start?.(actorCtx) ?? state;
```
Andarist, line 104: It's really nit-picky, I know, but according to existing type restrictions I feel like this should be:
```suggestion
  state = behavior.start ? behavior.start(actorCtx) : state;
```

This has a very minor semantic difference - if there is a `start` method then its result will replace the initial state that is kept in the `state` at this point in time. It could matter for twisted scenarios with unions, rehydrating etc.

packages/core/test/actor.test.ts
```diff
@@ -919,6 +920,49 @@ describe('actors', () => {
...
+          success: {
+            type: 'final'
+          }
+        }
+      });
+
+      const countService = interpret(countMachine)
+        .onTransition((s) => {
+          console.log(s._event);
+        })
```
Andarist, line 959: this looks like a leftover :wink:

packages/core/src/behaviors.ts
```diff
@@ -0,0 +1,123 @@
...
+      switch (event.type) {
+        case 'fulfill':
+          parent?.send(doneInvoke(id, event.data));
+          return event.data;
+        case 'reject':
+          parent?.send(error(id, event.error));
+          observers.forEach((observer) => {
+            observer.error(event.error);
+          });
+          return undefined;
```
Andarist, line 45: isn't returning `undefined` here meaning that we won't be able to display the error in a component when using smth like this?
```js
const actorRef = useSpawn(fromPromise(() => Promise.reject(new Error('Foo.'))));
const [state] = useActor(actorRef);
```

should we allow it though? if not - how the user should handle those errors in the component's context

packages/core/src/behaviors.ts
```diff
@@ -0,0 +1,123 @@
...
+    transition: (state, event, { parent, id, observers }) => {
+      switch (event.type) {
+        case 'fulfill':
+          parent?.send(doneInvoke(id, event.data));
+          return event.data;
+        case 'reject':
+          parent?.send(error(id, event.error));
+          observers.forEach((observer) => {
+            observer.error(event.error);
+          });
```
Andarist, line 44: it seems a little bit quirky that observers are called here in the error case while they are called from within `flush` for the next item case

I'm not sure how to change this atm though

packages/core/src/behaviors.ts
davidkpiano: By using an observer:

```js
useEffect(() => {
  actorRef.subscribe({ error: e => setError(e) });
}, [actorRef]);
```

Andarist: isn't this boilerplate-y though? as a user I would expect error handling to be more of primary concern and not something that I have to wire up myself. By making it harder to handle this we risk people not handling errors as people are lazy (me included 😉 ). Some will handle them, some will not - the API doesn't make it easy though, it doesn't create a pit of success here.

davidkpiano: What would a more convenient API look like? What we _can't_ do is shove the error in the `state`; the `state` is meant to represent the emitted value.

Maybe we can do something like this:
```ts
const [state, send, { error }] = useActor(actorRef);
```

davidkpiano: Another alternative:
```js
const error = useActorError(actorRef); // usually `undefined`
```

Andarist: We might want to throw an error during rendering - which would require wrapping the internal state of `useActor` into a wrapper object, subscribing to errors like you have showcased here and throwing during rerender. Similarly to how, for example, `react-query` handles that:
https://github.com/tannerlinsley/react-query/blob/16b7d290c70639b627d9ada32951d211eac3adc3/src/react/useBaseQuery.ts#L124-L131

Andarist (approved):
LGTM, dont forget about adding a changeset before merging 😉

packages/core/src/behaviors.ts
davidkpiano: @Andarist I'm revisiting this, because you're right, the way we're representing promise state is a bit awkward.

What are your thoughts on this?

```js
const [{ data, error, status }, send] = useActor(fromPromise(/* ... */));
```

davidkpiano: Made the changes here: 30c310c

Andarist: Oh, I have sort of commented on that a minute ago on Discord so let me just copy-paste my comment from there:
> Huh, a lot of catching up here - I was thinking about the error stuff as well, just didnt have time to write down my thoughts here. I came to the conclusion that, in my opinion, we should throw errors and let error boundaries catch them - what @mattpocock#9242 said initially. This solution~ might be framework-specific, for now I'm thinking mostly about React and `@xstate/react`. I think it's reasonable for APIs of framework integration packages to diverge because they should feel native to the users of those frameworks, exact API parity between them is nice but imho a conceptual non-goal.

> I'm not saying that API definitely has to diverge, this would have to be explored - individually for each framework. I just lack experience with Vue etc so I can't say with any certainty how what I'm saying applies to those frameworks.

> It might feel a little bit unnatural but I think it's the reasonable choice for React. Note that there is never anything stopping people from normalizing "errors" into states~ and just having "error data" available as the first item returned from `useActor`. 

> A big problem with returning rejection/error values anyhow from any hook is that... errors are basically untyped in TS, so this doesn't provide any actual type safety and is unsound by definition. Guiding people towards normalizing their errors into "states" is creating a pit of success - they have to type this and normalize their error data into that expected shape. It's clear where and how one should do it - just add `.catch(() => ...)` and do it there, whatever you return from there has to match the expected type. It IMHO feels nice and is what I would, as a user, prefer.

> And unhandled errors would just go to an error boundary where it's expected that the received error value is any/unknown and that it has to be refined there.

> This is also similar to what React Query does - however, it's currently opt-in and called as experimental: https://react-query.tanstack.com/guides/suspense

PR #2282, opened 2021-06-06
Minor LifecycleSignal-related cleanup

Andarist:
Just some minor cleanup of redundant~ checks, followup to https://github.com/davidkpiano/xstate/pull/2242

### DIFF
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -18,7 +18,6 @@ import {
   createPromiseBehavior,
   createObservableBehavior,
   createBehaviorFrom,
-  LifecycleSignal,
   startSignal,
   stopSignal
 } from './behavior';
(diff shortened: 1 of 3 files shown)


davidkpiano:
There are more potential signals than starting and stopping:

```ts
export type ActorSignal =
  | { type: ActorSignalType.Start }
  | { type: ActorSignalType.PostStop }
  | { type: ActorSignalType.Watch; ref: ActorRef<any> }
  | { type: ActorSignalType.Terminated; ref: ActorRef<any> }
  | { type: ActorSignalType.Subscribe; ref: ActorRef<any> }
  | { type: ActorSignalType.Emit; value: any };
```

We have three choices:
- Give each of these signals their own method, like `.start()` and `.stop()` (not a fan of this idea)
- Add a `.receiveSignal(...)` method just for these signals (would revert the previous PR, also not a fan)
- Handle signals in `.receive(...)` (no change, current behavior)

davidkpiano:
Is this still applicable?

Andarist:
Sort of, I think it's worth re-examining how we handle those signals today and maybe making some changes related to that. I will have to take a look at the exact shape of this PR and `next` to assess how applicable this PR really is in its current form

davidkpiano:
Closing as stale

PR #2283, opened 2021-06-06
Add links to XState docs and API in JS Docs

santicros:
This PR adds JSDoc links to the docs and API urls of the core user-facing features, as inspired by the message sent by @Andarist here https://discord.com/channels/795785288994652170/799416911259369472/851189445705072640.

Let me know of anything! 😊

For example:
Captura de pantalla 2021-06-07 a las 0 44 23


ChrisShank:
🔥🔥

packages/core/src/Machine.ts
```diff
@@ -11,6 +11,15 @@ import {
 import { StateNode } from './StateNode';
 import { Model, ModelContextFrom, ModelEventsFrom } from './model';
 
+/**
+ * Docs: {@link https://xstate.js.org/docs/guides/machines.html#configuration Machines}.
+ * API: {@link https://xstate.js.org/api/globals.html#machine Machine}.
```
davidkpiano, line 16: I wonder if this can be automated somehow? 🤔 (just thinking for the future)

packages/core/src/Machine.ts
```diff
@@ -11,6 +11,15 @@ import {
...
+ *
+ * Recommendation: Prefer `createMachine`.
+ *
+ * Creates a new finite state machine from the config.
```
davidkpiano, line 20:
```suggestion
 * Creates a new finite-state machine (statechart) from the config.
```

packages/core/src/Machine.ts
```diff
@@ -49,6 +58,15 @@ export function Machine<
...
 }
 
+/**
+ * Docs: {@link https://xstate.js.org/docs/guides/machines.html#configuration Machines}.
+ * API: {@link https://xstate.js.org/api/globals.html#createmachine createMachine}.
+ *
+ * Creates a new finite state machine from the config.
+ *
+ * @param config The config object for creating the machine.
+ * @param options The optional options object.
```
davidkpiano, line 68:
```suggestion
 * @param options The implementation details for actions, guards, services, and delays.
```

packages/core/src/Machine.ts
santicros: That would be super awesome!! No idea how though 😂

with-heart: @davidkpiano As in like something that would automatically add the links to jsdoc blocks?

ChrisShank: Could you write a script that extracts a map the `id` of a headers/sub headers to its url from the documentation, then parses through the source code looking for jsdoc blocks that contain that  an certain`id` and injects the mapped link into the jsdoc?

ChrisShank: I guess more generically you need a script the extract a concept and its link from documentation, then another script that finds jsdoc blocks and determines what concept it best matches to and injects that concepts link into the jsdoc block!

davidkpiano:
The code has changed quite a bit since this, but we are already in the process of adding the relevant links.

Thank you!

PR #2285, opened 2021-06-07
Add utility function to xstate/immer to create assign action for models produced by createModel.

keithernet:


keithernet:
This is a WIP. I'm going to be offline for a few days, but wanted to get this in to make sure that I understood your proposed solution correctly. 

I wrapped the existing immer tests in a new describe which changed the whitespace and made the diff a bit of a mess. Let me know if you want me to undo that. 

When I'm back on Wednesday I'll have time to incorporate your feedback.

davidkpiano (approved):
cc. @Andarist

johtso:
Would be great to have this in! Spent a few hours trying to reinvent this wheel with all its generic hairiness with my limited typescript knowledge, then I found out this was here :)

Andarist:
Since we removed the `createModel` API in v5 ([here](https://github.com/statelyai/xstate/pull/3187)), I think there is no need for this one anymore. Please correct me if I'm wrong - we can always reconsider and reopen this PR

PR #2286, opened 2021-06-07
Deprecate useService()

davidkpiano:
With `.getSnapshot()`, the `useService()` hooks are virtually the same as `useActor()`.

### DIFF
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -83,7 +83,6 @@ export function createDeferredActor(
 
   if (isMachine(entity)) {
     // "mute" the existing service scope so potential spawned actors within the `.initialState` stay deferred here
-    // @ts-ignore
     const initialState = serviceScope.provide(
       undefined,
       () => (data ? entity.withContext(data) : entity).initialState
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -83,10 +83,10 @@ export function createDeferredActor(
 
   if (isMachine(entity)) {
     // "mute" the existing service scope so potential spawned actors within the `.initialState` stay deferred here
-    const initialState = serviceScope.provide(
+    const initialState = ((tempActor as any).state = serviceScope.provide(
       undefined,
       () => (data ? entity.withContext(data) : entity).initialState
-    );
+    ));
     tempActor.getSnapshot = () => initialState;
   }
 
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -84,10 +84,11 @@ export function createDeferredActor(
   if (isMachine(entity)) {
     // "mute" the existing service scope so potential spawned actors within the `.initialState` stay deferred here
     // @ts-ignore
-    tempActor.state = serviceScope.provide(
+    const initialState = serviceScope.provide(
       undefined,
       () => (data ? entity.withContext(data) : entity).initialState
     );
+    tempActor.getSnapshot = () => initialState;
   }
 
   return tempActor;
--- a/packages/xstate-react/README.md
+++ b/packages/xstate-react/README.md
@@ -115,6 +115,12 @@ A [React hook](https://reactjs.org/hooks) that interprets the given `machine` an
 
 ### `useService(service)`
 
+::: warning Deprecated
+
+In the next major version, `useService(service)` will be replaced with `useActor(service)`. Prefer using the `useActor(service)` hook for services instead, since services are also actors.
+
+:::
+
 A [React hook](https://reactjs.org/hooks) that subscribes to state changes from an existing [service](https://xstate.js.org/docs/guides/interpretation.html).
 
 **Arguments**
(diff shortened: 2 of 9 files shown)


packages/core/src/Actor.ts
```diff
@@ -83,11 +83,11 @@ export function createDeferredActor(
 
   if (isMachine(entity)) {
     // "mute" the existing service scope so potential spawned actors within the `.initialState` stay deferred here
-    // @ts-ignore
-    tempActor.state = serviceScope.provide(
```
Andarist, line 87: we might need both properties here for the time being - removing `.state` is a compatibility risk with `@xstate/react` etc

packages/xstate-react/README.md
```diff
@@ -126,7 +132,7 @@ A [React hook](https://reactjs.org/hooks) that subscribes to state changes from
 - `state` - Represents the current state of the service as an XState `State` object.
 - `send` - A function that sends events to the running service.
 
-### `useActor(actor, getSnapshot)`
+### `useActor(actor, getSnapshot?)`
```
Andarist, line 135: q: should we support `getSnapshot` in the future? or should we just expect that `actor.getSnapshot` is available?

packages/xstate-react/src/useActor.ts
```diff
@@ -31,11 +31,7 @@ const noop = () => {
 function defaultGetSnapshot<TEmitted>(
   actorRef: ActorRef<any, TEmitted>
 ): TEmitted | undefined {
-  return isActorWithState(actorRef)
-    ? actorRef.state
-    : 'getSnapshot' in actorRef
-    ? actorRef.getSnapshot()
-    : undefined;
+  return 'getSnapshot' in actorRef ? actorRef.getSnapshot() : undefined;
```
Andarist, line 34: similarly here - we can't remove accessing `state` because it's a breaking change. A user should be free to update `@xstate/react` without updating `xstate` at the same time (and vice versa)

packages/xstate-react/src/useService.ts
```diff
@@ -11,20 +11,20 @@ export function getServiceSnapshot<
...
   TEvent extends EventObject,
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(
   service: Interpreter<TContext, any, TEvent, TTypestate>
 ): [State<TContext, TEvent, any, TTypestate>, PayloadSender<TEvent>] {
-  if (process.env.NODE_ENV !== 'production' && !('machine' in service)) {
-    throw new Error(
-      `Attempted to use an actor-like object instead of a service in the useService() hook. Please use the useActor() hook instead.`
-    );
-  }
```
Andarist, line 25: IMHO the tip was still useful here and we could keep it as long as `useService` exists.

packages/xstate-vue/src/useActor.ts
```diff
@@ -35,7 +35,7 @@ export function useActor(
     | ActorRef<EventObject, unknown>
     | Ref<ActorRef<EventObject, unknown>>,
   getSnapshot: (actor: ActorRef<EventObject, unknown>) => unknown = (a) =>
-    isActorWithState(a) ? a.state : undefined
+    'getSnapshot' in a ? a.getSnapshot() : undefined
```
Andarist, line 38: same as here: https://github.com/davidkpiano/xstate/pull/2286/files#r651275518

packages/xstate-react/README.md
davidkpiano: We should not (but this is a major breaking change). We should expect `.getSnapshot` to be available.

packages/core/src/Actor.ts
davidkpiano: Done in b256bf3bc1a9ff02f8c81d594f5c9bc691b6e3e5

packages/xstate-react/src/useActor.ts
davidkpiano: Done in 3be0918b94ea24074fd63e2697f2c9b72e74d815

packages/xstate-react/src/useService.ts
davidkpiano: Done in ae2f97e5ba9423c354b626a7ea31eb7a4ab2c487

packages/xstate-vue/src/useActor.ts
davidkpiano: Done here: e218f1f2dd12d2ea29a69834060454bfa9add1b6

packages/xstate-react/test/useService.test.tsx
```diff
@@ -137,7 +137,7 @@ describe('useService hook', () => {
     expect(countEl.textContent).toBe('1');
   });
 
-  it('should throw if provided an actor instead of a service', (done) => {
+  it.skip('should throw if provided an actor instead of a service', (done) => {
```
Andarist, line 140: this can be unskipped now, right?

Andarist (approved):

packages/xstate-react/test/useService.test.tsx
davidkpiano: Unskipped

PR #2290, opened 2021-06-09
feat: add event meta to test machine event execution

SimeonC:
I have a use case where I wanted to check that the maximum length validation of a lot of fields was respected.

One other thing I slipped in was re-exporting the types file from @xstate/test - this is really useful when trying to build wrappers or utility functions around @xstate/test in strict typescript environments.

Thinking about it, there may be a better way of doing this (and I have a workaround) but I spent the time making the change before I realised that so thought I'd submit the MR anyway as it might be more useful in some other case.

### DIFF
--- a/.changeset/fifty-ants-deliver.md
+++ b/.changeset/fifty-ants-deliver.md
@@ -1,0 +1,6 @@
+---
+'@xstate/graph': minor
+'@xstate/test': patch
+---
+
+Adding event meta to the arguments for a test event execution. This should allow more flexibility for reusing the same event execution logic with a different config. For example, filling out multiple fields with random text where the only thing that changes is the input selector.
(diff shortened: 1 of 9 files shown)


davidkpiano:
Thanks for the PR! Thinking through this a bit more, instead of `eventMeta`, it might be useful to just pass the current `State`, which includes those transitions (+ meta) and potentially even more meta data that would be useful in executing the event.

SimeonC:
Just an update, other stuff got really busy, and I'm not sure when I'll get back around to this PR. If someone else wants to pick this up and push it over the finish line before I get the time to, please feel free!

SimeonC:
Going to close this, ended up not needing this at all and having other things to do. If someone else wants to pick this up feel free.

PR #2294, opened 2021-06-10
[v5] Restrict `context` to object 

davidkpiano:
This PR restricts `context` passed into `createMachine(...)` or `createModel(...)` to be an object.

Closes https://github.com/davidkpiano/xstate/projects/1#card-40598611

### DIFF
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -9,7 +9,8 @@ import {
   Typestate,
   HistoryValue,
   NullEvent,
-  ActorRef
+  ActorRef,
+  MachineContext
 } from './types';
 import { matchesState, keys, isString } from './utils';
 import { StateNode } from './StateNode';
@@ -17,7 +18,7 @@ import { isInFinalState, nextEvents, getMeta } from './stateUtils';
 import { initEvent } from './actions';
 
 export function isState<
-  TContext,
+  TContext extends MachineContext,
   TEvent extends EventObject,
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(state: object | string): state is State<TContext, TEvent, TTypestate> {
@@ -27,17 +28,20 @@ export function isState<
 
   return 'value' in state && 'history' in state;
 }
-export function bindActionToState<TC, TE extends EventObject>(
-  action: ActionObject<TC, TE>,
-  state: State<TC, TE, any>
-): ActionObject<TC, TE> {
+export function bindActionToState<
+  TContext extends MachineContext,
+  TEvent extends EventObject
+>(
+  action: ActionObject<TContext, TEvent>,
+  state: State<TContext, TEvent, any>
+): ActionObject<TContext, TEvent> {
   const { exec } = action;
-  const boundAction: ActionObject<TC, TE> = {
+  const boundAction: ActionObject<TContext, TEvent> = {
     ...action,
     exec:
       exec !== undefined
         ? () =>
-            exec(state.context, state.event as TE, {
+            exec(state.context, state.event as TEvent, {
               action,
               state,
               _event: state._event
@@ -49,7 +53,7 @@ export function bindActionToState<TC, TE extends EventObject>(
 }
 
 export class State<
-  TContext,
+  TContext extends MachineContext,
   TEvent extends EventObject = EventObject,
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 > {
@@ -95,15 +99,18 @@ export class State<
    * @param stateValue
    * @param context
    */
-  public static from<TC, TE extends EventObject = EventObject>(
-    stateValue: State<TC, TE, any> | StateValue,
-    context?: TC | undefined
-  ): State<TC, TE, any> {
+  public static from<
+    TContext extends MachineContext,
+    TEvent extends EventObject = EventObject
+  >(
+    stateValue: State<TContext, TEvent, any> | StateValue,
+    context?: TContext | undefined
+  ): State<TContext, TEvent, any> {
     if (stateValue instanceof State) {
       if (stateValue.context !== context) {
-        return new State<TC, TE>({
+        return new State<TContext, TEvent>({
           value: stateValue.value,
-          context: context as TC,
+          context: context as TContext,
           _event: stateValue._event,
           _sessionid: null,
           history: stateValue.history,
@@ -118,11 +125,11 @@ export class State<
       return stateValue;
     }
 
-    const _event = initEvent as SCXML.Event<TE>;
+    const _event = initEvent as SCXML.Event<TEvent>;
 
-    return new State<TC, TE>({
+    return new State<TContext, TEvent>({
       value: stateValue,
-      context: context as TC,
+      context: context as TContext,
       _event,
       _sessionid: null,
       history: undefined,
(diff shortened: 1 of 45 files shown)


VanTanev:
In typescript, `object | undefined` basically means "Anything that is object-like." All these pass:
```ts
type X = object | undefined

let a: X = []
let b: X = new Map()
let c: X = new String('a')
let d: X = new Number(1)
// ... etc
```

What is the restriction you're trying to apply?

davidkpiano:
> In typescript, `object | undefined` basically means "anything." All these pass:
> 
> ```ts
> type X = object | undefined
> 
> let a: X = []
> let b: X = new Map()
> let c: X = new String('a')
> let d: X = new Number(1)
> let e: X = new Number(NaN)
> let f: X = new Boolean(false)
> // ... etc
> ```
> 
> What is the restriction you're trying to apply?

Only literal objects `{ ... }` or `undefined`. I did this late at night 😅  Let me know if there's a better type for this.

VanTanev:
Unfortunately, as far as I know, there is no such type. 
This is because TS has no concept of a hash/map literal as a separate type.

I think that usually `Record<string, any> | undefined` communicates the idea better, but all the above examples pass the same way. I /think/ that there were some esoteric cases where it provided slightly better results than `object`, but can't remember off the top of my head.

Andarist:
Yep, this is a very unfortunate (and quite frankly - misleading) quirk of TS. Some simple examples showcased [here](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAQwhApgBygJXROAJwBMAeAFUXQA8p0xiBnRPAk05MATwBoVuAfAIAUyAFyJyASgkA3ODGKIA3gF8AUOtQZsrIsWEBGAAzGpWtFlz59wsOgDuiALLJMwqefWhIsBCktsAFEAW2wuCipaeiYVVRFxSRlEeUU4zW0rUPCjU3NM4LCoLjtHFzcPLx9oeCQCqAB5ACMAK3woSJo6BmY4VvaEiWk5BSU1DMDG-uhcswsdKbaZ+ydXd09NIA). I think the `object` reads best and since I'm not sure when `Record` would give us anything beyond `object` I think we can stick with it to keep things simple. And actually `object` might be better in some situations (see [here](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABABwE5wCYggUwPIBGAVjtADwAqiOAHlDmBgM6JzGlQB8AFAIYBciAN6IICenQD8gqgF8AlIJFiwEqDMSzhAKESJUOKCFRIwIADbnEvFrzABPbbO3bQkWAhTosuAEqk4VAxKajoGZkR-MSCyJihUGDAAcwAaawdOHgFhUXFaKGlEOUUclTUNLSFdfUNjUwsrG3THZ20VONZiRABeL0xsfHZobiEFbTYiADoy-MnUOww4AFsXdqgaiB6+nxwowIwRsYMIaby6OYXloA))

I wonder though if we have to support `undefined` - or rather: what should be the result of this:
```
const m = createMachine<undefined>({})
m.context // {} or undefined?
```
I think that defaulting to an empty object could potentially lead to less special-casing certain code paths in the implementation and that there is no much value in supporting undefined contexts.

packages/core/src/scxml.ts
```diff
@@ -104,24 +104,21 @@ const evaluateExecutableContent = <
...
     .filter(Boolean)
     .join('\n');
 
   const args = ['context', '_event'];
 
   const fnBody = `
-    ${scope}
-    ${body}
+${scope}
+with (context) {
```
davidkpiano, line 115: Asking for forgiveness for using the deprecated `with` statement, but ✨ _it works so well_ ✨

packages/core/test/machine.test.ts
```diff
@@ -202,6 +202,12 @@ describe('machine', () => {
         });
       }).toThrow();
     });
+
+    it('machines defined without context should have a default empty object for context', () => {
+      const machine = createMachine({});
+
+      expect(machine.initialState.context).toEqual({});
+    });
```
davidkpiano, line 210: ❕ New test

packages/core/test/machine.test.ts
```diff
@@ -239,7 +245,7 @@ describe('machine', () => {
         bar: 42
       });
 
-      expect(changedBarMachine.initialState.context).toBeUndefined();
+      expect(changedBarMachine.initialState.context).toEqual({ bar: 42 });
```
davidkpiano, line 248: ❕ Changed test outcome (`context` is no longer `undefined` ever)

packages/core/test/types.test.ts
```diff
@@ -297,3 +298,17 @@ describe('Typestates', () => {
...
+      // @ts-expect-error
+      context: 'string'
+    });
+  });
+
+  it('defined context passed to createModel() should be an object', () => {
+    // @ts-expect-error
+    createModel('string');
+  });
+});
```
davidkpiano, line 314: ❕ New tests

davidkpiano:
Reviewers: you can safely ignore most of the changes in these files (there's a ton), because they're mostly changing the types to enforce that `TContext` is of type `MachineContext`.

Look for the comments with an ❕ icon.

packages/core/src/State.ts
```diff
@@ -95,15 +99,18 @@ export class State<
...
+    stateValue: State<TContext, TEvent, any> | StateValue,
+    context?: TContext | undefined
+  ): State<TContext, TEvent, any> {
     if (stateValue instanceof State) {
       if (stateValue.context !== context) {
-        return new State<TC, TE>({
+        return new State<TContext, TEvent>({
           value: stateValue.value,
-          context: context as TC,
+          context: context as TContext,
```
Andarist, line 113: Since `context` can be `undefined` here and this is passed to the `State`'s constructor I think we should default to an empty object within that constructor as well. This would avoid `state.context` from being potentially `undefined` in more cases

packages/core/src/StateNode.ts
```diff
@@ -44,14 +45,17 @@ import { StateMachine } from './StateMachine';
...
+  TEvent extends EventObject
+> {
   _key: string;
   _parent?: StateNode<TContext, TEvent>;
   _machine: StateMachine<TContext, TEvent>;
 }
 
 export class StateNode<
-  TContext = any,
+  TContext extends MachineContext = {},
```
Andarist, line 58: Shouldn't it be like this?
```suggestion
  TContext extends MachineContext = MachineContext,
```

packages/xstate-graph/src/graph.ts
```diff
@@ -62,9 +62,11 @@ export function getChildren(stateNode: StateNode): StateNode[] {
   return children;
 }
 
-export function serializeState<TContext>(state: State<TContext, any>): string {
+export function serializeState<TContext extends MachineContext>(
+  state: State<TContext, any>
+): string {
   const { value, context } = state;
-  return context === undefined
+  return context === undefined || Object.keys(context).length === 0
```
Andarist, line 69: `context` should never be undefined noow, right? :)
```suggestion
  return Object.keys(context).length === 0
```

packages/core/src/State.ts
davidkpiano: 5cd04ee

Andarist (approved):

PR #2235, opened 2021-05-23
[docs] Verify TypeScript examples in docs 

VanTanev:
This is an addendum to #2234. Sorry that I have the original commits in this PR, couldn't figure out a better way to do it. 

## Verify and fix all TypeScript examples in docs

This PR  also introduces `yarn verifyTypescriptExamples` command that checks every piece of TypeScript code in the docs and runs it against the TS compiler.

PR #2298, opened 2021-06-13
Fix non-working code example in typescript.md

dmitry-guryev:
`createMachine` doesn't support this syntax: <LightContext, LightStateSchema, LightEvent>, but `Machine `supports

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -29,7 +29,7 @@ interface LightContext {
   elapsed: number;
 }
 
-const lightMachine = createMachine<LightContext, LightStateSchema, LightEvent>({
+const lightMachine = Machine<LightContext, LightStateSchema, LightEvent>({
   key: 'light',
   initial: 'green',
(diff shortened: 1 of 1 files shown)


docs/guides/typescript.md
```diff
@@ -29,7 +29,7 @@ interface LightContext {
   elapsed: number;
 }
 
-const lightMachine = createMachine<LightContext, LightStateSchema, LightEvent>({
+const lightMachine = Machine<LightContext, LightStateSchema, LightEvent>({
```
davidkpiano, line 32:
```suggestion
const lightMachine = createMachine<LightContext, LightEvent>({
```

We want to use `createMachine` and discourage the use of `StateSchema` for now.

davidkpiano (changes requested):

VanTanev:
This is fixed in #2235 >_>

davidkpiano:
Addressed in https://github.com/statelyai/xstate/pull/2413

PR #2316, opened 2021-06-20
[v5] Add Self type to MachineConfig

davidkpiano:


devanshj:
👀 I'm happy txstate is trailing the path for better types :P

(nitpick: I call them "Self-Reconstructing" type just saying so that we don't end up with names like "Typestate" ;P)

PR #2318, opened 2021-06-21
Patch environment check for use in web-dev-server

carpie:
I've started playing with WebComponents (`lit`) and no-build environments.  I created a test project with open-wc, which uses `web-dev-server` to import ES modules.  With `xstate`,  `environment.ts`  was checking `process.env` which caused `web-dev-server` to throw an exception.  This PR simply guards against `undefined` to allow `xstate` to work directly imported into a browser.

### DIFF
--- a/.changeset/six-windows-impress.md
+++ b/.changeset/six-windows-impress.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Patch environment check for use in web-dev-server
+The global `process` variable is now checked for existence before trying to read the `NODE_ENV` environment variable.
--- a/.changeset/six-windows-impress.md
+++ b/.changeset/six-windows-impress.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Patch environment check for use in web-dev-server
--- a/packages/core/src/environment.ts
+++ b/packages/core/src/environment.ts
@@ -1,1 +1,4 @@
-export const IS_PRODUCTION = process.env.NODE_ENV === 'production';
+export const IS_PRODUCTION =
+  typeof process !== 'undefined'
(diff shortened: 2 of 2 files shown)


.changeset/six-windows-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Patch environment check for use in web-dev-server
```
davidkpiano, line 5:
```suggestion
The global `process` variable is now checked for existence before trying to read the `NODE_ENV` environment variable.
```

davidkpiano (approved):
Just a changeset suggestion; otherwise LGTM 

cc. @Andarist

davidkpiano (approved):

Andarist:
This would cause extra code to be added for the majority of webpack users and would prevent dead code elimination from dropping dev-only code branches. You can import `xstate/dist/xstate.web.js` to use `process`-free bundle. I'd like to revisit this in v5 but I'm rather hesitant to include this in v4 unless we are OK with the bundlesize being increased by this.

davidkpiano:
> This would cause extra code to be added for the majority of webpack users and would prevent dead code elimination from dropping dev-only code branches. You can import `xstate/dist/xstate.web.js` to use `process`-free bundle. I'd like to revisit this in v5 but I'm rather hesitant to include this in v4 unless we are OK with the bundlesize being increased by this.

It's a very minimal increase, to be fair.

@carpie Can you see if the workaround with `xstate.web.js` works instead?

carpie:
> @carpie Can you see if the workaround with `xstate.web.js` works instead?

Yes! That works!  Sorry, I missed `xstate.web.js` in the docs.  I'm ok with using that. Thanks for the info!

PR #2320, opened 2021-06-21
[core] Improve generics for InvokeCallback type

### LINKED ISSUE #1471: [TypeScript] Generic types for `InvokeCallback`
**Description**

I believe the `InvokeCallback` type definition could be improved by adding generic types for its params.

**Expected Result**
Something like this:

```typescript
export type InvokeCallback<SEvent extends EventObject = any, REvent extends EventObject = EventObject> = ( 
  callback: Sender<SEvent>, 
  onReceive: Receiver<REvent> 
) => any; 
```

**Actual Result**

https://github.com/davidkpiano/xstate/blob/34a309717eff4ca04477a02c79e3d2ee3f142f8a/packages/core/src/types.ts#L205-L208

davidkpiano:
Fixes #1471


packages/core/src/types.ts
```diff
@@ -230,10 +230,13 @@ export type Receiver<TEvent extends EventObject> = (
   listener: (event: TEvent) => void
 ) => void;
 
-export type InvokeCallback<TEvent extends EventObject = AnyEventObject> = (
+export type InvokeCallback<
+  TEvent extends EventObject = AnyEventObject,
+  TReceivedEvent extends EventObject = AnyEventObject
+> = (
```
Andarist, line 236: For most types, we use `TEvent` as a type for events that the actor **receives** and not for events that it sends. So it kinda feels like the naming here should be flipped - `TReceivedEvent` should become `TEvent` and `TEvent` should become `TSentEvent`. There is already a precedence for `TSentEvent` here: https://github.com/davidkpiano/xstate/blob/18821577bd7572a00eecaf563c098526db06c854/packages/core/src/actions.ts#L196

packages/core/src/types.ts
```diff
@@ -230,10 +230,13 @@ export type Receiver<TEvent extends EventObject> = (
...
   callback: Sender<TEvent>,
-  onReceive: Receiver<TEvent>
-) => any;
+  onReceive: Receiver<TReceivedEvent>
+) => (() => void) | Promise<any> | void;
```
Andarist, line 239: Returning `Promise` can result in some very hard to debug issues - I actually spent some time recently figuring this out while helping out Kent.

So I wonder if maybe we shouldn't encourage this pattern in the types? I couldn't find a simple way to add a dev-only warning or anything about the issue that I was investigating.

davidkpiano: This is necessary because this pattern is supported:

```ts
src: () => async (sendBack, receive) => {
 // ...
}
```

The typings would fail if I remove `Promise<any>`, but I'm open to suggestions; I can't just omit it.

Andarist: This is exactly the problematic pattern that I've mentioned. Sorry for not being more specific earlier.

The problem here is that if the created promise resolves then:
- [this actor is removed from the parent](https://github.com/davidkpiano/xstate/blob/34ee27c008ca88798da1c4579fd94597f2467335/packages/core/src/interpreter.ts#L1010-L1018) and we no longer can't send any events to it (the `receive` won't be able to receive anything)
- `sendBack` should stop working (it probably doesn't stop working right now) because this actor has already sent `done.invoke.*` event to the parent and thus sending anything after that should be treated as invalid delivery

There is no easy way to distinguish `async` function from a regular one before actually calling it and inspecting its result (especially if you consider transpiled code as well). I think that in v5 we might provide both `fromCallback`, `fromAsync`, and `fromPromise` to differentiate between those patterns properly.

However, there is still a problem with the timing of the `async` completion - it feels natural to treat a resolved async actor as done but it effectively breaks the communication between it and the parent from that point in time. This sounds conceptually OK - if the actor has completed then no further communication should happen. It's a real gotcha though and I think that at least we should provide dev warnings for those cases but I'm unsure how this can be done effectively right now for `send(ev, { to: 'asyncActor' })`.

davidkpiano: > However, there is still a problem with the timing of the `async` completion - it feels natural to treat a resolved async actor as done but it effectively breaks the communication between it and the parent from that point in time. This sounds conceptually OK - if the actor has completed then no further communication should happen. It's a real gotcha though and I think that at least we should provide dev warnings for those cases but I'm unsure how this can be done effectively right now for `send(ev, { to: 'asyncActor' })`.

Let's discuss this in a separate issue/PR; seems tangential to this one.

For now, we have to keep `Promise<any>` in here just to make the types happy. We can refine in v5.

packages/core/src/types.ts
davidkpiano: 1e883f44cf6d3cddb823944da847f793121e13fc

packages/core/src/types.ts
davidkpiano: @Andarist Should we make this a new issue, or is it blocking this PR?

Andarist (approved):
Please just add a changeset before merging - other than that: LGTM

PR #2327, opened 2021-06-22
Use templates for examples

davidkpiano:
This fixes CI and uses templates (instead of PlopJS), and introduces the following templates: 

- [x] Vanilla TS
- [x] React + TS (with `@xstate/react`)
- [x] Vue + TS (with `@xstate/vue`)
- [ ] Svelte + TS (with `@xstate/svelte`)


examples/template-react-ts/src/index.css
```diff
@@ -0,0 +1,13 @@
+body {
```
mattpocock, line 1: Should we ditch the favicons and the CSS? Might mean a bit more noise in examples

mattpocock (approved):

examples/template-react-ts/src/index.css
davidkpiano: We should probably replace the favicons with XState favicons 🤔 will do that in a follow-up

PR #2334, opened 2021-06-24
Ensure that context is defined when model type provided to createMachine()

davidkpiano:
When using a model type in `createMachine<typeof someModel>(...)`, TypeScript will no longer compile machines that are missing the `context` property in the machine configuration:

```ts
const machine = createMachine<typeof someModel>({
  // missing context - will give a TS error!
  // context: someModel.initialContext,
  initial: 'somewhere',
  states: {
    somewhere: {}
  }
});
```

### DIFF
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -63,6 +63,8 @@ export function createMachine<
   TEvent extends EventObject = AnyEventObject,
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(
+  // Ensure that only the first overload matches models, and prevent
+  // accidental inference of the model as the `TContext` (which leads to cryptic errors)
   config: TContext extends Model<any, any, any>
     ? never
     : MachineConfig<TContext, any, TEvent>,
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -55,7 +55,7 @@ export function createMachine<
   TEvent extends EventObject = ModelEventsFrom<TModel>,
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(
-  config: MachineConfig<TContext, any, TEvent>,
+  config: MachineConfig<TContext, any, TEvent> & { context: TContext },
   options?: Partial<MachineOptions<TContext, TEvent>>
 ): StateMachine<TContext, any, TEvent, TTypestate>;
 export function createMachine<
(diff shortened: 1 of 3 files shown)


packages/core/src/Machine.ts
```diff
@@ -55,15 +55,17 @@ export function createMachine<
...
 ): StateMachine<TContext, any, TEvent, TTypestate>;
 export function createMachine<
   TContext,
   TEvent extends EventObject = AnyEventObject,
   TTypestate extends Typestate<TContext> = { value: any; context: TContext }
 >(
-  config: MachineConfig<TContext, any, TEvent>,
+  config: TContext extends Model<any, any, any>
+    ? never
+    : MachineConfig<TContext, any, TEvent>,
```
Andarist, line 68: This is supposed to help with cryptic errors when we fail to match the first overload and Model ges used in the second one as TContext, right? maybe worth adding a comment about this here?

davidkpiano: Yes, will do. The main use-case is that this:

```ts
createMachine<typeof model>({
  // no context
});
```

Will go to the second overload and assume that the `typeof model` is actually the `TContext`, which is bad.

davidkpiano: 8a1587e

packages/core/test/model.test.ts
```diff
@@ -220,4 +220,19 @@ describe('createModel', () => {
...
+
+  it('should not compile if missing context', () => {
+    const toggleModel = createModel({ count: 0 });
+
+    // @ts-expect-error
+    const m = createMachine<typeof toggleModel>({
+      id: 'machine',
+      initial: 'inactive',
+      // missing context:
+      // context: toggleModel.initialContext,
```
Andarist, line 232: A question that I feel is somewhat related to this.

There are cases for which the initial context is not known upfront - for example when the machine requires user data but it can only be passed from the parent since it first needs to be acquired from somewhere. With this change, you force a user to provide some dummy data here or you steer them towards casting which loses the type safety. Of course, currently, there is no type safety when you don't provide any context value, and yet you expect context to be of a certain type but that is an orthogonal problem. And it has existed before the model has even been introduced.

So the question is - what the user should do when using modal for cases like the one that I've described. What's the official recommendation?

davidkpiano: With this pattern, that's not an issue:

```ts
const createUserMachine = user => createMachine({
  context: { user }
});
```

davidkpiano: Another idea (from Discord) for reference:

```js
// TENTATIVE API
  // input is of type Partial<TContext> maybe
  context: (input) => ({
    some: 'default value',
    ...input
  })
});

machine.transition(/* ... */); // throws error; expecting input

machine.withContext({ the: 'input' }).transition(/* ... */);

interpret(machine); // throws error

interpret(machine.withContext({ ... })); // does not throw error
```

Andarist (approved):

PR #2343, opened 2021-06-25
Fix issue where always transition not taken in absence of enabled transitions

davidkpiano:
This PR fixes an issue where, if an event is sent to a machine in a state where it does not accept that event (no enabled transitions), any transient (`always`) transitions will be _ignored_. According to the [SCXML algorithm](https://www.w3.org/TR/scxml/#AlgorithmforSCXMLInterpretation) (see `mainEventLoop`), they should not be ignored.

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -1268,19 +1268,31 @@ class StateNode<
       delete history.history;
     }
 
-    if (!resolvedStateValue) {
+    // There are transient transitions if the machine is not in a final state
+    // and if some of the state nodes have transient ("always") transitions.
+    const isTransient =
+      !isDone &&
+      (this._transient ||
+        configuration.some((stateNode) => {
+          return stateNode._transient;
+        }));
+
+    // If there are no enabled transitions, check if there are transient transitions.
+    // If there are transient transitions, continue checking for more transitions
+    // because an transient transition should be triggered even if there are no
+    // enabled transitions.
+    //
+    // If we're already working on an transient transition (by checking
+    // if the event is a NULL_EVENT), then stop to prevent an infinite loop.
+    //
+    // Otherwise, if there are no enabled nor transient transitions, we are done.
+    if (!resolvedStateValue && (!isTransient || _event.name === NULL_EVENT)) {
       return nextState;
     }
 
     let maybeNextState = nextState;
 
     if (!isDone) {
-      const isTransient =
-        this._transient ||
-        configuration.some((stateNode) => {
-          return stateNode._transient;
-        });
-
       if (isTransient) {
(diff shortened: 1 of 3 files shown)


packages/core/test/transient.test.ts
```diff
@@ -670,4 +676,51 @@ describe('transient states (eventless transitions)', () => {
...
+          type: 'final'
+        }
+      }
+    });
+    const service = interpret(machine).onDone(() => {
+      done();
+    });
+
+    service.start();
+  });
```
Andarist, line 725: It feels like this test is longer than it could be and as such requires more mental overhead to understand it, whereas the basic requirement here can be represented in a quite simple way:

```suggestion
  it('should be taken even in absence of other transitions', () => {
    let shouldMatch = false;

    const machine = createMachine({
      initial: 'a',
      states: {
        a: {
          always: {
            target: 'b',
            // TODO: in v5 remove `shouldMatch` and replace this guard with:
            // cond: (ctx, ev) => ev.type === 'WHATEVER'
            cond: () => shouldMatch
          }
        },
        b: {}
      }
    });
    const service = interpret(machine).start();

    shouldMatch = true;
    service.send({ type: 'WHATEVER' });

    expect(service.state.value).toBe('b');
  });
```

packages/core/src/StateNode.ts
```diff
@@ -1268,19 +1268,31 @@ class StateNode<
...
+          return stateNode._transient;
+        }));
+
+    // If there are no enabled transitions, check if there are transient transitions.
+    // If there are transient transitions, continue checking for more transitions
+    // because an transient transition should be triggered even if there are no
+    // enabled transitions.
+    //
+    // If we're already working on an transient transition (by checking
+    // if the event is a NULL_EVENT), then stop to prevent an infinite loop.
```
Andarist, line 1286: Actually... we should continue checking transient transitions **if** we have successfully selected a transition. Not sure how hard it would be to cover this in the current implementation though.

davidkpiano: What I meant by this is that the way we "trigger" transient transitions is by raising a `NULL_EVENT`, and if we didn't have this check here, we would have transient transitions infinitely since there would always be a state with a transient transition (if we didn't transition to a new state).

So this check is necessary. Try removing it ➡️  infinite loop 😱

Andarist: I've added a test for the case I had in mind - it works after all 🎉 I got confused by the fact that you didn't check `willTransition` here but as it turns out `resolvedStateValue` check was synonymous to it in this context.

Andarist (approved):
Please don't forget to add a changeset for this - I've merged my PR to this branch so you may want to pull it from origin first.

PR #2351, opened 2021-06-27
[examples] New updated Vue 3 TODO-MVC

santicros:
This is a new version created from scratch of the Todo-MVC with Vue 3.
Previous version linked on the docs is not working and is outdated :)

Now it also uses the new `createModel`


examples/todo-mvc-vue/src/todos.machine.ts
```diff
@@ -0,0 +1,127 @@
...
+const todosModel = createModel(
+  {
+    todo: '',
+    todos: [] as Todo[],
+    filter: 'all'
+  },
+  {
+    events: {
+      'NEWTODO.CHANGE': (value: string) => ({ value }),
+      'NEWTODO.COMMIT': (value: any) => ({ value }),
```
santicros, line 31: Not sure why setting the value type a string gives an enormous TS error
```js
  NEWTODO.COMMIT': (value: string) => ({ value }), // this gives a TS error
```

davidkpiano: It's because of two things:

```diff
-          todos: (context, event: { value: string }) => {
+         todos: (context, event) => {
```

And this pesky conditional:

```ts
cond: (_, event) => event.value.trim().length
```

Gonna investigate this one.

santicros: @davidkpiano got it!! With your indications I removed the manual string and changed the conditional to:
```diff
-  cond: (_, event) => event.value.trim().length
+  cond: (_, event) => !!event.value.trim().length

```
and now TypeScript is happy :)

examples/todo-mvc-vue/src/todos.machine.ts
```diff
@@ -0,0 +1,127 @@
+import { createMachine, spawn, ActorRef } from 'xstate';
+// @ts-expect-error
```
davidkpiano, line 2: What's the error here?

davidkpiano (approved):

examples/todo-mvc-vue/src/todos.machine.ts
santicros: `uuid-v4` has no typescript definitions and `@types/uuid-v4` does not exist...
I used the same package from your example but could use a different package that has TS definitions :)

santicros: Changed uuid-v4 to nanoid in ba4c936 to remove this error :)

PR #2356, opened 2021-06-28
Add `model.createMachine(...)`

davidkpiano:
The model created from `createModel(...)` now provides a `.createMachine(...)` method that does not require passing any generic type parameters:

```diff
const model = createModel(/* ... */);

-const machine = createMachine<typeof model>(/* ... */);
+const machine = model.createMachine(/* ... */);
```


mattpocock:
When you call `model.createMachine` without passing `context`, does it infer the context from the model? I'd expect that to happen.

davidkpiano:
> When you call `model.createMachine` without passing `context`, does it infer the context from the model? I'd expect that to happen.

You mean the types or the actual value of the `initialContext`?

mattpocock:
The actual value

davidkpiano:
> The actual value

Added here: 6100fa085d7a3a20539adc91a5630db2eb7b934e

farskid (approved):

PR #2358, opened 2021-06-28
Add Symbol.observable to interpreter.

### LINKED ISSUE #1997: TS Type inference when converting interpreter to an Observable
Just as noticeable in this test - it seems to be impossible to use `rxjs.from()` with xstate interpreter without casting it to `any`.
Are there plans to support proper types for this? Is it even feasible?

https://github.com/davidkpiano/xstate/blob/07586ccbe165e5b88ff9149cd13022d9cbdb331b/packages/core/test/interpreter.test.ts#L1698

davidkpiano:
Fixes #1997


woutermont:
I opened #2361 with some extra changes to include in this branch. It builds and tests fine now, but Typescript still fails to infer the correct type of the resulting observable, always thinking it is an `Observable<never>`, and thus errors when trying to use its emissions. However, this seems to be easily overcome by explicitly stating the typings of the call `from<ObservableInput<State<TContext, TEvent, TStateSchema, TTypeState>>>(interpreter)`.

woutermont:
When would this be merged/released?

.changeset/itchy-moles-count.md
```diff
@@ -0,0 +1,9 @@
+---
+'xstate': minor
+'@xstate/fsm': minor
+'@xstate/inspect': minor
+'@xstate/react': minor
+'@xstate/vue': minor
+---
+
+add typings for interop observables (in core, affects dep versions in other packages)
```
Andarist, line 9: I don't think that other packages are impacted by this change in a meaningful way 🤔 
```suggestion
---
'xstate': patch
---

Add type for `Symbol.observable` to the `Interpreter` to improve the compatibility with RxJS.
```

woutermont: Me neither. If version bumps are not considered a minor change, then only the core package is enough.

woutermont:
Does this need another review, or when could we expect this be merged/released?

Andarist:
@woutermont please be patient. If this is blocking you anyhow please consider using [`patch-package`](https://github.com/ds300/patch-package) or similar until this gets released

We still have one unresolved issue with this PR (see https://github.com/statelyai/xstate/pull/2370 ). It's really minor - although I would prefer fixing it. I think it's somewhat OK to fix this later though if @davidkpiano agrees on that.

davidkpiano:
@Andarist I think we can fix that issue later, if it doesn't cause any actual problems.

Andarist (approved):

woutermont:
@Andarist, my bad, hadn't noticed your other PR. I can also take a look at that.

PR #2366, opened 2021-07-01
[core] Add ability to spawn actors in lazy initial context

davidkpiano:
Actors can now be spawned directly in the initial `machine.context` using lazy initialization, avoiding the need for intermediate states and unsafe typings for immediately spawned actors:

```ts
const machine = createMachine<{ ref: ActorRef<SomeEvent> }>({
  context: () => ({
    ref: spawn(anotherMachine, 'some-id') // spawn immediately!
  })
  // ...
});
```

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -289,7 +289,9 @@ class StateNode<
     /**
      * The initial extended state
      */
-    public context: Readonly<TContext> = undefined as any // TODO: this is unsafe, but we're removing it in v5 anyway
+    private _context:
+      | Readonly<TContext>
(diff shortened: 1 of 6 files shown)


packages/core/src/StateNode.ts
```diff
@@ -289,7 +289,9 @@ class StateNode<
     /**
      * The initial extended state
      */
-    public context: Readonly<TContext> = undefined as any // TODO: this is unsafe, but we're removing it in v5 anyway
```
Andarist, line 292: This was public property - hard to tell if somebody was using it but since `StateNode` is also used for the "root" node it's not that unlikely that this was accessed by somebody at runtime.

Andarist (approved):

PR #2367, opened 2021-07-01
[core] Improve raise() typings

davidkpiano:
This PR adds overloads to raise() to improve the type inference and prevent TS errors.


packages/core/src/actions.ts
```diff
@@ -159,6 +159,12 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
  *
  * @param eventType The event to raise.
  */
+export function raise<TEvent extends EventObject>(
```
Andarist, line 162: could you add some tests for this? it's hard to review this as it's not obvious what does this fix exactly

davidkpiano: Might be tricky; but I noticed it in places like this:

CleanShot 2021-07-01 at 09 29 17@2x

davidkpiano: @Andarist These fix `packages/core/test/types.test.ts` (see below)

Andarist: I wouldn't exactly say that those mentioned call sites were fixed. This, somehow, made just any event accepted at those positions.

Screenshot 2022-12-03 at 17 55 53

This is most likely related to the fact that our types are incorrect here:
https://github.com/statelyai/xstate/blob/b9995f0844bbbff0c813fee020935dfd7562184b/packages/core/src/types.ts#L227

The problem is that we should have 2 event-related generics flying around here.

One should be the filtered down `TEventExpression` that would come from this `Extract` here:
https://github.com/statelyai/xstate/blob/b9995f0844bbbff0c813fee020935dfd7562184b/packages/core/src/types.ts#L458
and the other one should just be `TEvent` coming from the top-level.

One is about what the given expression can receive and the other one is about what raise can "dispatch"

PR #2370, opened 2021-07-01
Avoid undefined property on the Interpreter's instance

Andarist:
Well, this one is somewhat iffy and I had some problems when preparing this.

I don't think it's entirely OK to do this https://github.com/davidkpiano/xstate/pull/2358/files#diff-fc91146eff717b2bcd6adbc0955ca0549346ad7bef9aa2278e0a75a9793a66b9R1279 . It will result in the method named `"undefined"` being added to our `Interpreter` class for a good chunk of our users (I doubt that many people polyfill `Symbol.observable`. This would look like a bug in our code.

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -107,7 +107,7 @@ declare global {
   }
 }
 
-export class Interpreter<
+class PrivateInterpreter<
   // tslint:disable-next-line:max-classes-per-file
   TContext,
   TStateSchema extends StateSchema = any,
@@ -164,7 +164,7 @@ export class Interpreter<
   public status: InterpreterStatus = InterpreterStatus.NotStarted;
 
   // Actor
-  public parent?: Interpreter<any>;
+  public parent?: PrivateInterpreter<any>;
   public id: string;
 
   /**
@@ -185,10 +185,10 @@ export class Interpreter<
    */
   constructor(
     public machine: StateMachine<TContext, TStateSchema, TEvent, TTypestate>,
-    options: Partial<InterpreterOptions> = Interpreter.defaultOptions
+    options: Partial<InterpreterOptions> = PrivateInterpreter.defaultOptions
   ) {
     const resolvedOptions: InterpreterOptions = {
-      ...Interpreter.defaultOptions,
+      ...PrivateInterpreter.defaultOptions,
       ...options
     };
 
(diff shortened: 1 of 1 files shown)


packages/core/src/interpreter.ts
```diff
@@ -164,7 +164,7 @@ export class Interpreter<
   public status: InterpreterStatus = InterpreterStatus.NotStarted;
 
   // Actor
-  public parent?: Interpreter<any>;
+  public parent?: PrivateInterpreter<any>;
```
Andarist, line 167: I will have to change most of the `PrivateInterpreter` occurrences to `Interpreter`. It's a little bit problematic as `this as Interpreter` isn't quite OK for TS 😢 So the most sensible solution is to just use `as any` but usually I try to only keep that for special occasions and with this change I will have to do that in a lot of places here.

packages/core/src/interpreter.ts
```diff
@@ -1349,13 +1348,24 @@ export function interpret<
...
     TStateSchema,
     TEvent,
     TTypestate
   >(machine, options);
 
-  return interpreter;
+  return interpreter as any;
+}
+
+export declare class Interpreter<
```
Andarist, line 1362: This removes `Interpreter` class from being available as runtime value - we have never encouraged using it directly but it could be a breaking change for somebody.

I couldn't figure out how to cast a class properly while preserving both its type and the runtime value (since classes in TS are both at the same time). It almost feels like it would be easier to just add `Symbol.observable` to the `Interpreter` once `.d.ts` gets emitted - it seems hacky as f but would fix a lot of annoyances here.

davidkpiano:
Do we need this now that https://github.com/davidkpiano/xstate/pull/2371 is merged?

Andarist:
@davidkpiano this is unrelated to #2371 

If we paste in this code into the console (on the site that doesn't polyfill `Symbol.observable`, please remember that it's not standard):
```js
class Interpreter {
    [Symbol.observable]() {
        return this;
    }
}
const service = new Interpreter();
console.log(service);
```
then we'll see that there is a property named `"undefined"` on the prototype of this class:
Screenshot 2021-07-02 at 08 58 44

This is not causing any problems per see but looks like a bug and raises eyebrows.

I'm not super fond of the solution implemented in this PR though - it seems to be too problematic. See the comment here: https://github.com/davidkpiano/xstate/pull/2370#discussion_r662247926

PR #2372, opened 2021-07-01
Allow modeling "done.invoke" events for model.assign(...)

davidkpiano:
This PR allows you to specify `'done.invoke'` as a restricted event type for modeling inline `model.assign(...)` actions in `onDone` blocks:

```ts
const model = createModel(
  {
    name: ''
  },
  {
    events: {
      SOME_EVENT: () => ({})
    }
  }
);

createMachine<typeof model>({
  context: model.initialContext,
  invoke: {
    src: () => new Promise((res) => res(42)),
    onDone: {
      // No type errors!
      actions: model.assign(
        {
          name: (_, e) => e.data
        },
        'done.invoke'
      )
    }
  }
});
```


johtso:
Shouldn't you also be able to type things explicitly?

Doing something like this doesn't seem to work when you have events defined in your model:
```ts
            actions: myModel.assign(
              (ctx, event: DoneInvokeEvent<MyResult>) => ({
                count: event.data.n,
                msg: event.data.s
              })
            )
```

https://codesandbox.io/s/explicit-invoke-done-event-typing-kf0dt?file=/src/index.ts

davidkpiano:
> Shouldn't you also be able to type things explicitly?
> 
> Doing something like this doesn't seem to work when you have events defined in your model:
> 
> ```ts
>             actions: myModel.assign(
>               (ctx, event: DoneInvokeEvent<MyResult>) => ({
>                 count: event.data.n,
>                 msg: event.data.s
>               })
>             )
> ```
> 
> https://codesandbox.io/s/explicit-invoke-done-event-typing-kf0dt?file=/src/index.ts

The current problem is that DoneInvoke events don't have a `type`, so you can't narrow it this way.

johtso:
Hmm.. would it make sense to be able to declare your invoke events in your model with your other events?

davidkpiano:
> Hmm.. would it make sense to be able to declare your invoke events in your model with your other events?
> 
> Or something like:
> 
> ```
> events: {
>   ...
> },
> invoke: {
>   'myInvoke.done': {foo: '', bar: ''},
>   'myInvoke.error': {msg: ''}
> }
> ```
> 
> Just brainstorming..
> 
> And also by default the model's eventtype could include generic event types for the different invoke events.. then you'd be able to do what I was trying to do?

Yeah, that can probably be covered in another PR. It's something I also want to support.

davidkpiano:
@Andarist This is ready for re-review

packages/core/src/model.types.ts
```diff
@@ -28,12 +29,18 @@ export interface Model<
...
       | Assigner<TContext, ExtractEvent<TEvent, TEventType>>
       | PropertyAssigner<TContext, ExtractEvent<TEvent, TEventType>>,
     eventType?: TEventType
-  ) => AssignAction<TContext, ExtractEvent<TEvent, TEventType>>;
+  ): AssignAction<TContext, ExtractEvent<TEvent, TEventType>>;
+  assign(
+    assigner:
+      | Assigner<TContext, DoneInvokeEvent<any>>
+      | PropertyAssigner<TContext, DoneInvokeEvent<any>>,
+    eventType: `done.invoke` | `done.invoke.${string}`
```
Andarist, line 42: The syntax for this has only been introduced in TS 4.1 (not even a year ago). Introducing this here would break all projects using older TS versions (which is very unfortunate :/)

Andarist: Since that string doesn't really matter at all - I think you could go with this:
```suggestion
    eventType: 'done.invoke'
```

davidkpiano: What happens if this is used in an older version of TS? Does it just not compile?

Andarist: I haven't rechecked it but I think that TS would not be able to create any types based on this file - it's effectively a syntax error for it. You can check out how `.D.TS` tab is empty on the right of this playground (it's setup to pre-4.1 version):
https://www.typescriptlang.org/play?ts=4.0.5#code/C4TwDgpgBAglC8UAGwIGdhIFCklAQgsqhgCQDeMAvtrtAMJEDkJwATE0A

Which in turn would make models unusable (or typed to any or smth like that) on those TS versions

davidkpiano: 7cdddb48a75836c0d1f5059f6a430ab0300c8d82

davidkpiano:
@Andarist When you get the chance, can this be re-reviewed? It's not perfect, but it's a stop-gap solution to a problem many are facing with `onDone` transitions.

Andarist:
We phased out `createModel` API so no new additions to it are planned.

PR #2374, opened 2021-07-03
Allow existing actors in spawn() to be referenced by name

davidkpiano:
Existing actors can now be identified in `spawn(...)` calls by providing an `id`. This allows them to be referenced by string:

```ts
  context: () => ({
    someRef: spawn(someExistingRef, 'something')
  }),
  on: {
    SOME_EVENT: {
      actions: send('AN_EVENT', { to: 'something' })
    }
  }
});
```

### DIFF
--- a/.changeset/red-parrots-judge.md
+++ b/.changeset/red-parrots-judge.md
@@ -1,0 +1,18 @@
+---
+'xstate': patch
+---
+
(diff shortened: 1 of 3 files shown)


mattpocock:
Does this also make them available in `service.children.get('something')`

davidkpiano:
> Does this also make them available in `service.children.get('something')`

Yes it does. But maybe not on `state.children['something']`, will investigate.

Andarist (approved):

PR #2379, opened 2021-07-04
Add `.preserveActionOrder` machine config option

davidkpiano:
This PR adds the opt-in `.preserveActionOrder` setting to machine configuration, which is `false` by default (current behavior).

When `.preserveActionOrder` is `true`, all actions will be called _in order_ instead of prioritizing `assign(...)` actions. The intermediate `context` result of `assign(...)` actions will be used to evaluate actions instead of the final `context`:

```ts
const machine = createMachine<{ count: number }>({
  context: { count: 0 },
  entry: [
    (ctx) => console.log(ctx.count), // 0
    assign({ count: (ctx) => ctx.count + 1 }),
    (ctx) => console.log(ctx.count), // 1
    assign({ count: (ctx) => ctx.count + 1 }),
    (ctx) => console.log(ctx.count) // 2
  ],
  preserveActionOrder: true
});
```

⚠️ This is a _stop-gap opt-in fix_ for the faulty `assign(...)` prioritization behavior that is already fixed in the upcoming v5: **preserved action order will be the default.**


packages/core/src/actions.ts
```diff
@@ -582,107 +582,142 @@ export function resolveActions<TContext, TEvent extends EventObject>(
...
+              actionObject,
+              machine.options.actions
+            );
+            const { exec } = resolvedActionObject;
+            if (exec && preservedContexts) {
+              const contextIndex = preservedContexts.length - 1;
+              resolvedActionObject.exec = (_ctx, ...args) => {
+                exec?.(preservedContexts[contextIndex], ...args);
+              };
+            }
```
Andarist, line 716: This will have to be ported to v5 as well since it doesn't preserve the order of custom actions yet. 

In general, I think we should simplify how we handle this (implementation-wise) but it's a good step forward.

packages/core/src/actions.ts
```diff
@@ -582,107 +582,142 @@ export function resolveActions<TContext, TEvent extends EventObject>(
...
+            return resolvedActionObject;
         }
-        default:
-          return toActionObject(actionObject, machine.options.actions);
-      }
-    })
+      })
+      .filter((a): a is ActionObject<TContext, TEvent> => !!a)
```
Andarist, line 720: What has changed that this `filter` is now needed? We were not filtering before those changes

packages/core/src/actions.ts
```diff
@@ -582,107 +582,142 @@ export function resolveActions<TContext, TEvent extends EventObject>(
...
-              )
+          case actionTypes.pure: {
+            const matchedActions = (actionObject as PureAction<
+              TContext,
+              TEvent
+            >).get(updatedContext, _event.data);
+            if (!matchedActions) {
+              return [];
+            }
+            const [resolvedActionsFromPure, resolvedContext] = resolveActions(
```
Andarist, line 677: `preserveActionOrder` should be passed to this `resolveActions` call as well and the appropriate test for this should be added

packages/core/src/actions.ts
```diff
@@ -582,107 +582,142 @@ export function resolveActions<TContext, TEvent extends EventObject>(
...
+            })?.actions;
+
+            if (!matchedActions) {
+              return [];
+            }
+
+            const [
+              resolvedActionsFromChoose,
+              resolvedContextFromChoose
+            ] = resolveActions(
```
Andarist, line 658: Similar thing as here: https://github.com/statelyai/xstate/pull/2379/files#r665115941

packages/core/src/actions.ts
davidkpiano: `actionTypes.assign` returns `undefined`, not an `ActionObject`

packages/core/src/actions.ts
davidkpiano: 5ae1959d41ac90eab15a46c57e135d59dd2ca05f

packages/core/src/actions.ts
davidkpiano: 5ae1959d41ac90eab15a46c57e135d59dd2ca05f

packages/core/src/actions.ts
davidkpiano: Yep, the v4 tests will also verify the correct behavior in v5.

Andarist (approved):
Please don't forget about a changeset! 😉

PR #2391, opened 2021-07-06
[core] Add `ContextFrom<T>` and `EventFrom<T>`

davidkpiano:
This PR introduces two new helper types:

- `ContextFrom<T>` which extracts the `context` from any type that uses context
- `EventFrom<T>` which extracts the `event` type (which extend `EventObject`) from any type which uses events

You can extract `context` and `event` types from a model using the `ContextFrom<T>` and `EventFrom<T>` types:

```ts 
import { ContextFrom, EventFrom } from 'xstate';
import { createModel } from 'xstate/lib/model';

const someModel = createModel(
  {
    /* ... */
  },
  {
    events: {
      /* ... */
    }
  }
);

type SomeContext = ContextFrom<typeof someModel>;
type SomeEvent = EventFrom<typeof someModel>;
```


packages/core/src/types.ts
```diff
@@ -1349,3 +1350,25 @@ export interface Behavior<TEvent extends EventObject, TEmitted = any> {
...
+    ? TEvents0
+    : T extends Model<any, infer TEvents1, any>
+    ? TEvents1
+    : T extends State<any, infer TEvents2, any, any>
+    ? TEvents2
+    : T extends Interpreter<any, any, infer TEvents3, any>
+    ? TEvents3
+    : never);
+
+export type ContextFrom<T> = object &
```
Andarist, line 1365: what `object &` brings to the table here? 🤔

johtso:
Wouldn't we want to also be able to get the type for a specific `eventType`? Just having the type of all defined events doesn't let us specify the type of event an action is expecting to receive for example..

packages/core/src/types.ts
davidkpiano: Restricts the resulting type to an object; is there a better way?

davidkpiano:
> Wouldn't we want to also be able to get the type for a specific `eventType`? Just having the type of all defined events doesn't let us specify the specific types of events an action is expecting to receive for example..

Let's do that in a separate PR and keep the scope of this one small.

johtso:
> > Wouldn't we want to also be able to get the type for a specific `eventType`? Just having the type of all defined events doesn't let us specify the specific types of events an action is expecting to receive for example..
> 
> Let's do that in a separate PR and keep the scope of this one small.

Aha.. just figured out I can do:
```ts
ExtractEvent<SeatPickerEvents, "userUpdateTickets">
```

packages/core/src/types.ts
Andarist: Since all the slots you are inferring should be restricted to something (like an `object`), you shouldn't worry about it here - an incompatible type should never be created in the first place (it wouldn't compile). And if an incompatible type can be created then your restriction here is not correct anyway, as in - it could create problems down the road.

davidkpiano: 1568e0ac9a84135c06be2ed54f01d5005dd70d7d

packages/core/src/types.ts
```diff
@@ -1349,3 +1350,29 @@ export interface Behavior<TEvent extends EventObject, TEmitted = any> {
   initialState: TEmitted;
   start?: (actorCtx: ActorContext<TEvent, TEmitted>) => TEmitted;
 }
+
+export type EventsFrom<T> = EventObject &
```
Andarist, line 1354: Same suggestion as here: https://github.com/statelyai/xstate/pull/2391#discussion_r666537803 . I would drop the `EventObject &` bit.

packages/core/src/types.ts
```diff
@@ -1349,3 +1350,29 @@ export interface Behavior<TEvent extends EventObject, TEmitted = any> {
...
+  (T extends StateMachine<any, any, infer TEvents0, any>
+    ? TEvents0
+    : T extends Model<any, infer TEvents1, any>
```
Andarist, line 1357: Within all of those types here, we use `TEvent` (singular) and not `TEvents`, I would probably go for consistency and rename all those occurrences here to a singular form. This also brings up a question - maybe this helper should be named `EventFrom` and not `EventsFrom`? 🤔

packages/core/src/types.ts
```diff
@@ -1316,8 +1317,8 @@ export type ActorRefFrom<
     }
   : T extends Promise<infer U>
   ? ActorRef<never, U>
-  : T extends Behavior<infer TEvent, infer TEmitted>
-  ? ActorRef<TEvent, TEmitted>
+  : T extends Behavior<infer TEvent1, infer TEmitted>
```
Andarist, line 1320: If you have switched to using the "ordinal" suffix here then I would expect to see `TEvent0` in the `StateMachine` match above, OTOH - maybe it's more ergonomic to not use those suffixes at all? do they improve readability for you?

packages/core/src/model.ts
```diff
@@ -118,6 +44,7 @@ export function createModel(initialContext: object, creators?): unknown {
     reset: () => assign(initialContext),
     createMachine: (config, implementations) => {
       return createMachine(
+        // @ts-ignore
```
Andarist, line 47: I've tried removing this locally and it seems that it works just OK without this ignore comment

Andarist (approved):
I have left some very minor comments, other than that LGTM

packages/core/src/types.ts
davidkpiano: TypeScript warns when I use the same names for type parameters.

packages/core/src/types.ts
davidkpiano: 6a197fcacd22876770c350ead150ec8e1d9d13cc

packages/core/src/types.ts
davidkpiano: 6a197fcacd22876770c350ead150ec8e1d9d13cc

packages/core/src/model.ts
davidkpiano: d3413bb2f75562ea39f1d2301b8261821368e9ae

packages/core/src/types.ts
davidkpiano: e308b9b

Issue #2410, opened 2021-07-10
using `useSpawn` in @xstate/react@>=1.5.0 requires at least xstate@v4.22


karfau:
**Description**
`useSpawn` tries to import from `xstate/[src|lib]/behaviors` which seems to only exist since v4.22.

**Expected Result**
Not sure, ideas:
- correct version range for 1.5.2 version in 
```json5
"peerDependencies": {
  //...,
  "xstate": ""
}
```
- conditionally check if the behaviors import is required/are available?
- update changelog/release notes of 1.5.0 and 1.5.1 regarding this issue
- ???

**Actual Result**
With an earlier version of xstate there are errors when trying to resolve that "xstate/src/behaviors"(1.5.0) or `xstate/lib/behaviors`(1.5.1).

**Reproduction**

https://codesandbox.io/s/elegant-wind-08c99?file=/src/index.js

**Additional context**

davidkpiano:
cc. @Andarist

Andarist:
See the closing comment below the PR that was meant to fix this in the past: https://github.com/statelyai/xstate/pull/2411#issuecomment-1018362868

PR #2411, opened 2021-07-12
Fix compat issue between useSpawn in @xstate/react and XState version

Andarist:
fixes https://github.com/statelyai/xstate/issues/2410

⚠️ this is not rly an ideal solution because we'll end up with the same problem when any user updates `@xstate/react` without updating `xstate`. The best solution would be to move `spawnBehavior` to a new package (`@xstate/private-do-not-use` or something) and add it as a dependency to both `xstate` and `@xstate/react`. This is the only solution that I can think of that would be compatible with any combination of upgrades. Thoughts?

Moral of the story: peer deps are hard 😬

### DIFF
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -9,7 +9,7 @@ import {
 } from './types';
 import { isMachine, mapContext, toInvokeSource } from './utils';
 import * as serviceScope from './serviceScope';
-import { ActorRef, BaseActorRef } from '.';
+import { ActorRef, BaseActorRef } from './types';
 
 export interface Actor<
   TContext = any,
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -75,4 +75,3 @@ export {
 };
 
 export * from './types';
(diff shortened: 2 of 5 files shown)


davidkpiano:
Hmm, would it be better to just copy the implementation of `spawnBehavior` in `@xstate/react`?

Andarist:
@davidkpiano in this instance - yeah, this would work as well, at the expense of the increased bundlesize. So in a sense, this solution doesn't scale well - we might not need this to scale though (just yet). 

I've pushed out the proposed changes - notice how this can very often create a domino effect as we need to copy additional utils and types that were only introduced recently. Let me know what do you think.

Andarist:
We didn't get that many issues about this and affected users just updated their XState versions to overcome the problem. This was a problem at a time but it shouldn't really be a problem now. With our new [Semver Policy](https://github.com/statelyai/xstate#packages-1) this wouldn't be a classified as a problem today since we take liberty to update the peer dep range in packages like `@xstate/react`

PR #2422, opened 2021-07-16
[core] Remove `context` from `StateNodeConfig` interface

davidkpiano:
The `context` property has been removed from `StateNodeConfig`, as it has never been allowed, nor has it ever done anything. The previous typing was unsafe and allowed `context` to be specified on nested state nodes:

```ts
createMachine({
  context: {
    /* ... */
  }, // ✅ This is allowed
  initial: 'inner',
  states: {
    inner: {
      context: {
        /* ... */
      } // ❌ This will no longer compile
    }
  }
});
```


packages/core/src/StateNode.ts
```diff
@@ -291,7 +292,9 @@ class StateNode<
      */
     private _context:
       | Readonly<TContext>
-      | (() => Readonly<TContext>) = config.context as any // TODO: this is unsafe, but we're removing it in v5 anyway
+      | (() => Readonly<TContext>) = ('context' in config
+      ? (config as StateMachine<TContext, any, TEvent>).context
```
Andarist, line 296: This has tripped me a little bit. I know it will be a subject of a refactor in v5 so it doesn't matter - just want to leave a short comment for posterity.

The `context` has been removed from `StateNodeConfig` so I was wondering how this cast has produced a type that had a `context` property. `StateMachine` extends the `StateNode` class (and not the `StateNodeConfig` interface) which still has a `context` property:
https://github.com/statelyai/xstate/blob/ce5a6295e08e0d356cff88f11ab19af19c3962c6/packages/core/src/StateNode.ts#L494

PR #2428, opened 2021-07-18
[v5] Consistent renaming: options & definition

davidkpiano:


packages/core/src/StateMachine.ts
```diff
@@ -96,7 +96,7 @@ export class StateMachine<
     options?: Partial<MachineImplementations<TContext, TEvent>>
```
Andarist, line 96: This argument should also be renamed and `createDefaultOptions` should be removed. `createDefaultOptions` could be renamed to `createDefaultImplementations` but I feel like it kinda doesn't really make sense to have "default implementations". Perhaps we should just do this:
```ts
this.implementations = {
  actors: implementations.actors || {},
  actions: implementations.actions || {},
  delays: implementations.delays || {},
  guards: implementations.guards || {}
}
```

davidkpiano:
Superseded by https://github.com/statelyai/xstate/pull/4080

PR #2430, opened 2021-07-18
[@xstate/react] Ensure `asEffect*` works with `preserveActionOrder: true`

davidkpiano:
This PR refactors effects in `@xstate/react` to use an object instead of a function, so that it works with `preserveActionOrder: true` and also in v5 (currently broken, which this change will fix).

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -382,11 +382,11 @@ class StateNode<
     // TODO: deprecate (entry)
     this.onEntry = toArray(
       this.config.entry || this.config.onEntry
-    ).map((action) => toActionObject(action));
+    ).map((action) => toActionObject(action, this.machine.options.actions));
     // TODO: deprecate (exit)
     this.onExit = toArray(
       this.config.exit || this.config.onExit
-    ).map((action) => toActionObject(action));
+    ).map((action) => toActionObject(action, this.machine.options.actions));
     this.meta = this.config.meta;
     this.doneData =
       this.type === 'final'
@@ -1858,7 +1858,10 @@ class StateNode<
 
     const transition = {
       ...transitionConfig,
-      actions: toActionObjects(toArray(transitionConfig.actions)),
+      actions: toActionObjects(
+        toArray(transitionConfig.actions),
+        this.machine.options.actions
+      ),
       cond: toGuard(transitionConfig.cond, guards),
       target,
       source: this as any,
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -652,6 +652,11 @@ export function resolveActions<TContext, TEvent extends EventObject>(
               return [];
             }
 
+            const actionObjects = toActionObjects(
+              toArray(matchedActions),
+              machine.options.actions
+            );
+
             const [
               resolvedActionsFromChoose,
               resolvedContextFromChoose
@@ -660,7 +665,7 @@ export function resolveActions<TContext, TEvent extends EventObject>(
               currentState,
               updatedContext,
               _event,
-              toActionObjects(toArray(matchedActions), machine.options.actions),
+              actionObjects,
               preserveActionOrder
             );
             updatedContext = resolvedContextFromChoose;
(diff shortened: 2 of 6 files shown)


packages/core/src/actions.ts
```diff
@@ -713,7 +718,7 @@ export function resolveActions<TContext, TEvent extends EventObject>(
             if (exec && preservedContexts) {
               const contextIndex = preservedContexts.length - 1;
               resolvedActionObject.exec = (_ctx, ...args) => {
-                exec?.(preservedContexts[contextIndex], ...args);
+                return exec?.(preservedContexts[contextIndex], ...args);
```
davidkpiano, line 721: Important to return here because the `asEffect*` implementation in `@xstate/react` needs to know the return value

davidkpiano:
@Andarist This will also fix the v5 `next` branch.

davidkpiano:
I'll merge this in tomorrow to unblock the `next` branch, unless there are any concerns

packages/core/src/StateNode.ts
```diff
@@ -1858,7 +1858,10 @@ class StateNode<
 
     const transition = {
       ...transitionConfig,
-      actions: toActionObjects(toArray(transitionConfig.actions)),
+      actions: toActionObjects(
+        toArray(transitionConfig.actions),
+        this.machine.options.actions
+      ),
```
Andarist, line 1864: You can't resolve action objects here without breaking existing functionality. I've added a failing test case for this here: https://github.com/statelyai/xstate/commit/272ca85e2529432543197a19f667ed0ccfabe3d5

The same goes for entry & exit actions (those 2 changed lines a little above this one).

packages/core/src/actions.ts
Andarist: This will create yet another mismatch between `xstate` & `@xstate/react` versions (peer dependencies 😭  )

davidkpiano: Are you saying this shouldn't go in? Because this fixes `@xstate/react`. Previous versions of `xstate` worked, and this change "restores" the behavior to those previous versions.

packages/core/src/actions.ts
```diff
@@ -713,7 +718,7 @@ export function resolveActions<TContext, TEvent extends EventObject>(
             if (exec && preservedContexts) {
               const contextIndex = preservedContexts.length - 1;
               resolvedActionObject.exec = (_ctx, ...args) => {
```
Andarist, line 720: This is a dangerous mutation - it kinda works OK now (I wasn't able to find a compelling failing test case) but I would call this to be rather coincidental. 

However, in combination with this PR it leads to a memory leak and a broken behavior. I can't showcase easily the memory leak but basically, a function action gets converted to an `{ exec }` object, and then it gets mutated here (each time). So a new `exec` always wraps the previous which wraps the previous one which wraps the previous one and so on.

I was able to write down a failing test for the broken behavior though: https://github.com/statelyai/xstate/commit/c49e058e094c4d565b95668a6cbad5b609158d78 . To fix this you should follow the advice given here: https://github.com/statelyai/xstate/pull/2430/files#r675457307 . Just do not resolve action objects in the constructor and in the `formatTransition`.

This is quite a foot gun and how the whole action resolving works is a little bit convoluted for what it is. I think this is a place where the codebase could be refactored and cleaned up in the process - it's rather a v5 item though.

packages/core/src/actions.ts
Andarist: the problem that I'm seeing is that, according to your words, `@xstate/react` needs to know the return value - which in turn means that it starts to require a future release of XState, while it states that it works with `^4.11.0`:
https://github.com/statelyai/xstate/blob/10944fa6195a3146c7c120d7c3c5c5defeff4e72/packages/xstate-react/package.json#L50

It's a technically OK thing to do to **only** update `@xstate/react` without updating `xstate` - because that's a contract that we have defined, by using peer dependencies, version ranges, and so on. But, from what I understand, this is not true anymore here - one can have a functioning pair of `xstate` and `@xstate/react` and they might break their app by upgrading only `@xstate/react` (which should be OK to do).

So the question would be - is there a way not to break that contract here?

davidkpiano:
Closing - will fix in v5 instead, and readdress if this becomes a publicly reported issue.

PR #2436, opened 2021-07-19
Fixed an issue with states coming from `model.createMachine` resulting in contexts of type `any` after type refinements

### LINKED ISSUE #2423: Context type becomes `any` after `matches` conditional
**Description**
Context type becomes `any` after `matches` conditional

**Expected Result**
Context type maintained.

**Actual Result**
Context type becomes `any` after `matches` conditional

**Reproduction**
https://codesandbox.io/s/objective-noether-11jjk?file=/src/index.tsx

**Additional context**
Strict mode typescript, see codesandbox index.ts line 39 for the error.

Andarist:
fixes #2423 

Note that this PR contains 2 separate fixes for the reported issue. Any of them could be removed and the reported issue would still be fixed. I will comment on each one "inline"

### DIFF
--- a/.changeset/warm-spies-cry.md
+++ b/.changeset/warm-spies-cry.md
@@ -1,0 +1,12 @@
+---
+'xstate': patch
+---
+
+Fixed an issue with states coming from `model.createMachine` resulting in contexts of type `any` after type refinements such as here:
+
+```ts
+// `state.context` became `any` erroneously
+if (state.matches('inactive')) {
+  console.log(state.context.count);
+}
+```
--- a/.changeset/warm-spies-cry.md
+++ b/.changeset/warm-spies-cry.md
@@ -2,7 +2,7 @@
 'xstate': patch
 ---
 
-Fixed an issue where, when using `model.createMachine`, state context was incorrectly inferred as `any` after refinement with `.matches(...)`, e.g.
+Fixed an issue where, when using `model.createMachine`, state's context was incorrectly inferred as `any` after refinement with `.matches(...)`, e.g.
 
 ```ts
 // `state.context` became `any` erroneously
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -19,6 +19,12 @@ import { StateNode } from './StateNode';
 import { getMeta, nextEvents } from './stateUtils';
 import { initEvent } from './actions';
 
+type Equals<A1 extends any, A2 extends any> = (<A>() => A extends A2
+  ? 1
+  : 0) extends <A>() => A extends A1 ? 1 : 0
+  ? 1
+  : 0;
+
 export function stateValuesEqual(
   a: StateValue | undefined,
   b: StateValue | undefined
@@ -293,11 +299,13 @@ export class State<
   public matches<TSV extends TTypestate['value']>(
     parentStateValue: TSV
   ): this is State<
-    (TTypestate extends any
-      ? { value: TSV; context: any } extends TTypestate
-        ? TTypestate
-        : never
-      : never)['context'],
+    Equals<TTypestate, any> extends 1
+      ? TContext
+      : (TTypestate extends any
+          ? { value: TSV; context: any } extends TTypestate
+            ? TTypestate
+            : never
+          : never)['context'],
     TEvent,
     TStateSchema,
     TTypestate
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -19,12 +19,6 @@ import { StateNode } from './StateNode';
 import { getMeta, nextEvents } from './stateUtils';
 import { initEvent } from './actions';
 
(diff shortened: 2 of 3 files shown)


.changeset/warm-spies-cry.md
```diff
@@ -0,0 +1,12 @@
+---
+'xstate': patch
+---
+
+Fixed an issue with states coming from `model.createMachine` resulting in contexts of type `any` after type refinements such as here:
```
Andarist, line 5: I'm super tired already and I have a feeling that I've written some hard-to-understand gibberish here. If you have any ideas on how I could rephrase this I would appreciate the help.

packages/core/src/model.types.ts
```diff
@@ -32,7 +32,7 @@ export interface Model<
   createMachine: (
     config: MachineConfig<TContext, any, TEvent>,
     implementations?: Partial<MachineOptions<TContext, TEvent>>
-  ) => StateMachine<TContext, any, TEvent, any>;
+  ) => StateMachine<TContext, any, TEvent>;
```
Andarist, line 35: This resulted in the default type not being chosen here:
https://github.com/statelyai/xstate/blob/141c91cffd1d7c1ec2e82186834cb977b72fb4d4/packages/core/src/State.ts#L93
which in turn resulted in `any` being selected here:
https://github.com/statelyai/xstate/blob/68692c1d11df7a513e68d4ce888603c1ffe41552/packages/core/src/State.ts#L298

The alternative fix here would be to use the default type explicitly here. It would be more explicit but it's probably not worth it at the moment.

packages/core/src/State.ts
```diff
@@ -293,11 +299,13 @@ export class State<
   public matches<TSV extends TTypestate['value']>(
     parentStateValue: TSV
   ): this is State<
-    (TTypestate extends any
-      ? { value: TSV; context: any } extends TTypestate
-        ? TTypestate
-        : never
-      : never)['context'],
+    Equals<TTypestate, any> extends 1
```
Andarist, line 302: This is not strictly necessary because the other. simple, fix already fixes the reported issue.

This seems a little bit heavy (adding more local complexity here) but handles "more" scenarios. I feel like if the `TTypestate` is not configured then the most sensible approach is to prefer using `TContext`. However, this still doesn't handle `unknown` and `never` and if we do this then it would probably be the most sensible approach for all of those special types. 

I think I'm leaning towards removing this part of the fix for the time being. Thoughts?

packages/core/src/State.ts
```diff
@@ -19,6 +19,12 @@ import { StateNode } from './StateNode';
 import { getMeta, nextEvents } from './stateUtils';
 import { initEvent } from './actions';
 
+type Equals<A1 extends any, A2 extends any> = (<A>() => A extends A2
+  ? 1
+  : 0) extends <A>() => A extends A1 ? 1 : 0
+  ? 1
+  : 0;
```
davidkpiano, line 26: 😳

davidkpiano: Could this be `true` or `false` instead of `1` or `0`?

davidkpiano (approved):

.changeset/warm-spies-cry.md
johtso: Maybe something like?

Fixed issue where, when using `model.createMachine`, state context was incorrectly inferred as `any` after refinement with `.matches(..)`, for example:

packages/core/src/State.ts
Andarist: It could. It's just taken from [here](https://github.com/millsp/ts-toolbelt/blob/319e55123b9571d49f34eca3e5926e41ca73e0f3/sources/Any/Equals.ts) and this type collection is using the 1/0 pattern in all its types.

I'm still wondering if we should use this though, see the comment here: https://github.com/statelyai/xstate/pull/2436#discussion_r672664520

.changeset/warm-spies-cry.md
Andarist: Thank you very much for taking a stab at rewording this! Would you like to prepare this as a PR suggestion that I could just commit with a button click? The change would also be contributed to you that way which is nice :)

davidkpiano:
```suggestion
Fixed an issue where, when using `model.createMachine`, state context was incorrectly inferred as `any` after refinement with `.matches(...)`, such as in here:
```

Using @johtso's suggestion

packages/core/src/State.ts
davidkpiano: Yeah, let's remove this and revisit later 👍

packages/core/src/model.types.ts
Blitz2145: Thanks for digging into this and fixing it.

Andarist: No problem. That's my job 😍

PR #2442, opened 2021-07-21
Fix condition for selected filter class

annaghi:
Selected filter is not a finite state, it is stored in context instead.

The selected filter has border around it.

Before: no border
Screenshot from 2021-07-21 07-16-26

After: with border
Screenshot from 2021-07-21 07-16-44


Andarist:
Thank you!

PR #2446, opened 2021-07-21
XState CLI integration experiments

mattpocock:


davidkpiano:
Interesting! Regarding the package name, I was thinking of an `@xstate/cli` as being a package that contains CLI utilities _for_ machines, not as a prompt builder. So the naming here might be confusing.

packages/xstate-cli/src/cli.ts
```diff
@@ -0,0 +1,32 @@
+import prompts from 'prompts';
+import {
+  InvokeConfig,
+  SingleOrArray,
+  TransitionConfig
+} from 'xstate/src/types';
+
+export const makePromptService = <T extends string, TContext>(
```
Andarist, line 8: It's more like `makePromptServiceState`, or `makePromptServiceStateConfig`

packages/xstate-cli/src/cli.ts
```diff
@@ -0,0 +1,32 @@
...
+        >;
+    onError?:
+      | string
+      | SingleOrArray<
+          TransitionConfig<TContext, { type: string; data: prompts.Answers<T> }>
+        >;
+  }
+): InvokeConfig<TContext, any> => {
+  return {
+    src: async () => {
```
Andarist, line 24: It's so thin of an abstraction that I wonder if we should provide stuff like this. The implementation for this is literally 2 lines of code:

```js
const makeAsyncServiceState = (asyncFn, opts) => ({ src: asyncFn, ...opts });
const makePromptServiceState = (prompt, opts) =>
  makeAsyncServiceState(async () => prompts(prompt), opts);
```

This could be rather easily typed but I have skipped that for brevity.

It feels like something like this is very opinionated, and only integrates with a very specific API. Making more generalized utils that can be easily adapted seems to scale better. 🤔

PR #2480, opened 2021-07-30
Example to avoid `useActor` when `send` is needed

gepd:


docs/recipes/react.md
```diff
@@ -103,6 +103,29 @@ export const SomeComponent = (props) => {
...
+export const SomeComponent = (props) => {
+  const globalServices = useContext(GlobalStateContext);
+  const isLoggedIn = useSelector(globalServices.authService, loggedInSelector);
+  const send = globalServices.authService.send;
+
+  return <>
+      {isLoggedIn ? 'Logged In' : 'Logged Out'}
+      <button onClick={() => send('TOGGLE')}>Toggle</button>
+   </>;
+};
```
Andarist, line 126: Maybe something like this would demonstrate the pattern a little bit better? The problem with previous snippet was that a `TOGGLE` event was sent to the `authService` - which conceptually looked a little bit broken
```suggestion
export const SomeComponent = (props) => {
  const globalServices = useContext(GlobalStateContext);
  const isLoggedIn = useSelector(globalServices.authService, loggedInSelector);
  const send = globalServices.authService.send;

  return <>
      {isLoggedIn && <button type="button" onClick={() => send('LOG_OUT')}>Logout</button>}
   </>;
};
```

docs/recipes/react.md
```diff
@@ -103,6 +103,29 @@ export const SomeComponent = (props) => {
 };
 ```
 
+If you need to use `send` in the same component to transition use `globalServices`
```
davidkpiano, line 106:
```suggestion
If you need to send an event in the component that consumes a service, you can use the `service.send(...)` method directly:
```

docs/recipes/react.md
```diff
@@ -103,6 +103,29 @@ export const SomeComponent = (props) => {
...
+import { useSelector } from '@xstate/react';
+
+const loggedInSelector = (state) => {
+  return state.matches('loggedIn');
+};
+
+export const SomeComponent = (props) => {
+  const globalServices = useContext(GlobalStateContext);
+  const isLoggedIn = useSelector(globalServices.authService, loggedInSelector);
+  const send = globalServices.authService.send;
```
davidkpiano, line 120:
```suggestion
  // Get `send()` method from a service
  const { send } = globalServices.authService;
```

docs/recipes/react.md
gepd: You're totally right! I've push the commit with your sugestion

gepd:
@davidkpiano @Andarist All suggestion commited!

davidkpiano (approved):

mattpocock:
@davidkpiano @Andarist I think we should revert this, it's a duplicate of the section below it.

mattpocock:
https://xstate.js.org/docs/recipes/react.html#global-state-react-context

PR #2483, opened 2021-07-31
Let generics be inferred for `assign()` in fsm

tajakobsen:
The need to pass in the Context as a generic parameter in `assign()` (when using `PropertyAssigner`) feels redundant when we already give the Context to `createMachine` trough `StateMachine.Config`.

This commit changes `assign()` so that the type of the `assignment` parameter in `assign()` is inferred, and then `TContext` is also inferred, and so we don't need an explicit generic parameter when using `PropertyAssigner`.

I also created a new type `Assignment` which is the union of `Assigner` and `PropertyAssigner` to make the new code in `assign` a little bit easier to read.

---

These changes are only to the types, and should not affect the outputted JS-code at all.

There are some new blank lines inserted in *types.ts*. Those come from prettier, so I just left them in.

### DIFF
--- a/.changeset/wicked-chicken-chew.md
+++ b/.changeset/wicked-chicken-chew.md
@@ -1,0 +1,5 @@
+---
+'@xstate/fsm': patch
+---
+
+Changed the type of `assign()` so that an explicit generic parameter is not needed for input using `PropertyAssigner` any more.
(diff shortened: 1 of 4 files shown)


tajakobsen:
@Andarist Take a look at the changes to `fsm.test.ts` first.

PR #2491, opened 2021-08-03
Remove `.toString()` method on action objects

davidkpiano:
This method was undocumented and unused (except incorrectly in tests), and removing this improves performance.

Fixes #2488

### DIFF
--- a/.changeset/dry-kangaroos-roll.md
+++ b/.changeset/dry-kangaroos-roll.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-The custom `.toString()` method on action objects is now removed. See #2488
+The custom `.toString()` method on action objects is now removed which improves performance in larger applications (see [#2488](https://github.com/statelyai/xstate/discussions/2488) for more context).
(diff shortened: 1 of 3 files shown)


.changeset/dry-kangaroos-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The custom `.toString()` method on action objects is now removed. See #2488
```
Andarist, line 5:
```suggestion
The custom `.toString()` method on action objects is now removed which improves performance in larger applications (see [#2488](https://github.com/statelyai/xstate/discussions/2488) for more context).
```

Andarist:
We should also create a ticket for v5 to consider removing a very similar `Object.defineProperty` call in the `State` class: https://github.com/statelyai/xstate/blob/fb7ea97465dfba0b7ef17edbf327c7c21848c7e8/packages/core/src/State.ts#L253-L257

Andarist (approved):

PR #2492, opened 2021-08-05
Fixed a minor issue with sometimes sending `undefined` state to the inspector

Andarist:
When interpreting machines, when the inspect machine is already in the `connected` state, this makes this `service.event` sent to the inspector right within the `.start()` call:
https://github.com/statelyai/xstate/blob/fb7ea97465dfba0b7ef17edbf327c7c21848c7e8/packages/xstate-inspect/src/browser.ts#L171-L175

but the `service._state` has never been assigned yet up to this point so it's `undefined`. That gets forwarded to the inspector and throws here on `JSON.parse(undefined)`:
https://github.com/statelyai/xstate/blob/fb7ea97465dfba0b7ef17edbf327c7c21848c7e8/packages/xstate-inspect/src/utils.ts#L38

### DIFF
--- a/packages/xstate-inspect/src/browser.ts
+++ b/packages/xstate-inspect/src/browser.ts
@@ -168,6 +168,10 @@ export function inspect(
     };
 
     service.subscribe((state) => {
+      // filter out synchronous notification from within `.start()` call when the `service.state` has not yet been assigned
+      if (state === undefined) {
+        return;
+      }
       inspectService.send({
         type: 'service.state',
         state: stringify(state),
--- a/.changeset/nasty-ways-trade.md
+++ b/.changeset/nasty-ways-trade.md
@@ -1,0 +1,5 @@
+---
+'@xstate/inspect': patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/xstate-inspect/src/browser.ts
```diff
@@ -168,6 +168,10 @@ export function inspect(
     };
 
     service.subscribe((state) => {
+      // filter out synchronous notification from within `.start()` call when the `service.state` has not yet been assigned
+      if (state === undefined) {
+        return;
+      }
```
Andarist, line 174: A more semantic fix would be to add a local variable like `skipSyncNotification = true`, filter based on it, and assign it to `false` right after `service.subscribe` returns. 

It would look quite weird though so I've kept the initial condition that I've written - let me know if you'd like the other one to get introduced. This kinda showcases how calling the listener synchronously on the subscription time is often problematic where the intention is to react to changes.

davidkpiano: This fix is fine for now; we can reinvestigate later

davidkpiano (approved):

PR #2496, opened 2021-08-06
Add utility-type EmittedFrom<T> that extracts Emitted type from any type which can emit data

### LINKED ISSUE #2493: After "state" deprecation in ActorRefFrom, there is no easy way to get the type of an ActorRef
**Description**
Previously, when I needed a way to extract the context from an ActorRef of a generic machine, I could do something like this:
```ts
interface MyMachineActor<T> extends ActorRefFrom<MyMachine<T>> {}

// but now I get
// 'state' is depricated
const myMachineSelector<T, R>(selector: (state MyMachineActor<T>['state']) => R) { ... }
```

**Expected Result**
We probably can provide a utility type like:
```ts
type EmittedFrom<T> = T extends ActorRef<any, infer TEmitted> 
  ? TEmitted 
  : T extends Behavior<any, infer TEmitted>
  ? TEmitted
  : never 
```

Used like this:
```ts
const myMachineSelector<T, R>(selector: (state EmittedFrom<MyMachineActor<T>>) => R) { ... }
```

Let me know what you think and if I should open a PR.
Thanks!

VanTanev:
Closes #2493

### DIFF
--- a/.changeset/quiet-islands-sit.md
+++ b/.changeset/quiet-islands-sit.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Add utility-type EmittedFrom<T> that extracts Emitted type from any type which can emit data
--- a/.changeset/quiet-islands-sit.md
+++ b/.changeset/quiet-islands-sit.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Add utility-type EmittedFrom<T> that extracts Emitted type from any type which can emit data
+Add utility type `EmittedFrom<T>` that extracts `Emitted` type from any type which can emit data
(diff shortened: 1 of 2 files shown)


.changeset/quiet-islands-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Add utility-type EmittedFrom<T> that extracts Emitted type from any type which can emit data
```
davidkpiano, line 5:
```suggestion
Add utility type `EmittedFrom<T>` that extracts `Emitted` type from any type which can emit data
```

davidkpiano (approved):

PR #2497, opened 2021-08-06
[examples] Add React Todo MVC

annaghi:
This PR is based on https://codesandbox.io/s/xstate-todomvc-33wr94qv1?from-embed with couple of modifications listed above, if they are fine?

Project created with
```sh
pnpx create-vite todo-mvc-react --template react-ts
```

Temporary CodeSandbox link:
https://codesandbox.io/embed/github/annaghi/xstate/tree/examples/todo-mvc-react/examples/todo-mvc-react


examples/todo-mvc-react/src/App.jsx
```diff
@@ -0,0 +1,25 @@
...
+        <p>
+          Template by <a href="http://sindresorhus.com">Sindre Sorhus</a>
+        </p>
+        <p>
+          Created by <a href="http://todomvc.com">you</a>
+        </p>
+        <p>
+          Part of <a href="http://todomvc.com">TodoMVC</a>
+        </p>
+      </footer>
```
annaghi, line 20: Moved `footer` from `index.html` to `App.jsx` in order to avoid showing it up for a moment before the App renders.

examples/todo-mvc-react/src/Todos.jsx
```diff
@@ -0,0 +1,168 @@
...
+    todo: 'Learn state machines',
+    todos: (() => {
+      try {
+        return JSON.parse(localStorage.getItem('todos-xstate')) || [];
+      } catch (e) {
+        console.error(e);
+        return [];
+      }
+    })(),
+    filter: 'all'
```
annaghi, line 43: Store `filter` in LocalStorage too.

examples/todo-mvc-react/src/Todos.jsx
```diff
@@ -0,0 +1,168 @@
...
+  const mark = !allCompleted ? 'completed' : 'active';
+  const markEvent = `MARK.${mark}`;
+  const filteredTodos = filterTodos(filter, todos);
+
+  const inputRef = useRef(null);
+  useEffect(() => {
+    if (todos.length === 0) {
+      inputRef.current.focus();
+    }
+  }, [todos]);
```
annaghi, line 73: If there is no todo, then set focus on input field.

examples/todo-mvc-react/src/Todos.jsx
```diff
@@ -0,0 +1,168 @@
...
+          }
+          value={todo}
+        />
+      </header>
+
+      {!!todos.length && (
+        <>
+          <section className="main">
+            <input
+              id="toggle-all"
```
annaghi, line 100: Moved `section` inside the condition, because of the original example: https://todomvc.com/
It has sense not displaying the `toggle-all` when there is no todo.

examples/todo-mvc-react/src/todo.machine.ts
```diff
@@ -0,0 +1,135 @@
...
+      BLUR: () => ({}),
+      CANCEL: () => ({})
+    }
+  }
+);
+
+type TodoEvents = EventFrom<typeof todoModel>[keyof EventFrom<
+  typeof todoModel
+>];
+
```
annaghi, line 29: Is this probably temporary until `xstate/lib/model` settles down?

It was needed with `entry` transitions. I checked `exit` and `always` those require this type parameter too. (Did not tried delay transitions)

examples/todo-mvc-react/src/todo.machine.ts
```diff
@@ -0,0 +1,135 @@
...
+              actions: [todoModel.assign({ completed: false }), 'commit']
+            },
+            EDIT: {
+              target: 'editing',
+              actions: 'focusInput'
+            }
+          }
+        },
+        editing: {
+          entry: todoModel.assign<TodoEvents>({
```
annaghi, line 82: Using generic type here ^

examples/todo-mvc-react/src/todos.machine.ts
```diff
@@ -0,0 +1,134 @@
...
+      'MARK.active': () => ({}),
+      CLEAR_COMPLETED: () => ({})
+    }
+  }
+);
+
+type TodosEvents = EventFrom<typeof todosModel>[keyof EventFrom<
+  typeof todosModel
+>];
+
```
annaghi, line 45: Same comment here with types.

examples/todo-mvc-react/src/todos.machine.ts
```diff
@@ -0,0 +1,134 @@
...
+      actions: (context) => {
+        context.todos.forEach((todo) => todo.ref.send('SET_ACTIVE'));
+      }
+    },
+    CLEAR_COMPLETED: {
+      actions: [
+        todosModel.assign({
+          todos: (context) => context.todos.filter((todo) => !todo.completed)
+        }),
+        'persist'
```
annaghi, line 130: Added a `persist` action here.

examples/todo-mvc-react/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "target": "ESNext",
+    "lib": ["DOM", "DOM.Iterable", "ESNext"],
+    "allowJs": false,
+    "skipLibCheck": true,
```
annaghi, line 6: A fresh project generated with `create-vite` has `skipLibCheck: false`, and it worked when outside of the `xstate` repo.
However there were type errors when running `pnpm run build` inside this repo, so I had to change `skipLibCheck` to `true`.

annaghi:
@davidkpiano What do you think about renaming `template-{js|ts}` to `template-vanilla-{js|ts}`?

Benefits
- it is popular that `vanilla` means `no-framework`, e.g. https://github.com/vitejs/vite/tree/main/packages/create-vite or https://github.com/krausest/js-framework-benchmark/tree/master/frameworks/keyed
- in lists sorted alphabetically they will appear next to each other https://github.com/statelyai/xstate/tree/main/examples
- the explanation can be removed here: https://github.com/statelyai/xstate#templates
- if in the future the templates will be moved into their own folder, it might be easy to remove the `template-` prefix in the project names, and keeping `vanilla-{js|ts}` will be self explanatory

Drawbacks
- need to refactor the name in this repo (or maybe outside of this repo too if references were used externally)

VanTanev:
Just a note, this example does not stop the spawned todo machines properly.

They need to be stopped like this:

https://codesandbox.io/s/xstate-todomvc-fixed-todo-actor-cleanup-and-some-persist-bugs-yqhqf?file=/todosMachine.js:1598-2061

```ts
  "TODO.DELETE": {
      actions: [
        actions.pure((context, event) => {
          const assignAction = assign({
            todos: context.todos.filter((todo) => todo.id !== event.id)
          });

          const stopActions = context.todos
            .filter((todo) => todo.id === event.id)
            .map((deletedTodo) => actions.stop(deletedTodo.ref));
          return [assignAction, ...stopActions];
        }),
        "persist"
      ]
    },
```

Basically, anywhere where the list of todos is filtered down you also need a stop action, as in the above example. The stop action needs the list of machines to stop before the assign action executes. This can be done either with `actions.pure()` as shown above, or with machine-level config `preserveActionOrder: true` and doing the assign after the stop - @davidkpiano can chime in which approach is preferred?

annaghi:
Thank you @VanTanev for the fix and warning.

I was not careful enough and did not check if the todo machines across the repo are the same or they are not!

I reused the machines from here:
https://github.com/statelyai/xstate/blob/main/examples/todo-mvc-vue/src/
instead of here:
https://github.com/statelyai/xstate/tree/main/docs/sandboxes/todomvc

Your comment revealed this inconsistency, so I take my PR back to draft, and will use `actually` the machines from here:
https://github.com/statelyai/xstate/tree/main/docs/sandboxes/todomvc

davidkpiano:
> @davidkpiano What do you think about renaming `template-{js|ts}` to `template-vanilla-{js|ts}`?

That's a good idea; let's do that.

VanTanev:
This was merged without addressing the buggy behavior?

davidkpiano:
> This was merged without addressing the buggy behavior?

Can you make a follow-up PR fixing it? Or I can do it.

PR #2501, opened 2021-08-08
[@xstate/react] Improve deprecation warning for send payloads

### LINKED ISSUE #2412: [@xstate/react] send function from useActor
**Description**
The `send` function returned from the `useActor` hook has not the same API as the send function returned from `useService`. Since the docs recommend replacing the now deprecated `useService` with `useActor` their accepted arguments should be equivalent.

**Expected Result**
```javascript
const [, send] = useActor(service)
send({ type: 'MY_EVENT', data: 42 }) // works
send('MY_EVENT', { data: 42 }) // should work too, because this format is accepted by
                               // the send function returned from useService
```

**Actual Result**
```javascript
// in the state machine
on: {
  MY_EVENT: {
    actions: (ctx, e) => console.log(e.data)
  }
}

// in a React component
const [, send] = useActor(service)
send({ type: 'MY_EVENT', data: 42 }) // logs "42"
send('MY_EVENT', { data: 42 }) // logs "undefined"

```

**Reproduction**
https://codesandbox.io/s/hungry-sound-gmbny?file=/src/index.js

davidkpiano:
This PR adds warnings for unexpected arguments when migrating from `useService(...)` to `useActor(...)`.

The warnings look like this:

CleanShot 2021-08-08 at 17 30 46

Documentation is also updated. Closes #2412

### DIFF
--- a/docs/packages/xstate-react/index.md
+++ b/docs/packages/xstate-react/index.md
@@ -90,7 +90,7 @@ A [React hook](https://reactjs.org/hooks) that interprets the given `machine` an
 
 In the next major version, `useService(service)` will be replaced with `useActor(service)`. Prefer using the `useActor(service)` hook for services instead, since services are also actors.
 
-Also, keep in mind that only a single argument (the event object) can be sent to `send(eventObject)` from `useActor(...)`. When migrating to `useActor(...)`, refactor `send(...)` calls to use only a single event object:
+Also, keep in mind that only a single argument (the event object) can be sent to `send(eventObject)` from `useActor(...)`. When migrating to `useActor(...)`, refactor `send(...)` calls to use only a single argument (event object or event type string):
 
 ```diff
 const [state, send] = useActor(service);
--- a/docs/packages/xstate-react/index.md
+++ b/docs/packages/xstate-react/index.md
@@ -90,6 +90,15 @@ A [React hook](https://reactjs.org/hooks) that interprets the given `machine` an
 
 In the next major version, `useService(service)` will be replaced with `useActor(service)`. Prefer using the `useActor(service)` hook for services instead, since services are also actors.
 
+Also, keep in mind that only a single argument (the event object) can be sent to `send(eventObject)` from `useActor(...)`. When migrating to `useActor(...)`, refactor `send(...)` calls to use only a single event object:
+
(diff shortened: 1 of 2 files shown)


docs/packages/xstate-react/index.md
```diff
@@ -90,6 +90,15 @@ A [React hook](https://reactjs.org/hooks) that interprets the given `machine` an
 
 In the next major version, `useService(service)` will be replaced with `useActor(service)`. Prefer using the `useActor(service)` hook for services instead, since services are also actors.
 
+Also, keep in mind that only a single argument (the event object) can be sent to `send(eventObject)` from `useActor(...)`. When migrating to `useActor(...)`, refactor `send(...)` calls to use only a single event object:
```
Andarist, line 93: From the used wording I would think that only event objects are supported but in reality we also support plain strings for "simple" events

packages/xstate-react/src/useActor.ts
```diff
@@ -52,7 +52,17 @@ export function useActor(
...
-  const send: Sender<EventObject> = useConstant(() => (event) => {
+  const send: Sender<EventObject> = useConstant(() => (...args) => {
+    const event = args[0];
+
+    if (process.env.NODE_ENV !== 'production' && args.length > 1) {
+      console.warn(
+        `Unexpected argument: ${JSON.stringify(
+          (args as any)[1]
+        )}. Only a single event object can be sent to actor send() functions.`
+      );
```
Andarist, line 63: Especially the "Unexpected argument" part sounds like something that would be included in a hard error whereas this, for now, is just a deprecation warning. Maybe a little bit more mild wording could be used here?

Andarist (approved):

packages/xstate-react/src/useActor.ts
davidkpiano:
```suggestion
      console.warn(
        `Unexpected payload: ${JSON.stringify(
          (args as any)[1]
        )}. Only a single event object can be sent to actor send() functions.`
      );
```

docs/packages/xstate-react/index.md
Andarist: Works for me!

PR #2502, opened 2021-08-09
Improve `model.assign` inference in some cases by using separate overloads

Andarist:
This is related to this weird case: https://www.loom.com/share/51f67097b6ca43b790ce5c215b2a63ae

I've found out that by putting less stress on the inference algorithm we can sometimes achieve better results. Overloads in our codebase are sometimes a little bit hard to get right - there are sensitive to declaration order and can lead to cryptic errors. I think this specific case is a good candidate for using overloads though. Unless we can think of a scenario when the first overload shouldn't be preferred and when the second one should be chosen instead then this should be OK to go in.

EDIT:// disregard what I have just said. If anything - overloads can make it harder to infer stuff correctly.

### DIFF
--- a/.changeset/proud-ghosts-hear.md
+++ b/.changeset/proud-ghosts-hear.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Improve `model.assign` inference in some cases by using separate overloads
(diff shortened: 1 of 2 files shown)


davidkpiano:
Seems like this is causing test failures

Andarist:
The CI has failed with errors that I have not seen locally (not unless I've restarted my TS server 🙈 ). I'm going to close this because even if this fixes the reported problem it breaks inference in a lot of other situations 😬 I'm not sure how to satisfy both

PR #2504, opened 2021-08-09
Call `Inspector`'s `subscribe` callback immediately with the current state at the subscription time

Andarist:
As requested last week. I wonder though if this really should be exposed like that - this is exposing the `inspectMachine`'s state as part of the public API. It could mean that any change to it could be considered a breaking change.

### DIFF
--- a/packages/xstate-inspect/src/browser.ts
+++ b/packages/xstate-inspect/src/browser.ts
@@ -203,6 +203,7 @@ export function inspect(
       const observer = toObserver(next, onError, onComplete);
 
       listeners.add(observer);
+      observer.next(inspectService.state);
 
       return {
         unsubscribe: () => {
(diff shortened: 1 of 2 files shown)


packages/xstate-inspect/src/browser.ts
```diff
@@ -203,6 +203,7 @@ export function inspect(
       const observer = toObserver(next, onError, onComplete);
 
       listeners.add(observer);
+      observer.next(inspectService.state);
```
Andarist, line 206: an alternative fix would be to just use the existing `inspectService.subscribe` as this proxying doesn't give us a lot:
https://github.com/statelyai/xstate/blob/5010717d216fe74b36826b193f985b337b19f71a/packages/xstate-inspect/src/browser.ts#L92-L94

davidkpiano: I think your current fix makes the most sense

davidkpiano (approved):

PR #2508, opened 2021-08-11
[core] Fix race condition when stopping child of initializing parent machine

### LINKED ISSUE #2507: Cannot read property 'children' of undefined when spawned machine is at the final state
**Description**
When you have multiple levels of machines in context by spawning machines, if a grand child machine's state is `final`, you get an error saying `Cannot read property 'children' of undefined.`

**Expected Result**
This error should not happen.

**Actual Result**
`Cannot read property 'children' of undefined` error occurs.

**Reproduction**

parent -> child -> grand child (final)
https://codesandbox.io/s/reverent-austin-uztit

parent -> child (final) this one won't throw the error.
https://codesandbox.io/s/empty-paper-s2l4u

davidkpiano:
This PR fixes #2507 

The problem was a classic race condition:

1. Child machine determining initial state (`this.state` is still `undefined`)
2. Grandchild machine immediately stopping, notifying child machine
3. Child machine trying to remove grandchild from `this.state.children` (remember: `this.state` is still `undefined` at this point) ❌ 

The fix is to _schedule_ the removal of the child, so that it can first complete determining its initial state, so that `this.state` is defined, and then try to remove the child afterward.

### DIFF
--- a/.changeset/large-crabs-serve.md
+++ b/.changeset/large-crabs-serve.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-A race condition occurred when a grandchild service is immediately stopped and the child service tried to remove it from its undefined state; this has been fixed, and the race condition no longer occurs. See https://github.com/statelyai/xstate/issues/2507 for details.
+A race condition occurred when a child service is immediately stopped and the parent service tried to remove it from its undefined state; this has been fixed, and the race condition no longer occurs. See https://github.com/statelyai/xstate/issues/2507 for details.
--- a/.changeset/large-crabs-serve.md
+++ b/.changeset/large-crabs-serve.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+A race condition occurred when a grandchild service is immediately stopped and the child service tried to remove it from its undefined state; this has been fixed, and the race condition no longer occurs. See https://github.com/statelyai/xstate/issues/2507 for details.
--- a/.changeset/large-crabs-serve.md
+++ b/.changeset/large-crabs-serve.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-A race condition occurred when a child service is immediately stopped and the parent service tried to remove it from its undefined state; this has been fixed, and the race condition no longer occurs. See https://github.com/statelyai/xstate/issues/2507 for details.
+A race condition occurred when a child service is immediately stopped and the parent service tried to remove it from its undefined state (during its own initialization). This has been fixed, and the race condition no longer occurs. See [this issue](https://github.com/statelyai/xstate/issues/2507) for details.
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -1013,8 +1013,8 @@ export class Interpreter<
 
     childService
       .onDone((doneEvent) => {
-        this.send(toSCXMLEvent(doneEvent as any, { origin: childService.id }));
         this.removeChild(childService.id);
+        this.send(toSCXMLEvent(doneEvent as any, { origin: childService.id }));
       })
       .start();
 
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -932,11 +932,12 @@ export class Interpreter<
   }
 
   private removeChild(childId: string): void {
-    this.scheduler.schedule(() => {
-      this.children.delete(childId);
-      this.forwardTo.delete(childId);
-      delete this.state.children[childId];
-    });
+    this.children.delete(childId);
+    this.forwardTo.delete(childId);
+
+    // this.state might not exist at the time this is called,
+    // such as when a child is added then removed while initializing the state
+    delete this.state?.children[childId];
   }
 
   private stopChild(childId: string): void {
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -1011,8 +1011,12 @@ export class Interpreter<
 
     childService
       .onDone((doneEvent) => {
-        this.removeChild(childService.id);
-        this.send(toSCXMLEvent(doneEvent as any, { origin: childService.id }));
+        this.scheduler.schedule(() => {
+          this.removeChild(childService.id);
+          this.send(
+            toSCXMLEvent(doneEvent as any, { origin: childService.id })
+          );
+        });
       })
       .start();
 
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -932,10 +932,11 @@ export class Interpreter<
   }
 
   private removeChild(childId: string): void {
-    this.children.delete(childId);
-    this.forwardTo.delete(childId);
-
-    delete this.state.children[childId];
+    this.scheduler.schedule(() => {
+      this.children.delete(childId);
+      this.forwardTo.delete(childId);
+      delete this.state.children[childId];
+    });
   }
 
   private stopChild(childId: string): void {
@@ -1011,12 +1012,8 @@ export class Interpreter<
 
     childService
       .onDone((doneEvent) => {
-        this.scheduler.schedule(() => {
-          this.removeChild(childService.id);
-          this.send(
-            toSCXMLEvent(doneEvent as any, { origin: childService.id })
-          );
-        });
+        this.send(toSCXMLEvent(doneEvent as any, { origin: childService.id }));
+        this.removeChild(childService.id);
       })
       .start();
 
--- a/packages/core/test/actor.test.ts
+++ b/packages/core/test/actor.test.ts
@@ -1107,8 +1107,8 @@ describe('actors', () => {
   });
 
   // https://github.com/statelyai/xstate/issues/2507
-  it('should not attempt to clean up children before stopping', () => {
-    const grandChildMachine = createMachine({
+  it('should not crash on sync child completion before stopping', () => {
+    const childMachine = createMachine({
       initial: 'idle',
       states: {
         idle: {
@@ -1123,42 +1123,15 @@ describe('actors', () => {
         }
       }
     });
-    const childMachine = createMachine<any>(
-      {
-        initial: 'idle',
-        context: {
-          child: null
-        },
-        states: {
-          idle: {
-            entry: ['setup']
-          }
-        }
-      },
-      {
-        actions: {
-          setup: assign({
-            child: () => spawn(grandChildMachine)
-          })
-        }
-      }
-    );
 
-    const parentMachine = createMachine(
+    const parentMachine = createMachine<{
+      child: ActorRefFrom<typeof childMachine> | null;
+    }>(
       {
-        initial: 'idle',
         context: {
           child: null
         },
-        states: {
-          idle: {
-            on: {
-              SETUP: {
-                actions: ['setup']
-              }
-            }
-          }
-        }
+        entry: 'setup'
       },
       {
         actions: {
@@ -1168,12 +1141,9 @@ describe('actors', () => {
         }
       }
     );
-
     const service = interpret(parentMachine);
-    service.start();
-
     expect(() => {
-      service.send('SETUP');
+      service.start();
     }).not.toThrow();
   });
 });
--- a/packages/core/test/actor.test.ts
+++ b/packages/core/test/actor.test.ts
@@ -1164,7 +1164,7 @@ describe('actors', () => {
     }).not.toThrow();
   });
 
-  it('should not crash on immediate spawned observable completion', () => {
+  it('should not crash on child observable sync completion during self-initialization', () => {
     const createEmptyObservable = (): any => ({
       subscribe(_next, _error, complete) {
         complete();
--- a/packages/core/test/actor.test.ts
+++ b/packages/core/test/actor.test.ts
@@ -1149,7 +1149,7 @@ describe('actors', () => {
 
   it('should not crash on immediate spawned promise completion', () => {
     const parentMachine = createMachine<{
-      child: ActorRef<any> | null;
+      child: ActorRef<never, any> | null;
     }>({
       context: {
         child: null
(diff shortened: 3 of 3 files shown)


packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1106,75 @@ describe('actors', () => {
...
+      }
+    );
+
+    const service = interpret(parentMachine);
+    service.start();
+
+    expect(() => {
+      service.send('SETUP');
+    }).not.toThrow();
+  });
```
Andarist, line 1179: A potentially simpler test case can be written for to validate this behavior:
```suggestion
  // https://github.com/statelyai/xstate/issues/2507
  it('should not attempt to clean up children before stopping', () => {
    const childMachine = createMachine({
      initial: 'idle',
      states: {
        idle: {
          always: [
            {
              target: 'stopped'
            }
          ]
        },
        stopped: {
          type: 'final'
        }
      }
    });

    const parentMachine = createMachine(
      {
        context: {
          child: null
        },
        entry: 'setup'
      },
      {
        actions: {
          setup: assign({
            child: () => spawn(childMachine)
          })
        }
      }
    );

    const service = interpret(parentMachine);

    expect(() => {
      service.start();
    }).not.toThrow();
  });
```

Andarist: It roughly tests for the same thing with just a parent and a child instead of using a 3 layer actor hierarchy

packages/core/src/interpreter.ts
```diff
@@ -1011,8 +1011,12 @@ export class Interpreter<
 
     childService
       .onDone((doneEvent) => {
-        this.removeChild(childService.id);
-        this.send(toSCXMLEvent(doneEvent as any, { origin: childService.id }));
+        this.scheduler.schedule(() => {
```
Andarist, line 1014: Any particular reason why this is scheduled now instead of being guarded in the `removeChild`? I can't really think of an edge-case to break it right now but scheduling can potentially introduce edge cases. The problem is that we don't know how many tasks and what kind of tasks are already in the queue - so I'm a little bit afraid that somehow one could create a bizarre situation in which this would crash once again or remove a child that should no longer be removed. Or that changing the timing of the `doneEvent` could change behavior in somebody's code.

Andarist: Also - the fix has only been applied for machines but observables, callbacks, behaviors etc can suffer from the very same problem. Guarding in the `removeChild` would fix all of those cases at once.

packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1106,75 @@ describe('actors', () => {
       })
       .start();
   });
+
+  // https://github.com/statelyai/xstate/issues/2507
+  it('should not attempt to clean up children before stopping', () => {
```
Andarist, line 1111: This IMHO describes a vague internal implementation detail. The actual user-like story would sound something like:
"should not crash on sync child completion during own setup"

.changeset/large-crabs-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+A race condition occurred when a grandchild service is immediately stopped and the child service tried to remove it from its undefined state; this has been fixed, and the race condition no longer occurs. See https://github.com/statelyai/xstate/issues/2507 for details.
```
Andarist, line 5: As mentioned in other comments - the fixed situation was not unique to setups with grandchildren etc so a little bit more general description would be more appropriate here.

packages/core/test/actor.test.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
Andarist, line 1:
```suggestion
```

Andarist:
I've realized now how `state` is mutated 😬 this probably should be refactored in v5 somehow. I think ideally the state object should be immutable.

packages/core/src/interpreter.ts
davidkpiano: It's supposed to be scheduled because we need to consider `.onDone(...)` to be an "event", and events need to be queued.

davidkpiano: Also, if we don't schedule this, there are other potential race conditions like this one, so we need to schedule regardless. I'll fix it for the others.

packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1105,79 @@ describe('actors', () => {
...
+      },
+      entry: assign({
+        child: () => spawn(EMPTY)
+      })
+    });
+    const service = interpret(parentMachine);
+    expect(() => {
+      service.start();
+    }).not.toThrow();
+  });
```
Andarist, line 1182: This actually still throws in the `removeChild` method - it's just not obvious due to how RxJS handles errors.

They wrap handlers:
https://github.com/ReactiveX/rxjs/blob/3d062a2b8c2e49efc92e1205ae926c7182745b8c/src/internal/Subscriber.ts#L179
catch them and report:
https://github.com/ReactiveX/rxjs/blob/3d062a2b8c2e49efc92e1205ae926c7182745b8c/src/internal/Subscriber.ts#L195-L202
where reporting is done in a timeout:
https://github.com/ReactiveX/rxjs/blob/3d062a2b8c2e49efc92e1205ae926c7182745b8c/src/internal/util/reportUnhandledError.ts#L13-L14

While RxJS is by far the most popular stream library we don't specify that XState only works with that stream library. I've adjusted this test to showcase how this still crashes with other observable implementations and also added a second one that doesn't pass even with RxJS. That's because `removeChild` throws and we never actually get to sending the done event to a service.

```suggestion
it('should not crash on immediate spawned observable completion', () => {
    const createEmptyObservable = (): any => ({
      subscribe(_next, _error, complete) {
        complete();
      }
    });
    const parentMachine = createMachine<{
      child: ActorRef<any> | null;
    }>({
      context: {
        child: null
      },
      entry: assign({
        child: () => spawn(createEmptyObservable())
      })
    });
    const service = interpret(parentMachine);
    expect(() => {
      service.start();
    }).not.toThrow();
  });

  it('should receive done event from an immediately completed observable when self-initializing', () => {
    const parentMachine = createMachine<{
      child: ActorRef<any> | null;
    }>({
      context: {
        child: null
      },
      entry: assign({
        child: () => spawn(EMPTY, 'myactor')
      }),
      initial: 'init',
      states: {
        init: {
          on: {
            'done.invoke.myactor': 'done'
          }
        },
        done: {}
      }
    });
    const service = interpret(parentMachine);

    service.start();

    expect(service.state.value).toBe('done');
  });
```

packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1105,79 @@ describe('actors', () => {
...
+      },
+      entry: assign({
+        child: () => spawn(new Promise((res) => res(null)))
+      })
+    });
+    const service = interpret(parentMachine);
+    expect(() => {
+      service.start();
+    }).not.toThrow();
+  });
```
Andarist, line 1165: Since `then` callbacks are always invoked asynchronously this couldn't really throw for this specific scenario. While completion might be immediate and sync - the notification to the parent isn't.

A better test (or an additional one) would be to use a "thenable" here that would actually notify the parent synchronously about its completion:
```suggestion
  it('should not crash on immediate spawned promise completion', () => {
    const parentMachine = createMachine<{
      child: ActorRef<any> | null;
    }>({
      context: {
        child: null
      },
      entry: assign({
        child: () => spawn({ then: (fn) => fn(null) } as any)
      })
    });
    const service = interpret(parentMachine);
    expect(() => {
      service.start();
    }).not.toThrow();
  });
```

Note that this particular test wouldn't fail, even with a thenable. That's only because we cast thenables to promises here:
https://github.com/statelyai/xstate/blob/8d7cbc36432c26fb33e485dbddbe41b784de3fc6/packages/core/src/interpreter.ts#L959

Without this cast this test would fail and this is something that we should have a test for.

.changeset/large-crabs-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+A race condition occurred when a child service is immediately stopped and the parent service tried to remove it from its undefined state; this has been fixed, and the race condition no longer occurs. See https://github.com/statelyai/xstate/issues/2507 for details.
```
Andarist, line 5:
```suggestion
'xstate': patch
---

A race condition occurred when a child service is immediately stopped and the parent service tried to remove it from its undefined state (during its own initialization). This has been fixed, and the race condition no longer occurs. See [this issue](https://github.com/statelyai/xstate/issues/2507) for details.
```

packages/core/src/interpreter.ts
Andarist: I agree that `this.send` should be scheduled here... but it already is. It already happens in the `send` method itself:
https://github.com/statelyai/xstate/blob/8d7cbc36432c26fb33e485dbddbe41b784de3fc6/packages/core/src/interpreter.ts#L593

I'm pretty sure that we shouldn't schedule this twice - it makes the code execution harder to understand and a little bit iffy. At the moment I see the scheduler as a way to express the external events queue and double scheduling breaks that mental model. Consider this scheduler state:
```
// initial~
[A, B]
// DONE, C arrives
[A, B, scheduled(DONE), C]
// we process 3 items and we end up with
[C, DONE]
```
As we may notice here - the items were processed a little bit out of order. I understand that **conceptually** it doesn't quite matter if we assume that actor communication is async and that done event can arrive at any time. However, this still complicates the mental model and execution path a little bit. I think that in v5 we actually might want to align the implementation closer to the mailbox concept - this actually have already happened in some parts of the codebase, and even on the v4 branch:
https://github.com/statelyai/xstate/blob/f135dfae27c43afcbf52c0960c11c2c4011fe597/packages/core/src/behaviors.ts#L116

So if we follow that in the future then that refactor would change the behavior a little bit - because in the example above DONE would get actually processed before C. I know this is really an edge case and hopefully, such a change wouldn't impact anyone but if we could avoid it then this would still IMHO be preferable.

I see how we might want to `removeChild` together with processing the done event - but then IMHO we should just schedule it and let `this.send` to schedule the event. This would avoid the double scheduling problem that got outlined above while processing both things right after another, as a group of 2 scheduled tasks.

Alternatively `removeChild` could be called when actually processing the done event - as it's sort of its result - but that would be bigger refactor so I wouldn't really recommend this at the moment.

davidkpiano: I removed the double-schedule issue here: https://github.com/statelyai/xstate/pull/2508/commits/a0a835bd00a6264814aa4006847666fec72e53f4

packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1105,111 @@ describe('actors', () => {
       })
       .start();
   });
+
+  // https://github.com/statelyai/xstate/issues/2507
+  it('should not crash on sync child completion before stopping', () => {
```
Andarist, line 1110: I ain't sure what the "before stopping" part is referring to here, it might be that I'm reading this in a wrong way 🤔  Maybe something like:
```suggestion
  it('should not crash on sync child completion during self-initialization', () => {
```

nit: all added tests are actually testing the same thing, just with different data types (machines, promises, observables etc) but yet their respective test titles are different. Could we unify them? It would get more clear that they are related - you could maybe even create a `describe` block for all of them

packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1105,111 @@ describe('actors', () => {
...
+        }
+      }
+    );
+    const service = interpret(parentMachine);
+    expect(() => {
+      service.start();
+    }).not.toThrow();
+  });
+
+  it('should not crash on immediate spawned promise completion', () => {
```
Andarist, line 1150:
```suggestion
  it('should not crash on immediate spawned promise-like completion', () => {
```

packages/core/src/interpreter.ts
```diff
@@ -1011,8 +1013,8 @@ export class Interpreter<
 
     childService
       .onDone((doneEvent) => {
-        this.removeChild(childService.id);
         this.send(toSCXMLEvent(doneEvent as any, { origin: childService.id }));
+        this.removeChild(childService.id);
```
Andarist, line 1017: any particular reason why this got moved to after the `this.send` call? I'm a little bit afraid that this has a potential of removing from the incorrect state - the one after the `this.send` "settles" and updates `this` with a new state. I could try to think of a failing test case for this tomorrow (not entirely sure if this is an actual possibility yet)

davidkpiano: Thought it would be safer but I'll move it back for now

davidkpiano: https://github.com/statelyai/xstate/pull/2508/commits/5025f1599dc1cca86b497481e3a2f4d3710148c8

packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1105,111 @@ describe('actors', () => {
       })
       .start();
   });
+
+  // https://github.com/statelyai/xstate/issues/2507
+  it('should not crash on sync child completion during self-initialization', () => {
```
Andarist, line 1110:
```suggestion
  it('should not crash on child machine sync completion during self-initialization', () => {
```

packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1105,111 @@ describe('actors', () => {
...
+        }
+      }
+    );
+    const service = interpret(parentMachine);
+    expect(() => {
+      service.start();
+    }).not.toThrow();
+  });
+
+  it('should not crash on immediate spawned promise-like completion', () => {
```
Andarist, line 1150:
```suggestion
  it('should not crash on child promise-like sync completion during self-initialization', () => {
```

packages/core/test/actor.test.ts
```diff
@@ -1105,4 +1105,111 @@ describe('actors', () => {
...
+        child: () => spawn({ then: (fn) => fn(null) } as any)
+      })
+    });
+    const service = interpret(parentMachine);
+    expect(() => {
+      service.start();
+    }).not.toThrow();
+  });
+
+  it('should not crash on immediate spawned observable completion', () => {
```
Andarist, line 1167:
```suggestion
  it('should not crash on child observable sync completion during self-initialization', () => {
```

Andarist (approved):

PR #2512, opened 2021-08-14
[examples] Add CameoParison Svelte example with machine

annaghi:
This example has a single machine, so it can be seen with the inspector.
Pair of this example is implemented with actors: https://github.com/statelyai/xstate/pull/2513

Temporary CodeSandbox link:
https://codesandbox.io/embed/github/annaghi/xstate/tree/examples/cameoparison-svelte-machine/examples/cameoparison-svelte-machine


examples/cameoparison-svelte-machine/src/machine.js
```diff
@@ -0,0 +1,204 @@
...
+
+  return {
+    celebs: Array.from(subset),
+    lookup
+  };
+};
+
+const loadRounds = (selection) => {
+  return Promise.all(selection.map((round) => loadCelebPair(round)));
+};
```
annaghi, line 34: This should be (and all the consequences need to be implemented):
```js
const loadRounds = (selection) => selection.map((round) => loadCelebPair(round));
```

The example works as it is, but I will improve it in the next weeks. My first approach was to put all the logic into the machines, and now I need to check what should be moved back into the components.

Same goes with the other example with actors.

annaghi:
Closing because of ruined Git on my side.

PR #2518, opened 2021-08-15
Action schema

davidkpiano:
Actions are now supported in the machine config's `.schema` property:

```ts
import { createMachine, createSchema } from 'xstate';

  schema: {
    actions: createSchema<
      { type: 'greet'; message: string } | { type: 'sayHello' }
    >()
  },
  entry: [
    { type: 'greet', message: 'hello' },
    { type: 'sayHello' },

    // @ts-expect-error (missing `message`)
    { type: 'greet' },

    // @ts-expect-error
    { type: 'other' }
  ]
});
```

Passing in a `Model` type as a generic type in `createMachine<typeof someModel>` is no longer allowed:

```ts
// Will not compile and show error containing:
// "Model type no longer supported as generic type.
// Please use `model.createMachine(...)` instead."
const machine = createMachine<typeof someModel>({
  // ...
});
```

CleanShot 2021-08-21 at 09 11 53

The `model.createMachine(...)` method should be used instead:

```diff
-const machine = createMachine<typeof someModel>({
+const machine = someModel.createMachine({
  // ...
});
```


packages/core/test/schema.test.ts
```diff
@@ -96,4 +96,22 @@ describe('schema', () => {
...
+        >()
+      },
+      entry: [
+        { type: 'greet', message: 'hello' },
+        { type: 'sayHello' },
+        // @ts-expect-error
+        { type: 'greet' },
+        // @ts-expect-error
+        { type: 'fake' }
+      ]
```
davidkpiano, line 114: @Andarist This test is failing because the machine is inferring the actions on `entry` here to be part of `TActions`, instead of _only_ inferring `schema.actions` (do we need `NoInfer` here?)

packages/core/src/actions.ts
```diff
@@ -330,22 +334,23 @@ export function log<TContext, TEvent extends EventObject>(
...
+  } as LogAction<TContext, TEvent>;
 }
 
-export const resolveLog = <TContext, TEvent extends EventObject>(
+export function resolveLog<TContext, TEvent extends EventObject>(
   action: LogAction<TContext, TEvent>,
   ctx: TContext,
   _event: SCXML.Event<TEvent>
-): LogActionObject<TContext, TEvent> => ({
-  // TODO: remove .expr from resulting object
```
Andarist, line 341: should the `expr` property stay on the resolved action?

packages/core/src/types.ts
```diff
@@ -886,10 +921,20 @@ export enum ActionTypes {
...
-export interface RaiseAction<TEvent extends EventObject> {
+// Used as a "phantom type" for built-in actions
+export interface BuiltInActionObject<TContext, TEvent extends EventObject> {
+  type: string;
+  __context: TContext;
+  __event: TEvent;
+}
+
+export type RaiseAction<TEvent extends EventObject> = BuiltInActionObject<
+  any,
```
Andarist, line 932: After some recent type-related problems in the viz I was wondering if it wouldnt be better to use `unknown` instead of `any` in situations like this. WDYT? With `any` it's easy to spill it somewhere unknowingly and get runtime errors because one thought that some property access was safe - it type "checked" after all ;p

packages/core/src/types.ts
```diff
@@ -886,10 +921,20 @@ export enum ActionTypes {
...
+  TEvent
+> & {
```
Andarist, line 934: TS team recommendation is to prefer interfaces (and extending from interfaces, rather creating types that are intersections) - they supposedly perform better (perf-wise) in some situations.

packages/core/test/schema.test.ts
Andarist: > do we need NoInfer here?

This is the only solution to this problem that I know of - so yes. I'm experimenting locally to see how to apply this technique in this situation

packages/core/src/types.ts
davidkpiano: I had interfaces at first; type inference didn't work. It magically works with types instead. (Try it out yourself)

packages/core/test/schema.test.ts
Andarist: I've applied this technique here:
https://github.com/statelyai/xstate/commit/b42b0a4cc81a40d253d9aebff43b31a9978d11b6

And it works... sort of. If we only make a single mistake the error is reported at correct position (it doesnt matter which one of those 2 invalid ones I include in the config):
Screenshot 2021-08-20 at 10 47 40
Note that red squiggles are shown because TS reports "Unused '@ts-expect-error' directive, ts(2578)" there as that other action is commented out.

If we include both mistakes the error "bubbles up":
Screenshot 2021-08-20 at 10 49 49
Which is somewhat correct in a sense that this won't allow invalid code but it's not the most user-friendly thing in the world.

The problem is also compounded by the reported error:
Screenshot 2021-08-20 at 10 53 26

This most likely happens because the message is only reported against the last member of the union, like here ([TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8AoKUA+UDeVSQFxQCIBzAJwgmAIG4oBbCAZwYEMiJ8HgSBLAOyKgBfZGkzZw7QixAAJCABt5AewJCR6LDkmce-GgyX18BAG6sArhGpQwJJWGMQSoVYKqJEAYyW9OUZvhw8GJaxqTklELuQA)):
```ts
type A =
  | { type: "greet"; message: string }
  | { type: "sayHello" }
  | { type: string; some: "vague"; prop: "erty" };

// Type '{ type: "greet"; }' is not assignable to type 'A'.
//   Type '{ type: "greet"; }' is missing the following properties from type '{ type: string; some: "vague"; prop: "erty"; }': some, prop(2322)
const a: A = { type: "greet" };
```
An interesting observation that can be made here is that is we change the `type` property of the last member to a literal type (so all types at the `type` property become literal strings) then TS can improve its error messaging here ([TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8AoKUA+UDeVSQFxQCIBzAJwgmAIG4oBbCAZwYEMiJ8HgSBLAOyKgBfZGkzZw7QixAAJCABt5AewJCR6LDkkBybTQZL6+AgDdWAVwjUoYEkrDGIJUKsFVEiAMZLenKM3w4eDEtY1JySiF3IA)): 
```ts
type A =
  | { type: "greet"; message: string }
  | { type: "sayHello" }
  | { type: ''; some: "vague"; prop: "erty" };

// Type '{ type: "greet"; }' is not assignable to type 'A'.
//   Property 'message' is missing in type '{ type: "greet"; }' but required in type '{ type: "greet"; message: string; }'.(2322)
const a: A = { type: "greet" };
```

Anyway... I've tried to fix the problem in our codebase but I couldn't. It seems that it's somehow related to the fact that `createMachine` has overloads because once I remove the first overload (the model-related one) then it works like intended:
Screenshot 2021-08-20 at 11 04 45
And when I reintroduce the first overload and change it to something non-sensical, that just can't be matched here, like this:
```ts
export function createMachine(config: never): void
```
I just land again with the same problem - the whole `createMachine` gets underlined and a vague error message is reported.

packages/core/src/types.ts
Andarist: It seems to work in the same way with interfaces and types: https://github.com/statelyai/xstate/commit/7e1448c5be9ec4f36409d4af39e32083f30435ed

packages/core/src/actions.ts
davidkpiano: I don't think so; the resulting action object is the result of the dynamic action; `.expr` is an implementation detail.

packages/core/src/types.ts
davidkpiano: Done 👍

packages/core/src/types.ts
davidkpiano: 🍒 ⛏️

packages/core/test/schema.test.ts
davidkpiano: I removed the first overload.

packages/core/src/actions.ts
Andarist: Since we are spreading the `action` here - the `expr` becomes part of it. So if this doesn't have to stay on the resolved action it might be worth removing it.

packages/core/src/types.ts
```diff
@@ -894,7 +932,15 @@ export enum ActionTypes {
   Choose = 'xstate.choose'
 }
 
-export interface RaiseAction<TEvent extends EventObject> {
+// Used as a "phantom type" for built-in actions
+export interface BuiltInActionObject<TContext, TEvent extends EventObject> {
+  type: string;
+  __context: TContext;
+  __event: TEvent;
```
Andarist, line 939: you could make those optional to avoid casting return values in the action resolvers: 
https://github.com/statelyai/xstate/commit/8ca10ac7e56daf8bd4d5355d51f1ec5535a01a4f

Andarist:
Types for `choose` became overly permissive with this change. This correctly errors on the `main` branch (but it doens't on this one):
Screenshot 2021-08-23 at 11 18 41

I was experimenting with this using this machine: https://github.com/statelyai/xstate/blob/6cf01b4cf8dbf26b266674de1fa415e43b248bcc/packages/core/test/actions.test.ts#L1046-L1056

It's probably caused by incorrect/inaccurate types being used in the `ChooseCondition` here:
https://github.com/statelyai/xstate/blob/6cf01b4cf8dbf26b266674de1fa415e43b248bcc/packages/core/src/types.ts#L86

`Actions` type is using the `Action` type that contains an `ActionObject`:
https://github.com/statelyai/xstate/blob/6cf01b4cf8dbf26b266674de1fa415e43b248bcc/packages/core/src/types.ts#L91

This in turn `extends BaseActionObject`:
https://github.com/statelyai/xstate/blob/6cf01b4cf8dbf26b266674de1fa415e43b248bcc/packages/core/src/types.ts#L37

And which in turns is defined like this:
https://github.com/statelyai/xstate/blob/6cf01b4cf8dbf26b266674de1fa415e43b248bcc/packages/core/src/types.ts#L24-L30

Since `assign({ other: 42 })` doesn't embed any generic references or anything like that it's basically treated as `{ type: string }` in this context - thanks to TS structural typing and all.

How to fix this? Well, this is a little bit tricky. It's just that the structure of the current types are prone to such regressions. And that we have quite a bit of types related to actions, things are extending each other etc. Creating a hierarchy of types (a diagram perhaps) and assigning conceptual roles to them would help quite a bit - both to fix this problem but also for documentation purposes. I would happily hop on a call regarding this tomorrow or something - we could hash this out quickly.

davidkpiano:
@Andarist This correctly errors for me (even better than `main`):

CleanShot 2021-08-23 at 06 11 02

Andarist:
@davidkpiano notice that on my screenshot I'm not providing an explicit generic to `assign`. Try removing `<Ctx>`.

This also turns out to be a more bigger issue than what has been described in relation to the `choose` action. Even those ones don't error correctly:
```ts
createMachine<{ counter: number }>({
  entry: assign({ counter: 'string' })
});

createMachine<{ counter: number }>({
  entry: assign(() => ({ counter: 'string' }))
})
```

packages/core/src/types.ts
Andarist: Hm, it turns out that this doesn't report those are problems anymore:
https://github.com/statelyai/xstate/blob/6cf01b4cf8dbf26b266674de1fa415e43b248bcc/packages/core/test/schema.test.ts#L110-L113

It seems highly related to this comment: https://github.com/statelyai/xstate/pull/2518#issuecomment-903603539 . It's because both builtin action objects and `TAction` use `type: string`, so those types are overlapping - where in reality they are 2 different concepts

davidkpiano:
Closing as we're exploring better options in v5

PR #2525, opened 2021-08-16
Remove unwanted imports, other tiny fixes

annaghi:
`react-tic-tac-toe/src/ticTacToe.machine.ts` still has two type errors.

### DIFF
--- a/examples/react-tic-tac-toe/src/ticTacToe.machine.ts
+++ b/examples/react-tic-tac-toe/src/ticTacToe.machine.ts
@@ -1,8 +1,5 @@
-import {
-  createModel,
-  ModelContextFrom,
-  ModelEventsFrom
-} from 'xstate/lib/model';
+import { createModel } from 'xstate/lib/model';
+import { ContextFrom, EventFrom } from 'xstate';
 
 type Player = 'x' | 'o';
 
@@ -22,13 +19,13 @@ const model = createModel(
 );
 
 const isValidMove = (
-  ctx: ModelContextFrom<typeof model>,
-  e: ModelEventsFrom<typeof model> & { type: 'PLAY' }
+  ctx: ContextFrom<typeof model>,
+  e: EventFrom<typeof model> & { type: 'PLAY' }
 ) => {
   return ctx.board[e.value] === null;
 };
 
-function checkWin(ctx: ModelContextFrom<typeof model>) {
+function checkWin(ctx: ContextFrom<typeof model>) {
   const { board } = ctx;
   const winningLines = [
     [0, 1, 2],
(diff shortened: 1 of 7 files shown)


examples/todo-mvc-react/src/App.jsx
```diff
@@ -6,7 +6,7 @@ function App() {
     <>
       <Todos />
 
-      <footer class="info">
+      <footer className="info">
```
annaghi, line 9: This was my mistake :recycle:

examples/react-tic-tac-toe/src/ticTacToe.machine.ts
```diff
@@ -22,13 +19,13 @@ const model = createModel(
 );
 
 const isValidMove = (
-  ctx: ModelContextFrom<typeof model>,
-  e: ModelEventsFrom<typeof model> & { type: 'PLAY' }
+  ctx: ContextFrom<typeof model>,
+  e: EventFrom<typeof model> & { type: 'PLAY' }
```
davidkpiano, line 23: Thinking out loud... wondering how useful something like this would be:
```ts
event: EventFrom<typeof model, 'PLAY'> // restricts to `{ type: 'PLAY' }`
```

davidkpiano (approved):

examples/react-tic-tac-toe/src/ticTacToe.machine.ts
annaghi: Sounds good!

Continue the thinking out loud.
I pick up TypeScript notation on the fly, so if I misuse the symbols then my words have preference.

If the restriction is flexible then it can serve wide range of needs. For example there can be a case where a parametrized action need to be reused for couple of events.

Let assume we have these events:
```ts
{
  events: {
    PLAY: (value: number) => ({ value }),
    SAVE: (value: number, date: Date) => ({ value, date }),
    RESET: () => ({})
  }
}
```

and let `Type1 === Type2` mean that the two types are the same.

Some restriction levels I can see now:

1. No restriction, event can be any event from the model
The function body can only use the properties which are common (intersection) for all events, and this is the `event.type`
```ts
EventFrom<typeof model> ===
{ type: 'PLAY'; value: number; } | { type: 'RESET'; } | { type: 'SAVE'; value: number; date: Date; }
```

2. Restriction to a subset of events, might be useful for general purpose parametrized actions.
The intersection is `event.type` and `event.value`, so these are the ones the function can use:
```ts
EventFrom<typeof model, 'PLAY', 'SAVE'> ===
{ type: 'PLAY'; value: number; } | { type: 'SAVE'; value: number; date: Date; }
```

3. Restriction to a single event (this is your case), all properties of a single event are available inside the function:
```ts
EventFrom<typeof model, 'PLAY'> ===
{ type: 'PLAY'; value: number; }
```

4. Restricts to a subset of properties, might be useful for general parametrized actions BUT without knowing the concrete events
The function can use `event.value` and `event.type` too implicitly
```ts
EventFrom<typeof model, { value: number }> ===
{ type: 'PLAY'; value: number; } | { type: 'SAVE'; value: number; date: Date; }
```

PR #2526, opened 2021-08-17
fix action.exec && action.exec()?

Kazintsev:
should be `typeof action.exec === 'function' && action.exec();`?

because got an error ("TypeError: action.exec is not a function") on `action.exec && action.exec();`

### DIFF
--- a/docs/guides/interpretation.md
+++ b/docs/guides/interpretation.md
@@ -217,7 +217,7 @@ function send(event) {
 
   actions.forEach((action) => {
     // If the action is executable, execute it
-    action.exec && action.exec();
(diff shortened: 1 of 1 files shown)


davidkpiano:
Looks good, can you add a changeset by running `yarn changeset` (this would be a `patch` change)?

Kazintsev:
I do apologize. I don't have local repository with yarn/npm installed to make a patch.

davidkpiano:
> I do apologize. I don't have local repository with yarn/npm installed to make a patch.

No worries, I'm sorry because I didn't realize this was just a documentation change; you're all good!

PR #2530, opened 2021-08-18
Deprecate activities

davidkpiano:
This PR adds TypeScript annotations for deprecating `activities`.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -592,6 +592,8 @@ export interface StateNodeConfig<
   /**
    * The activities to be started upon entering the state node,
    * and stopped upon exiting the state node.
+   *
+   * @deprecated Use `invoke` instead.
    */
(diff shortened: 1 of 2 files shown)


mattpocock (approved):

bayssmekanique:
@davidkpiano, I see that activities are being deprecated but I can't seem to find where the alternative is listed. None of the services types support long lived processes that need to stay alive for the life of the state node, they all expect the service to resolve on their own when their task is complete.

davidkpiano:
> 

See here: https://github.com/statelyai/xstate/pull/2878

---

Activites are deprecated and will be removed in XState version 5. The recommended approach is to [invoke an actor](./actors.md) instead:

```diff
-activities: [(context, event) => {
-  // do something
-
-  return () => {/* cleanup */}
-}],
+invoke: {
+  src: (context, event) => (sendBack, receive) => {
+    // do something
+
+    return () => {/* cleanup */}
+  }
+}
```

PR #2537, opened 2021-08-19
Expand the StateFrom<T> type helper

VanTanev:
Minor expansion of the `StateFrom<T>` helper to handle `Interpreter<...>` as well as `ActorRefFrom<Machine>` types.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -828,9 +828,14 @@ export interface StateMachine<
   ): StateMachine<TContext, TStateSchema, TEvent, TTypestate>;
 }
 
-export type StateFrom<
-  TMachine extends StateMachine<any, any, any>
-> = ReturnType<TMachine['transition']>;
+export type StateFrom<T> = T extends StateMachine<any, any, any>
+  ? ReturnType<T['transition']>
+  : T extends Interpreter<any, any, any, any>
+  ? ReturnType<T['nextState']>
+  : // infer state if given ActorRefFrom<Machine>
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -828,9 +828,14 @@ export interface StateMachine<
...
-export type StateFrom<
-  TMachine extends StateMachine<any, any, any>
-> = ReturnType<TMachine['transition']>;
+export type StateFrom<T> = T extends StateMachine<any, any, any>
+  ? ReturnType<T['transition']>
+  : T extends Interpreter<any, any, any, any>
+  ? ReturnType<T['nextState']>
+  : // infer state if given ActorRefFrom<Machine>
+  T extends ActorRef<any, any> & { state: infer S }
+  ? S
```
davidkpiano, line 837: This will be deprecated; maybe we should avoid adding this? The `extends Interpreter` case will catch this anyway.

Otherwise, looks good!

VanTanev: @davidkpiano I don't think that `extends Interpreter` catches this.

The idea is to be able to extract state from 
```ts
type MachineActorRef = ActorRefFrom<Machine>
```

I can't say
```ts
type StateFrom<T extends ActorRef<any, infer S>> = S`
```

Because not all actor refs actually have a "state" in the sense of a machine state.

I don't know a better approach to deal with 
```ts
type MachineActorRef = ActorRefFrom<Machine>
type MachineState  = StateFrom<MachineActorRef>
```

PR #2538, opened 2021-08-20
[examples] CameoParison Svelte with machine

annaghi:
Original example: https://github.com/Rich-Harris/cameoparison-starter/tree/complete
Example with actors: https://github.com/statelyai/xstate/pull/2539

CodeSandbox
https://codesandbox.io/embed/github/annaghi/xstate/tree/examples-cameoparison-svelte-machine/examples/cameoparison-svelte-machine

Issue #1792, opened 2020-12-30
Warning: No implementation found for action type 'xstate.assign'


davydof:
**Description**
Hello, after I've updated on `4.14.1` I get warning:
```
Warning: No implementation found for action type 'xstate.assign'
```
from place:
https://github.com/davidkpiano/xstate/blob/b64c8719b043a8f283ab12e6f4984e39e56c7b4c/packages/core/src/interpreter.ts#L906-L913

```js
import { assign, Machine } from 'xstate';

const createMachine = () => Machine({
  initial: 'pending',
  context: {
    foo: undefined,
  },
  states: {
    pending: {
      // with event instead of invoking actor warning appears as well
      /*on: {
        DO: 'done',
      },*/
      invoke: {
        src: () => Promise.resolve(),
        onDone: 'done',
      },
    },
    done: {
      exit: assign({
        foo: () => undefined,
      }),
    },
  },
});
```

Test:
```js
import { interpret } from 'xstate';

interpret(createMachine()).start();
```

No warning: `4.14.0`
Warning appears: `4.14.1`

**Expected Result**
No warnings

**Actual Result**
```
console.warn node_modules/xstate/lib/utils.js:451
Warning: No implementation found for action type 'xstate.assign'
```

**Reproduction**
https://codesandbox.io/s/xstate-warning-no-implementation-found-for-action-type-xstateassign-1699k?file=/index.test.js

davidkpiano:
The reproduction is not showing that warning.

davydof:
@davidkpiano 
Yes, it's true, but why I can get it in tests locally?

davidkpiano:
Not sure, but a reproduction is necessary. A repo is fine.

davydof:
Looks strange
I get it in tests.
And in production but not every time.

What I can do?

Andarist:
Debug.

davydof:
Reproduced, please check [codesandbox](https://codesandbox.io/s/xstate-warning-no-implementation-found-for-action-type-xstateassign-1699k?file=/index.test.js).

After test added:
```
interpreted.stop()
```

Andarist:
I can confirm the issue and got a basic understanding of it in our codebase - gonna comment more about this sometime later.

davidkpiano:
For now, you can safely ignore that warning (for `xstate.assign` specifically).

Andarist:
The problem comes from this call `exec` call:
[...]

PR #2539, opened 2021-08-20
[examples] CameoParison Svelte with actors

annaghi:
Original example: https://github.com/Rich-Harris/cameoparison-starter/tree/complete
Example with machine: https://github.com/statelyai/xstate/pull/2538

CodeSandbox
https://codesandbox.io/embed/github/annaghi/xstate/tree/examples-cameoparison-svelte-actors/examples/cameoparison-svelte-actors

cameoparison-actor-model


examples/cameoparison-svelte-actors/src/app/machine.js
```diff
@@ -0,0 +1,38 @@
...
+              spawn(gameMachine(event.data), 'gameActor')
+          })
+        }
+      }
+    },
+    game: {
+      on: {
+        GREET: 'welcome'
+      },
+      exit: stop('gameActor')
```
Andarist, line 35: while it might not matter here - maybe it would be a good idea to also remove this from the context?

annaghi: Absolutely good advice: calling `stop` and cleaning the `context` should go hand in hand.

annaghi: I thought I resolved this piece, but there are some warnings in the console now:
```
Warning: No implementation found for action type 'xstate.assign'
```
It was reported here: https://github.com/statelyai/xstate/issues/1792

The warning appears a couple of places, for example after playing the game it lands on the Feedback page, and when clicking on the back button the warning is logged.

I tried adding the `preserveActionOrder: true` thinking that the fixed order will help, but it did not, so I removed it. I am not sure if the `preserveActionOrder` is needed here or not.

annaghi: Found two different cases when the warning appeared for me:
- modelling: I defined `exit` action when the machine was not leaving that state ever -> fixed here: https://github.com/statelyai/xstate/pull/2539/commits/5409f9b0a6010e0b175eb4ca2c2b1451baa44cdd
- Vite HMR: when making some change in the source and the running app was reloaded

PR #2546, opened 2021-08-22
[core] Add `state.can(event)`

davidkpiano:
This PR adds `state.can(...)`:

```js
someState.can({ type: 'TOGGLE' }); // `true` if it will cause a transition
```


VanTanev:
If we're adding this in core, I would like to see the full functionality of https://github.com/VanTanev/xstate-helpers/blob/development/src/react/useIsXStateTransitionAvailable.ts - eg, type inference for both event types as strings and event objects - XState supports that in other places.

Also, I strongly believe that the result of `state.can()` should always be a `boolean` and never `undefined`. If there is no machine available on the state, `can()` should just return `false`.

Also, what is the best practice around memorizing this? Is `machine.transition()` expensive enough to want to use `React.useMemo()` around `state.can()`?

VanTanev:
```diff
diff --git a/packages/core/src/State.ts b/packages/core/src/State.ts
index 17435556..2762c9b6 100644
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -129,7 +129,7 @@ export class State<
    */
   public children: Record<string, ActorRef<any>>;
   public tags: Set<string>;
-  private machine: StateMachine<TContext, any, TEvent, any> | undefined;
+  protected machine: StateMachine<TContext, any, TEvent, any> | undefined;
   /**
    * Creates a new State instance for the given `stateValue` and `context`.
    * @param stateValue
@@ -321,16 +321,16 @@ export class State<
   /**
    * Determines whether sending the `event` will cause a transition.
    *
-   * If this state was created outside of a machine (e.g., with `State.from(...)`), `undefined` will be returned.
+   * If this state was created outside of a machine (e.g., with `State.from(...)`), `false` will be returned.
    *
    * @param event The event to test
    * @returns Whether the event will cause a transition
    */
-  public can(event: TEvent): boolean | undefined {
+  public can(event: TEvent | TEvent['type']): boolean {
     if (!this.machine) {
-      return undefined;
+      return false;
     }
 
-    return this.machine.transition(this, event).changed;
+    return !!this.machine.transition(this, event).changed;
   }
 }
diff --git a/packages/core/test/state.test.ts b/packages/core/test/state.test.ts
index 8b56738f..70145a3f 100644
--- a/packages/core/test/state.test.ts
+++ b/packages/core/test/state.test.ts
@@ -590,6 +590,10 @@ describe('State', () => {
             on: {
               NO_CHANGE: 'a',
               ACTION: { actions: 'someAction' },
+              ACTION_WITH_COND: {
+                actions: 'someAction',
+                cond: (_: any, e: any) => e.cond
+              },
               CHANGE: 'b'
             }
           },
@@ -598,15 +602,30 @@ describe('State', () => {
       });
 
       expect(machine.initialState.can({ type: 'UNKNOWN' })).toEqual(false);
+      expect(machine.initialState.can('UNKNOWN')).toEqual(false);
+
       expect(machine.initialState.can({ type: 'NO_CHANGE' })).toEqual(false);
+      expect(machine.initialState.can('NO_CHANGE')).toEqual(false);
+
       expect(machine.initialState.can({ type: 'ACTION' })).toEqual(true);
+      expect(machine.initialState.can('ACTION')).toEqual(true);
+
       expect(machine.initialState.can({ type: 'CHANGE' })).toEqual(true);
+      expect(machine.initialState.can('CHANGE')).toEqual(true);
+
+      expect(
+        machine.initialState.can({ type: 'ACTION_WITH_COND', cond: true })
+      ).toEqual(true);
+      expect(
+        machine.initialState.can({ type: 'ACTION_WITH_COND', cond: false })
+      ).toEqual(false);
+      expect(machine.initialState.can('ACTION_WITH_COND')).toEqual(false);
     });
 
-    it('should return undefined for states created without a machine', () => {
+    it('should return false for states created without a machine', () => {
       const state = State.from('test');
 
-      expect(state.can({ type: 'ANY_EVENT' })).toEqual(undefined);
+      expect(state.can({ type: 'ANY_EVENT' })).toEqual(false);
     });
   });
 });
 ```

**Comment by VanTanev** [comment]
After the last commit, this is *chef's kiss~*!
💯 

Only thing left to fix is an incompatibility with `StateConfig` type.

It might be OK to make `machine` a public property?

File: packages/core/src/State.ts [hunk 1]
```diff
@@ -128,6 +131,7 @@ export class State<
    */
   public children: Record<string, ActorRef<any>>;
   public tags: Set<string>;
+  private machine: StateMachine<TContext, any, TEvent, any> | undefined;
```
VanTanev, line 134:
```suggestion
  private machine: StateMachine<TContext, any, TEvent, any> | undefined;
  public machine: StateMachine<TContext, any, TEvent, TTypestate> | undefined;
```

davidkpiano: @Andarist Deferring to you; would this be okay to do?

packages/core/src/State.ts
```diff
@@ -314,4 +319,23 @@ export class State<
   public hasTag(tag: string): boolean {
     return this.tags.has(tag);
   }
+
+  /**
+   * Determines whether sending the `event` will cause a transition.
+   *
+   * If this state was created outside of a machine (e.g., with `State.from(...)`), `undefined` will be returned.
+   *
```
davidkpiano, line 327:
```suggestion
```

VanTanev:
Oops, my bad for leaving the 2nd machine prop there, sorry.

davidkpiano:
> How should errors thrown by assign actions and guards be treated? How they should be treated when we implement error-related changes and we actually start capturing all unhandled errors? Should an error be counted as a change (errors will, most likely, try to select a transition - why way or another)?

Can this be resolved in a separate PR? Let's keep this one low-scope.

packages/core/src/State.ts
```diff
@@ -314,4 +319,20 @@ export class State<
...
+   * @param event The event to test
+   * @returns Whether the event will cause a transition
+   */
+  public can(event: TEvent | SimpleEventsOf<TEvent>['type']): boolean {
+    if (IS_PRODUCTION) {
+      warn(
+        !!this.machine,
+        `state.can(...) used outside of a machine-created State object; this will always return false.`
+      );
+    }
```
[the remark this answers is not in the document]
davidkpiano: https://github.com/statelyai/xstate/issues/2576

davidkpiano:
@Andarist Ready for re-review

.changeset/nice-pugs-shout.md
```diff
@@ -0,0 +1,34 @@
+---
+'xstate': minor
+---
+
+You can now know if an event will cause a state change by using the new `state.can(event)` method, which will return `true` if the machine will change the state when sent the `event`, or `false` otherwise:
```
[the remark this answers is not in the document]
farskid: 💯 agree with both comments

packages/core/test/state.test.ts
```diff
@@ -574,4 +581,213 @@ describe('State', () => {
       expect(machine.transition(undefined, 'TO_FINAL').done).toBeTruthy();
     });
   });
+
+  describe('.can', () => {
+    it('should return true for a simple event that results in a transition to a different state', () => {
+      const machine = createMachine({
```
farskid, line 587: Should we add an example for when only the context changes?

farskid (approved):
Besides the comments, LGTM.

Andarist:
> Can this be resolved in a separate PR? Let's keep this one low-scope.

At this stage, this is a design question about this API - both for v4 and for v5 (the answer can differ for both). I feel like it should be answered before merging this PR in because whatever lands on the main branch should be "releasable" and we shouldn't release something that might change (behavior-wise) soon-ish.

If you don't want to keep this PR low-scope but you'd like to make the design of this more tight & explicit before releasing then you could always create a PR targeting this PR. 

If the answer is: "nothing should happen in v4 and errors should bubble to the caller" then it's fine. We can refine this in v5 (but a ticket~ should be created so we don't forget about it) but the appropriate tests should be added to validate this and to prevent us from regressing.

davidkpiano:
> If the answer is: "nothing should happen in v4 and errors should bubble to the caller" then it's fine. We can refine this in v5 (but a ticket~ should be created so we don't forget about it) but the appropriate tests should be added to validate this and to prevent us from regressing.

This is the answer IMO. Just added a test for this 👍 

How would this be refined in v5?

Andarist:
> How would this be refined in v5?

That depends on the outcome of the error handling refactor. The behavior has been settled for the v4 though so let's just don't bother with thinking how this related to v5

.changeset/nice-pugs-shout.md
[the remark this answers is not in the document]
davidkpiano: Added to changeset. Will add the button example to the documentation in a subsequent PR

packages/core/test/state.test.ts
davidkpiano: Added!

davidkpiano:
@Andarist Ready for final review 🏁

packages/core/test/state.test.ts
```diff
@@ -615,6 +615,41 @@ describe('State', () => {
...
+              NEXT: {
+                actions: 'newAction'
+              }
+            }
+          }
+        }
+      });
+
+      expect(machine.initialState.can({ type: 'NEXT' })).toBe(true);
+    });
```
Andarist, line 633: note - this is basically a duplicate of https://github.com/statelyai/xstate/blob/65c7e62a506d34648fb046278ec0427943236981/packages/core/test/state.test.ts#L702-L717

Andarist (approved):

PR #2548, opened 2021-08-23
adds complete listener from subscribe to the Interpreter's stopListeners

chanced:
I'm not 100% on this as complete sort of implies done. If you stop a machine, does that mean it is complete?

### DIFF
--- a/.changeset/shaggy-rabbits-boil.md
+++ b/.changeset/shaggy-rabbits-boil.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+adds complete listener from subscribe to the Interpreter's stopListeners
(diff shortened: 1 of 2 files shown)


Andarist:
Being stopped is not the exact same thing as being done - stopped is more like being canceled whereas being done means reaching a final state. 

Could you tell us a little bit more about your use case and how would you utilize this?

chanced:
> Being stopped is not the exact same thing as being done 

Agreed, but the handler is for complete. 

> Could you tell us a little bit more about your use case and how would you utilize this?

There is currently no way to determine whether or not an actor has been stopped and subscribe'scomplete handler is invoked upon a call to `stop`.

In my case, I have actors that create subscriptions to other actors. As part of their cleanup, I'd like to send an event indicating the subscription has ended. However, doing can cause warnings for sending to stopped actors.

See: https://github.com/chanced/xstate-pubsub/blob/main/src/subscribe.ts#L80-L84

davidkpiano:
Closing as XState v5 beta now calls completion listeners upon an actor being stopped.

PR #2549, opened 2021-08-24
Export createModel from index.ts

chanced:
Given how common it is to `createModel` if using TypeScript, it'd be nice if it were exported off of index.ts like `createMachine`.

### DIFF
--- a/.changeset/loud-onions-tickle.md
+++ b/.changeset/loud-onions-tickle.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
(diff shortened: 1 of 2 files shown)


davidkpiano:
Thanks, but it's intentionally left off of the main import, as it is an opt-in helper.

chanced:
Then you definitely won't like this pull request: https://github.com/statelyai/xstate/pull/2552

Everything is sort of opt in though

Anyways, cheers.

PR #2551, opened 2021-08-24
Widened the *From utility types to allow extracting from factory functions

mattpocock:
Copied from changeset:

Widened the *From utility types to allow extracting from factory functions.

This allows for:

```ts
const makeMachine = () => createMachine({});

type Interpreter = InterpreterFrom<typeof makeMachine>;
type Actor = ActorRefFrom<typeof makeMachine>;
type Context = ContextFrom<typeof makeMachine>;
type Event = EventsFrom<typeof makeMachine>;
```

This also works for models, behaviours, and other actor types.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -2,7 +2,6 @@ import { StateNode } from './StateNode';
 import { State } from './State';
 import { Interpreter, Clock } from './interpreter';
 import { Model } from './model.types';
-import { createMachine } from './Machine';
 
 export type EventType = string;
 export type ActionType = string;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -828,9 +828,11 @@ export interface StateMachine<
   ): StateMachine<TContext, TStateSchema, TEvent, TTypestate>;
 }
 
-export type StateFrom<
-  TMachine extends StateMachine<any, any, any>
-> = ReturnType<TMachine['transition']>;
+export type StateFrom<T> = T extends StateMachine<any, any, any>
+  ? ReturnType<T['transition']>
+  : T extends (...args: any[]) => StateMachine<any, any, any>
+  ? ReturnType<ReturnType<T>['transition']>
+  : never;
 
 export interface ActionMap<TContext, TEvent extends EventObject> {
   onEntry: Array<Action<TContext, TEvent>>;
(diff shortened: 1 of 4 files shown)


davidkpiano (approved):
Really useful change. Wondering if it would be too much abstraction to have a type like this?

```ts
type FactoryOrValue<T> = T extends ((...args: any[]) => infer U) ? U : T;
```

Then we can do:
```ts
export type StateFrom<T> = FactoryOrValue<T> extends StateMachine<any,any,any> ? FactoryOrValue<T>['transition'] : ...
```

Just thinking out loud 🤔

packages/core/src/types.ts
```diff
@@ -828,7 +828,11 @@ export interface StateMachine<
   ): StateMachine<TContext, TStateSchema, TEvent, TTypestate>;
 }
 
-export type StateFrom<T> = T extends StateMachine<any, any, any>
+export type StateFrom<
+  T extends
```
davidkpiano, line 832: Is the idea to add to this `extends` whenever we add a new thing that `StateFrom` can extract a state from?

mattpocock: Yes, that will work.

mattpocock:
> Really useful change. Wondering if it would be too much abstraction to have a type like this?
> 
> ```ts
> type FactoryOrValue<T> = T extends ((...args: any[]) => infer U) ? U : T;
> ```
> 
> Then we can do:
> 
> ```ts
> export type StateFrom<T> = FactoryOrValue<T> extends StateMachine<any,any,any>
>   ? FactoryOrValue<T>['transition']
>   : ...
> ```
> 
> Just thinking out loud 🤔

Probably a good idea? I'm really nervous to do stuff like this in a codebase that doesn't have `strict: true` in it, but it would probably work.

mattpocock:
@davidkpiano Tried the above, doesn't work

Andarist:
Re `FactoryValue` - I have tried a slightly different approach and it seems to work: https://github.com/mattpocock/xstate/pull/1/files . It would be great to add some tests for this though.

packages/core/src/types.ts
Andarist: I've removed this on my branch - https://github.com/mattpocock/xstate/pull/1/files . I don't have super strong opinion about this and definitely having those constraints in place could improve where type errors are reported and so on. We didn't have those in other *From utilities - so I've just gone for the unification for the time being. We can always unify this other way around - although keeping the list of those things is a little bit tedious 🤷‍♂️

davidkpiano:
> Re `FactoryValue` - I have tried a slightly different approach and it seems to work: https://github.com/mattpocock/xstate/pull/1/files . It would be great to add some tests for this though.

Nice; well, we can do refactor later. I think this is good to go in, wdyt?

davidkpiano (approved):

PR #2552, opened 2021-08-24
Simplifies exports on index, including all exports from pertinent files

chanced:
This is just a nice-to-have. It simplifies the surface of xstate imports.

### DIFF
--- a/packages/core/src/Actor.ts
+++ b/packages/core/src/Actor.ts
@@ -7,9 +7,9 @@ import {
   Spawnable,
   SCXML
 } from './types';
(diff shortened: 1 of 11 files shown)


chanced:
Wasn't sure what this would be considered in terms of a changeset. It shouldn't impact anything downstream though, so patch?

chanced:
There is actually one possible breaking change in this. The function `isActor` was replicated in both Actor.ts and utils.ts. I removed the one from Actor.ts in favor of the utils.

Also apologize about accidentally including commit & revert of screaming snakecase change of actionTypes. I can toss this pull req and open a new one that is rebased.

PR #2560, opened 2021-08-26
feat(core/behaviors): add stop callback

ChrisShank:
To build out more complicated behaviors I think we need to have to ability for behaviors to know when they are stopped, similar to invoked callbacks.

I would like this functionally to build out some helpers located [here](https://github.com/ChrisShank/xstate-behaviors). I have been using invoked callbacks so far, but the more I think about it the more I think I need to use the behavior API.

### DIFF
--- a/packages/core/src/behaviors.ts
+++ b/packages/core/src/behaviors.ts
@@ -89,7 +89,7 @@ export function fromPromise<T>(
         }
       );
 
-      return initialState;
+      return { state: initialState };
     }
   };
 }
@@ -104,6 +104,7 @@ export function spawnBehavior<TEvent extends EventObject, TEmitted>(
   options: SpawnBehaviorOptions = {}
 ): ActorRef<TEvent, TEmitted> {
   let state = behavior.initialState;
+  let stop: (() => void) | undefined;
   const observers = new Set<Observer<TEmitted>>();
   const mailbox: TEvent[] = [];
   let flushing = false;
@@ -138,7 +139,8 @@ export function spawnBehavior<TEvent extends EventObject, TEmitted>(
           observers.delete(observer);
         }
       };
-    }
+    },
+    stop: () => stop?.()
   });
 
   const actorCtx = {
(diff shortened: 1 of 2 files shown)


ChrisShank:
@Andarist do you like this implementation better? No breaking changes now and much simpler imo. Also added a `done` flag so that events cannot be sent to a stopped behavior.

packages/core/src/behaviors.ts
```diff
@@ -138,6 +139,11 @@ export function spawnBehavior<TEvent extends EventObject, TEmitted>(
           observers.delete(observer);
         }
       };
+    },
+    stop: () => {
+      state = behavior.stop ? behavior.stop(actorCtx) : state;
+      observers.clear();
```
ChrisShank, line 145: I *think* we want to clear the observers?

packages/core/src/behaviors.ts
```diff
@@ -107,9 +107,10 @@ export function spawnBehavior<TEvent extends EventObject, TEmitted>(
   const observers = new Set<Observer<TEmitted>>();
   const mailbox: TEvent[] = [];
   let flushing = false;
+  let done = false;
 
   const flush = () => {
-    if (flushing) {
+    if (flushing || done) {
```
ChrisShank, line 113: I made it so that a behavior cannot be flushed after its stopped. I think I might also have to make sure the mailbox cannot be filled after its done since that might be a memory leak.

packages/core/src/behaviors.ts
ChrisShank: Would we also want to clear the mailbox too?

ChrisShank:
For reference this is what `fromWebSocket` looks like with this new functionality: https://github.com/ChrisShank/xstate-behaviors/blob/f9bc2766f92bf4bc76810188445ff6fe5d223a79/src/real-time/from-web-socket.ts

ChrisShank:
I refactored more helpers to use the behavior API https://github.com/ChrisShank/xstate-behaviors/pull/2

Andarist:
This, unfortunately, got stale and the v5 APIs are significantly different in that area. If those APIs are not enough to cover the use case here - let's start a thread about it to discuss this feature.

PR #2522, opened 2021-08-16
Allow for lazy context in `machine.withConfig`

farskid:

PR #2567, opened 2021-08-27
Add a failing test for context composition using multiple `provide` calls

Andarist:
When working on this PR: https://github.com/statelyai/xstate/pull/2522 and comparing the implementation with the `next` branch I've noticed that this doesn't work properly. I didn't have time to dig into this more but hopefully, this shouldn't be too hard to fix - I'm opening this so we don't lose track of it.

### DIFF
--- a/packages/core/test/machine.test.ts
+++ b/packages/core/test/machine.test.ts
@@ -146,7 +146,7 @@ describe('machine', () => {
     });
   });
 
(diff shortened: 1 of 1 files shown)


davidkpiano:
This is likely to be irrelevant when #3675 goes in, which will be recommended over "extending" context.

PR #2579, opened 2021-08-30
update @xstate/inspect README about new beta viz URL

jebeck:
Related to my other PR #2580. Just my suggestion for how you might want to handle the new vs. old viz situation at the moment.

Issue #2542, opened 2021-08-21
documented syntax for entry actions crashes the inspector


jebeck:
**Description**
The configuration that exists in the documentation for `entry` actions and does _not_ cause any errors in application code crashes the inspector.

**Expected Result**
Can view the working machine and spawned child machine in the inspector.

**Actual Result**
"An unexpected error has occurred." in the inspector. Dev tools shows:
<img width="685" alt="Screen Shot 2021-08-20 at 18 21 42" src="https://user-images.githubusercontent.com/1588547/130306087-8f3aa743-b728-474b-ae96-3a384090247a.png">

**Reproduction**
Reproduction here: https://codesandbox.io/s/morning-moon-qdjnp

The core issues is specifying `entry` like this (crashes inspector):
```
entry: ["createNewNode", log()],
```
vs. this, which works but which is not how entry actions are documented (as far as syntax) in [the docs on actions](https://xstate.js.org/docs/guides/actions.html?#actions):
```
entry: { type: "createNewNode" },
```
(I also don't know how to include the log action in the second type of syntax... I tried an array of objects, but it didn't work.)

**Additional context**
Maybe of interest to note: things like this happen to me all the time with the inspector, where one syntax works for a machine without errors but not for the inspector. If there was a way to view all possible syntaxes of different types of machine properties, that would be helpful to reduce my search space when things like this happen. As I said above, I had no idea that the second way of configuring `entry` was even possible until I combed through a lot of issues here in the repo and saw it referenced [here](https://github.com/statelyai/xstate/issues/1437). Thanks for the tooling tho, the inspector is beyond valuable, which is why I find it so hard to continue working when it crashes!

Andarist:
This is using the old inspector - we've recently released a new one and the issue seems to be fixed there. You just need to change the URL to https://stately.ai/viz?inspect=1

https://codesandbox.io/s/affectionate-antonelli-0spuz

jebeck:
> This is using the old inspector - we've recently released a new one and the issue seems to be fixed there. You just need to change the URL to https://stately.ai/viz?inspect=1

Yep, you're right, both syntaxes work in the new one. Would you like a PR on the docs and/or the @xstate/inspect package updating the URL? Happy to do those small changes to pay it forward to future people like me who might similarly get frustrated.

davidkpiano:
> > This is using the old inspector - we've recently released a new one and the issue seems to be fixed there. You just need to change the URL to https://stately.ai/viz?inspect=1
[...]

Issue #227, opened 2018-11-10
Unable to use xstate in the browser


TimvdLippe:
## Bug or feature request?
Bug
## Description:
In `StateNode.ts`, there is a reference to `process`. Since a web browser does not have this variable, you can not load xstate using ES modules in a browser.
## (Bug) Expected result:
I can load xstate using ES modules in a browser
## (Bug) Actual result:
`process is not defined`
## (Bug) Potential fix:
This is a similar issue to https://github.com/reduxjs/redux/issues/2907 which I fixed in https://github.com/reduxjs/redux/pull/3143 that correctly generates an ES bundle.
I went ahead and did this locally, but I saw your `es/` folder output actually contains all `tsc` build targets. The fix would therefore be impossible if someone would directly use `es/StateNode.ts`.
I would propose to generate a new bundle called `es/index.mjs` just like Redux, but with the reference to `process` removed.

davidkpiano:
Fix made in `master` - can you please test it out and see if it works for your use-cases?

TimvdLippe:
@davidkpiano Finally had time to verify and yes it has been fixed :tada:

PR #2580, opened 2021-08-30
default to new visualizer URL in @xstate/inspect

jebeck:
As requested in #2542, here's a small change to update the URL for the inspector to the new visualizer. After working with the new viz further and seeing some of its issues, like [this one](https://github.com/statelyai/xstate-viz/issues/227), but also some general fragility*, I think my suggestion would be that you publish this change to @xstate/inspect under a @beta or @next tag so that the beta status of the new viz is clear to users? I have suggested this as well in the docs change in #2579.

*Fragility = times when the visualization stopped showing up, just a blank page. Unfortunately I didn't have time to stop and capture the instances when this happened by extracting the machines into sandboxes... but I'll try to find the time in the future.

### DIFF
--- a/packages/xstate-inspect/src/browser.ts
+++ b/packages/xstate-inspect/src/browser.ts
@@ -61,7 +61,7 @@ export function createDevTools(): XStateDevInterface {
 }
 
 const defaultInspectorOptions: InspectorOptions = {
-  url: 'https://statecharts.io/inspect',
+  url: 'https://stately.ai/viz',
   iframe: () =>
     document.querySelector<HTMLIFrameElement>('iframe[data-xstate]'),
   devTools: () => {
--- a/.changeset/dirty-icons-grow.md
+++ b/.changeset/dirty-icons-grow.md
@@ -1,0 +1,5 @@
+---
+'@xstate/inspect': minor
+---
+
+update @xstate/inspect URL to the new visualizer on stately.ai


packages/xstate-inspect/src/browser.ts
```diff
@@ -61,7 +61,7 @@ export function createDevTools(): XStateDevInterface {
 }
 
 const defaultInspectorOptions: InspectorOptions = {
-  url: 'https://statecharts.io/inspect',
+  url: 'https://stately.ai/viz',
```
Andarist, line 64: We'll be adjusting the public URL this week - probably to something like https://stately.ai/inspect . So until then, I would hold off merging this PR

Andarist:
> I think my suggestion would be that you publish this change to @xstate/inspect under a @beta or @next tag so that the beta status of the new viz is clear to users

I'm afraid that this is not easy to do with our current releasing pipeline. From what I understand the #2542 got fixed and is no longer an issue. We are not aware of any substantial problems with the new inspector so I think it's rather OK to just publish a new version of the `@xstate/inspect` as the latest version. If issues arise when people start using it more we'll just have to deal with them and fix those issues promptly

> *Fragility = times when the visualization stopped showing up, just a blank page. Unfortunately I didn't have time to stop and capture the instances when this happened by extracting the machines into sandboxes... but I'll try to find the time in the future.

It would be highly appreciated if you could try to get exact repro steps for this.

jebeck:
> > I think my suggestion would be that you publish this change to @xstate/inspect under a @beta or @next tag so that the beta status of the new viz is clear to users
> 
> I'm afraid that this is not easy to do with our current releasing pipeline. From what I understand the #2542 got fixed and is no longer an issue. We are not aware of any substantial problems with the new inspector so I think it's rather OK to just publish a new version of the `@xstate/inspect` as the latest version. If issues arise when people start using it more we'll just have to deal with them and fix those issues promptly

Sure, it was just a suggestion. And partly it is also feedback: I don't find the new visualizer adequate to my needs, and I'm still heavily using the old one. The layout is still much more readable in the old visualizer (related to [xstate-viz #227](https://github.com/statelyai/xstate-viz/issues/227) where there are lingering weird edges after the fix...I know further improvements are planned!), and some other things are better, like I'm much more likely to see the initial cascade of events that fire when my app loads in the old visualizer, and that's one of my most frequent sites of debugging. (If I understand correctly, @davidkpiano confirmed that events that don't make it to the visualizer on load are due to time to create connection to the inspector, and I guess the difference between old and new is that new takes longer to make that connection and misses more events, at least in my recent experience.)

Overall, my biggest suggestion is that stately provide clearer documentation about what the statuses of the old vs. new visualizers are. Just a few lines in the @xstate/inspect README with the URLs for both and some verbiage to the effect of "old viz not being developed further, existing bugs won't be fixed; new viz is new and has some bugs, please file as you encounter" (ideally w/link to a template code sandbox for the new inspector) would be super helpful for setting expectations and letting us users help ourselves. (I don't love having to go to a Discord to get questions of this sort answered, and I think self-serve documentation would make it easier for you all as well.)

> 
> > *Fragility = times when the visualization stopped showing up, just a blank page. Unfortunately I didn't have time to stop and capture the instances when this happened by extracting the machines into sandboxes... but I'll try to find the time in the future.
> 
> It would be highly appreciated if you could try to get exact repro steps for this.

I understand that exact repro is desired, and I will try to make adjustments in my own codebases/workflow to make it easier for me to extract them in the future. At present it requires a lot of replacing actions and such with stubs and careful reading to make sure I don't leak any company stuff.

And let me be clearer about _why_ I report things without a repro sometimes, in case this is a source of frustration for y'all. It's *not* because I expect a magic fix! I understand it's very difficult to fix something without exact steps to reproduce, and I don't expect a vague bug report to even result in an _effort_ to try to fix something, really. Rather, what I hope is that my report of frequent problems with the new tool is useful feedback in a general sense—i.e., that some feedback is better than no feedback.

davidkpiano:
This is now outdated as https://github.com/statelyai/inspect is now recommended for XState v5+.

PR #2581, opened 2021-08-30
feat(lit-controller): scaffold lit-controller package

ChrisShank:
POC implementation of https://github.com/statelyai/xstate/discussions/2186

TODO:
- [ ] Implement helpers for `@xstate/fsm`
- [ ] Add tests
- [ ] Add documentation

cc @justinfagnani @santicros


packages/xstate-lit-controller/README.md
```diff
@@ -0,0 +1,29 @@
...
+```html
+<script src="https://unpkg.com/@xstate/lit-controller/dist/xstate-lit-controller.min.js"></script>
+```
+
+By using the global variable `XStateLitController`
+
+or
+
+```html
+<script src="https://unpkg.com/@xstate/vue/dist/xstate-lit-controller.fsm.min.js"></script>
```
ChrisShank, line 24: Fix

ChrisShank:
Is this something that y’all want to add to this repo? Or would it be better to publish it myself?

davidkpiano:
> @davidkpiano Is this something that y’all want to add to this repo? Or would it be better to publish it myself?

Sorry, I'll get to this later today and let you know!

ChrisShank:
No rush, just something i should have asked before i put up a PR! Still a little work on my end to complete it!

packages/xstate-lit-controller/src/ActorController.ts
```diff
@@ -0,0 +1,45 @@
...
+  get state() {
+    return this._state;
+  }
+
+  get send() {
+    return this.actorRef.send;
+  }
+
+  hostDisconnected() {
+    this.subscription.unsubscribe();
```
justinfagnani, line 36: You should also have a `hostConnected()` implementation to re-subscribe. Remember that these callbacks will be called when the element is moved (like with `repeat()`) in addition to when they're removed and discarded.

packages/xstate-lit-controller/src/ActorController.ts
```diff
@@ -0,0 +1,45 @@
...
+    });
+  }
+
+  get state() {
+    return this._state;
+  }
+
+  get send() {
+    return this.actorRef.send;
+  }
```
justinfagnani, line 33: Tearing the `send()` method off like this will possibly give it the wrong `this` reference. It'll also make inlining by the VM harder. Consider doing:

```ts
  send(event, options) {
    return this.actorRef.send(event, options);
  }
```

packages/xstate-lit-controller/src/MachineController.ts
```diff
@@ -0,0 +1,61 @@
...
+  get service() {
+    return this.serviceController.service;
+  }
+
+  get state() {
+    return this.service.state;
+  }
+
+  get send() {
+    return this.service.send;
```
justinfagnani, line 40: same comment about tear-offs as above.

packages/xstate-lit-controller/src/MachineController.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  hostConnected() {
+    this.service.onTransition((state) => {
```
justinfagnani, line 44: Remember that `hostConnected` can be called multiple times. Should you guard that this is only called once? Alternatively, maybe you can do this in the constructor?

packages/xstate-lit-controller/src/SelectorController.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+    });
+  }
+
+  selected() {
+    return this._selected;
+  }
+
+  hostDisconnected() {
+    this.subscription.unsubscribe();
```
justinfagnani, line 41: Add `hostConnected`

packages/xstate-lit-controller/src/ServiceController.ts
```diff
@@ -0,0 +1,55 @@
...
+      typeof getMachine === 'function' ? getMachine() : getMachine;
+    const { state: rehydratedState, ...interpreterOptions } = options;
+
+    this.service = interpret(machine, interpreterOptions).start(
+      rehydratedState ? (State.create(rehydratedState) as any) : undefined
+    );
+  }
+
+  hostDisconnected() {
+    this.service.stop();
```
justinfagnani, line 40: add `hostConnected`

packages/xstate-lit-controller/src/ServiceController.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+export function connectService<
```
justinfagnani, line 44: Add jsdocs here and for other exported APIs?

ChrisShank:
@justinfagnani appreciate the review 🙏

cbarratt:
Hey folks, I wonder if there is any movement on this, as a consumer of xstate/lit I think it'd be super helpful.

Andarist:
This has completely flown under the radar for me - I'll try to review this next week.

Andarist:
Given my limited knowledge about Lit and web components, I really miss usage examples here (that could be tests in the repo 😉 )

cbarratt:
@Andarist - I can't claim to be a lit expert myself, as I've only just started using it in the last month or so! So other folks here can probably input better than me 😅 

I can throw some examples of an app I'm writing at the minute with a main machine + child actor.

export class Checkout extends LitElement {
  service = interpret(CheckoutMachine(), { devTools: true })

  @property({ attribute: false })
  state = this.service.initialState

  connectedCallback(): void {
    super.connectedCallback()

    this.service.onTransition((state, event) => {
      this.state = state
      const label = `[${this.service.id}]`.toUpperCase()
      console.log(label, event, state.value)
    })

    this.service.start({})
  }

  // Stop service on removal from the Light DOM
  disconnectedCallback(): void {
    this.service.stop()
  }

  ....
 Then in my render method, I'm rendering a new lit element for representing a child actor

  render() {
    return html`<handoff .actor=${this.state.context.handOffActor}></handoff>`
  }
```

Then in my other lit elem

```
export class HandOff extends LitElement {
  @property()
  actor: HandOffActor | null = null

  @property()
  state: any

  @property()
  subscription!: Subscription | undefined

  connectedCallback(): void {
    super.connectedCallback()

    this.subscription = this?.actor?.subscribe((emitted) => {
      this.state = emitted
      const label = `[${emitted?.machine?.id}]`.toUpperCase()
      console.log(label, emitted._event.data, emitted.value)
    })
  }

  disconnectedCallback(): void {
    this.subscription?.unsubscribe()
  }
```

using subscribe here I can use `this.state` and get the updated child machine on any transitions/events.

I've yet to correctly manage to type the `this.state` property which is good fun 😅

kiastorm:
Would love to see progress here!

justinfagnani:
@ChrisShank let me know if you're still interested pursuing this and need more review from the Lit side!

davidkpiano:
Thanks for the PR! Doing a bit of cleanup and closing this as outdated - would be happy to accept an updated lit-controller package though 🧹

PR #2582, opened 2021-08-30
Added XState Parser/XState CLI to core

mattpocock:


Andarist:
We've decided to keep this in a separate monorepo: https://github.com/statelyai/xstate-tools

PR #2587, opened 2021-08-31
Fixed an issue with updating options in React's effect

### LINKED ISSUE #2586: [@xstate/react] v1.6.0 breaks merging of service definitions
**Description**

`@xstate/react` v1.6.0 seems to break the ability to define services in your machine if you also include service definitions in your `useMachine` call. Specifically, the service definitions passed to `useMachine` seem to overwrite the service definitions that already existed in the machine definition.

**Expected Result**

I expected the service definitions to be merged, as they were in v1.5.1.

**Actual Result**

The service definitions included in the machine definition appear to be wiped out and you get the following warning:

```
Warning: No service found for invocation 'internalService' in machine 'toggle'. 
```

**Reproduction**

https://codesandbox.io/s/repro-for-xstate-issue-2586-8b3fz

**Additional context**

`xstate 4.23.2`
`@xstate/react 1.6.0`

Andarist:
fixes #2586 (a regression caused by #2241)

i will add a changeset and tests in a few hours

### DIFF
--- a/.changeset/fast-bees-applaud.md
+++ b/.changeset/fast-bees-applaud.md
@@ -1,0 +1,5 @@
+---
+'@xstate/react': patch
+---
+
+Fixed an issue with implementations provided outside of React being wiped out and unusable.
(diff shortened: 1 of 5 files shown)


mattpocock (approved):
Makes sense why this would break!

davidkpiano (approved):

PR #2606, opened 2021-09-02
Fix ContextFrom<T>, EventFrom<T>, EmittedFrom<T>

### LINKED ISSUE #2600: *From utility types returning `never`
**Description**
```typescript
const model = createModel(
  { editions: [] as Edition[], error: null as string | null, invalidFields: [] as string[] },
  {
    events: {
      submit: (values: Record<string, Date | null>) => ({ values }),
    },
  },
);

type Context = ContextFrom<typeof model>;
type Event = EventFrom<typeof model>;
```

The `ContextFrom` and `EventFrom` utility types were working great, but since a recent update they return `never`.

**Expected Result**
I expected them to return the inferred context and event types.

**Actual Result**
They returned `never`

**Reproduction**
Weirdly enough a reproduction in CodeSandbox does not have the same results. I also added my `tsconfig.json` to it.
https://codesandbox.io/s/thirsty-water-dyxqi?file=/src/index.ts 

**Additional context**
The types start breaking for me from version 4.23.2. The project is based on the [Vite React TS](https://github.com/vitejs/vite/tree/main/packages/create-vite/template-react-ts) template.

davidkpiano:
This PR fixes an issue with the `*From<T>` types returning `never`. This was occurring because doing something like:

```ts
T extends Model<infer T, any, any, any>
```

was not working when `T` was actually `Model<Something, never, never, ...>`, since `never` doesn't extend `any` (I think).

The fix was to do this instead:

```ts
T extends Model<infer T, infer _, infer __, infer ___>
```

This also uses the trick that @Andarist proposed in the previous PR 👏 

Fixes #2600

### DIFF
--- a/packages/core/test/model.test.ts
+++ b/packages/core/test/model.test.ts
@@ -544,7 +544,7 @@ describe('createModel', () => {
       }
     );
 
-    const val = (null as unknown) as ContextFrom<typeof model>;
+    const val = ({} as unknown) as ContextFrom<typeof model>;
 
     // expect no type error here
     // with previous ContextFrom behavior, this will not compile
@@ -564,7 +564,7 @@ describe('createModel', () => {
       }
     );
 
-    const val = (null as unknown) as EventFrom<typeof model>;
+    const val = ({} as unknown) as EventFrom<typeof model>;
 
     // expect no type error here
     // with previous ContextFrom behavior, this will not compile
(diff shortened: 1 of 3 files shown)


packages/core/test/model.test.ts
```diff
@@ -535,4 +535,42 @@ describe('createModel', () => {
...
+        events: {
+          INC: () => ({})
+        }
+      }
+    );
+
+    const val = ({} as unknown) as EventFrom<typeof model>;
+
+    // expect no type error here
+    // with previous ContextFrom behavior, this will not compile
```
davidkpiano, line 570:
```suggestion
    // with previous EventFrom behavior, this will not compile
```

Andarist (approved):

PR #2614, opened 2021-09-04
Add useSelector to xstate-svelte

DavKato:
This PR adds useSelector functiont to xstate-svelte.

```svelte
<script>
  // It won't be updated unless the selected value changed.
  const value = useSelector(service, (state) => state.context.value);
</script>

<p>{$value}</p>
```


packages/xstate-svelte/src/useSelector.ts
```diff
@@ -0,0 +1,30 @@
...
+const defaultCompare = (a, b) => a === b;
+
+export const useSelector = <
+  TActor extends ActorRef<any, any>,
+  T,
+  TEmitted = TActor extends Subscribable<infer Emitted> ? Emitted : never
+>(
+  actor: TActor,
+  selector: (emitted: TEmitted) => T,
+  compare: (a: T, b: T) => boolean = defaultCompare
```
farskid, line 13: There is a missing 4th argument `getSnapshot` compared to `@xstate/react`.

farskid (approved):

davidkpiano (approved):

packages/xstate-svelte/src/useSelector.ts
davidkpiano: I think it's fine for that to be added later or emitted; actors passed into `useSelector` should adhere to the `ActorRef` interface (i.e., it's not compatible with other things), so when I added `getSnapshot` as an option, I think I was being too flexible 😓

PR #2620, opened 2021-09-06
change callback terminology in docs

Spankyed:
No code changes made to typescript files with callback terminology. As well, actors.md & stencil.md reference code that contains non trivially changed callback terminology and those sections we're not changed.

### DIFF
--- a/docs/guides/actors.md
+++ b/docs/guides/actors.md
@@ -38,7 +38,7 @@ They may have optional methods:
 All the existing invoked service patterns fit this interface:
 
 - [Invoked promises](./communication.md#invoking-promises) are actors that ignore any received events and send at most one event back to the parent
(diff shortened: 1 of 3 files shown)


mattpocock:
Invoked callback feels more descriptive to me than invoked function - especially because invoked functions can also resolve promises, which introduces confusion.

Spankyed:
Invoked callbacks

> Invoked callback feels more descriptive to me than invoked function - especially because invoked functions can also resolve promises, which introduces confusion.

Invoked callbacks is an incorrect description that adds unneeded complexity, IMO. The function being invoked is not itself a callback, but rather, a callback handler that is passed a callback as its first argument. 

See [discussion #2064](https://github.com/statelyai/xstate/discussions/2604)

mattpocock:
Sorry, I don't think we can make this change as a lot of educational material not on the docs site would instantly become legacy.

Spankyed:
I am ok with this practical decision.

PR #2625, opened 2021-09-07
Update milliseconds to match minutes

programbo:
Corrects a typo where the millisecond/minute conversion doesn’t match.


programbo:
Fixed by https://github.com/statelyai/xstate/pull/2626

PR #2632, opened 2021-09-10
[core] Allow any action if action creators are not specified in model

### LINKED ISSUE #2619: `createModel` second param cause TS errors
**Description**
Second parameter in `createModel` cause type errors in machine.

For example:

```
import { createModel } from "xstate/lib/model";

export const model = createModel(
  { apiKey: "" },
  // If you remove this everything works just fine
  {
    /* ... */
  }
);

export const machine = model.createMachine(
  {
    initial: "ready",
    states: {
      ready: {
        // Type 'string' is not assignable to type 'BaseActions ...
        entry: "initWidget",
      },
    },
  },
  {
    actions: {
      // Parameter 'ctx' implicitly has an 'any' type.
      initWidget: (ctx) => {
        console.log(ctx.apiKey);
      },
    },
  }
);
```

**Reproduction**

https://codesandbox.io/s/dark-darkness-5pwig?file=/src/index.ts

**Additional context**

 - `xstate`: `4.23.4`
 - `typescript`: `4.2.3` / `4.4.2`

davidkpiano:
This PR fixes a regression noticed by some users - when they specified `events` but not `actions` in `createModel(...)`, actions would be typed as `never`:

```ts
    const model = createModel(
      {},
      {
        events: {}
      }
    );

    model.createMachine({
      // TS will not compile
      entry: 'someAction',
      exit: { type: 'someObjectAction' },
    });
```

Fixes #2619

### DIFF
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -22,9 +22,11 @@ export function createModel<
   TComputedEvent = 'events' extends keyof TFinalModelCreators
     ? UnionFromCreatorsReturnTypes<TFinalModelCreators['events']>
     : never,
-  TComputedAction = 'actions' extends keyof TFinalModelCreators
-    ? UnionFromCreatorsReturnTypes<TFinalModelCreators['actions']>
-    : never
+  TComputedAction = 'actions' extends keyof TModelCreators
+    ? 'actions' extends keyof TFinalModelCreators
+      ? UnionFromCreatorsReturnTypes<TFinalModelCreators['actions']>
+      : any
+    : any
 >(
   initialContext: TContext,
   creators: TModelCreators
--- a/.changeset/slow-timers-nail.md
+++ b/.changeset/slow-timers-nail.md
@@ -1,0 +1,20 @@
+---
+'xstate': patch
+---
+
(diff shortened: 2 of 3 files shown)


farskid (approved):

packages/core/src/model.ts
```diff
@@ -22,9 +22,11 @@ export function createModel<
   TComputedEvent = 'events' extends keyof TFinalModelCreators
     ? UnionFromCreatorsReturnTypes<TFinalModelCreators['events']>
     : never,
-  TComputedAction = 'actions' extends keyof TFinalModelCreators
-    ? UnionFromCreatorsReturnTypes<TFinalModelCreators['actions']>
-    : never
+  TComputedAction = 'actions' extends keyof TModelCreators
```
Andarist, line 25: This change has completely broken the inference of actions:
```ts
const model = createModel({ foo: 100 }, {
  events: {
    BAR: () => ({})
  }
})

model.createMachine({
  // ctx is being any
  entry: (ctx) => {},
  exit: assign({
    // ctx is being unknown
    foo: (ctx) => 42
  })
})
```

Issue #2138, opened 2021-04-23
Passing DOM event crashes when inspector is enabled


SimeonC:
**Description**
It seems that any usage of `onFocus={() => send({ type: 'ANY', data: event });}` will cause the inspector to crash.
It seems from my local problems that this is some sort of circular reference.

**Expected Result**
No crash to happen. Like we handle 'circular data' we could also just make the inspector do `event is Event ? '[DOM Event]' : event`.

**Actual Result**
The app and inspector crashes!

**Reproduction**
https://codesandbox.io/s/xstateinspect-bug-with-events-mr0wd

ronaldborman:
@davidkpiano I just have experienced this bug when trying to send a DOM event directly. Initially I was looking in other directions before seeing this issue. A warning in the inspect documentation would be useful imho.

Andarist:
I've been discussing this briefly last week with David. One of the options to solve this would be to accept an option like `serialize` that would allow us to serialize complex data types before sending them through serialization boundaries like 2 different windows. We'd probably have to ship some built-in, simple, serializer but allow this to be configured.

davidkpiano:
This addition has been made here: https://github.com/statelyai/xstate/pull/2640

So you will be able to do something like this:

```js
inspect({
  // ...
  serialize: (key, value) => {
    if (value instanceof UIEvent) {
      return { type: value.type }
    }
    return value;
  }
});
```

PR #2640, opened 2021-09-12
[@xstate/inspect] Add `.serialize`

davidkpiano:
This PR adds `.serialize` as an option to `inspect`.

A serializer can now be specified as an option for `inspect(...)` in the `.serialize` property. It should be a [replacer function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter):

```js
// ...

inspect({
  // ...
  serialize: (key, value) => {
    if (value instanceof Map) {
      return 'map';
    }

    return value;
  }
});

// ...

// Will be inspected as:
// {
//   type: 'EVENT_WITH_MAP',
//   map: 'map'
// }
someService.send({
  type: 'EVENT_WITH_MAP',
  map: new Map()
});
```

Fixes https://github.com/statelyai/xstate/issues/2138


packages/xstate-inspect/src/browser.ts
```diff
@@ -138,19 +139,24 @@ export function inspect(
...
 
+  const stringifyWithSerializer = (value: any) =>
+    stringify(value, options?.serialize);
+
   devTools.onRegister((service) => {
     inspectService.send({
       type: 'service.register',
-      machine: stringify(service.machine),
-      state: stringify(service.state || service.initialState),
+      machine: stringifyWithSerializer(service.machine),
```
Andarist, line 148: I think we should only run the serializer over "user values" - the vast majority of `service.machine` itself is needed untouched for us to work with the inspector etc. 

I see how this solves the problem of serializing context embedded in the machine, parametrized actions, and stuff like this but it feels a little bit iffy to just run the custom serializer over the whole thing. 

What do you think? Either way - I'm OK with this right now but would prefer to add a TODO comment or something here for the future, unless you think that my concerns are not valid.

packages/xstate-inspect/src/browser.ts
```diff
@@ -138,19 +139,24 @@ export function inspect(
     inspectService.send({ type: 'unload' });
   });
 
+  const stringifyWithSerializer = (value: any) =>
+    stringify(value, options?.serialize);
```
Andarist, line 143: nit: all other options are "unpacked" at the top of the outer function - maybe the same could be done with this to keep the style consistent?

packages/xstate-inspect/src/browser.ts
```diff
@@ -138,19 +139,24 @@ export function inspect(
...
+
   devTools.onRegister((service) => {
     inspectService.send({
       type: 'service.register',
-      machine: stringify(service.machine),
-      state: stringify(service.state || service.initialState),
+      machine: stringifyWithSerializer(service.machine),
+      state: stringifyWithSerializer(service.state || service.initialState),
```
Andarist, line 149: In a similar vein as above - we expect a lot of the state to stay untouched 🤔

packages/xstate-inspect/src/types.ts
```diff
@@ -6,10 +6,13 @@ export type MaybeLazy<T> = T | (() => T);
 
 export type ServiceListener = (service: Interpreter<any>) => void;
 
+export type Replacer = (key: string, value: any) => any;
+
 export interface InspectorOptions {
   url: string;
   iframe: MaybeLazy<HTMLIFrameElement | null | false>;
   devTools: MaybeLazy<XStateDevInterface>;
+  serialize: Replacer | undefined;
```
Andarist, line 15: Usually, such types can be consumed by people - and making all of those fields required make it harder so. I would probably recommend making the optional things optional here and creating a separate that for "resolved" options that would only be used within our implementation

farskid (approved):

packages/xstate-inspect/src/browser.ts
davidkpiano: There isn't an easy way to do this if we want to target parameterized actions, context, etc. Let's revisit this if it becomes a problem.

davidkpiano:
@Andarist Re-review? Let's save the selective serialization of the machine/state because it's probably a premature optimization at this point.

davidkpiano:
@Andarist An original idea I had for this (which you didn't agree with at the time, but might now?) was to separate the serialization functions:

```diff
inspect({
- serialize: (key, value) => { ... },
+ serializeState: (key, value) => { ... },
+ serializeEvent: (key, value) => { ... }
});
```

This addresses the concern of serializing the entire machine/state. What do you think?

davidkpiano:
@Andarist I added selective serialization

packages/xstate-inspect/src/serialize.ts
```diff
@@ -2,47 +2,34 @@ import { State, StateMachine } from 'xstate';
...
+    selected[key] = value[key];
+  }
 
-  return {
-    ...state.toJSON(),
-    context: replacer('context', state.context),
-    event: scxmlEvent.data,
-    _event: scxmlEvent
-  };
+  const serialized = JSON.parse(stringify(selected, replacer));
```
davidkpiano, line 16: This is the simplest way to get the replacer to work _recursively_.

packages/xstate-inspect/src/browser.ts
```diff
@@ -94,7 +94,7 @@ export function inspect(
     return undefined;
   }
 
-  const inspectMachine = createInspectMachine(devTools);
+  const inspectMachine = createInspectMachine(devTools, options ?? {});
```
Andarist, line 97: nit: probably better to handle this as a default param within `createInspectMachine` since it doesn't really require any options

packages/xstate-inspect/src/types.ts
```diff
@@ -6,13 +6,17 @@ export type MaybeLazy<T> = T | (() => T);
...
+export type Replacer = (key: string, value: any) => any;
+
 export interface InspectorOptions {
-  url: string;
-  iframe: MaybeLazy<HTMLIFrameElement | null | false>;
-  devTools: MaybeLazy<XStateDevInterface>;
+  url?: string;
+  iframe?: MaybeLazy<HTMLIFrameElement | null | false>;
+  devTools?: MaybeLazy<XStateDevInterface>;
+  serialize?: Replacer | undefined;
```
Andarist, line 15: nit: the different naming of the property and the type is slightly confusing

packages/xstate-inspect/src/browser.ts
```diff
@@ -175,13 +180,14 @@ export function inspect(
...
-      // filter out synchronous notification from within `.start()` call when the `service.state` has not yet been assigned
+      // filter out synchronous notification from within `.start()` call
+      // when the `service.state` has not yet been assigned
       if (state === undefined) {
         return;
       }
       inspectService.send({
         type: 'service.state',
-        state: stringify(state),
+        state: stringifyState(state, options?.serialize),
```
Andarist, line 190: Note for a future improvement - when sending things through `postMessage` the browser uses "structured clone algorithm" (a browser API to use this algorithm, `structuredClone`, is becoming standardized lately) and this algorithm handles cyclic structures just fine. By dumping this to a string we lose information here - and maybe there would be a way to not do that in the future.

packages/xstate-inspect/src/browser.ts
```diff
@@ -138,19 +138,24 @@ export function inspect(
...
+  const stringifyWithSerializer = (value: any) =>
+    stringify(value, options?.serialize);
+
   devTools.onRegister((service) => {
+    const state = service.state || service.initialState;
     inspectService.send({
       type: 'service.register',
-      machine: stringify(service.machine),
-      state: stringify(service.state || service.initialState),
+      // TODO: determine how only parts of the machine can go through serializer
```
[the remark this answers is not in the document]
davidkpiano: Nope!
```suggestion
```

packages/xstate-inspect/src/types.ts
davidkpiano: It is a replacer (copying the same type from `JSON.stringify(value, replacer, spacing)`), and `serialize` is how it's commonly known.

packages/xstate-inspect/src/browser.ts
davidkpiano:
```suggestion
        // TODO: investigate usage of structuredClone in browsers if available
        state: stringifyState(state, options?.serialize),
```

Andarist (approved):

PR #2645, opened 2021-09-14
fix: return empty context from event.type protective code

pke:
It seems the TS compiler is complaining about the return value. A simple return is not allowed.
Returning and empty object hash works though.

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -248,7 +248,7 @@ The reason this errors is because inside the `consoleLogData` function, we don't
 createMachine<Context, Event>(machine, {
   actions: {
     consoleLogData: (context, event) => {
-      if (event.type !== 'EVENT_WITH_FLAG') return
+      if (event.type !== 'EVENT_WITH_FLAG') return {}
       // No more error at .flag!
       console.log(event.flag);
     };


davidkpiano:
That's strange, it shouldn't be complaining about that... cc. @mattpocock

Andarist:
@pke could you share a repro case? I can't see the problem with what we have in the docs: https://codesandbox.io/s/modest-microservice-54gsm

pke:
```
Argument of type '(context: Context, event: Event) => { guesses: string; solved: string; remaining: number; } | undefined' is not assignable to parameter of type 'Assigner<Context, Event> | PropertyAssigner<Context, Event>'.
  Type '(context: Context, event: Event) => { guesses: string; solved: string; remaining: number; } | undefined' is not assignable to type 'Assigner<Context, Event>'.
    Type '{ guesses: string; solved: string; remaining: number; } | undefined' is not assignable to type 'Partial<Context>'.
      Type 'undefined' is not assignable to type 'Partial<Context>'.ts(2345)
type Context = {
    word: string;
    solved: string;
    guesses: string;
    remaining: number;
    maxTries: number;
    animation: string;
}
```

Screenshot 2021-09-14 at 22 07 32

Andarist:
In your screenshot, you are having a problem with an `assign` action, while this PR edits a plain function. That being said we could allow `undefined` return here at the type level but I wonder if that wouldn't make users catch fewer problems with their code since one could unintentionally forget to return something here.

pke:
Ah you are right, total oversight from my side, sorry!

PR #2651, opened 2021-09-14
Added the ability to tighten TS declarations of machine via generated metadata

mattpocock:
https://www.loom.com/share/34bfa23a60f24cd985ea08acaee93b21


packages/core/src/Machine.ts
```diff
@@ -61,6 +62,18 @@ export function createMachine<
...
+  TEvent extends EventObject,
+  TMeta extends TypegenMeta
+>(
+  // Ensure that only the first overload matches models, and prevent
+  // accidental inference of the model as the `TContext` (which leads to cryptic errors)
+  config: TContext extends Model<any, any, any, any>
+    ? never
+    : MachineConfig<TContext, any, TEvent> & { types: TMeta },
+  options?: Partial<MaybeTypegenMachineOptions<TContext, TEvent, TMeta>>
+): StateMachine<TContext, any, TEvent, any, any, TMeta>;
```
mattpocock, line 76: We should try to have fewer overloads, not more - I wonder if there's a way we can add this to a different overload?

packages/core/src/State.ts
```diff
@@ -298,7 +300,9 @@ export class State<
    * @param parentStateValue
    */
   public matches<TSV extends TTypestate['value']>(
-    parentStateValue: TSV
+    parentStateValue: TMeta extends IsTypegenActive
+      ? TMeta['matchesStates']
+      : TSV
```
mattpocock, line 305: I had to do a bit of trampling over `TTypestate` here - basically overriding the inference in favour of the meta info we get from `TMeta['matchesStates']`

mattpocock:
Looks like I have some work to do with getting this working on models

packages/core/src/Machine.ts
davidkpiano: We should deprecate (or straight up remove) `createMachine<typeof model>`. We can detect if a model is passed in through the original type (the one below this one) and warn.

packages/core/src/State.ts
```diff
@@ -316,7 +321,9 @@ export class State<
    * Whether the current state configuration has a state node with the specified `tag`.
    * @param tag
    */
-  public hasTag(tag: string): boolean {
+  public hasTag(
+    tag: TMeta extends IsTypegenActive ? TMeta['tags'] : string
```
davidkpiano, line 325: Nit: maybe we can condense this with a special type:

```js
tag: WithTypegen<TMeta, 'tags', string>
```

packages/core/src/Machine.ts
mattpocock: Let's straight up remove it. `createMachine<typeof model>()` is super easy to change to `model.createMachine()`.

mattpocock: Added in #2657

packages/core/src/State.ts
mattpocock: Having fiddled around with implementing it, I think this:

image

reads better than this:

image

farskid (approved):

packages/core/src/State.ts
```diff
@@ -297,17 +299,22 @@ export class State<
    * Whether the current state value is a subset of the given parent state value.
    * @param parentStateValue
    */
-  public matches<TSV extends TTypestate['value']>(
-    parentStateValue: TSV
+  public matches<TSV extends TTypestate['value'] = never>(
```
Andarist, line 302: What's that `= never` here for? 🤔 it looks a little bit fishy

packages/core/src/State.ts
```diff
@@ -297,17 +299,22 @@ export class State<
...
+    parentStateValue: TMeta extends TypegenIsActive
+      ? TMeta['matchesStates']
+      : TSV
   ): this is State<
-    (TTypestate extends any
+    (TMeta extends TypegenIsActive
+      ? never
```
Andarist, line 308: could you explain this part? won't this result in `TContext` to be given as `never` to the resulting `State`?

mattpocock: Apologies, left this in accidentally.

packages/core/src/State.ts
mattpocock: Also in accidentally, good eyes

Andarist (commented):
This is a deep-dive, technical PR - it really could use some writeup about the problem, proposed solution, design of the `TMeta` and potentially tricks used in the implementation.

mattpocock:
@Andarist Where you do you want me to add it? In the PR description?

packages/core/src/Machine.ts
```diff
@@ -50,23 +51,26 @@ export function Machine<
...
-  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext },
+  TMeta extends TypegenMeta = DefaultTypegenMeta
 >(
   config: TContext extends Model<any, any, any, any>
     ? 'Model type no longer supported as generic type. Please use `model.createMachine(...)` instead.'
-    : MachineConfig<TContext, any, TEvent>,
-  options?: Partial<MachineOptions<TContext, TEvent>>
-): StateMachine<TContext, any, TEvent, TTypestate>;
+    : MachineConfig<TContext, any, TEvent> & { types?: TMeta },
```
Andarist, line 59: q: why don't we just put `types?: Meta` in the `MachineConfig`?

packages/core/src/State.ts
```diff
@@ -21,6 +21,7 @@ import { StateNode } from './StateNode';
 import { getMeta, nextEvents } from './stateUtils';
 import { initEvent } from './actions';
 import { IS_PRODUCTION } from './environment';
+import { DefaultTypegenMeta, TypegenIsActive, TypegenMeta } from '.';
```
Andarist, line 24: It's best to avoid cycles in the module graph
```suggestion
import { DefaultTypegenMeta, TypegenIsActive, TypegenMeta } from './types';
```

packages/core/src/Machine.ts
mattpocock: We could do, but frankly I couldn't be bothered to go through and add all the generics. If this is the only entry point (at the root of the config), we should only allow it to be passed in here.

mattpocock:
@Andarist added a deep-dive loom

packages/core/src/Machine.ts
Andarist: And `MachineConfig` only represents the root so it IMHO makes sense for it to be included there. This also improves some minor things like jump to type declaration, gets displayed better in tooltips and stuff.

mattpocock: OK, will add it there

mattpocock: I've just added it, but there's one thing on my mind... Let's say that a user a defines their machine config, typed like this, outside of the machine call:

```ts
const config: MachineConfig<Context, any, Event> = {};
```

They'd now have to change the declaration to this:

```ts
const config: MachineConfig<Context, any, Event, any, Meta> = {};
```

Which is not particularly fun.

Andarist: Is it a problem though? you have added a default type param to `MachineConfig`

packages/core/src/types.ts
```diff
@@ -740,6 +740,119 @@ export type DelayConfig<TContext, TEvent extends EventObject> =
...
+export type RequiredTypegenMachineOptions<
+  TContext,
+  TEvent extends EventObject,
+  TMeta extends TypegenMeta
+> = Required<
+  Pick<
+    TypegenMachineOptions<TContext, TEvent, TMeta, 1>,
+    GetRequiredTypegenOptionKeys<TMeta>
+  >
+>;
```
Andarist, line 752: As far as I can tell this `Required` is redundant because all the keys of `TypegenMachineOptions` are already required
```suggestion
> = Pick<
  TypegenMachineOptions<TContext, TEvent, TMeta, 1>,
  GetRequiredTypegenOptionKeys<TMeta>
>;
```

packages/core/src/Machine.ts
mattpocock: I mean they'd need to do it if they wanted the inference to carry over to the machine. But maybe we just don't mind that much.

packages/core/src/types.ts
```diff
@@ -740,6 +740,117 @@ export type DelayConfig<TContext, TEvent extends EventObject> =
...
+export interface TypegenMachineOptions<
+  TContext,
+  TEvent extends EventObject,
+  TMeta extends TypegenMeta,
+  TRequired extends 0 | 1 = 0
+> {
+  actions: TRequired extends 1
+    ? {
+        // @ts-ignore
+        [K in TMeta['requiredActions']]: Pick<
```
Andarist, line 811:
```suggestion
        [K in NonNullable<TMeta['requiredActions']>]: Pick<
```

mattpocock: Added

packages/core/src/Machine.ts
Andarist: We'd have to see what would happen with:
```js
const config: MachineConfig<Context, any, Event> = {};
createMachine({ ...config, types: {} as Meta })
```
but either way, I think it's OK - I don't think the `Meta` should be used manually anyway because that's just asking for problems

davidkpiano:
Huge nit: Can we rename this to `TTypeMeta` instead of `TMeta`, as `TMeta` makes me think this is for the `meta` property?

packages/core/src/model.types.ts
```diff
@@ -37,10 +37,21 @@ export interface Model<
...
   reset: () => AssignAction<TContext, any>;
-  createMachine: (
-    config: MachineConfig<TContext, any, TEvent, TAction>,
-    implementations?: Partial<MachineOptions<TContext, TEvent, TAction>>
-  ) => StateMachine<TContext, any, TEvent>;
+  createMachine: {
+    <TMeta extends DefaultTypegenMeta>(
+      config: MachineConfig<TContext, any, TEvent, TAction, TMeta>,
+      implementations?: Partial<
+        MaybeTypegenMachineOptions<TContext, TEvent, TMeta, TAction>
```
davidkpiano, line 44: `TMeta` and `TAction` switched here is a bit confusing (nit) - whereas it's `TAction` and `TMeta` above

packages/core/src/types.ts
```diff
@@ -740,6 +740,113 @@ export type DelayConfig<TContext, TEvent extends EventObject> =
   | number
   | DelayExpr<TContext, TEvent>;
 
+export type RequiredTypegenMachineOptions<
```
davidkpiano, line 743: Another nit: wondering if we should put this into a `types/metaTypes.ts` file or similar (the one huge file with all the types is my fault 🙈)

packages/xstate-react/src/useInterpret.ts
```diff
@@ -39,16 +42,37 @@ function toObserver<T>(
...
+        getMachine: MaybeLazy<
+          StateMachine<TContext, any, TEvent, TTypestate, any, TMeta>
+        >,
+        options?: Partial<InterpreterOptions> &
+          Partial<UseMachineOptions<TContext, TEvent>> &
+          MaybeRequiredTypegenMachineOptions<TContext, TEvent, TMeta>,
+        observerOrListener?:
+          | Observer<State<TContext, TEvent, any, TTypestate, TMeta>>
+          | ((value: State<TContext, TEvent, any, TTypestate, TMeta>) => void)
+      ]
```
davidkpiano, line 74: Nit: can this be simplified? 😅

packages/xstate-react/src/useMachine.ts
```diff
@@ -59,16 +62,30 @@ export interface UseMachineOptions<TContext, TEvent extends EventObject> {
...
+          MaybeRequiredTypegenMachineOptions<TContext, TEvent, TMeta>
+      ]
+    : [
+        getMachine: MaybeLazy<
+          StateMachine<TContext, any, TEvent, TTypestate, any, TMeta>
+        >,
+        options?: Partial<InterpreterOptions> &
+          Partial<UseMachineOptions<TContext, TEvent>> &
+          MaybeRequiredTypegenMachineOptions<TContext, TEvent, TMeta>
+      ]
```
davidkpiano, line 84: Nit: same here (simplification) Worried about how this will read on hover for end-users too

mattpocock:
@Andarist ideas around how we can infer `done.invoke.*` events:

https://www.loom.com/share/de46bd18c393411e8c633eb00dfb671d

mattpocock:
@Andarist We should extend `DoneInvokeEvent` so that it takes the name of the service too:

```ts
export interface DoneInvokeEvent<TData, TServiceName extends string = never>
  extends EventObject {
  data: TData;
  type: TServiceName extends string ? `done.invoke.${TServiceName}` : any;
}
```

packages/xstate-react/src/useMachine.ts
mattpocock: Because the members of the tuple are labelled, it actually reads really nicely (and normally)

packages/xstate-react/src/useInterpret.ts
mattpocock: We can probably condense options and observerOrListener into utility types.

Andarist:
This work is included in https://github.com/statelyai/xstate/pull/2674

PR #2657, opened 2021-09-15
Removed the model overload from createMachine

mattpocock:


packages/core/src/Machine.ts
```diff
@@ -69,7 +55,7 @@ export function createMachine<
   // Ensure that only the first overload matches models, and prevent
   // accidental inference of the model as the `TContext` (which leads to cryptic errors)
   config: TContext extends Model<any, any, any, any>
-    ? never
+    ? 'Passing typeof model to the first generic of createMachine is no longer supported - use model.createMachine instead'
```
mattpocock, line 58: @Andarist will this work as intended?

mattpocock: Also, there are now only two overloads left in this file. They look very similar in signature - can we condense them into... only... ONE OVERLOAD...?

mattpocock:
```ts
  TContext,
  TEvent extends EventObject = AnyEventObject,
>(
  // Ensure that only the first overload matches models, and prevent
  // accidental inference of the model as the `TContext` (which leads to cryptic errors)
  config: TContext extends Model<any, any, any, any>
    ? 'Passing typeof model to the first generic of createMachine is no longer supported - use model.createMachine instead'
    : MachineConfig<TContext, any, TEvent>,
  options?: Partial<MachineOptions<TContext, TEvent>>
): StateMachine<TContext, any, TEvent, TTypestate>;
  TContext,
  TEvent extends EventObject = AnyEventObject,
>(
  config: MachineConfig<TContext, any, TEvent>,
  options?: Partial<MachineOptions<TContext, TEvent>>
): StateMachine<TContext, any, TEvent, TTypestate> {
  return new StateNode<TContext, any, TEvent, TTypestate>(
    config,
    options
  ) as StateMachine<TContext, any, TEvent, TTypestate>;
}
```

.changeset/weak-zebras-pull.md
```diff
@@ -0,0 +1,21 @@
+---
+"xstate": minor
+---
+
+Removed the ability to pass a model as a generic to `createMachine`, in favour of `model.createMachine`. This lets us cut an overload from the definition of `createMachine`, meaning errors become more targeted and less cryptic.
```
Andarist, line 5: This was already being implemented in https://github.com/statelyai/xstate/pull/2518/files#diff-00f777398251992f42d7061fc6a79e0c6e7c8aebfd257cb1d8c04129c84e0273R51 and accepted back then so I think it's an OK change.

packages/core/src/Machine.ts
```diff
@@ -69,7 +55,7 @@ export function createMachine<
   // Ensure that only the first overload matches models, and prevent
```
Andarist, line 55: this comment should be adjusted now (or even removed)

packages/core/src/Machine.ts
Andarist: Note that `createMachine` has only a single overload here now - the second thing is like a private declaration. If possible you could still consolidate both though

Andarist:
@mattpocock this test should be removed now:
https://github.com/statelyai/xstate/blob/b0799ac8e68cabe36eff5ad61be3f0aec9314a24/packages/core/test/model.test.ts#L470-L483

mattpocock:
@Andarist good spot, removed

Andarist (approved):

davidkpiano (approved):

farskid (approved):

PR #2658, opened 2021-09-15
Added createSelector

mattpocock:


packages/xstate-react/src/createSelector.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Allows you to specify a typesafe selector based on a machine definition.
+ *
+ * @example
+ * const getIsToggedOn = createSelector(machine, state => state.matches('toggledOn'));
+ */
+export const createSelector = <TMachine extends StateMachine<any, any, any>, T>(
+  _machine: TMachine,
+  selector: (state: StateFrom<TMachine>) => T
```
Andarist, line 11: We could save TS from resolving this through a helper since we know what the input type is:
```suggestion
  selector: (state: TMachine['initialState']) => T
```

packages/xstate-react/src/createSelector.ts
```diff
@@ -0,0 +1,12 @@
...
+) => selector;
```
Andarist, line 12: I wonder if it wouldnt be more ergonomic to embed `useSelector` directly in this:
```suggestion
import { StateMachine, ActorRefFrom } from 'xstate';
import { useSelector } from './useSelector';

/**
 * Allows you to specify a typesafe `useSelector` based on a machine definition.
 *
 * @example
 * const useIsToggedOn = createUseSelector(machine, state => state.matches('toggledOn'));
 */
export const createUseSelector = <TMachine extends StateMachine<any, any, any>, T>(
  _machine: TMachine,
  selector: (state: TMachine['initialState']) => T
) => (actorRef: ActorRefFrom<TMachine>) => useSelector(actorRef, selector);
```

Andarist: Especially that this way it makes more sense for this to live in the `@xstate/react`. The previous version had no dependency on React whatsoever 🤔

mattpocock: @Andarist I have thoughts about this! This is worse because you sometimes want to write selectors which depend on other selectors, and compose them. `createSelector` is a pattern common in Redux, and could also be used with `reselect` if you fancied it.

Andarist: Ok, I see that point but them it would IMHO make sense to move this to the `xstate` package itself - since this might be useful for all of the "integration" packages like `@xstate/react`, `@xstate/vue`, etc. Or maybe to a separate package that those integration packages would reexport?

Andarist (changes requested):
It would be great if you could add some tests for this feature.

packages/xstate-react/src/createSelector.ts
mattpocock: Yep, seems like it'll be useful in all packages. I think it'll be useful enough (and small enough) that it can fit into core, but also happy to put it into a separate package.

.changeset/fresh-pens-walk.md
```diff
@@ -0,0 +1,27 @@
...
+import { useInterpret, useSelector, createSelector } from '@xstate/react';
+
+const machine = createMachine({
+  initial: 'toggledOn',
+  states: {
+    toggledOn: {}
+  }
+});
+
+const getIsToggedOn = createSelector(machine, (state) =>
```
tatchi, line 18:
```suggestion
const getIsToggledOn = createSelector(machine, (state) =>
```

packages/xstate-react/src/createSelector.ts
```diff
@@ -0,0 +1,12 @@
+import { StateFrom, StateMachine } from 'xstate';
+
+/**
+ * Allows you to specify a typesafe selector based on a machine definition.
+ *
+ * @example
+ * const getIsToggedOn = createSelector(machine, state => state.matches('toggledOn'));
```
tatchi, line 7:
```suggestion
 * const getIsToggledOn = createSelector(machine, state => state.matches('toggledOn'));
```

packages/xstate-react/src/createSelector.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Allows you to specify a typesafe selector based on a machine definition.
+ *
+ * @example
+ * const getIsToggledOn = createSelector(machine, state => state.matches('toggledOn'));
+ */
+export const createSelector = <TMachine extends StateMachine<any, any, any>, T>(
+  _machine: TMachine,
+  selector: (state: TMachine['initialState']) => T
+) => selector;
```
davidkpiano, line 12: Is it possible to have `createSelector` generic enough for any actor? This is more of a `createMachineSelector` 🤔

mattpocock: Yes true, it should probably use an ActorRef and grab the TEmitted

mattpocock: @davidkpiano What would the first parameter take? Agree, this is more like a createMachineSelector. Not sure it can take an actor ref, because it needs to be created from the actor _specification_, not the actor. I.e. from the machine, not the running service.

mattpocock:
@Andarist @davidkpiano Added tests and moved the function into core

packages/xstate-react/src/createSelector.ts
mattpocock: Put it in core.

Andarist:
Spiritually the ideas from this PR were incorporated into https://github.com/statelyai/xstate/pull/3778

PR #2659, opened 2021-09-15
Fixed a regression in the inline actions type inference

Andarist:
fixes a regression from here: https://github.com/statelyai/xstate/pull/2632/files#r709015048

### DIFF
--- a/packages/core/src/model.ts
+++ b/packages/core/src/model.ts
@@ -7,7 +7,9 @@ import {
   UnionFromCreatorsReturnTypes,
   FinalModelCreators,
   Model,
-  ModelCreators
+  ModelCreators,
+  Prop,
+  IsNever
 } from './model.types';
 
 export function createModel<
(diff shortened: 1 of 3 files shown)


packages/core/src/model.ts
```diff
@@ -22,11 +22,9 @@ export function createModel<
   TComputedEvent = 'events' extends keyof TFinalModelCreators
     ? UnionFromCreatorsReturnTypes<TFinalModelCreators['events']>
     : never,
-  TComputedAction = 'actions' extends keyof TModelCreators
-    ? 'actions' extends keyof TFinalModelCreators
-      ? UnionFromCreatorsReturnTypes<TFinalModelCreators['actions']>
-      : any
```
Andarist, line 28: Using `any` should be almost forbidden in our types at the "output" positions. It has a domino effect on types and the inference - any unions containing that etc are becoming `any` themselves with that.

farskid (approved):

davidkpiano (approved):

PR #2665, opened 2021-09-16
[docs] Add docs for `state.can()`

davidkpiano:


docs/guides/states.md
```diff
@@ -217,6 +219,52 @@ const canGo = state.hasTag('go');
 // => `true` if in 'green' or 'yellow' state
 ```
 
+### `state.can(event)`
+
+_Since 4.25.0_
+
+This method determines whether an `event` will cause a state change if sent to the interpreted machine. It will return `true` if the state will change due to the `event` being sent; otherwise `false`:
```
farskid, line 226: can we get a reference here as to what we mean by a state change?
https://xstate.js.org/docs/guides/states.html#state-changed

farskid (approved):

docs/guides/states.md
mattpocock:
```suggestion
This method determines whether an `event` will cause a change in state (or context) if sent to the interpreted machine. It will return `true` if the state will change due to the `event` being sent; otherwise `false`:
```

mattpocock (approved):

PR #2668, opened 2021-09-17
Added some model-related typings tests in response to a possible bug report this morning

mattpocock:


packages/core/test/model.test.ts
```diff
@@ -555,6 +561,24 @@ describe('createModel', () => {
...
+
+      const machine = model.createMachine({
+        context: model.initialContext,
+        states: {
+          a: {}
+        }
+      });
+
+      // This would error if it didn't match
+      const service: ActorRefFrom<typeof machine> = interpret(machine);
```
davidkpiano, line 578: Might need to `noop(service)` to prevent CI from failing here (unused variable)

Andarist:
We removed the `createModel` API in v5 ([here](https://github.com/statelyai/xstate/pull/3187))

PR #2669, opened 2021-09-17
Add `when` action creator

Silverwolf90:
As discussed in https://github.com/statelyai/xstate/discussions/2662, this PR adds a `when` action creator to simplify the creation of single-branch `choose` actions.

Let me know if I need to make any changes!


packages/core/test/actions.test.ts
```diff
@@ -1423,6 +1430,20 @@ describe('choose', () => {
...
+    const cond = 'checkCondition';
+    const actions = ['doThing', 'doOtherThing'];
+
+    const chooseAction = when(cond, actions);
+
+    expect(chooseAction).toEqual({
+      type: actionTypes.choose,
+      conds: [{ cond, actions }]
+    });
+  });
```
davidkpiano, line 1444: Can we add a test that this works in an actual machine? I can help with this.

packages/core/test/actions.test.ts
```diff
@@ -1423,6 +1430,20 @@ describe('choose', () => {
...
+});
```
davidkpiano, line 1445:
```suggestion
describe('when', () => {
  it('should execute an action only if the condition is true', (done) => {
    let oneExecuted = false;
    let twoExecuted = false;

    const machine = createMachine<{ someCondition: boolean }>({
      context: {
        someCondition: true
      },
      entry: [
        when(
          (ctx) => ctx.someCondition,
          () => {
            oneExecuted = true;
          }
        ),
        when(
          (ctx) => !ctx.someCondition,
          () => {
            twoExecuted = true;
          }
        )
      ]
    });

    interpret(machine)
      .onTransition(() => {
        if (oneExecuted && !twoExecuted) {
          done();
        }
      })
      .start();
  });

  it('should work with named guards', (done) => {
    let oneExecuted = false;
    let twoExecuted = false;

    const machine = createMachine<{ someCondition: boolean }>(
      {
        context: {
          someCondition: true
        },
        entry: [
          when('someCondTrue', () => {
            oneExecuted = true;
          }),
          when('someCondFalse', () => {
            twoExecuted = true;
          })
        ]
      },
      {
        guards: {
          someCondTrue: (ctx) => ctx.someCondition,
          someCondFalse: (ctx) => !ctx.someCondition
        }
      }
    );

    interpret(machine)
      .onTransition(() => {
        if (oneExecuted && !twoExecuted) {
          done();
        }
      })
      .start();
  });
});
```

Silverwolf90:
Merged in `main` and fixed issues. Looks like everything is building correctly now.

davidkpiano:
cc. @Andarist

Silverwolf90:
@Andarist -- here's the ping you asked for in the discord's #contributing channel :)

davidkpiano:
@Silverwolf90 Can you merge `main` into this?

Andarist:
My main concerns here are:
- I somewhat like explicit `{ actions: [...] }`, if u assume that the second argument is always actions (which is reasonable) then this makes sense API-wise that u dont have to wrap this in an object at the call site. I feel though that in here the readability suffers a little bit, just to make things a little bit more terse. Is it really worth it? And if we choose to wrap the second argument in the object then we just spare ourselves writing the `cond` property manually which is not much.
- we'd have to add support for this in many places outside of the core. Mainly in our machine extractor used by the vscode extension etc and then we'd have to potentially remember *how* this was declared just so we dont sync it back to the source code as regular `choose` (this might not be relevant to the current implementation as IIRC it doesnt care about the shape of the inline stuff but if we ever allow for those being editable in the editor  then we would face this problem)

IMHO the added terseness is not quite worth the added complexity. It's also something that needs to be in the docs, will pop up in learning materials etc and all of that raises the cognitive overload for newcomers

davidkpiano:
> * we'd have to add support for this in many places outside of the core. Mainly in our machine extractor used by the vscode extension etc and then we'd have to potentially remember _how_ this was declared just so we dont sync it back to the source code as regular `choose` (this might not be relevant to the current implementation as IIRC it doesnt care about the shape of the inline stuff but if we ever allow for those being editable in the editor  then we would face this problem)

@Andarist This shouldn't be an issue (as in it might currently be an issue, but it shouldn't in the future) - we can't forbid the use of helpers (3rd-party or our own) just because it might be incompatible with the machine extractor.

Silverwolf90:
Thanks for your replies!

Over time, using `choose` for single-branch conditions starts to feel _extremely_  cumbersome.  In the original discussion I posted:

> xstate suffers enough from being "syntax heavy" and `choose` is very noisy for simple cases. If I had to choose 1 of these to make it into the core library, it would be `when` which we use extensively -- we have ~60 occurrences across our machines.

I just checked and we're actually at ~82 occurences of `when` now (as opposed to 5 for `choose`). Obviously our codebase could be an outlier, but for us single branch conditionals are by far the most common.

> This shouldn't be an issue (as in it might currently be an issue, but it shouldn't in the future) - we can't forbid the use of helpers (3rd-party or our own) just because it might be incompatible with the machine extractor.

Definitely agree that it _shouldn't_ be an issue in theory -- but it does feel very optimistic! I imagine there are lots of really useful patterns that can't be easily supported by tooling except by releasing "official" functionality is designed with tooling integration in mind. That's the primary reason I opened up this PR -- my `when` actions aren't recognized in the various visualizers and I wanted official support for a pattern I consider to be extremely useful. 😄

davidkpiano:
@Andarist I'd like to revisit this, as I've ran into the need for this multiple times.

davidkpiano:
May be superseded by #4429

davidkpiano:
Thanks for the PR! We'll hold off on this as [enqueue actions](https://stately.ai/docs/actions#enqueue-actions) is the recommended pattern for this now:

```ts
actions: enqueueActions(({ enqueue, check }) => {
  if (check('someCond')) {
    enqueue(someAction)
  }
})
```

PR #2672, opened 2021-09-19
Add description property to state and transition

davidkpiano:
This PR adds `description` as a top-level property for state nodes and transitions:

```js
  // ...
  states: {
    active: {
      // ...
      description: 'The task is in progress',
      on: {
        DEACTIVATE: {
          // ...
          description: 'Deactivates the task'
        }
      }
    }
  }
});
```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -222,6 +222,7 @@ export interface TransitionConfig<TContext, TEvent extends EventObject> {
   internal?: boolean;
   target?: TransitionTarget<TContext, TEvent>;
   meta?: Record<string, any>;
+  description?: string;
 }
 
 export interface TargetTransitionConfig<TContext, TEvent extends EventObject>
@@ -647,6 +648,10 @@ export interface StateNodeConfig<
    * @default false
    */
   preserveActionOrder?: boolean;
+  /**
+   * A description of the state node, rendered in Markdown
+   */
+  description?: string;
 }
 
 export interface StateNodeDefinition<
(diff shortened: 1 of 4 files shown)


mattpocock:
Is this an alternative to `meta.description?`. I like it.

davidkpiano:
> Is this an alternative to `meta.description?`. I like it.

You got it!

mattpocock:
Agree! I will add this to the parser.

Andarist:
What the `meta` property is supposed to hold now? Is it just a bag of whatever one wants to put on a state/transition (basically `Record<string, any>`)? What has prompted moving this from the `meta` to the "core"?

mattpocock:
@Andarist Having autocomplete on description should be nice! And having it as a key on the event should also be handy

packages/core/src/types.ts
```diff
@@ -647,6 +648,10 @@ export interface StateNodeConfig<
    * @default false
    */
   preserveActionOrder?: boolean;
+  /**
+   * A description of the state node, rendered in Markdown
```
Andarist, line 652: We probably should mention here that "rendering in Markdown" applies to the XState visual tooling. It would also be nice to specify what flavor of Markdown this is interpreted as (I assume that it's GFM but could be Commonmark or something else)

davidkpiano:
> What the `meta` property is supposed to hold now? Is it just a bag of whatever one wants to put on a state/transition (basically `Record<string, any>`)? What has prompted moving this from the `meta` to the "core"?

Devs use `meta` for many different things, some of which may even conflict with `description` etc. We ultimately don't want to mix things that XState will natively support in tooling (like `description`) with "free-form" `meta` objects.

Motivation is also here: https://github.com/statelyai/xstate/pull/2672#issuecomment-922543276

packages/core/src/types.ts
davidkpiano: I removed the markdown part. Will specify in documentation for tooling, not here

Andarist (approved):
This probably deserves a small changeset - other than that: LGTM

PR #2674, opened 2021-09-20
Andarist/typegen types

Andarist:


.changeset/red-ants-lick.md
```diff
@@ -0,0 +1,39 @@
...
+```ts
+const model = createModel({}, {
+  events: {
+    FOO: () => ({}),
+    BAR: () => ({}),
+  }
+});
+
+interface Meta {
+  __generated: 1;
```
mattpocock, line 19:
```suggestion
  '@@xstate/typegen': true;
```

mattpocock:
@Andarist Question:

https://www.loom.com/share/e9737e8c1e3749c2afff7104a8355b79

packages/xstate-test/src/types.ts
```diff
@@ -1,5 +1,5 @@
 import { EventObject, State, StateNode } from 'xstate';
-export interface TestMeta<T, TContext> {
+export interface TesTTypesMeta<T, TContext> {
```
mattpocock, line 2: Is this a typo?

mattpocock (approved):
Approved, with a couple of comments. Working locally with VSCode extension.

mattpocock:
@davidkpiano Could we get a review on this one?

davidkpiano:
> @davidkpiano Could we get a review on this one?

Yes, will review later today 👍

packages/core/src/Machine.ts
```diff
@@ -50,23 +58,64 @@ export function Machine<
...
+  TEvent,
+  TTypestate,
+  BaseActionObject,
+  TTypesMeta
+> {
   return new StateNode<TContext, any, TEvent, TTypestate>(
     config,
-    options
-  ) as StateMachine<TContext, any, TEvent, TTypestate>;
+    options as any
```
davidkpiano, line 112: What needs to be fixed here?

packages/core/src/State.ts
```diff
@@ -21,6 +21,8 @@ import { StateNode } from './StateNode';
 import { getMeta, nextEvents } from './stateUtils';
 import { initEvent } from './actions';
 import { IS_PRODUCTION } from './environment';
+import { TypegenDisabled, TypegenEnabled } from './typegenTypes';
+import { BaseActionObject } from '.';
```
davidkpiano, line 25:
```suggestion
import { BaseActionObject } from './types';
```

packages/core/src/Machine.ts
mattpocock: What do you mean?

mattpocock: The options type is basically too dynamic now - it can either be specified loosely or tightly, depending on whether we have generated types. Typescript isn't smart enough, so we patch it with any.

packages/core/src/State.ts
```diff
@@ -298,7 +323,10 @@ export class State<
    * @param parentStateValue
    */
   public matches<TSV extends TTypestate['value']>(
-    parentStateValue: TSV
+    parentStateValue: TResolvedTypesMeta extends TypegenEnabled
+      ? // @ts-ignore
+        TResolvedTypesMeta['matchesStates']
```
davidkpiano, line 328: What's the problem here? (And in the other one)

mattpocock: I don't know specifically with this one - might need a comment here.

.changeset/red-ants-lick.md
```diff
@@ -0,0 +1,39 @@
+---
+"xstate": minor
+"@xstate/react": minor
+---
+
+Added the ability to tighten TS declarations of machine with generated metadata. This opens several exciting doors to being able to use typegen seamlessly with XState to provide an amazing typing experience.
+
+The basics work like this:
```
Andarist, line 8: This is IMHO too technical and also describes a lot of implementation details that people shouldn't focus on, nor they should try to declare `Meta` type on their own.

mattpocock: How would you word it?

packages/core/src/Machine.ts
Andarist: Actually, after some changes, this cast is no longer needed. 🤷‍♂️ I've removed it locally

packages/core/src/State.ts
Andarist: I will write more about the design of the whole thing (would love to describe this in a blog post as well) but to answer this question immediately:

The idea here is that I don't check if the `TResolvedTypesMeta extends TypegenMeta` because I trust the injected type (it is only supposed to be injected using our tooling). Checking all of the properties would impose some extra performance overhead - after all, you need to check if every property in the constraint type is satisfied by the provided/instantiated type. But since this never provided by the user we can only check a single property to see if we actually got a type with `'@@xstate/typegen': true` and continue assuming that the rest of this type is "OK".

I've used `// @ts-ignore` here since it felt cumbersome to use things like `Prop` and `Cast` all over the place to satisfy the compiler. It also felt that this might come with some perf boost since we wouldn't have to go through additional layers of types, check more constraints etc. I've tested this on a simple example, and `// @ts-ignore` performed better when it comes to the number of type instantiations than `Prop`. I've consulted this with Devan and he said that my example could be too simple and that `Prop` should perform better. It feels that this might be true but it also might depend on how sophisticated the actual implementation of TS is.

The `// @ts-ignore` approach has some drawbacks associated with it though. The problem is that TS can remove/misplace those comments when generating `.d.ts` and that in turn can raise errors in users' projects if they are not using `skipLibCheck: true`. They really should use that option but that's not our decision to make and we shouldn't enforce stuff like this on them - especially given the fact that the issue might be hard to diagnose on their end.

So I've decided that I will be changing all of those to `Prop`/`Cast` appropriately. The plan is to revisit the `// @ts-ignore` approach in the future though. I would love to test both versions on a real codebase - like XState Visualizer or smth - and make a more informed decision about it then. If the `// @ts-ignore` approach would come out of this experiment victorious then we could write a script that would include this comment at correct positions in the `.d.ts`.

packages/core/src/StateNode.ts
```diff
@@ -465,16 +471,22 @@ class StateNode<
    * @param context Custom context (will override predefined context)
    */
   public withConfig(
-    options: Partial<MachineOptions<TContext, TEvent>>,
+    options: MaybeTypegenMachineOptions<
+      TContext,
+      TEvent,
+      BaseActionObject,
+      TResolvedTypesMeta,
+      true
```
Andarist, line 479: At the moment `withConfig` will require all missing implementation to be provided - just like `interpret`, `useMachine`, and `useInterpret`.

This doesn't quite reflect how one can use it - as people can use `withConfig` to "inject" implementations partially (through multiple `withConfig` calls). Supporting this is possible and I've prototypes it [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgzgOQgNwgJwDwBUB8UC8UA2lgLpQQAewEAdgCZzG0rrkD8UAjFAFxQAGANwAoUJCgA1AIYAbAK4Q42PISxEA1hBAB7AGZQyo8dACy6AOYQMpgDRQEqqAHkAtgEtgN+1t0HHUABkDsbg0AAKaDpg2PYA0k5xFNR0jFC++oZQnOpx5PwsqGiiYmGwtHro5sDS2Mk0DEym0gDGABbuLBjStCA4Tlj1qU2tHV2dlWhQpnicpnxQheglnTRoeq3QWGHV0lAA3iJQUNLI0u6y0gBGshAwrmC3rnQ1wO46tHD8h8fHrW8fL5QODANCdCxHKAAX0hHjgcHB90eEGetFe70+30hfxaAMxwNB4MhMJhpQkAFFPvI0BAAIJnC7XW603EY5S4AiGEQiVboDYtMyjToQZxgPHs3ZDRqGHYQGr2ADKEFkBkI+yheB+J1ZgPYWN+xCSnXS2kykWiGCVKvsAHJ-mybThSPwklK0lhdkQAESnc6XG53B5PF7ScVe0je+2A8PZKDmmJWvS2qOfR1EPILJZoYkiUkmONRZDueiB5Go9GAjCQj1yvZUBru2XyyHmosl+i2SGmdzwxFBlEh8WchAfBDyWT+27YT1euEI2gWJHBtGhtnhnAiJzmNBWKvHGvN45a45zvtlwds-UG7Xi-jkygtBQlvfX47xmw9+eL-vl1eA+w+jqnxejgnavscGQGO+rbFhAHZQIBYYbq+OCiAaUJodCm4rGifKbNMQpdAedYpNK2yQLsmqQgA7p4bQAMIfHo7gWHU9bDAR7TCqK4rTrW9i4DgAAU2JQC0TEsfwWCQgAlPwzRcV0MElkuA4rrxxECf0oikry6z4QpYwQIxFQsXxNRUccJhwHqhi7Dp3J6fyZi1gcsKfmey4Vp8tnHjel4IXoOg6F6UAAD4IVc0hoF6mEYbm3Ilo+0XQHo8i0EBYk0qGECGcK5kkQ2TDkbltbCZC4mmRY8mEcZEmscRoGQtE4q2XlLA8WyBX2OqG5yZxRkFah3L5rSNIKjozyqb+4rdr2C4vsRbrFU20hgdMHkLtNF6AsOo7jpO1jEd6p5bT+O3AaQG5OPASBFC+MgKEoGB+emUDGpBG3zd+57qWyzpfV+21-YC6akHF13LdwkKcAA9LD70GOJrhXMK9BQOlGJQJkciyFAp0WO950g7QUDWe9TBZmTbR0BQRQgMAYyE200hMFcEC02AhawfBIDSCAACEokCJC-BcCUSWXDSUBpRleIY3AuW1QVUMlZRInHAAdNrRCuLV9gtf9-BjRAE1TcT3lwHNX5DVDXCiZwuu1TVilHZRBtigF7Uip7laaQcGqkKJ-BO67LuDY1HuteH3G+583UB06Ij9cgOjFiU+b3tI-aSmqkK+oyAbA5bV44regXBaFEVelFMVxe533F3++J+SmQJekFIXhZF0WxcSOklJVIL45yLTZTQ3tCVq1nfFCJxMFnOf8bmMklPDUAAALAHAAC0VCQLie9oFE2byIrk+uKv3JD8A+MAOp0QAYsFnKuJrtGMyZzEWFP+dAUCflO78CEjJAgmooT2F+MSTsUIr4iHXlvXe+8ICH3QCfEQZ8lauyEq4B+jNn46DgTfe+T9gq0gYAAIWiq-PBbQCHvzol-Fiv8y4BT8rXYBoD8DgPWhAlea8EauHkO0KA1EdDUXsMjMAOhFbox6OjTwoidBoA0HAIWmCL60IIeQ+gVC0BwNwaQnQDDP71RYf5QEpcoAcKgCAsBAd1rHAQdvPelAD7ACPugg0ABHai6BQCcPseqXhMC4GZ0oNnZEuwABMnItQF0Ok3curd-4FFYNmY4MITybR+l5ZuADRJtzSUUOKA9r6Alvq4WJhAx4QBypPaeYQAFz1ZlAReUTazRNCQIxYOh8Y5KJr9WsQ44B9MZtARWlU5E7nkOWCmvTb40m8fIdwNJ6AYPPrVHB0Sr5AA) but it comes with a drawback of the auto-completions not working properly at the TS level (in theory, we could bring them back through VScode extension). 

Given this drawback and increased complexity, I've ditched implementing this in XState for the time being. Let's see if people even need it in the first place.

mattpocock: The way I implemented it on my initial branch was to make it a deep partial of the options. I think that's a good compromise.

packages/core/src/StateNode.ts
```diff
@@ -147,7 +149,8 @@ class StateNode<
   TContext = any,
   TStateSchema extends StateSchema = any,
   TEvent extends EventObject = EventObject,
-  TTypestate extends Typestate<TContext> = { value: any; context: TContext }
+  TTypestate extends Typestate<TContext> = { value: any; context: TContext },
+  TResolvedTypesMeta = TypegenDisabled
```
Andarist, line 153: Adjusting those classes to support more generics is a real chore. I wonder if something can be done about this in the future. It seems that exposing a public overload and an internal implementation type is possible but I don't quite understand the assignability rules between those two. It would be worth experimenting with this in the future.

packages/core/src/model.types.ts
```diff
@@ -37,10 +34,24 @@ export interface Model<
...
-    implementations?: Partial<MachineOptions<TContext, TEvent, TAction>>
-  ) => StateMachine<TContext, any, TEvent>;
+  createMachine: {
+    <TTypesMeta extends TypegenConstraint = TypegenDisabled>(
+      config: MachineConfig<TContext, any, TEvent, TAction, TTypesMeta>,
+      implementations?: MaybeTypegenMachineOptions<
+        TContext,
+        TEvent,
+        TAction,
+        ResolveTypegenMeta<TTypesMeta, TEvent, TAction>
```
Andarist, line 44: calling `ResolveTypegenMeta` once here was impossible because of the TS issue with the inference that I've been debugging, inlining it into both call sites fixed the issue

packages/core/src/typegenTypes.ts
```diff
@@ -0,0 +1,260 @@
...
+  missingImplementations: {
+    actions: string;
+    delays: string;
+    guards: string;
+    services: string;
+  };
+  eventsCausingActions: Record<string, string>;
+  eventsCausingDelays: Record<string, string>;
+  eventsCausingGuards: Record<string, string>;
+  eventsCausingServices: Record<string, string>;
```
Andarist, line 34: I would probably prefer combining those under a single property - but didn't come up with good names for this. WDYT?

mattpocock: `eventsCausing: { actions: Record<string, string> }`

Is there a benefit to this beyond looking prettier?

packages/core/src/typegenTypes.ts
```diff
@@ -0,0 +1,260 @@
...
+  TRequireMissingImplementations,
+  // @ts-ignore
+  TMissingActions = TResolvedTypesMeta['missingImplementations']['actions']
+> = MaybeMakeMissingImplementationsRequired<
+  'actions',
+  TMissingActions,
+  TRequireMissingImplementations
+> & {
+  actions?: TypegenMachineOptionsActions<TContext, TResolvedTypesMeta>;
+};
```
Andarist, line 179: This looks a little bit complicated - the idea is that we make all named implementations optional and only "mark" required ones as required using `unknown` value.

This might look like a trick but it makes perfect sense if you think about what a type intersection is - a resulting type has to satisfy **all** member types. So by making the property required, even with `unknown` used as a value, we make it required because an intersection of "it has to be here and it might be here" is "it has to be here". Check a simple demo here:
[TS playground](https://www.typescriptlang.org/play?#code/FAFwngDgpgBAggYxASwPYDsDOMC8MDewMxMAZqqgPwBcMAFFAG634zjS0DkAYgPK+cYAXwCUuAHwxGqZABNgQ4KEiwASlACOAV2QAnKLMQoM2PIRJkKtLegDW6VAHd0Cpe1jdk6AIYAbI2hYuPBIgdgAZDDq2noGASZKslAIvt76ZDahGGTodN5ZWLSePv4FmCK00nJKpLn4osC1dKzkqLR0YjiS9cIiQA)

This is somewhat a workaround for not being able to conditionally change the required/optional modifier in a single mapped type.

This also has a somewhat unfortunate effect that `undefined` stays assignable to the property:
[TS playground](https://www.typescriptlang.org/play?#code/FAFwngDgpgBAggYxASwPYDsDOMC8MDewMxMAZqqgPwBcMAFFAG634zjS0DkAYgPK+cYAXwCUuAHwxGqZABNgQ4KEiwASlACOAV2QAnKLMQoM2PIRJkKtLegDW6VAHd0Cpe1jdk6AIYAbI2hYuPBIgdgAZDDq2noGASZKslAIvt76ZDahGGTodN5ZWLSePv4FmCK00nJKpLn4osC1dKzkqLR0YjiS9cIijXWWbTA2SaReBr1AA)

However, I feel like this is not something that can easily trip people - we don't support passing `undefined` at those positions anywhere and it's not an easy mistake to make. I plan to add a runtime dev-only validation for this which I think is a good enough solution. In addition to that, we could validate this in the VScode extension so I don't think it's a big deal. By not handling this at the type level we avoid creating two mapped types with conditionals etc that would intersect with each other here - so I think that it's worth it.

The issue also actually goes away with `exactOptionalPropertyTypes: true`:
[TS playground](https://www.typescriptlang.org/play?exactOptionalPropertyTypes=true#code/FAFwngDgpgBAggYxASwPYDsDOMC8MDewMxMAZqqgPwBcMAFFAG634zjS0DkAYgPK+cYAXwCUuAHwxGqZABNgQ4KEiwASlACOAV2QAnKLMQoM2PIRJkKtLegDW6VAHd0Cpe1jdk6AIYAbI2hYuPBIgdgAZDDq2noGASZKslAIvt76ZDahGGTodN5ZWLSePv4FmCK00nJKpLn4osC1dKzkqLR0YjiS9cIijXWWbTA2SaReBr1AA)

But we can't expect everyone to turn this on - and this option has not been included in the `strict` settings by the TS team.

packages/core/src/typegenTypes.ts
```diff
@@ -0,0 +1,260 @@
...
+}
+
+export interface ResolvedTypegenMeta extends TypegenMeta {
+  indexedActions: Record<string, BaseActionObject>;
+  indexedEvents: Record<string, EventObject>;
+}
+
+export type TypegenConstraint = TypegenEnabled | TypegenDisabled;
+
+// if combined union of all missing implementation types is never then everything has been provided
```
Andarist, line 44: This is a neat trick that allows us to compare again **fewer** strings than more (since the check is inversed). In a way we don't have to match against any strings - we just create a union and check if the result is `never`.

packages/core/src/types.ts
```diff
@@ -56,10 +75,7 @@ export type Event<TEvent extends EventObject> = TEvent['type'] | TEvent;
 export interface ActionMeta<
   TContext,
   TEvent extends EventObject,
-  TAction extends ActionObject<TContext, TEvent> = ActionObject<
```
Andarist, line 59: Changes like this across the codebase are actually bug fixes~ since the constraint and the default was incorrect previously.

packages/core/src/types.ts
```diff
@@ -821,27 +830,58 @@ export interface HistoryStateNode<TContext> extends StateNode<TContext> {
   target: StateValue | undefined;
 }
 
+// @ts-ignore TS complains about withConfig & withContext not being compatible here when extending StateNode
```
Andarist, line 833: this should hopefully get fixed on the next branch since `StateMachine` shouldn't overload `StateNode` like this there

packages/xstate-test/src/types.ts
Andarist: Ye, somehow this got south despite me using `Rename Symbol` functionality

packages/xstate-react/src/useMachine.ts
```diff
@@ -1,14 +1,18 @@
...
   Interpreter,
   InterpreterOptions,
-  MachineOptions,
+  MaybeTypegenMachineOptions,
+  State,
   StateConfig,
+  StateMachine,
   Typestate,
-  ActionFunction
+  AreAllImplementationsAssumedToBeProvided,
```
Andarist, line 12: This introduces a type-level dependency on the newest version of XState. In theory, we declare that `@xstate.react` works with `xstate@>=4.11.0`:
https://github.com/statelyai/xstate/blob/385d29847a1462ee7cd564fdec0a8d66cbf34406/packages/xstate-react/package.json#L50
and this stops being true because of changes like this.

Issues like this have been reported to us in the past (like [here](https://github.com/statelyai/xstate/issues/2410)) - so it's not like a theoretical problem. People don't always upgrade `@xstate/react` together with `xstate`.

Should we care about issues like this? My general stance is that I try not to inflict such issues on users - it is quite problematic in practice though.

Note that this could be avoided by introducing a package like `@xstate/internal-types` that would be a regular dependency of both `xstate` and `@xstate/react`.

cc @davidkpiano

.github/workflows/nodejs.yml
```diff
@@ -27,4 +27,6 @@ jobs:
         run: yarn build
 
       - name: Test
-        run: yarn test --silent
+        run: yarn test:ci --silent
+        env:
+          NODE_OPTIONS: '--max_old_space_size=4096'
```
Andarist, line 32: The memory issue is quite weird. I've been testing this locally and I couldn't notice any substantial increase in the memory usage when using Jest. It has also used "only" 600MB of the total memory heap while the limit on CI is 2GB so it shouldn't cause any issues. Apparently, it did though and increasing the limit to 4GB helps. 🤷‍♂️

packages/xstate-react/src/useMachine.ts
davidkpiano: One of my concerns is that the typegen types are bleeding into other XState packages (like `@xstate/react`, and will soon become necessary for others); is there any way to prevent this and have them only apply to `xstate` core? Even at the small expense of not providing as helpful typing for the XState helpers

mattpocock: @davidkpiano What's the downside of the typegen types bleeding into other packages?

Andarist:
@mattpocock the `unknown` issue is a result of using `// @ts-ignore`. At the moment there is a special "init" event that is used internally - so it should actually be added somehow to here. It should:
- be included in the generated meta type
- be included in the `indexedEvents` coming from the `ResolveTypegenMeta`

Does this sound OK to you?

packages/xstate-react/src/useMachine.ts
Andarist: It's IMHO crucial for this to be supported in the integration packages - the DX would be quite severely impacted by not including this in those packages.

mattpocock:
@Andarist I think having `unknown` there (in places where an event cannot be inferred) is reasonable. I can't see a use case for typing the exact event, especially when it has no payload.

packages/core/src/typegenTypes.ts
Andarist: > Is there a benefit to this beyond looking prettier?

Just this 😅 So ye - probably not worth it. When I think about it now - not nesting them might actually make it easier to index those properties (since I always~ pass the "root" meta-type around).

packages/core/src/StateNode.ts
Andarist: But then the user can easily miss the need to provide all implementations, don't they? 🤔

packages/core/src/typegenTypes.ts
mattpocock: Yep I was thinking that too

packages/core/src/StateNode.ts
mattpocock: Yes, but it's not recommended in most cases to use `.withConfig`, unless for unit testing.

Andarist:
I'm unsure how to accurately replicate this situation on the TS playground cause I got `any` instead of `unknown` on my first attempt. However, it feels important because we **must** include the init event on the list of `eventsCausingX`. Otherwise, we would end up with a broken control flow analysis. Imagine this ([TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgdgEwgDwgqBeKBvAUFAqAMQHkSAuHKUSSgclJLoBooA3AQwBsBXCSuDwC2AIwgAnKAF98hAEIBBAEqVs1cPyh1FSlu259KAZ2DiAlnADm03DNw1oAWQjAOmHLIIQ2EOMCMAwhw8RhaWCgDGwGYA9nBGqp6EBAD0KVAiPMBQFlDiENxmoFARHHBQ3EYxeRAREGY+UAAGyCYcwBAAdBZFTVBJyR0m9Ix0UAA+Wjp0STJ2DlAkYNFxRu54yRxRsfGJyYMQw1AAFN6U8EioCADazq7XAETevv5BIWGRK-EPALqPQ8Bfj8AJSYAB87BiZgQs1suFwSAiXA4+SgADMeHBtnEKhE6ssAMqmMLHEziYzEqzAyhsKEwhG1ZGojFYr4lfLtCCOLYACwsEGOMWWOwSi2Fq2pkOh8IiHI63IifLgAo2hC2X1Fqv21EOwEopzYoIwEK12sIZjRJ28nQWGDtWlGoNNZv2aWajCaAzNUigEC4RmgzpdhDdTR0TVYmWyRh5MR4XHQYeUExycF6XuDUDdHDWbnyRnj2WAfLWYgiMSEhwq+i40JK3C4Jws5fE+SiXBAwIzwa2+OARPMVgNnU4vAgXczBBkLunhDmwKAA)):
```ts
type Indexed = {
    FOO: { type: 'FOO', value: number }
    BAR: { type: 'BAR', value: string }
}

type Meta = {
    eventsCausingActions: {
        // but in reality can also receive `xstate.init` 
        test: 'FOO' | 'BAR'
    }
}

type Options = {
    actions: {
        test: (ev: Indexed[Meta["eventsCausingActions"]["test"]]) => void
    }
}

declare function acceptString(str: string): void

declare function createMachine(options: Options): void

createMachine({
    actions: {
        test: (ev) => {
            if (ev.type === 'FOO') {
                // `FOO`
            } else {
                // `BAR`, but should `BAR | init`
                // as a result this becomes a valid call (incorrectly)
                acceptString(ev.value)
            }
        }
    }
})
```

And if we need this for cases like this then it should be also included for implementation types that can only receive that special `xstate.init` event.

mattpocock:
@Andarist I really don't see a need for this! Since the user never provides a generic to cover `xstate.init`, it seems redundant to check for it. I'm not even sure how we'd check for it on the typegen side, because it isn't covered in state.transitions.

I think we should check usage and see if anyone asks for this - but it's never been requested from usage in `xstate-codegen` and I can't see it being useful here.

mattpocock:
I think my reluctance is less about its usefulness, and more that it feels like a feature request on the typegen side that I'm not sure how to satisfy. Since this is an edge case that can be fixed by a bug, I think we should ship and fix the bug later if required.

Andarist:
I feel pretty strongly that we should support this. I don't think that shipping bugs knowingly is ever a good option for a library code (app-level code is different). At the app level, we take the responsibility for our bugs and for how they affect our business. I don't feel comfortable at all about us taking the responsibility for bugs affecting other people's businesses. Note that I think the same about crucial parts of services that we provide - even if from our perspective they might feel more like an app code than like a library code. Basically, bugs that are somewhat local and don't affect the core of our products are acceptable to me but a library/service bug is, by definition, not local and I feel like it's crucial to minimize the negative impact that we might have on others - if we can only predict if beforehand.

I've implemented support for this in the types [here](https://github.com/statelyai/xstate/pull/2674/commits/8805f6a7deb71857805fb382d5a582443bb66d15). Note how this doesn't require the user to provide this event at all - it can be completely provided on our side.

As to how to compute this at the typegen side - this gets a little bit tricky. We almost could use `machine.initialState.actions` **but** this will include all the actions from the initial macrostep and won't limit this to a microstep. We could either port [`microstep` from v5](https://github.com/statelyai/xstate/blob/20671291f89e0d26a6f9bb7390b946b0fad76bd3/packages/core/src/StateMachine.ts#L215) to v4 or reimplement the logic for resolving this at the typegen side. I think that I would actually prefer the latter as adjusting the code in v4 and fixing conflicts later with the `next` branch might just be not worth it. Since we already have methods like `getStateNodeById`, `getStateNodeByPath`, etc it actually shouldn't be that complex to do this.

mattpocock:
@Andarist I think it's probably fine to compute it. We would:

1. Look the initial state.
2. Capture all entry actions/invokes/delays on that state
3. If it has children, look at their initial states and repeat from 2.

Though this also raises a question about 'always', 'delays' and others. Should we also be computing those? I am worried about this pandora's box you've opened, when I think it'll have little actual impact on users.

Andarist:
> Though this also raises a question about 'always', 'delays' and others. Should we also be computing those? 

If they can be called with that special event (and they do) then it should be handled as well.

> I am worried about this pandora's box you've opened, when I think it'll have little actual impact on users.

The pandora's box was already opened and the worms were already crawling out of it 🤣 Since this was never handled anywhere in our types etc. This puts us in a unique position now for fixing those mistakes. 

It's hard to say how much of an impact this will have on users. However, even if we spare a single user a tragedy that can happen without supporting this then it's still worth it in my eyes. For me, it's about solid architecture, a clean mental model, and trust. If I would be put on the user's side and discover this randomly - possibly after debugging a production issue - I would probably get pissed because of this issue. And I'm a nice guy, I don't want to piss people off 😉

mattpocock:
@Andarist Some of these events have dynamic `type` attributes, especially those generated by delays. How can we generate those without TS 4.1?

mattpocock:
Part of what I don't like about this is that we're exposing more XState internals to the user through the types. It's crystal-clear to us what `xstate.init` means, but users? It's more reading/code-diving in the source. If we're trying to protect users from their own errors, I think this is the wrong way to go about it.

Andarist:
> @Andarist Some of these events have dynamic type attributes, especially those generated by delays. How can we generate those without TS 4.1?

The dynamic type is an interesting challenge - we should list all of those possible event patterns and evaluate them one by one. I don't think that we have to support TS<4.1 with the typegen though - it's a new feature and can require a newer version of TS than the rest of the codebase. This whole point of `typesVersions` shenanigans is to enable exactly that. 

> Part of what I don't like about this is that we're exposing more XState internals to the user through the types. It's crystal-clear to us what xstate.init means, but users? It's more reading/code-diving in the source. If we're trying to protect users from their own errors, I think this is the wrong way to go about it.

The type of this event is quite self-explanatory - I can see though how one can be slightly surprised by its existence. However, this feels like a public API to me - only a one that is not well-document at the moment. If we are calling user-provided functions with some values then those values are immediately becoming a part of the API. I'm not super fond of the `xstate.init` event myself but I don't see any better alternatives. It can be a special event, or null, or undefined, or some other special kind of value. Keeping the shape (`{ type: string }`) consistent with "regular"/explicit events makes it easier to work with though.

mattpocock:
Let's have a call about this tomorrow properly with @davidkpiano - I think we should prioritise shipping without this because of:

1. Its complexity - it'll add a significant amount of time to this work.
2. It's an edge case we can improve later

We'll only know if we need to add this feature _once_ we've shipped it and started receiving feedback.

.changeset/red-ants-lick.md
Andarist: I would only mention how you can use this, that it's a beta feature, give a short example - but I would omit the `interface Meta` declaration from here

.changeset/red-ants-lick.md
```diff
@@ -0,0 +1,18 @@
+---
+'xstate': minor
+---
+
+author: @Andarist
+author: @mattpocock
+
+Added the ability to tighten TS declarations of machine with generated metadata. This opens several exciting doors to being able to use typegen seamlessly with XState to provide an amazing typing experience.
+
+With the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode), you can specify a new attribute called `tsTypes: true` in your machine definition:
```
Andarist, line 10: the URL here will require to be updated, right? I've seen that you were setting up Stately publisher with David

mattpocock: Yep, we can adjust later

Andarist: Sure, just remember about updating that before releasing the new version of XState - people subscribe to our releases and it would be preferable if the final URL would end up in the emails sent out to them.

packages/core/src/State.ts
```diff
@@ -298,7 +323,9 @@ export class State<
    * @param parentStateValue
    */
   public matches<TSV extends TTypestate['value']>(
-    parentStateValue: TSV
+    parentStateValue: TResolvedTypesMeta extends TypegenEnabled
+      ? Prop<TResolvedTypesMeta, 'matchesStates'>
```
davidkpiano, line 327: Nit: wondering if a utility `PropOrDefault<TResolvedTypesMeta, 'matchesStates', TSV>` would be useful for DRYing some of the types

mattpocock:
@Andarist we should change the ActorRefFrom, StateFrom and other `*From` generics to work with this.

mattpocock (changes requested):
Comments coming

mattpocock:
@Andarist an example from our private repo of some strangeness with inference when it comes to onDone/onError types:

https://github.com/statelyai/registry/pull/97

packages/core/src/typegenTypes.ts
```diff
@@ -0,0 +1,262 @@
...
+  TResolvedTypesMeta,
+  TEventsCausingServices = Prop<TResolvedTypesMeta, 'eventsCausingServices'>,
+  TIndexedEvents = Prop<TResolvedTypesMeta, 'indexedEvents'>,
+  TInvokeSrcNameMap = Prop<TResolvedTypesMeta, 'invokeSrcNameMap'>
+> = {
+  [K in keyof TEventsCausingServices]?: InvokeCreator<
+    TContext,
+    Cast<Prop<TIndexedEvents, TEventsCausingServices[K]>, EventObject>,
+    Prop<Prop<TIndexedEvents, Prop<TInvokeSrcNameMap, K>>, 'data'>
+  >;
```
mattpocock, line 149: There's some strangeness here that's causing this:

https://github.com/statelyai/registry/pull/97

packages/core/src/typegenTypes.ts
```diff
@@ -0,0 +1,262 @@
...
+    TContext,
+    Cast<Prop<TIndexedEvents, TEventsCausingGuards[K]>, EventObject>
+  >;
+};
+
+export type TypegenMachineOptionsServices<
+  TContext,
+  TResolvedTypesMeta,
+  TEventsCausingServices = Prop<TResolvedTypesMeta, 'eventsCausingServices'>,
+  TIndexedEvents = Prop<TResolvedTypesMeta, 'indexedEvents'>,
```
mattpocock, line 142: When a service does not have any `onDone` attribute and it's declared as a promise, it returns `Promise<never>` - which is not correct - it should return `Promise<unknown>`

mattpocock: Actually, I might be able to fix this on the typegen side - I don't think we're providing all the `internalEvents` that we should be.

packages/core/src/typegenTypes.ts
mattpocock: Could be the same stuff as below, ignore until I've had a chance to look.

packages/core/src/typegenTypes.ts
Andarist: ah, this makes sense - if there is no transition defined for a particular `invoke` then you don't have that on the list of `.transitions`

mattpocock: Fixed this on the typegen side, but still erroring in the same way, discussed on Discord

mattpocock (approved):
It's an approve from me. QA'd for the whole day and PR'd suggested changes.

.changeset/red-ants-suck.md
```diff
@@ -0,0 +1,8 @@
+---
+'@xstate/react': major
+---
+
+author: @Andarist
+author: @mattpocock
+
+To avoid breaking any consumers and to leverage the newly introduced typegen support the major version of this package had to be bumped. While you can still use it with older versions of TS the typegen support in this package required at least TS 4.0.
```
davidkpiano, line 8:
```suggestion
To avoid breaking any consumers and to leverage the newly introduced typegen support, the major version of this package had to be bumped. While you can still use it with older versions of TS, the typegen support in this package requires TS version 4.0 or greater.
```

davidkpiano (commented):
Can we remove the unrelated changes to the docs (and other `.md`) files?

packages/core/src/StateNode.ts
```diff
@@ -102,6 +103,8 @@ import {
 } from './stateUtils';
 import { createInvocableActor } from './Actor';
 import { toInvokeDefinition } from './invokeUtils';
+import { TypegenDisabled } from './typegenTypes';
+import { ServiceMap } from '.';
```
davidkpiano, line 107: Can this be imported from types? (nit)

packages/core/src/StateNode.ts
```diff
@@ -925,18 +964,18 @@ class StateNode<
...
     }
 
-    const nodes: Array<StateNode<TContext, any, TEvent, any>> = [];
+    const nodes: Array<StateNode<TContext, any, TEvent, any, any, any>> = [];
     let marker:
       | StateNode<TContext, any, TEvent, any>
       | undefined = childStateNode;
 
-    while (marker && marker !== this) {
+    while (marker && marker !== (this as any)) {
```
davidkpiano, line 978: Needed still?

Andarist: Nope, it doesn't seem so

packages/core/src/StateNode.ts
Andarist: vscode autoimports are annoying at times :s i agree - we shouldn't ever import from `'.'`. It tends to introduce cycles in the dependency graph (and while they work, it's best to avoid them).

Andarist:
> Can we remove the unrelated changes to the docs (and other .md) files?

Sure, I'll try to do that. It seems that they were introduced as part of [this merge](https://github.com/statelyai/xstate/pull/2674/commits/1bdd690d209241b10565fd27bb307fa150ebbc0a). The files in question were not formatted with Prettier, I got conflicts in other files, resolved them, and then when committing the recommit hooks has reformatted all of the files that were staged.

.changeset/red-ants-lick.md
```diff
@@ -0,0 +1,18 @@
+---
+'xstate': minor
+---
+
+author: @Andarist
+author: @mattpocock
+
+Added the ability to tighten TS declarations of machine with generated metadata. This opens several exciting doors to being able to use typegen seamlessly with XState to provide an amazing typing experience.
+
+With the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode), you can specify a new attribute called `tsTypes: true` in your machine definition:
```
Andarist, line 10:
```suggestion
With the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode), you can specify a new attribute called `tsTypes: {}` in your machine definition:
```

.changeset/red-ants-lick.md
```diff
@@ -0,0 +1,18 @@
...
+
+```ts
+const machine = createMachine({
+  tsTypes: true,
```
Andarist, line 14:
```suggestion
  tsTypes: {},
```

.changeset/red-ants-lick.md
```diff
@@ -0,0 +1,18 @@
...
+})
+```
+
+The extension will automatically add some extra generics to `createMachine`, which allow for type-safe access to nearly all of XState's API's.
```
Andarist, line 18:
```suggestion
The extension will automatically add a type assertion to this property, which allows for type-safe access to a lot of XState's API's.
```

.changeset/red-ants-lick.md
```diff
@@ -0,0 +1,18 @@
...
+
+With the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode), you can specify a new attribute called `tsTypes: {}` in your machine definition:
+
+```ts
+const machine = createMachine({
+  tsTypes: {},
+})
+```
+
+The extension will automatically add a type assertion to this property, which allows for type-safe access to a lot of XState's API's.
```
Andarist, line 18:
```suggestion
The extension will automatically add a type assertion to this property, which allows for type-safe access to a lot of XState's API's.

⚠️ This feature is in beta. Actions/services/guards/delays might currently get incorrectly annotated if they are called "in response" to always transitions or raised events. We are working on fixing this, both in XState and in the typegen.
```

PR #2675, opened 2021-09-20
Added documentation for typegen with the VSCode extension

mattpocock:
https://www.loom.com/share/bea9c84af6564d4d9134358628f7dd2c

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -16,6 +16,8 @@ JavaScript and TypeScript [finite state machines](https://en.wikipedia.org/wiki/
 
 New to state machines and statecharts? [Read our introduction](/guides/introduction-to-state-machines-and-statecharts/).
 
+[Download our VSCode extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode).
+
 📑 Adheres to the [SCXML specification](https://www.w3.org/TR/scxml/)
 
 💬 Chat on the [Stately Discord Community](https://discord.gg/KCtSX7Cdjh)
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -81,6 +81,8 @@ Here's how you can get started:
 2. Open a new file and create a typed model using `createModel`. [Want more info on models?](./models.md). For instance:
 
 ```ts
+import { createModel } from 'xstate/lib/model';
+
 const model = createModel(
   {
     value: ''
@@ -115,11 +117,11 @@ const machine = model.createMachine({
 });
 ```
 
-4. Finally, add a 'types' attribute to the machine and save the file:
+4. Finally, add `tsTypes: true` to the machine and save the file:
 
 ```ts
 const machine = model.createMachine({
-  types: {},
+  tsTypes: true,
   initial: 'a',
   states: {
     a: {
@@ -137,11 +139,11 @@ const machine = model.createMachine({
 });
 ```
 
-5. The extension should automatically add an imported type to the machine:
+5. The extension should automatically add a generic to the machine:
 
 ```ts
-const machine = model.createMachine({
-  types: {} as import('./filename.typegen').Typegen[0],
+const machine = model.createMachine<import('./filename.typegen').Typegen[0]>({
+  tsTypes: true,
   initial: 'a',
   states: {
     a: {
@@ -162,7 +164,7 @@ const machine = model.createMachine({
 6. Add a second parameter into the `createMachine` call - this is where you implement the actions, services, guards and delays for the machine.
 
 ```ts
-const machine = model.createMachine(
+const machine = model.createMachine<import('./filename.typegen').Typegen[0]>(
   {
     // Config here
   },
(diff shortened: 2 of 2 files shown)


docs/guides/typescript.md
```diff
@@ -71,6 +71,201 @@ Providing the context and events as generic parameters for the `createMachine()`
...
+>({
+  types: {} as import('./filename.typegen').Typegen[0]
+});
+```
+
+A future update to our VSCode extension may make this easier, but if it's possible for you to use `createModel`, we recommend you do so.
+
+#### The generated files
+
+We recommend you commit the generated files (`filename.demo.typegen`) to the repository. We currently don't have a way to generate the files en masse on a CI, for instance via a CLI. This is our next job.
```
mattpocock, line 261:
```suggestion
We recommend you commit the generated files (`filename.typegen.ts`) to the repository. We currently don't have a way to generate the files en masse on a CI, for instance via a CLI. This is our next job.
```

davidkpiano (approved):

farskid (approved):

docs/README.md
```diff
@@ -16,6 +16,8 @@ JavaScript and TypeScript [finite state machines](https://en.wikipedia.org/wiki/
 
 New to state machines and statecharts? [Read our introduction](/guides/introduction-to-state-machines-and-statecharts/).
 
+[Download our VSCode extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode).
```
Andarist, line 19: do we have an organization account setup so we could publish this in the Stately namespace?

docs/guides/typescript.md
```diff
@@ -71,6 +71,229 @@ Providing the context and events as generic parameters for the `createMachine()`
 - The event type (`TEvent`) ensures that only specified events (and built-in XState-specific ones) are used in transition configs. The provided event object shapes are also passed on to actions, guards, and services.
 - Events which you send to the machine will be strongly typed, offering you much more confidence in the payload shapes you'll be receiving.
 
+## The VSCode Extension <Badge text="4.26+" />
```
Andarist, line 74: The header here seems to focus on a slightly wrong thing - I would expect a typegen approach to be mentioned here and then the paragraph below could go into mentioning that VScode it's currently the only way to get this working.

docs/guides/typescript.md
```diff
@@ -71,6 +71,229 @@ Providing the context and events as generic parameters for the `createMachine()`
...
+  {
+    events: {
+      FOO: (value: string) => ({ value }),
+      BAR: () => ({})
+    }
+  }
+);
+```
+
+3. Create a machine from the model:
```
Andarist, line 97: a note that we should stick to "named" implementations here would be highlight this as a requirement for this feature

docs/guides/typescript.md
```diff
@@ -71,6 +71,229 @@ Providing the context and events as generic parameters for the `createMachine()`
...
+    }
+  }
+});
+```
+
+5. The extension should automatically add an imported type to the machine:
+
+```ts
+const machine = model.createMachine({
+  types: {} as import('./filename.typegen').Typegen[0],
```
Andarist, line 144: q: should we use a different pseudo-file extension than `.typegen`? something perhaps more focused on `.xstate`? What do other code-generation tools do?

Andarist: I wonder if in the future we won't use this file for other stuff that goes beyond typegen - like it could hold layout information for the editor etc. We probably would like to stick to a single xstate-governed file so we don't generate too many files in people projects

docs/guides/typescript.md
```diff
@@ -71,6 +71,229 @@ Providing the context and events as generic parameters for the `createMachine()`
...
+6. Add a second parameter into the `createMachine` call - this is where you implement the actions, services, guards and delays for the machine.
+
+```ts
+const machine = model.createMachine(
+  {
+    // Config here
+  },
+  {
+    actions: {
+      consoleLogValue: (context, event) => {
```
Andarist, line 171: React-focused recipe would be also much appreciated so people don't get lost on this one

docs/guides/typescript.md
Andarist: At the moment for XState itself there is no special TS version required for this feature but I wonder if it wouldn't be OK to mention that we recommend using at least TS 4.1 or something. The `@xstate/react` will require TS 4.0 for this to work.

tonivj5:
This is awesome @mattpocock!! 👏🏻 Is this typegen going to be launched with v5 or we can start using in in 4.x.x? I've tried adding `types: {}` and `tsTypes: true`, but extension do nothing now, I'm doing something wrong or it isn't released yet?

mattpocock:
@tonivj5 As you can see from this PR, it hasn't been released yet!

mattpocock:
It'll be done as part of v4.

mattpocock:
Closed in favour of #2963

PR #2683, opened 2021-09-23
Added a syncToContext option to useInterpret and useMachine

mattpocock:
This implementation is exceptionally naive, but it appears to work. I'd love advice on how I can improve this.

- [ ] Add docs
- [ ] Add changeset
- [ ] Improve implementation?
- [ ] https://twitter.com/zhzhng/status/1441032589947518979 - Consider whether to handle dependency changes explicitly, or implicitly

### DIFF
--- a/packages/xstate-react/src/useInterpret.ts
+++ b/packages/xstate-react/src/useInterpret.ts
@@ -9,7 +9,8 @@ import {
   InterpreterOptions,
   MachineOptions,
   Typestate,
-  Observer
+  Observer,
+  assign
 } from 'xstate';
 import { MaybeLazy } from './types';
 import useConstant from './useConstant';
@@ -50,7 +51,20 @@ export function useInterpret<
     | ((value: State<TContext, TEvent, any, TTypestate>) => void)
 ): Interpreter<TContext, any, TEvent, TTypestate> {
   const machine = useConstant(() => {
-    return typeof getMachine === 'function' ? getMachine() : getMachine;
+    const machineToReturn =
+      typeof getMachine === 'function' ? getMachine() : getMachine;
+
+    if (!machineToReturn.config.on) {
+      (machineToReturn as any).config.on = {};
+    }
+
+    (machineToReturn.config as any).on['xstate.update.syncToContext'] = {
+      actions: assign((_context, event: any) => {
+        return event.props;
+      })
+    };
+
+    return machineToReturn;
   });
 
   if (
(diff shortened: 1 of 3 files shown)


dougmacklin:
Hi @mattpocock, I just wanted to say that this change looks fantastic and will really simplify some things on our end! The only thing I'm wondering is if shortening it to just `sync` would be clear enough? Just a thought, very happy with it being called `syncToContext` as well.

packages/xstate-react/src/useInterpret.ts
```diff
@@ -50,7 +51,20 @@ export function useInterpret<
...
+
+    if (!machineToReturn.config.on) {
+      (machineToReturn as any).config.on = {};
+    }
+
+    (machineToReturn.config as any).on['xstate.update.syncToContext'] = {
+      actions: assign((_context, event: any) => {
+        return event.props;
+      })
+    };
```
davidkpiano, line 65: Not sure how I feel about an option changing the actual machine config :thinking: this seems like something easy to add but is going to be difficult to remove later, and also might be something unexpected for the user inspecting their machine.

I don't have any immediate better ideas right now, though, but let's keep thinking about this and not merge it in just yet.

mattpocock: I agree, it feels hacky but it appears to work nicely.

@Andarist What do you think?

simonflk:
Would there be any way to make this work with `useActor` also, or is it too late at that point? 
Could that screw things up if you have multiple consumers of that actor?

mattpocock:
@simonflk No, I don't think it could - part of the reason it works here is that it patches the machine definition in order to get it to work. By the point you're consuming the machine, you're too late to modify it.

This could work later on, though, if we got more support in core (a proper `service.updateContext` method or similar).

But it gets hairy if you have multiple actor call sites all trying to sync their context to the same actor. A little nasty IMO.

davidkpiano:
Re: making it work for `useActor` etc., we also have to be careful that we're not allowing unsafe behavior. Editing the `context` from anywhere, despite being convenient, also makes the logic much less robust and more unpredictable. The "slippery slope" here is `context` becoming a magical bag of values that can change at any time (and arguably, it's still magical even if there is a special "change" transition patched onto the original machine definition).

One potential idea for this is scoping this under `input` (imagine a `state.input = { ... }` property). The role of `input` would be precisely for external values coming into the machine; it would not be mixed with `context`, and it can only be changed externally (never modified with `assign(...)`). It would also make machines more pluggable into flow-based programming contexts (visualization++).

mattpocock:
@davidkpiano 100%! It would be much clearer. But, for v5 so it can be given first-class support

davidkpiano:
Superseded by https://github.com/statelyai/xstate/pull/2779

PR #2684, opened 2021-09-23
Update docs to encourage using someRef.getSnapshot() over someRef.state

pollen8:


docs/guides/actors.md
```diff
@@ -371,7 +371,22 @@ To do this, set `{ sync: true }` as an option to `spawn(...)`:
 // ...
 ```
 
-This will automatically subscribe the machine to the spawned child machine's state, which is kept updated in `ref.state`:
+This will automatically subscribe the machine to the spawned child machine's state, which is kept update and can be access via `getSnapshot()`:
```
mattpocock, line 374:
```suggestion
This will automatically subscribe the machine to the spawned child machine's state, which is kept updated and can be accessed via `getSnapshot()`:
```

docs/guides/actors.md
```diff
@@ -371,7 +371,22 @@ To do this, set `{ sync: true }` as an option to `spawn(...)`:
...
+
+  console.log(someRef.getSnapshot());
+  // => State {
+  //   value: ...,
+  //   context: ...
+  // }
+});
+```
+
+Prior to version 4.20.0 this was kept updated in `ref.state`:
```
mattpocock, line 389:
```suggestion
Prior to version 4.20.0 this was kept updated in `ref.state`:
```

I think we can remove this - state is documented elsewhere.

mattpocock (approved):

mattpocock:
Thanks @pollen8!

PR #2685, opened 2021-09-24
`@xstate/react`: Correct machine reference in the `useInterpret` hook

farskid:
This PR:
- Makes sure we target the correct machine reference in `useInterpret`
- Makes sure the warning is shown when machine reference updates

TODO:
- [x] Add changeset

### DIFF
--- a/.changeset/long-icons-pull.md
+++ b/.changeset/long-icons-pull.md
@@ -1,0 +1,37 @@
+---
+'@xstate/react': patch
+---
+
+Correct reference to the machine in `useInterpret` hook.
+Make sure a warning is shown in the console in the development mode when machine reference is updated during the hook lifecycle. This usually happens when machine options are dependent on external values and they're passed via `withConfig`.
+
+```js
+const machine = createMachine({
+  initial: 'foo',
+  context: { id: 1 },
+  states: {
+    foo: {
+      on: {
+        CHECK: {
+          target: 'bar',
+          cond: 'hasOverflown'
+        }
+      }
+    },
+    bar: {}
+  }
+});
+
+const [id, setId] = useState(1);
+const [current, send] = useMachine(
+  machine.withConfig({
+    guards: {
+      hasOverflown: () => id > 1 // id is a reference to an outside value
+    }
+  })
+);
+
+// later when id updates
+setId(2);
+// Now the reference to `machine.withConfig` from above is updated but in the guard, the id value is still stale (id=1).
+```
--- a/.changeset/long-icons-pull.md
+++ b/.changeset/long-icons-pull.md
@@ -3,7 +3,7 @@
 ---
 
 Correct reference to the machine in `useInterpret` hook.
-Make sure a warning is shown in the console in the development mode when machine reference is updated during the hook lifecycle. This usually happens when machine options are dependent on external values and they're passed via `withConfig`.
+Fixed a regression with a development-only warning not being shown when a machine reference is updated during the hook lifecycle. This usually happens when machine options are dependent on external values and they're passed via `withConfig`.
 
 ```js
 const machine = createMachine({
--- a/.changeset/long-icons-pull.md
+++ b/.changeset/long-icons-pull.md
@@ -2,6 +2,9 @@
 '@xstate/react': patch
 ---
 
+author: @farskid
+author: @Andarist 
+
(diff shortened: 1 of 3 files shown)


mattpocock:
Nice catch!

mattpocock (approved):

packages/xstate-react/test/useInterpret.test.tsx
```diff
@@ -111,4 +111,52 @@ describe('useInterpret', () => {
...
+          </button>
+        </>
+      );
+    };
+
+    const { findByRole } = render(<App />);
+
+    (await findByRole('button')).click();
+
+    expect(console.warn).toHaveBeenCalledTimes(1);
```
Andarist, line 160: It would be great to confirm that the reported message is actually ours:
```suggestion
    expect(console.warn).toHaveBeenCalledTimes(1);
    expect(console.warn.mock.calls[0]).toMatchInlineSnapshot();
```

.changeset/long-icons-pull.md
```diff
@@ -0,0 +1,37 @@
+---
+'@xstate/react': patch
+---
+
+Correct reference to the machine in `useInterpret` hook.
```
Andarist, line 5: This is an implementation detail - it doesn't quite matter for our consumers so it's best to leave this out
```suggestion
```

davidkpiano (approved):

.changeset/long-icons-pull.md
```diff
@@ -0,0 +1,37 @@
...
+Make sure a warning is shown in the console in the development mode when machine reference is updated during the hook lifecycle. This usually happens when machine options are dependent on external values and they're passed via `withConfig`.
```
Andarist, line 6:
```suggestion
Fixed a regression with a development-only warning not being shown when a machine reference is updated during the hook lifecycle. This usually happens when machine options are dependent on external values and they're passed via `withConfig`.
```

I've checked that this was introduced correctly https://github.com/statelyai/xstate/pull/918/files but regressed in https://github.com/statelyai/xstate/pull/1472/files

.changeset/long-icons-pull.md
```diff
@@ -0,0 +1,37 @@
...
+  machine.withConfig({
+    guards: {
+      hasOverflown: () => id > 1 // id is a reference to an outside value
+    }
+  })
+);
+
+// later when id updates
+setId(2);
+// Now the reference to `machine.withConfig` from above is updated but in the guard, the id value is still stale (id=1).
```
Andarist, line 36:
```suggestion
// Now the reference passed to `useMachine` (the result of `machine.withConfig`) is updated but the interpreted machine stays the same. So the guard is still the previous one that got passed to the `useMachine` initially, and it closes over the stale `id`.
```

.changeset/long-icons-pull.md
```diff
@@ -0,0 +1,37 @@
...
+```
```
Andarist, line 37: It would be great if we could also mention how one can fix this here, add a note to the docs about this and maybe even add a link to that note in the warning itself

.changeset/long-icons-pull.md
farskid: How do we word it so that the consumers are aware of a bug fix?

.changeset/long-icons-pull.md
farskid: Good idea. I'm updating the docs in a separate PR. in that same PR, I'll also refer to it here. sounds good?

.changeset/long-icons-pull.md
```diff
@@ -0,0 +1,37 @@
+---
+'@xstate/react': patch
+---
+
```
Andarist, line 4:
```suggestion

author: @farskid
author: @Andarist 

```

.changeset/long-icons-pull.md
Andarist: The other suggestion I've made mentions that it's a fix for a regression

.changeset/long-icons-pull.md
Andarist: Slightly unsure about the exact shape of that other PR - but if you prefer to make those changes in a separate PR, then it's fine for me.

farskid: it'll be a doc-only change. Let's take it there.

Andarist (approved):

PR #2686, opened 2021-09-24
`@xstate/react`: Update docs with instructions for using external values in machine configuration

farskid:
In this PR, additional instructions are added to keep machines up to date in cases in which machine options and context are dependent on an external mutable value.

TODO:
- [ ] Refer to these instructions in the tests of `useInterpret`
- [ ] Add changeset

### DIFF
--- a/packages/xstate-react/README.md
+++ b/packages/xstate-react/README.md
@@ -453,6 +453,74 @@ const Fetcher = ({ onResolve }) => {
 };
 ```
 
+### A note on external values
+
+When the machine options or context need to read from an external value, it's important that they can read the latest value in case it's changed! Therefore, using `withConfig` and `withContext` on the passed machine might end up being a recipe for bug. Instead, you can use one of the two approaches below:
+
+#### Pass a callback to `useInterpret` and `useMachine`
+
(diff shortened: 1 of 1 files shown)


packages/xstate-react/README.md
```diff
@@ -453,6 +453,74 @@ const Fetcher = ({ onResolve }) => {
...
+  const [id, setId] = useState(1);
+  // Passing a callback to `useMachine` makes sure the value of `id` is always up to date with the useState above
+  const [state, send] = useMachine(() =>
+    machine.withContext({ id: 1 }).withConfig({
+      guards: {
+        hasOverflown: () => id > 1
+      }
+    })
+  );
+};
```
mattpocock, line 489: I don't think this is true - and I don't think we should recommend this either! The second option is both much cleaner, and soon will be perfectly typesafe.

Andarist: Yep - this isn't true, the `id` would still be closed over here. The lazy initialization in here only makes it slightly more apparent that this can be stale. If you want auto-updates you should use the second argument of the `useMachine`

PR #2689, opened 2021-09-26
Doc: Add doc to useInterpret() in index.md

redbar0n:
To clarify that `useInterpret` works with component lifecycles (like `useMachine` does), and how it compares to `useMachine`.
Had to dig into #1915 and aea4a2f to figure out the difference between `useInterpret` and `useMachine`. Realised the only difference was the observer, since `useInterpret` was previously named more clearly as `useMachineObserver`.

### DIFF
--- a/packages/xstate-react/README.md
+++ b/packages/xstate-react/README.md
@@ -141,7 +141,7 @@ const [state, send] = useActor(customActor, (actor) => {
 
 ### `useInterpret(machine, options?, observer?)`
 
-A React hook that returns the `service` created from the `machine` with the `options`, if specified. It also sets up a subscription to the `service` with the `observer`, if provided.
+A React hook that returns the `service` created from the `machine` with the `options`, if specified. It also sets up a subscription to the `service` with the `observer`, if provided. It starts the service and runs it for the lifetime of the component. Similar to using `useMachine`, but with an observer.
 
 _Since 1.3.0_
 


packages/xstate-react/README.md
```diff
@@ -141,7 +141,7 @@ const [state, send] = useActor(customActor, (actor) => {
 
 ### `useInterpret(machine, options?, observer?)`
 
-A React hook that returns the `service` created from the `machine` with the `options`, if specified. It also sets up a subscription to the `service` with the `observer`, if provided.
+A React hook that returns the `service` created from the `machine` with the `options`, if specified. It also sets up a subscription to the `service` with the `observer`, if provided. It starts the service and runs it for the lifetime of the component. Similar to using `useMachine`, but with an observer.
```
Andarist, line 144: We could somehow mention here that the main difference here is that `useMachine` pumps all changes to React's state (using the observer, since it's currently built on top of the `useInterpret`). While `useInterpret` gives you a more fine-grained control over this

mattpocock:
```suggestion
A React hook that returns the `service` created from the `machine` with the `options`, if specified. It starts the service and runs it for the lifetime of the component.

It also sets up a subscription to the `service` with the `observer`, if provided.

Using `useInterpret` is useful when you want to minimise re-renders. Where `useMachine` flushes each update from the machine to the component, but `useInterpret` returns a static reference which never causes a render.

To use a piece of state from the service inside a render, you'll need to use `useSelector` to subscribe to it.
```

redbar0n: @mattpocock should it be "never causes a render" or "never causes a re-render" ?

packages/xstate-react/README.md
```diff
@@ -141,7 +141,13 @@ const [state, send] = useActor(customActor, (actor) => {
 
 ### `useInterpret(machine, options?, observer?)`
 
-A React hook that returns the `service` created from the `machine` with the `options`, if specified. It also sets up a subscription to the `service` with the `observer`, if provided. It starts the service and runs it for the lifetime of the component. Similar to using `useMachine`, but with an observer.
+A React hook that returns the `service` created from the `machine` with the `options`, if specified. It starts the service and runs it for the lifetime of the component.
+
+It also sets up a subscription to the `service` with the `observer`, if provided.
+
+Using `useInterpret` is useful when you want to minimise re-renders. Where `useMachine` flushes each update from the machine to the component, but `useInterpret` returns a static reference which never causes a render.
```
davidkpiano:
```suggestion
Using `useInterpret` is useful when you want to minimise re-renders. Whereas `useMachine` flushes each update from the machine to the component, `useInterpret` instead returns a static reference to just the interpreted machine, which never causes rerenders.
```

Andarist (approved):

packages/xstate-react/README.md
mattpocock:
```suggestion
Using `useInterpret` is useful when you want to minimise re-renders. `useMachine` flushes each update from the machine to the component, but `useInterpret` returns a static reference which never causes a render.
```

mattpocock (approved):

mattpocock:
@redbar0n Could you resolve conflicts?

redbar0n:
@mattpocock I’m not sure how. It lists over half the file as a conflict. Someone may have moved the docs/contents somewhere else in the meanwhile? If so, then I don’t want to negate that, as I don’t have overview of the codebase.

PR #2690, opened 2021-09-26
[core] Improve action types with specific event generics

davidkpiano:
This PR adds a 3rd generic type, `TSpecificEvent`, to `Action<TContext, TEvent, TSpecificEvent>`. The purpose of this is to be a little more permissive of action objects that can take all events (`TEvent`), _without_ losing type inference for inline action functions.

EDIT: Depends on #2912 (`davidkpiano/extra`)


davidkpiano:
There are some weird differences between the TypeScript type checking in the test file and running the actual tests that I can't figure out (cc. @mattpocock @Andarist if you know how to fix this?)

Notice how the event is inferred properly in the `model.test.ts` file, but apparently not when the test is actually ran.

CleanShot 2021-09-26 at 08 59 25

mattpocock:
@davidkpiano are you using the same version of TS in your VSCode as the one in the workspace?

davidkpiano:
> @davidkpiano are you using the same version of TS in your VSCode as the one in the workspace?

Yes, regardless of which TS version I use in VS Code (even tried upgrading TS), I get the same result: expected types in the editor but not in the tests.

packages/core/test/model.test.ts
```diff
@@ -72,6 +72,60 @@ describe('createModel', () => {
...
+    const userModel = createModel<UserContext, UserEvent>({
+      name: 'David',
+      age: 30
+    });
+
+    userModel.assign({
+      name: (_, event) => {
+        // Incompatible because event.value can be string | number | undefined
+        // @ts-expect-error
+        return event.value;
```
Andarist, line 98: I think this is somewhat checking an incorrect thing here. It looks like your intention is to check assignability to the return value but actually you are suppressing property access error here:
https://www.typescriptlang.org/play?#code/C4TwDgpgBAqgzhATgUQG4QHbCgXgFBSFQA+UA3gUVaJAFxQDkArmACYCGwEAcuwLYQGAbkpVCqdgBsmEenGCIAlhgDmIqgF9RJclBqzGLDlwCCKwUKgTpBjEz4AjJFC1VSZPeAMN2GAPbAABZIaJjADC4ieABmTBgAxsCKfhhQ7HBwiioYABQQ6Fj08CEFwACU5NqIEMBMiKn5YQB01jIiGkA

This is because your `anotherEvent` doesn't declare `value` property at all - so you can't just freely access this property here without refining stuff first.

To fix that you could add `value?: undefined`  to the `anotherEvent`

packages/core/src/types.ts
```diff
@@ -215,9 +226,25 @@ export type TransitionTargets<TContext> = Array<
...
+  ? T extends (...args: any[]) => void
+    ? T
+    : T extends object
+    ? InferNarrowestObject<T>
+    : T
+  : never;
+export type InferNarrowestObject<T> = {
+  readonly [K in keyof T]: InferNarrowest<T[K]>;
+};
+export type NoInfer<T> = [T][T extends any ? 0 : never];
```
Andarist, line 239: are those used anywhere? 🤔

Andarist:
A very similar thing got implemented in https://github.com/statelyai/xstate/pull/3694

PR #2691, opened 2021-09-27
[core] Invoke meta data

davidkpiano:
This PR adds `meta` data to `invoke: { ... }`

### DIFF
--- a/packages/core/test/invoke.test.ts
+++ b/packages/core/test/invoke.test.ts
@@ -2600,6 +2600,19 @@ describe('invoke', () => {
       .onDone(() => done())
       .start();
   });
+
+  it('should show meta data', () => {
+    const machine = createMachine({
+      invoke: {
+        src: 'someSource',
+        meta: {
+          url: 'stately.ai'
+        }
+      }
+    });
+
+    expect(machine.invoke[0].meta).toEqual({ url: 'stately.ai' });
+  });
 });
 
 describe('services option', () => {
--- a/packages/core/test/invoke.test.ts
+++ b/packages/core/test/invoke.test.ts
@@ -2601,17 +2601,36 @@ describe('invoke', () => {
       .start();
   });
 
-  it('should show meta data', () => {
-    const machine = createMachine({
-      invoke: {
-        src: 'someSource',
-        meta: {
-          url: 'stately.ai'
+  describe('meta data', () => {
+    it('should show meta data', () => {
+      const machine = createMachine({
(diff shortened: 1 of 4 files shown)


mattpocock (approved):

packages/core/test/invoke.test.ts
```diff
@@ -2600,6 +2600,38 @@ describe('invoke', () => {
...
+      });
+
+      expect(machine.invoke[0].meta).toEqual({ url: 'stately.ai' });
+    });
+
+    it('meta data should be available in the invoke source function', () => {
+      expect.assertions(1);
+      const machine = createMachine({
+        invoke: {
+          src: (_ctx, _e, { meta }) => {
```
Andarist, line 2622: this looks pretty much the same as what we already support for actions but we use `action` there and not `meta`:
https://github.com/statelyai/xstate/blob/a728060353c9cb9bdb0cd37aacf793498a8750c8/packages/core/test/model.test.ts#L361

I don't care that much what we use here but I think we should be consistent. It makes sense for me to not use `meta` though since `meta` is something that is completely untyped at the moment, it's a bag for random stuff. And hopefully, in the future, we might be able to provide some types for this

Andarist: It's also worth looking into how one declares those - cleaning everything up might not be possible in the v4 but we should decide on how this should look in v5 and create a task for unifying this across all implementation types. Maybe it would make sense to decouple the type and other params for everything and use `params` here on the meta object? This would require changing enforcing this kind of a shape for parametrized actions: `{ type: string; params: Record<string, any> }`. While this looks like smth harder to declare I think it has the potential of being better for the runtime as we don't have to do any custom parsing/merging/excluding/whatever - we just need to pass the `params` property around and it can nicely be unified for all of the implementation types. Note that a DSL and the visual builder would make declaring those at state nodes easier - so the proposed additional property nesting doesn't have to be that bad

Andarist: > EDIT: Are you proposing this shape for meta?

No, if this is meant to act as a "bag for random stuff" then it's OK to me as it is. The given example here, in the test code, just made me think more about the params than the meta object.

Andarist (approved):

packages/core/test/invoke.test.ts
davidkpiano: Yeah, sorry about that - the test code example is not the best example.

PR #2697, opened 2021-09-29
Separate package documentation from package readmes

laurakalbag:


docs/README.md
```diff
@@ -171,7 +171,7 @@ Read [📽 the slides](http://slides.com/davidkhourshid/finite-state-machines) (
 - [The World of Statecharts](https://statecharts.github.io/) by Erik Mogensen
 - [Pure UI](https://rauchg.com/2015/pure-ui) by Guillermo Rauch
 - [Pure UI Control](https://medium.com/@asolove/pure-ui-control-ac8d1be97a8d) by Adam Solove
-- [Spectrum - Statecharts Community](https://spectrum.chat/statecharts) (For XState specific questions, please use the [GitHub Discussions](https://github.com/davidkpiano/xstate/discussions))
+- [Spectrum - Statecharts Community](https://spectrum.chat/statecharts) (For XState specific questions, please use the [GitHub Discussions](https://github.com/statelyai/xstate/discussions))
```
davidkpiano, line 174:
```suggestion
```

This community is no longer active

docs/packages/core/index.md
```diff
@@ -83,7 +83,7 @@ Read [📽 the slides](http://slides.com/davidkhourshid/finite-state-machines) (
 - [The World of Statecharts](https://statecharts.github.io/) by Erik Mogensen
 - [Pure UI](https://rauchg.com/2015/pure-ui) by Guillermo Rauch
 - [Pure UI Control](https://medium.com/@asolove/pure-ui-control-ac8d1be97a8d) by Adam Solove
-- [Spectrum - Statecharts Community](https://spectrum.chat/statecharts) (For XState specific questions, please use the [GitHub Discussions](https://github.com/davidkpiano/xstate/discussions))
+- [Spectrum - Statecharts Community](https://spectrum.chat/statecharts) (For XState specific questions, please use the [GitHub Discussions](https://github.com/statelyai/xstate/discussions))
```
davidkpiano, line 86:
```suggestion
```

davidkpiano (approved):
🎉 🎉

Andarist (approved):
It makes me a little bit sad since it makes README pages on npm (like this one https://www.npmjs.com/package/@xstate/fsm ) much less useful when it comes to the docs (and making them available offline in `node_modules`) but I see how it simplifies maintenance so :+1:

docs/package.json
```diff
@@ -3,14 +3,13 @@
   "private": true,
```
Andarist, line 3: q: will this be automatically released when merged to `main`? I'm just wondering if there is any risk of breaking the published docs after we land this

laurakalbag:
> It makes me a little bit sad since it makes README pages on npm (like this one https://www.npmjs.com/package/@xstate/fsm ) much less useful when it comes to the docs (and making them available offline in `node_modules`) but I see how it simplifies maintenance so 👍

It might be worth exploring whether we can do some content fragments/headless-ish content where we can pull different content into the package indexes and readmes through templates while only having to maintain one source.

docs/package.json
laurakalbag: Shouldn’t break anything unless I’ve missed something major. The files were just being copied across during the build process, so removing that stops any content being overwritten. And I’ve amended both versions of the files to be better suited to their destinations… that should be enough?

Andarist: Sounds OK - I'm not too familiar with the existing process though, just wanted to double-check

PR #2712, opened 2021-10-04
chore(core): update `raise` typings to match `send`

### LINKED ISSUE #2710: using `actions.raise` with a callback results in a type error.
### Description
Using `actions.raise` with a callback results in a type error.

### Expected Result
- [x] The `raise` action is executed with the same outcome as `send`.
- [x] No type errors are present.

### Actual Result
- [x] The `raise` action is executed with the same outcome as `send`.
- [ ] No type errors are present.
```log
Argument of type '() => { type: string; }' is not assignable to parameter of type 'Event<EventObject>'.
Property 'type' is missing in type '() => { type: string; }' but required in type 'EventObject'.ts(2345)
```

### Reproduction
https://codesandbox.io/s/empty-silence-7t6c2?file=/src/index.ts:459-504

Enteleform:
Fixes #2710

`raise` was programmatically handling `send`-shaped events, including callbacks; however, it wasn't typed properly to signal that it was capable of doing so.

I merged the relevant parts of the `send` function signature with `raise`, and refactored the test into a more concise implementation as per [**this discussion**](https://discord.com/channels/795785288994652170/801127053092716596/894393154743197717) with @davidkpiano on Discord.


packages/core/src/actions.ts
```diff
@@ -152,11 +152,16 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
  *
  * @param eventType The event to raise.
  */
-export function raise<TContext, TEvent extends EventObject>(
-  event: Event<TEvent>
-): RaiseAction<TEvent> | SendAction<TContext, AnyEventObject, TEvent> {
+export function raise<
+  TContext,
+  TEvent extends EventObject,
+  TSentEvent extends EventObject = AnyEventObject
```
davidkpiano, line 158: This shouldn't be necessary; it should be the same as `TEvent` for raising events

Enteleform: `TSentEvent` ?

so like; `SendExpr<TContext, TEvent, TEvent>`, etc. instead ?

Enteleform: Haven't dug that deep into the internal differences between `send` and `raise`, just assumed that since non-string events are being forwarded to `send` that using the same typings would provide the same IntelliSense & whatnot.

davidkpiano: > `TSentEvent` ?
> 
> so like; `SendExpr<TContext, TEvent, TEvent>`, etc. instead ?

Yes, like this, I'm pretty sure

Enteleform:
@Andarist Thanks for the thorough review & details! Sorry for the delay, been swamped lately. I just took another shot at this after checking out your TS playground example, but haven't been able to figure out a solution that addresses the issues that you highlighted just yet.

Might have to come back to this after I wrap up my current project, been kind of short on time & the mental energy required for type programming (*been spending way too much time in React render optimization land lately 💀*).

Andarist:
Superseded by https://github.com/statelyai/xstate/pull/3393

PR #2717, opened 2021-10-05
Align send functionality for `spawnBehavior` with interpreted machine

### LINKED ISSUE #2706: Spawned Behavior Does Not Fully Support Actorref's `send` Type
I noticed that the `send` function for actors that are spawned from behavior does not behave the same as the `send` function for interpreted machines. Generally, the `ActorRef` interface uses the `Sender<TEvent>` type for its send function, i.e. both `send("hello")` and  `send({type: "hello"})` are supported.

This works as expected if the `ActorRef` refers to an interpreted machine. However, `Behavior` that is spawned with `spawnBehavior` allows `send("hello")` since it is an `ActorRef`, but it is not correctly processed at runtime. The implementation around `spawnBehavior`, `toActorRef`, and `BaseActorRef` works with `send: (e: TEvent) => void`.

I would like to unify the functionality around `Behaviors` and `StateMachines` and this might be the first step in the right direction. Before I can contribute a pull-request, we should clarify what type definition is expected in the future, i.e. should `send("hello")` be supported? A [*TODO*](https://github.com/statelyai/xstate/blob/9fb87ab0b637dda6fb21aaedcd0b0ff62e3e230f/packages/core/src/types.ts#L1382) suggests otherwise in the source code. However, @davidkpiano [mentioned](https://discord.com/channels/795785288994652170/799416911259369472/893270046376341534) that he would like to keep the `send("hello")` shortcut on Discord.

christoph-fricke:
This PR fixes #2706.

Actors that are spawned with `spawnBehavior` are typed as `ActorRefs`. This means that they should support the `Sender` type, i.e. `send("event")`.  However, a behavior normally expects event objects so string events are ignored. This PR fixes the problem by converting string events to event objects before they are passed to the spawned behavior in `spawnBehavior`, which is equivalent to the functionality in the interpreter.

### DIFF
--- a/.changeset/smooth-ducks-judge.md
+++ b/.changeset/smooth-ducks-judge.md
@@ -1,0 +1,13 @@
+---
+'xstate': patch
+---
+
+Adds support for calling `send` with a string as the event type to `spawnBehavior`.
+This aligns actors that are spawned from behavior with actors that are spawned from machines.
+
+```typescript
+const actor = spawnBehavior(/* ... */);
+
+actor.send('test'); // Will be converted to an event object {type: "test"}
+actor.send({ ŧype: 'test' }); // Continuous to work as before
+```
--- a/.changeset/smooth-ducks-judge.md
+++ b/.changeset/smooth-ducks-judge.md
@@ -9,5 +9,5 @@ This aligns actors that are spawned from behavior with actors that are spawned f
 const actor = spawnBehavior(/* ... */);
 
 actor.send('test'); // Will be converted to an event object {type: "test"}
-actor.send({ ŧype: 'test' }); // Continuous to work as before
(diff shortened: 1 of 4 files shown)


davidkpiano:
This change makes sense to me 👍 awaiting comments from @mattpocock @farskid @Andarist

christoph-fricke:
Awesome :+1:
I'm not totally sure about the semver version in the changeset. When I first tried to use `send("event")` with spawned behavior, I was surprised that it is not converted to an object. So I have classified this PR as a _patch_. However, depending on the point of view this might as well be a _minor_ release. What do you think?

Andarist:
I'm sorry that this one has been overlooked in the past. In v5 the related APIs got a bigger overhaul and I think that with those changes this PR can be closed.

PR #2720, opened 2021-10-06
Pass events with type `xstate.*` through the inspector to the receiver

farskid:
The primary reasoning behind this PR is to allow `xstate.disconnect` to be received by the devtools receiver. This is useful in cases in which we want to locally re-interpret an inspected machine. One of such cases is embedded inspected machines in blogs where you want to reset the machine to the initial state.

### DIFF
--- a/.changeset/fair-eggs-rule.md
+++ b/.changeset/fair-eggs-rule.md
@@ -1,0 +1,5 @@
+---
+'@xstate/inspect': minor
+---
+
+Bypass events with type `xstate.*` to pass through xstate/inspect to the receiver
--- a/packages/xstate-inspect/src/browser.ts
+++ b/packages/xstate-inspect/src/browser.ts
@@ -135,7 +135,7 @@ export function inspect(
   window.addEventListener('message', messageHandler);
 
   window.addEventListener('unload', () => {
-    inspectService.send({ type: 'unload' });
+    inspectService.send({ type: 'inspector.restart' });
   });
 
   devTools.onRegister((service) => {
--- a/packages/xstate-inspect/src/inspectMachine.ts
+++ b/packages/xstate-inspect/src/inspectMachine.ts
@@ -55,7 +55,7 @@ export function createInspectMachine(
               service?.send(scxmlEventObject);
             }
           },
-          unload: {
+          'inspector.restart': {
             actions: (ctx) => {
               ctx.client!.send({ type: 'xstate.disconnect' });
             }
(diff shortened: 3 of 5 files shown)


Andarist:
I'm not saying that we should pursue the ideal solution for now in terms of the `@xstate/inspect` implementation, or the API that would allow for other types of integrations (like installing the inspector at any time, rather than only before starting any machines. 

Introducing a new, dedicated, event type should be an easy thing to do - `xstate.*` seems to be rather reserved for internal events dispatched by the core interpreter (such as `xstate.init`). What I'm proposing is basically something along those lines:
```diff
diff --git a/packages/xstate-inspect/src/browser.ts b/packages/xstate-inspect/src/browser.ts
index f243de1f0..9c10e4820 100644
--- a/packages/xstate-inspect/src/browser.ts
+++ b/packages/xstate-inspect/src/browser.ts
@@ -135,7 +135,7 @@ export function inspect(
   window.addEventListener('message', messageHandler);
 
   window.addEventListener('unload', () => {
-    inspectService.send({ type: 'unload' });
+    inspectService.send({ type: 'inspector.restart' });
   });
 
   devTools.onRegister((service) => {
diff --git a/packages/xstate-inspect/src/inspectMachine.ts b/packages/xstate-inspect/src/inspectMachine.ts
index 540dd4aee..77999ff0c 100644
--- a/packages/xstate-inspect/src/inspectMachine.ts
+++ b/packages/xstate-inspect/src/inspectMachine.ts
@@ -55,10 +55,8 @@ export function createInspectMachine(
               service?.send(scxmlEventObject);
             }
           },
-          unload: {
-            actions: (ctx) => {
-              ctx.client!.send({ type: 'xstate.disconnect' });
-            }
+          'inspector.restart': {
+            actions: (ctx, e) => ctx.client!.send(e)
           },
           disconnect: 'disconnected'
         }
diff --git a/packages/xstate-inspect/src/types.ts b/packages/xstate-inspect/src/types.ts
index e30964908..bf1ab4990 100644
--- a/packages/xstate-inspect/src/types.ts
+++ b/packages/xstate-inspect/src/types.ts
@@ -45,7 +45,8 @@ export type ReceiverEvent =
       state: string;
       sessionId: string;
     }
-  | { type: 'service.event'; event: string; sessionId: string };
+  | { type: 'service.event'; event: string; sessionId: string }
+  | { type: 'inspector.restart' };
 
 export type ParsedReceiverEvent =
   | {
```

davidkpiano:
Closing as outdated

PR #2737, opened 2021-10-14
Add failing tests for regressions introduced in the initial changes of #2736

Andarist:


Andarist:
Ok, I'll land this into your PR - even though a lot has changed since your approval. Both PRs should really be reviewed and landed as a whole so it makes sense to just merge them now and review the whole thing.

PR #2743, opened 2021-10-19
add optional event type parameter for narrowing types returned by `EventFrom`

janovekj:
From @davidkpiano's [idea on the Discord](https://discord.com/channels/795785288994652170/799416911259369472/877211065878999101). 

```ts
const userModel = createModel(
  {},
  {
    events: {
      updateName: (value: string) => ({ value }),
      updateAge: (value: number) => ({ value }),
      anotherEvent: () => ({})
    }
  }
);

type UpdateNameEvent = EventFrom<typeof userModel, "updateName">
// =>  { type: "updateName", value: string }  
```

It also comes with autocomplete:
bilde

### DIFF
--- a/.changeset/neat-timers-deny.md
+++ b/.changeset/neat-timers-deny.md
@@ -2,4 +2,8 @@
 'xstate': minor
 ---
 
-Add optional type parameter to narrow type returned by EventFrom
+Add optional type parameter to narrow type returned by `EventFrom`. You can use it like this:
+
+```ts
+type UpdateNameEvent = EventFrom<typeof userModel>;
+```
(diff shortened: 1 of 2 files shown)


janovekj:
> Looks good! I'm assuming this works as normal with `EventFrom<T>` (no narrowed event)

Thanks! Yep, correct!

Andarist:
It would be great if we could add some tests for this type helper as it becomes slightly more complicated now

Andarist:
> @Andarist I can look into this! Are there any existing type tests in the project I should look at for reference?

I think we have some [here](https://github.com/statelyai/xstate/blob/c2d6a24bcc7174af543e30aefd2a62c95a9b83ca/packages/core/test/types.test.ts) - but for those, you could create a dedicated file, like `typeHelpers.test.ts`. I have also been lately implementing similar tests [here](https://github.com/statelyai/xstate/blob/feb7044d036a0fe0dae822c191a692fb5579936a/packages/xstate-react/test/typegenTypes.test.tsx) so this could serve as an inspiration as well.

PR #2754, opened 2021-10-26
docs: Typings for the context and the service

ilabacheuskicastor:
This PR will properly type the context.
With such way we could inform the users how to properly type our service, since right now it is not clear that we could use the generic `InterpreterFrom<>`

### DIFF
--- a/docs/recipes/react.md
+++ b/docs/recipes/react.md
@@ -63,33 +63,13 @@ Using `useInterpret` returns a service, which is a static reference to the runni
 
 #### Usage with typescript
 
-To have the strict typings with typescript for your service, you could use the generic `InterpreterFrom` from the `xstate`.
+The return type from `useInterpret` could be created by the generic `InterpreterFrom` from the `xstate`.
 
 ```typescript
-import { createContext, FC } from 'react';
-import { useInterpret } from '@xstate/react';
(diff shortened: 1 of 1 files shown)


mattpocock:
@davidkpiano @ilabacheuskicastor An extra TS example here feels a bit too verbose, we could just add:

`export const GlobalStateContext = createContext({} as InterpreterFrom<typeof authMachine>);`

to the existing example to make it fully TS compliant?

mattpocock:
Or add it as a:

```
> For Typescript, you can type the context as `createContext({} as InterpreterFrom<typeof authMachine>);` to ensure you get proper autocomplete.
```

davidkpiano:
> Or add it as a:
> 
> ```
> > For Typescript, you can create the context as `createContext({} as InterpreterFrom<typeof authMachine>);` to ensure strong typings.
> ```

From the two options, adding it as a note (like this) is sufficient

ilabacheuskicastor:
@mattpocock @davidkpiano I will update

ilabacheuskicastor:
LGTM, @mattpocock I had written the full example, bcs in my work I was unaware of the existing InterpreterFrom generic in the xstate. And Other generics was not very helpful :)

PR #2769, opened 2021-10-29
Added a tip for `.onTransition()` regarding microsteps.

jenny-tru:


docs/guides/interpretation.md
```diff
@@ -134,8 +134,13 @@ service.onTransition(state => {
 });
 ```
 
+:::
```
davidkpiano, line 137:
```suggestion
```

PR #2776, opened 2021-10-31
Add `self` to invoke meta

davidkpiano:
This PR adds `self` (reference to invokee) to the invoke meta.

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -880,7 +880,8 @@ export class Interpreter<
             ? serviceCreator(context, _event.data, {
                 data: resolvedData,
                 src: invokeSource,
-                meta: activity.meta
+                meta: activity.meta,
+                self: this
               })
             : serviceCreator;
 
--- a/packages/core/test/invoke.test.ts
+++ b/packages/core/test/invoke.test.ts
@@ -2632,6 +2632,31 @@ describe('invoke', () => {
       interpret(machine).start();
     });
   });
+
+  it('invoke creator should provide reference to self (invokee)', (done) => {
+    const machine = createMachine({
+      invoke: {
+        src: (_ctx, _e, { self }) => () => {
+          self.send('FROM_INVOKED');
+        }
+      },
+      initial: 'pending',
+      states: {
+        pending: {
+          on: {
+            FROM_INVOKED: 'success'
+          }
+        },
+        success: {
+          type: 'final'
+        }
+      }
+    });
+
+    interpret(machine)
+      .onDone(() => done())
+      .start();
+  });
 });
 
 describe('services option', () => {
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -299,9 +299,10 @@ export type InvokeCallback<
   onReceive: Receiver<TEvent>
 ) => (() => void) | Promise<any> | void;
 
-export interface InvokeMeta {
+export interface InvokeMeta<TSelfRef extends ActorRef<any, any>> {
   data: any;
   src: InvokeSourceDefinition;
+  self: TSelfRef;
   meta?: MetaObject;
 }
 
@@ -325,7 +326,7 @@ export type InvokeCreator<
 > = (
   context: TContext,
   event: TEvent,
-  meta: InvokeMeta
+  meta: InvokeMeta<ActorRef<TEvent, State<TContext, TEvent>>>
 ) =>
   | PromiseLike<TFinalContext>
   | StateMachine<TFinalContext, any, any>


packages/core/test/invoke.test.ts
```diff
@@ -2632,6 +2632,31 @@ describe('invoke', () => {
       interpret(machine).start();
     });
   });
+
+  it('invoke creator should provide reference to self (invokee)', (done) => {
```
Andarist, line 2636: Isn't the child actor an invokee here? 
```suggestion
  it('invoke creator should provide reference to self', (done) => {
```

packages/core/src/interpreter.ts
```diff
@@ -880,7 +880,8 @@ export class Interpreter<
             ? serviceCreator(context, _event.data, {
                 data: resolvedData,
                 src: invokeSource,
-                meta: activity.meta
+                meta: activity.meta,
+                self: this
```
Andarist, line 884: Could we add `this.self = toActorRef(this)` to the constructor (the exact code might differ from this example) and pass that here?

We should limit what the user has access to from the callback - it's already limited on the type-level (since it got typed as `ActorRef<TEvent, State<TContext, TEvent>>`) but we should also limit it in the runtime.

Do we support restarting interpreters (as in `s.start(); s.stop(); s.start()`)? If yes then the `this.self` would have to be created in the `start` method.

We should probably also ensure that the `stop` method is not exposed on this (the `ActorRef` currently allows it). It sounds like a bad idea to allow a child to stop its parent imperatively, right?

Andarist:
Just a reminder - this needs a changeset 😉

packages/core/src/interpreter.ts
davidkpiano: Done!

ivandotv:
Waiting on this.
This feature is going to open up a lot of possibilities.

Also, I think that the `self` reference should also be added to plain `actions` arguments like discussed [here](https://discord.com/channels/795785288994652170/799416943324823592/900053743071010827).

packages/core/test/invoke.test.ts
```diff
@@ -2709,6 +2709,31 @@ describe('invoke', () => {
       interpret(machine).start();
     });
   });
+
+  it('invoke creator should provide reference to self', (done) => {
+    const machine = createMachine({
+      invoke: {
+        src: (_ctx, _e, { self }) => () => {
+          self.send('FROM_INVOKED');
```
Andarist, line 2717: It would be great if we could have some type tests that would verify that only accepted events can be sent to `self`. This test types `self` as:
```ts
ActorRef<AnyEventObject, State<unknown, AnyEventObject, any, {
    value: any;
    context: unknown;
}>>
```
so there is no such validation happening here. Any event gets accepted.

Also - the `stop` method is still a part of `self` (but it's optional). I wonder if we want to hide it completely in the future?

packages/core/src/interpreter.ts
```diff
@@ -883,7 +888,8 @@ export class Interpreter<
             ? serviceCreator(context, _event.data, {
                 data: resolvedData,
                 src: invokeSource,
-                meta: activity.meta
+                meta: activity.meta,
+                self: toActorRef(this)
```
Andarist, line 892: Re chosen naming - I wonder how using `self` will play out here. It will get used in callbacks etc and I wonder if using `parent.send()` wouldn't make more sense in that context.

packages/core/test/invoke.test.ts
```diff
@@ -2709,6 +2709,31 @@ describe('invoke', () => {
       interpret(machine).start();
     });
   });
+
+  it('invoke creator should provide reference to self', (done) => {
```
Andarist, line 2713: Especially `getSnapshot` is going to be quite useful too. I think it does deserve an additional, basic, test.

packages/core/src/interpreter.ts
Andarist: Also - it would probably make sense to create this once, instead of recreating it for each invoke. This ensures that those refs are isolated between invokes but I don't see why this would matter in practice.

davidkpiano: `self` is what is used in Akka, and it makes sense to me: it is a _self-reference_. Naming it something like `parent` would get confusing, and actually be completely wrong, since it's not a reference to the parent of this actor.

The alternative is calling it `ref`, `selfRef`, etc. I'd prefer to stick with Akka naming.

packages/core/test/invoke.test.ts
davidkpiano: > It would be great if we could have some type tests that would verify that only accepted events can be sent to self.

This is recently added to 934fa862c5cc13bcc01fec829309e9e4a184ee3a

> Also - the stop method is still a part of self (but it's optional). I wonder if we want to hide it completely in the future?

That makes sense - will do in a separate PR for v5.

packages/core/src/interpreter.ts
davidkpiano: > Also - it would probably make sense to create this once, instead of recreating it for each invoke. This ensures that those refs are isolated between invokes but I don't see why this would matter in practice.

This is done in 2a6a822f89a6c120b586f0cc17de281df5e88d67

packages/core/test/invoke.test.ts
davidkpiano: This is done in 4c3b9b9653a333bb347c00f715a6e0e3784f95d0

Andarist:
There are two type problems that have to be addressed before landing this, the call sites with errors are shown here:
```diff
diff --git a/packages/core/src/State.ts b/packages/core/src/State.ts
index 6aef9796d..96ef66de7 100644
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -72,11 +72,15 @@ export function bindActionToState<TC, TE extends EventObject>(
     exec:
       exec !== undefined
         ? () =>
-            exec(state.context, state.event as TE, {
-              action,
-              state,
-              _event: state._event
-            })
+            exec(
+              state.context,
+              state.event as TE,
+              {
+                action,
+                state,
+                _event: state._event
+              } as any
+            )
         : undefined
   };
 
diff --git a/packages/xstate-react/src/useReactEffectActions.ts b/packages/xstate-react/src/useReactEffectActions.ts
index d7657ed0b..527f07d66 100644
--- a/packages/xstate-react/src/useReactEffectActions.ts
+++ b/packages/xstate-react/src/useReactEffectActions.ts
@@ -13,7 +13,7 @@ function executeEffect<TContext, TEvent extends EventObject>(
     action,
     state,
     _event: state._event
-  });
+  } as any);
 
   originalExec();
 }

```

packages/core/src/interpreter.ts
Andarist: While the points related to `.origin` raised in https://github.com/statelyai/xstate/pull/2887#issue-1086967204 were invalid, I think that we should still revisit naming this "self". In Akka (maybe just in Classic, but that doesn't affect naming considerations) one gets access to both `context.self` and `context.parent` and I think that was is provided here is much closer to `context.parent`.

It's because it's the child actor that sends an event to its parent and not either that parent sending an event to self or that child sending an event to self.

davidkpiano: Maybe `invoker`?

Andarist: `invoker` would work OK. Any particular objections to just `parent` though?

davidkpiano: > `invoker` would work OK. Any particular objections to just `parent` though?

Just worried it would be confused with "parent of this machine" or "parent of the invoked actor"

Andarist: Can't something like "invoker" be misinterpreted in the same way? Both names are sensitive to the context and if somebody assumes a different context here then they would misinterpret both in the same way. I guess maybe `invoker` is harder to misinterpret but it's also not that clear if it's the invoker of this machine of the invoker of the invoked actor 🤔 

I've gone back to other comments by @ivandotv and @erikras though and I think that after all the chosen name (self) is right. I still believe that this might be somewhat confusing because very often u might want to `self.send()` from within a child actor and this won't really send to its own "self" but this is somewhat a closure/nesting problem. At the very time when `serviceCreator` gets called, there is no "self" for the child actor yet - and if we'd make this available to other types of implementations (such as actions) then there would be no "self" for those implementation types at all. So whatever is given at this very position/time in the `meta` param has to refer to the machine that is "calling" those implementations and `self` suits that.

davidkpiano: From the Akka docs:

> **parent**
> There is no `parent` in Typed. Instead you have to explicitly include the `ActorRef` of the parent as a parameter when constructing the Behavior.
>
> The reason for not having a parent in Typed is that it wouldn’t be possible to know the type for the parent ActorRef[T] at compile time without having an additional type parameter in the Behavior. For testing purposes it’s also better to pass in the parent since it can be replaced by a probe or being stubbed out in tests.

davidkpiano: We can also provide good patterns in the docs for naming:

```js
invoke:

Andarist: > The reason for not having a parent in Typed is that it wouldn’t be possible to know the type for the parent ActorRef[T] at compile time without having an additional type parameter in the Behavior. For testing purposes it’s also better to pass in the parent since it can be replaced by a probe or being stubbed out in tests.

Yeah, I agree with this reasoning - if we follow that advice though then we should deprecate `sendParent` though (and similar things too, like `origin`) and the question is: is it pragmatic for us to deprecate it? 😅

ivandotv: This `self` thing can be a little bit confusing but the example that David has provided is pretty self-explanatory.

davidkpiano: > self-explanatory

I see what you did there @ivandotv 😂

Andarist:
Closing this PR since this is already done one the next branch and we don't plan to introduce this in v4.

PR #2779, opened 2021-11-01
Implemented input

mattpocock:
Changeset copied here for ease:

# Input

Introduced 'input' - a way of specifying dependencies to the machine which cannot be assigned to. Input can be specified like so:

```ts
  {
    entry: (context, event, meta) => {
      console.log(meta.state.input.lyric); // 'In west philadelphia, born and raised'
    }
  },
  {
    input: {
      lyric: 'In west philadelphia, born and raised'
    }
  }
);
```

### Re-specifying initial inputs

You can re-specify them using `.withConfig`:

```ts
const machineWithDifferentInput = machine.withConfig({
  input: {
    lyric: 'In the playground, where I spend most of my days'
  }
});
```

### Updating inputs

You can update inputs on a running service using the `service.input()` method:

```ts

service.input({
  lyric: 'Chilling out, maxing, relaxing all cool'
});
```

### State

Input is available on the `state` property:

```ts
console.log(service.state.input.lyric); // 'Chilling out, maxing, relaxing all cool'
```

### Listening for new inputs

You can listen for updates to the inputs on the `xstate.input` event:

```ts
  on: {
    'xstate.input': {
      actions: (context, event, meta) => {
        console.log(meta.state.input);
      }
    }
  }
});
```

> Inputs always resolve _before_ any actions relating to the `xstate.input` transitions are run. This means that `meta.state.input` will always be up to date.

### createModel

Inputs can be made type-safe using `createModel`:

```ts
const model = createModel(
  {},
  {
    input: {} as {
      lyric: string;
    }
  }
);

  {
    entry: [
      (context, event, meta) => {
        // Input is type-safe
        console.log(meta.state.input.lyric);
      }
    ]
  },
  {
    input: {
      lyric: 'And all shooting some b-ball outside of the school'
    }
  }
);
```

### Use with other hooks

Input can now be passed into `useMachine` and `useInterpret`:

```ts
  const [result] = useQuery();

  const [state, send] = useMachine(machine, {
    input: {
      // props.id and result.data will be kept up to date
      // in the machine above
      id: props.id,
      data: result.data
    }
  });
};
```

Each render, the input will be kept up to date inside the machine, allowing for easy integration with other hooks and props.


mattpocock:
@Andarist this is failing for the JS Heap out of memory thing, what was the fix?

packages/core/src/State.ts
```diff
@@ -21,6 +21,7 @@ import { StateNode } from './StateNode';
 import { getMeta, nextEvents } from './stateUtils';
 import { initEvent } from './actions';
 import { IS_PRODUCTION } from './environment';
+import { ExtraGenerics } from '.';
```
davidkpiano, line 24: Will `ExtraGenerics` work fine (not conflict) with the typegen work?

mattpocock: Yeah, we'd just need to have another generic for the Typegen. If we start recommending `createModel` as the default it means this isn't so user-facing.

Using `ExtraGenerics` as an object means we only need to do this refactor once, too - we can include any other optional generics on the `ExtraGenerics` object

packages/core/src/StateNode.ts
```diff
@@ -1242,9 +1291,24 @@ class StateNode<
       ? currentState.configuration
       : [];
 
+    const currentInput: NonNullable<TExtra['input']> =
+      currentState?.input || this.options.input!;
+
+    let updatedInput: NonNullable<TExtra['input']> | undefined = undefined;
+
+    if (_event?.name === actionTypes.input) {
+      updatedInput = Object.assign(currentInput, (_event.data as any).input);
```
davidkpiano, line 1300:
```suggestion
      updatedInput = Object.assign({}, currentInput, (_event.data as any).input);
```

We want to keep track of previous input in `state.history` instead of mutating it

mattpocock: Need to add a test to that effect before committing this

packages/core/test/input.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      {},
+      {
+        input: {
+          // @ts-expect-error
+          bar: ''
+        }
+      }
+    );
+  });
+});
```
davidkpiano, line 182: Can we add a test for partial input? If the initial input is:

```js
input: {
  foo: 'bar',
  count: 0
}
```

We should be able to send `service.input({ count: 9 })` and have the resulting input be:

```js
input: {
  foo: 'bar',
  count: 9
}
```

davidkpiano (commented):
Really excited for this! Just a couple of tests and changes to be made; I can help out with the tests if needed.

@Andarist if you're reviewing, most of the files here are just `TExtra` additions; otherwise it's fairly straightforward.

davidkpiano:
Actually one more question: can we confirm that `TExtra` won't conflict with the typegen work (cc. @Andarist)?

.changeset/old-flies-peel.md
```diff
@@ -0,0 +1,124 @@
...
+### createModel
+
+Inputs can be made type-safe using `createModel`:
+
+```ts
+const model = createModel({}).withInput(
+  {} as {
+    lyric: string;
+  }
+);
```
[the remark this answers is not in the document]
mattpocock: @davidkpiano Could you clarify what you mean here?

mattpocock:
@davidkpiano It will be a pain to merge them, but mostly it'll be just grunt work, adding the extra generics.

packages/core/src/StateNode.ts
mattpocock: Nice, added a test which failed without this change and passed with it. Good eyes!

packages/core/src/interpreter.ts
```diff
@@ -541,6 +579,17 @@ export class Interpreter<
...
+
+  public input(
+    input: Partial<TExtra['input']>
+  ): State<TContext, TEvent, TStateSchema, TTypestate, TExtra> {
+    this.send(({
+      type: actionTypes.input,
+      input
+    } as any) as SCXML.Event<TEvent>);
+
+    return this._state!;
```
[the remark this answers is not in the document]
mattpocock: Done

packages/core/src/interpreter.ts
```diff
@@ -541,6 +579,13 @@ export class Interpreter<
 
     return this;
   }
+
+  public input(input: Partial<TExtra['input']>): void {
```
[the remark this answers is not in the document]
Andarist: If we ever decide to make the argument partial here we will still be able to do that. Although it will become a little bit more complex because we'll have to figure out **where** to resolve the input to the whole input (assuming we would still want to maintain the `event.input` to have all properties). 

We can't resolve here, because when this is called we can be in the middle of the event queue processing and we won't have the "current" input if the queue looks like this: `[foo, xstate.input]`. Any already scheduled `xstate.input` can update the "current" input value and everything should be processed assuming in that sequence etc.

packages/core/src/model.ts
```diff
@@ -66,6 +67,21 @@ export function createModel(
         'context' in config ? config : { ...config, context: initialContext },
         implementations
       );
+    },
+    withInput: (input: any) => {
+      const model = createModel(initialContext, creators as any);
```
Andarist, line 72: we should be able to just spread the existing `model` from the outer scope, right? any particular reason that we recreate the model here?

packages/core/src/model.ts
```diff
@@ -66,6 +67,21 @@ export function createModel(
...
+
+      return {
+        ...model,
+        createMachine: (config, implementations) => {
+          return createMachine(
+            'context' in config
+              ? config
+              : { ...config, context: initialContext },
+            { input, ...implementations }
+          );
```
Andarist, line 82: that's equivalent to:
```suggestion
          return model.createMachine(config, { input, ...implementations });
```
which makes the behavior (regarding the `config.context`) of those two the same, out of the box

packages/core/src/model.ts
```diff
@@ -66,6 +67,21 @@ export function createModel(
...
+      const model = createModel(initialContext, creators as any);
+
+      return {
+        ...model,
+        createMachine: (config, implementations) => {
+          return createMachine(
+            'context' in config
+              ? config
+              : { ...config, context: initialContext },
+            { input, ...implementations }
```
Andarist, line 81: q: should this merge with "existing" input or should it overwrite it? I feel like our rules for this sort of thing are somewhat inconsistent, or at least under-specified, and that all our APIs should behave in the same way when it comes to this sort of thing. Unless there is a strong reason why they shouldn't.

At the moment I feel like a newcomer has to consult the docs or the implementation a lot, to know which behavior is used where. And even users who are using XState for a long time already are regularly confused by our APIs not being consistent in such scenarios, a recent tweet: https://twitter.com/rafalfilipek/status/1460236700207271938

packages/xstate-react/src/useInterpret.ts
```diff
@@ -128,7 +140,8 @@ export function useInterpret<
     Object.assign(service.machine.options.activities, activities);
     Object.assign(service.machine.options.services, services);
     Object.assign(service.machine.options.delays, delays);
-  }, [actions, guards, activities, services, delays]);
+    Object.assign(service.machine.options.input, input);
+  }, [actions, guards, activities, services, delays, input]);
```
Andarist, line 144: Could you point me to a test that would fail if we'd remove this dependency from here (and the associated `Object.assign` call)? It would help me to understand this better

packages/xstate-react/src/useInterpret.ts
```diff
@@ -119,6 +125,12 @@ export function useInterpret<
     };
   }, []);
 
+  useIsomorphicLayoutEffect(() => {
+    if (input) {
+      service.input(input);
```
Andarist, line 130: Without bumping the peer dep requirement we can't be sure if the installed version of XState supports this method.
```suggestion
      service.input?.(input);
```

packages/xstate-react/src/useInterpret.ts
```diff
@@ -119,6 +125,12 @@ export function useInterpret<
...
+    }
+  }, [...Object.keys(input || {}), ...Object.values(input || {})]);
```
Andarist, line 132: As part of the "dont allow users to write weird code"... I would just check the values here. The "input" object should really be only used inline and should always contain the same keys. Even if we assume that this wouldn't be a requirement then this code would still not be entirely correct - if we allow the shape of the `input` to change then we can't rely on spreading those things here because we can't be sure about the number of keys/values (it could change between renders) and that breaks the "rules of hooks".

So in a sense - we already make an assumption that the input object is always of the same shape, so let's just embrace it more

packages/xstate-react/src/useInterpret.ts
```diff
@@ -85,7 +90,8 @@ export function useInterpret<
       actions,
       activities,
       services,
-      delays
+      delays,
+      input
     };
     const machineWithConfig = machine.withConfig(machineConfig, () => ({
```
Andarist, line 96: The initial render should already contain the "current" input value - we shouldn't have to wait for the first commit phase to get access to it.

Basically this should hold true:
```js
const [state] = useMachine(m, { input: { foo: 'bar' } })
expect(state.input.foo).toBe('bar')
```

packages/core/src/model.types.ts
```diff
@@ -38,9 +40,18 @@ export interface Model<
...
+  ) => StateMachine<
+    TContext,
+    any,
+    TEvent,
+    { value: any; context: TContext },
+    TExtra
+  >;
+  withInput: <TInput extends {}>(
+    input: TInput
+  ) => Model<TContext, TEvent, TAction, TModelCreators, { input: TInput }>;
```
Andarist, line 54: I'm a little bit confused about the `TExtra` and `TInput`. Is `TExtra` an attempt to simplify the list of the generics by moving more stuff into a single generic? If yes - then this one here should be smth like:
```suggestion
  ) => Model<TContext, TEvent, TAction, TModelCreators, TExtra & { input: TInput }>;
```
or maybe even this:
```suggestion
  ) => Model<TContext, TEvent, TAction, TModelCreators, Omit<TExtra, 'input'> & { input: TInput }>;
```

What should happen here?
```js
createModel({}, {}).withInput({ foo: '' }).withInput({ foo: 42 })
```

I guess the last one should win? Or should they be merged?

As long as we don't allow access to a runtime value somewhere within the `createModel` call having the `TInput` to be defined "later" should be OK-ish I guess.

Andarist (commented):
First batch of the review comments

packages/core/src/actionTypes.ts
```diff
@@ -19,3 +19,4 @@ export const error = ActionTypes.ErrorCustom;
 export const update = ActionTypes.Update;
 export const choose = ActionTypes.Choose;
 export const pure = ActionTypes.Pure;
+export const input = ActionTypes.Input;
```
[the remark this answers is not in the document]
mattpocock: Yep, I was unsure about this myself

packages/core/src/interpreter.ts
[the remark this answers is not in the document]
mattpocock: I agree, I think dependencies should always be declared in their entirety. Having the partial made me a touch uncomfortable.

packages/core/src/interpreter.ts
```diff
@@ -541,6 +579,13 @@ export class Interpreter<
...
+    this.send(({
```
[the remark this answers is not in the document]
mattpocock: One thing that might be useful conceptually is imagining both `input` and `output` as first-class parts of the XState API. Imagine we have a `model.withRequiredData()` method to represent a typesafe output from the machine. This could make `data` typesafe.

The benefit of having both `input` and `output` is that we can then use XState machines to represent async functions.

You can, of course, represent input as 'context' - but having it be its own property with its own constraints (must be complete, cannot be assigned to) is IMO important.

packages/xstate-react/src/useInterpret.ts
mattpocock: Yes, makes sense.
```suggestion
  }, [...Object.values(input || {})]);
```

packages/core/src/model.types.ts
mattpocock: I think the last one should win, meaning your second suggestion should be merged in

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
+# Input <Badge text="4.25+" />
+
+[:rocket: Quick Reference](#quick-reference)
+
+When you're building software, it's common to split things up into small pieces. These might be functions in a node process, or components in a React app. These small pieces often need to declare **dependencies** - bits of information/code that they're not responsible for, but that they rely on to serve their purpose.
```
laurakalbag, line 5: > things

What do you mean by “things” here? Is it “code”? “the project structure”? Something else? You give some examples in the next line, could we summarise the “things” with something more specific?

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+  }
+);
+```
+
+Or, by using the `.withConfig` method:
+
+```ts
+const machineWithDifferentInput = machine.withConfig({
+  input: {
+    lyric: 'In the playground, where I spend most of my days'
```
laurakalbag, line 29: Should this use the same lyric example as above, as it’s two different ways to achieve the same goal?

laurakalbag (dismissed):
Love the structure and the basic examples. Importantly, this newbie understood the explanations! I did a pass just for understanding the content, and it reads well, with just a few areas where I think the text could be clearer/more educational with more specific words and terms.

I’d like to do another pass for copyediting after you make these changes (if you agree with them!)

docs/guides/input.md
mattpocock:
```suggestion
When you're building software, it's common to split your code up into small pieces. These might be functions in a node process, or components in a React app. These small pieces often need to declare **dependencies** - bits of information/code that they're not responsible for, but that they rely on to serve their purpose.
```

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+    'xstate.input': {
+      actions: (context, event, meta) => {
+        console.log(meta.state.input);
+      }
+    }
+  }
+});
+```
+
+Note that `meta` will be updated _before_ actions and guards that listen for this transition fire. This means you don't need to read the updated input from the event:
```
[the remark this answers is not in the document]
mattpocock: 😂 

```suggestion
Note that `meta` will be updated _before_ the actions and guards that listen for this transition are executed. This means you don't need to read the updated input from the event:
```

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+
+```ts {5-6,8-9}
+const machine = createMachine({
+  on: {
+    'xstate.input': {
+      actions: (context, event, meta) => {
+        // BAD
```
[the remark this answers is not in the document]
mattpocock: It's bad because then you need to type the event properly. I'm comfortable here with recommending an approach without explaining why.

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+});
+```
+
+## Usage
+
+### Frontend
+
+'Dependencies which change over time', otherwise known as **props**, is a common pattern in modern frontend frameworks. Input allows you to use props in your state machines. See the docs on each individual framework to learn more:
```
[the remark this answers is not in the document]
mattpocock: It's common in frontend, but not in backend. I figured I'd give a short definition here to clarify it.

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+        return Boolean(meta.state.input.error);
+      }
+    }
+  }
+);
+```
+
+## Quick Reference
+
+#### Declare it in the options object
```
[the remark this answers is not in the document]
mattpocock: Yes
```suggestion
#### Declare input in the options object
```

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+  {},
+  {
+    input: {
+      foo: 'bar'
+    }
+  }
+);
+```
+
+#### Declare it using .withConfig
```
mattpocock, line 195:
```suggestion
#### Redeclare input using .withConfig
```

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+
+```ts
+const newMachine = machine.withConfig({
+  input: {
+    foo: 'bar'
+  }
+});
+```
+
+#### Update it in a running service with `.input`
```
mattpocock, line 205:
```suggestion
#### Update input in a running service with `.input`
```

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+    'xstate.input': {
+      actions: (context, event, meta) => {
+        console.log(meta.state.input);
+      }
+    }
+  }
+});
+```
+
+#### Declare its type using `createModel`
```
mattpocock, line 233:
```suggestion
#### Declare the input type using `createModel`
```

.changeset/old-flies-peel.md
```diff
@@ -0,0 +1,124 @@
...
+
+Introduced 'input' - a way of specifying dependencies to the machine which cannot be assigned to. Input can be specified like so:
+
+```ts
+const machine = createMachine(
+  {
+    entry: (context, event, meta) => {
+      console.log(meta.state.input.lyric); // 'In west philadelphia, born and raised'
+    }
+  },
```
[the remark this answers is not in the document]
Andarist: The most recent comment on the matter: https://github.com/statelyai/xstate/discussions/2770#discussioncomment-1650122 . It IMHO shows nicely some of the usability problems with `meta.state` - but there is even more to that, could elaborate more but that would probably require a separate thread or something.

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+```ts
+const machine = createMachine(
+  {
+    entry: (context, event, meta) => {
+      console.log(meta.state.input.lyric); // 'In west philadelphia, born and raised'
+    }
+  },
+  {
+    input: {
+      lyric: 'In west philadelphia, born and raised'
```
Andarist, line 18:
```suggestion
    entry: (context, event, meta) => {
      console.log(meta.state.input.lyric); // 'In west Philadelphia, born and raised'
    }
  },
  {
    input: {
      lyric: 'In west Philadelphia, born and raised'
```

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+
+```ts
+const machineWithDifferentInput = machine.withConfig({
+  input: {
+    lyric: 'In the playground, where I spend most of my days'
+  }
+});
+```
+
+The input is then available in any action/service/guard, in the `meta` property.
```
Andarist, line 34:
```suggestion
The input is then available in any action/service/guard, in the `meta` argument.
```

docs/guides/input.md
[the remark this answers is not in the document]
davidkpiano:
```suggestion
        // ⚠️ Avoid - prefer `meta.state.input` instead
```

docs/guides/input.md
```diff
@@ -0,0 +1,294 @@
...
+
+```ts {5-6,8-9}
+const machine = createMachine({
+  on: {
+    'xstate.input': {
+      actions: (context, event, meta) => {
+        // BAD
+        console.log(event.input);
+
+        // GOOD
```
davidkpiano, line 110:
```suggestion
        // ✅
```

.changeset/old-flies-peel.md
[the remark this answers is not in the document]
mattpocock: I'm happy with the above

docs/guides/input.md
mattpocock: No, because the inference is you can override it.

mattpocock:
@davidkpiano @Andarist The last question we have about this is whether service.input should accept a partial. I don't think it should, and neither does @Andarist.

I've added a commit to this effect that we can revert if we like.

davidkpiano:
> @davidkpiano @Andarist The last question we have about this is whether service.input should accept a partial. I don't think it should, and neither does @Andarist.
> 
> I've added a commit to this effect that we can revert if we like.

That's fine with me. My concern was that devs would have to do this:

```js
service.input({
  ...state.input,
  theOnlyThingYouWantToChange: 42
});

// versus:
// service.input({ theOnlyThingYouWantToChange: 42 });
```

Which has been found to not be a great DX with `machine.withContext({ ...machine.context, a: 'b' })` as reported by devs. It also makes it slightly more computationally expensive to detect granular input changes, since now we have to check _all_ values for equality rather than just the inputted ones, but that might not be that big of a problem admittedly.

I'm fine with not accepting a partial; we can always change it in the future if needed. 👍

docs/guides/input.md
```diff
@@ -0,0 +1,290 @@
+# Input <Badge text="4.25+" />
```
davidkpiano, line 1:
```suggestion
# Input

_Since version 4.27.0_
```

docs/guides/input.md
```diff
@@ -0,0 +1,290 @@
...
+
+[:rocket: Quick Reference](#quick-reference)
+
+When you're building software, it's common to split your code up into small pieces. These might be functions in a node process, or components in a React app. These small pieces often need to declare **dependencies** - bits of information/code that they're not responsible for, but that they rely on to serve their purpose.
```
davidkpiano, line 5:
```suggestion
When you're building software, it's common to split your code up into small pieces. These might be functions in a node process, or framework components. These small pieces often need to declare **dependencies** - bits of information/code that they're not responsible for, but that they rely on to serve their purpose.
```

docs/guides/input.md
```diff
@@ -0,0 +1,290 @@
...
+  },
+  {
+    input: {
+      foo: 'bar'
+    }
+  }
+);
+```
+
+Unlike [context](./context.md), inputs cannot be assigned to from _inside_ the machine. If you do need to mutate the input locally, consider putting it in context instead.
```
davidkpiano, line 64:
```suggestion
Unlike [context](./context.md), inputs cannot be assigned to from _inside_ the machine. If you do need to mutate the input locally, consider putting it in `context` instead.
```

docs/recipes/react.md
```diff
@@ -209,32 +209,29 @@ This also works for services, guards, and delays.
 
 > If you use this technique, any references you use inside `goToOtherPage` will be kept up to date each render. That means you don't need to worry about stale references.
 
-### Syncing data with useEffect
+### Syncing data with input <Badge text="4.25+" />
```
davidkpiano, line 212:
```suggestion
### Syncing data with input

_Since version 4.27.0_
```

packages/xstate-react/src/useInterpret.ts
```diff
@@ -119,6 +125,12 @@ export function useInterpret<
     };
   }, []);
 
+  useIsomorphicLayoutEffect(() => {
+    if (input) {
+      service.input?.(input);
+    }
+  }, [...Object.values(input || {})]);
```
davidkpiano, line 132: Clever!

mattpocock:
@davidkpiano I think that since the only use case we can think of for `service.input()` specifically is prop updates, I think it's fair to assume providing the whole input is fine, similar to props.

packages/xstate-react/src/useInterpret.ts
mattpocock: One thing I'm nervous about with this (and the implementation in general) is that passing an un-memoised input to `useInterpret` will not end well:

```ts
input: {
  data: { id: '1', foo: false }
}
```

I _think_ this will end up in an infinite loop, and I'm unsure how to warn for this.

docs/guides/input.md
```diff
@@ -0,0 +1,292 @@
...
+  {
+    entry: [
+      (context, event, meta) => {
+        console.log(meta.state.input.foo);
+      }
+    ],
+    invoke: {
+      src: (context, event, meta) => {
+        console.log(meta.state.input.foo);
+      },
```
Andarist, line 49: Since `meta.state` is currently not available to invoke creators this should be removed from an example and from other places mentioning this.

docs/guides/input.md
```diff
@@ -0,0 +1,292 @@
...
+        console.log(meta.state.input);
+      }
+    }
+  }
+});
+```
+
+Note that `meta` will be updated _before_ the actions and guards that listen for this transition are executed. This means you don't need to read the updated input from the event:
+
+```ts {5-6,8-9}
```
Andarist, line 100: what do those curly braces mean here? it's the first time I see this and I can't find any reference for this in https://github.github.com/gfm/#fenced-code-blocks

docs/guides/input.md
mattpocock: Isn't it? Ah, crap

docs/guides/input.md
mattpocock: They're for highlighting lines of code, used elsewhere in the repo

packages/core/src/interpreter.ts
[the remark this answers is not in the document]
Andarist: I agree that input/output, as a concept, might be very handy and might increase the type safety of things. 

It would be great to achieve the parity - have the `input` be available on `xstate.init` event and `output` on the `done.invoke.*`.

> You can, of course, represent input as 'context' - but having it be its own property with its own constraints (must be complete, cannot be assigned to) is IMO important.

I'm not sold on this argument though - IMHO those constraints are not that useful because I don't expect people to think in those terms, I don't expect them to choose one (context vs input) based on those. The completeness can be solved at the type level - by typing event payloads and by grouping the keys in context on a property that has no optional members.

Based on that I really think that event+context should always be encouraged over input. I see how the props thing is similar to, observable stream and one could do something like:
```js
obs.subscribe(data => {
  service.input(data)
})
```
but I'm not sure how useful in practice that is. In this context, we can also deal with multiple **implicit** "sources" so the Partial argument swings back to allowing a Partial (I'm still more inclined more to ship without it though to see what happens as it could be very interesting to read community discussions about this).

I just wonder if making **implicit** updates is not a feature that we should stick with and only open up for them when we have strong-enough use cases for them.

docs/guides/input.md
davidkpiano: Yep, they're for Vuepress

packages/core/src/model.ts
Andarist: actually - the order should be reversed here, `implementations` can already contain `input` and that shouldn't take precedence over the argument received by `withInput`:
```suggestion
          return model.createMachine(config, { ...implementations, input });
```

packages/core/src/model.types.ts
```diff
@@ -38,9 +40,24 @@ export interface Model<
...
+    config: MachineConfig<TContext, any, TEvent, TExtra, TAction>,
+    implementations?: Partial<MachineOptions<TContext, TEvent, TExtra, TAction>>
+  ) => StateMachine<
+    TContext,
+    any,
+    TEvent,
+    { value: any; context: TContext },
+    TExtra
+  >;
+  withInput: <TInput extends {}>(
```
Andarist, line 52: Most likely you've wanted a different constraint here:
```suggestion
  withInput: <TInput extends Record<string, unknown>>(
```

Unfortunately, `{}` when used as a constraint is not really constraining the generic to be a subtype of an empty object. It's some kind of "allow all" (well, probably not exactly that but this always confuses me and I've stopped trying to understand it). Check out [the demo on TS playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFApgZygHgCqJQD1TABM1EBvAXwD4AKASgC5FdyAoRRAJxShC6RgQAG2GIAhmXFgAngG42lNm1QZMULiBR16C1VgBEBnXvRYAjACYAzCZVnMVO-seJgcOMwMAjcVwOINAwKbKCQsAjIZpY4eIQoJGQASigQcFzEmBhcMGAA5gA0iOAA1mBwAO5g1DrMrBzcvPyCImKSErIKSvYYMRpaztGYRoO9mFa2wT1QMU5T+rNuHl6+-oF2QA)

docs/guides/input.md
Andarist: That is not hard to fix, but I... don't want to 😅 this is pending on the outcome of this: https://github.com/statelyai/xstate/pull/2779#discussion_r749321830 . I think we should decide on this asap - as it would be bad to introduce a new thing into a potentially deprecated "argument". If we can spare people from refactoring this in the future this would be awesome.

packages/xstate-react/src/useInterpret.ts
Andarist: Well, it's React world - people should already be ~tired with~ accustomed to memoization. I'm not sure there is much that we can do about this within `@xstate/react`

However, I see a **huge** problem that goes into the performance area of the design here. 

I think this expectation at the end is reasonable:
```js
  it('should not re-render after input changes', () => {
    let renders = 0;
    const model = createModel({}, {}).withInput(
      {} as {
        foo: string;
      }
    );

    const machine = model.createMachine({}, {});

    const App = (props: { foo: string }) => {
      renders++;
      useMachine(machine, {
        input: {
          foo: props.foo
        }
      });
      return null;
    };

    const { rerender } = render(<App foo="1" />);
    rerender(<App foo="2" />);
    
    expect(renders).toBe(2);
  });
```

An input change should not result in a "followup" rerender - and this is now "required" because the `state` changes. The main goal of the feature is to make `input` available **within** the machine, right?

To avoid this rerender problem we'd have to remove `input` from the `state` entirely - because if we keep it there then it should be kept in sync and we can't allow for stale data to be "preserved" there until some next, random, update.

Alternatively, we could bail out from rerendering on input changes and "fake" the `state.input` update in the render phase. This goes into dangerous territory though and opens new challenges related to proper memoization, user expectations and more

packages/core/src/StateNode.ts
```diff
@@ -466,14 +485,25 @@ class StateNode<
...
+      input
+    } = this.options;
 
     return new StateNode(
       this.config,
       {
+        input: {
+          ...input,
+          ...options.input
+        },
```
Andarist, line 506: I think I've mentioned this somewhere else too - it seems inconsistent how we approach input in this PR. In some cases, it's being overridden and in some cases, it's being merged. I know this is not unique to this PR though - similar things are inconsistent in more places of the codebase. However, I think it would still be nice to decide what's best for input and other things right now - so we don't have to change the input's behavior in the future.

Andarist:
I'm pretty confident that we shouldn't forward this event even if `autoForward` is being set to true. It's too implicit and could wreak havoc in some setups.

packages/core/src/model.ts
```diff
@@ -26,7 +26,8 @@ export function createModel<
   >,
   TComputedAction = UnionFromCreatorsReturnTypes<
     Prop<TFinalModelCreators, 'actions'>
-  >
+  >,
+  TComputedInput = Prop<TFinalModelCreators, 'input'>
```
Andarist, line 30: This is unused and probably should stay that way as `withInput` already is responsible for accepting this
```suggestion
  >
```

packages/core/src/model.types.ts
```diff
@@ -1,3 +1,4 @@
+import { ExtraGenerics } from '.';
```
Andarist, line 1: could you sum up why you have chosen this approach and how do you imagine this evolving in the future?

packages/core/src/types.ts
```diff
@@ -900,7 +1043,12 @@ export enum ActionTypes {
   ErrorCustom = 'xstate.error',
   Update = 'xstate.update',
   Pure = 'xstate.pure',
-  Choose = 'xstate.choose'
+  Choose = 'xstate.choose',
+  Input = 'xstate.input'
```
Andarist, line 1047: as `EventTypes` has been created in this PR this is being redundant
```suggestion
  Choose = 'xstate.choose'
```

packages/core/test/input.test.ts
```diff
@@ -0,0 +1,264 @@
...
+        any,
+        {
+          input: {
+            foo: string;
+          };
+        }
+      >({
+        on: {
+          'xstate.input': {
+            actions: [
```
Andarist, line 195: could you verify that this fails?
```js
assign((_, __, meta) => {
  expect(meta.state.input).toEqual({ foo: 'bar' });
  return {}
})
```
and this:
```js
cond: (_, __, meta) => {
  expect(meta.state.input).toEqual({ foo: 'bar' });
  return true
}
```

I strongly suspect that both will fail and is just yet another argument in my opinion to deprecate `meta.state` 😜

packages/core/test/input.test.ts
```diff
@@ -0,0 +1,264 @@
...
+      foo: 'changed'
+    });
+
+    expect(service.state.history?.input).toEqual({
+      foo: 'foo'
+    });
+  });
+});
+
+describe('createModel', () => {
```
Andarist, line 246: could you add tests that would validate the runtime behavior of `createModel({}).withInput`?

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -987,3 +988,113 @@ describe('useMachine (strict mode)', () => {
...
+
+    const machine = model.createMachine({}, {});
+
+    const App = (props: { foo: string }) => {
+      const [state] = useMachine(machine, {
+        input: {
+          foo: props.foo
+        }
+      });
+      return <>{state.input.foo}</>;
```
Andarist, line 1031: a question: is it expected for the user that sometimes this isn't true: `props.foo === state.input.foo`?

Andarist:
> I'd even be in favor of deprecating autoForward in favor of something more explicit in v5.

Definitely, let's remove or redefine this. How it is defined in SCXML is IMHO harmful - this can't forward **all** the events (including "internal" ones).

.changeset/old-flies-peel.md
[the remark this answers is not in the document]
mattpocock: Ping for @davidkpiano, we should resolve this in a call.

packages/xstate-react/src/useInterpret.ts
mattpocock: You're right, this is tricky. It depends if the user ever wants to _read_ `state.input`. I'm not sure what the point of that would be, since by definition you always have the values in scope anyway.

That would mean the input shouldn't be on `meta.state` at all, but `meta.input`.

And the input should live on the `Interpreter`, not the `State`?

Andarist: The only thing why one would like to read `state.input` is convenience - you could pass the whole `state` around instead of 2 separate values if you need access to both of those things. We also should think about how this affects selectors - maybe those should receive "meta" too?

> And the input should live on the Interpreter, not the State?

That's my initial impression as well - maybe some other object should be created within interpreter that would be used for use cases like this. This could live in smth like `actorCtx` that we might need for other stuff too - like the actual `ref` for the "self", maybe also children? 

We probably should allow for this stuff to be serialized for serialization & rehydration purposes - to achieve this goal/use case we don't have to limit ourselves to keeping everything in a single object. Sure, it makes some things easier - but some other things harder. 

This is a very broad discussion though - and I think we could really use a design meeting for this.

acollazomayer:
@mattpocock Any news about this. Do you need some help?

mattpocock:
@acollazomayer We're debating internally whether we want to do this. @Andarist is focusing more on v5 release right now, so we're soft feature-freezing v4 while that gets worked on.

Andarist:
iIRC, the main problem I see with this implementation is that by using `input` we are basically opting into rerendering after each input change. I'm not quite yet sure how to best approach this since there isn't an easy way in React's world to overcome this problem given how our integration with React works.

Perhaps if we'd constrain the design somehow then we could overcome this problem somehow. For instance, if we don't allow this to be used with `useMachine` and only allow usage with `useInterpret` then those rerenders can be mitigated because in such a situation people would have to rely on `useSelector` to *read* the machine and on that level changing `input` wouldn't have to end up in additional rerenders.

However, this also kinda touches on how to best organize this kind of stuff in the React's world. Cause... deferring this kind of stuff till the commit phase/effects is fundamentally not ideal in this context. Ideally, some side-effects like refetching data would be already triggered in the render phase - we "only" need to ensure this stuff is idempotent etc so this becomes a caching problem etc.

I think it would be great to revisit this in the upcoming weeks - before switching full gears on the v5 work (I'm still working on some v4 stuff right now). The outcome of this discussion could influence some direction for the v5 etc.

And, of course, we'd need to figure out how to best integrate this with typegen - but I guess this is just a matter of including this event type explicitly in the `TEvent` when declaring the machine.

acollazomayer:
@mattpocock and @Andarist Thanks for all the input (pun intended). I understand perfectly that there might be other priorities. If you need any help don't hesitate to ask. 

In my opinion, I don't see that much of a problem with re-rendering after each input change. I don't see a way around it. I think that it would be up to the implementation to solve these issues. In our current implementation, we are "hacking" our way to have input support using useEffect and events. In order to save re-renders we use useInterpret and selectors every time we need a value from the state machine. This approach works and looks kind of clean but I also understand that it might look counterintuitive

Andarist:
One of the things I'm worried about is that syncing stuff from within effects is not very React-y. So I would just like to research this more and rethink if we have any alternatives. While we don't have to necessarily do stuff in React way, it would be ideal if we could - so it's worth giving this another shot to see if we can somehow fit in their model with our stuff or if perhaps both models are so incompatible that we just can't do that.

It would also be best to gather the most common use cases that would be enabled by this feature as that would drive us to figure out the best design.

Andarist:
Input was introduced in v5 beta as something different: https://github.com/statelyai/xstate/pull/3815

PR #2780, opened 2021-11-01
Revert "`activities` should be `optional`"

Andarist:
Reverts statelyai/xstate#2751

I'm unsure if this change has been sound. We can, of course, make this optional at the type level but in reality this property is **always** defined on the `StateNodeDefinition` object (that is returned from the `stateNode.definition` getter):
https://github.com/statelyai/xstate/blob/c1503b1219d995ebf0f45de46036c5a1d7e6442f/packages/core/src/StateNode.ts#L522

A side note is that the value fallback (`|| []`) is actually redundant because `this.activities` is always an array:
https://github.com/statelyai/xstate/blob/c1503b1219d995ebf0f45de46036c5a1d7e6442f/packages/core/src/StateNode.ts#L442-L444

cc @huan

### DIFF
--- a/.changeset/optional-activities.md
+++ b/.changeset/optional-activities.md
@@ -1,5 +1,0 @@
----
-"xstate": patch
----
-
-`activities` should be `optional` because it has been deprecated.
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -680,7 +680,7 @@ export interface StateNodeDefinition<
   /**
    * @deprecated
    */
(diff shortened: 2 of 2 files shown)


huan:
I'm ok with the reversion.

I sent this PR because when I'm studying the XState I found that the `activities` have to be set even I do not need it at all. After reading docs I saw it was documented as `deprecated` then I think it should be optional.

Thanks for the careful reviewing BTW!

davidkpiano (approved):

mattpocock (approved):

PR #2783, opened 2021-11-02
Added createActorContext

mattpocock:
https://www.loom.com/share/efa306bd87384acf952a1fbb27ec5d27

- [ ] Write docs
- [ ] Write changeset
- [ ] Write tests

### DIFF
--- a/packages/xstate-react/src/createXStateContext.tsx
+++ b/packages/xstate-react/src/createXStateContext.tsx
@@ -1,18 +1,22 @@
-import { createContext, useContext } from 'react';
-import { InterpreterFrom, StateFrom, StateMachine } from 'xstate';
+import React, { createContext, useContext } from 'react';
+import {
+  InterpreterFrom,
+  MachineOptions,
+  StateFrom,
+  StateMachine
+} from 'xstate';
 import { useActor } from './useActor';
+import { useInterpret } from './useInterpret';
 import { useSelector } from './useSelector';
 
-export const createXStateContext = <
+export function createXStateContext<
   TMachine extends StateMachine<any, any, any, any>
->(
-  _machine: TMachine
-) => {
+>(machine: TMachine) {
   const context = createContext({} as InterpreterFrom<TMachine>);
 
-  const createSelector = <T>(selector: (state: StateFrom<TMachine>) => T) => {
+  function createSelector<T>(selector: (state: StateFrom<TMachine>) => T) {
     return selector;
-  };
+  }
 
   const useXStateContext = (): InterpreterFrom<TMachine> => {
     const contextValue = useContext(context);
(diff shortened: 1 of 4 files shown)


packages/xstate-react/src/createXStateContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+      children?: React.ReactNode;
+    }) => {
+      const service = useInterpret(machine, props.options) as any;
+
+      return (
+        <context.Provider value={service}>{props.children}</context.Provider>
+      );
+    },
+    Provider: context.Provider,
+    Consumer: context.Consumer,
```
Andarist, line 64: I almost don't remember when I've used Consumer for the last time 😅 maybe we can drop it and only expose the `useContext`?

packages/xstate-react/src/createXStateContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+export function createXStateContext<
+  TMachine extends StateMachine<any, any, any, any>
+>(machine: TMachine) {
+  const context = createContext({} as InterpreterFrom<TMachine>);
+
+  function createSelector<T>(selector: (state: StateFrom<TMachine>) => T) {
+    return selector;
+  }
+
+  const useXStateContext = (): InterpreterFrom<TMachine> => {
```
Andarist, line 21: q: when this would be useful publicly if we have `useActor` and `useSelector` helpers here? for what kind of a task they are not enough?

packages/xstate-react/src/createXStateContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+  StateMachine
+} from 'xstate';
+import { useActor } from './useActor';
+import { useInterpret } from './useInterpret';
+import { useSelector } from './useSelector';
+
+export function createXStateContext<
+  TMachine extends StateMachine<any, any, any, any>
+>(machine: TMachine) {
+  const context = createContext({} as InterpreterFrom<TMachine>);
```
Andarist, line 15: it would be nice if we'd accept a `displayName` parameter and we'd set it on this created context - it will display better in the React devtools etc

packages/xstate-react/src/createXStateContext.tsx
mattpocock: For when you want to send a running service an event, but not subscribe to updates to it or select any data from it.

packages/xstate-react/src/createXStateContext.tsx
mattpocock: Yes, agree - do you think it should be required or optional?

packages/xstate-react/src/createXStateContext.tsx
mattpocock: Pretty cheap to expose it? Don't mind either way

packages/xstate-react/src/createXStateContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+        : never;
+      children?: React.ReactNode;
+    }) => {
+      const service = useInterpret(machine, props.options) as any;
+
+      return (
+        <context.Provider value={service}>{props.children}</context.Provider>
+      );
+    },
+    Provider: context.Provider,
```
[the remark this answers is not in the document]
mattpocock: I'm unsure whether it is - sometimes you want to do this:

```ts

const context = createXStateContext(machine);

  const service = useInterpret(machine);

  useEffect(() => {
    service.send({ type: 'ID', id: props.id });
  }, [props.id]);

  return <context.Provider value={service}></context.Provider>;
};
```

Maybe we should rename it `reactContext` and provide the raw context we created? That would also solve the issue of whether we provide the Consumer or not.

packages/xstate-react/src/createXStateContext.tsx
davidkpiano: > Yes, agree - do you think it should be required or optional?

Perhaps optional, and use the machine ID for a default

mattpocock: 👍

davidkpiano:
Like discussed, `createXStateContext` -> `createActorContext`

packages/xstate-react/src/createXStateContext.tsx
davidkpiano: Let's expose it; useful for those using class components (yes, many devs/companies still are)

Andarist:
Spiritually the ideas from this PR were incorporated into https://github.com/statelyai/xstate/pull/3778

PR #2786, opened 2021-11-04
Removed installation exports

mattpocock:


laurakalbag (commented):
Just to satisfy my curiosity: why are you removing these? Are they no longer used?

mattpocock:
It just feels very weird to have them here! Never seen this on any other docs page.

laurakalbag:
Fair enough!

laurakalbag (approved):

PR #2792, opened 2021-11-08
Rewrote the docs landing page

mattpocock:
https://www.loom.com/share/9800b3f03fbb4bf39a18d57e101deafe


mattpocock:
@laurakalbag Also added this cute visualisation:

image

laurakalbag (commented):
Britishism… we’re all speaking English here 😂

On `XState Visualizer`/`Stately Visualizer`, I still get a bit confused about this. While it is visualis/zing XState, it is the Visualizer belonging to Stately… so maybe we need to clarify this with @davidkpiano.

docs/README.md
```diff
@@ -3,7 +3,7 @@
   <br />
   <img src="https://user-images.githubusercontent.com/1093738/101672561-06aa7480-3a24-11eb-89d1-787fa7112138.png" alt="XState" width="150"/>
   <br />
-    <sub><strong>State machines and statecharts for the modern web.</strong></sub>
+    <sub><strong>Handle complexity anywhere JavaScript runs</strong></sub>
```
laurakalbag, line 6: Ok, so I agree with your point about “for the modern web” being limited, but here we miss the specificity of state machines and statecharts, and “handle complexity” is perhaps too unspecific for a docs landing page. (More of an overall landing page vibe.)

How about using your line as below: “Handle complexity anywhere JavaScript runs using statecharts.” (Just revised as I prefer the wording you used.)

Then we’re:
- *handle complexity*: WHY - benefit of statecharts/XState
- *with statecharts*: HOW - what we’re using to handle the complexity, just statecharts because its technically redundant to say “state machines and statecharts”, and XState is always statecharts and usually more than state machines
- *anywhere JavaScript runs*: WHERE - where you can handle complexity

This isn’t a perfect solution:

- Do we need a WHAT? Specify that the complexity is in applications and/or state specifically?
- Does the length make it less of a tagline and more of just a sentence?

I’m only getting into mega detail on this because it’s the first line on the docs homepage 😊

docs/README.md
```diff
@@ -12,14 +12,23 @@
 [![npm version](https://badge.fury.io/js/xstate.svg)](https://badge.fury.io/js/xstate)
 <img src="https://opencollective.com/xstate/tiers/backer/badge.svg?label=sponsors&color=brightgreen" />
 
-JavaScript and TypeScript [finite state machines](https://en.wikipedia.org/wiki/Finite-state_machine) and [statecharts](https://www.sciencedirect.com/science/article/pii/0167642387900359/pdf) for the modern web.
+Handle complexity anywhere JavaScript runs using statecharts.
```
laurakalbag, line 15: Could this be “Handle complex states anywhere JavaScript runs using statecharts.”? Just to be a bit more specific about the complexity, and as an intro for the following list?

laurakalbag (commented):
Should we bring back another real-life example of a state machine? I’ve got some reservations about replacing the tagline, though I agree with the critique of the current tagline, and otherwise think this is a big improvement.

docs/README.md
```diff
@@ -3,7 +3,7 @@
   <br />
   <img src="https://user-images.githubusercontent.com/1093738/101672561-06aa7480-3a24-11eb-89d1-787fa7112138.png" alt="XState" width="150"/>
   <br />
-    <sub><strong>State machines and statecharts for the modern web.</strong></sub>
```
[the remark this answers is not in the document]
davidkpiano: Yep, these are for the GitHub styles - unfortunately we can't have a stylesheet in here for the GItHub readme.

docs/README.md
mattpocock:
```suggestion
    <sub><strong>Build with statecharts, anywhere JavaScript runs</strong></sub>
```

⬆️ ❓

docs/README.md
```diff
@@ -12,14 +12,23 @@
 [![npm version](https://badge.fury.io/js/xstate.svg)](https://badge.fury.io/js/xstate)
 <img src="https://opencollective.com/xstate/tiers/backer/badge.svg?label=sponsors&color=brightgreen" />
 
-JavaScript and TypeScript [finite state machines](https://en.wikipedia.org/wiki/Finite-state_machine) and [statecharts](https://www.sciencedirect.com/science/article/pii/0167642387900359/pdf) for the modern web.
+Handle complexity anywhere JavaScript runs using statecharts.
 
-New to state machines and statecharts? [Read our introduction](/guides/introduction-to-state-machines-and-statecharts/).
+- Create **robust applications**, both on the **frontend and backend**
+- **Visualize your app's logic** in our awesome **dev tools**
+- Integrates with Node, [React](./packages/xstate-react/index.md), [Vue](./packages/xstate-vue/index.md), [Svelte](./packages/xstate-svelte/index.md) - anywhere JavaScript runs
```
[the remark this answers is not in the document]
mattpocock: Yyyy... Unsure, we probably need a page for how to implement XState in the backend. But since there's no strong recommendations/integration package, it's a little dicey.

docs/README.md
davidkpiano: Agree that the tagline should be more descriptive of what XState actually is

docs/README.md
davidkpiano: I would rephrase "anywhere JavaScript runs" to "in JavaScript" as that reads a bit more naturally

docs/README.md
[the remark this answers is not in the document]
laurakalbag: Ok, I’ll have a look into what we can do that gives us slightly better semantics for assistive tech and seo.

docs/README.md
mattpocock: Yep I'm down for this, anything that isn't 'modern web' is fine by me and we can workshop a proper tagline later

PR #2796, opened 2021-11-08
Condense introductions in each section to redirect to the introduction to statecharts

mattpocock:
https://www.loom.com/share/12f6b90aa48c453c91d1213adff63553


laurakalbag (changes requested):
I think the restructuring works overall.

This is another review with a lot of copyediting that may be more for @davidkpiano (as the original author) than @mattpocock.

mattpocock:
@laurakalbag I think a lot of the comments relate to text I haven't changed, and because of the code folding issue I'm not always sure what you're referring to - especially because line numbers change as edits are made. Any chance we could move these comments to their own issue, and correct them in a separate PR?

laurakalbag:
> @laurakalbag I think a lot of the comments relate to text I haven't changed, and because of the code folding issue I'm not always sure what you're referring to - especially because line numbers change as edits are made. Any chance we could move these comments to their own issue, and correct them in a separate PR?

Good shout. I’ll find a better way to do the editing for older content too.

PR #2798, opened 2021-11-08
Added a toPromise function to allow modelling machines as promises

mattpocock:
https://www.loom.com/share/01794774bfbc43c685e26282c47a74ed

### DIFF
--- a/packages/core/test/toPromise.test.ts
+++ b/packages/core/test/toPromise.test.ts
@@ -1,0 +1,287 @@
+import { createMachine } from '../src/Machine';
+import { toPromise, validatePromiseMachine } from '../src/toPromise';
+
+describe('toPromise', () => {
+  describe('validatePromiseMachine', () => {
+    it('Should not allow states without invocations or final', () => {
+      const errors = validatePromiseMachine(
+        createMachine({
+          id: 'machine',
+          initial: 'pending',
+          states: {
+            pending: {}
+          }
+        })
+      );
+
+      expect(errors).toEqual([
+        {
+          stateId: 'machine.pending',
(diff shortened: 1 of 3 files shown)


packages/core/test/toPromise.test.ts
```diff
@@ -0,0 +1,287 @@
...
+              type: 'final'
+            },
+            final: {
+              type: 'final'
+            }
+          }
+        })
+      );
+
+      expect(state.value).toEqual('final');
```
davidkpiano, line 283: Separate discussion/PR: in the future, this would be a good opportunity to represent "done data" in the `state` itself (`undefined` by default). That way, we can use promise machines like this:

```js
const { doneData } = await toPromise(machine);

return doneData;
```

mattpocock: Agree!

Andarist:
[`waitFor`](https://github.com/statelyai/xstate/pull/3190) can be used as an approximation of this. It's not exactly the same since it doesn't come with any constraints for the given machine config though.

davidkpiano:
> [`waitFor`](https://github.com/statelyai/xstate/pull/3190) can be used as an approximation of this. It's not exactly the same since it doesn't come with any constraints for the given machine config though.

I assume #3977 supersedes this too.

Andarist:
Not entirely but it's definitely related. This PR here was also about restricting the shape of the machine config - to give some guarantees around the completion of the given machine etc. We currently don't do any of that, the config stays "free-form" and `waitFor`/`toPromise` are just helpers to await the execution of any given actor. They don't impose any constraints on the completion, suspension points, etc

PR #2800, opened 2021-11-10
Add `sendTo`

davidkpiano:
This PR adds the `sendTo(actor, event)` action creator, which will replace the external use-cases for `send(...)` (the internal use-cases will be replaced with `raise(...)`).


packages/core/test/actions.test.ts
```diff
@@ -1446,6 +1447,67 @@ describe('sendParent', () => {
...
+      child: ActorRefFrom<typeof childMachine>;
+      count: number;
+    }>({
+      context: () => ({
+        child: spawn(childMachine),
+        count: 42
+      }),
+      entry: sendTo(
+        (ctx) => ctx.child,
+        // TODO: figure out how to avoid `as const`
```
Andarist, line 1502: A simplified repro - [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggxsA9gJwEoQGYB4AqA+KAXigG8AoKSqAZwgDsATALigAoIA3FnASiII6IAlgzIBfMmVCQoAUQ71gAMWSIAtlhgFiMKBAAeweg2qwEKdNiF0MEZHIIB+OVBYBDOiEkMIcADZuyNAYAK50CEKIdDTGOIiaeobGpvBIaJhYHiB4eKxu5ukYLDAANFAA5hDA8oosrHyEBDV0yqoaWjwsgiKS0tDyRKRQfSwA5EoA8hOjANxQHtQA7nYs1MDI1uVQYlAAPkMjUKMAQjCos1BwiGHALHQhagBGdtvevgFBl1Fr83DFBZYsPI8JJaIw4nk4GV6vw2CRhuAIGNJtNtjweJIgA)

I'm not sure what to do about it, I've reached out to Devan. It's wild that autocomplete works at this position but yet the type is inferred as `string` 🤦

Andarist: @devanshj is a wizard. Based on his suggestions and some trial & error in our codebase and in the TS playground I got this: https://github.com/statelyai/xstate/pull/2847

packages/core/src/actions.ts
```diff
@@ -271,6 +272,39 @@ export function sendParent<
   });
 }
 
+type InferEvent<E extends EventObject> = {
+  [T in E['type']]: { type: T } & Extract<E, { type: T }>;
+}[E['type']];
```
Andarist, line 277: I am somewhat worried about the performance implications of this, but don't know how to make the whole thing work without it so 🤷‍♂️ I also might worry too much - but I definitely experience performance problems related to types when working with XState. It's hard to profile this though so... 🤷‍♂️

Andarist (approved):

PR #2802, opened 2021-11-10
Remove `xstate-viz` package (old)

davidkpiano:
This PR removes the `/packages/xstate-viz` package.

The actual XState Viz is located at github.com/statelyai/xstate-viz


package.json
```diff
@@ -4,8 +4,7 @@
   "description": "Finite State Machines and Statecharts for the Modern Web.",
   "workspaces": {
     "packages": [
-      "packages/*",
-      "packages/xstate-viz/example"
+      "packages/*"
```
Andarist, line 7: We should also remove this from here:
https://github.com/statelyai/xstate/blob/6245be838581e2cf84ac580303262e46b6cebd56/.changeset/config.json#L14-L15

PR #2804, opened 2021-11-11
Optimize `state.can()`

### LINKED ISSUE #2803: can() will trigger assign()
**Description**
can() will trigger assign()

**Expected Result**
do not trigger

**Actual Result**
can() will trigger assign()

**Reproduction**
```
...
       test: {
            on: {
              TEST: {
                target: 'a',
                actions: 'testAction'
              },
            },
          }
...
actions: {
    testAction: assign({
      behavior: (_context, event) => {
        console.log('xxxxxxxxxxxxxxxxxxxxxxxx');
    })
}
```

```
xState.can('TEST')
```

davidkpiano:
This PR optimizes `state.can()` by reading the transition data before evaluating the actions (such as `assign(...)`), which do not need to be evaluated in order to determine if the state will change.

Fixes #2803


davidkpiano:
@Andarist Is this good to go?

packages/core/test/state.test.ts
```diff
@@ -825,13 +825,13 @@ describe('State', () => {
       expect(state.can({ type: 'ANY_EVENT' })).toEqual(false);
     });
 
-    it('should allow errors to propagate', () => {
+    it('should not execute assignments', () => {
```
Andarist, line 828: based on this test we can't truly know if it tests what the test title suggests - the error could be just swallowed somewhere within our implementation and the test would pass. Could we refactor this to use a variable that changes value within `assign` and assert that it has not been changed?

davidkpiano: Done

PR #2812, opened 2021-11-15
[v5] Improve memoization

davidkpiano:
This PR introduces internal `memo(...)` and `getMemo(...)` functions that obviate the `__cache` stuff in a more GC-friendly way.

No changeset; internal only.

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -24,7 +24,6 @@ import type {
   InvokeDefinition,
   Mapper,
   PropertyMapper,
-  NullEvent,
   SCXML,
   TransitionDefinitionMap,
   InitialTransitionDefinition,
@@ -42,6 +41,7 @@ import {
 } from './stateUtils';
 import { evaluateGuard } from './guards';
 import type { StateMachine } from './StateMachine';
+import { memo } from './memo';
 
 const EMPTY_OBJECT = {};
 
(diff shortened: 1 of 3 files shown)


packages/core/src/StateNode.ts
```diff
@@ -222,6 +196,14 @@ export class StateNode<
     this.tags = toArray(config.tags);
   }
 
+  public initialize() {
```
Andarist, line 199: this should be private and if it can't be easily made private cause of typing reasons then we should at least prefix this with an underscore

A separate note is that it would be great if we could remove this one in the future. The "resolved" shape of a transition is only needed first when we try to select it - no reason to resolve **all** of the transitions upfront. For huge machines the cost here can add up and it would be great if we could minimize it.

packages/core/src/memo.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function getMemo<T>(
+  stateNode: StateNode<any, any>,
+  key: string
+): T | undefined {
+  return stateNodeMap.get(stateNode)?.[key];
+}
+
+export function memo<T>(
+  stateNode: StateNode<any, any>,
```
Andarist, line 13: this is quite a generic module - we could decouple its implementation from state nodes, the "owner" of the memo cache can be just anything (well, anything that is accepted by a `WeakMap`), it's not limited to state nodes

packages/core/src/memo.ts
```diff
@@ -0,0 +1,29 @@
+import type { StateNode } from '.';
+
+const stateNodeMap = new WeakMap<StateNode, any>();
+
+export function getMemo<T>(
+  stateNode: StateNode<any, any>,
+  key: string
+): T | undefined {
+  return stateNodeMap.get(stateNode)?.[key];
+}
```
Andarist, line 10: this one stays unused, right? it's simple enough that we can always add it when the need arise - so I would recommend dropping it for the time being if it truly stays unused

packages/core/src/memo.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function memo<T>(
+  stateNode: StateNode<any, any>,
+  key: string,
+  fn: () => T
+): T {
+  let memoizedData = stateNodeMap.get(stateNode);
+
+  if (!memoizedData || !memoizedData[key]) {
```
Andarist, line 19: if the intention is to always cache the first run of `fn` then the more appropriate check here would be to check if the key exists in the `memoizedData`. Falsy values should, technically, be cacheable too - even if we don't quite intend to cache such

packages/core/src/memo.ts
```diff
@@ -0,0 +1,29 @@
...
+    memoizedData = {
+      ...memoizedData,
+      [key]: fn()
+    };
```
Andarist, line 23: there is also no reason to use immutability here - let's save some perf here by avoiding spread

Andarist:
> This PR introduces internal memo(...) and getMemo(...) functions that obviate the __cache stuff in a more GC-friendly way.

Note that this has almost no effect on GC. The previous implementation was just as good, memory-wise - it just involved more manual cache management and this got now abstracted in the form of this `memo` helper

packages/core/src/memo.ts
davidkpiano: You're right; it was used during the refactor and then I removed it's only usage 😜

davidkpiano: Removed

Andarist (approved):

PR #2813, opened 2021-11-15
Duplicate state resolution

aerhard:
This PR removes redundant `resolveState()` invocations from the interpreter tests and the "Persisting State" section of the docs. See https://github.com/statelyai/xstate/discussions/2811.

### DIFF
--- a/docs/guides/states.md
+++ b/docs/guides/states.md
@@ -280,7 +280,7 @@ try {
 }
 ```
 
-State can be restored using the static `State.create(...)` method and resolved using the public `machine.resolveState(...)` method:
+State can be restored using the static `State.create(...)` method:
 
 ```js
 import { State, interpret } from 'xstate';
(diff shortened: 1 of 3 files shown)


mattpocock:
@aerhard No need for a changeset on this one - changesets are only for updates requiring an npm release, which the docs don't.

davidkpiano (approved):

docs/guides/states.md
```diff
@@ -280,7 +280,7 @@ try {
 }
 ```
 
-State can be restored using the static `State.create(...)` method and resolved using the public `machine.resolveState(...)` method:
```
Andarist, line 283: Resolving can actually be important in some situations though. For instance - `state.can` won't work without it (for that one state, any state that will replace this "rehydrate" state will already work with `state.can`).

davidkpiano: Right but this is only for the persist/restore use-case which doesn't need `resolveState`. We can mention this in the `state.can` docs instead, but using `machine.resolveState(...)` there should be discouraged as well.

Andarist: I'm not sure if having a code here that breaks other functionality is good. By sparing a line of code here we can cause a headache for somebody who won't be aware of this. The ROI on this PR seems negative to me - I see how it looked redundant (and I think that we should simplify the API around this in the future so it wouldn't be a problem at all) but it turned out that it isn't

lodi-g: Yes, it did cause a headache and the loss of a whole day 😅. This is definitely worth documenting.

davidkpiano: > Yes, it did cause a headache and the loss of a whole day 😅. This is definitely worth documenting.

Very sorry about that.

In the upcoming version 5, you'll be happy to know that it'll be impossible to end up with a broken `state.can` because states will only be able to be created from `machine.fromState(...)`, which pretty much guarantees a non-broken state.

PR #2814, opened 2021-11-15
Removed StateNode as a valid type for transition targets

mattpocock:
https://www.loom.com/share/f9ff66b6ca9d4db4aefa87900bfcd4b7

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -206,28 +206,23 @@ export type Condition<TContext, TEvent extends EventObject> =
   | ConditionPredicate<TContext, TEvent>
   | Guard<TContext, TEvent>;
 
-export type TransitionTarget<
-  TContext,
-  TEvent extends EventObject
-> = SingleOrArray<string | StateNode<TContext, any, TEvent>>;
+export type TransitionTarget = SingleOrArray<string>;
 
-export type TransitionTargets<TContext> = Array<
-  string | StateNode<TContext, any>
->;
+export type TransitionTargets = Array<string>;
 
 export interface TransitionConfig<TContext, TEvent extends EventObject> {
   cond?: Condition<TContext, TEvent>;
   actions?: Actions<TContext, TEvent>;
   in?: StateValue;
   internal?: boolean;
-  target?: TransitionTarget<TContext, TEvent>;
+  target?: TransitionTarget;
   meta?: Record<string, any>;
   description?: string;
 }
 
 export interface TargetTransitionConfig<TContext, TEvent extends EventObject>
   extends TransitionConfig<TContext, TEvent> {
-  target: TransitionTarget<TContext, TEvent>; // TODO: just make this non-optional
+  target: TransitionTarget; // TODO: just make this non-optional
 }
 
 export type ConditionalTransitionConfig<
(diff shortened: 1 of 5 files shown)


packages/core/src/types.ts
```diff
@@ -206,28 +206,23 @@ export type Condition<TContext, TEvent extends EventObject> =
...
-  target?: TransitionTarget<TContext, TEvent>;
+  target?: TransitionTarget;
   meta?: Record<string, any>;
   description?: string;
 }
 
 export interface TargetTransitionConfig<TContext, TEvent extends EventObject>
   extends TransitionConfig<TContext, TEvent> {
-  target: TransitionTarget<TContext, TEvent>; // TODO: just make this non-optional
+  target: TransitionTarget; // TODO: just make this non-optional
```
mattpocock, line 225: Maybe we can remove this TODO?

packages/core/src/types.ts
```diff
@@ -1135,7 +1125,7 @@ export interface ChooseAction<TContext, TEvent extends EventObject>
 }
 
 export interface TransitionDefinition<TContext, TEvent extends EventObject>
-  extends TransitionConfig<TContext, TEvent> {
+  extends Omit<TransitionConfig<TContext, TEvent>, 'source' | 'target'> {
```
mattpocock, line 1128: There might be a better way of handling this? This looks fairly ugly.

Andarist:
This is something that we've wanted to do in v5 (and it actually might already be done there, but would have to double-check this). We can't do it easily in the current release line though - this is a breaking change.

mattpocock:
@Andarist Is this documented? I don't remember it in the docs.

Is there a different way we can handle this? I.e. with an `extends` check. Is there a way to add `// @deprecated` for a single member of a union?

Andarist:
You could try to do smth like this: [TS playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEkoGdnwMoBcqZAOQHtR4BvAKHkvgH1qAPZbXagOyJAC55MYBXEClWQEAtiAAKMAgAcujGAEsWAczIBfMmUwBPaQgDCBFgDMFy+AF5Sgyka4AlcARjAAPPKXKANPA8r4AD4YTPjsAHzqmmBGjL4slvAsIADuwTihoAAUAJSaoEhw8Ma8LGCYCkaIcOmGJmaZ0XXKXLWmytlcAG4ECsBR1bit9eRU8HbWo6MAYgDyM1wjk0sA9MvwULyYBNEi0hAguGgABsJikjJHNpMaoxpquWRgAyBDypmLtiwLV9NzcixXO4PHR6eCvABMCQ+Yy+8Ec0Rc7h4nh8fnMQRINHojHSrHYXB4-HgagiGjI+WghWKpXKlSeIBqRja4IaTLMLTZynBHXg3V6-QZg05LOh42hv3mEyWk1W8GABGQLEw60221EewOIGOpwkUmkl2lxMB6ge9MZTRFVzFPyos0liuN9yAA)

Note that the `__xstate_node` property (or a similar) already exists.

Andarist:
And the docs are here: https://xstate.js.org/docs/guides/ids.html#avoiding-strings

mattpocock:
@Andarist That's pretty wild - and breaks quite a few XState rules around determinism. Should we remove it from the docs?

Andarist:
A wild getter definitely breaks determinism but this section only mentioned a specific use case for this that shouldn't lead to any non-determinism. Either way - I'm totally fine with removing this section as this is something that we want to remove entirely in the v5. I'm just uncomfortable with making this a type error now (since, unfortunately, this has been in the docs for over 2 years or more).

Andarist:
We no longer allow state nodes as targets in v5 so this PR can be closed.

PR #2815, opened 2021-11-15
Added XStateError for attaching metadata to errors within XState

mattpocock:
https://www.loom.com/share/9e409d477f0c4f6d9d1651199d019a0f

### DIFF
--- a/packages/core/src/errors.ts
+++ b/packages/core/src/errors.ts
@@ -4,7 +4,7 @@ export interface XStateErrorMeta {
   /**
    * The path the error occurred in
    */
-  state?: State<any, any>;
+  state?: State<any, any, any, any>;
   /**
    * If it occurred within a state node, log the state node
    */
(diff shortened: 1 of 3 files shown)


packages/core/src/errors.ts
```diff
@@ -0,0 +1,20 @@
+import { State, StateNode } from '.';
+
+export interface XStateErrorMeta {
+  /**
+   * The path the error occurred in
+   */
+  state?: State<any, any>;
```
davidkpiano, line 7:
```suggestion
  state?: State<any, any, any, any>;
```

Probably need a few more `any`s to make TS happy unfortunately

Andarist:
Most of the adjusted errors are things that shouldn't happen in production, they are soft validations and stuff like that. For errors like this the error message should be enough, they are not meant to be handled at runtime (beyond passing them to the bug tracking services).

PR #2816, opened 2021-11-15
Added a small FAQ addition to the transitions page

mattpocock:


mattpocock:
These are very common questions in the Discord - handy to have them here to allow for better linking between resources.

docs/guides/transitions.md
```diff
@@ -561,3 +561,26 @@ quietMachine.transition(quietMachine.initialState, { type: 'WHISPER' });
...
+
+- If _something_ is true, go to this state
+- If _something else_ is true, go to this state
+- Else, go to this state
+
+You can use [guarded transitions](./guards.md#guarded-transitions) to achieve this.
+
+### How do I transition to _any_ state?
+
+You can transition to _any_ state by giving that state a custom id, and using `target: '#customId'`. You can read the [full docs on custom ids here](./ids.md#custom-ids).
```
Andarist, line 579: nit: other sections of the docs use this casing:
```suggestion
You can transition to _any_ state by giving that state a custom id, and using `target: '#customId'`. You can read the [full docs on custom IDs here](./ids.md#custom-ids).
```

docs/guides/transitions.md
```diff
@@ -561,3 +561,26 @@ quietMachine.transition(quietMachine.initialState, { type: 'WHISPER' });
...
+
+### How do I do if/else logic on transitions?
+
+Sometimes, you'll want to say:
+
+- If _something_ is true, go to this state
+- If _something else_ is true, go to this state
+- Else, go to this state
+
+You can use [guarded transitions](./guards.md#guarded-transitions) to achieve this.
```
Andarist, line 575: I think this should also be accompanied by a simple example - just referring to another section seems like a needless overhead for the user. The link to the section is still useful though, if they want to learn more about it etc

mattpocock: I actually think a simple example leads to excess duplication/maintenance. I've added PR's to that effect elsewhere. @laurakalbag What do you think?

mattpocock: I've added a conceptual example to illustrate the question, but I don't think a concrete example is needed when the link to the explicit docs section is there.

davidkpiano (approved):

docs/guides/transitions.md
laurakalbag: I think simple and short examples are useful, especially for folks who are more visual learners and/or aren’t native English speakers.

PR #2817, opened 2021-11-15
Removed section on getters from docs

mattpocock:
This section should be removed because it's not best practice, and encourages non-deterministic statecharts.


Andarist (approved):
Let's wait for @davidkpiano's ✅ too

davidkpiano:
Long overdue, thanks!

PR #2820, opened 2021-11-16
Check if `process` is defined

### LINKED ISSUE #787: Process references

davidkpiano:
This PR checks if `process` is defined before reading it. I ran into this when trying to use XState with Cloudflare Durable Objects.

Fixes #787

### DIFF
--- a/packages/core/src/environment.ts
+++ b/packages/core/src/environment.ts
@@ -1,1 +1,4 @@
-export const IS_PRODUCTION = process.env.NODE_ENV === 'production';
+export const IS_PRODUCTION =
+  typeof process !== 'undefined'
+    ? process.env.NODE_ENV === 'production'
+    : false;
--- a/packages/core/src/environment.ts
+++ b/packages/core/src/environment.ts
@@ -1,4 +1,4 @@
 export const IS_PRODUCTION =
   typeof process !== 'undefined'
     ? process.env.NODE_ENV === 'production'
-    : false;
+    : true;
--- a/packages/xstate-fsm/src/index.ts
+++ b/packages/xstate-fsm/src/index.ts
@@ -50,7 +50,10 @@ function toActionObject<TContext extends object, TEvent extends EventObject>(
     : action;
 }
 
-const IS_PRODUCTION = process.env.NODE_ENV === 'production';
+const IS_PRODUCTION =
+  typeof process !== 'undefined'
+    ? process.env.NODE_ENV === 'production'
+    : false;
 
 function createMatcher(value: string) {
   return (stateValue) => value === stateValue;
--- a/packages/xstate-react/src/fsm.ts
+++ b/packages/xstate-react/src/fsm.ts
@@ -9,6 +9,11 @@ import {
 import { useSubscription, Subscription } from 'use-subscription';
 import useConstant from './useConstant';
 
+export const IS_PRODUCTION =
+  typeof process !== 'undefined'
+    ? process.env.NODE_ENV === 'production'
+    : false;
+
 const getServiceState = <
   TContext extends object,
   TEvent extends EventObject = EventObject,
@@ -39,7 +44,7 @@ export function useMachine<
   StateMachine.Service<TContext, TEvent, TState>['send'],
   StateMachine.Service<TContext, TEvent, TState>
 ] {
-  if (process.env.NODE_ENV !== 'production') {
+  if (!IS_PRODUCTION) {
     const [initialMachine] = useState(stateMachine);
 
     if (stateMachine !== initialMachine) {


Andarist:
How have you tried to consume XState in this environment? Have you used any of the templates mentioned [here](https://developers.cloudflare.com/workers/learning/using-durable-objects#uploading-a-durable-object-worker)?

davidkpiano:
> How have you tried to consume XState in this environment? Have you used any of the templates mentioned [here](https://developers.cloudflare.com/workers/learning/using-durable-objects#uploading-a-durable-object-worker)?

Yes, I used the TS template.  I think it's a good idea to _not_ assume that `process` is defined everywhere XState can run; this problem has come up in the past quite a few times.

I fixed it by using the Rollup `replace` plugin, but developers shouldn't need to do that. This problem also occurs when trying to use XState in the browser (e.g., via `unpkg`) without using `xstate.web.js`.

Andarist:
I just find it a little bit ironic in general that the ecosystem has converged on using a tool that was designed to work with node (npm registry and CLIs interacting with it) and then started to complain about code distributes there to use node-specific conventions :shrug: 

I understand the overall frustration with this but I'm a little bit torned between various goals here. Somewhat luckily we now have better ways of distributing code for various target environments, or when using different conditions (like development/production). Not that long time ago there was no standard for doing this - and even today most people probably won't be able to configure tooling correctly to pick up correct things. This is very nuanced and that's why I believe that using higher-level tooling that configured stuff on our behalf is very important.

We need to decide what are our defaults and priorities though, some of the goals include:
- good DX (in a form of additional warnings/errors, only available in the development)
- bundlesize (thanks to the former being hidden behind a development flag this one gets optimized) 
- 0-config consumption

The problem with the current solution is that it violates first two. With the shape of this PR we regress on the second goal.

Take a look at what happens with a simple test:
```js
// input
const IS_PRODUCTION =
  typeof process !== 'undefined'
    ? process.env.NODE_ENV === 'production'
    : false;

export default function () {
  if (IS_PRODUCTION) {
    console.log('prod');
  } else {
    console.log('dev');
  }
}

// terser output
const o="undefined"!=typeof process&&"production"===process.env.NODE_ENV;export default function(){o?console.log("prod"):console.log("dev")}
```

As we might see - the dev-only code has stayed untouched and is still present in the output. This is of no surprise, if we consider the type of `IS_PRODUCTION` it's a `boolean` here. A static tool can't evaluate the condition because it doesn't have access to the runtime information.

We might think that  at least this will get fixed when one is using a "replacement" plugin. Take a look at such a situation then:
```js
// input
const IS_PRODUCTION =
  typeof process !== 'undefined' ? 'production' === 'production' : false;

export default function () {
  if (IS_PRODUCTION) {
    console.log('prod');
  } else {
    console.log('dev');
  }
}

// output
const o="undefined"!=typeof process;export default function(){o?console.log("prod"):console.log("dev")}
```

Unfortunately, not much of a change here - the dev-only code is still here. Why is that? Well, the type of the `IS_PRODUCTION` didn't change here - it's still a `boolean`.

We can try to flip the situation and assume that we default to the production code (the previous version uses `false` as the default for `IS_PRODUCTION`):
```js
// input
const IS_PRODUCTION =
  typeof process !== 'undefined' ? 'production' === 'production' : true;

export default function () {
  if (IS_PRODUCTION) {
    console.log('prod');
  } else {
    console.log('dev');
  }
}

// output
export default function(){console.log("prod")}
```

Ka-ching 💰 ! We have it! Well... do we? If we consult this with our overall goals then we might see that we've fixed the bundlesize problem but we've regressed on the DX part (by default, it can be gained back when using a bundler). It really is a [Fast Good Cheap](https://fastgood.cheap/) situation that we are having here.

Don't take me wrong - I'm open for adjusting this but I would like to make sure that we are making a well-informed decision here, taking all of that into consideration. 

Nowadays we could remove `process.env.NODE_ENV` stuff from the distributes files **entirely**, thanks to `package.json#exports`. Unfortunately, in v4 we kinda allow accessing stuff from `xstate/lib` and `xstate/es` and this becomes more tricky than it should be and I'm somewhat afraid of making such changes in v4. And even with that in place - the webpack's recommendation is to default to the production code and make dev code opt-in. If we apply that we end up with the last situation that I've depicted - no improved DX by default. Fortunately, I think that most of modern, popular, bundling "wrappers" (like Vite, Snowpack, webpack 5, etc) know how to pick development/production code based on `package.json#exports`. So I guess that those who are hand-crafting their configs would just need to know how to opt into the development code on their own.

packages/core/src/environment.ts
```diff
@@ -1 +1,4 @@
-export const IS_PRODUCTION = process.env.NODE_ENV === 'production';
+export const IS_PRODUCTION =
+  typeof process !== 'undefined'
+    ? process.env.NODE_ENV === 'production'
+    : false;
```
Andarist, line 4: Based on the end goal described [here](https://github.com/statelyai/xstate/pull/2820#issuecomment-970406584) and that this would minify better I think we should switch to this:
```suggestion
export const IS_PRODUCTION =
  typeof process !== 'undefined'
    ? process.env.NODE_ENV === 'production'
    : true;
```

davidkpiano:
@Andarist Should we close this or go with your solution (which I committed)?

Andarist:
superseded by https://github.com/statelyai/xstate/pull/4016

PR #2822, opened 2021-11-17
Update types.ts to work with --exactOptionalPropertyTypes

dai-shi:
While I was trying to enable `exactOptionalPropertyTypes` in tsconfig.json, I got tsc errors.
https://github.com/pmndrs/jotai/pull/839

To work around it, it's using a patch now. This is the same fix, without understanding anything.


dai-shi:
Maybe we should add `| undefined` to the base type, instead of removing it.

davidkpiano (approved):

davidkpiano:
@Andarist Is this good to go in?

PR #2824, opened 2021-11-18
[v5] Ensure that null events are not recorded

davidkpiano:
This PR ensures that when taking eventless transitions, the resulting event that is read by subsequent guards and actions is the event that _triggered_ the initial transition; not "null events" (`{ type: '' }`)", which should not be recorded anyway.

Addresses https://github.com/statelyai/xstate/projects/1#card-34972159


packages/core/test/guards.test.ts
```diff
@@ -107,9 +107,11 @@ describe('guard conditions', () => {
 
   it('should not transition if no condition is met', () => {
     const nextState = lightMachine.transition(
-      State.from('green', {
-        elapsed: 9000
-      }),
+      lightMachine.resolveState(
```
Andarist, line 110: q: in this file you wrap all `State.from` with `machine.resolveState` - this is correct and better, i just wonder if there is any particular logic in this PR that has made this obligatory right now?

packages/core/src/stateUtils.ts
```diff
@@ -505,7 +498,7 @@ export function formatTransitions<
       )
     : [];
   const eventlessConfig = stateNode.config.always
```
Andarist, line 500: it feels like this should be put on a separate key on the state - instead of mixing this together with other, regular, transitions

this would make the `selectEventlessTransitions` slightly easier and would make it more performant - we wouldn't have to iterate through the whole list of transitions just to get to the eventless ones. We could move `eventlessConfig` to the the beginning of the `formattedTransitions` but then we'd still have to implement a stop condition in the `selectEventlessTransitions` looping. Separating those transition types would just make it working well out of the box - and it would make reading this information in the editor and stuff easier as one would only have to access a single property to read those special transitions

packages/core/src/StateNode.ts
```diff
@@ -24,7 +24,6 @@ import type {
   InvokeDefinition,
```
Andarist, line 24: idea: could we add a dev-only warning to the `machine.transition` that would warn if somebody tries to send `event.type === ''`? Let's require a non-empty string there

packages/core/src/stateUtils.ts
```diff
@@ -1842,27 +1827,41 @@ export function macrostep<
...
   let maybeNextState = nextState;
 
-  while (_internalQueue.length && !maybeNextState.done) {
-    const _previousEvent = maybeNextState._event;
-    const raisedEvent = _internalQueue.shift()!;
-    const currentActions = maybeNextState.actions;
-
-    maybeNextState = machine.microstep(
+  while (!maybeNextState.done) {
+    const enabledTransitions = selectEventlessTransitions(
```
Andarist, line 1831: `enabledTransitions` sounds like **any** enabled transitions
```suggestion
    const eventlessTransitions = selectEventlessTransitions(
```

packages/core/src/types.ts
```diff
@@ -503,7 +501,6 @@ type TransitionsConfigArray<
   | (TEvent extends EventObject
       ? TransitionConfig<TContext, TEvent> & { event: TEvent['type'] }
       : never)
-  | (TransitionConfig<TContext, TEvent> & { event: '' })
```
Andarist, line 506: sideways note -> what do you think about dropping support for `TransitionsConfigArray` entirely? This has been mainly introduced to support compatibility with SCXML better but it imho doesn't bring a lot of value and nobody is using this publicly. This isn't even documented.

packages/xstate-scxml/test/fixtures/actionSend/send4.ts
```diff
@@ -9,14 +9,14 @@ const m = createMachine({
     },
```
Andarist, line 9: this PR should probably go over our SCXML-compatibility test suite and try to enable some new tests. Most of the tests are associated by the comments on why they are failing so locating the ones concerned with the changed behavior here should be easy to locate.

packages/core/test/guards.test.ts
davidkpiano: Yes, `state.done` needs to read the machine from `state.configuration[0].machine` but `state.configuration` is empty from `State.from(...)`; `machine.resolve` provides that.

Same if we want to read from `state.machine` - that does not exist in `State.from(...)`

davidkpiano: This PR checks `state.done` in the `while` loop, and it will throw an error without this.

packages/core/src/StateNode.ts
davidkpiano: Yes, good idea

packages/core/src/stateUtils.ts
davidkpiano: Agree, will change - this is just copied from the SCXML algorithm

packages/xstate-scxml/test/fixtures/actionSend/send4.ts
davidkpiano: I took a brief look and didn't immediately find any tests that this enabled.

packages/core/src/types.ts
davidkpiano: I'm wary to do this because an array offers interesting (and sometimes useful) patterns for resolving wildcard events.

Andarist: The wildcard case was partially a reason why i've added array support in the first place - after all that time I never really wanted to actually use this though, nor I've seen anyone using this. We are currently deprioritizing `*` in the object syntax (it's always matched after other transitions with exact descriptors) but the question is - should the same happen to other inexact descriptors, like `foo.*`? I'm on the fence here - if we shuffle priority orders on our own it seems a little bit like adding magic rules. If instead we would just trust the document order (for all descriptors, `*` included) then our implementation would get simpler but also the public behavior would be more predictable (we'd only have one rule -> document order).

While having an array support in place we can do more than with an object - question is: is this even a good thing in the first place? With time I'm getting more and more into KISS, as in - don't try to be interesting, let the code be dull and boring. Any clever solution can be refactored to a more dull one and the dull one will probably be easier to grasp over time. So I'm wondering if removing support for this wouldn't be actually a feature, forcing people not to use clever solutions. 

It also seems that full array support is just harder to visualize - although document-order solution comes with similar problems, but at least the same descriptor cant be interleaved with other descriptors if we go with object-only syntax.

cc @johnyanarella @ChrisShank @ojkelly what are your thoughts on array support? how descriptor matching is working in StateML so far?

davidkpiano: Another use-case for a transition array: the planned feature of letting _multiple events_ be handled by the same transition.

```js
on: [
  { event: ['A_EVENT', 'B_EVENT', target: 'goHere' }
]
```

ChrisShank: @Andarist in John’s original DSL multiple events for a single transition was supported since it was largely compliant with SCXML. We have since removed this ability due to its apparently little to no usage and confusing semantics. Potential confusion for a user stems from whether to interpret the array of events as `or` or `and`. In other parts of the DSL an array means `and` (e.g. multiple targets) while this array means `or`.

By removing this feature we don't loss expressiveness since there is a simple rewrite that represents the same model. There are also concerns of how to represent this feature in the visualization that implies the correct semantics.

If there are legit use cases in the future we would reconsider adding this to the language.

packages/core/src/StateNode.ts
davidkpiano: I've forbidden these

packages/core/src/stateUtils.ts
davidkpiano: These are now on `stateNode.always`

davidkpiano:
Good to go @Andarist?

davidkpiano:
@Andarist ping 😃

packages/core/src/types.ts
Andarist: > Another use-case for a transition array: the planned feature of letting multiple events be handled by the same transition.

IMHO, this sounds neat at the surface and I've also wanted to introduce this in the past. But now I don't think that introducing a syntax that makes you refactor the whole chunk of code is good. A good syntax should be "appendable" and focused, should allow local modifications. If we'd process event keys we could allow stuff like:
```js
on: { 
  "A_EVENT, B_EVENT": "go here"
}
```
This can even be written like this:
on: { 
  [["A_EVENT", "B_EVENT"]]: "go here"
}
that is allowed because JS would just stringify the array put on the computed property and stringifying an array joins elements with a comma.

A different delimited could be chosen, like whitespace, but then we couldn't leverage the computed property syntax.

I agree that this doesn't look that great and makes us "decode" the event descriptors from the property name. I'm not super attached to this idea but I think it's better not to allow this at all than to keep the array syntax in place. Having to copy-paste stuff can be a feature and not a limitation 😉 And note that the visual editor can simplify this - even if the target export syntax would involve some "duplicates" to be "emitted".

> By removing this feature we don't loss expressiveness since there is a simple rewrite that can represents the same model.

💯

packages/core/src/interpreter.ts
```diff
@@ -765,6 +765,9 @@ export class Interpreter<
           } else {
             this.logger(value);
           }
+        },
+        [actionTypes.assign]: () => {
+          /* No-op to prevent no implementation warning */
```
Andarist, line 770: this should be flagged with a TODO or something as it's a strong indication that things are wired weirdly here

packages/core/src/stateUtils.ts
```diff
@@ -1511,14 +1502,17 @@ function selectEventlessTransitions<
     loop: for (const s of [stateNode].concat(
       getProperAncestors(stateNode, null)
     )) {
-      for (const t of s.transitions) {
+      if (!s.always) {
+        continue;
+      }
+      for (const t of s.always) {
         if (
           t.eventType === NULL_EVENT &&
```
Andarist, line 1510: This is redundant here now
```suggestion
```

Andarist (approved):

PR #2827, opened 2021-11-19
Fix websocket receiver

erlendfh:
I couldn't get the websocket receiver to work until I changed this line, so I'm assuming this is how it should read (this matches how `createWindowReceiver` is implemented).

### DIFF
--- a/.changeset/wise-wasps-drive.md
+++ b/.changeset/wise-wasps-drive.md
@@ -2,4 +2,4 @@
 '@xstate/inspect': patch
 ---
 
-Fixed a bug in createWebsocketReceiver
+Fixed a bug in `createWebsocketReceiver` so that it works as expected with a WebSocket connection.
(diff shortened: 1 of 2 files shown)


davidkpiano (approved):

davidkpiano:
Can you add a changeset (via `yarn changeset` - make it a patch change) briefly describing the fix? That way it can be versioned properly.

.changeset/wise-wasps-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/inspect': patch
+---
+
+Fixed a bug in createWebsocketReceiver
```
davidkpiano, line 5:
```suggestion
Fixed a bug in `createWebsocketReceiver` so that it works as expected with a WebSocket connection.
```

PR #2835, opened 2021-11-25
Added interop observable symbols to ActorRef

### LINKED ISSUE #2834: Let ActorRef be an RxJS InteropObservable
A while back we fixed the `Interpreter` so that it was recognized by RxJS as an `InteropObservable` (using interop symbols). However, when retrieving children of a machine, this does not suffice, since these children are typed as `ActorRef`. Would it be possible to have `ActorRef`, or even it's parent interface `Subscribable`, to either mandate their subclasses to contain the necessary symbols, or change them into abstract classes containing the symbols themselves?

If this is not an immediate NO, and I find some time, I can try to implement a suggestion.

woutermont:
Closes #2834


.changeset/tasty-terms-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Added interop observable symbols to ActorRef and Subscribable
```
davidkpiano, line 5:
```suggestion
Added interop observable symbols to `ActorRef` and `Subscribable` so that actor refs are compatible with libraries like RxJS.
```

davidkpiano (approved):

woutermont:
Making `Subscribable` be an `InteropObservable` is obviously not the way to go, so I reverted that part, simply added the `InteropObservable` from RxJS and let only `ActorRef` extend it.

davidkpiano:
@Andarist LGTY?

packages/core/src/types.ts
```diff
@@ -3,6 +3,12 @@ import { State } from './State';
 import { Interpreter, Clock } from './interpreter';
 import { IsNever, Model, Prop } from './model.types';
 
+declare global {
+  interface SymbolConstructor {
+    readonly observable: symbol;
+  }
+}
```
Andarist, line 10: I don't think a global type should be added by us here. It's not our responsibility to do this on behalf of our users and this is risking some conflicts with other types that may be loaded implicitly and extending the same thing.

packages/core/src/utils.ts
```diff
@@ -530,8 +530,18 @@ export function isObservable<T>(value: any): value is Subscribable<T> {
...
-  (typeof Symbol === 'function' && (Symbol as any).observable) ||
-  '@@observable')();
+  (typeof Symbol === 'function' && Symbol.observable) || '@@observable')();
+
+export const interopSymbols = {
+  [symbolObservable]: function () {
+    return this;
+  },
+  // this gets stripped by Babel to avoid having "undefined" property in environments without this non-standard Symbol
+  // it has to be here to be included in the generated .d.ts
```
Andarist, line 540: have you confirmed that this works? is the `Symbol.observable` emitted in the `.d.ts`? is this property in the dist files from here by Babel?

.changeset/tasty-terms-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+Added interop observable symbols to `ActorRef` and `Subscribable` so that actor refs are compatible with libraries like RxJS.
```
Andarist, line 5: As far as I can tell this has only been added to the `ActorRef` interface, right?
```suggestion
Added interop observable symbols to `ActorRef` so that actor refs are compatible with libraries like RxJS.
```

packages/xstate-inspect/src/browser.ts
```diff
@@ -7,7 +7,12 @@ import {
   Observer
 } from 'xstate';
 import { XStateDevInterface } from 'xstate/lib/devTools';
-import { toSCXMLEvent, toEventObject, toObserver } from 'xstate/lib/utils';
+import {
+  toSCXMLEvent,
+  toEventObject,
+  toObserver,
+  interopSymbols
```
Andarist, line 14: There might be compatibility issues between mismatched versions of `xstate` and `@xstate/inspect` - I would just touch the core directory as part of this PR and revert the changes here.

Andarist (changes requested):
It would be great if we could add some tests for this.

.changeset/tasty-terms-remain.md
woutermont: Yes, was reverted later for `Subscribable`

packages/xstate-inspect/src/browser.ts
woutermont: How would you suggest to do this? `InspectReceiver` of `@xstate/inspect` is an alias for an `ActorRef`, which is impacted by these changes in the core.

packages/core/src/utils.ts
woutermont: Yes, only `[symbolObservable]` is used in the ES file, and `[Symbol.observable]` is present in the type declarations.

packages/core/src/types.ts
woutermont: Weird, I believe we added this in the `Interpreter.ts` module when first looking into interop observables, to make TS believe the existence of `Symbol.observable`, but somehow it now doesn't seem to be necessary anymore (TS version bump?). Should not matter for the user, since the global type is not exported, but I'll try to remove them anyhow.

woutermont:
That would indeed be great, but the only thing this PR does is adding an interface, and making an existing interface extend from it (with some sugar to implement it) ... where/how should I test this?

woutermont:
@Andarist, can you confirm my changes and comments are satisfactory, or reply once more to them?

davidkpiano:
Confirmed to be working. Here is without the change:

CleanShot 2021-12-30 at 18 36 28

And here is with this PR:

CleanShot 2021-12-30 at 18 35 35

packages/xstate-inspect/src/browser.ts
davidkpiano: We can revise these in a follow-up PR.

Andarist:
@woutermont are you using this feature? I'm wondering **how** you are using this at runtime because I have a suspicion that this might not work correctly for all the cases. So I wonder - for which exact cases is this working for you (I can think of one).

woutermont:
We're using this to combine XState ActorRefs (specifically the Interpreter, I believe) with RxLitElement. I'm currently on holiday, but can provide more info when back.

PR #2837, opened 2021-11-27
Add a failing test for `state.can` that involves parallel regions

Andarist:
Add a failing test for regression from https://github.com/statelyai/xstate/pull/2804

### DIFF
--- a/packages/core/test/state.test.ts
+++ b/packages/core/test/state.test.ts
@@ -846,6 +846,42 @@ describe('State', () => {
 
       expect(executed).toBeFalsy();
     });
+
+    it('should return true when non-first parallel region changes value', () => {
+      let executed = false;
+      const machine = createMachine({
+        type: 'parallel',
+        states: {
+          a: {
+            initial: 'a1',
+            states: {
+              a1: {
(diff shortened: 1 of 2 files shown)


davidkpiano:
Please see https://github.com/statelyai/xstate/pull/2839 for a potential fix, but also one that reveals a bigger problem: not all targets appear in the transition data.

IMO this is a much smaller edge-case than what #2804 fixed, and should not prevent #2804 from going in.

packages/core/test/state.test.ts
```diff
@@ -846,6 +846,42 @@ describe('State', () => {
...
+        type: 'parallel',
+        states: {
+          a: {
+            initial: 'a1',
+            states: {
+              a1: {
+                id: 'foo',
+                on: {
+                  // first region doesn't change value here
+                  EVENT: ['#foo', '#bar']
```
davidkpiano, line 862: It should be this:

```suggestion
                  EVENT: { target: ['#foo', '#bar'] }
```

Andarist:
closing this as it became a part of the https://github.com/statelyai/xstate/pull/2839

PR #2839, opened 2021-11-27
[core] Add potential fix for failing `state.can` test

davidkpiano:
This PR potentially fixes #2837, but reveals another problem in which transitions with multiple targets might be incorrect, preventing this fix from working.

### DIFF
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -343,17 +343,24 @@ export class State<
      * The state will change from this event if:
      * - There are entry actions
      * - There are exit actions
-     * - There are transitions in which either the target has changed or there are transition actions
+     * - There are transitions in which either the targets have changed or there are transition actions
      */
-    return (
+    if (
       transitionData.entrySet.some((sn) => sn.onEntry.length > 0) ||
-      transitionData.exitSet.some((sn) => sn.onExit.length > 0) ||
-      !!transitionData.transitions.some((t) => {
-        return (
-          (t.target && t.target.length > 0 && t.target[0] !== t.source) ||
-          t.actions.length > 0
-        );
-      })
-    );
+      transitionData.exitSet.some((sn) => sn.onExit.length > 0)
+    ) {
+      return true;
+    }
+
+    for (const t of transitionData.transitions) {
+      if (t.actions.length) {
+        return true;
+      }
+      if (t.target?.some((target) => !this.configuration.includes(target))) {
+        return true;
+      }
+    }
+
+    return false;
   }
 }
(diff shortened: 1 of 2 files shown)


Andarist:
> another problem in which transitions with multiple targets might be incorrect, preventing this fix from working.

Do you have a test case for this? When that could happen?

packages/core/test/state.test.ts
```diff
@@ -846,6 +846,44 @@ describe('State', () => {
...
+          }
+        }
+      });
+
+      const { initialState } = machine;
+
+      expect(machine.transition(undefined, { type: 'EVENT' }).value).toEqual({
+        a: 'a1',
+        b: 'b2'
+      });
```
Andarist, line 883: Weirdly... this passes for me locally but fails on CI. Gonna investigate it on Monday

davidkpiano:
@Andarist This passes on CI now 👍

packages/core/src/State.ts
```diff
@@ -335,21 +335,32 @@ export class State<
...
-      ? false
-      : transitionData.entrySet.some((sn) => sn.onEntry.length > 0) ||
-          transitionData.exitSet.some((sn) => sn.onExit.length > 0) ||
-          !!transitionData?.transitions.some((t) => {
-            return (
-              (t.target && t.target.length > 0 && t.target[0] !== t.source) ||
-              t.actions.length > 0
-            );
-          });
+    if (
```
Andarist, line 348: resolving this correctly is quite complex, just added yet another failing test with this logic: https://github.com/statelyai/xstate/pull/2850

I wonder if it wouldn't be just easier to redefine `state.can` as "will transition be selected based on the given event". This way we'd only have to check something like:
```js
return !!transitionData?.getTransitionData(this, event).transitions.length
```

davidkpiano: @Andarist Changed the behavior here: 9c1a28e

packages/core/test/state.test.ts
```diff
@@ -716,7 +716,7 @@ describe('State', () => {
       expect(machine.initialState.can({ type: 'EV' })).toBe(true);
     });
 
-    it('should return false for a forbidden transition', () => {
+    it('should return true for a defined transition without a target', () => {
```
Andarist, line 719: We explicitly call this out in the docs as a "forbidden" transition and I think this should return `false`. When suggesting the simplified version of the check for `.can` I didn't realize that `.transitions` actually contain those forbidden transitions. In such a case the check should be slightly modified to account for this situation.

davidkpiano: Done

Andarist (approved):
I took the liberty of landing https://github.com/statelyai/xstate/pull/2850 into this branch. 

LGTM!

PR #2840, opened 2021-11-27
[v5] Remove `interpreter.children`

davidkpiano:
This PR removes `interpreter.children` in favor of `interpreter.state.children` (redundant).


packages/core/src/interpreter.ts
```diff
@@ -818,12 +816,11 @@ export class Interpreter<
...
   private stopChild(childId: string): void {
-    const child = this.children.get(childId);
+    const child = this.state.children[childId];
     if (!child) {
       return;
     }
 
-    this.children.delete(childId);
     this.forwardTo.delete(childId);
     delete this.state.children[childId];
```
Andarist, line 825: This probably should be refactored to avoid mutation of `this.state` - I think most people treat this as immutable. Maybe grabbing the actor ref to stop within machine APIs~ and embedding that in the action would help? This way we wouldn't have to access this here by ID

davidkpiano: Not a good idea; prevents them from getting GC'd. This is one of those cases where we might need to allow these to be mutated.

davidkpiano: Let's address this in a separate PR. This isn't part of this PR's changes.

Andarist:
Let's not forget about a changeset here - this is a pretty big user-facing change.

packages/core/src/interpreter.ts
Andarist: Do we have any ticket anywhere to track this?

davidkpiano: #2851

PR #2841, opened 2021-11-28
[core] feat(interpreter): allow to provide errorListeners

roggervalf:
This PR tries to handle this missing feature, where error listeners were not attached.


packages/core/src/interpreter.ts
```diff
@@ -1057,6 +1086,9 @@ export class Interpreter<
             if (this.devTools) {
               this.devTools.send(errorEvent, this.state);
             }
+            if (this.errorListeners.size) {
+              this.sendError(error);
+            }
```
Andarist, line 1091: There are more places where errors could happen and I think they should be covered as well **IF** we decide to add this feature.

I think that in general there might be a need for a top-level `onError` listener but we should think carefully through how this works with multiple nested machines etc.

roggervalf: thank you @Andarist, I tried to update my pr, please when you get a chance :eyes: to take a look

.changeset/famous-icons-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': major
```
davidkpiano, line 2: This should be either a `minor` or target the `next` branch (for v5)

roggervalf: oh sorry I just had that option by using `yarn changeset` (first time using it), now it should be updated

Visot:
Interesting that the errorListener is missing, I was also wondering why errors were not appearing, any chance to have a fix for it?

packages/core/test/interpreter.test.ts
```diff
@@ -1732,6 +1732,61 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
       );
     });
 
+    it('should be handle child errors with errorListener', (done) => {
```
manast, line 1735: typo: should be handle -> should handle

packages/core/src/interpreter.ts
manast: > There are more places where errors could happen and I think they should be covered as well IF we decide to add this feature.

Could you help us to find those places where the errors could happen please?

packages/core/test/interpreter.test.ts
roggervalf: thank you, I missed that

roggervalf:
> Interesting that the errorListener is missing, I was also wondering why errors were not appearing, any chance to have a fix for it?

hi @Visot, I'm still trying to do it in this pr

packages/core/src/interpreter.ts
```diff
@@ -552,13 +583,22 @@ export class Interpreter<
    */
   public send = (
     event: SingleOrArray<Event<TEvent>> | SCXML.Event<TEvent>,
-    payload?: EventData
+    payload?: EventData,
+    sendError = false
```
davidkpiano, line 587: Let's remove this; errors should either be unhandled `error.*` events or "natural" errors thrown from execution of the machine.

roggervalf: thank you @davidkpiano I just see your message, in my last commit I was replacing this by checking the type in the event, is that ok or should I do in a different way?

Visot:
+1 on this one, super useful if implemented

packages/core/src/interpreter.ts
```diff
@@ -1053,10 +1092,20 @@ export class Interpreter<
             // Send "error.platform.id" to this (parent).
             this.send(toSCXMLEvent(errorEvent as any, { origin: id }));
           } catch (error) {
-            reportUnhandledExceptionOnInvocation(errorData, error, id);
+            if (this.parent) {
+              this.parent.send({
+                type: 'xstate.error',
```
roggervalf, line 1097: @davidkpiano in this case, is it ok this type of error or should it be a different one?

eliecerUribe (commented):
Very useful feature for this. Good job! 💯

manast:
What's left in order to get this PR merged?

packages/core/src/interpreter.ts
```diff
@@ -559,6 +590,14 @@ export class Interpreter<
       return this.state;
     }
 
+    if (((event as EventObject)?.type || '').includes('error')) {
```
Andarist, line 593: If anything - this should be checked on the `_event` (and on the appropriate field). It might be "too soon" to check this as the `event` might not be an `EventObject`, it actually might be already a `SCXML.Event`.

This should also check if the type/name of the event starts with `error.` and not if the type contains an `error` substring. 

also, a nit: optional chaining here and defaulting this to an empty string seems unnecessary:
- the `event` parameter is required
- all events should have valid types/names

packages/core/src/interpreter.ts
```diff
@@ -559,6 +590,14 @@ export class Interpreter<
...
+      if (this.parent) {
+        this.parent.sendError(event as any);
+      } else {
+        this.sendError(event as any);
```
Andarist, line 597: This one sounds like it's going to send this event somewhere (like some kind of an actor or something), but it's more like notifying current subscribers.

I also do not (overall) liking how this might make more errors to be swallowed. In general, the error story in XState is currently under-specified so it's somewhat hard to answer "how this should behave **now**?". I would love to hear out your thoughts about this RFC: https://github.com/statelyai/rfcs/pull/4 . Going through that could also show you my perspective better.

packages/core/src/interpreter.ts
roggervalf: thank you, I updated it

packages/core/src/interpreter.ts
roggervalf: Looks pretty nice, could you please guide me on the proper changes when you get some chance?

Andarist:
I'm sorry that this PR didn't get the attention it deserved in the past. In v5 those parts changed and I think that the issue at hand was already addressed on that branch.

PR #2845, opened 2021-11-30
[core] Add failing test for general `model.assign(...)` actions

davidkpiano:
This PR demonstrates the failure for transitions to accept general `model.assign(...)` functions with the following characteristics:

1. Are defined _outside_ of the machine definition
2. Can accept any event

Potentially addressed by https://github.com/statelyai/xstate/pull/2690

### DIFF
--- a/packages/core/test/model.test.ts
+++ b/packages/core/test/model.test.ts
@@ -41,6 +41,11 @@ describe('createModel', () => {
       'updateName'
     );
 
+    // Action that can take any event
+    const clearName = userModel.assign({
+      name: ''
+    });
+
     const machine = createMachine<UserContext, UserEvent>({
       context: userModel.initialContext,
       initial: 'active',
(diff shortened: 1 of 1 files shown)


packages/core/test/model.test.ts
```diff
@@ -58,6 +63,9 @@ describe('createModel', () => {
                   age: e.value
                 };
               })
+            },
+            anotherEvent: {
+              actions: clearName
```
Andarist, line 68: Removing `exec` property (a contravariant property) from here:
https://github.com/statelyai/xstate/blob/c6ead65609a591bc6949e90d572ff43932dfc4a4/packages/core/src/types.ts#L44
fixes the problem

However, then we stop inferring the event type correctly here:
https://github.com/statelyai/xstate/blob/2ab14af07b5ef4b07279d729086dedcd1ea352d1/packages/core/test/model.test.ts#L61

It's not being narrowed down to a specific event and instead `e` becomes a union of all events.

davidkpiano:
Should this be tackled in v5 instead, where we have better typings for actions?

Andarist:
Since we phased out `createModel` API this is no longer needed and types are being reworked on anyway.

PR #2848, opened 2021-12-01
docs(strict): add strict documentation

roggervalf:
strict mode is reference from https://xstate.js.org/docs/guides/hierarchical.html#events but strict mode docs are missing

### DIFF
--- a/docs/guides/machines.md
+++ b/docs/guides/machines.md
@@ -37,9 +37,11 @@ const lightMachine = createMachine({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


davidkpiano:
Thanks for the PR! Strict mode is going away in v5, so we don't want to emphasize it in the docs, but I'll wait for other's thoughts.

roggervalf:
> Thanks for the PR! Strict mode is going away in v5, so we don't want to emphasize it in the docs, but I'll wait for other's thoughts.

Thanks @davidkpiano just wanted to know if it was on purpose, I can close this pr as this will be removed anyway

PR #2850, opened 2021-12-01
Add a failing test for `state.can` that involves targeting a node in the current configuration

Andarist:


packages/core/test/state.test.ts
```diff
@@ -875,14 +875,35 @@ describe('State', () => {
...
+                on: {
+                  NEXT: '#foo'
+                }
+              }
+            }
+          }
+        }
       });
 
-      expect(initialState.can('EVENT')).toBeTruthy();
```
huan, line 885: How about let's keep this line, permit this unit test to get passed?

Because

> it makes sense for a use-case to be "I want to know if this event with any parameter will cause a transition to be selected". - @davidkpiano

PR #2864, opened 2021-12-07
[core] Do not use `src` in place of missing invoke `id`

### LINKED ISSUE #464: onDone not being scoped per state
## Description:

Generated done event is being scoped only per service name and not being scope per state, this means that if we invoke the same service simultaneously on different parallel states (or even on different machines if we use autoForward) then all of those pending states will resolve whenever first invocation resolves.

This is basically similar problem to the one described in #456 and #427 . 

## (Bug) Expected result:

onDone being independent on parallel states

## (Bug) Actual result:

onDone being called for resolved invocations from other states

## (Bug) Potential fix:

Generated event types for done & error events would have to include both a service name AND a machine id+state id. This would produce ugly event names though, so maybe we should attach this data to the event payload & attach appropriate guards to the generated transitions.

If we agree on the solution I would be happy to work on the fix.

```js
const createSingleState = () => ({
  initial: "fetch",
  states: {
    fetch: {
      invoke: {
        src: "fetchSmth",
        onDone: {
          target: "done",
          actions: "notifySingleSuccess"
        }
      }
    },
    done: {
      type: "final"
    }
  }
});

const testMachine = Machine(
  {
    id: "test",
    type: "parallel",
    onDone: {
      actions: "notifyBothSuccess"
    },
    states: {
      first: createSingleState(),
      second: createSingleState()
    }
  },
  {
    actions: {
      notifySingleSuccess: () => console.log(Date.now(), "single success"),
      notifyBothSuccess: () => console.log(Date.now(), "both success")
    },
    services: {
      fetchSmth: () => {
        const ms = Math.floor(Math.random() * 1000);
        console.log("gonna resolve in", ms);
        return new Promise(resolve =>
          setTimeout(() => {
            console.log(Date.now(), "resolving single");
            resolve();
          }, ms)
        );
      }
    }
  }
);

```

davidkpiano:
This PR replaces the erroneous "id or src" logic for an invocation's `id` with a unique and predictable generated `id`.

Fixes #464


davidkpiano:
@Andarist LGTY? This is more of a bugfix than a breaking change in v4.

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -43,6 +43,7 @@ describe('useMachine hook', () => {
       },
       loading: {
         invoke: {
+          id: 'fetchData',
```
Andarist, line 46: q: was there a reason behind adding this here? couldn't this rely on the auto-generated id? I'm just wondering if this had to be added here or if it could use the auto-generated id to assess the scope of the change

Andarist:
> @Andarist LGTY? This is more of a bugfix than a breaking change in v4.

What changed from [the assessment made in 2020](https://github.com/statelyai/xstate/issues/464#issuecomment-670939641)? :P

I'm OK with the change though - but a test case should be added to cover for this before merging.

A related issue is that this will only fix scoping within a single machine - but the problem will still be there if we consider manual or auto-forwarding of events. In general, I think that the `autoForward` should be removed altogether in v5 (it's a major footgun when it comes to "internally" generated events, like `done.invoke`, and breaks type-safety) - so I'm not that concerned about this. However, I wonder how the situation can be improved for manual forwarding - I guess this is a broad topic though and deserves a separate discussion. We should make sure that forwarding can be done in a type-safe manner and we could forbid forwarding "internally" generated events altogether (apart from error events) - I don't think there is much value in forwarding `done.invoke.*`, `done.state.*`, `xstate.start`, `xstate.stop` and similar events and if those "escape" the relevant machine they can wreck havoc in other machines (lead to weird situations).

davidkpiano:
> What changed from the assessment made in 2020? :P

The realization that the existing behavior is indeed buggy 🐛 

> I think that the autoForward should be removed altogether in v5

Agreed 👍

packages/xstate-react/test/useMachine.test.tsx
davidkpiano: It _can_ use the autogenerated ID, but the test is specifically testing:

```js
//          vvvvvvvvv
doneInvoke('fetchData', 'persisted data')
```

So I wanted to keep that part of the test the same.

davidkpiano:
Also, the reason why it's buggy behavior currently is because there is an obvious potential conflict:

```js
createMachine({
  type: 'parallel',
  states: {
    a: { invoke: { src: 'something' } },
    b: { invoke: { src: 'something' } },
  }
});
```

Both invocations will run simultaneously _with the same ID_. Ideally, no collisions should occur, and this is guaranteed by using `<state id>.invocation[<index>]`.

packages/xstate-react/test/useMachine.test.tsx
Andarist: 👌

Andarist:
I know why this was wrong, this was my bug report! 😅

packages/core/src/StateNode.ts
```diff
@@ -414,7 +414,7 @@ class StateNode<
         return toInvokeDefinition({
           ...invokeConfig,
 
-          id: invokeConfig.id || (invokeConfig.src as string),
+          id: invokeConfig.id || `${this.id}:invocation[${i}]`,
```
Andarist, line 417: I believe that we still have the same problem in other branches of this `.map`:
1. in the `isMachine(invokeConfig)` branch we should always use the auto-generated id as [reusing `invokeConfig.id`](https://github.com/statelyai/xstate/blob/4c2fdbdcefafa5db71f150506814f49ef70e0ee6/packages/core/src/StateNode.ts#L411) is dangerous for the very same reason, this would also match the behavior of the branch that checks `isMachine(invokeConfig.src) || isFunction(invokeConfig.src)` as in there the [`invokeSrc` is always autogenerated](https://github.com/statelyai/xstate/blob/4c2fdbdcefafa5db71f150506814f49ef70e0ee6/packages/core/src/StateNode.ts#L421)
2. the else branch for parametrized invocations is also [reusing `invokeConfig.src.type`](https://github.com/statelyai/xstate/blob/4c2fdbdcefafa5db71f150506814f49ef70e0ee6/packages/core/src/StateNode.ts#L436) which is even more close to the bug already fixed in this PR as this is always~ a string referencing an actor definition that can be looked up in the `implementations` argument

packages/core/test/invoke.test.ts
```diff
@@ -2709,6 +2709,44 @@ describe('invoke', () => {
       interpret(machine).start();
     });
   });
+
+  it('invoke generated ID should be predictable based on the state node where it is defined', (done) => {
```
Andarist, line 2713: this is an OK test but doesn't directly test the bug that is being fixed 😉 a regression test can be created based on the code included in the original report [here](https://github.com/statelyai/xstate/issues/464)

Andarist:
Just a reminder - this also deserves a changeset :)

packages/core/test/invoke.test.ts
davidkpiano: It is not easy to make a repro from the original code, since it depends on non-working parallel state behavior (`onDone` isn't called on a top-level parallel state).

Andarist: > It is not easy to make a repro from the original code, since it depends on non-working parallel state behavior (onDone isn't called on a top-level parallel state).

Apparently, this worked 2.5 years ago - the old viz still call that `onDone` with the very exact code from the issue: https://xstate.js.org/viz/?gist=812dc699e7161ba31887af2d19e7b997 . It also loops infinitely for some reason because both services (in parallel states) are constantly being restarted~., oops 🤫

Andarist: I've tweaked that test case [here](https://github.com/statelyai/xstate/pull/2909) and rechecked with the `main` branch - it fails there.

packages/core/src/StateNode.ts
Andarist: Just a reminder about this comment that blocks merging this PR. If you want I can add failing tests for those tomorrow and just patch those lines as those are really one-liners.

davidkpiano: Added here: 9ad6b8fd7f09cc70cc1e377628f27c83aff835fc

Andarist: The first described scenario was not fixed in this commit, I've added a failing test case and fixed this issue in: https://github.com/statelyai/xstate/pull/2915

Andarist (approved):

PR #2869, opened 2021-12-09
[v5] Remove batching

davidkpiano:
This PR removes batching and (somewhat related) the scheduler - refactored to be a simple mailbox.


packages/core/src/interpreter.ts
```diff
@@ -112,7 +110,9 @@ export class Interpreter<
   public options: Readonly<InterpreterOptions>;
 
   public id: string;
-  private scheduler: Scheduler = new Scheduler();
+  // private scheduler: Scheduler = new Scheduler();
```
tivac, line 113: Can this be removed?

Andarist:
```suggestion
```

packages/core/test/event.test.ts
```diff
@@ -67,9 +69,14 @@ describe('SCXML events', () => {
       }
     });
 
-    const service = interpret(machine).start();
-
-    expect(service.state.context.childOrigin).toBe('callback_child');
+    interpret(machine)
```
Andarist, line 72: q: did anything change here that this test has to be asynchronous right now?

packages/core/src/interpreter.ts
```diff
@@ -409,16 +406,33 @@ export class Interpreter<
...
-        this.attachDevTools();
-      }
-    });
     return this;
   }
 
+  private flush() {
+    this.mailboxStatus = 'processing';
+    while (this.mailbox.length) {
+      const event = this.mailbox.shift()!;
```
Andarist, line 425: I wonder if we should implement a better data structure for the mailbox, shifting arrays is not the best for performance

packages/core/src/interpreter.ts
```diff
@@ -409,16 +406,33 @@ export class Interpreter<
...
+
+      this.forward(event);
+
+      const nextState = this.nextState(event);
+
+      this.update(nextState);
```
Andarist, line 431: I'm a little bit worried about potential errors coming from those calls, this goes a little bit into the overall error handling story and we probably should not try to handle this anyhow as part of this PR. This is a crucial part of the implementation though so I would add a TODO comment to revisit this once we decide what to do about errors

packages/core/src/interpreter.ts
```diff
@@ -409,16 +406,33 @@ export class Interpreter<
...
 
-      if (this.options.devTools) {
-        this.attachDevTools();
-      }
-    });
     return this;
   }
 
+  private flush() {
+    this.mailboxStatus = 'processing';
```
Andarist, line 423: i wonder - is this status required? couldn't it be just derived from the length of the mailbox?

packages/core/src/interpreter.ts
```diff
@@ -409,16 +406,33 @@ export class Interpreter<
             State.from(initialState, this.machine.context)
           );
 
-    this.scheduler.initialize(() => {
-      this.update(resolvedState, initEvent as SCXML.Event<TEvent>);
```
Andarist, line 413: i suspect that the `initEvent` is still a thing, right? where is it handled now?

packages/core/src/interpreter.ts
davidkpiano: Using a linked list would make dequeuing O(1) instead of O(n), at the cost of more memory 🤔

packages/core/src/interpreter.ts
davidkpiano:
```suggestion
      // TODO: handle errors
      this.forward(event);

      const nextState = this.nextState(event);

      this.update(nextState);
```

packages/core/src/interpreter.ts
davidkpiano: I don't think it can be easily derived because we need to prevent the race condition that sending an event causes another event to be fired, which would flush the mailbox again, even though it's already being flushed. In that case, the mailbox length doesn't matter.

packages/core/src/interpreter.ts
davidkpiano: It comes from here:

```ts
export function resolveMicroTransition<
  TEvent extends EventObject
>(
  machine: StateMachine<TContext, TEvent>,
  transitions: Transitions<TContext, TEvent>,
  currentState: State<TContext, TEvent, any>,
  _event: SCXML.Event<TEvent> = initEvent as SCXML.Event<TEvent> // <<< here
): State<TContext, TEvent, any> {
```

The machine has this:
```ts
    const nextState = resolveMicroTransition(
      this,
      [],
      this.preInitialState,
      undefined // <<< resolves to initEvent
    );
```

packages/core/src/interpreter.ts
davidkpiano: Added to #2870

packages/core/test/event.test.ts
davidkpiano: Nope, just a refactoring to match how other services are "idiomatically" tested.

Although, we _should_ discourage reading `service.state` directly, and maybe disallow it in favor of `service.getSnapshot()`

packages/core/src/StateMachine.ts
```diff
@@ -240,32 +240,42 @@ export class StateMachine<
     return transitionNode(this.root, state.value, state, _event) || [];
   }
 
-  public get first(): State<TContext, TEvent> {
-    const pseudoinitial = this.resolveState(
+  private get preInitialState(): State<TContext, TEvent> {
```
Andarist, line 243: q: how do we define a "pre initial state"? could you add a comment above this method that would explain this?

packages/core/test/event.test.ts
Andarist: > Nope, just a refactoring to match how other services are "idiomatically" tested.

I agree that this might be "idiomatically" correct, however, I've found out that writing "non-idiomatic" tests (aka tests executed synchronously) can reveal more edge cases while writing those tests. It's because some problems only manifest themselves when we deal with synchronous execution, ping-ponging between different actors - this is true for XState, Redux-Saga, and even RxJS because those problems can be classified as scheduling problems and all of those libs deal with this and try to manage the complexity involved with communicating different "processes" etc

packages/core/src/StateMachine.ts
davidkpiano: Added a comment. This is the initial state _before_ evaluating any microsteps. The `preInitialState` may differ from the `initialState` e.g. if the `preInitialState` has an eventless transition to a different state.

packages/core/src/Mailbox.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!event) {
+      return undefined;
+    }
+
+    this.index++;
+
+    if (this.index > this.events.length - 1) {
+      this.events.length = 0;
+      this.index = 0;
+    }
```
Andarist, line 32: This strategy doesn't scale because we retain all the queued items up until we drain the queue. The items that were already processed should be immediately disposed because we don't have any use for them from that point in time.

I've reimplemented this mailbox using a simple linked list to avoid this problem altogether:
https://github.com/statelyai/xstate/pull/2916/commits/37547c356cf5f6fe5e073c9c5e1df57e4ddcb43b

packages/core/src/interpreter.ts
```diff
@@ -399,16 +395,35 @@ export class Interpreter<
...
-      if (this.options.devTools) {
-        this.attachDevTools();
-      }
-    });
     return this;
   }
 
+  private flush() {
+    this.mailbox.status = 'processing';
+    let event = this.mailbox.dequeue();
```
Andarist, line 413: This implementation doesn't scale too well with multiple consumers of the mailbox because a lot of the delicate queueing/dequeuing management becomes the responsibility of each consumer. I think I've managed to avoid this in this PR:
https://github.com/statelyai/xstate/pull/2916/commits/37547c356cf5f6fe5e073c9c5e1df57e4ddcb43b

With these changes, the consumer only needs to provide a "process"  callback to the created mailbox and call simple `start`, `clear` and `enqueue` methods. Thanks to that I've been able to reuse easily this Mailbox implementation in other 2 places that were managing this:
https://github.com/statelyai/xstate/pull/2916/commits/00f1d74d13d8105c7ebb52e771a82098b327ba79
https://github.com/statelyai/xstate/pull/2916/commits/e530b5b94c147e68fe3d51eea38d83469d8d49ac

Andarist (approved):

PR #2871, opened 2021-12-11
docs: Simplified Chinese Version

lecepin:
Simplified Chinese version of the document.

image


davidkpiano:
Thank you for this, @lecepin! We will be reviewing this soon; it's probably a good idea for us to find someone else who also knows simplified Chinese so we can have multiple people reviewing it.

laurakalbag:
Thanks for all your hardwork, @lecepin. I don’t read the language, but have otherwise reviewed the content with the assistance of a translation engine and really appreciate your hard work. This is ready to merge, I’m just going to add an update in the docs updates to direct people to this translation.

PR #2875, opened 2021-12-12
[v5] Remove strict mode

davidkpiano:
This PR removes `strict: true` behavior.


.changeset/breezy-boxes-press.md
```diff
@@ -0,0 +1,19 @@
+---
+'xstate': major
+---
+
+Strict mode is no longer available. Instead, wildcard transitions should be used to explicitly handle undesired events:
```
Andarist, line 5: This sort of has already been removed from the `next` branch but it might not be obvious to users and since this changeset mentions "strict mode" being removed I think it's worth mentioning that here too. In v4 the `.strict` property has also controlled the behavior related to rejected promise actors:
https://github.com/statelyai/xstate/blob/80bfe1542f1d10261698562c80954aa8bad81246/packages/core/src/interpreter.ts#L1058

davidkpiano: > This sort of has already been removed from the `next` branch but it might not be obvious to users and since this changeset mentions "strict mode" being removed I think it's worth mentioning that here too.

Sorry, what should be mentioned exactly in the changeset?

Andarist, line 5: Note that this might require some better wording/grammar:
```suggestion
Strict mode is no longer available. You can no longer control if a machine gets stopped or not when receiving unhandled rejection events from children, nor you can control if a machine throws an error when receiving events with unknown types. Instead of the latter, wildcard transitions should be used to explicitly handle undesired events:
```

davidkpiano:
Superseded by https://github.com/statelyai/xstate/pull/3943

PR #2876, opened 2021-12-12
[v5] Remove `Typestate`

davidkpiano:
This PR removes `Typestate` typings from XState core.


tonivj5:
Hey, I'm starting to use xstate on backend and I'm recently discovered TypeStates. I was looking for some way to define a concrete state of the context on an determined state point... Seeing this PR, I wonder what is the new way to do it?

Andarist:
> I wonder what is the new way to do it?

There is no new way to do it. What is being removed in this PR wasn't quite type-safe and we have to think about other ways to achieve this, this time in a type-safe manner.

packages/xstate-react/test/types.test.tsx
```diff
@@ -73,11 +73,11 @@ describe('useMachine', () => {
     type: 'YES';
   }
 
-  type YesNoTypestate =
-    | { value: 'no'; context: { value: undefined } }
-    | { value: 'yes'; context: { value: number } };
+  // type YesNoTypestate =
+  //   | { value: 'no'; context: { value: undefined } }
+  //   | { value: 'yes'; context: { value: number } };
```
Andarist, line 79:
```suggestion
```

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -292,17 +292,17 @@ describe('useMachine hook', () => {
...
+    // type TestState =
+    //   | {
+    //       value: 'loading';
+    //       context: { count: number; user: undefined };
+    //     }
+    //   | {
+    //       value: 'loaded';
+    //       context: { user: { name: string } };
+    //     };
+
```
Andarist, line 304:
```suggestion
```

Andarist (approved):

davidkpiano:
> Hey, I'm starting to use xstate on backend and I'm recently discovered TypeStates. I was looking for some way to define a concrete state of the context on an determined state point... Seeing this PR, I wonder what is the new way to do it?

Like @Andarist said, we're currently figuring it out. One idea is to use a design-by-contract approach so that you can _guarantee_ that certain conditions hold in a state:

```js
// idea
states: {
  loading: {
    invoke: {
      src: 'loadUser',
      onDone: { target: 'loaded', actions: assign({ user: (_, e) => e.data }) }
    }
  },
  loaded: {
    // precondition - assert this is true whenever entering this state
    pre: (ctx) => !!ctx.user // in this state, the `user` should always exist
    // ...
  }
}
```

Then, a typegen can derive that in `loaded`, `ctx.user` will exist.

Issue #414, opened 2019-04-07
Array conditions


Andarist:
## Bug or feature request?

Feature request

## Description:

Would be neat if `cond` could accept array of guards, without this I have to make multiple transient states when my logic is more complex.

## (Feature) Potential implementation:

It's not a breaking change and the API would be something like this:
```js
'': [
    { target: 'foo', cond: ['isOnline', 'hasForm'] }
]
```

davidkpiano:
This seems like it would work well with [custom guards 📘](https://xstate.js.org/docs/guides/guards.html#custom-guards) - the tricky thing with "just allowing arrays" is that it's ambiguous whether it means `isOnline && hasForm` or `isOnline || hasForm`. But you can easily do this:

```js
const every = (...guards) => ({
  type: 'every',
  guards
});

const guards = {
  isOnline: ctx => ctx.status === 'online',
  hasForm: ctx => ctx.user.hasForm === true
};

// ...
'': [
  {
    target: 'foo',
    cond: every('isOnline', 'hasForm')
    // => { type: 'every', guards: ['isOnline', 'hasForm'] }
  }
]

// ...
// machine options
{
  guards: {
    ...guards,
    every: (ctx, event, { guard }) => {
      const { guards } = guard;
      return guards.every(guardKey => guards[guardKey](ctx, event));
    }
  }
}
```

What do you think about that for a solution? Much more flexible than having XState be opinionated (and potentially too magical/black-box like) about custom guards.

Andarist:
I don't have much problem with the proposed solution, I've tackled the problem at hand in a similar fashion at the moment. My reasoning behind requesting this was that basically OR might be already represented by:
```
'': [ { target: 'foo', guard: 'a' }, { target: 'foo', guard: 'b' } ]
```
and that custom guards are harder to serialize.

davidkpiano:
I'll consider it. Good to look at prior art too: https://states-language.net/spec.html#choice-state

ooloth:
This feature would be really nice. An array of two strings (e.g. `cond: ['isOnline', 'hasForm']`) would be much easier to read and write than custom guards are.

I think an array would work as an assumed `&&` since the `||` alternative can already be implemented using an array of transitions (as @Andarist mentioned) and the pattern already exists with the `actions` arrays (i.e. every function in the array is fired).

🤞

davidkpiano:
[...]

PR #2879, opened 2021-12-16
[core] Multiple guards

davidkpiano:
This PR introduces the ability to specify guards in an array, where all of them must evaluate to `true` ("and" behavior) for the transition to be taken:

```js
// ...
on: {
  WITHDRAW: {
    cond: ['isLoggedIn', 'hasMoney'],
    target: 'withdrawn'
  }
}
// ...
```

### DIFF
--- a/.changeset/empty-clocks-rest.md
+++ b/.changeset/empty-clocks-rest.md
@@ -12,4 +12,5 @@ on: {
     target: 'withdrawn'
   }
 }
+// ...
 ```
--- a/.changeset/empty-clocks-rest.md
+++ b/.changeset/empty-clocks-rest.md
@@ -1,0 +1,15 @@
+---
+'xstate': minor
+---
+
+An array of guards can now be specified in a transition's `cond: ...` property. All guards in the array must pass ("and" behavior) for the transition to be taken.
+
+```js
+// ...
+on: {
+  WITHDRAW: {
+    cond: ['isLoggedIn', 'hasMoney'],
+    target: 'withdrawn'
+  }
+}
+```
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -198,7 +198,7 @@ export type TransitionTargets<TContext> = Array<
 >;
 
 export interface TransitionConfig<TContext, TEvent extends EventObject> {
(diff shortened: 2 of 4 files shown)


.changeset/empty-clocks-rest.md
```diff
@@ -0,0 +1,16 @@
+---
+'xstate': minor
+---
+
+An array of guards can now be specified in a transition's `cond: ...` property. All guards in the array must pass ("and" behavior) for the transition to be taken.
```
Andarist, line 5: That was one of my first feature requests to XState (see https://github.com/statelyai/xstate/issues/414)! I think though that I agree with your past self, this is ambiguous because it's not obvious if this represents AND or OR. 

Since this is a somewhat addressed issue in v5 - is it worth introducing this in v4?

davidkpiano: One thing I realized is that OR can be represented by separate transitions:

```js
EVENT: [
  { target: 'next', cond: 'thisOne' },
  { target: 'next', cond: 'orThatOne' }
]
```

So maybe it's not ambiguous after all.

I'm still not 100% sure if this should be a core feature, as opposed to `and(...)`, but I put it up here so we can talk about it.

Andarist: > One thing I realized is that OR can be represented by separate transitions:

This is what I've initially responded in that old issue: https://github.com/statelyai/xstate/issues/414#issuecomment-480838103 🤣 

But you don't necessarily have them side by side to look at and compare so when looked at in isolation - this isn't that obvious. Both use arrays, the "nesting level" is different and this makes a difference - I'm somewhat worried that this ain't beginner-friendly. This might not be obvious even for seasoned devs, or a tired dev :P - how many times we have been looking at parenthesis or curly braces put at the wrong place? How many times we have been rechecking operator precedence in non-basic scenarios?

I'm not saying that I'm super opposed to this - I'm just raising concerns. Also - if we are introducing those "operands" in v5 I think this would get deprecared/removed there anyway, in favor of `and(['cond1', 'cond2'])`. So the question is - is it even worth it to introduce a temporary syntax?

PR #2881, opened 2021-12-19
[v5] Improve target resolution

### LINKED ISSUE #346: Root transitions are not working properly

### LINKED ISSUE #500: Static visualizer for Machine
## Feature

It is nice that the xstate has the visualizer app where we can paste the code, however, if we decide to use xstate, we would not be able to do it manually with every state machine in the app.

I wonder if you plan a visualization to a static file feature, something like the javascript-state-machine has: https://github.com/jakesgordon/javascript-state-machine/blob/master/docs/visualization.md

Ideally, we would like to generate static diagram files in our CI/CD pipeline as a part of build process and dump them to file(s) in sort of open format (such as DotGraph) so that they can be viewed later, at any point in time. E.g something like `dump (fsm, path, format = "dotgraph")`

Is this something that is being considered?

Regards,
Dmitry.

davidkpiano:
This PR improves target resolution:

- Targeting sibling nodes from the root is no longer valid, since the root node has no siblings

```diff
createMachine({
  id: 'direction',
  initial: 'left',
  states: {
    left: {},
    right: {}
  },
  on: {
-   LEFT_CLICK: 'left',
+   LEFT_CLICK: '.left'
  }
});
```

- Delimiters in state node IDs resolve to the state node from the path relative to the resolved state node ID

```js
createMachine({
  // ...
  on: {
    SOME_EVENT: '#foo.bar' // resolves to whatever.bar
  },
  states: {
    whatever: {
      id: 'foo',
      states: {
        bar: {} // resolves to this state node
      }
    }
  }
});
```

fixes #346


davidkpiano:
@Andarist Would this be a breaking change?

Andarist:
Gonna test this asap - I've tried to do so a few days ago but the test output is spammed with some unrelated stuff and this made playing with it a little bit harder. Gonna take a second look on Monday or so and also I will try to clean up that test output.

packages/core/test/id.test.ts
```diff
@@ -43,7 +43,7 @@ const idMachine = createMachine({
           }
         },
         dot: {
-          id: 'B.dot'
+          id: '.dot'
```
Andarist, line 46: that looks incorrect/confusing, I think that it should be completely forbidden to an `id` with dots in it (it doesn't matter if the dot is leading or just in the middle of it). It introduces ambiguity because it's not obvious if it should be interpreted as a literal or if it should be interpreted as a path chained from the first segment

davidkpiano: You're right; not sure how that snuck in here. Will review.

.changeset/loud-books-teach.md
```diff
@@ -0,0 +1,22 @@
+---
+'xstate': major
+---
+
+Target resolution improvements: delimiters in state node IDs now resolve to the state node from the path relative to the resolved state node ID:
```
Andarist, line 5: this isn't new as far as I can tell: 
https://codesandbox.io/s/dawn-brook-tbt0ht?file=/src/index.js

Andarist: 👋

davidkpiano: Removed

PR #2882, opened 2021-12-20
[v5] Remove `state.history`

davidkpiano:
This PR removes the `history` property from the `State` object, for these reasons:

- `state.history` breaks immutability from states since every penultimate `state.history` is deleted to prevent memory leaks
- `state.history` is _not used_ for the statechart history mechanism (`state.historyValue` is, and that is staying)
- Users can/should keep track of previous state(s) themselves

I'm still on the fence whether this should be completely removed or just renamed to `state.previous` (cc. @Andarist) - I'd like to get an idea of how useful accessing the previous state is from within the machine, which is the most compelling use-case to keep this feature.

_**Note:** This does not remove the history mechanism from XState machines._


packages/xstate-analytics/test/analytics.test.ts
```diff
@@ -81,29 +81,18 @@ describe('@xstate/analytics', () => {
...
-              "state": "{\\"value\\":\\"green\\",\\"context\\":{}}",
-              "weight": 0.3333333333333333,
-            },
-          },
-          "{\\"value\\":\\"green\\",\\"context\\":{}}": Object {
             "{\\"type\\":\\"TIMER\\"}": Object {
-              "count": 1,
-              "currentWeight": 1,
-              "relativeWeight": 1,
+              "count": 2,
```
Andarist, line 85: Hm, it's somewhat surprising that this has been affected anyhow by this PR. Do you know why that is?

Andarist:
> I'm still on the fence whether this should be completely removed or just renamed to state.previous (cc. @Andarist)

Renaming would have the same downsides as the current solution. I think it's safe to remove this completely and to figure out a better solution in the future **IF** we find any compelling use cases to provide access for this.

packages/xstate-analytics/test/analytics.test.ts
Andarist: 👋

davidkpiano: Disregard this, I want to rework `@xstate/analytics` (very old project) before releasing it.

davidkpiano: The reason this has been affected is because it was using `state.history`, but now we have to use something different (which I added).

Andarist (approved):

PR #2883, opened 2021-12-20
[core] Add support for handling communication errors

davidkpiano:
This PR adds support for handling communication errors when the targets of events sent between actors does not exist:

```js
// ...
entry: sendParent('SOME_EVENT'), // pretend the parent does not exist...
on: {
  'error.communication': {
    actions: (_, e) => { console.log(e.message); }  // this will happen instead of an error being thrown
  }
}
```

If there is no `'error.communication'` transition, then an error will be thrown (existing behavior).

### DIFF
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -8,7 +8,8 @@ import {
   MachineOptions,
   StateMachine,
   StateSchema,
-  Typestate
+  Typestate,
+  ErrorEventObject
 } from './types';
 
 /**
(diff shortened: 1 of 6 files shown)


huan:
Can we have `error.communication` defined as an enum or literal string so that we can import and use it with static typed?

i.e.:

```ts
enum XStateTypes {
  ErrorCommnication = 'error.communication'
}
```

```ts
import { XStateTypes } from 'xstate/types'

// ...
entry: sendParent('SOME_EVENT'), // pretend the parent does not exist...
on: {
  [XStateTypes.ErrorCommunication]: {
    actions: (_, e) => { console.log(e.message); }  // this will happen instead of an error being thrown
  }
}
```

Andarist:
Error handling might still require some work in v5 but I think that it will look different than what we have here so this PR can be closed for now.

PR #2888, opened 2021-12-23
[v5] `onError`

davidkpiano:


packages/core/src/interpreter.ts
```diff
@@ -215,10 +215,16 @@ export class Interpreter<
     // Attach session ID to state
     state._sessionid = this.sessionId;
 
+    // Execute actions
```
davidkpiano, line 218: Here, actions are executed _before_ settling on the next state. If an error occurs, we will transition from the _previous state_ on the `error.*` event.

packages/core/src/interpreter.ts
```diff
@@ -765,38 +783,18 @@ export class Interpreter<
...
-        return action.execute(state);
-      } catch (err) {
-        this.parent?.send({
-          type: 'xstate.error',
-          data: err
-        });
-
-        throw err;
-      }
+      return action.execute(state);
```
davidkpiano, line 786: Errors caught in `.update()` method instead

packages/core/src/interpreter.ts
```diff
@@ -765,38 +783,18 @@ export class Interpreter<
...
-        if (this.parent) {
-          this.parent.send({
-            type: 'xstate.error',
-            data: err
-          } as EventObject);
-        }
-
-        throw err;
-      }
+      return exec(state.context, _event.data, {
```
davidkpiano, line 793: Errors caught in `.update()` method instead

Andarist:
I think that for now, the changes from https://github.com/statelyai/xstate/pull/4145 are sufficient. We can revisit adding root-level or state-level `onError` in the future but usually one shouldn't need anything like that.

PR #2889, opened 2021-12-23
Add `assertEventType` helper

tom-sherman:
Adds a helper function that can be used as an invariant to narrow the type of events that are passed to actions, services, guards, etc.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -71,7 +71,8 @@ export {
   spawn,
   doneInvoke,
   createMachine,
-  createSchema
+  createSchema,
+  assertEventType
 };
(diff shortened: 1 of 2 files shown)


Andarist:
I don't mind introducing such a helper but I'm worried that it will give users a false sense of safety. After all - this **can** throw, how do you respond to the thrown error?

tom-sherman:
Good point, I wonder if that can be communicated (or is already communicated) in the name and API. I only see this useful when using typescript and in Dev mode.

tom-sherman:
Maybe an alternative could be a type predicate function, not as ergonomic though IMO

PR #2892, opened 2021-12-24
[core] Resolve state from a state config object

davidkpiano:
This PR makes it easier to restore state by allowing a state config object to be passed into `machine.resolveState(object)`:

```js
const stateConfig = JSON.parse(localStorage.get('some-state-key'));

const restoredState = someMachine.resolveState(stateConfig); // no need to convert to State instance
```


packages/core/src/State.ts
```diff
@@ -47,21 +47,21 @@ export function stateValuesEqual(
...
-  if (isString(state)) {
+export function isStateConfig<TContext, TEvent extends EventObject>(
+  state: any
+): state is StateConfig<TContext, TEvent> {
+  if (typeof state !== 'object' || state === null) {
     return false;
   }
 
-  return 'value' in state && 'history' in state;
+  return 'value' in state && '_sessionid' in state;
```
Andarist, line 57: `_sessionid` is not essential - cause it won't really be reused/used anyhow (IIRC). I see how this helps to check if something is a persisted state or not but maybe it would be better to check for something more "required". 

It's kinda a nit level so it doesn't matter.

packages/core/src/StateNode.ts
```diff
@@ -707,14 +708,25 @@ class StateNode<
...
+    const resolvedState =
+      state instanceof State
+        ? state
+        : isStateConfig(state)
+        ? State.create(state)
+        : undefined;
+
+    if (!resolvedState) {
+      throw new Error(`Cannot resolve state: invalid state object`);
+    }
```
Andarist, line 722: Those lines should kinda not be needed at all - the argument is typed and thus an assignment like this would already be exhaustive:
```js
const resolvedArg = state instanceof State ? state : State.create(state)
```

packages/core/src/State.ts
davidkpiano: `_event` perhaps?

Andarist: Works for me.

packages/core/src/StateNode.ts
davidkpiano: They are needed because even though the types are exhaustive, data coming in here is often from an untrustworthy source, and needs to be checked in runtime:

```js
const stateConfig = JSON.parse(localStorage.get('state')); // cannot guarantee that this is a valid state config!

machine.resolveState(stateConfig); // might fail, despite types not failing
```

packages/core/test/machine.test.ts
```diff
@@ -328,6 +328,62 @@ describe('machine', () => {
...
+      });
+
+      const barState = machine.transition(undefined, 'NEXT');
+
+      const jsonBarState = JSON.parse(JSON.stringify(barState));
+
+      expect(machine.resolveState(jsonBarState).matches('bar')).toBeTruthy();
+    });
+
+    it('should resolve from a state config object (interpreter)', (done) => {
```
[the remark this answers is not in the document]
davidkpiano: Only creating the state can be done synchronously; resolving in a started interpreter is tested "async" since it's more real-life.

davidkpiano:
Ready for re-review

packages/core/test/machine.test.ts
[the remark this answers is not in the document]
Andarist: Those are IMHO two separate tests:
- resolves the state synchronously
- tests that `onDone` gets called immediately when starting with a "done" state

A problem with putting too many things into a single test case is that those inner/extra requirements/assertions will be missed by the reader and can accidentally be removed during refactors. When I see this test I don't know that this is supposed to test that `onDone` call, it just feels like its implementation detail here.

packages/core/src/StateNode.ts
Andarist: The counter-point is that... this can probably still fail or result in an invalid/incomplete value despite those checks, so this is not buying us that much here.

Andarist (approved):

packages/core/test/machine.test.ts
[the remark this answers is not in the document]
davidkpiano: Renamed this test, since we have the "resolves the state synchronously" test above this one and the second is the one that's really being tested.

packages/core/src/StateNode.ts
davidkpiano: > The counter-point is that... this can probably still fail or result in an invalid/incomplete value despite those checks, so this is not buying us that much here.

Okay, removed

PR #2895, opened 2021-12-27
Fixed an issue with some exit handlers being executed more than once when stopping a machine

Andarist:


packages/core/src/StateNode.ts
```diff
@@ -991,13 +989,6 @@ class StateNode<
       }
     }
 
-    if (!transition.source) {
-      transition.exitSet = [];
-
-      // Ensure that root StateNode (machine) is entered
-      transition.entrySet.push(this);
-    }
```
Andarist, line 999: this was redundant as in this scenario (when there is no `transition.source` - which in turn means "when resolving the initial state") because the `exitSet` is already empty and nothing pushes to it AND `entrySet` is already resolved with `this` in it - so we were always adding a duplicate root to the list here

packages/core/src/StateNode.ts
```diff
@@ -673,21 +673,19 @@ class StateNode<
     }
 
     const subStateKeys = keys(stateValue);
-    const subStateNodes: Array<
-      StateNode<TContext, any, TEvent, TTypestate>
-    > = subStateKeys.map((subStateKey) => this.getStateNode(subStateKey));
```
Andarist, line 678: Similarly here - the logic that follows this is already pushing `this.getStateNode(subStateKey)` to the `subStateNodes` so this line here could be dropped to avoid duplicate entries being pushed to the final `subStateNodes`

packages/core/src/StateNode.ts
```diff
@@ -673,21 +673,19 @@ class StateNode<
...
 
-        return allSubStateNodes.concat(subStateNode);
-      }, [] as Array<StateNode<TContext, any, TEvent, TTypestate>>)
+    subStateNodes.push(
+      ...flatten(
+        subStateKeys.map((subStateKey) =>
+          this.getStateNode(subStateKey).getStateNodes(stateValue[subStateKey])
+        )
+      )
     );
```
Andarist, line 686: it's an equivalent to the reduce on the left - I've found it harder to read this reduce with concat so I've just refactored this to push+map

davidkpiano (approved):

PR #2896, opened 2021-12-27
[core] Ensure compound state have `initial` defined

### LINKED ISSUE #674: Always resolve state value to leaf nodes

davidkpiano:
This PR throws an error if compound states (states with `state: { ... }`) do not have an `initial` state defined.

Closes #674


Andarist (commented):
Isn't this a huge breaking change? Even some of our tests (not a lot of them though) had no initial states on some compound states and they were successfully used in those tests.

I also wonder if a config like this should be forbidden:
```js
createMachine({
    initial: 'a',
    states: {
        a: {
            on: {
                EV: '#foo'
            }
        },
        b: {
            b1: {
                id: 'foo'
            },
            b2: {}
        }
    }
})
```

In here we never enter `b` directly so, in a sense, it doesn't have to have an initial state defined on it - because the machine always "routes" directly to one of its descendants

davidkpiano:
Should we warn instead?

Andarist:
> It should not; this is a valid configuration, but currently you need to specify initial, which doesn't make sense... wondering what we should do here.

Maybe this should be a warning/error at the tooling level? like our CLI/VScode extension? In there we could perform an exhaustive analysis of all the existing transitions and "prove" that a compound state can be entered directly, and if it does then it needs an initial state to be configured on it.

Note that some APIs (like when resolving a partial state value) could throw (or warn in v4) when the initial state for a compound state is missing. This is because those APIs should resolve to a valid configuration if a state value is "partial" then the "leaf" compunt state encoded in it has to have the initial state configured. Otherwise, we won't be able to resolve its value.

So we can start warning about this even today, just maybe not at the state node creation time.

Andarist:
@davidkpiano any new thoughts about this?

PR #2898, opened 2021-12-28
[v5] Pass error to parent for errorListener

roggervalf:


packages/core/src/interpreter.ts
```diff
@@ -345,8 +345,12 @@ export class Interpreter<
         listener(errorEvent.data.data);
       });
     } else {
-      this.stop();
-      throw errorEvent.data.data;
+      if (this.parent) {
+        this.parent?.send(errorEvent);
```
Andarist, line 349: you have just checked if there is a parent so we can just access `send` method here directly - without using the optional chaining:
```suggestion
        this.parent.send(errorEvent);
```

packages/core/src/interpreter.ts
```diff
@@ -345,8 +345,12 @@ export class Interpreter<
         listener(errorEvent.data.data);
       });
     } else {
-      this.stop();
-      throw errorEvent.data.data;
+      if (this.parent) {
```
Andarist, line 348: I think it's important that this branch still stops the current machine - if I understand correctly this whole function can only be called from within `service.send` and is called when there are no explicit error transitions defined in the machine. In such a case I would expect the machine with the unhandled error to be stopped by the XState's runtime

packages/core/src/interpreter.ts
Andarist, line 349: is it guaranteed that this event will always have the `origin` set? it feels important for clearing up the child in the parent

packages/core/src/interpreter.ts
Andarist: More than that - I think it makes sense to perform this logic regardless of the `this.errorListeners`, even if there is an error listener we should still stop the current machine and forward the event to the parent (if there is any)

packages/core/src/interpreter.ts
```diff
@@ -486,6 +490,13 @@ export class Interpreter<
       return;
     }
 
+    if (((event as EventObject)?.type || '').includes('error')) {
+      if (this.parent) {
+        this.parent.send(event);
+        return;
+      }
+    }
```
Andarist, line 498: how this relates to the logic introduced within `handleErrorEvent`? It feels like we'd never get to that other logic if we return from here. Is the distinction that in here we might handle "non-scxml" error event?

packages/core/src/interpreter.ts
```diff
@@ -486,6 +490,13 @@ export class Interpreter<
       return;
     }
 
+    if (((event as EventObject)?.type || '').includes('error')) {
```
Andarist, line 493: it feels like this should be performed (if it needs to be done in the first place) after converting the `event` to `_event` (just a couple of lines below this)

We should also always test this with `startsWith`-like logic - because `some_error_event` should not be treated as an error. This logic is reserved for event types starting with `errror.`

packages/core/src/interpreter.ts
roggervalf: it sounds good to me, let me refactor it

packages/core/src/interpreter.ts
roggervalf: you are right, the logic introduced within handleErrorEvent is not needed with this change

packages/core/src/interpreter.ts
roggervalf: I refactored this part too, thank you

packages/core/src/interpreter.ts
roggervalf: this part is not needed anymore, it was introduced when developing so we should be good now

Andarist:
This was recently reworked as part of https://github.com/statelyai/xstate/pull/4145 . I also included tests cases similar to what we have here in https://github.com/statelyai/xstate/pull/4517

PR #2903, opened 2021-12-29
Fixed an issue with exit actions being called in random order when stopping a machine

Andarist:


.changeset/perfect-paws-explode.md
```diff
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Fixed an issue with some exit handlers being executed more than once when stopping a machine.
+Fixed an issue with some exit actions being executed more than once when stopping a machine.
```
Andarist, line 5: I've just tweaked this as usually we don't call those 'handlers" in the docs so this could be confusing for people

packages/core/src/interpreter.ts
```diff
@@ -516,11 +516,13 @@ export class Interpreter<
       return this;
     }
 
-    this.state.configuration.forEach((stateNode) => {
-      for (const action of stateNode.definition.exit) {
-        this.exec(action, this.state);
-      }
-    });
+    [...this.state.configuration]
+      .sort((a, b) => a.order - b.order)
```
Andarist, line 520: I've considered sorting this when we create States:
https://github.com/statelyai/xstate/blob/b6dde9075adb3bb3522b4b8f8eeb804d3221a527/packages/core/src/State.ts#L251
but this has resulted in some test failures. I think that all of them were either snapshot tests (so we could just update them) or were related to rehydration. When dealing with the latter it turns out that `config.configuration` can be undefined. Well, our tests say that this can be undefined but our types assume that this is a required property:
https://github.com/statelyai/xstate/blob/b6dde9075adb3bb3522b4b8f8eeb804d3221a527/packages/core/src/types.ts#L1247

I could just default this to an empty array but this would also potentially cover up another problem. The problem is that `state.configuration` might be invalid/incomplete. I think this is "patched" for most cases within our `resolveState`:
https://github.com/statelyai/xstate/blob/b6dde9075adb3bb3522b4b8f8eeb804d3221a527/packages/core/src/StateNode.ts#L709-L722
I've added test cases for this as part of this PR here:
https://github.com/statelyai/xstate/pull/2903/commits/c4781eb0b57a929c38d6d41faf719ee416647340

I've decided not to patch it and sort it in the constructor of the `State` for now though as that would incur the sorting cost for all created States (and we create them after each transition). By only sorting this here we make this operation "lazy" in a sense that it's only performed when it is actually needed

packages/core/test/actions.test.ts
```diff
@@ -1354,96 +1475,6 @@ describe('choose', () => {
 
     expect(service.state.context).toEqual({ answer: 42 });
   });
-
-  // https://github.com/davidkpiano/xstate/issues/1109
-  it('exit actions should be called when invoked machine reaches final state', (done) => {
```
Andarist, line 1359: I've moved those to the new describe blocks as those were nested in the `"choose"` block which didn't quite make sense for those tests.

packages/core/test/actions.test.ts
```diff
@@ -595,6 +595,127 @@ describe('entry/exit actions', () => {
...
+            }
+          }
+        }
+      });
+
+      interpret(machine).start().stop();
+      expect(actual).toEqual(['root', 'a', 'a1']);
+    });
+
+    it('should call exit handlers in document order when the service gets stopped', () => {
```
Andarist, line 692: This is the only test added in this file. It's sort of basic and acts as a regression test here. Its implementation depends on the fact that transitions "mess up" the `state.configuration` order and this is an implementation detail - I think I'm going to add in a sec another, similar, test for parallel states, just in case (I know it will pass right now)

Andarist: Added those tests for parallel regions here:
https://github.com/statelyai/xstate/pull/2903/commits/6b5a57c4e9a05f53500f8857dafdd598b4b7953d
I've created 3 very similar tests but each one of them is "messing" up the `state.configuration` in a different order so I've decided to include all of those cases

packages/core/src/interpreter.ts
Andarist: Actually... I've made a huge mistake at first and I've sorted those exit actions in the document order while this should be done in the reversed document order (just how it's done when sorting & calling exit actions based on the exitSet during regular transitions).

The `exitOrder` is defined [here](https://www.w3.org/TR/scxml/#Predicates) as follows:
> exitOrder  // Descendants precede ancestors, with reverse document order being used to break ties 
    (Note: since descendants follow ancestors, this is equivalent to reverse document order.)

So sorting this in the State's constructor wouldn't be good anyway because most likely a document order would be better suited there while we need the reversed document order here. But, of course, if that would be sorted in the documented order then we could just iterate in reverse rather than sorting it but the "lazy" argument still stands.

davidkpiano (approved):

PR #2905, opened 2021-12-30
[v5] Remove all usage of `Object.defineProperty`

davidkpiano:
This PR removes all usage of `Object.defineProperty`


packages/core/src/State.ts
```diff
@@ -190,13 +186,6 @@ export class State<
       ? new Set(config.tags)
       : new Set();
     this.machine = config.machine;
-
-    Object.defineProperty(this, 'nextEvents', {
-      enumerable: false,
-      get: () => {
-        return nextEvents(this.configuration);
-      }
-    });
```
Andarist, line 199: Just a note for now - these weird lines of code were introduced here for a reason. We need to ensure that this change doesn't introduce a regression for those issues:
https://github.com/statelyai/xstate/issues/511
https://github.com/statelyai/xstate/pull/1023

The former is probably not an issue because right now on the `next` branch we transpile using Babel and they have correct semantics for getters and the latter is probably not the issue because we still reference `this` and not `config`. However, a closer look is required to make sure that this change is OK.

davidkpiano: I'll let you make the judgment call on whether this is okay or not. Feel free to merge if you think it is.

davidkpiano: #1023 is no longer applicable because `.history` was removed.

PR #2913, opened 2022-01-02
docs: move "...crosswalkEast..." comment to correct line

dlcmh:


docs/guides/final.md
```diff
@@ -60,11 +60,11 @@ const lightMachine = createMachine({
               }
             },
             stop: {
+            // 'stop' is a final state node for 'crosswalkEast'
```
Andarist, line 63:
```suggestion
              // 'stop' is a final state node for 'crosswalkEast'
```

PR #2915, opened 2022-01-03
Andarist/fix done invoke events for direct machines

Andarist:


packages/core/test/invoke.test.ts
```diff
@@ -2749,35 +2749,56 @@ describe('invoke', () => {
   });
 
   it.each([
-    'someSrc',
-    createMachine({ id: 'someId' }),
-    () => () => {
-      /* ... */
-    }
-  ])('something', (arg) => {
```
Andarist, line 2757: Screenshot 2022-01-03 at 13 26 53

PR #2916, opened 2022-01-03
Andarist/mailbox unification

Andarist:


packages/core/src/actor.ts
```diff
@@ -160,46 +158,36 @@ export class Actor<TEvent extends EventObject, TEmitted> {
...
+        ? (event as LifecycleSignal)
+        : toSCXMLEvent(event as TEvent);
+
+    this.current = this.behavior.transition(
+      this.current,
+      typeof this.context._event.type !== 'string'
+        ? (this.context._event as LifecycleSignal)
+        : (this.context._event as SCXML.Event<TEvent>).data,
+      this.context
+    );
```
Andarist, line 190: This is quite ugly but I didn't want to touch this any further in this PR. Refining the type with TS was not possible here because apparently, TS has problems when it comes to distinguishing strings and symbols:
[TS playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBNIAcDKBLA5mAhgGwCoE8EBTGAXhiXwFsAjEbAOgDMQAnACgHIAPaTKIg3gJOASgDcAKElRCJAKIA3MjADeMWcQBcMTixCcYAXxgAfNRrk7ONTK0Mnz6zUR0uQTOFESoMOAsTG0gAmRMDYdiSgkLBEijpK0iie7O6ecQwuZKTknNCsKGBoYmqSMDBxZTAA9NUwAHoA-JJGSSlpFYqZcjAAhDm6+YXFoqXlleW1Dc2tkskwqXIend2BObkQ1HTYJapVEzV1TS1tCx0ZWf0bW-S7+4pVU8dGQA)

I've just found `isSignal` helper in the codebase and I could use it here - but OTOH I think that using symbols here is not buying as anything and it's just annoying to work with so I would like to propose adjusting those signals to use plain strings.

packages/core/test/actor.test.ts
```diff
@@ -33,6 +33,8 @@ import {
 } from '../src/behaviors';
 import { invokeMachine } from '../src/invoke';
 
+jest.setTimeout(1000);
+
```
Andarist, line 37:
```suggestion
```

packages/xstate-react/src/useSpawn.ts
```diff
@@ -16,5 +17,10 @@ export function useSpawn<TState, TEvent extends EventObject>(
     return spawnBehavior(behavior);
   });
 
+  useEffect(() => {
+    actorRef.start!();
+    return () => actorRef!.stop!();
+  });
```
Andarist, line 23: Top-level actorRefs are required to have `start` method because we need to start them in a "deferred" manner. We can't start them when rendering in the React world - this has to be done in an effect

packages/core/src/behaviors.ts
```diff
@@ -172,6 +161,12 @@ export function spawnBehavior<TEvent extends EventObject, TEmitted>(
         }
       };
     },
+    start() {
+      mailbox.start();
+    },
+    stop() {
+      mailbox.clear();
+    },
```
Andarist, line 169: I think there is an urgent need to define "primitive types" of XState. Quite frankly, I'm getting lost in that - there are a lot of similar, yet slightly different abstractions introduced in the codebase and it's hard to manage their responsibilities because they are not entirely well-defined. I think that I understand most of them but yet the implementation doesn't mirror what I assume and this is confusing.

packages/core/src/behaviors.ts
```diff
@@ -183,8 +178,6 @@ export function spawnBehavior<TEvent extends EventObject, TEmitted>(
     _event: null as any
   };
 
-  state = behavior.start ? behavior.start(actorCtx) : state;
```
Andarist, line 186: Same argument as [here](https://github.com/statelyai/xstate/pull/2916/files#r777570632). We can't execute `start` when rendering.

packages/core/src/behaviors.ts
```diff
@@ -172,6 +161,12 @@ export function spawnBehavior<TEvent extends EventObject, TEmitted>(
         }
       };
     },
+    start() {
+      mailbox.start();
```
Andarist, line 165: this most likely should also start a behavior or something, but I'm not sure about the relationships between different primitive types (as mentioned in another comment). All tests are passing at the moment - so at the very least this is under-specified and we should make sure those primitive types and communication protocols etc are well-defined and covered by tests.

packages/core/src/behaviors.ts
davidkpiano: Can you list exactly what's confusing?

davidkpiano:
We'll merge this into `v5/mailbox` and continue work from there.

packages/core/src/behaviors.ts
Andarist: Basically what we've discussed last week - atm behaviors are not so pure but they should be. Whatever responsibilities are managed by behaviors should be pulled into actors.

Issue #2923, opened 2022-01-06
`sendTo` not directly importable from xstate


devanfarrell:
**Description**
The `sendTo` action was introduced in `4.27.0` per the [release notes](https://github.com/statelyai/xstate/releases/tag/xstate%404.27.0) but is not usable without importing from `xstate/lib/actions`

**Expected Result**
`sendTo` should be defined in at least one of these scenarios.

``` js
import { actions } from 'xstate';
const { sendTo } = actions;
```
``` js
import { sendTo } from 'xstate';
```

**Actual Result**
`sendTo` is undefined when imported as done above.

**Additional context**
Version `4.27.0`

The below does work but I believe it was omitted as an oversight.
```js
import { sendTo } from 'xstate/lib/actions'
```

Likely the change needs to be made here. https://github.com/statelyai/xstate/blob/a26897174a85672e70cba788a717c95b53b26235/packages/core/src/index.ts#L7-L25

devanfarrell:
I'm willing to take this on. Only question is if it should be accessible in the former or both access patterns?
``` js
import { actions } from 'xstate'
actions.sendTo()
```
``` js
import { sendTo } from 'xstate'
sendTo()
```

davidkpiano:
> I'm willing to take this on. Only question is if it should be accessible in the former or both access patterns?
> 
> ```js
> import { actions } from 'xstate'
> actions.sendTo()
> ```
> 
> ```js
> import { sendTo } from 'xstate'
> sendTo()
> ```

Would love a PR for this. I'd choose the first pattern `import { actions } from 'xstate'`

devanfarrell:
Verified fixed in [4.28.0](https://github.com/statelyai/xstate/releases/tag/xstate%404.28.0)

PR #2925, opened 2022-01-07
Added sendTo to the rest of the exported actions

devanfarrell:
Resolves issue https://github.com/statelyai/xstate/issues/2923

Contributing guidelines requested a test but none was added due to there not being an existing pattern to cleanly do so.


Andarist:
Ye, let's skip over adding a test for this - it's not worth it. Could you add a changeset though?

devanfarrell:
@Andarist The changeset tool thinks this is a major version bump. Feels like it should be a patch but I don't want to mess with the flow of anything without some direction.

.changeset/forty-turtles-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': major
```
Andarist, line 2:
```suggestion
'xstate': patch
```

.changeset/forty-turtles-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The `sendTo(actorRef, event)` action creator introduced in `4.27.0` was not accessible from the package exports.
```
davidkpiano, line 5:
```suggestion
The `sendTo(actorRef, event)` action creator introduced in `4.27.0`, which was not accessible from the package exports, can now be imported just like other actions:

```js
import { sendTo } from 'xstate';
```
```

Comment (Andarist) [reply .changeset/forty-turtles-protect.md:5]: @davidkpiano this should be more like:
```suggestion
The `sendTo(actorRef, event)` action creator introduced in `4.27.0`, which was not accessible from the package exports, can now be used just like other actions:

```js
import { actions } from 'xstate';
actions.sendTo
```
```

Andarist: This is what you have requested as a preferred option [here](https://github.com/statelyai/xstate/issues/2923#issuecomment-1007035960)

Note that applying this suggestion might be slightly unpredictable as it contains a triple backtick at the end and I'm unsure how to "escape" this properly here. It would be better to apply this manually.

Andarist (approved):

Andarist:
> @Andarist The changeset tool thinks this is a major version bump. Feels like it should be a patch but I don't want to mess with the flow of anything without some direction.

You had to accidentally accept a major bump type when creating changeset file through the CLI. The CLI itself is just a convenience tool and changeset files can be created/modified manually so I've just patched this here:
https://github.com/statelyai/xstate/pull/2925/commits/f8d89576e8ad8274622ea41d202f9c8ea9042f07

Thank you for you contribution!

Issue #2919, opened 2022-01-05
Testing `onError` transition using Jest surfaces unexpected console.error 


mindy:
## Description
When trying to test the `onError` transition logic using Jest, it surfaced a console.error stating that the `onError` handler was missing, even though it was defined in the machine. However, adding a `target` property fixes the error: 

```
onError: {
  // target: "fetchComplete" < adding a target fixes error
  actions: escalate((_, event) => event)
}
``` 

The error only appears during testing and does not appear when interacting with the machine in the browser. 

## Expected Result
Should not see any console errors when testing the `onError` transition without a `target` property. 

## Actual Result
```
  console.error
    Missing onError handler for invocation 'apiRequest', error was 'removal failed'.

      at console.error (node_modules/@testing-library/react/dist/act-compat.js:52:34)
      at Object.reportUnhandledExceptionOnInvocation (node_modules/xstate/lib/utils.js:535:15)
      at node_modules/xstate/lib/interpreter.js:1016:17
```

## Reproduction
https://codesandbox.io/s/festive-star-juruk?file=/src/index.test.ts

_Note: There's an [issue](https://github.com/codesandbox/codesandbox-client/issues/513) with using `jest.mock` in Codesandbox, so running the tests in browser doesn't work._ 

## Additional context
- Discord link: https://discord.com/channels/795785288994652170/809564635614150686/928385438652186674 
- Relevant packages and versions: 
```
"xstate": "^4.23.1",
 "@xstate/react": "^1.5.1",
"jest": "^26.0.0",
"jest-dom": "^3.5.0",
```

Andarist:
Thank your for the report and the repro case. From the description it looks like a bug but given that escalate is also used here - maybe there is some parent & child relationship that makes a difference here. Gonna examine this carefully asap

Andarist:
This is almost the same problem as the one described [here](https://github.com/statelyai/xstate/issues/2920#issuecomment-1006778813). You have mocked `escalate` and when that function is called it returns `undefined` but it should return something resembling an action (like a function).

The logic responsible for throwing the error that you are observing is based on this:
https://github.com/statelyai/xstate/blob/9f9184a534c9f49e9e76a75d8113aa4d0c812f2d/packages/core/src/StateNode.ts#L1797-L1807
And since your `actions: escalate()` is really an equivalent to `actions: undefined` then the normalized `transitions.actions` is just an empty array and as such your targetless `onError` is not treated as a transition that does something meaningful.

[...]

PR #2935, opened 2022-01-11
[do not merge] Allow wildcard transitions to act as "error handlers"

Andarist:
While investigating https://github.com/statelyai/xstate/issues/2919 I've noticed some things and played around with minor code adjustments.


packages/core/test/invoke.test.ts
```diff
@@ -1784,44 +1782,48 @@ describe('invoke', () => {
...
-                invoke: {
-                  src: () => () => {
-                    throw new Error('test');
-                  },
-                  onError: {
-                    target: 'failed',
-                    cond: () => {
-                      errorHandlersCalled++;
-                      return false;
+                initial: 'waiting',
```
Andarist, line 1785: this test was weirdly written so I've refactored it in the process, this should land on the main branch regardless of this PR - I will create a branch with just this simple adjustment later on

packages/core/src/interpreter.ts
```diff
@@ -718,9 +718,7 @@ export class Interpreter<
 
     if (
       _event.name.indexOf(actionTypes.errorPlatform) === 0 &&
-      !this.state.nextEvents.some(
-        (nextEvent) => nextEvent.indexOf(actionTypes.errorPlatform) === 0
-      )
+      !this.state.can(_event as any)
```
Andarist, line 721: Note that both the current logic (the one on the main branch) and the proposed logic in this PR sound incorrect to me, regardless of the decisions that have to be made in response to [the other comment](https://github.com/statelyai/xstate/pull/2935/files#r781981212).

Incorrect scenarios are really edge-case-y though and applicable more in v5 context than in v4. This whole if statement might be also dropped entirely as part of the work resulting from [The Errors RFC](https://github.com/statelyai/rfcs/pull/4) so it's not worth addressing this now.

With the current logic, based on `nextEvents`, this won't quite work correctly when we consider parallel regions because a single `onError` in one of the regions "turns off" this if statement for another parallel region.

The proposed logic feels incorrect because we can create such structure:
```js
on: {
  [errorDescriptor]: { /* do smth */ },
},
states: {
  a: {
    invoke: {
      src: 'fetch',
      onError: undefined
    }
  }
}
```
As we might see here - the inner error handler is forbidden but the outer one is not (the latter doesn't quite matter but maybe other edge cases can be created using such setup). This would return false from `state.can(errorEvent)` when in the `a` state but the `onError: undefined` looks explicit enough to count this as an "explicit" error handler. It feels that it would make sense to swallow errors with such a forbidden `onError` - note that it's not obvious how to express "error swallowing" differently. The only thing that comes to my mind is this:
```js
onError: { actions: noop }
```
And I wouldn't count this as intuitive.

packages/core/test/invoke.test.ts
```diff
@@ -2641,6 +2643,33 @@ describe('invoke', () => {
         })
         .start();
     });
+
+    it('a wildcard transition should be able to receive an error event', (done) => {
```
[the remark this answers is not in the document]
Andarist: Yeaaah, I think that `xstate.error.*` events should be treated in a special way. Wildcard transitions should not match them.

packages/core/src/interpreter.ts
Andarist: I added v5 essential label to this PR just to remember to incorporate thoughts from this comment above. The PR itself has to be redone.

davidkpiano:
Let's prefer `onError` on state?

davidkpiano:
Decision:
- Do not allow built-in events to be captured by wildcards
- Do not allow errors to be captured by wildcards (these are built-in events anyway)

davidkpiano:
🧹 We should have `onError` instead for states

PR #2939, opened 2022-01-13
Refactor `@xstate/react` to useSyncExternalStore

### LINKED ISSUE #2821: [@xstate/react] Leverage useSyncExternalStore() now that React 18 is in beta
**Description**
[React 18 beta announcement](https://github.com/reactwg/react-18/discussions/112)
[useSyncExternalStore() discussion](https://github.com/reactwg/react-18/discussions/86)
[Redux PR for React 18 compat](https://github.com/reduxjs/react-redux/pull/1808)

### LINKED ISSUE #3196: Bug: @xstate/react does not support react 18
### Description

When I try a fresh project with react 18.0.0 and install xstate it fails to resolve dependency tree.

### Expected result

`npm i xstate @xstate/react` gives no errors.

### Actual result

`npm i xstate @xstate/react`

Gives the following error:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: frontend@0.1.0
npm ERR! Found: react@18.0.0
npm ERR! node_modules/react
npm ERR!   react@"18.0.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.8.0 || ^17.0.0" from @xstate/react@2.0.1
npm ERR! node_modules/@xstate/react
npm ERR!   @xstate/react@"*" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR! 
npm ERR! See /Users/nvortega/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/nvortega/.npm/_logs/2022-04-06T09_26_02_261Z-debug.log
```

### Reproduction

I cant because there is no way of installing it

### Additional context

_No response_

Andarist:
closes #3196
closes #2821


packages/core/src/interpreter.ts
```diff
@@ -525,6 +525,7 @@ export class Interpreter<
         child.stop();
       }
     });
+    this.children.clear();
```
Andarist, line 528: Children, while stopped, were being preserved in this map. So when doing this we could end up with some broken states~
```js
service.start()
service.stop()
service.start()
```

packages/core/src/interpreter.ts
```diff
@@ -534,6 +535,7 @@ export class Interpreter<
     this.scheduler.clear();
     this.initialized = false;
     this.status = InterpreterStatus.Stopped;
+    this._initialState = undefined;
```
Andarist, line 538: Atm the `service.initialState` is being cached:
https://github.com/statelyai/xstate/blob/4bb3a91f232599034a25159e89f5c82ce9b3776d/packages/core/src/interpreter.ts#L213-L222
and it should stay cached, otherwise the `useSyncExternalStore` complains about `getSnapshot` not yielding stable results when `service.status === InterpreterStatus.NotStarted`.

However, this has to be reset in the same scenario as mentioned here:
https://github.com/statelyai/xstate/pull/2939/files#r784159286

packages/xstate-react/package.json
```diff
@@ -73,8 +74,8 @@
     "jsdom": "^14.0.0",
     "jsdom-global": "^3.0.2",
     "lerna-alias": "3.0.3-0",
-    "react": "^16.12.0",
-    "react-dom": "^16.12.0",
+    "react": "18.0.0-rc.0-next-811634762-20220110",
+    "react-dom": "18.0.0-rc.0-next-811634762-20220110",
```
Andarist, line 78: Before landing this we should set up tests to run against both React 17 and React 18. It's totally doable - I just didn't bother to do it right now.

This should also only land **after** React 18 gets finally released.

packages/xstate-react/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { useMachine, asEffect, asLayoutEffect } from './useMachine';
```
Andarist, line 1: Those became hard to support after the changes. Maybe I could make them work after digging into this more but since we've wanted to remove those anyway... I've decided to just do it here.

It means that after landing this we should update the `@xstate/react`'s version

packages/xstate-react/src/useInterpret.ts
```diff
@@ -36,18 +35,15 @@ function toObserver<T>(
   };
 }
 
-export function useInterpret<
+export function useIdleInterpreter<
```
Andarist, line 38: Not the best name... but naming things is hard™ and naming some hooks is even harder 😅

packages/xstate-react/src/useMachine.ts
```diff
@@ -81,21 +76,36 @@ export function useMachine<
...
+    getSnapshot,
+    undefined,
+    identity,
+    isEqual
   );
 
-  const service = useInterpret(getMachine, options, listener);
+  useEffect(() => {
+    const rehydratedState = options.state;
+    service.start(
```
Andarist, line 101: After those changes, `useMachine` no longer depends on the `useInterpret`. `useInterpret` starts the machine in the layout effect but this one has to do that in the passive effect. Maybe we should change the behavior of the `useInterpret` too - but I've wanted to avoid changing too much now. I've also wanted to change this for now because how the listener/subscriber gets wired is now different and has to be managed here, in the `useMachine`, but it was previously managed by `useInterpret`. Technically, we could just not pass any subscriber to the `useInterpret` and just subscribe with `useSyncExternalStore` here... and maybe that is something that we should do.

packages/xstate-react/src/useMachine.ts
```diff
@@ -81,21 +76,36 @@ export function useMachine<
...
+      const { unsubscribe } = service.subscribe(handleStoreChange);
+      return unsubscribe;
     },
-    []
+    [service]
+  );
+  const storeSnapshot = useSyncExternalStoreWithSelector(
+    subscribe,
+    getSnapshot,
+    undefined,
```
Andarist, line 94: This is an argument slot for `getServerSnapshot`. I'm not yet sure how this should work... maybe we should just use the initialState for it? It somewhat doesn't matter for `useMachine` because nothing should be able to change its state between render and the commit phase but all the other hooks are also using the `useSyncExternalStore` stuff and the same cannot be said about them - so we should think how this should be handled there.

packages/xstate-react/src/useMachine.ts
```diff
@@ -81,21 +76,36 @@ export function useMachine<
...
+
+  const subscribe = useCallback(
+    (handleStoreChange) => {
+      const { unsubscribe } = service.subscribe(handleStoreChange);
+      return unsubscribe;
     },
-    []
+    [service]
+  );
+  const storeSnapshot = useSyncExternalStoreWithSelector(
```
Andarist, line 91: `useMachine` kinda shouldn't have to use `useSyncExternalStoreWithSelector` - it could use the simpler `useSyncExternalStore`, but to make this possible we'd have to have immutable states without `.changed`

packages/xstate-react/src/useInterpret.ts
```diff
@@ -93,15 +72,13 @@ export function useInterpret<TMachine extends AnyStateMachine>(
     context,
     guards,
     actions,
+    activities,
```
davidkpiano, line 75: Why do we add this back in?

packages/xstate-fsm/src/types.ts
```diff
@@ -11,6 +14,44 @@ export interface EventObject {
 
 export type InitEvent = { type: 'xstate.init' };
 
+export type AnyMachine = StateMachine.Machine<any, any, any>;
```
Andarist, line 17: I've added a bunch of helpers here that we already have in `xstate` itself. I wasn't sure if I should add those in the `StateMachine` namespace or if they should just live at the top of the module (like in XState)

packages/xstate-react/src/index.ts
davidkpiano: Are we still going to export `useMachine`?

packages/xstate-fsm/src/types.ts
```diff
@@ -11,6 +14,44 @@ export interface EventObject {
...
+    : never
+  : never;
+
+export type ServiceFrom<T> = ReturnTypeOrValue<T> extends infer R
+  ? R extends StateMachine.Machine<infer TContext, infer TEvent, infer TState>
+    ? StateMachine.Service<TContext, TEvent, TState>
+    : never
+  : never;
+
+export type MachineImplementationsFrom<TMachine extends AnyMachine> = {
```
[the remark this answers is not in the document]
davidkpiano: Implementations is good 👍

packages/xstate-react/index.v3.d.ts
```diff
@@ -1,45 +1,79 @@
...
-  StateMachine,
-  Typestate
+  StateFrom
 } from 'xstate';
-import { MaybeLazy, UseMachineOptions } from './lib/types';
+import { MaybeLazy, Prop, UseMachineOptions } from './lib/types';
 
 export * from './lib/index';
 
-export declare function useMachine<
```
Andarist, line 15: I've unified signature declarations that target `TS@<4`. It was a mistake to have different generic parameters for `TS@<4` and `TS@>=4` - as upgrading TS version shouldn't immediately break somebody's code. I've fixed the issue as this release will be a major release for `@xstate/react` anyway.

Those still need to be in a separate file since labeled tuple elements were introduced in TS 4. Those declarations here work almost the same as the ones for TS 4. I didn't realize before that I could just ditch the labels and have the rest to behave the same.

packages/xstate-react/src/fsm.ts
```diff
@@ -56,58 +56,54 @@ export function useMachine<
...
         options ? options : (stateMachine as any)._options
       )
-    ).start()
+    )
   );
 
   const [state, setState] = useState(() => getServiceState(service));
 
-  useEffect(() => {
+  useIsomorphicLayoutEffect(() => {
```
Andarist, line 64: This aligns the behavior with the same logic that we have for the XState itself. I will also add a test covering this

packages/xstate-react/src/fsm.ts
```diff
@@ -56,58 +56,54 @@ export function useMachine<
...
     if (options) {
       (service as any)._machine._options = options;
     }
   });
 
   useEffect(() => {
     service.subscribe(setState);
+    service.start();
```
Andarist, line 72: This aligns the behavior with the same logic that we have for the XState itself. I will also add a test covering this

packages/xstate-react/src/useInterpret.ts
Andarist: Those were handled before too ([here](https://github.com/statelyai/xstate/pull/2939/files#diff-3efb60b28006844ac5e7d12c3877125205eef496823208e94488eadb16792942L103)) - but the type-level plumbing has changed slightly here and we no longer need to access this property separately.

packages/xstate-react/src/useMachine.ts
```diff
@@ -91,28 +65,63 @@ export function useMachine<TMachine extends AnyStateMachine>(
...
+    return service.state;
+  }, [service]);
 
-  const [state, setState] = useState(() => {
-    const { initialState } = service.machine;
-    return (options.state
-      ? State.create(options.state as any)
-      : initialState) as StateFrom<TMachine>;
-  });
+  const isEqual = useCallback(
```
Andarist, line 80: I would love to ditch this in the future, it's kinda hacky (I don't even check the `_prevState`) but this would require some changes in XState itself

Andarist: I guess I've already mentioned this in other comment from Jan: https://github.com/statelyai/xstate/pull/2939/files#r784176656

packages/xstate-react/test/useActor.test.tsx
```diff
@@ -65,7 +63,8 @@ describe('useActor', () => {
     );
   });
 
-  it('invoked actor should be able to receive (deferred) events that it replays when active', (done) => {
+  // TODO: how deferred actors are currently handled is quite hacky, we should preserve actor identity for deferred actors
+  it.skip('invoked actor should be able to receive (deferred) events that it replays when active', (done) => {
```
Andarist, line 67: oops, I forgot to recheck this - I'll do that later today, but this is yet again a problem caused by deferred children :S

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -353,201 +342,6 @@ describe('useMachine hook', () => {
     render(<App />);
   });
 
-  it('should capture all actions', (done) => {
```
Andarist, line 356: those removed tests were about `asEffect`+`asLayoutEffect` and those are being removed in this PR

packages/xstate-react/src/index.ts
Andarist: Yeah, the export is still here :)

packages/xstate-react/src/useSelector.ts
```diff
@@ -26,62 +25,27 @@ export function useSelector<
...
-    if (currentChanged) {
-      // required so we don't cause a rerender by setting state (this could create infinite rerendering loop with inline selectors)
-      // at the same time we need to update the value within the subscription so new emits can compare against what has been returned to the user as current value
-      subscription.setCurrentValue(currentSelected);
-    }
-  });
-
-  return currentSelected;
+): T {
+  const subscribe = useCallback(
```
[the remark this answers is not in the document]
davidkpiano: Wow you're not lying! 💯

packages/xstate-react/package.json
Andarist: OTOH - I don't think now that this would add much value here. Stuff that works with React 18 should just work with React 17. So it might be enough to just test with React 18. We can always also revisit adding extra tests for React 17 in the future.

packages/xstate-react/src/useMachine.ts
Andarist: Actually, I've added a comment on why we can't use `useInterpret` here right now:
https://github.com/statelyai/xstate/pull/2939/commits/47f566cc83e06bac152cfd599a4bd6c786f1183b#diff-19f47c3f58ef51d8a79a32343f5c32c02d6d7237ecc29f636cb53846d00424d5R68-R69

packages/xstate-react/test/useActor.test.tsx
Andarist: Somewhat weirdly those tests now "just pass". In a way - that's good, in another way - I'm not yet 100% sure if this is not a false positive. I need to recheck this.

packages/core/src/interpreter.ts
```diff
@@ -587,15 +587,21 @@ export class Interpreter<
...
     });
+    this.children.clear();
 
     // Cancel all delayed events
     for (const key of Object.keys(this.delayedEventsMap)) {
       this.clock.clearTimeout(this.delayedEventsMap[key]);
     }
 
     this.scheduler.clear();
+    this.scheduler = new Scheduler({
```
Andarist, line 598: this assigns a fresh scheduler because the old one already has `this.initialized === true` and thus when we start to send events to a stopped machine the processing in the scheduler goes wrong (it process the event instead of queeing it). This happens because the "active" states of the interpreter and the scheduler are not "synced"~

packages/xstate-react/src/fsm.ts
```diff
@@ -56,58 +56,54 @@ export function useMachine<
...
-          if (state.changed !== false) {
-            currentState = state;
-            callback();
-          }
-        });
-        return unsubscribe;
-      }
-    };
-  }, [service]);
+  const isEqual = useCallback(
```
davidkpiano, line 87: Can this just be declared externally?

Andarist: ah, yes - this one here, for `/fsm`, doesn't require access to the local `service` - so it can be hoisted

good 👁️  👍

packages/xstate-fsm/src/types.ts
Andarist: I've now spotted that there is already `StateMachine.AnyState` - so I probably should move most of those types to this namespace. Thoughts?

packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -462,4 +390,150 @@ describe('interpreter', () => {
 
     expect(nextState.actions.map((a) => a.type)).toEqual(['action']);
   });
+
+  describe('`start` method', () => {
```
Andarist, line 394: this block has just been moved to the bottom as tests were sort of mixed and the order in the file didn't match the order in the test report

packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -462,4 +390,150 @@ describe('interpreter', () => {
...
+      const context = { hello: 'world' };
+      const service = interpret(machine).start({ value: 'bar', context });
+      expect(service.state.value).toBe('bar');
+      expect(service.state.context).toBe(context);
+
+      service.send('NEXT');
+      expect(service.state.matches('baz')).toBe(true);
+    });
+
+    it('should execute initial actions when re-starting a service', () => {
```
Andarist, line 465: this is a new test

packages/xstate-react/src/useMachine.ts
```diff
@@ -91,28 +65,67 @@ export function useMachine<TMachine extends AnyStateMachine>(
...
+
+      // Only change the current state if:
+      // - the incoming state is the "live" initial state (since it might have new actors)
+      // - OR the incoming state actually changed.
+      //
+      // The "live" initial state will have .changed === undefined.
+      const initialStateChanged =
+        nextState.changed === undefined &&
+        (Object.keys(nextState.children).length > 0 ||
+          typeof prevState.changed === 'boolean');
```
Andarist, line 96: this additional condition here allows us to "restart" to the initial state after Fast Refresh, whereas previously we were displaying the outdated state that was no longer true for the current, restarted, service

packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -462,4 +390,150 @@ describe('interpreter', () => {
...
+      service.stop();
+
+      entryActionCalled = false;
+
+      service.start();
+
+      expect(entryActionCalled).toBe(true);
+    });
+
+    it('should execute initial actions when re-starting a service that transitioned to a different state', () => {
```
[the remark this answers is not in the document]
davidkpiano: The behavior should be consistent with core

.changeset/rude-laws-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/react': major
+---
+
+The signatures of `useMachine` and `useService` integrating with `@xstate/fsm` were changed. They now only accept a single generic each (`TMachine` and `TService` respectively). This has been done to match their signatures with the related hooks that integrate with `xstate` itself.
```
davidkpiano, line 5: Can we remove `useService` in here, in favor of `useActor`? Or is there a reason we still need `useService`?

packages/core/src/interpreter.ts
davidkpiano: Should this instead be assigned in `this.start(...)`? Reason being that for use-cases outside of React, a stopped service should be stopped, and not allocate a new scheduler at all, since it should no longer receive events. Maybe a micro-optimization, but we should allocate this object in `.start()` instead of `.stop()`.

packages/xstate-fsm/src/types.ts
davidkpiano: For consistency, might be a good idea 👍

.changeset/rude-laws-battle.md
Andarist: There is no `useActor` for `@xstate/fsm` and, so far, we've been using "service" as the name for the thing returned from `interpret`. So this makes it specific for how it can actually be used today.

We can rename it though but at the type-level this only accepts `TService` now. At the runtime level, the internal `getSnapshot` is also tied to a service type. I think it might be better to leave it as-is for now and reconsider later.

packages/core/src/interpreter.ts
Andarist: We can't do that because we need to start buffering events sent to a "stopped" service (that we treat as not started) before the `.start()` is even called.

So we either create this here or we hack this stuff around and assign this private property from within `@xstate/react`'s code. In `@xstate/react` we could add it here:
https://github.com/statelyai/xstate/blob/360e8546298c4a06b6d51d8f12c0563672dd7acf/packages/xstate-react/src/useInterpret.ts#L143-L146

packages/xstate-fsm/src/types.ts
Andarist: I've moved `AnyMachine` and `AnyService`. Any thoughts about `*From` helpers? For now, I've left them at the top level. I think this is an acceptable "difference" and would match the XState exports shape more closely.

packages/xstate-react/test/useMachine-fsm.test.tsx
```diff
@@ -322,4 +314,122 @@ describe('useMachine hook for fsm', () => {
...
+      React.useLayoutEffect(() => {
+        send('EXEC_ACTION');
+      }, [value]);
+
+      return null;
+    };
+
+    const { rerender } = render(<App value={1} />);
+
+    expect(actual).toEqual(suiteKey === 'strict' ? [1, 1] : [1]);
```
davidkpiano, line 350: Sigh...

.changeset/hungry-dryers-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Fixed issues with not disposing some cached internal values when stopping interpreters. This could lead to issues when starting such an interpreter again.
```
davidkpiano, line 5:
```suggestion
Fixed issues with not disposing some cached internal values when stopping interpreters, which could have led to issues when starting such an interpreter again.
```

Rephrased to clarify that this change _fixes_ the potential issues, instead of introduces them 😅

davidkpiano (approved):
🚢 ⚛️ 1️⃣ 8️⃣

PR #2941, opened 2022-01-17
Typegen PR feedback

mattpocock:


packages/core/src/typegenTypes.ts
```diff
@@ -22,19 +22,74 @@ export interface TypegenEnabled {
   '@@xstate/typegen': true;
 }
 export interface TypegenMeta extends TypegenEnabled {
-  matchesStates: string;
+  /**
+   * Allows you to specify all the results of state.matches
+   */
+  matchesStates: string | {};
```
mattpocock, line 28: Added `{}` here so that we can pass `{ a: 'c' }` as the state.matches value

packages/core/test/typegenTypes.test.ts
```diff
@@ -317,6 +317,23 @@ describe('typegen types', () => {
     machine.initialState.matches('a');
   });
 
+  it('should allow valid object `matches`', () => {
```
mattpocock, line 320: Added tests for the `{}` syntax of state.matches

packages/core/test/typegenTypes.test.ts
```diff
@@ -338,6 +355,24 @@ describe('typegen types', () => {
     machine.initialState.matches('d');
   });
 
+  it('should allow invalid object `matches`', () => {
```
Andarist, line 358:
```suggestion
  it('should not allow invalid object `matches`', () => {
```

mattpocock: Doh

packages/core/src/typegenTypes.ts
Andarist: Hm, `{}` might be somewhat "loose" since it doesn't quite mean "an empty object"  but IIRC we don't really even use this `TypegenMeta` that much (we might only use this in tests). It's just like a declaration and is used more as a documentation than implementation.

Andarist (approved):

packages/core/src/typegenTypes.ts
mattpocock: Yeah I'm intentionally using the 'loose' aspect of it to allow us to pass anything resembling an object.

PR #2942, opened 2022-01-17
Add Semver Policy to the README

Andarist:


README.md
```diff
@@ -434,3 +434,13 @@ const previousState = paymentMachine.transition(reviewState, 'PREVIOUS').value;
...
+
+XState itself executes a good chunk of the user logic and as such almost any change to its behavior might be considered a breaking change. We recognize this as a potential problem but we believe that treating every change as a breaking one is not practical. When implementing new features we do our best to do that in a thoughtful manner, in a way that can enable our users to implement their logic in better and safer way. This, sometimes, can affect how existing XState machines behave if they are using particular configurations. This doesn't mean that we introduce behavior changes on a whim or that they will be introduced in a way that affects most of the existing machines. It just means that we reserve the right to do some behavior changes in minor releases and that such changes will always be dictated by our best judgement of the situation. Please always read our release notes before making a decision to upgrade.
```
mattpocock, line 442:
```suggestion
XState itself executes a good chunk of the user logic and as such almost any change to its behavior might be considered a breaking change. We recognize this as a potential problem but we believe that treating every change as a breaking one is not practical. When implementing new features we do our best to do so in a thoughtful manner, to enable our users to implement their logic in a better, safer way. This _can_ affect how existing XState machines behave if they are using particular configurations. This doesn't mean that we introduce behavior changes on a whim or that they will be introduced in a way that affects most of the existing machines. It means that we reserve the right to make _some_ behavior changes in minor releases and that such changes will always be dictated by our best judgement of the situation. Please always read our release notes before making a decision to upgrade.
```

README.md
```diff
@@ -434,3 +434,13 @@ const previousState = paymentMachine.transition(reviewState, 'PREVIOUS').value;
...
+
+At the same time we reserve a similar right to adjust declared TypeScript definitions or to drop support for older versions of TypeScript in a minor release. The language itself evolves so quickly and tends to introduce breaking changes in its minor releases. At the same time our team learns how to use TypeScript more effectively and in a better way. Both of those things mean that it becomes very impractical for our team to be bound by decisions taken when an older version of TypeScript was its latest version or when we just didn't know how to declare our types in a better way. This also doesn't mean that we'll introduce declaration changes often but the likelihood of that is also higher than when it comes to runtime changes.
```
mattpocock, line 444:
```suggestion
At the same time we reserve a similar right to adjust declared TypeScript definitions or to drop support for older versions of TypeScript in a minor release. The language itself evolves so quickly and tends to introduce breaking changes in its minor releases. At the same time, our team learns how to leverage TypeScript more effectively - and the types improve as a result. Both of those things mean that it becomes very impractical for our team to be bound by decisions taken when an older version of TypeScript was its latest version or when we just didn't know how to declare our types in a better way. This also doesn't mean that we'll introduce declaration changes often - but the likelihood of that is also higher than when it comes to runtime changes.
```

laurakalbag (approved):

README.md
```diff
@@ -434,3 +434,23 @@ const previousState = paymentMachine.transition(reviewState, 'PREVIOUS').value;
...
+
+Any change _could_ affect how existing XState machines behave if those machines are using particular configurations. We do not introduce behavior changes on a whim and aim to avoid making changes that affect most existing machines. But we reserve the right to make _some_ behavior changes in minor releases. Our best judgment of the situation will always dictate such changes. Please always read our release notes before deciding to upgrade.
+
+### TypeScript changes
+
+We also reserve a similar right to adjust declared TypeScript definitions or drop support for older versions of TypeScript in a minor release. The TypeScript language itself evolves quickly and often introduces breaking changes in its minor releases. Our team is also continuously learning how to leverage TypeScript more effectively - and the types improve as a result. 
+
+For these reasons, it is impractical for our team to be bound by decisions taken when an older version of TypeScript was its latest version or when we didn’t know how to declare our types in a better way. We won’t introduce declaration changes often - but we are more likely to do so with runtime changes.
```
Andarist, line 452:
```suggestion
For these reasons, it is impractical for our team to be bound by decisions taken when an older version of TypeScript was its latest version or when we didn’t know how to declare our types in a better way. We won’t introduce declaration changes often - but we are more likely to do so than with runtime changes.
```

Andarist:
@davidkpiano is this good to go?

davidkpiano (approved):

Issue #2565, opened 2021-08-26
A sendParent used in the entry action of the initial state of a lazily spawned machine crashes the process


riccardo-forina:
**Description**

A sendParent used in the entry action of the initial state of a lazily spawned machine crashes the process. When executing the machine in a Jest test, it causes high CPU usage and eventually a crash of the node process.

**Expected Result**
Should work.

**Actual Result**
Laptop gets hot 😂

**Reproduction**
A bunch of tests here, a few with the actual machine that caused the issue in my project and one minimal reproduction with just the sendParent setup I described.
https://github.com/riccardo-forina/xstate-jest-lazy-actors-high-cpu

**Additional context**

xstate 4.23.1

davidkpiano:
Is this still reproducible with the latest `xstate` and `@xstate/react` versions?

riccardo-forina:
I'm afraid so

```
~/Code/jest-high-cpu main*
❯ npm ls xstate
jest-ok@1.0.0 /Users/riccardoforina/Code/jest-high-cpu
├─┬ @xstate/react@1.6.3
│ └── xstate@4.27.0 deduped
└── xstate@4.27.0


~/Code/jest-high-cpu main*
❯ npm run test:high-cpu

> jest-ok@1.0.0 test:high-cpu
> jest src/HighCpu.test.tsx


 RUNS  src/HighCpu.test.tsx

<--- Last few GCs --->

[14420:0x150008000]    80918 ms: Mark-sweep 4006.9 (4134.4) -> 3992.6 (4135.9) MB, 2975.1 / 0.0 ms  (average mu = 0.107, current mu = 0.005) allocation failure scavenge might not succeed
[14420:0x150008000]    83947 ms: Mark-sweep 4008.7 (4135.9) -> 3994.5 (4137.7) MB, 3012.1 / 0.0 ms  (average mu = 0.059, current mu = 0.005) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x10057c46c node::Abort() [/var/folders/09/9bv34hm11vb94tmwhtqyyx880000gn/T/fnm_multishells/13931_1640939242744/bin/node]
[...]

PR #2943, opened 2022-01-17
Fixed an infinite loop when initially spawned actor responded synchronously to its parent

Andarist:
fixes https://github.com/statelyai/xstate/issues/2565

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -1129,7 +1129,12 @@ class StateNode<
 
     stateTransition.configuration = [...resolvedConfig];
 
-    return this.resolveTransition(stateTransition, currentState, _event);
+    return this.resolveTransition(
+      stateTransition,
+      currentState,
+      currentState.context,
+      _event
+    );
   }
 
   private resolveRaisedTransition(
@@ -1151,9 +1156,9 @@ class StateNode<
 
   private resolveTransition(
     stateTransition: StateTransition<TContext, TEvent>,
-    currentState?: State<TContext, TEvent, any, any>,
-    _event: SCXML.Event<TEvent> = initEvent as SCXML.Event<TEvent>,
-    context: TContext = this.machine.context
+    currentState: State<TContext, TEvent, any, any> | undefined,
+    context: TContext,
+    _event: SCXML.Event<TEvent> = initEvent as SCXML.Event<TEvent>
   ): State<TContext, TEvent, TStateSchema, TTypestate> {
     const { configuration } = stateTransition;
     // Transition will "apply" if:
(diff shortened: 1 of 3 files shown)


packages/core/src/StateNode.ts
```diff
@@ -1151,9 +1156,9 @@ class StateNode<
 
   private resolveTransition(
     stateTransition: StateTransition<TContext, TEvent>,
-    currentState?: State<TContext, TEvent, any, any>,
-    _event: SCXML.Event<TEvent> = initEvent as SCXML.Event<TEvent>,
-    context: TContext = this.machine.context
+    currentState: State<TContext, TEvent, any, any> | undefined,
+    context: TContext,
+    _event: SCXML.Event<TEvent> = initEvent as SCXML.Event<TEvent>
```
Andarist, line 1161: arguments were reordered to make `context` **required** and a required param can't follow an optional one (`_event`)

packages/core/test/actor.test.ts
```diff
@@ -498,6 +498,84 @@ describe('actors', () => {
...
+        },
+        end: { type: 'final' }
+      }
+    });
+    interpret(parent).start();
+    expect(spawnCounter).toBe(1);
+  });
+
+  // https://github.com/statelyai/xstate/issues/2565
+  it('should only spawn an initial actor once when it synchronously responds with an event', () => {
```
Andarist, line 542: this is the only new test here

packages/core/src/StateNode.ts
```diff
@@ -1151,9 +1156,9 @@ class StateNode<
 
   private resolveTransition(
     stateTransition: StateTransition<TContext, TEvent>,
-    currentState?: State<TContext, TEvent, any, any>,
-    _event: SCXML.Event<TEvent> = initEvent as SCXML.Event<TEvent>,
-    context: TContext = this.machine.context
```
Andarist, line 1156: this is the only line that has truly changed - the other stuff is just reordering etc

The reason behind the infinite loop was that this code was running within a "service scope" and thus it has been executing the `spawnMachine`. The problem was that this `this.machine.context` is actually a getter (yet one reason to hate getters for me :P) and it was executed almost **always** because the default value for the `context` parameter had to be assigned here. In the vast majority of use cases, we are only interested in `currentState.context` though - and we were chosing that a couple of lines below:
https://github.com/statelyai/xstate/blob/56842cb39110923afe70280dde744bea4f81d6dd/packages/core/src/StateNode.ts#L1174
but that was only happening **after** a default value for `context` was already assigned and the contained side-effect logic executed.

So new child actors were constantly created and their initial actions were constantly executed in response to `sendParent` that was contained in those initial actions - thus creating an infinite loop and a memory leak.

I've fixed this by getting rid of this default value and by just providing the appropriate context explicitly at 2 call sites of this function

packages/core/src/StateNode.ts
Andarist: oh, and this was OK to do because `resolveTransition` is a private method, so it's not a breaking change

davidkpiano (approved):

PR #2944, opened 2022-01-17
Added failing test with multiple state.matches

mattpocock:


packages/xstate-react/test/typegenTypes.test.tsx
```diff
@@ -423,4 +423,48 @@ describe('useInterpret', () => {
 
     render(<App />);
   });
+
+  it('Should handle multiple state.matches', () => {
```
mattpocock, line 427:
```suggestion
  it('Should handle multiple state.matches when passed TypegenMeta', () => {
```

packages/xstate-react/test/typegenTypes.test.tsx
```diff
@@ -423,4 +423,48 @@ describe('useInterpret', () => {
...
+      tsTypes: {} as TypesMeta
+    });
+
+    () => {
+      const [state] = useMachine(machine, {});
+      if (state.matches('a')) {
+        return <div>a</div>;
+      }
+
+      // matches should be defined, but isn't
```
mattpocock, line 448:
```suggestion
      // matches should still be defined
```

packages/core/src/State.ts
```diff
@@ -322,10 +322,12 @@ export class State<
...
    */
-  public matches<TSV extends TTypestate['value']>(
-    parentStateValue: TResolvedTypesMeta extends TypegenEnabled
+  public matches<
+    TSV extends TResolvedTypesMeta extends TypegenEnabled
       ? Prop<TResolvedTypesMeta, 'matchesStates'>
-      : TSV
+      : TTypestate['value']
+  >(
+    parentStateValue: TSV
```
Andarist, line 330: It seems that both of those patches could also lead us to fixing this:
<details>
<summary>Diff 1</summary>

```diff
diff --git a/packages/core/src/Machine.ts b/packages/core/src/Machine.ts
index 1db863345..2ec94ea9a 100644
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -94,7 +94,7 @@ export function createMachine<
   TContext,
   any,
   TEvent,
-  any,
+  { value: never; context: TContext },
   BaseActionObject,
   ResolveTypegenMeta<TTypesMeta, TEvent, BaseActionObject>
 >;
```
</details>
<details>
<summary>Diff 2</summary>

```diff
diff --git a/packages/core/src/Machine.ts b/packages/core/src/Machine.ts
index 1db863345..b1bebf1c3 100644
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -94,7 +94,7 @@ export function createMachine<
   TContext,
   any,
   TEvent,
-  any,
+  { value: ''; context: TContext },
   BaseActionObject,
   ResolveTypegenMeta<TTypesMeta, TEvent, BaseActionObject>
 >;
```
</details>

They both fix the problem for `createMachine` call within `xstate` directory but for some reason, the problem doesn't go away within `xstate-react` directory. I'm not sure why... maybe it's a caching problem or something is still off within `useMachine` type that loses this "fake" `TTypestate` generic.

Either way... we can probably roll with this and revisit later if we find any problems with the approach used in this PR

PR #2957, opened 2022-01-24
Update `package.json` repository links

davidkpiano:
The repository links have been updated from `github.com/davidkpiano` to `github.com/statelyai`.

### DIFF
--- a/.changeset/cuddly-hats-swim.md
+++ b/.changeset/cuddly-hats-swim.md
@@ -1,0 +1,15 @@
+---
+'xstate': patch
+'@xstate/analytics': patch
+'@xstate/fsm': patch
+'@xstate/graph': patch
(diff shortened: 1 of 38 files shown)


.changeset/cuddly-hats-swim.md
```diff
@@ -0,0 +1,15 @@
+---
```
Andarist, line 1: This PR doesn't need a changeset unless you believe that this should be released asap~. From my PoV this change can just be included in the next release without being mentioned explicitly.

Andarist (approved):

.changeset/cuddly-hats-swim.md
davidkpiano: I still think it's good to do so that the changes propagate up to `npmjs.com`.

PR #2962, opened 2022-01-25
Added t() to allow for simpler use of schema

mattpocock:


packages/core/src/schema.ts
```diff
@@ -1,3 +1,4 @@
 export function createSchema<T>(schema?: any): T {
```
davidkpiano, line 1:
```suggestion
/**
 * @deprecated Use `t<SomeType>()` instead.
 **/
export function createSchema<T>(schema?: any): T {
```

davidkpiano (approved):

packages/core/src/schema.ts
mattpocock: Will this work as intended? Since t is an alias it might rely on the createSchema TSDocs

davidkpiano: Maybe move the function definition to `t` instead?

This is nit-level but useful for folks using `createSchema`; can be done in a separate PR by the way

PR #2963, opened 2022-01-25
Added documentation for typegen & the VS Code extension

mattpocock:
This PR also changes the recommended TS approach from `createMachine<TContext, TEvent>()` to `schema`, and documents use of `t<{}>()` (as PR'd here: #2962)


docs/README.md
```diff
@@ -16,6 +16,8 @@ JavaScript and TypeScript [finite state machines](https://en.wikipedia.org/wiki/
 
 New to state machines and statecharts? [Read our introduction](/guides/introduction-to-state-machines-and-statecharts/).
 
+[Download our VSCode extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode).
```
mattpocock, line 19: TODO

- [ ] change this link when we deploy the new VS Code extension

docs/guides/typescript.md
```diff
@@ -65,12 +63,210 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
-Providing the context and events as generic parameters for the `createMachine()` function gives many advantages:
+Providing the context and events to the `schema` attribute gives many advantages:
 
 - The context type/interface (`TContext`) is passed on to actions, guards, services and more. It is also passed to deeply nested states.
 - The event type (`TEvent`) ensures that only specified events (and built-in XState-specific ones) are used in transition configs. The provided event object shapes are also passed on to actions, guards, and services.
 - Events which you send to the machine will be strongly typed, offering you much more confidence in the payload shapes you'll be receiving.
 
+## The VSCode Extension <Badge text="4.29+" />
+
+Using our [VSCode extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode), you can automatically generate intelligent typings for XState.
```
davidkpiano, line 74:
```suggestion
Using our [VS Code extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode), you can automatically generate intelligent typings for XState.
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,210 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+
+Here's how you can get started:
+
+1. Download and install the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode).
```
davidkpiano, line 78:
```suggestion
1. Download and install the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode).
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,210 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+2. Open a new file and create a new machine, passing the schema attributes:
+
+```ts
+const machine = createMachine({
```
davidkpiano, line 82:
```suggestion
import { createMachine, t } from 'xstate';

const machine = createMachine({
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,210 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+        FOO: {
+          actions: 'consoleLogValue',
+          target: 'b'
+        }
+      }
+    },
+    b: {
+      entry: 'consoleLogValueAgain'
+    }
+  }
```
davidkpiano, line 128: Just for brevity/easier reading:

```suggestion
  states: {/* ... */}
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,210 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
```
davidkpiano, line 154:
```suggestion
  states: {/* ... */}
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,210 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+);
+```
+
+You'll notice that the events in the options are _strongly typed to the events that cause the action to be triggered_. This is true for actions, guards, services and delays.
+
+You'll also notice that `state.matches`, `tags` and other parts of the machine are now type-safe.
+
+### How to get the most out of the VSCode extension
```
davidkpiano, line 160:
```suggestion
### How to get the most out of the VS Code extension
```

docs/README.md
Andarist:
```suggestion
[Download our VSCode extension](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode).
```

docs/guides/context.md
```diff
@@ -323,17 +323,22 @@ const countMachine = createMachine({
 
 ## TypeScript
 
-For proper type inference, add the context type as the first type parameter to `createMachine<TContext, ...>`:
+For proper type inference, add the context type to the schema property of the machine:
 
 ```ts
+import { t, createMachine } from 'xstate';
```
Andarist, line 329: I wonder if we should emphasize `t` helper here - or if that should only be something for people with aggressive linters.

Kinda fine either way for me, the only reason why I'm just a little bit hesitant about `t` as the default is that I'm worried about the quality of the IDE auto-imports for it. The identifier is so short (to improve part of the DX) that it feels like some other stuff might get often preferred. From my PoV using `{} as` is just as fine as this as it's super easy to write it and there are no gotchas to it (no more than with `t`). It feels "simpler" to me.

But as mentioned - I'm fine both ways, both will work anyway.

docs/guides/context.md
```diff
@@ -323,17 +323,22 @@ const countMachine = createMachine({
 
 ## TypeScript
 
-For proper type inference, add the context type as the first type parameter to `createMachine<TContext, ...>`:
+For proper type inference, add the context type to the schema property of the machine:
```
Andarist, line 326: Maybe we should emphasize somewhere here that using explicit generics is not recommended? We could label that note with the version of XState that will include the typegen support.

docs/guides/typescript.md
```diff
@@ -3,18 +3,16 @@
...
+const lightMachine = createMachine({
+  schema: {
+    // The context (extended state) of the machine
+    context: t<{ elapsed: number }>(),
+    // The events this machine handles
+    events: {} as
+      | { type: 'TIMER' }
+      | { type: 'POWER_OUTAGE' }
+      | { type: 'PED_COUNTDOWN'; duration: number }
+  },
```
Andarist, line 15: I would probably stick with a single way of providing those types throughout the documentation. Using stuff consistently raises less questions from the readers

docs/guides/typescript.md
```diff
@@ -65,12 +63,188 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
 - The event type (`TEvent`) ensures that only specified events (and built-in XState-specific ones) are used in transition configs. The provided event object shapes are also passed on to actions, guards, and services.
 - Events which you send to the machine will be strongly typed, offering you much more confidence in the payload shapes you'll be receiving.
 
+## The VSCode Extension <Badge text="4.29+" />
+
+Using our [VS Code extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode), you can automatically generate intelligent typings for XState.
+
+Here's how you can get started:
+
+1. Download and install the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode).
```
Andarist, line 78:
```suggestion
1. Download and install the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode).
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,188 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
-Providing the context and events as generic parameters for the `createMachine()` function gives many advantages:
+Providing the context and events to the `schema` attribute gives many advantages:
 
 - The context type/interface (`TContext`) is passed on to actions, guards, services and more. It is also passed to deeply nested states.
 - The event type (`TEvent`) ensures that only specified events (and built-in XState-specific ones) are used in transition configs. The provided event object shapes are also passed on to actions, guards, and services.
 - Events which you send to the machine will be strongly typed, offering you much more confidence in the payload shapes you'll be receiving.
 
+## The VSCode Extension <Badge text="4.29+" />
+
+Using our [VS Code extension](https://marketplace.visualstudio.com/items?itemName=mattpocock.xstate-vscode), you can automatically generate intelligent typings for XState.
```
Andarist, line 74:
```suggestion
Using our [VS Code extension](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode), you can automatically generate intelligent typings for XState.
```

docs/guides/context.md
Andarist: `as` has also that small advantage that we don't have to care about adding `t` to the appropriate import when inserting the typegen meta to the file

docs/guides/typescript.md
```diff
@@ -65,12 +63,188 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+      }
+    },
+    b: {
+      entry: 'consoleLogValueAgain'
+    }
+  }
+});
+```
+
+3. Add `tsTypes: true` to the machine and save the file:
```
Andarist, line 106: Perhaps this should use something that is more compatible with the injected replacement
```suggestion
3. Add `tsTypes: {}` to the machine and save the file:
```
```suggestion
3. Add `tsTypes: {}` to the machine and save the file:
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,188 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+    actions: {
+      sayHello: () => {
+        console.log('Hello!');
+      }
+    }
+  }
+);
+```
+
+This is useful, but less optimal:
```
Andarist, line 183: this will prevent loading it in the editor, right? so in a way - it's not that useful 🤣

docs/guides/typescript.md
```diff
@@ -223,8 +397,12 @@ type Event =
       type: 'EVENT_WITHOUT_FLAG';
     };
 
-createMachine<Context, Event>(
+createMachine(
   {
+    schema: {
+      context: t<Context>(),
```
Andarist, line 403: just an extra argument for `{} as ...` - if somebody lands on a doc site which doesnt mention `import { t } from 'xstate'` they might not be able to easily find our what that `t` is and how to get it. Using cmd/ctrl+F: "t" won't help them much either 😉

docs/guides/context.md
mattpocock: We discussed today (after you left this comment) that we'd only do linting, not auto-refactoring on the schema, so we're good - I think.

docs/guides/typescript.md
mattpocock: Iiiish, you'll still be able to load the inspector, and generate typegen from it.

docs/guides/typescript.md
mattpocock: @davidkpiano, what do you think of `t()`?

docs/guides/typescript.md
```diff
@@ -65,12 +63,188 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+});
+```
+
+Named actions/services/guards allow for:
+
+- Better visualisation, because the names appear in the statechart
+- Easier-to-understand code
+- Overrides in `useMachine`, or `machine.withConfig`
+
+#### Typing onDone/onError functions
```
mattpocock, line 201:
```suggestion
#### Typing promise services
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,188 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+
+You can use the generated types to specify the result of `onDone` services, by using the `services` schema property:
```
mattpocock, line 203:
```suggestion
You can use the generated types to specify the return type of promise-based services, by using the `services` schema property:
```

docs/guides/typescript.md
```diff
@@ -65,12 +63,188 @@ const lightMachine = createMachine<LightContext, LightEvent>({
...
+    }
+  }
+});
+```
+
+3. Add `tsTypes: {}` to the machine and save the file:
+
+```ts
+const machine = createMachine({
+  tsTypes: true,
```
janovekj, line 110: Above it says "Add `tsTypes: {}` ...". What does `true` mean here?

mattpocock: True, this shouldn't be here! An old version snuck in.

docs/guides/context.md
Andarist: We've been discussing refactors from generics to schema property but you still need to insert `as typeof import('').Typegen0` at the position of `tsTypes` property. So if we are recommending `t` then it will look somewhat weird that this will use a different pattern (`as` over `t`) or we'll have to potentially import `t` somehow.

docs/guides/typescript.md
Andarist: should we even allow generating typegen with generics? what would happen in such a situation? would it be pushed to the last place of the generics list? would it be inserted at the position of `tsTypes` property?

docs/guides/typescript.md
```diff
@@ -1,75 +1,213 @@
...
+
+## Typegen with the VSCode Extension <Badge text="4.29+" />
 
-const lightMachine = createMachine<LightContext, LightEvent>({
-  key: 'light',
-  initial: 'green',
-  context: { elapsed: 0 },
+Using our [VS Code extension](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode), you can automatically generate intelligent typings for XState.
+
+> This feature is in beta! See the section on known limitations below to see what we're actively looking to improve.
```
Andarist, line 33:
```suggestion
::: warning
This feature is in beta! See the section on known limitations below to see what we're actively looking to improve.
```

with-heart:
One thing that isn't clear to me is how typegen works with `createModel`. Is there anything worth documenting there?

docs/guides/typescript.md
```diff
@@ -1,75 +1,212 @@
...
+    /* ... */
+  }
+});
+```
+
+4. The extension should automatically add a generic to the machine:
+
+```ts
+const machine = createMachine({
+  tsTypes: {} as import('./filename.typegen').Typegen[0]
```
Andarist, line 84:
```suggestion
  tsTypes: {} as import('./filename.typegen').Typegen0
```

Andarist:
> One thing that isn't clear to me is how typegen works with createModel. Is there anything worth documenting there?

It works pretty much in the same way - you declare events/context within `createModel` call (like before). Services have to be declared through `model.createMachine({ schema: { services } })` though and the typegen metadata is injected into this `model.createMachine` call too.

Note that I have a draft to be posted with a proposal to deprecate/fade away `createModel` as something endorsed by the code XState.

docs/guides/context.md
mattpocock: @Andarist This is covered by the VSCode extension itself. It gives you an error line if you use explicit generics. Should we also add a note here?

docs/guides/typescript.md
mattpocock: It gets inserted at the position - then you get a red line at the generics saying 'refactor this to schema'

docs/guides/typescript.md
mattpocock: Changed it to this:

image

mattpocock:
@with-heart Yep I think I think the fact that it's semi-deprecated AND it works pretty seamlessly using the same instruction set should cover us.

docs/guides/context.md
Andarist: I guess it's OK as it is - let's ride on and observe what happens 😉

with-heart:
Gotcha! I didn't notice that it's semi-deprecated! RIP `createModel`, you were a great little creator

PR #2965, opened 2022-01-26
improve exporting of 'actions'

satyasinha:
Hello xstate team :wave: 

I [asked a question](https://discord.com/channels/795785288994652170/799416943324823592/929512618157285447) some time ago if there are any functions that can retrieve the event names for done/error of services but the answer: 
image

DIdn't seem to work. I realised later that some of exports from the `actions.ts` file aren't actually exported properly. Especially the ones we're after e.g. `doneInvoke` and `error`.

So this the change I'm making to improve how actions are exported. I don't think I've broken any exports but now everything that's exported from `actions.ts` is now exposed. Let me know if this not the right fix and I can change this to only export the ones we need and were assumed to be exported.

Cheers!
Satya

### DIFF
--- a/.changeset/flat-dolls-suffer.md
+++ b/.changeset/flat-dolls-suffer.md
@@ -2,4 +2,4 @@
 'xstate': minor
 ---
 
-improve exporting of 'actions'
+All actions are now available in the `actions` variable when importing: `import { actions } from 'xstate'`
--- a/.changeset/flat-dolls-suffer.md
+++ b/.changeset/flat-dolls-suffer.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+improve exporting of 'actions'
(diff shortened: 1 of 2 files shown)


davidkpiano:
Great contribution; thanks! Can you add a changeset? This would be a minor change, I believe (cc. @Andarist)

.changeset/flat-dolls-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+improve exporting of 'actions'
```
davidkpiano, line 5:
```suggestion
All actions are now available in the `actions` variable when importing: `import { actions } from 'xstate'`
```

satyasinha:
@davidkpiano @Andarist was there anything else for this PR? Otherwise bump for visibility and hopefully merge :crossed_fingers: 

Also great work on the typegen, can't wait to integrate it to our own codebase!

PR #2968, opened 2022-01-26
Added a failing test for onReceive

mattpocock:


packages/core/test/typegenTypes.test.ts
```diff
@@ -967,4 +967,58 @@ describe('typegen types', () => {
...
+              // @ts-expect-error
+              event.unknown;
+            });
+          }
+        }
+      }
+    );
+  });
+
+  it("should allow specifying `onReceive`'s argument type manually", () => {
```
Andarist, line 999: This test isn't really showcasing a completely type-safe behavior. This relies on the introduced `bivarianceHack` - while it's unsafe it feels like a balance between correct & friendly to the user.

In the future, I hope to figure out how to make communication between actors completely type-safe.

PR #2969, opened 2022-01-26
Fixed issue where you could not make services type-safe via schema when using createModel

mattpocock:


packages/core/src/model.types.ts
```diff
@@ -36,28 +36,31 @@ export interface Model<
   actions: Prop<TModelCreators, 'actions'>;
   reset: () => AssignAction<TContext, any>;
   createMachine: {
-    <TTypesMeta extends TypegenConstraint = TypegenDisabled>(
+    <
+      TTypesMeta extends TypegenConstraint = TypegenDisabled,
+      TServiceMap extends ServiceMap = ServiceMap
+    >(
```
Andarist, line 42:
```suggestion
    <
      TServiceMap extends ServiceMap = ServiceMap,
      TTypesMeta extends TypegenConstraint = TypegenDisabled
    >(
```

Andarist (approved):

PR #2970, opened 2022-01-26
Added failing test for strange Promise.reject issue

mattpocock:


Andarist:
I think this is a "wont fix" for now. It looks like a bug. I was able to create a [minimal repro case](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEkoGdnwMoBcqZAWSjAAsBLAOxAB4AVAPngG8AoeeABxhIDccEyB7ALbkoEAFzxqAbhbxkIMsAkAKKBOoBKeAF56XfiWAyAvkyagkceADMArmTCYS-Moji8CxclTrKSgtggQQQVsJxdkCWZWeRguEjAQSPgAJXB+GGBKZExOMgBzABp4ZS1deAAFGCESeRp6AB8MMPxCUgp62hMNCX1DGSYwd1xPdqoyW0EAIxAYWmVouVn4xOTF1lwciRQATwcSsvoqmvkAOjgAK3BMZQAiZCEEWZhbjULZVl39zcwAJlLGB9WPA4JhbDBXMdhGdLtc7g8QvBnq8ZMD4MZ3mivmB4D8AMwA9bAsARTDwKA6SrVaEgc4gK6OeGPJEwF4aVFokEgMEQ8kc1gYoE-AAsKkOVJOtNhjIA5NYUPKZRpZILjOygA) for this problem.

I think that this is what happens:
1. TS grabs the function type from this service property
2. it gets its return type and provides the contextual type (the provided `string`), this expands the union from like `Union<T>` to `Union<string>`
3. it recognized the function as being async so it wraps that return type in a Promise, so the type becomes `Promise<Union<string>>`
4. it checks if the type is a union and if it is then it would try to narrow it down, this is probably what happens for my `test4`, However, at this time this type is not a union so this stage is being skipped

I think that, potentially, this could even be an easy fix in TS itself -> just move the narrowing logic **before** wrapping the type into a Promise type. However, promises have this "auto-flattening" behavior that is recursive and maybe this is why this (unconfirmed) order of operation has been chosen. Although, it feels that this wouldn't change that much as the final "flattening"/unwrapping should happen at the end - hard to tell without examining more situations and edge cases.

Andarist:
This is an upstream issue in TS (see [here](https://github.com/microsoft/TypeScript/issues/47682)) and the fix for it awaits review [here](https://github.com/microsoft/TypeScript/pull/51196). It's not actionable on out side here.

Issue #2979, opened 2022-01-28
Compilation issue with xstate@4.29.0 and typescript@4.5.5


mlegait:
**Description**
Since I've upgraded xstate to 4.29.0, my project doesn't compile anymore.

**Expected Result**
TypeScript compilation ok.

**Actual Result**
Here is the error I get:
```
node_modules/xstate/lib/types.d.ts(567,18): error TS2430: Interface 'StateMachine<TContext, TStateSchema, TEvent, TTypestate, TAction, TServiceMap, TResolvedTypesMeta>' incorrectly extends interface 'StateNode<TContext, TStateSchema, TEvent, TTypestate, TServiceMap, TResolvedTypesMeta>'.
  The types of 'withConfig(...).withContext(...).withConfig' are incompatible between these types.
    Type '(options: InternalMachineOptions<TContext, TEvent, AreAllImplementationsAssumedToBeProvided<TResolvedTypesMeta, Prop<TResolvedTypesMeta, "missingImplementations">> extends false ? MarkAllImplementationsAsProvided<...> : TResolvedTypesMeta, true, Prop<...>>, context?: TContext | ... 1 more ... | undefined) => StateMa...' is not assignable to type '(options: InternalMachineOptions<TContext, TEvent, TypegenDisabled, true, never>, context?: TContext | (() => TContext) | undefined) => StateNode<...>'.
      Types of parameters 'options' and 'options' are incompatible.
        Type 'InternalMachineOptions<TContext, TEvent, TypegenDisabled, true, never>' is not assignable to type 'InternalMachineOptions<TContext, TEvent, AreAllImplementationsAssumedToBeProvided<TResolvedTypesMeta, Prop<TResolvedTypesMeta, "missingImplementations">> extends false ? MarkAllImplementationsAsProvided<...> : TResolvedTypesMeta, true, Prop<...>>'.
          Type 'InternalMachineOptions<TContext, TEvent, TypegenDisabled, true, never>' is not assignable to type 'IsNever<Prop<Prop<AreAllImplementationsAssumedToBeProvided<TResolvedTypesMeta, Prop<TResolvedTypesMeta, "missingImplementations">> extends false ? MarkAllImplementationsAsProvided<...> : TResolvedTypesMeta, "missingImplementations">, "actions">> extends true ? {} : { ...; }'.
```

Thank you for your help :)

mattpocock:
[...]

PR #2981, opened 2022-01-28
Move @ts-ignore to a JSDoc comment so it gets preserved after compilation

Andarist:
fixes https://github.com/statelyai/xstate/issues/2979

### DIFF
--- a/.changeset/sour-candles-exist.md
+++ b/.changeset/sour-candles-exist.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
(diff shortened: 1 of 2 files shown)


davidkpiano (approved):

mlegait:
Hi 😄 

When do you plan to release a version with this fix?

Andarist:
@mlegait soon (today/tomorrow), let me just handle some other type-related stuff that could potentially go into the next patch release

For the time being you can use `skilLibCheck` in your `tsconfig.json` (I feel like there is probably no reason to use its default)

PR #2988, opened 2022-01-30
Updated React recipe context typing advice

eostrom:
The context value in the [example](https://xstate.js.org/docs/recipes/react.html#global-state-react-context) isn't an interpreter, but contains an interpreter. The type in the accompanying TypeScript note now matches it.


davidkpiano:
You can remove the changeset - this doesn't need to cause a patch update.

Thanks for the PR!

eostrom:
Changeset removed!

PR #2989, opened 2022-01-30
Removed TS troubleshooting sections

mattpocock:


mattpocock:
These troubleshooting sections have been solved by typegen, so we can remove them.

davidkpiano (changes requested):
These are still relevant to developers using older versions of XState or using XState without typegen (whether unable to or by choice). We should leave these in, and perhaps encourage the usage of typegen in a `TIP:` here instead.

mattpocock:
@davidkpiano I was thinking about this, but I'm not sure they are - if the official advice of the library is to 'use typegen', which fixes these problems, then it doesn't make sense to keep around detailed explanations of lesser solutions.

davidkpiano:
> @davidkpiano I was thinking about this, but I'm not sure they are - if the official advice of the library is to 'use typegen', which fixes these problems, then it doesn't make sense to keep around detailed explanations of lesser solutions.

But there currently are situations where using typegen is not possible (different IDEs, CodeSandbox, etc.)

IMO, if the answer to the question "Is it possible for a user to run into these problems under normal usage?" is "yes", then the troubleshooting advice should still stay there. Removing it has, at best, a net-zero benefit for the user, and at worst, it's removed so they will inevitably ask for troubleshooting help instead of referring to the documentation.

Using XState without typegen should still be considered normal usage.

mattpocock:
@davidkpiano Agreed, maybe when we release the CLI we can ditch this.

I _do_ think removing excessive documentation is a positive for the user. I.e. why have things on the docs site if they're not useful?

davidkpiano:
> @davidkpiano Agreed, maybe when we release the CLI we can ditch this.
> 
> I _do_ think removing excessive documentation is a positive for the user. I.e. why have things on the docs site if they're not useful?

That's a good discussion point (cc. @laurakalbag) - IMO, troubleshooting information is not read like normal docs. Users typically won't read through an entire troubleshooting section; they're likely going to just find (`ctrl-f`) the exact problem they're having.

In that case, having more troubleshooting info is much better than having less. Not finding the thing you're trying to troubleshoot is much more frustrating than thinking "there is too much troubleshooting info".

mattpocock:
@davidkpiano Yes, though maybe we should present it in a different way - i.e. in an accordion or on a separate page.

davidkpiano:
> @davidkpiano Yes, though maybe we should present it in a different way - i.e. in an accordion or on a separate page.

Good idea, maybe `<summary><details>`?

PR #2992, opened 2022-01-30
Fixed an issue with state.context becoming any after `state.matches` when typegen is used

mattpocock:


davidkpiano (approved):

mattpocock:
@Andarist looks good

PR #2995, opened 2022-01-31
Added a section about not using enums

mattpocock:


docs/guides/typescript.md
```diff
@@ -206,6 +206,31 @@ We recommend you commit the generated files (`filename.typegen.ts`) to the repos
 
 If you want to remove the generated file, just remove the `tsTypes` attribute from your machine and it'll stop being generated.
 
+#### Don't use enums
+
+Enums are a common pattern in the XState Typescript commmunity. They're often used to declare state names. like this:
```
Andarist, line 211: This kinda reads like they are common and still useful - maybe using past tense here would be more fitting here?

Andarist (approved):

docs/guides/typescript.md
davidkpiano, line 211:
```suggestion
Enums were a common pattern used with XState TypeScript. They were often used to declare state names. like this:
```

mattpocock: Done

PR #3003, opened 2022-02-02
allow machines to use sendParent when they are not invoked/spawned

pollen8:
A machine should be able to call sendParent regardless of whether it is spawned/invoked or run on its own.
This allows greater re-usability of machines.

### DIFF
--- a/.changeset/proud-carpets-tap.md
+++ b/.changeset/proud-carpets-tap.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-allow machines to use sendParent when they are not invoked/spawned
+Machines can now use `sendParent(...)` without error, even when they are not invoked/spawned from a parent.
--- a/.changeset/proud-carpets-tap.md
+++ b/.changeset/proud-carpets-tap.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+allow machines to use sendParent when they are not invoked/spawned
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -722,8 +722,7 @@ export class Interpreter<
     event: SCXML.Event<AnyEventObject>,
     to: string | number | ActorRef<any>
   ) => {
-    const isParent =
-      this.parent && (to === SpecialTargets.Parent || this.parent.id === to);
+    const isParent = to === SpecialTargets.Parent || this?.parent?.id === to;
     const target = isParent
       ? this.parent
       : isString(to)
--- a/packages/core/test/interpreter.test.ts
+++ b/packages/core/test/interpreter.test.ts
@@ -979,6 +979,26 @@ Event: {\\"type\\":\\"SOME_EVENT\\"}"
         }
       }
     });
+
+    const orphanMachine = Machine<Ctx>({
+      id: 'orphan',
+      initial: 'start',
+      context: {
+        password: ''
+      },
+      states: {
+        start: {
+          onEntry: sendParent((ctx) => ({
+            type: 'NEXT',
+            password: ctx.password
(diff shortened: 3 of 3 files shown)


davidkpiano (approved):
cc. @Andarist Should we show a dev-time warning?

pollen8:
> Also, can you please add a changeset? pray Thanks!
 just checking that this was aimed at @Andarist  and not me?

davidkpiano:
> > Also, can you please add a changeset? pray Thanks!
> > just checking that this was aimed at @Andarist  and not me?

That's for you! 😉 Run `yarn changeset` and let's mark this as a `minor` change.

pollen8:
done :)

Andarist:
Could you back this feature up with a real life use case? Im really curious about this

pollen8:
Sure its  bit convoluted,
I was trying to get around the issue in this codesandbox https://codesandbox.io/s/sad-fog-ypdcw where by using a useSelector to access the spawned child's state, seems to be an update behind.

My fix/hack was to fire a sendParent event to the parent machine which realigned the child machines reported state.
That worked great but meant I couldn't then use my form machine unless it was a child machine. 

Maybe there's a better fix for my initial problem, but in general it just seems wrong to error if a machine attempts to send an event, even if its to a non-existant parent machine.

simonflk:
I have a use-case for this. I have 2 machines that work fine standalone:

* An auth machine handling login, logout & refreshing tokens
* A separate machine which decides which content to show the user

while both work standalone, I also want to be able to compose them and expose events that a parent machine can respond to.

pollen8:
@davidkpiano @Andarist is there anything else you'd need from me to get this merged?

Andarist:
@simonflk could you illustrate this with a runnable example?

@pollen8 your particular example here should be fixed in another way. `getSnapshot()` used within `useSelector `"loses reactivity". You should first grab the `formActor` from the context and then subscribe to it: https://codesandbox.io/s/misty-tree-w27luf?file=/src/App.tsx:331-634

mdpratt:
This is incredibly useful for our usecase where machines can be run as standalone, or as part of a wizard. Is there any way to get some eyes on this again, or should I create a new PR?

davidkpiano:
> This is incredibly useful for our usecase where machines can be run as standalone, or as part of a wizard. Is there any way to get some eyes on this again, or should I create a new PR?

Consider my eyes on it for v5. This should go further than just failing silently when a parent doesn't exist, so here is a proof-of-concept for a dead-letter queue: https://github.com/statelyai/xstate/pull/3904

mdpratt:
Let me give another (better) concrete example.

I have created a generic wrapper that can handle any AWS SDK client and request, with built in pagination. These machines are designed to be called standalone for simple pages, but also integrated as part of larger workflow.

image

However, this means that we basically need to have a factory method where our actions are defined as follows:

```typescript
actions: {
  sendPageReady: settings?.sendParentEvents
    ? sendParent((context) => ({ type: 'PAGE_READY', data: context }))
    : NoOp,

  sendRequestComplete: settings?.sendParentEvents
    ? sendParent((context) => ({ type: 'REQUEST_COMPLETE', data: context }))
    : NoOp,

  sendRequestError: settings?.sendParentEvents
    ? sendParent((context) => ({ type: 'REQUEST_ERROR', data: context }))
    : NoOp,
}
```

davidkpiano:
Closing as stale; in the next version of XState you can send to a potentially undefined parent:

```ts
EVENT: ({ event, parent }, enq) => {
  if (event.value === 42) {
    // parent may be undefined; that's fine
    enq.sendTo(parent, { type: 'SUCCESS' });
  }
}
```

Issue #3010, opened 2022-02-04
Bug: Node.js types are included in 4.29.0


chbdetta:
### Description

After update to 4.29.0, type check fails:

```
src/index.ts:5:7 - error TS2322: Type 'Timeout' is not assignable to type 'number'.

5 const a: number = setTimeout(() => {}, 0);
        ~

```

Somehow the Node.js types are brought into my project. This was not an issue in 4.28.1

### Expected result

No type error should arise

### Actual result

gets error

### Reproduction

https://github.com/chbdetta/xstate-node-js-type-reproduce

PR #3012, opened 2022-02-04
Fixed an issue with a reference to `@types/node` being inserted into XState's compiled output

Andarist:
fixes https://github.com/statelyai/xstate/issues/3010

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -140,7 +140,7 @@ export class Interpreter<
       clearTimeout: (id) => {
         return clearTimeout(id);
       }
-    },
+    } as Clock,
     logger: global.console.log.bind(console),
     devTools: false
   }))(typeof self !== 'undefined' ? self : global);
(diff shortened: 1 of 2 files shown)


packages/core/src/interpreter.ts
```diff
@@ -140,7 +143,7 @@ export class Interpreter<
       clearTimeout: (id) => {
         return clearTimeout(id);
       }
-    },
+    } as Clock,
```
Andarist, line 146: For some reason, `tsc` has started to infer the `setTimeout`'s return type of this default clock to be of type `NodeJS.Timeout`. That in turn has caused it to include `/// <reference types="node" />` at the top of the file and that has created the reported issue.

davidkpiano:
A couple of type errors, but after that LGTM

Andarist:
@davidkpiano i've removed that `ReturnType` thing - it made mocking the clock harder. This should help CI to pass 🤞

davidkpiano (approved):

mattpocock (approved):

Issue #3006, opened 2022-02-03
Bug: Impossible to use generic TContext with `createMachine`


hedgepigdaniel:
### Description

The types for `createMachine` have a [check](https://github.com/statelyai/xstate/blob/7e17c41f68d3805181ebaad09118caf39847872f/packages/core/src/Machine.ts#L89-L98) that aims to prevent passing `Model` instead of a `MachineConfig`. But this check triggers incorrectly for any generic type, because any generic type might add properties such that it extends `Model<any, any, any, any>`.

This results in a type error whenever using createMachine with a `MachineConfig<TContext, ...>` where `TContext` is generic.

```typescript
export const createMachineWithExtras = <
  TContext,
  TEvent extends EventObject,
  TStateSchema extends StateSchema
>(
  makeConfig: (options: {}) => MachineConfig<TContext, TStateSchema, TEvent>
): StateMachine<TContext, TStateSchema, TEvent> =>
  createMachine(makeConfig({}));
```

### Expected result

There should not be a type error.

### Actual result

```
Argument of type 'MachineConfig<TContext, TStateSchema, TEvent, ActionObject<TContext, TEvent>>' is not assignable to parameter of type 'TContext extends Model<any, any, any, any> ? "Model type no longer supported as generic type. Please use `model.createMachine(...)` instead." : MachineConfig<TContext, any, TEvent, ActionObject<...>>'.ts(2345)
```

### Reproduction

https://codesandbox.io/s/compassionate-cray-xvjyr?file=/src/index.tsx

Andarist:
I think it's reasonable to remove this "type-check" now. @hedgepigdaniel would you like to prepare a PR for this?

PR #3027, opened 2022-02-08
Remove type assertion in createMachine

hedgepigdaniel:
Fixes https://github.com/statelyai/xstate/issues/3006

This assertion made it impossible to use `createMachine` with a generic TContext type.


davidkpiano (approved):

mattpocock:
cc @Andarist

Andarist:
We should probably add a type-level test for this change - it could potentially guard us against future regressions. @hedgepigdaniel would you be up for working on it?

.changeset/lemon-mugs-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+fix: allow usage of generic TContext with createMachine
```
Andarist, line 5:
```suggestion
Fixed an issue with not being able to call `createMachine` in a generic context when the type for the context was generic and not concrete.
```

PR #3036, opened 2022-02-12
[@xstate/test] Next version low-level API

davidkpiano:
## Summary of changes:

**Test Model**

```js
import { createTestModel } from '@xstate/test';

// Creating a test model from an existing machine
const testModel = createTestModel(someMachine);

// Provide assertions for each state
const testModel = createTestModel(someMachine, {
  states: {
    loading: async (state) => {
      // expect some SUT to be in 'loading' state
    },
    success: async (state) => {
      // ...
    },
    '*': async (state) => {
      // Tests any unmatched state
    }
  }
});

// Provide cases and execution for events
const testModel = createTestModel(someMachine, {
  // ...
  events: {
    'form.submit': {
      cases: [{ inputValue: 'invalid' }, { inputValue: 'valid' }],
      exec: async ({ state, event }) => {
        // execute event, or throw error if event can't be executed
      }
    },
    'options.click': {
      cases: (state) => state.context.possibleOptions.map(...), // dynamically generate events
      exec: async (step) => { ... }
    }
  }
});
```

**Custom test model**

```js
import { TestModel } from '@xstate/test';

// creating a test model from any custom behavior
const testModel = new TestModel({
  initialState: { ... },
  transition: (state, event) => { ... },
  getEvents: (state) => { ... }
});
```

**Generating paths**
```js
// Get shortest paths
const paths = testModel.getShortestPaths();

// Get simple paths
const paths = testModel.getSimplePaths();

// Get paths that reach specific state
const paths = testModel.getShortestPathsTo(state => state.matches('success'));

const paths = testModel.getSImplePathsTo(state => state.matches('success'));
```

**Custom plan generation**
```js
// Provide a custom path generator (to be documented)
const randomPathGenerator = (behavior, options) => {
  // ...

  // Generates an array of paths
  return [ ... ];
}

const paths = testModel.getPaths(randomPathGenerator);
```

**Testing plans**
```js
const paths = testModel.getShortestPaths();

// Test all plans
await testModel.testPaths(paths);

// Test one plan at a time
for (const path of paths) {
  await testModel.testPath(path);
}
```

**Prevent infinite traversal**
```js
// Will throw error if traversal limit reached
const paths = testModel.getShortestPaths({ traversalLimit: 100 });
```

Todo:

- [x] Add `configure(...)`
- [x] Add `coverage` property to `configure(...)`
- [x] Add `planGenerator` property to `configure(...)` #3302
- [x] Add default for `.getPlans()` (use `planGenerator` as default) #3298
- [x] Add default for `.testPlans()` (use `planGenerator` as default) #3298 
- [x] Add changeset for `.getCoverage(...)` and `.testCoverage(...)`
- [ ] Add changeset for `configure(...)`
- [ ] Add changeset for `createTestModel()` including options (`{ states, events }`)
- [ ] Add changeset for `.getPlans()` and `.testPlans()`
- [ ] Add changeset for `.testPlan()`
- [ ] Add changeset for `.getShortestPlans()` and `.getShortestPlansTo()`
- [ ] Add changeset for `.getSimplePlans()` and `.getSimplePlansTo()`
- [ ] Add changeset for `.getPlanFromEvents()`
- [ ] Add changeset for `.testPath()`
- [ ] Add changeset for `{ traversalLimit }` property on `.get*Plans()`
- [ ] Add changeset for custom plan generators
- [ ] Add changeset for `events` options: `{ cases, exec }`
- [x] Decide whether to remove actions?
- [x] Add wildcard for `events` testing (deferred)
- [x] Remove `beforePath` and `afterPath`
- [x] Deduplicate path generation by default #3304
- [x] Prevent users passing `invoke` or `after` in machine configs #3303
- [x] Add `testSync` methods #3297
- [x] Refactor to `paths` only, not `plans`
- [ ] Transition traversal and coverage by default (configurable)
- [ ] @davidkpiano add JSDoc comment to getAdjancencyMap - https://github.com/statelyai/xstate/pull/3036#discussion_r874806807
- [ ] Remove branded `SerializedState` and `SerializedEvent` types - https://github.com/statelyai/xstate/pull/3036#discussion_r874843683
- [ ] Add comment to ValueMapAdjOptions - https://github.com/statelyai/xstate/pull/3036/files/3f06f9f97ee6e468b786d27ef04b4cdc0a4982c5#r874914838


davidkpiano:
This is now ready for review

packages/xstate-test/src/types.ts
```diff
@@ -125,20 +130,137 @@ export interface TestEventConfig<TTestContext> {
...
+    | EventExecutor<TState, TEvent>
+    | TestEventConfig<TState, TEvent>;
+};
+
+export interface TestModelEventConfig<TState, TEvent extends EventObject> {
+  cases?:
+    | ((state: TState) => Array<EventCase<TEvent>>)
+    | Array<EventCase<TEvent>>;
+  exec?: EventExecutor<TState, TEvent>;
+}
```
Silverwolf90, line 147: Minor quality of life thing I noticed while refactoring some experimental code to use new API:

When using `getPlanForEvents` it's not necessary to define `cases` since they are unsused. Previously, the events config allowed mapping to the `EventExecutor` directly without needing to wrap it in `{ exec: ... }` which was a nice convenience when you've got an object with tons of events. Often my event executors are one-liners, like the following:

```ts
'click': () => cy.findByRole('button', /click me/i).click(),
```

so it feels extra cumbersome to have to expand that to:

```ts
'click': {
  exec: () => cy.findByRole('button', /click me/i).click(),
},
```

especially when you've got dozens of these one-liners.

Silverwolf90:
```ts
// Provide assertions for each state
const testModel = createTestModel(someMachine, {
  states: {
    loading: async (state) => {
      // expect some SUT to be in 'loading' state
    },
    success: async (state) => {
      // ...
    },
    '*': async (state) => {
      // Tests any unmatched state
    }
  }
});
```

Can these keys use state ids? Or must they be the full path?

packages/xstate-test/src/types.ts
davidkpiano: I fixed this so that you can either specify an executor by itself, or in an `{ exec: ... }` object.

davidkpiano:
> Can these keys use state ids? Or must they be the full path?

Added to [0c483ec](https://github.com/statelyai/xstate/pull/3036/commits/0c483ec736c02f05a0b0593c2d91a88049a86e4d)

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,365 @@
...
+ *
+ * ```js
+ * const toggleModel = createModel(toggleMachine).withEvents({
+ *   TOGGLE: {
+ *     exec: async page => {
+ *       await page.click('input');
+ *     }
+ *   }
+ * });
+ * ```
```
Silverwolf90, line 48: this example still uses the old API

Silverwolf90:
> > Can these keys use state ids? Or must they be the full path?
> 
> They can now be both; added to [0c483ec](https://github.com/statelyai/xstate/pull/3036/commits/0c483ec736c02f05a0b0593c2d91a88049a86e4d)

Ah I've just tested this out and realize my question wasn't precise enough. I was expecting paths w/ id to work -- just like any transition target (eg: `#someId.a.b.c`)

davidkpiano:
> > > Can these keys use state ids? Or must they be the full path?
> > 
> > 
> > They can now be both; added to [0c483ec](https://github.com/statelyai/xstate/pull/3036/commits/0c483ec736c02f05a0b0593c2d91a88049a86e4d)
> 
> Ah I've just tested this out and realize my question wasn't precise enough. I was expecting paths w/ id to work -- just like any transition target (eg: `#someId.a.b.c`)

Let's save this as a future feature request - you can workaround this by using IDs per state or the full `foo.bar.baz` path.

packages/xstate-test/src/index.ts
```diff
@@ -1,422 +1,23 @@
...
-  TestModelOptions,
-  TestPlan,
-  StatePredicate,
-  TestPathResult,
-  TestSegmentResult,
-  TestMeta,
-  EventExecutor,
-  CoverageOptions
-} from './types';
+import { EventObject } from 'xstate';
```
Silverwolf90, line 1: I believe this file is missing the export of `TestModel`.

davidkpiano: Fixed here: [57acfad](https://github.com/statelyai/xstate/pull/3036/commits/57acfadd5ec27ca329eb6061f94353110460e003)

tivac:
@davidkpiano do you have an ETA for this to land?

horacioh:
> // expect some SUT to be in 'loading' state
what is SUT?

davidkpiano:
> > // expect some SUT to be in 'loading' state
> 
> what is SUT?

System Under Test: https://en.wikipedia.org/wiki/System_under_test

Silverwolf90:
> @davidkpiano do you have an ETA for this to land?

I am also curious about this. Perhaps it could be published to npm as a prerelease tag or `@next` tag?

mattpocock:
Some thoughts:

## Move coverage settings to global config

We should move the coverage settings to a global config, similar to other tools like Playwright.

```ts
import { configure } from '@xstate/test';

configure({
  coverageMetrics: ['all-events-covered', 'all-states-covered'],
});
```

These can then either be set inside each test file, or set at a global level in `setupTestsAfterEnv` (in jest, for instance).

## Automatic deduplication of paths

We MUST by default deduplicate any paths to be tested. I've added a POC here:

https://github.com/mattpocock/xstate-test-playwright/blob/main/src/utils.ts

It's crazy that this isn't the default behaviour - it spawns a bunch of useless tests that ONLY test what is already covered elsewhere.

## Before and after path hooks

I don't see a use case for this - they should be removed.

## Default `getPlans` function

Calling `getPlans` with `undefined` should fall back to some sensible default behaviour, perhaps configurable in `configure`. We shouldn't be asking users to choose or know the difference between `getShortest` and `getSimple`. We should present `getPlans` as the default API. This reduces friction.

The sensible default behaviour should be, IMO:

1. All transitions should be traversed
2. All states should be traversed

Perhaps this should be calculated by the coverage metrics suggested above.

## `testSync` function

You should be allowed to test in a sync environment instead of always requiring a promise to be returned. This will allow `@xstate/test` to work with Cypress.

mattpocock:
We should also consider an API that is more implicit for running tests. For instance, `model.runTestsInBlocks` (bikesheddable) could try to wrap each test in `it` or `describe` if found in globals (from jest) or you could pass it into `configure`.

That would make the API:

```ts

const model = createModel(machine, {
  states: {},
  events: {},
});

describe('myMachine', () => {
  model.runTestsInBlocks();
});
```

davidkpiano:
@mattpocock Some of the changes you are asking for are for a _higher-level API_; this PR is meant to represent the base-level API which can enable a higher-level API layer that can achieve what you want. Just something to keep in mind; we don't want to limit flexibility and paint ourselves in a corner for the sake of a (subjectively) nicer API.

- Move coverage settings to global config

This makes sense, but it should cascade: there will be use-cases where the coverage metrics need to be _different_ per test model, or at least slightly tweaked. For example, a dev might want one particular test model to cover all states except those with a specific tag, so we still need to provide the ability to configure coverage at an individual test-model level and provide custom coverage metrics, which is the use-case of that function-call-in-function:

```js
testModel.testCoverage(coversAllStates({ filter: ... }))
```

- Automatic deduplication of paths

✅  That makes sense; this should be done at the `@xstate/test` level but not the `@xstate/graph` level, since there may be other valid use-cases for generating all paths, even if some are subpaths of other paths.

- Before and after path hooks

There are valid use-cases for this. For example, if we're using Puppeteer/Playwright, we would want to close a browser page (`afterPath`) and open a new one (`beforePath`).

- Default getPlans function

✅  Yes, a sensible default can be simple paths (with deduplication) that covers states and transitions. The risk is that it's not explicit by the function name alone (`getPlans()` is ambiguous), so we need to make this clear in documentation and also _not_ avoid presenting `getSimplePlans()` and `getShortestPlans()` in the examples.

- testSync function

Having `@xstate/test` work with Cypress should already be possible as-is: https://timdeschryver.dev/blog/generated-tests-with-xstate-and-cypress

Did anything change in this PR that would make it not work?

- coverage functions

I want to push back against this statement:

> Passing the result of called functions to other function calls feels unnecessarily boilerplate-y.

Again, this PR is for a _lower-level API_. We can discuss a higher-level API but that should not replace the full flexibility of a lower-level API, even if it might feel more boilerplate-y.

- testPlans function

✅  Having a sensible default makes sense.

> We should also consider an API that is more implicit for running tests. 

This is something that can be discussed after this PR, since it is a higher-level API. We need to avoid tying `@xstate/test` to any specific testing frameworks, regardless of popularity (Mocha, Jest, Vitest, etc.), especially since there can be different globals/aliases used. One idea is to have this as part of configuration:

```js
import { configure } from '@xstate/test';

configure({
  // explicitly pass in the globals
  it: it, 
  describe: test,
  // ...
});

```

mattpocock:
@davidkpiano Will reply to this fully tomorrow, but wanted to reply to your top-level point:

> Some of the changes you are asking for are for a higher-level API; this PR is meant to represent the base-level API which can enable a higher-level API layer that can achieve what you want. Just something to keep in mind; we don't want to limit flexibility and paint ourselves in a corner for the sake of a (subjectively) nicer API.

Yes, I think we should be opinionated about our API so user's don't have to be. If we only provide API's for power users, we risk alienating _most_ of our users who will have to always write boilerplate to get the library to perform its basic functions.

I think that the decisions you're offering users (how to measure coverage, how to traverse the graph) are decisions users _will not want to make_. I.e. they won't care about the trade-offs enough to need to read up on them. We should prepare a golden path for users to call the simplest parts of the API and get _exactly what most users need_.

mattpocock:
> before/after path hooks

My point here is that you can do these very effectively with `beforeEach`, `afterEach` (present in most test frameworks) or by simply doing setup before/after each `plan.test()` call:

```ts
// Run something before the test
await plan.test();
// Run something after the test
```

mattpocock:
> Also, I don't see how using a called function (which provides more configurability) is significantly more boilerplate than a magic string:

My point here is that the magic string is discoverable by TypeScript and autocomplete, while the called function approach is not. We can, of course, have both!

I'm also open to the idea of not doing the magic strings and having sensible defaults if nothing is specified.

mattpocock:
```ts
import { configure } from '@xstate/test';

configure({
  // explicitly pass in the globals
  it: it, 
  describe: test,
  // ...
});
```

Yes, I was thinking this too!

I do think that it makes sense to make it fall back to the globals (`describe` is `global.describe`, `test` is `global.test`) but configuring it here makes sense too.

I'm also a bit meh over whether this API is needed, probably it's easier just to do the `for` loops. That way you can do setup per test or mocks etc.

davidkpiano:
> @davidkpiano Will reply to this fully tomorrow, but wanted to reply to your top-level point:
> 
> > Some of the changes you are asking for are for a higher-level API; this PR is meant to represent the base-level API which can enable a higher-level API layer that can achieve what you want. Just something to keep in mind; we don't want to limit flexibility and paint ourselves in a corner for the sake of a (subjectively) nicer API.
> 
> Yes, I think we should be opinionated about our API so user's don't have to be. If we only provide API's for power users, we risk alienating _most_ of our users who will have to always write boilerplate to get the library to perform its basic functions.
> 
> I think that the decisions you're offering users (how to measure coverage, how to traverse the graph) are decisions users _will not want to make_. I.e. they won't care about the trade-offs enough to need to read up on them. We should prepare a golden path for users to call the simplest parts of the API and get _exactly what most users need_.

Discussed externally:

- Provide sensible defaults as happy path, with ability to customize
- Will consider strings (with TS autocompletion) after PR

davidkpiano:
@mattpocock Added a checklist for release: https://github.com/statelyai/xstate/pull/3036#issue-1134110169

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,383 @@
...
+        const testStepResult: TestStepResult = {
+          step,
+          state: { error: null },
+          event: { error: null }
+        };
+
+        testPathResult.steps.push(testStepResult);
+
+        try {
+          await this.testState(step.state, options);
```
mattpocock, line 223: Should this be receiving resolvedOptions instead? Same as below.

mattpocock:
@davidkpiano Some missing things:

`testSync` function
Automatic dedup of paths

mattpocock:
@davidkpiano After a chat with Andarist, we should also talk about whether we should error when we encounter an `invoke`, an `after` or a delayed `send`. All of these behaviours will not go as expected in `@xstate/test` - invokes will not be run, `after` will not be queued up, delayed `send` will not be sent.

To combat this, I think we should allow passing a `TestMachineConfig` direct to `createModel`:

```ts
const model = createTestModel({
  initial: 'a',
  states: {
    a: {},
  },
}, {
  states: { a },
})
```

And also expose a `createTestMachine` API which also takes `TestMachineConfig`:

```ts
const machine = createTestMachine({
  initial: 'a',
  states: {
    a: {},
  },
});

const model = createTestModel(machine);
```

`createTestModel` will still be able to receive a `StateMachine` (from `createMachine`) so nothing will break.

.changeset/great-spies-exist.md
```diff
@@ -0,0 +1,51 @@
...
+---
+
+Coverage can now be obtained from a tested test model via `testModel.getCoverage(...)`:
+
+```js
+import { configure, createTestModel } from '@xstate/test';
+import {
+  coversAllStates,
+  coversAllTransitions
+} from '@xstate/test/lib/coverage';
```
Andarist, line 12: is there any advantage for this here? I mean - cant we just reexport everything from the main entrypoint?

packages/xstate-graph/src/graph.ts
```diff
@@ -67,59 +73,49 @@ export function getChildren(stateNode: AnyStateNode): AnyStateNode[] {
...
   filter: () => true,
-  stateSerializer: serializeState,
+  stateSerializer: serializeMachineState,
   eventSerializer: serializeEvent
 };
 
-function getValueAdjMapOptions<TContext, TEvent extends EventObject>(
-  options?: ValueAdjMapOptions<TContext, TEvent>
-): Required<ValueAdjMapOptions<TContext, TEvent>> {
+function getValueAdjMapOptions<TState, TEvent extends EventObject>(
```
Andarist, line 94: Shouldn't this have a constraint like this?
```suggestion
function getValueAdjMapOptions<TState extends AnyState, TEvent extends EventObject>(
```

.changeset/great-spies-exist.md
```diff
@@ -0,0 +1,51 @@
...
+
+// Filters state coverage
+testModel.testCoverage(
+  coversAllStates({
+    filter: (stateNode) => {
+      return stateNode.key !== 'somePassedState';
+    }
+  })
+);
+```
```
Andarist, line 51: we should also update (or at least note down that we need to update it) the README of the `@xstate/test` package

packages/xstate-graph/src/graph.ts
```diff
@@ -67,59 +73,49 @@ export function getChildren(stateNode: AnyStateNode): AnyStateNode[] {
...
-export function getAdjacencyMap<
-  TContext = DefaultContext,
-  TEvent extends EventObject = AnyEventObject
->(
-  node:
-    | StateNode<TContext, any, TEvent, any, any, any>
-    | StateMachine<TContext, any, TEvent, any, any, any, any>,
-  options?: ValueAdjMapOptions<TContext, TEvent>
-): AdjacencyMap<TContext, TEvent> {
+export function getAdjacencyMap<TMachine extends AnyStateMachine>(
```
Andarist, line 105: could we add a jsdoc comment above this that would explain what it does and what is it useful for? I only see some tests related to this within `@xstate/graph`'s test suite but the tests are mostly concerned about the shape of the returned object and I'm not exactly sure how to utilize that returned object creatively 

it would be cool to understand the difference between this util and other utils available in this pkg (some of which are actually used by `@xstate/test`)

davidkpiano: Yes, I will add the JS docs 👍

packages/xstate-graph/src/graph.ts
```diff
@@ -171,182 +167,141 @@ export function getAdjacencyMap<
...
+  const { serializeState } = optionsWithDefaults;
+
+  const adjacency = performDepthFirstTraversal(behavior, optionsWithDefaults);
 
   // weight, state, event
   const weightMap = new Map<
-    string,
-    [number, string | undefined, string | undefined]
+    SerializedState,
+    [number, SerializedState | undefined, TEvent | undefined]
```
Andarist, line 212: You could remove the comment above this declaration if you would use tuple element labels here:
```suggestion
    [weight: number, state: SerializedState | undefined, event: TEvent | undefined]
```

packages/xstate-graph/src/graph.ts
```diff
@@ -171,182 +167,141 @@ export function getAdjacencyMap<
...
-  stateMap.set(initialVertex, machine.initialState);
+  const stateMap = new Map<SerializedState, TState>();
+  const initialSerializedState = serializeState(behavior.initialState, null);
+  stateMap.set(initialSerializedState, behavior.initialState);
 
-  weightMap.set(initialVertex, [0, undefined, undefined]);
-  const unvisited = new Set<string>();
+  weightMap.set(initialSerializedState, [0, undefined, undefined]);
+  const unvisited = new Set<SerializedState>();
   const visited = new Set<string>();
```
Andarist, line 220: If the `unvisited` type has been changed here then the same thing should happen for `visited` since both are highly related (if i read this right):
```suggestion
  const visited = new Set<SerializedState>();
```

packages/xstate-graph/src/types.ts
```diff
@@ -57,72 +56,112 @@ export type DirectedGraphNode = JSONSerializable<
...
+
+export interface VisitedContext<TState, TEvent> {
+  vertices: Set<SerializedState>;
+  edges: Set<SerializedEvent>;
+  a?: TState | TEvent; // TODO: remove
+}
+
+export interface SerializationOptions<TState, TEvent> {
+  serializeState: (state: TState, event: TEvent | null) => SerializedState;
+  serializeEvent: (event: TEvent) => SerializedEvent;
```
Andarist, line 138: If I read things correctly then those options are public. Requiring people to provide functions that return those branded types feels like a chore on their side - it won't be super obvious and we'd have to document it. I don't think it's worth it. We should expect a simple `string`s here and cast those functions internally if we want to use those branded types internally for improved type safety

packages/xstate-graph/src/graph.ts
```diff
@@ -171,182 +167,141 @@ export function getAdjacencyMap<
...
           ]
     };
   });
 
-  return statePathMap;
+  return Object.values(statePlanMap);
 }
 
-export function getSimplePaths<
+export function getSimplePlans<
```
Andarist, line 289: what's the current purpose of this helper? what's a meaningful difference between this and the `traverseSimplePlans` (that is being called here)?

packages/xstate-graph/src/graph.ts
```diff
@@ -393,64 +348,295 @@ export function toDirectedGraph(
...
+function resolveTraversalOptions<TState, TEvent>(
+  traversalOptions?: Partial<TraversalOptions<TState, TEvent>>,
+  defaultOptions?: TraversalOptions<TState, TEvent>
+): Required<TraversalOptions<TState, TEvent>> {
+  const serializeState =
+    traversalOptions?.serializeState ??
+    defaultOptions?.serializeState ??
+    ((state) => JSON.stringify(state) as any);
+  return {
+    serializeState,
```
Andarist, line 482: I'm pretty sure that u can simplify this to this:
```suggestion
  return {
    serializeState: (state) => JSON.stringify(state) as any,
```
After all, you are already spreading `traversalOptions` and `defaultOptions` with the same "priorities" at the end of this object.

packages/xstate-graph/src/types.ts
```diff
@@ -57,72 +56,112 @@ export type DirectedGraphNode = JSONSerializable<
...
+}
+
+export interface TraversalOptions<TState, TEvent>
+  extends SerializationOptions<TState, TEvent> {
+  visitCondition?: (
+    state: TState,
+    event: TEvent,
+    vctx: VisitedContext<TState, TEvent>
+  ) => boolean;
+  filter?: (state: TState, event: TEvent) => boolean;
```
Andarist, line 148: especially those two could get dedicated jsdoc comments here as both accept similar arguments and both return booleans so it's not overly obvious how they are actually used by the implementation

packages/xstate-graph/src/graph.ts
Andarist: is the main difference that `get*` functions accept a state machine whereas the `traverse*` ones receive a behavior?

packages/xstate-graph/src/graph.ts
```diff
@@ -393,64 +348,295 @@ export function toDirectedGraph(
...
+  const simplePlansMap = traverseSimplePlans(behavior, resolvedOptions);
+
+  return filterPlans(simplePlansMap, predicate);
+}
+
+export function traverseSimplePathsFromTo<TState, TEvent extends EventObject>(
+  behavior: SimpleBehavior<TState, TEvent>,
+  fromPredicate: (state: TState) => boolean,
+  toPredicate: (state: TState) => boolean,
+  options: TraversalOptions<TState, TEvent>
```
Andarist, line 603: maybe those could be merged in a single options param? 4 positional arguments is already a lot to memorize

packages/xstate-graph/src/graph.ts
```diff
@@ -393,64 +348,295 @@ export function toDirectedGraph(
...
+): Array<StatePlan<TState, TEvent>> {
+  const resolvedOptions = resolveTraversalOptions(options);
+  const simplePlansMap = traverseSimplePlans(behavior, resolvedOptions);
+
+  // Return all plans that contain a "from" state and target a "to" state
```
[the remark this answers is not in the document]
Andarist: The filtering remark is ofc applicable to all functions that are actually filtering the whole computed `plansMap`

packages/xstate-graph/src/types.ts
```diff
@@ -57,72 +56,112 @@ export type DirectedGraphNode = JSONSerializable<
...
+  Step<TState, TEvent>
 >;
 
 export type ExtractEvent<
   TEvent extends EventObject,
   TType extends TEvent['type']
 > = TEvent extends { type: TType } ? TEvent : never;
 
-export interface ValueAdjMapOptions<TContext, TEvent extends EventObject> {
+export interface ValueAdjMapOptions<TState, TEvent extends EventObject> {
```
Andarist, line 119: q: why is this a `ValueAdjMapOptions` and not `AdjMapOptions`? 🤔 

as a sidenote - it would be cool to not use the `adj` abbreviation in any type name - I remember that when I was first reading the codebase I had to look up what the adj stands for and I couldn't simply google it because it was a custom~ abbreviation, whereas even if I didn't know what adjacency is i would be able to google for it

packages/xstate-graph/test/graph.test.ts
```diff
@@ -158,63 +197,111 @@ describe('@xstate/graph', () => {
...
 
     it('the initial state should have a zero-length path', () => {
+      const shortestPaths = getShortestPlans(lightMachine);
+
       expect(
-        getShortestPaths(lightMachine)[
-          JSON.stringify(lightMachine.initialState.value)
-        ].paths[0].segments
+        shortestPaths.find((plan) =>
+          plan.state.matches(lightMachine.initialState.value)
```
Andarist, line 217: looking here - I would have no idea that `plan.state` is holding the "final"/target/end state of the whole "plan". Maybe this could use a more descriptive name?

packages/xstate-graph/test/graph.test.ts
```diff
@@ -432,3 +571,102 @@ describe('@xstate/graph', () => {
...
+        }
+        if (e.type === 'b' && s === 1) {
+          return 2;
+        }
+        if (e.type === 'reset') {
+          return 0;
+        }
+        return s;
+      },
+      initialState: 0 as number
```
Andarist, line 590: nit: 0 is already a number 😅 
```suggestion
      initialState: 0
```

packages/xstate-graph/src/types.ts
davidkpiano: Good point regarding `adj` - I'll spell it out.

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+  coverage: [coversAllStates<any, any>(), coversAllTransitions<any, any>()],
+  planGenerator: traverseShortestPlans
+};
+
+/**
+ * Creates a test model that represents an abstract model of a
+ * system under test (SUT).
+ *
+ * The test model is used to generate test plans, which are used to
+ * verify that states in the `machine` are reachable in the SUT.
```
Andarist, line 50: Since this is about an abstract system we shouldn't refer to a machine here
```suggestion
 * verify that states in the `machine` are reachable in the SUT.
```

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+      stateMatcher: (_, stateKey) => stateKey === '*',
+      getStates: () => [],
+      execute: () => void 0,
+      logger: {
+        log: console.log.bind(console),
+        error: console.error.bind(console)
+      }
+    };
+  }
+  public static defaults: TestModelDefaults<any, any> = testModelDefaults;
```
Andarist, line 76: Those two are quite confusing - I have no idea what's the purpose of each of them and where I should add a new option/default in the future.

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+  }
+
+  public getShortestPlans(
+    options?: Partial<TraversalOptions<TState, TEvent>>
+  ): Array<StatePlan<TState, TEvent>> {
+    return this.getPlans({ ...options, planGenerator: traverseShortestPlans });
+  }
+
+  public getShortestPlansTo(
+    stateValue: StatePredicate<TState>
```
Andarist, line 106:
```suggestion
    statePredicate: StatePredicate<TState>
```

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+  ): Array<StatePlan<TState, TEvent>> {
+    let minWeight = Infinity;
+    let shortestPlans: Array<StatePlan<TState, TEvent>> = [];
+
+    const plans = this.filterPathsTo(stateValue, this.getShortestPlans());
```
Andarist, line 111:
```suggestion
    const plans = this.filterPathsTo(statePredicate, this.getShortestPlans());
```

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+        try {
+          await this.testState(step.state, options);
+        } catch (err) {
+          testStepResult.state.error = err;
+
+          throw err;
+        }
+
+        try {
+          await this.testTransition(step);
```
Andarist, line 278: re duplication between `testPlan` and `testPlanSync` etc... the main (and only?) difference is that `testPlan` automatically awaits some results, right? so... what if we'd just check if the result is a promise and conditionally await it? This way we could just have a single implementation.

The outermost call would still have to be an async function though to use `await` conveniently in the implementation **if** the return value turns out to be a promise. OTOH we could technically ditch `await` sugar and deal with bare promises here in an attempt to have just a single implementation.

The inner calls wouldn't be able to use `async` sugar either because even if they don't return anything their void return value would still be promisified. The whole async call stack is just "colored" this way and we can't escape it within the bounds of the sugar syntax.

We could ditch the sugar altogether and it's totally doable **but** it complicates the control flow with loops, try/catches etc. Ultimately - it wouldn't quite be worth it.

There is a neat language construct that would actually allow us to share the implementation, has a nice sugar for the control flow, and that allows a function~ to return any value, that doesn't mandate returning a promise... and that construct is called a generator function 😅 

Generators might not always be super ergonomic for public consumption but it's usually fairly easy to write a wrapper function that calls into the generator and return a plain value for the caller. This is actually how some methods in the Babel are implemented nowadays (for example you have `transformFile` and `transformFileSync` APIs that share the very same, generator-based, implementation, but that's completely opaque to the user). They have even made a package that makes implementing this pattern easier: https://github.com/loganfsmyth/gensync

It's probably not worth at all to refactor this at all right now, I'm just laying out the possibilities here. 

Generators have a nice trait though - they are lazy by default, you could have a plan generator that would generate a *single* path, execute it then and only generate another plan after you are done with the first one. We could even implement test parallelization at this level~ without paying the price to generate *all* the paths upfront which gives us somewhat of a cold start before any test can even be executed.

(sorry for the chaos, was just thinking out loud, in writing)

davidkpiano: cc. @mattpocock What do you think?

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+
+    for (const plan of plans) {
+      const currWeight = plan.paths[0].weight;
+      if (currWeight < minWeight) {
+        minWeight = currWeight;
+        shortestPlans = [plan];
+      } else if (currWeight === minWeight) {
+        shortestPlans.push(plan);
+      }
+    }
```
Andarist, line 121: I don't quite understand this algorithm - I mean I understand the filtering algorithm based on weights but I don't understand the concept of filtering *plans* based on a *single* path of that plan. How do we even guarantee within `@xstate/graph` that the first path is going to have the lowest weight within that plan? 🤔 

Are we first filtering paths within a plan based on weights in `getShortestPlans` and then filter plans here based on those results based on weights again? Why this second filter is not implemented in the `traverseShortestPathsTo`?

packages/xstate-test/src/TestModel.ts
Andarist: Note that the lazy generation probably wouldnt work for plans that require knowing all of the generated paths for weight-based filtering and stuff - but I wonder if that's not the limitation of the depth-first traversal that we use. It feels like with breadth-first traversal it could be possible to do this

packages/xstate-test/src/dedupPathPlans.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  // Sort by path length, descending
+  allPathsWithPlan.sort((a, z) => z.path.steps.length - a.path.steps.length);
+
+  const superpathsWithPlan: typeof allPathsWithPlan = [];
+
+  /**
+   * Filter out the paths that are subpaths of superpaths
+   */
+  pathLoop: for (const pathWithPlan of allPathsWithPlan) {
```
Andarist, line 43: Second time ever that I see a literal label combined with break/continue in JS. And the second time that I see this in XState - neat 😉

packages/xstate-test/src/testUtils.ts
```diff
@@ -0,0 +1,22 @@
+import { StatePlan } from '@xstate/graph';
```
Andarist, line 1: this file is only used for tests - it would make sense to move it there cause right now it looks like it's a public API when it's declared in `src`, unless... this is a public API?

packages/xstate-test/src/machine.ts
```diff
@@ -0,0 +1,135 @@
...
+export async function testStateFromMeta(state: AnyState) {
+  for (const id of Object.keys(state.meta)) {
+    const stateNodeMeta = state.meta[id];
+    if (typeof stateNodeMeta.test === 'function' && !stateNodeMeta.skip) {
+      await stateNodeMeta.test(state);
+    }
+  }
+}
+
+export function createTestMachine<
```
Andarist, line 33: what is the exact purpose of this - I mean: at the moment both `createMachine` and `createTestMachine` will be accepted as arguments for `createTestModel`, right? What is going to be put in the docs etc about this?

packages/xstate-test/src/machine.ts
```diff
@@ -0,0 +1,135 @@
...
+  TContext,
+  TEvent extends EventObject = AnyEventObject,
+  TTypesMeta extends TypegenConstraint = TypegenDisabled
```
Andarist, line 36: are we going to have a separate type of typegen for test machines?

packages/xstate-test/src/machine.ts
```diff
@@ -0,0 +1,135 @@
...
+ *   }
+ * });
+ * ```
+ *
+ * @param machine The state machine used to represent the abstract model.
+ * @param options Options for the created test model:
+ * - `events`: an object mapping string event types (e.g., `SUBMIT`)
+ * to an event test config (e.g., `{exec: () => {...}, cases: [...]}`)
+ */
+export function createTestModel<TMachine extends AnyStateMachine>(
```
Andarist, line 81: `TestModel` is abstract and capable of working with simple behaviors, this function here (that is named `createTestModel`) is strongly coupled to machines though. We should consider renaming this to smth like `createMachineTestModel` or smth.

packages/xstate-test/test/coverage.test.ts
```diff
@@ -0,0 +1,389 @@
...
+    const model = createTestModel(machine);
+    const shortestPlans = model.getShortestPlans();
+
+    for (const plan of shortestPlans) {
+      await model.testPlan(plan);
+    }
+
+    // TODO: determine how to handle missing coverage for transient states,
+    // which arguably should not be counted towards coverage, as the app is never in
+    // a transient state for any length of time
```
Andarist, line 189: isn't it still important for coverage to check if a transient state was visited though?

packages/xstate-test/test/plans.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    await testUtils.testPlans(testModel, plans);
+
+    expect(testModel.getCoverage(coversAllStates())).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "criterion": Object {
+            "description": "Visits \\"(machine)\\"",
+            "predicate": [Function],
+            "skip": false,
```
Andarist, line 79: I can't find any test that would end up with `skip: true`, could you create one?

packages/xstate-test/test/plans.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+      const model = createTestModel(machine);
+
+      const plans = model.getPlans();
+
+      expect(plans).toHaveLength(1);
+    });
+  });
+
+  describe('When the machine only has more than one path', () => {
```
Andarist, line 130:
```suggestion
  describe('When the machine has more than one path', () => {
```

packages/xstate-test/test/plans.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      const model = createTestModel(multiPathMachine);
+
+      const plans = model.getPlans();
+
+      expect(plans).toHaveLength(2);
+    });
+  });
+
+  describe('simplePathPlans', () => {
+    it('Should dedup simple path plans too', () => {
```
Andarist, line 141: "too" is rather confusing here as there is no good context for it - and especially with time it wont be obvious what it relates to
```suggestion
    it('Should dedup simple path plans', () => {
```

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+  public defaultTraversalOptions?: TraversalOptions<TState, TEvent>;
+  public getDefaultOptions(): TestModelOptions<TState, TEvent> {
+    return {
+      serializeState: (state) => simpleStringify(state) as SerializedState,
+      serializeEvent: (event) => simpleStringify(event) as SerializedEvent,
+      getEvents: () => [],
+      states: {},
+      events: {},
+      stateMatcher: (_, stateKey) => stateKey === '*',
+      getStates: () => [],
```
Andarist, line 68: Is this used anywhere? I can only see it being referenced here and declared on the `TestModelOptions` type.

packages/xstate-test/src/dedupPathPlans.ts
davidkpiano: That's how you know we do ✨ algorithms ✨ here

packages/xstate-test/src/testUtils.ts
davidkpiano: Yes, this should be moved to `/tests/testUtils.ts`

packages/xstate-test/src/machine.ts
davidkpiano: `createTestMachine` provides additional constraints on the machines that `createTestModel(...)` accepts, to prevent footguns

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+    transitions: {}
+  };
+  public options: TestModelOptions<TState, TEvent>;
+  public defaultTraversalOptions?: TraversalOptions<TState, TEvent>;
+  public getDefaultOptions(): TestModelOptions<TState, TEvent> {
+    return {
+      serializeState: (state) => simpleStringify(state) as SerializedState,
+      serializeEvent: (event) => simpleStringify(event) as SerializedEvent,
+      getEvents: () => [],
+      states: {},
```
Andarist, line 65: `states` is nice and short but also a little bit enigmatic - maybe worth considering `stateAssertions` for this or something?

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+      events: {},
```
Andarist, line 66: if you like the proposal to rename `states` to `stateAssertions` then this could be renamed too but I don't have any particular proposal for a better name myself

packages/xstate-test/test/coverage.test.ts
davidkpiano: Good question; I don't think so. For MBT, a transient state is used as a mechanism for going to non-transient states, not as an actual state to be tested, so its coverage doesn't matter.

If a "transient" state is somehow represented in the SUT, then 1) it should be represented as a real state, not as a transient state, and 2) it's not really a transient state anyway; there's a non-zero amount of time the SUT will be in that state.

packages/xstate-graph/src/graph.ts
Andarist: Ye, that difference is not obvious at all. It will be somewhat hard to name them differently unless we literally put machine/behavior in their respective names. So perhaps we should reach out for different entrypoints? or export machine-oriented helpers from the root entry and use `/behaviors` for those other ones?

In general `get*` clearly indicates that we are going to get something in return, while with `traverse*` I was somewhat expecting a provided callback to be called on each state/path/something (that's what often happens in AST-related projects like Babel and TypeScript)

packages/xstate-test/src/machine.ts
Andarist: if there are footguns with `createMachine` - shouldn't we disallow using `createMachine` together with `createTestModel`?

packages/xstate-graph/src/graph.ts
davidkpiano: Yeah, the naming can be improved. Will revise

packages/xstate-test/test/coverage.test.ts
Andarist: hm, i guess this somewhat depends on how do we define a transient state - is a `{ type: 'final' }` without the associated `onDone` considered to be transient?

packages/xstate-test/test/coverage.test.ts
```diff
@@ -0,0 +1,389 @@
...
+        coversAllStates({
+          filter: (stateNode) => {
+            return stateNode.key !== 'passthrough';
+          }
+        })
+      );
+    }).not.toThrow();
+  });
+
+  // https://github.com/statelyai/xstate/issues/981
```
Andarist, line 146: the linked issue doesnt seem to directly relate to the failure in this test - the issue mentions some invalid warnings to be logged when testing machines with final states, but this test fails on the coverage computation
```suggestion
```

packages/xstate-test/src/machine.ts
mattpocock: I don't think so, for migration reasons. I think we do want to allow cases where you create the machine in XState and pass it to `createTestModel` - even if there are footguns.

packages/xstate-test/src/machine.ts
mattpocock: It can work off the same definition, since this is a subset of the original createMachine.

We should add `createTestMachine` to the list of functions that `@xstate/machine-extractor` attempts to parse.

packages/xstate-test/src/TestModel.ts
mattpocock: I think it's fine given that all that users are using states for here are assertions. And not all the stuff you do in `states` will be assertions, it's also possible to use it for setup.

packages/xstate-test/src/TestModel.ts
mattpocock: See above comment

packages/xstate-test/test/plans.test.ts
davidkpiano: I could have sworn I had a test that had criterion skipping; @mattpocock this is for skipping coverage w/r/t coverage testing, not for skipping tests.

We talked about _not_ testing coverage though, which makes sense. If we agree on that (and remove `.testCoverage(...)`, then yes, we can remove coverage skipping.

packages/xstate-test/src/machine.ts
mattpocock: I'm comfortable leaving this the way it is. I don't think we're going to focus on other behaviours in this major version (I don't see the point in it atm) so we should assume that all users of `@xstate/test` are going to be using machines as test models.

If they want to use something else as a test model, they can use `new TestModel` directly.

packages/xstate-test/test/coverage.test.ts
mattpocock: No, I don't think so.

packages/xstate-test/src/TestModel.ts
mattpocock: I agree with Andarist, I don't mind this duplicated code since it's likely easier to read than the alternative proposed above.

packages/xstate-test/src/testUtils.ts
mattpocock: @davidkpiano This would mean it would be run as part of the tests, but I can add an ignore to the jest config.

packages/xstate-test/test/plans.test.ts
mattpocock: @davidkpiano I agree that we should remove `.testCoverage`, I'll add a PR to remove it.

packages/xstate-test/src/TestModel.ts
mattpocock: I'll add a PR to remove it, seems a bit redundant.

packages/xstate-test/src/TestModel.ts
mattpocock: cc @davidkpiano - I wonder if this still makes sense given that I changed this to paths only now.

packages/xstate-test/src/TestModel.ts
```diff
@@ -0,0 +1,491 @@
...
+        log: console.log.bind(console),
+        error: console.error.bind(console)
+      }
+    };
+  }
+  public static defaults: TestModelDefaults<any, any> = testModelDefaults;
+
+  constructor(
+    public behavior: SimpleBehavior<TState, TEvent>,
+    options?: Partial<TestModelOptions<TState, TEvent>>
```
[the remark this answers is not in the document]
mattpocock: What would you propose doing about this? Which approach do you prefer?

.changeset/great-spies-exist.md
mattpocock: Will add a PR changing this

packages/xstate-graph/src/graph.ts
```diff
@@ -67,59 +73,49 @@ export function getChildren(stateNode: AnyStateNode): AnyStateNode[] {
...
 }
 
-export function serializeState<TContext>(state: State<TContext, any>): string {
-  const { value, context } = state;
-  return context === undefined
-    ? JSON.stringify(value)
-    : JSON.stringify(value) + ' | ' + JSON.stringify(context);
+export function serializeMachineState(state: AnyState): SerializedState {
+  const { value, context, actions } = state;
+  return JSON.stringify({ value, context, actions }) as SerializedState;
```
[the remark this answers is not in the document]
mattpocock: cc @davidkpiano

.changeset/great-spies-exist.md
mattpocock: I'll handle this as part of the docs rewrite when we finalise the API, doesn't need to be included in this PR.

packages/xstate-graph/src/graph.ts
mattpocock: Added as task above.

packages/xstate-graph/src/types.ts
mattpocock: Added as a task above.

packages/xstate-graph/src/graph.ts
mattpocock: Done

packages/xstate-graph/src/graph.ts
```diff
@@ -393,64 +348,295 @@ export function toDirectedGraph(
...
+        };
+        queue.push([nextState, subEvent]);
+      }
+    }
+  }
+
+  return adj;
+}
+
+function resolveTraversalOptions<TState, TEvent>(
```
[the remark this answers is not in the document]
mattpocock: cc @davidkpiano

I consider this nit-level since it's not public facing.

packages/xstate-graph/src/graph.ts
mattpocock: Nit-level: I'm happy not to change this.

packages/xstate-graph/src/graph.ts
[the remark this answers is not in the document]
mattpocock: cc @davidkpiano

packages/xstate-graph/src/types.ts
mattpocock: Added as a task in the description

packages/xstate-graph/test/graph.test.ts
```diff
@@ -432,3 +571,102 @@ describe('@xstate/graph', () => {
...
+      serializeState: (v, e) =>
+        (JSON.stringify(v) + ' | ' + JSON.stringify(e)) as any
+    }
+  );
+
+  expect(getPathsMapSnapshot(a)).toMatchSnapshot();
+});
+
+describe('filtering', () => {
+  it('should not traverse past filtered states', () => {
```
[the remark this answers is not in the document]
mattpocock: cc @davidkpiano

packages/xstate-test/test/dieHard.test.ts
```diff
@@ -0,0 +1,348 @@
...
+      states: {
+        third: () => {
+          throw new Error('test error');
+        }
+      }
+    });
+
+    testModel
+      .getShortestPlansTo((state) => state.matches('third'))
+      .forEach((plan) => {
```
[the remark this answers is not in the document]
mattpocock: Outdated, I think, since we got rid of plans.

packages/xstate-test/test/forbiddenAttributes.test.ts
```diff
@@ -0,0 +1,49 @@
+import { createMachine, send } from 'xstate';
+import { createTestModel } from '../src';
+
+describe('Forbidden attributes', () => {
+  it('Should not let you declare invocations on your test machine', () => {
```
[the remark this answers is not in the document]
mattpocock: We can add this as a future improvement if folks are running into it.

packages/xstate-graph/test/graph.test.ts
mattpocock: Could you provide a suggestion?

.changeset/great-spies-exist.md
davidkpiano: Resolved to remove coverage functions

packages/xstate-graph/src/graph.ts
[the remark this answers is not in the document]
davidkpiano: https://github.com/statelyai/xstate/pull/3317

packages/xstate-graph/src/graph.ts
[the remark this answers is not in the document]
davidkpiano: Agree - will resolve in future PR

packages/xstate-graph/test/graph.test.ts
davidkpiano: We can discuss removing plans in a future PR, since the focus is on `xstate/test`

packages/xstate-graph/test/graph.test.ts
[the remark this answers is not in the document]
davidkpiano: Will add in future PR

packages/xstate-test/src/TestModel.ts
davidkpiano: Don't remove it; will add a test for it

davidkpiano: This can be removed

packages/xstate-test/src/TestModel.ts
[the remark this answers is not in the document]
davidkpiano: Agree that this needs to be consistent; will discuss offline and resolve in follow-up PR

packages/xstate-test/src/TestModel.ts
davidkpiano: We can change this to assume an array of paths and find the path satisfying the predicate with the least weight

packages/xstate-test/src/testUtils.ts
davidkpiano: We can do this in a follow-up PR

packages/xstate-test/test/coverage.test.ts
davidkpiano: We'll leave this comment in here and revisit it if someone reports it.

It was only an issue when test coverage was yelling that it was missing, but we're removing coverage testing.

packages/xstate-test/test/plans.test.ts
davidkpiano: Criterion removed

mattpocock (approved):

packages/xstate-test/src/testUtils.ts
Andarist: > @davidkpiano This would mean it would be run as part of the tests, but I can add an ignore to the jest config.

Tests in Jest (by default) are only executed for **all** files in `__tests__` and for all files with the `.test.(j|t)sx?` extension~. Our tests are located in a `test` directory and all of them have `.tests.tsx?` extension - so it's as easy as moving this to a test directory. This won't be recognized as a test file because it doesn't comply with its rules regarding the containing dir and the extension pattern

packages/xstate-graph/src/graph.ts
```diff
@@ -393,64 +351,297 @@ export function toDirectedGraph(
...
+    traversalOptions?.serializeState ??
+    defaultOptions?.serializeState ??
+    ((state) => JSON.stringify(state) as any);
+  return {
+    serializeState,
+    serializeEvent: serializeEvent as any, // TODO fix types
+    filter: () => true,
+    visitCondition: (state, event, vctx) => {
+      return vctx.vertices.has(serializeState(state, event));
+    },
```
Andarist, line 491: q: why this is customizable? it looks pretty internal at the first glance 🤔

packages/xstate-test/src/types.ts
```diff
@@ -1,144 +1,161 @@
...
-   * exec: async (page, event) => {
-   *   await page.type('.foo', event.value);
-   * }
-   * ```
-   */
-  exec?: EventExecutor<TTestContext>;
+export interface TestModelOptions<TState, TEvent extends EventObject>
+  extends TraversalOptions<TState, TEvent> {
+  // testState: (state: TState) => void | Promise<void>;
+  // testTransition: (step: Step<TState, TEvent>) => void | Promise<void>;
```
Andarist, line 119:
```suggestion
```

packages/xstate-graph/src/graph.ts
davidkpiano: Removed

quasor:
So excited for this, thanks David and team!

PR #3042, opened 2022-02-13
[core] Add `AnyStateConfig`

suerta-git:
This PR adds `AnyStateConfig` which can be used to type [persisted state as the document mentioned](https://xstate.js.org/docs/guides/states.html#persisting-state).


davidkpiano:
Great, thank you! Can you add a quick changeset by running `yarn changeset`?

This will be a **patch** change. Your message can be something like:

> Added the `AnyStateConfig` type, which represents any `StateConfig<...>`.

davidkpiano (approved):
Nice, love the example!

Issue #3033, opened 2022-02-10
Update @xstate/vue with 4.29.0 version on xstate


cedricradicia:
### Description

useMachine don't work as expected with the new version of XState (4.29.0), there are a typing issue.

### Expected result

useMachine have to work as the last react version (@xstate/react@2.0.0)

### Actual result

The following error happen:

```
Argument of type 'StateMachine<unknown, any, { type: "FOO"; }, any, BaseActionObject, ServiceMap, TypegenDisabled & { missingImplementations: { actions: never; delays: never; guards: never; services: never; }; } & AllowAllEvents & { ...; }>' is not assignable to parameter of type 'MaybeLazy<StateMachine<unknown, any, { type: "FOO"; }, any, BaseActionObject, ServiceMap, TypegenDisabled>>'.
  Type 'StateMachine<unknown, any, { type: "FOO"; }, any, BaseActionObject, ServiceMap, TypegenDisabled & { missingImplementations: { actions: never; delays: never; guards: never; services: never; }; } & AllowAllEvents & { ...; }>' is not assignable to type 'StateMachine<unknown, any, { type: "FOO"; }, any, BaseActionObject, ServiceMap, TypegenDisabled>'.
    Types of property 'getStateNodes' are incompatible.
      Type '(state: StateValue | State<unknown, { type: "FOO"; }, any, any, TypegenDisabled & { missingImplementations: { actions: never; delays: never; guards: never; services: never; }; } & AllowAllEvents & { ...; }>) => StateNode<...>[]' is not assignable to type '(state: StateValue | State<unknown, { type: "FOO"; }, any, any, TypegenDisabled>) => StateNode<unknown, any, { type: "FOO"; }, any, ServiceMap, TypegenDisabled>[]'.
        Types of parameters 'state' and 'state' are incompatible.
          Type 'StateValue | State<unknown, { type: "FOO"; }, any, any, TypegenDisabled>' is not assignable to type 'StateValue | State<unknown, { type: "FOO"; }, any, any, TypegenDisabled & { missingImplementations: { actions: never; delays: never; guards: never; services: never; }; } & AllowAllEvents & { ...; }>'.
            Type 'State<unknown, { type: "FOO"; }, any, any, TypegenDisabled>' is not assignable to type 'StateValue | State<unknown, { type: "FOO"; }, any, any, TypegenDisabled & { missingImplementations: { actions: never; delays: never; guards: never; services: never; }; } & AllowAllEvents & { ...; }>'.
[...]

Issue #3021, opened 2022-02-07
Bug: atomWithMachine typescript error after updating xstate to version 4.29.0


violabg:
### Description

I have updated xstate to version 4.29.0, and I'm getting a typescript error when adding a machine to atomWithMachine.

### Expected result

no typescript error

### Actual result

```
Argument of type '(get: Getter) => StateMachine<TimerContext, any, TimerEvents, any, BaseActionObject, ServiceMap, TypegenDisabled & { ...; } & AllowAllEvents & { ...; }>' is not assignable to parameter of type 'StateMachine<TimerContext, any, TimerEvents, any, BaseActionObject, ServiceMap, TypegenDisabled> | ((get: Getter) => StateMachine<...>)'.
  Type '(get: Getter) => StateMachine<TimerContext, any, TimerEvents, any, BaseActionObject, ServiceMap, TypegenDisabled & { ...; } & AllowAllEvents & { ...; }>' is not assignable to type '(get: Getter) => StateMachine<TimerContext, any, TimerEvents, any, BaseActionObject, ServiceMap, TypegenDisabled>'.
    Call signature return types 'StateMachine<TimerContext, any, TimerEvents, any, BaseActionObject, ServiceMap, TypegenDisabled & { ...; } & AllowAllEvents & { ...; }>' and 'StateMachine<TimerContext, any, TimerEvents, any, BaseActionObject, ServiceMap, TypegenDisabled>' are incompatible.
      The types of 'getStateNodes' are incompatible between these types.
        Type '(state: StateValue | State<TimerContext, TimerEvents, any, any, TypegenDisabled & { missingImplementations: { actions: never; delays: never; guards: never; services: never; }; } & AllowAllEvents & { ...; }>) => StateNode<...>[]' is not assignable to type '(state: StateValue | State<TimerContext, TimerEvents, any, any, TypegenDisabled>) => StateNode<TimerContext, ... 4 more ..., TypegenDisabled>[]'.
          Types of parameters 'state' and 'state' are incompatible.
            Type 'StateValue | State<TimerContext, TimerEvents, any, any, TypegenDisabled>' is not assignable to type 'StateValue | State<TimerContext, TimerEvents, any, any, TypegenDisabled & { missingImplementations: { actions: never; delays: never; guards: never; services: never; }; } & AllowAllEvents & { ...; }>'.
[...]

Issue #2984, opened 2022-01-28
Type error when using popular `createInterpreterContext` in v4.29.0, @xstate/react v2.0.0


josmithua:
**Description**
Type error when using popular `createInterpreterContext` util, after upgrading to `@xstate/react v2.0.0` and `xstate v4.29.0`.

Here is the full code that causes the error:

```tsx
import * as React from 'react';

import { useSelector } from '@xstate/react';
import { createMachine, Interpreter, InterpreterFrom } from 'xstate';

// error occurs on following line:
const _ = createInterpreterContext<InterpreterFrom<typeof fooMachine>>('Foo');

const fooMachine = createMachine(
  {
    // still an error if following line is commented out (different error message though):
    tsTypes: {} as import('./fooMachine.typegen').Typegen0,
    initial: 'foo',
    states: {
      foo: {
        entry: ['logFoo'],
      },
    },
  },
  {
    actions: {
      logFoo: () => {
        console.log('foo');
      },
    },
  }
);

function createInterpreterContext<TInterpreter extends Interpreter<any, any, any>>(displayName: string) {
  const [Provider, useContext] = createRequiredContext<TInterpreter>(displayName);

  function createUseSelector<Data>(selector: (state: TInterpreter['state']) => Data) {
    return () => {
      return useSelector(useContext(), selector);
    };
  }

  return [Provider, useContext, createUseSelector] as const;
}

function createRequiredContext<T>(displayName: string) {
  const context = React.createContext<T | null>(null);
  context.displayName = displayName;

  function useContext() {
    const ctx = React.useContext(context);
    if (!ctx) {
      throw new Error(`use${displayName} must be used inside ${displayName}Provider`);
    }
    return ctx;
  }

  return [context.Provider, useContext] as const;
}
```

Here is a screenshot with the error:

![Screen_Shot_2022-01-28_at_10 50 59](https://user-images.githubusercontent.com/3375071/151569118-6761915d-0dcc-4a17-8f50-d4e530e0f115.png)

**Expected Result**
Shouldn't be an error, unless there's a breaking change.

**Actual Result**
```
const fooMachine: StateMachine<unknown, any, AnyEventObject, any, BaseActionObject, ServiceMap, Typegen0 & {
    indexedActions: IndexByType<BaseActionObject>;
    indexedEvents: MergeWithInternalEvents<...>;
}>
[...]

PR #3051, opened 2022-02-15
Fixed type compatibility with typegen-less functions

Andarist:
fixes https://github.com/statelyai/xstate/issues/3033
fixes https://github.com/statelyai/xstate/issues/3021
fixes https://github.com/statelyai/xstate/issues/3052
and addresses https://github.com/statelyai/xstate/issues/2984

### DIFF
--- a/.changeset/long-insects-remember.md
+++ b/.changeset/long-insects-remember.md
@@ -1,0 +1,7 @@
+---
+'xstate': patch
+---
+
+Fixed type compatibility with functions accepting machines that were created before typegen was a thing in XState. This should make it possible to use the latest version of XState with `@xstate/vue`, `@xstate/react@^1` and some community packages.
+
+Note that this change doesn't make those functions to accept machines that have typegen information on them. For that the signatures of those functions would have to be adjusted.
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1005,7 +1005,12 @@ export interface StateMachine<
   TTypestate extends Typestate<TContext> = { value: any; context: TContext },
   TAction extends BaseActionObject = BaseActionObject,
   TServiceMap extends ServiceMap = ServiceMap,
-  TResolvedTypesMeta = TypegenDisabled
+  TResolvedTypesMeta = ResolveTypegenMeta<
+    TypegenDisabled,
+    TEvent,
+    TAction,
+    TServiceMap
+  >
 > extends StateNode<
     TContext,
     TStateSchema,
@@ -1484,7 +1489,7 @@ export interface StateConfig<TContext, TEvent extends EventObject> {
   machine?: StateMachine<TContext, any, TEvent, any, any, any, any>;
 }
 
-export type AnyStateConfig = StateConfig<any, AnyEventObject>
+export type AnyStateConfig = StateConfig<any, AnyEventObject>;
 
 export interface StateSchema<TC = any> {
   meta?: any;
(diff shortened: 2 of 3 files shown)


davidkpiano (approved):

.changeset/long-insects-remember.md
```diff
@@ -0,0 +1,7 @@
+---
+'xstate': patch
+---
+
+Fixed type compatibility with functions accepting machines that were created before typegen was a thing in XState. This should make it possible to use the latest version of XState with `@xstate/vue`, `@xstate/react@^1` and some community packages.
+
+Note that this change doesn't make those functions to accept machines that have typegen information on them. For that the signatures of those functions would have to be adjusted.
```
Andarist, line 7: This isn't possible because we are using `__TResolvedTypesMeta` phantom type [here](https://github.com/statelyai/xstate/blob/e75c47ad147453aa5ba7dd7723c21d30ef6af44f/packages/core/src/types.ts#L1067) and this makes a machine with typegen information incompatible with a machine that has typegen information disabled for it.

**Maybe, maybe** it would be possible to fix this by defaulting to `any` and pushing resolving logic for the `TypegenDisabled` case further down the road. This would require using `IsAny` and stuff like that and, in general, I've experienced `any`s wreaking havoc on the type machine. Maybe it can be wrangled but I'm somewhat scared of doing such a substantial change now cause while this type machinery works quite alright... it's somewhat fragile.

I think that, for the time being, this is a reasonable tradeoff for what the typegen enables us to do. It feels better to push it harder than to experiment with changes that might break it. I imagine that with some of the changes planned for v5 this might become much less of a problem - but also I can't be certain about it until we actually try to minimize the amount of generics etc.

davidkpiano: Do we want to create an `AnyTypegenMachine` type to help with this?

Andarist: As mentioned in the private conversation - this wouldn't work correctly because the whole type machinery is built on top of a type that has some specific structure. `ResolvedTypegenDisabled` is acting, sort of, like `UnknownTypegenMeta`. What I mean by this is that the typegen meta includes string literals etc when the `ResolvedTypegenDisabled` has just `string`, `Record<string, unknown>` etc. In a way probably this holds true:
```ts
const truthy: ResolvedTypegenMeta extends ResolvedTypegenDisabled ? true : false = true
```
This allows the whole machine after the initial resolution to work exactly the same - for the most part, I don't even need to check if the resolved type contains metadata created by the typegen or not, I don't branch the implementation based on this.

This is just part of the problem though. The first problem that would have to be addressed before even thinking about any kind of reactors here is related to `machine.withConfig`. This method utilizes the typegen information and might mark some implementations as **required**. If the consuming function doesn't specify the typegen slot anyhow we need to default it to something. Whatever we default it to though has to end up marking all implementations provided to `withConfig` as optional - because this is like the most "loose" shape.

So the consuming function that doesn't specify the typegen slot anyhow will always end up expecting a machine that has `.withConfig(options: EverythingOptional)` while the typegen-based provides something like `.withConfig(options: SomeMightBeRequired)`. Those types are just fundamentally incompatible because allowing this would be unsound, especially when we consider `strictFunctionTypes` (and I believe that it's our target to support this option, in a way - I don't intend to work on any special support for the opposite).

So it feels like the only viable solution here is to enforce the choice on those consuming functions:
1. they can either provide `any` for the typegen slot to say "I don't care about this as I don't do anything special with it"
2. or they can infer it or provide a generic for it to utilize the information or forward it to the result of their function or something

PR #3057, opened 2022-02-16
Run core tests using TS strict mode

Andarist:


packages/core/test/guards.test.ts
```diff
@@ -178,19 +178,22 @@ describe('guard conditions', () => {
               T1: [
                 {
                   target: 'B1',
-                  cond: (_state, _event, { state: s }) => s.matches('A.A1')
+                  cond: (_state: any, _event: any, { state: s }: any) =>
```
Andarist, line 181: Conditional transitions are acting weird when no generics are provided. Even adding something like this fixes the problem:
```js
schema: {
  context: {} as unknown
}
```

This could be investigated in the future.

packages/core/test/invoke.test.ts
```diff
@@ -69,7 +70,7 @@ const fetcherMachine = Machine({
       invoke: {
         src: fetchMachine,
         data: {
-          userId: (ctx) => ctx.selectedUserId
+          userId: (ctx: any) => ctx.selectedUserId
```
Andarist, line 73: places like this can't be typed in any better way because the `PropertyMapper` type is used here:
https://github.com/statelyai/xstate/blob/59fc815b288537e8ac1cf700a58ac541a95627fa/packages/core/src/types.ts#L366
and `PropertyMapper` looks like this:
https://github.com/statelyai/xstate/blob/59fc815b288537e8ac1cf700a58ac541a95627fa/packages/core/src/types.ts#L1319-L1327

The problem is that in here `TParams` are declared using `any` because we can't currently "source" the true type from anywhere. And since `PropertyMapper` includes a bit like this:
```
| TParams[K];
```
where `TParams` is effectively just `any`. And this `any` just downgrades the whole union to `any` because `T | any` is just `any`.

So TS actually can't provide a type for `ctx` here, even though the second member of the union could be used to infer this because that other member of the union makes it impossible.

packages/core/src/State.ts
```diff
@@ -215,7 +215,7 @@ export class State<
    * @param context
    */
   public static inert<TC, TE extends EventObject = EventObject>(
-    stateValue: State<TC, TE> | StateValue,
+    stateValue: State<TC, TE, any, any, any> | StateValue,
```
davidkpiano, line 218: Good candidate for `AnyState`

packages/core/test/actor.test.ts
```diff
@@ -1174,7 +1174,7 @@ describe('actors', () => {
       {
         actions: {
           setup: assign({
-            child: () => spawn(childMachine)
+            child: (_) => spawn(childMachine)
```
davidkpiano, line 1177: Why doesn't strict mode like empty arguments?

packages/core/src/State.ts
Andarist: Not really though, cause we'd lose the ability to pass `TC` and `TE` here. Unless maybe something like this would be implemented:
```diff
-type AnyState = State<any, any, any, any, any>
+type AnyState<
+  TContext = any,
+  TEvent extends EventObject = any,
+  TStateSchema extends StateSchema<TContext> = any,
+  TTypestate extends Typestate<TContext> = any,
+  TResolvedTypesMeta = any
+> = State<TContext, TEvent, TStateSchema, TTypestate, TResolvedTypesMeta>;
```

Although I would probably try to stay away from changing this across the codebase etc. This might become a refactor target for v5 branch but given that merging main to next will already be a nightmare I feel that it's better to hold off doing that for now

packages/core/test/actor.test.ts
Andarist: This is a long-standing issue with XState types - I don't understand it in full yet but TS is somehow inferring differently here when the parameter is used or not, even if it stays unused. This doesn't have that much in common with strict as it has to do with their inference algorithm. However, the strict mode changed assignability types related to function types and thus in here the inferred return stopped being assignable - the type that was previously computed here was already wrong here though.

davidkpiano (approved):

PR #3062, opened 2022-02-18
[@xstate/test][@xstate/graph] Fix `getAdjacencyMap` to ignore transitions for events not explicitly passed into the `options.events` 

Silverwolf90:
Wasn't entirely sure what the changeset should be -- this could be considered a breaking change, despite the original behavior being flawed.

Basically, `getAdjacencyMap` would always attempt to find adjacencies for **every** transition. It would first look in the `options.events` to determine the payload, but if none was found, then it would create a payload-less event and use that to find adjacencies.

This is problematic for a couple reasons:
1. We can't assume events can be run without a payload
2. The adjacency map is far more expensive to compute when you only need adjacencies for a subset of events (eg: `getPathFromEvents`

There are a couple threads in discord related to this:

**Expensive adjacency map**
https://discord.com/channels/795785288994652170/809564635614150686/943109510040551424

**getPlanFromEvents tried unwanted transitions**
https://discord.com/channels/795785288994652170/809564635614150686/943823319625519124

Anyway, curious to see what people think!

### DIFF
--- a/.changeset/cool-colts-jam.md
+++ b/.changeset/cool-colts-jam.md
@@ -1,0 +1,6 @@
+---
+'@xstate/graph': major
+'@xstate/test': minor
+---
+
+`getAdjacencyMap` will now only process events that are passed into `options.events`. Previously, it would attempt to process every event per state, and if there wasn't a case in the `options.events` then a payload-less event would be created -- which is problematic for events that required a payload and could lead to an invalid `context`. The consequence of this change is that `options.events` must always be explicitly passed into `getAdjacencyMap` (see test changes).
--- a/packages/xstate-graph/src/graph.ts
+++ b/packages/xstate-graph/src/graph.ts
@@ -130,17 +130,15 @@ export function getAdjacencyMap<
       nextEvents.map((nextEvent) => {
         const getNextEvents = events[nextEvent];
 
-        if (!getNextEvents) {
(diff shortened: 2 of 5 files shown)


Silverwolf90:
Perhaps this will be fixed as part of https://github.com/statelyai/xstate/pull/3036?

davidkpiano:
Hey, just getting back to this! Let's see if we can port some of this logic over to https://github.com/statelyai/xstate/pull/3036

Silverwolf90:
Apologies for the delayed reply!

> Hey, just getting back to this! Let's see if we can port some of this logic over to #3036

Is this not already incorporated? When I tested the PR locally (using `getPlanForEvents`) I didn't notice it to be particularly slow at all where the previous version could take 10+ seconds to build up adjacency map in some of our cases.

davidkpiano:
> Apologies for the delayed reply!
> 
> > Hey, just getting back to this! Let's see if we can port some of this logic over to #3036
> 
> Is this not already incorporated (in some form)? When I tested the PR locally (using `getPlanForEvents`) I didn't notice it to be particularly slow at all where the previous version could take 10+ seconds to build up adjacency map in some of our cases.

Potentially; I'll copy the tests over to verify.

davidkpiano:
This has been fixed in #3036

Issue #2533, opened 2021-08-19
Version 4.23.0 breaks typing of "$" prefixed service inside svelte component


DavKato:
**Given this**
```ts
const machine = model.createMachine(...)
const service = interpret(machine).start();
```

**Until 4.22**
```ts
const { someData } = $service.context // This returns the correct type
```

**From 4.23.0**
```ts
service.subscribe((state) => {
  state.context // This still returns a correct type
})
const { someData } = $service.context // $service becomes any
```

annaghi:
`service` is not a reactive store, you cannot use the `$` sign on it. `state` is the store:

```js
const { someData } = $state.context;
```

https://github.com/statelyai/xstate/blob/5da8724bb6c4dada5d6e8cc61e36c7e2c771838b/packages/xstate-svelte/src/index.ts#L62-L80

DavKato:
@annaghi Thank you for a quick response!
Actually I'm not refering to xstate-svelte, I'm just using a plain xstate.

Service here is just a xstate service like this.

```ts
const machine = model.createMachine(...)
const service = interpret(machine).start();
```

Service is subscribable and we still can access it's state with "$" syntax, only the typings is broken with 4.23.0.

annaghi:
@DavKato Oh, you are right! I was not aware of that the `service.subscribe` is compatible with the Svelte store contract. Sorry for my silly comment. :)

DavKato:
@annaghi No worries! I updated the issue to make the point clearer.

Andarist:
@DavKato could you prepare a runnable repro case? It would make it easier for us to jump straight into playing with this problem

DavKato:
I couldn't configure codesandbox, so here's a github repo.
https://github.com/DavKato/xstate-typefail

chanced:
I have a slightly different take on the svelte implementation here: https://github.com/chanced/xstate-svelte

I just spiked it out last night so it doesn't have tests but you can find a quick example of the usage here: https://github.com/chanced/xstate-svelte/blob/main/src/routes/index.svelte

If its something that you guys are interested in, I can work on a pull request with tests and such. Otherwise, I'll get around to testing and publishing it as a package. I'll be sure to annotate credit to the original as it is largely a fork.

DavKato:
I found out that reverting this commit solves this issue.
https://github.com/statelyai/xstate/pull/2371/files

davidkpiano:
> I found out that reverting this commit solves this issue.
[...]

PR #3063, opened 2022-02-18
Fixed a type compatibility with Svelte's readables

Andarist:
fixes https://github.com/statelyai/xstate/issues/2533

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,7 @@
     "test": "jest",
     "test:core": "npm test --prefix packages/core",
     "dev": "node ./scripts/dev.js",
-    "ci": "npm run build && npm run test",
     "changeset": "changeset",
-    "prerelease": "npm run build && npm run test",
     "release": "changeset publish",
     "version": "changeset version && node ./scripts/bump-peer-dep-ranges.js"
   },
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -33,7 +33,8 @@ import {
   StopActionObject,
   Subscription,
   AnyState,
-  StateConfig
+  StateConfig,
+  InteropSubscribable
 } from './types';
 import { State, bindActionToState, isStateConfig } from './State';
 import * as actionTypes from './actionTypes';
@@ -53,12 +54,11 @@ import {
   toEventObject,
   toSCXMLEvent,
   reportUnhandledExceptionOnInvocation,
-  symbolObservable,
   toInvokeSource,
   toObserver,
   isActor,
   isBehavior,
-  interopSymbols
+  symbolObservableRef
 } from './utils';
 import { Scheduler } from './scheduler';
 import { Actor, isSpawnedActor, createDeferredActor } from './Actor';
@@ -389,16 +389,16 @@ export class Interpreter<
     return this;
   }
   public subscribe(
+    observer: Observer<
+      State<TContext, TEvent, any, TTypestate, TResolvedTypesMeta>
+    >
+  ): Subscription;
+  public subscribe(
     nextListener?: (
       state: State<TContext, TEvent, any, TTypestate, TResolvedTypesMeta>
     ) => void,
     errorListener?: (error: any) => void,
     completeListener?: () => void
-  ): Subscription;
-  public subscribe(
-    observer: Observer<
-      State<TContext, TEvent, any, TTypestate, TResolvedTypesMeta>
-    >
   ): Subscription;
   public subscribe(
     nextListenerOrObserver?:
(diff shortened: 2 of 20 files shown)


.github/workflows/ci-checks.yml
```diff
@@ -0,0 +1,12 @@
+name: 'CI checks'
+runs:
+  using: 'composite'
```
Andarist, line 3: I had to add svelte-check to our CI jobs here and I've thought that it would be nice to experiment with those composite actions - we'll see how it goes, it's the first time when I've attempted to use this.

package.json
```diff
@@ -15,9 +15,7 @@
     "test": "jest",
     "test:core": "npm test --prefix packages/core",
     "dev": "node ./scripts/dev.js",
-    "ci": "npm run build && npm run test",
     "changeset": "changeset",
-    "prerelease": "npm run build && npm run test",
```
Andarist, line 20: "pre*" scripts dont work with all package managers, so I've decided to just move this directly to a github action

packages/core/test/types.test.ts
```diff
@@ -386,3 +387,23 @@ describe('Typestates', () => {
     expect(failed).toEqual({ result: none, error: 'oops' });
   });
 });
+
+describe('interpreter', () => {
+  it('should be convertable to Rx observable', () => {
```
Andarist, line 392: an explicit test covering for changes from: https://github.com/statelyai/xstate/pull/2371/files

packages/xstate-svelte/test/interpreterAsReadable.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
+  // this file acts as a type test
+  import { createMachine, interpret } from 'xstate';
```
Andarist, line 3: this is a new test for what is being fixed in this PR

packages/core/src/types.ts
```diff
@@ -1603,16 +1603,20 @@ export interface Subscription {
...
 
 export interface InteropObservable<T> {
-  [Symbol.observable]: () => Subscribable<T>;
+  [Symbol.observable]: () => InteropSubscribable<T>;
 }
 
-export interface Subscribable<T> {
+export interface InteropSubscribable<T> {
+  subscribe(observer: Observer<T>): Subscription;
+}
```
Andarist, line 1611: In a way, this breaks the previous contract exposed under `Symbol.observable` method. However, this is the only signature that is valid from Rx's PoV. The `InteropObservable` returns `Subscribable<T>`:
https://github.com/ReactiveX/rxjs/blob/67cb317a7a6b9fdbd3d2e8fdbc2ac9ac7e57179c/api_guard/dist/types/index.d.ts#L321-L323
and Subscribable is defined as follows:
https://github.com/ReactiveX/rxjs/blob/67cb317a7a6b9fdbd3d2e8fdbc2ac9ac7e57179c/api_guard/dist/types/index.d.ts#L690-L692

So, in fact, this interface only accepts observers - it doesn't accept separate listener functions for `next`, `error` and `complete`.

By implementing this change and using this as an explicit return type of the `Symol.observable` method on the `Interpreter` class we ensure that it's compatible with the Rx's signature for this:
https://github.com/statelyai/xstate/blob/d912a8d8c00d6dc00787235f31ae1344d8750269/packages/core/src/interpreter.ts#L1369-L1373

Thanks to that we can revert the reordering of the `subscribe`'s overloads from [this PR](https://github.com/statelyai/xstate/pull/2371/files) and fix the issue with Svelte's readables. This is because both libraries are using conditional types to match the input argument and to "extract" the emitted type and conditional types are only seeing the last overload.

So, in a sense, this is a hack and might break any minute... cause the whole thing is fitting both libraries thanks to the alignment of the stars and not because this is built on super solid principles. We trick Rx by only exposing a single overload with the observer as the parameter for its conditional type from the `Symbol.observable` method and we trick Svelte by reordering overloads so it only sees the one with callbacks parameters.

packages/core/src/interpreter.ts
```diff
@@ -1208,11 +1210,12 @@ export class Interpreter<
       id,
       send: (event) => receivers.forEach((receiver) => receiver(event)),
       subscribe: (next) => {
-        listeners.add(next);
+        const observer = toObserver(next);
+        listeners.add(observer.next);
```
Andarist, line 1214: this is a fix from the perspective of the `Symbol.observable` - since that method is accepting observers we convert the received callback (or observer) to an observer

Andarist:
CI is expected to fail right now. It rightfully reports errors in some svelte files related to `state.matches` typings - this is a new error from CI's perspective because we actually type-check those svelte files now. The error was there all along, it just slipped through the cracks because CI was not verifying those files.

I'll look into fixing this problem and I think it's OK to wait with this PR a little bit before I do so.

Andarist:
The CI should get unblocked in a sec since I've landed https://github.com/statelyai/xstate/pull/3065

davidkpiano (approved):

PR #3073, opened 2022-02-22
[core] Fix restarted root-level invocations

### LINKED ISSUE #3072: Bug: Root invocations should never be restarted
### Description

See this thread: https://discord.com/channels/795785288994652170/799416943324823592/945378092552179753

### Expected result

Root invocations are started when a machine is started, and only stopped when a machine is stopped.

### Actual result

Root invocations can be restarted on external root transitions (which arguably should not be allowed).

### Reproduction

https://codesandbox.io/s/unable-to-send-event-to-child-o5h24d?file=/src/index.ts:528-891

### Additional context

Working example: https://codesandbox.io/s/unable-to-send-event-to-child-forked-e6bsn0?file=/src/index.ts:529-914

davidkpiano:
External root-level transitions (which should be invalid – only internal root-level transitions are allowed) will no longer restart root-level invocations. See #3072 for more details.

Fixes #3072


packages/core/src/StateNode.ts
```diff
@@ -954,14 +955,22 @@ class StateNode<
 
     const isInternal = !!selectedTransition.internal;
 
-    const reentryNodes = isInternal
-      ? []
-      : flatten(allNextStateNodes.map((n) => this.nodesFromChild(n)));
+    const reentryNodes: AnyStateNode[] = [];
+
+    if (!isInternal) {
+      allNextStateNodes.forEach((n1) => {
```
davidkpiano, line 961: This is the actual change

Andarist:
I've taken a look at this from the SCXML perspective and I think that this is a good change. Let's start first with how this compares to the spec text.

We can read that `<scxml>` is never an explicit part of the configuration ([configuration definition](https://www.w3.org/TR/scxml/#PrelimaryDefinitions)). This could be read as if it is an implicit member of the configuration or just that it can't ever be exited and thus it's not worth keeping it in the configuration.

It's also worth noting that `entrySet` and `exitSet` are computed based on "transition domain". Only descendants of the domain can be exited and thus reentered (see the [computeExitSet](https://www.w3.org/TR/scxml/#computeExitSet)). For external transitions, this is based on the [`findLCCA`](https://www.w3.org/TR/scxml/#findLCCA) procedure that mentions this:
> The [Least Common Compound Ancestor]() is the <state> or <scxml> element s such that s is a proper ancestor of all states on stateList and no descendant of s has this property. Note that there is guaranteed to be such an element since the <scxml> wrapper element is a common ancestor of all states.

This clearly means that we can't "exit" the root "state". The difference between SCXML and XState though is that in XState we actually extend what the root (`<scxml>`) can do! In SCXML you can't define `<invoke>`, `<onentry>`, nor `<onexit>` on the `<scxml>` (those are not [valid children](https://www.w3.org/TR/scxml/#N1027F) for that element).

Ok, so I think we are in the agreement that this is something that we'd like to do. **However**, I wonder if we really should do this in v4. The current behavior was with us here for quite a long time and it's also not a deal-breaker, it's just something that we'd like to change - but by no means this is a pressing matter.

I know that we've defined the [semver policy about breaking changes](https://github.com/statelyai/xstate#breaking-changes) so we have some freedom to change this now. That doesn't mean though that we should introduce such changes lightly. So my question is - is the RoI of this change sufficiently high that we are OK with this "breaking change" (cause it definitely might be for some state machines out there, it's hardly an edge case)? Isn't this something that we should just change in v5?

packages/core/test/invoke.test.ts
```diff
@@ -2905,6 +2905,41 @@ describe('invoke', () => {
       done();
     }, 0);
   });
+
+  // https://github.com/statelyai/xstate/issues/3072
+  it('root invocations should not restart on external transitions', (done) => {
```
Andarist, line 2910: We should also add a separate test that would test, in a more explicit way, that the root state is never being reentered here. That `invoke`'s implementation is piggybacking on entry/exit actions is an implementation detail so I don't think that we should assume that this is also testing that more "basic" scenario.

packages/core/src/StateNode.ts
```diff
@@ -954,14 +955,22 @@ class StateNode<
 
     const isInternal = !!selectedTransition.internal;
 
-    const reentryNodes = isInternal
-      ? []
-      : flatten(allNextStateNodes.map((n) => this.nodesFromChild(n)));
+    const reentryNodes: AnyStateNode[] = [];
+
+    if (!isInternal) {
```
Andarist, line 960: Note that this algorithm is also broken in other subtle ways. States are not only reentered for external transitions - internal/external only change **what** states are potentially reentered.

For instance, on [this codesandbox](https://codesandbox.io/s/beautiful-benji-njkixn?file=/src/index.js:298-331) this entry action should technically be called twice. Even though the transition is internal, this target state should be reentered.

And I'm not saying that we should fix this right now. I'm just observing the fact. I somewhat don't see how fixing **non-critical** bugs in v4 gets us closer to anything. As mentioned before - I would rather halt the v4 development (just fix pressing issues) and focus on v5 (or on the "satellite" packages, as that development is mostly orthogonal)

.changeset/four-waves-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The `AnyStateNode` type, which represents a `StateNode` of any context, event, etc., has been introduced.
```
Andarist, line 5: This is a somewhat internal change - I'm not sure how relevant this is for our users and I would probably skip "documenting" it:
```suggestion
---
---
```

Andarist:
I've created https://github.com/statelyai/xstate/pull/3487 as a successor PR to this one. The new one is targeting the `next` branch

PR #3076, opened 2022-02-22
Fixed an issue with "maximum call stack size exceeded" errors being thrown when registering a machine with a very deep object in its context

SimeonC:
I found this error when using the code attached below in react. Running the code will hang the browser so a bit hard to setup a reproduction. I had a good debugging session on this and discovered that `__context`, `idMap` and a lot of other values (also nested values) sent through in machine have the DOM node ref referenced which is circular and breaks the serializer. It seems that what happens is that the serializer attempts to serialize my entire React tree as it's referenced on the DOM node.

```tsx
inspect({
  serialize: (key: string, value: unknown) => {
    if (key.match(/Ref$/g)) {
      console.log(`xstate inspector dom node "${key}"`, value);
      return `<<omitted dom node>>`;
    }
    return value;
  }
});
// Component
const baseElementRef = React.useRef<HTMLDivElement>(null);
const history = useHistory();
React.useEffect(() => {
  const actor = interpret(
    idlePrompterMachine.withContext({ baseElementRef })
  );
  actor.start();
  history.listen(() => actor.send({ type: 'Page Change' }));
}, []);
```


Andarist (changes requested):
Could you add a test case that would fail on the main branch?

We already have a basic one for circular context here:
https://github.com/statelyai/xstate/blob/f72691bbb6b929093e32e0e3a46fee1bbdf7c617/packages/xstate-inspect/test/inspect.test.ts#L5

So I wonder where exactly is the difference between this and your use case.

SimeonC:
@Andarist Thanks for pointing that out, you're correct it isn't a circular reference issue. I've updated the commit message and changelog to reflect this.

About building a reproduction, this doesn't occur in simple reproductions but only in the full app.

It seems that React adds a reference to the virtual DOM to the DOM Node, the virtual DOM references also have parent references so the serialiser attempts to stringify my entire react app via this reference. I have i18n files and context values with a ton of data in them and when it tries to go serialize that when starting the machine my browser hangs for several minutes.

Andarist:
Just to understand the issue in full - so it's not about crashing but about freezing the main thread for a very long time?

SimeonC:
I thought about it again - I realised there is a way we can cause a crash with this and I added a test for it.

In my app I assumed it crashed because everything including dev tools froze up and I couldn't manage to get anything out of it (or maybe I just wasn't patient enough). One time I did manage to get a pause in the sources debugger the current code was a good 200+ calls deep in the `decirc` function of the fast stringify.

packages/xstate-inspect/test/inspect.test.ts
```diff
@@ -141,4 +141,54 @@ describe('@xstate/inspect', () => {
       }
     });
   });
+
+  it('should handle using serializer to remove dangerous values', (done) => {
+    const circularStructure = {} as any;
+    let deepRef = circularStructure;
+    for (let i = 0; i < 1_000_000; i += 1) {
```
SimeonC, line 148: The actual depth errors start to occur is somewhere around about 5,275 on my machine. Not sure that that value would stay the same so I picked a really big one.

SimeonC:
@Andarist I created a reproduction sandbox here https://codesandbox.io/s/wispy-leftpad-72f7j7?file=/src/index.js

The reproduction is a bit different to what I was seeing locally, though it does cause noticable lag on the toggle button it doesn't crash the browser here which indicates that possibly my browser would have recovered if I'd given it enough time.

packages/xstate-inspect/test/inspect.test.ts
SimeonC: Did redo this back down to 20,000 so it's a bit faster but still shows an error.

Andarist:
Thanks, I will get to this next week.

SimeonC:
Awesome, thanks!

As a side note, I referenced this in one of my commits and got a comment from my co-worker that he doesn't use the devTools due to performance as well. His case is when there is a lot of data in context or a lot of events being fired. (He's been using xstate kinda like you would use redux for context).

.changeset/shy-chefs-chew.md
```diff
@@ -0,0 +1,6 @@
+---
+'@xstate/inspect': patch
+---
+
+Remove the selective stringify serialization for machine. The serialize function failed to be able to prevent serialization of extremely deep values, specifically inside context, that are referenced in multiple places inside the machine.
+This serialization was unpreventable and could cause the browser to hang for long periods of time.
```
Andarist, line 6:
```suggestion
author: @SimeonC 
author: @Andarist 

Fixed an issue with "maximum call stack size exceeded" errors being thrown when registering a machine with a very deep object in its context despite using a custom `serializer` capable of replacing it.
```

packages/xstate-inspect/test/inspect.test.ts
```diff
@@ -141,4 +141,46 @@ describe('@xstate/inspect', () => {
       }
     });
   });
+
+  it('should not crash when registering machine with very deep context when serializer manages to replace it', (done) => {
```
Andarist, line 145: I've analyzed this situation further and it turned out that this wasn't about circular references at all. The only required prerequisite for the crash was a very deep object in context.

packages/xstate-inspect/src/serialize.ts
```diff
@@ -20,13 +20,16 @@ export function selectivelyStringify<T extends object>(
...
+): string {
   return selectivelyStringify(state, ['context', 'event', '_event'], replacer);
 }
 
 export function stringifyMachine(
   machine: AnyStateMachine,
   replacer?: Replacer
 ): string {
-  return selectivelyStringify(machine, ['context'], replacer);
+  return selectivelyStringify(machine.definition, ['context'], replacer);
```
Andarist, line 34: I basically limit what is being stringified here as this should be everything that has to be potentially replaced. Previously we were passing the whole `machine` and it referenced context in a couple of properties other than `context` (such as `_context`, `config` and maybe something else too)

packages/xstate-inspect/src/serialize.ts
```diff
@@ -20,13 +20,16 @@ export function selectivelyStringify<T extends object>(
   });
 }
 
-export function stringifyState(state: AnyState, replacer?: Replacer): string {
+export function stringifyState(
+  { machine, configuration, ...state }: AnyState,
+  replacer?: Replacer
+): string {
```
davidkpiano, line 26:
```suggestion
  stateToStringify: AnyState,
  replacer?: Replacer
): string {
  const { machine: _machine, configuration: _configuration, ...state } = stateToStringify;
```

Alternative approach that doesn't destructure in the args (which might not look good for intellisense)

davidkpiano (approved):
Approved with tiny nit suggestion 👍

SimeonC:
Awesome, thanks for finishing this up!

Issue #3029, opened 2022-02-08
Bug: Generated types are setting matches as never after a conditional rendering on React


BrunoQuaresma:
### Description

Generated types are setting `matches` as `never` after a conditional rendering on React.

### Expected result

Do not mark `matches` as `never`.

### Actual result

Marking `matches` as `never`

<img width="889" alt="Screen Shot 2022-02-08 at 17 14 45" src="https://user-images.githubusercontent.com/3165839/153068973-99fe4007-9736-40c9-90b0-9c865db0cc33.png">

### Reproduction

https://codesandbox.io/s/react-typescript-forked-pk5z0?file=/src/App.tsx:2269-2334

### Additional context

I noticed this happens when there are nested conditionals and nested states.

Valkendorm:
I found that it happens when the `matches` call is nested in another `matches`. It's also only happening after the first string match, while objects matching will continue to work.

```ts
// This works
if (state.matches('topToo.nested')) return <div>{state.context.stuff}</div>;
if (state.matches('topToo.nestedToo')) return <div>{state.context.stuffToo}</div>;

// This works too
if (state.matches('top')) {
  if (state.matches({ top: 'nested' })) return <div>{state.context.stuff}</div>;
  if (state.matches({ top: 'nestedToo' })) return <div>{state.context.stuffToo}</div>;
}

// "Property 'matches' does not exist on type 'never'." after first string match
if (state.matches('topToo')) {
  if (state.matches('topToo.nested')) return <div>{state.context.stuff}</div>;
  if (state.matches('topToo.nestedToo')) return <div>{state.context.stuffToo}</div>; 
}

// "Property 'matches' does not exist on type 'never'." after first string match
if (state.matches('top')) {
  if (state.matches({ top: 'nestedToo' })) return <div>{state.context.stuffToo}</div>;
  if (state.matches({ top: 'nested' })) return <div>{state.context.stuff}</div>;
  if (state.matches('top.nested')) return <div>{state.context.stuff}</div>;
  if (state.matches('top.nestedToo')) return <div>{state.context.stuffToo}</div>; 
}

```

mattpocock:
@Valkendorm @BrunoQuaresma Are you on latest?

Andarist:
Yes, this still happens with 4.30.0. I've created a minimal repro case here:
[...]

PR #3077, opened 2022-02-22
Fixed an issue with nested state.matches calls causing state to be narrowed to `never` when the typegen was involved

Andarist:
fixes https://github.com/statelyai/xstate/issues/3029


Andarist:
It's utterly annoying but my local dev often caches TS-related stuff and I think that everything is OK... until I push out and CI verifies the whole without without the cache 🙈

davidkpiano (approved):

Issue #3028, opened 2022-02-08
Bug: Schemata + Typegen break string literals, event suggestions, and `ActorRef` assignments


christoph-fricke:
### Description

I noticed a few problem with schemata + typegen:

1. String literals unions (e.g. `"working" | "not-working"`) used in context are broadened to string, breaking selectors and other potential code that requires the string literal union.
2. Since 4.29 the result of spawning a machine is no longer assignable to an actor ref created with `ActorRefFrom`, which makes it nearly impossible to type the spawned machine for passing it around.
3. While preparing the code sandbox, I noticed that VS Code does not infer event types in the machine definition when schemata are used.

### Expected result

1. String literals unions should be typed as string literal unions and not `string`.
2. Spawned machines are assignable to their corresponding `ActorRef`.
3. VS Code should suggest event types in `<state>.on` like it does when events are specified through the generic.

### Actual result

1. Instead of preserving a literal union, the type is broadened to `string`.
2. A nasty TypeScript error appears.
3. VS Code only suggests `""` and `"*"`, not the actual event types.

### Reproduction

https://codesandbox.io/s/floral-bush-yxn10?file=/src/index.ts

### Additional context

In the sandbox, schemas are disabled by default and can be enabled by removing the comments (`createMachine` generics have to be removed as well). 

XState Version: 4.29.0

jonnyirwin:
I started changing my codebase over to typegen yesterday and was also struggling with typing errors on my `ActorRefFrom`. I assumed it was me missing something.

Andarist:
@christoph-fricke could you split this into 3 issues? being able to track those separately would make it easier to work with those.

For now, I've looked into the string literal case and I'm not yet sure what we can do about avoiding breaking changes. I already had a refactor plan for v5 that would just avoid this altogether but it's not an easy decision to use that strategy in v4. Potentially this could be fixed with overloads but overloads have their own set of problems that I'd like to avoid. We also could use `NoInfer` trick but that would break people relying on inference and using `config.context` to type this (instead of `config.schema.context`). 

[...]

PR #3087, opened 2022-02-24
Fixed an issue with `ActorRefFrom` not resolving the typegen metadata from machine types

Andarist:
Fixed the second issue reported here: https://github.com/statelyai/xstate/issues/3028

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1671,19 +1671,8 @@ export type ActorRefWithDeprecatedState<
   state: State<TContext, TEvent, any, TTypestate, TResolvedTypesMeta>;
 };
 
-export type ActorRefFrom<T> = T extends StateMachine<
-  infer TContext,
-  any,
-  infer TEvent,
-  infer TTypestate,
-  any,
-  any,
-  any
->
-  ? ActorRefWithDeprecatedState<TContext, TEvent, TTypestate>
-  : T extends (
-      ...args: any[]
-    ) => StateMachine<
+export type ActorRefFrom<T> = ReturnTypeOrValue<T> extends infer R
+  ? R extends StateMachine<
       infer TContext,
       any,
       infer TEvent,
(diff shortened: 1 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -1671,19 +1671,8 @@ export type ActorRefWithDeprecatedState<
...
 };
 
-export type ActorRefFrom<T> = T extends StateMachine<
-  infer TContext,
-  any,
-  infer TEvent,
-  infer TTypestate,
-  any,
-  any,
-  any
```
Andarist, line 1681: The issue was that we were not using `infer TResolvedTypesMeta` here and I could have fixed it with a two-line change (infer here + provide the result to StateMachine below). However, I've chosen to refactor this type to use `ReturnTypeOrValue<T> extends infer R` technique so we don't end up in such problems - as it ensures that we infer from direct types and factories in the very same way.

packages/core/test/types.test.ts
```diff
@@ -480,3 +482,17 @@ describe('interpreter', () => {
     });
   });
 });
+
+describe('spawn', () => {
+  it('spawned actor ref should be compatible with the result of ActorRefFrom', () => {
+    const createChild = () => createMachine({});
+
+    function createParent(_deps: {
+      spawnChild: () => ActorRefFrom<typeof createChild>;
```
Andarist, line 491: The issue was that in here we are unpacking the type from a factory but `spawn` below actually unpacks from a machine (since the result of a factory is passed to it). Since resolving from factories and types was not unified... it caused a type mismatch.

packages/core/test/types.test.ts
```diff
@@ -480,3 +482,17 @@ describe('interpreter', () => {
...
+    }) {}
+
+    createParent({
+      spawnChild: () => spawn(createChild())
+    });
```
Andarist, line 496: I wish I could write a better test case for this, but unfortunately, within `createMachine` call TS gets lost when computing variance types and it allows what shouldn't be allowed. Since this is a known problem and there is a fix for it I've rechecked with an insiders build but it didn't end up erroring there either. I've reported this as a potential issue here: https://github.com/microsoft/TypeScript/pull/45628#issuecomment-1049714681

davidkpiano (approved):

PR #3089, opened 2022-02-24
Fixed compatibility with Skypack

Andarist:
Basically, I just removed `keys` as we can easily use `Object.keys`. I've also inlined `flatten` into other packages since it's a one-liner.

Other things that caused "deopts" were moved to the root entry of XState, you can check out what exports have been added

### DIFF
--- a/packages/xstate-inspect/src/browser.ts
+++ b/packages/xstate-inspect/src/browser.ts
@@ -1,14 +1,16 @@
 import {
   ActorRef,
-  Interpreter,
-  interpret,
-  EventObject,
   EventData,
+  EventObject,
+  interpret,
+  Interpreter,
   Observer,
-  toActorRef
+  toActorRef,
+  toEventObject,
+  toObserver,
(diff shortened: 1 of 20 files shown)


packages/xstate-inspect/src/browser.ts
```diff
@@ -1,14 +1,16 @@
...
   Observer,
-  toActorRef
+  toActorRef,
+  toEventObject,
+  toObserver,
+  toSCXMLEvent,
+  XStateDevInterface
 } from 'xstate';
-import { XStateDevInterface } from 'xstate/lib/devTools';
-import { toSCXMLEvent, toEventObject, toObserver } from 'xstate/lib/utils';
```
Andarist, line 11: Such changes should be OK as per our semver policy: https://github.com/statelyai/xstate/tree/50c271dc1a1a05b035364f8247aa4d80d613864f#packages-1

davidkpiano (approved):
Very excited for this fix!

Issue #3026, opened 2022-02-07
Bug: actions not properly typed when using schema and assign


a-uz:
### Description

Given the following machine there are TypeScript errors for `addNumber` action and `event.number`.

```ts
import { assign, createMachine } from "xstate";

export const numbersListMachine = createMachine(
  {
    schema: {
      context: {} as { numbers: number[] },
      events: {} as { type: "ADD"; number: number }
    }
  },
  {
    actions: {
      addNumber: assign({
        numbers: (context, event) => {
          if (event.type !== "ADD") {
            return context.numbers;
          }

          return context.numbers.concat(event.number);
        }
      })
    }
  }
);
```

### Expected result

There shouldn't be TypeScript errors in this example machine.

### Actual result

Error for `addNumber` action:

```
Type 'AssignAction<{ numbers: number[]; }, EventObject>' is not assignable to type 'ActionObject<{ numbers: number[]; }, { type: "ADD"; number: number; }> | ((context: { numbers: number[]; }, event: { type: "ADD"; number: number; }, meta: ActionMeta<{ numbers: number[]; }, { type: "ADD"; number: number; }, BaseActionObject>) => void) | undefined'.
  Type 'AssignAction<{ numbers: number[]; }, EventObject>' is not assignable to type 'ActionObject<{ numbers: number[]; }, { type: "ADD"; number: number; }>'.
    Types of property 'exec' are incompatible.
      Type '((context: { numbers: number[]; }, event: EventObject, meta: ActionMeta<{ numbers: number[]; }, EventObject, BaseActionObject>) => void) | undefined' is not assignable to type '((context: { numbers: number[]; }, event: { type: "ADD"; number: number; }, meta: ActionMeta<{ numbers: number[]; }, { type: "ADD"; number: number; }, BaseActionObject>) => void) | undefined'.
        Type '(context: { numbers: number[]; }, event: EventObject, meta: ActionMeta<{ numbers: number[]; }, EventObject, BaseActionObject>) => void' is not assignable to type '(context: { numbers: number[]; }, event: { type: "ADD"; number: number; }, meta: ActionMeta<{ numbers: number[]; }, { type: "ADD"; number: number; }, BaseActionObject>) => void'.
          Types of parameters 'meta' and 'meta' are incompatible.
[...]

PR #3090, opened 2022-02-24
Fixed an issue with action objects not receiving correct event types

Andarist:
fixes https://github.com/statelyai/xstate/issues/3026


packages/core/src/Machine.ts
```diff
@@ -93,11 +93,18 @@ export function createMachine<
     TServiceMap,
     TTypesMeta
   >,
-  options?: InternalMachineOptions<
-    TContext,
-    TEvent,
-    ResolveTypegenMeta<TTypesMeta, TEvent, BaseActionObject, TServiceMap>
-  >
+  options?: TTypesMeta extends infer EvaluatedTypesMeta
```
Andarist, line 96: This somehow forces TS to "evaluate" the final~ type for `TTypesMeta` quicker. My hunch is that it couldn't be discovered soon enough without typegen because `tsTypes` is an optional property, so it was missing then TS tried to "explore" the stuff further instead of using the fallback (`TypegenDisabled` default) immediately and then it couldn't compute stuff further down the road.

By doing it we kinda force it to utilize that `TypegenDisabled` default and it can proceed computation with that additional knowledge and that in turns allow for other stuff to be computed correctly.

packages/core/src/typegenTypes.ts
```diff
@@ -165,7 +165,7 @@ type AllowAllEvents = {
 };
 
 export type ResolveTypegenMeta<
-  TTypesMeta extends TypegenConstraint,
+  TTypesMeta,
```
Andarist, line 168: `ResolveTypegenMeta` isn't supposed to be used publicly, so we can just trust its input. The problem was that after this `infer` trick above TS couldn't provide that the `EvaluatedTypesMeta` actually satisfies this constraint. An alternative approach to removing this generic would be to use `Cast<EvaluatedTypesMeta, TypegenConstraint>` at those "call sites". This seemed to be easier though.

Simplified demo of the problem:
[TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCoE8AOEDmEQDCA9iAM5hRyhjIDeAUMsgEQACrAHuXJAPRhZc+ZgC5kAIyJEANhDggA3PQC+9egOzIAShFIyAbtAA8qZBA6QQAE1JpBeQiXKVqAPmQBeNEvpWICaTgoFBgAVxAEMGASZEQkTDAAJhMzC3wbO2wHYjIKKnBXAApGWKgcMVNzSwzQGGhkAFF9OGlQnggrZAB+bV0DYyaWtsgrdzEQCEMoegBKMX0iYCslIA)

Andarist:
Ok, the CI has actually found some problems with this... so I'll have to convert this to a draft 😭

Andarist:
I think this might be a somewhat minimal repro for the problem:
[TS playground](https://www.typescriptlang.org/play?ts=4.7.0-dev.20220425#code/JYOwLgpgTgZghgYwgAgKIDcLgPICMBWECYyA3gFDLJgCeADhAFzIDOYUoA5gNzkC+5cgBMiAGzhQUtBsgAq9CJywBhAPYg2UOKBIBeOQqUhUIOLlEQhyAD4GGRgCLAWZi0N7kd0eEjuKsTi7mlmSUyABEAAKRAB5scJAA9NL+IOHM8KIsELwCwmISUgrIynBsADwAggA0yABCAHzI+pXIEDGQIEIs9cgA-MitzHW8IgjiktTFAApQqnTlsrUA0k36y20dWN3IANYQNKowcv1yANrLALrIzCAQmFAeKYPEwOoAYgCuIK-qixhYEjtTo7AE4AhEMBrZAACnugOYsjBYAAlM0muhVMB3IIvLBEChKr8QHhCMR-phwJsQT1kaTIU0KFQUsxNFxeFREolkAh1Ow4OgJMA4FS6KoWMAwG8QMh8J82MghKoICwqaAYNAwu0iMwiVKPt9iRTAQ0OcguTzVIKOCKSGKJfqZZIWJ9RGAeglkHQ5gwoLRkHAhEIAHKfAC2uE1nO5AH0kZSwIjkfxceBvATkABZRAAC1AEDUIBgwE45TC8cB1O2tIT9OI1XL8gYLEzEDAcGafkczlclnIjLCLAQOYgYbgzCZVDaCZYSYTZr4ZvdTZVfURK5bbbguUEz2zw-z2DojpYeulLDLVFkACUVapRJghBvW+2G1fkSxSvKuGf1D19LM8yLLeLD3o+z5brU4TwuAn5wN+ICcL+GjhA0b5yAAkl02pCB+naAQsN53g+lgQa+ESgCIMSWB+qH9p2k4XMgoB7AcRxyB+X4SohyEsJcurEnWYCXlOpQVARixYVRNEzrUFawVxP7EiwFyXGhaC1hCxANGEpr8B4eI+CgK5GCYvZWJOUSxPESQpEY6TUFAnw5Cm5DPCBYEQCZWAvnAImyGRHbAtWXYqH+-I6Oh8lAlsXQ1oCQn0fok4we6ik8cpzC3ryUBCOUbKIbUBWcHpVCUThH7rgom7tlWcWhcYpjBEIYRUAM2WqLl+XsFwcnIjpU43MgHVdcVfUJnpi6CO0Yp+oqBSTDAhqOjykgJBA+55nc-nInVoKaWSYBRYFe09N5ICFpo2hUvo52BOZ-YwmEvJFiWzCbfmhbFqW0VyYFaFhPMJ4Tq1AaZVmuaHse568eUHkkedvmLIF40mqVyACCizCYtiHhjBMKBLT8K1lBKnAgMaVLBfVdJaVCT1UKTJYgGGCKwqlc6AmiugYliLVYy8jpCZTUIeAga2QB9dwM6Eg0Wsu1UTnwAY9KQyAAOTRHE7a2YYWDqxkcBZCgfDoVQQ4jmOIODVQqWzmQytlGQUwMMw4SVA4DjhNwyAgOGkZQLc-vQBjZuh2EpthJOjPg9HU6BiGweByrZMgDCcIJtzA421O6cxogSDHkHEbQFnDsohngIAHR+yXUAomaOcWpE7oALQzZC7dQHMUCg4NecFxAResD1iFl6QfAV6lNdJw3fcYyiYeR1QmO8EAA)

Can't be completely sure until this gets fixed in TS though. I will be investigating this further.

Andarist:
That is wild... the workaround that I've discovered when debugging through TypeScript compiler... works. 

I feel like it's the greatest victory of man over a computer. Gonna be dancing to loud music today.

davidkpiano (approved):

PR #3095, opened 2022-02-26
fix example todo-mvc-react.

mar9000:
In this example the Todo final state is not declared `final` indeed I've noticed that Todos `state.children` keep growing while adding /removing Todo items. With just one line the problem get fixed.  
The problem is probably present also into the other todo-mvc examples.


davidkpiano (approved):

davidkpiano:
Thank you!

PR #3097, opened 2022-02-26
Fix `state.can()` from restored state

### LINKED ISSUE #3096: Bug: `interpreter.state.can()` always be `false` if the interpreter started with an initial state from DB
### Discussed in https://github.com/statelyai/xstate/discussions/3094

<div type='discussions-op-text'>

<sup>Originally posted by **suerta-git** February 26, 2022</sup>
### Description

I am storing states in DB as JSON string following doc's recommendation. And restore them like:

```ts
const state = JSON.parse(jsonString)
const service = interpret(machine).start(state)
```

Then I want to use `state.can()` to test an event, but its result always be `false`

```ts
service.state.can(validEvent)  // false
```

If the interpreter started with a non-parsed state, the result is correct.

### Expected result

`true` for a valid event

### Actual result

`false`

### Reproduction

https://codesandbox.io/s/wizardly-perlman-01g01l?file=/src/index.ts

### Additional context

_No response_</div>

davidkpiano:
Fixes #3096


davidkpiano:
@Andarist is this good to go?

packages/core/src/StateNode.ts
```diff
@@ -761,7 +762,8 @@ class StateNode<
       value: this.resolve(stateFromConfig.value),
       configuration,
       done: isInFinalState(configuration, this),
-      tags: getTagsFromConfiguration(configuration)
+      tags: getTagsFromConfiguration(configuration),
+      machine: (this.machine as unknown) as AnyStateMachine
```
Andarist, line 766: nit: right now `this.machine === this` so we could replace this with:
```suggestion
      machine: (this as unknown) as AnyStateMachine
```

packages/core/test/state.test.ts
```diff
@@ -905,6 +905,27 @@ describe('State', () => {
 
       expect(nextState.can({ type: 'NEXT' })).toBeTruthy();
     });
+
+    // https://github.com/statelyai/xstate/issues/3096
+    it('should work with restored state', () => {
```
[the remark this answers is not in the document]
Andarist: Only this one here depicts better what actually got reported - the fact that under the hood we call `machine.resolveState` could be considered an implementation detail.

packages/core/test/state.test.ts
```diff
@@ -905,6 +905,27 @@ describe('State', () => {
...
+        on: {
+          TOGGLE: {
+            actions: 'something'
+          }
+        }
+      });
+
+      const state = State.create(
+        JSON.parse(JSON.stringify(machine.initialState))
+      );
```
Andarist, line 921: this feels rather irrelevant for this test case - if we want to test that "machineless" state returns `false` then that should be moved to a dedicated test

.changeset/twenty-timers-cheat.md
```diff
@@ -0,0 +1,7 @@
+---
+'xstate': patch
+---
+
+State that is persisted and restored from `machine.resolveState(state)` will now have the correct `state.machine` value, so that `state.can(...)` and other methods will work as expected.
+
+See #3096 for more details.
```
Andarist, line 7:
```suggestion
State that is persisted and restored from `machine.resolveState(state)` will now have the correct `state.machine` value, so that `state.can(...)` and other methods will work as expected. See [#3096](https://github.com/statelyai/xstate/issues/3096) for more details.
```

davidkpiano: > q: do u want to make a rule that we should always reference the closed ticket in changesets?

Yes, this is a good idea

packages/core/src/StateNode.ts
davidkpiano: @Andarist There is a chance that `this` can refer to a `StateNode` instead of the root state node (machine)

Andarist: well, ye, but I can't imagine somebody doing something like this:
```js
machine.states.a.resolveState(/*...*/)
```

Either way is fine though

Issue #3115, opened 2022-03-04
@xstate/test


calvinchengx:
### Description

For `@xstate/test` both v2.0.0 and v2.0.1, this still happens for `xstate` v4.30.4

If I revert back to `@xstate/test` v2.0.0 and `xstate` v4.30.2, the test fails and complains that `getSimplePathPlans()` has a type error.

```bash
  ● Test suite failed to run


    TypeError: (0 , utils_1.keys) is not a function

      60 |   })
      61 |
    > 62 |   const testPlans = betaModel.getSimplePathPlans()
         |                               ^
      63 |
      64 |   testPlans.forEach((plan) => {
      65 |     describe(plan.description, () => {

      at getSimplePaths (node_modules/@xstate/graph/lib/graph.js:316:49)
      at TestModel.getSimplePathPlans (node_modules/@xstate/test/lib/index.js:157:54)
  ```

### Expected result

Works for v2.0.0 and v4.30.2.

I am expecting this to still work after upgrading to v2.0.1 (@xstate/test) and v4.30.4 (xstate)

### Actual result

```bash
Test suite failed to run

    TypeError: (0 , utils_1.keys) is not a function
```

### Reproduction

NA

mattpocock:
@calvinchengx Should be fixed in 4.30.3

Andarist:
@mattpocock this issue is specifically about this not being fixed by the latest release

mattpocock:
@Andarist Whoops

PR #3118, opened 2022-03-04
Fixed a bundling issue that prevented the `keys()` export to be preserved in the previous release

Andarist:
fixes https://github.com/statelyai/xstate/issues/3115


packages/core/rollup.config.js
```diff
@@ -48,9 +49,11 @@ const createUmdConfig = ({ input, output, target = undefined }) => ({
   ]
 });
 
+const npmInputs = glob.sync('src/!(scxml|invoke|model.types|typegenTypes).ts');
+
 export default [
   createNpmConfig({
-    input: 'src/index.ts',
+    input: npmInputs,
```
Andarist, line 56: Even though we were using `preserveModules` here the whole thing was still subject to tree-shaking and since `keys` was not really used by anything... it got removed by Rollup. By providing an array of inputs we let Rollup to know that we expect stuff to be importable directly from those entries so it has to preserve every public export of every input - and that makes `keys` to be preserved

packages/core/rollup.config.js
```diff
@@ -48,9 +49,11 @@ const createUmdConfig = ({ input, output, target = undefined }) => ({
   ]
 });
 
+const npmInputs = glob.sync('src/!(scxml|invoke|model.types|typegenTypes).ts');
```
Andarist, line 52: those were mainly type-only files and Rollup has been complaining about empty entries so I've filtered them out. The whole thing should still work correctly - and those files are still accessible at the type-level because `.d.ts` files are still produced for them

davidkpiano (approved):

PR #3124, opened 2022-03-07
docs: "东走" should be "动作"

JounQin:


Andarist (approved):
Based on Google Translate - LGTM 😅

PR #3126, opened 2022-03-07
Add `xstate.stop` event and fix exit handlers execution

### LINKED ISSUE #2057: Error invoke machine
Error! if the parent machine is stopped, and the invoke service has an exit - sandParent event, then the machines cannot be stopped.
```js
import { Machine, interpret, send, sendParent } from "xstate";

// Invoked child machine
const pongMachine = Machine({
  id: "pong",
  initial: "active",
  states: {
    active: {
      entry: sendParent("ENTRY_EVENT"),
      // The error is here!!!!!!!
      exit: sendParent("EXIT_EVENT"),
      on: {
        PING: {
          // Sends 'PONG' event to parent machine
          actions: sendParent("PONG", {
            delay: 1000
          })
        }
      }
    }
  }
});

// Parent machine
const pingMachine = Machine({
  id: "ping",
  initial: "active",
  states: {
    active: {
      invoke: {
        id: "pong",
        src: pongMachine
      },
      // Sends 'PING' event to child machine with ID 'pong'
      entry: send("PING", { to: "pong" }),
      on: {
        PONG: {
          actions: send("PING", {
            to: "pong",
            delay: 1000
          })
        }
      }
    }
  }
});

const service = interpret(pingMachine)
  .onTransition((s) => console.log(s))
  .start();
service.stop();
```

Please help me, I am using a library on a working project and cannot kill the running machines

https://codesandbox.io/s/xstate-typescript-template-forked-30nli?file=/src/index.ts:0-989

### LINKED ISSUE #2450: Sending events on `exit` throws if the machine is stopped
**Description**
The `send` action used as an exit action throws an error when the machine is stopped.

**Expected Result**
`send` in the `exit` actions should just send the event. If the target actor (e.g. the machine itself) is stopped already, the event should be ignored.
```javascript
  id: "child",
  initial: "Ready",
  states: {
    Ready: {
      entry: send("ENTERED"),
      exit: send("EXITED")
    }
  },
  on: {
    "*": {
      actions: (c, e) => console.log("event received", e)
    }
  }
});

  id: "root",
  initial: "Ready",
  states: {
    Ready: {
      invoke: {
        src: childMachine
      },
      on: {
        EXIT: "Done"
      }
    },
    Done: {
      type: "final"
    }
  }
});

const service = interpret(machine);
service.start();
service.send("EXIT"); // stops the child machine. The exit action in the child machine executes, but fails.
```

**Actual Result**
When the machine sends an event to itself when it was stopped, an error is thrown:
`TypeError
Cannot use 'in' operator to search for '$$type' in undefined
toSCXMLEvent
https://s3sgo.csb.app/node_modules/xstate/es/utils.js:693:38`

**Reproduction**
https://codesandbox.io/s/brave-frog-s3sgo

**Additional context**
Hint: When the send action is executed on exiting a state node, while the whole machine stops (e.g. on unmounting of the React component using the machine), the send action object is missing the internal `_event` property, which later causes the exception:
```javascript
// Interpreter.prototype.exec contains this line
this.send(sendAction._event); // _event is undefined
```

### LINKED ISSUE #2880: TypeError: Cannot read properties of undefined (reading 'name')
**Description**

When a child machine has been in a state with the `exit: sendParent` action set, if the parent machine stop, then the error will be thrown.

I guess this is because the parent has been destroyed before the child.

**Expected Result**

The interpreter should be stopped gracefully. (executed all the `exit` from all children before destroy parent?)

**Actual Result**

```
TypeError: Cannot read properties of undefined (reading 'name')
```

**Reproduction**

https://codesandbox.io/s/loving-sutherland-y3he8?file=/src/index.ts

**Additional context**

- https://github.com/statelyai/xstate/issues/2057

### LINKED ISSUE #3053: Bug: Sending events to an invoked child machine on `exit` throws when stopping machines
### Description

when a machine stops and has `send` action in its `exit` actions, the send action throws as there's no event to be sent

relates to: #2450 #2057 #2880

### Expected result

`exit` actions to be executed before machines are stopped

### Actual result

all ancestor machines are stopped except for the children of the state with `send` in the `exit` action list

when a machine sends an event to its invoked child on `exit` action, an error is thrown:
`TypeError Uncaught TypeError: Cannot read properties of undefined (reading 'name')` at `Interpreter.sendTo` on interpreter.js:134 

line 758 on `interpreter.ts` file where `event` is `undefined`

this happens on hot reload or when unmounting the context/component that created the machine 

when calling `machine.stop()` the thrown error is `TypeError
Cannot read properties of undefined (reading 'listeners')` at `Interpreter.stop`

### Reproduction

https://codesandbox.io/s/xstate-stop-parent-child-grandchild-tuv3nk?file=/src/App.tsx

### Additional context

xstate v4.30.0

### LINKED ISSUE #3221: Bug: Exit actions are not run when stop() is called on a child machine actor.
### Description

When a parent machine starts a child machine using spawn(), then later the parent machine calls stop() on the spawned child machine, exit actions specified on the child machine are not run if they are xstate action creators (eg: assign(), pure(),  sendParent(), stop() etc...)

### Expected result

I would expect the exit actions to run and still be recognized without warning.

### Actual result

The browser JS console shows the following warning message:
Warning: No implementation found for action type 'whatever the name of the action was'

The action is never run.

### Reproduction

https://codesandbox.io/s/hopeful-moore-wm86bi?file=/src/index.js

### Additional context

I think there are some legitimate use cases for action creators that run on machine exit, such as:
 - Updating the parent with sendParent() or pure() based on state internal to the child actor when it is closed by the parent,
-  Actions that are re-used both as normal transition actions internal to the child machine and child machine exit actions.

Andarist:
TODO:
- [ ] changesets
- [ ] establish how we rollout this, especially in consideration with the typegen (cc @mattpocock )
- [ ] discuss the minor TODO comments left in the code

fixes #2880 
fixes #2450
fixes #2057
fixes #3053
fixes #3221

cc @huan @ptaranto @rlaffers @Jimmy89 @erikras @AvivLocalize @AntonS86


packages/core/src/actions.ts
```diff
@@ -609,7 +609,7 @@ export function choose<TContext, TEvent extends EventObject>(
 
 export function resolveActions<TContext, TEvent extends EventObject>(
   machine: StateNode<TContext, any, TEvent, any, any, any>,
-  currentState: State<TContext, TEvent> | undefined,
+  currentState: State<TContext, TEvent, any, any, any> | undefined,
```
davidkpiano, line 612: Can use `AnyState` here, no? (nit)

Andarist: This would have a slightly different meaning though as we wouldn't require `State<...>` to be based on `TContext` and `TEvent`. I think it would be somewhat OK to change this *here* as `resolveActions` is really just a private helper but we couldn't make such a change in the whole codebase.

EIther way - I think we should leave such refactorings for the next branch after we freeze v4 development.

packages/core/src/interpreter.ts
```diff
@@ -568,30 +578,82 @@ export class Interpreter<
...
+          this.machine as any,
+          this.state,
+          this.state.context,
+          _event,
+          exitActions,
+          this.machine.config.preserveActionOrder
+        );
+
+        const newState = new State<TContext, TEvent, TStateSchema, TTypestate>({
+          value: this.state.value,
```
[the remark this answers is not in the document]
davidkpiano: I think we can deviate from SCXML here, since we assume the final state to be the state _before_ exiting all states of the machine completely, and having a state of "nothingness" (empty configuration, null state value) is not very practical, especially since `state.done === true` implicitly gives us that information already.

huan:
Thanks for creating this PR to fix the issue, looking forward to using the merged new version, and also the v5 as well! :smile:

packages/core/src/interpreter.ts
```diff
@@ -568,30 +578,82 @@ export class Interpreter<
...
+          context: updatedContext,
+          _event,
+          _sessionid: this.sessionId,
+          historyValue: undefined,
+          history: this.state,
+          // this filtering is somewhat questionable and we might reconsider this in the future
+          // we definitely don't want to execute raised events as the stop event is the last thing that should be processed by this machine
+          // the question is if we should allow sending events from here to other actors *while* preventing the parent that has stopped this machine from receiving an event from here
```
davidkpiano, line 627: I think we should allow events to be sent to other machines from here.

If the recipient is the parent, it should ignore all events after stopping (I _think_):

1. Parent starts "stop procedure" (🛑  no events can be received from this point)
2. Child receives `xstate.stop` event (🛑  no events can be received from this point)
3. Final actions are executed, which may include sending to parent
  - Any event sent to parent as an action is ignored _by parent_ (child doesn't handle this logic)
4. Procedure finished; machines terminated

packages/core/src/interpreter.ts
```diff
@@ -1018,6 +1080,9 @@ export class Interpreter<
     name: string,
     options?: SpawnOptions
   ): ActorRef<any> {
+    if (this.status !== InterpreterStatus.Running) {
```
davidkpiano, line 1083: This also should not do anything if the statis is `InterpreterStatus.Stopped`

Also, when will this occur (spawning before the interpreter is started) relevant to this issue?

packages/core/test/actions.test.ts
```diff
@@ -847,6 +847,254 @@ describe('entry/exit actions', () => {
...
+        invoke: child
+      });
+
+      const interpreter = interpret(parent);
+      interpreter.start();
+
+      expect(() => interpreter.stop()).not.toThrow();
+    });
+
+    it('send actions from exit handlers of a stopped child should not be executed', () => {
```
davidkpiano, line 888: This test should still pass, but by virtue of the parent _ignoring_ any events after being stopped, rather than filtering out `action.send` events.

packages/core/test/actions.test.ts
```diff
@@ -847,6 +847,254 @@ describe('entry/exit actions', () => {
...
+            }
+          }
+        }
+      });
+
+      const interpreter = interpret(parent).start();
+      interpreter.send({ type: 'STOP_CHILD' });
+    });
+
+    it('actors spawned in exit handlers of a stopped child should not be started', () => {
```
davidkpiano, line 920: Likewise, this test should still pass by virtue of _not_ spawning any actor after being stopped, rather than spawning a deferred actor

packages/core/test/actions.test.ts
```diff
@@ -847,6 +847,254 @@ describe('entry/exit actions', () => {
...
+    it('should clear all scheduled events when the interpreter gets stopped', () => {
+      const machine = createMachine({
+        on: {
+          INITIALIZE_SYNC_SEQUENCE: {
+            actions: () => {
+              // schedule those 2 events
+              service.send({ type: 'SOME_EVENT' });
+              service.send({ type: 'SOME_EVENT' });
+              // but also immediately stop *while* the `INITIALIZE_SYNC_SEQUENCE` is still being processed
+              service.stop();
```
davidkpiano, line 1003: So `service.stop()` takes precedence over `service.send()` events within the same "tick", if I'm understanding this right?

Is this because `service.stop()` is meant to be synchronous, and `service.send()` is (conceptually) async?

davidkpiano (commented):
Mostly looks good, just some comments/questions for clarification

packages/core/src/interpreter.ts
Andarist: > Any event sent to parent as an action is ignored by parent (child doesn't handle this logic)

Right, this is somewhat hard to handle right now though - I will take another look at this to see if I can make it happen.

packages/core/src/interpreter.ts
Andarist: Within this PR it's not as much about spawning before the interpreter is started but rather about spawning from within `exit` action where the interpreter is already stopped.

packages/core/test/actions.test.ts
Andarist: I probably agree with this, this would be better. Let's keep the discussion here: https://github.com/statelyai/xstate/pull/3126#discussion_r821626098

packages/core/test/actions.test.ts
Andarist: This would be problematic because some actions might refer to the "spawned" actor, like for example here:
```js
exit: [
  assign({
    childRef: () => spawn(child)
  }),
  send({ type: 'GRAB_SOME_DATA_FROM_ME' }, { to: ctx => ctx.childRef })
]
```

So we need to create **something** within this `spawn` call but also we shouldn't really `spawn` a real child. This also brings up the fact that we should introduce a warning like this to more places:
https://github.com/statelyai/xstate/blob/11d815ee92044e0fb8f4ce89a6e6b2f6cf8dd2b0/packages/core/src/interpreter.ts#L1416-L1422

Ideally, we'd always receive "lazy" entities from users within `assign`. This doesn't hold true right now and we allow eager entities but we should probably require all of them to be lazy. To avoid breaking changes we still need to process those eager entities but that should be accompanied by a dev-only warning. Thoughts?

packages/core/test/actions.test.ts
Andarist: If we just focus on the concept here then I think that my mental picture here is that `xstate.stop` is just a priority signal. And since we can have stuff in the "mailbox" while receiving this signal then we need to drop those items from the mailbox and process `xstate.stop` asap (after processing currently processed item, if applicable).

So conceptually it's more like:
- when we process an event new stuff can always arrive
- if we receive a regular event then we just append it to the mailbox
- if we receive a stop signal then we clear the mailbox and append this signal to the mailbox

So both are received and processed in an asynchronous manner. It's just that I didn't figure out a better way to push to this conceptual mailbox to represent this situation in the test code. Of course, this simplified situation isn't that realistic, but those sends and stop could be replaced with some calls that would end up calling those and that is, edge-casey, but somewhat realistic. This test case just represents a "flattened", to its bare bones, situation like this.

packages/core/test/actions.test.ts
davidkpiano: Yes, let's add a dev-only warning.

In v5 (the spawn improvements branch), this is fixed by making all actors "not started" at first, and then starting them with their given behavior, while using the exact same reference. All actors are lazy.

packages/core/test/actions.test.ts
davidkpiano: That makes sense - priority signals vs. messages 👍

packages/core/src/interpreter.ts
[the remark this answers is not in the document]
Andarist: I've adjusted the implementation slightly and now `state.done` is equivalent to `isInFinalState(state)` - so a stopped machine doesn't end up with `state.done === true`

packages/core/src/interpreter.ts
Andarist: I've managed to cover this use case, however, the implementation is slightly different. I'm filtering what is being sent out from a done/stopped machine (where the original suggestion was to reject the events on the receiver's end). I couldn't find a clear way in the parent to check if the received event was sent by one of its children. I don't think this makes any practical difference though

Andarist:
@davidkpiano it's ready for the re-review. I've addressed the important part of the feedback and gonna add a changeset tomorrow. Let me know if you think that there is anything else blocking this PR.

davidkpiano (approved):
No further blockers, looks good to me

.changeset/slow-cheetahs-try.md
```diff
@@ -0,0 +1,7 @@
+---
+'xstate': minor
```
Andarist, line 2: I've added a changeset here - @davidkpiano  maybe you could review the wording and the level of the detail included here

.changeset/slow-cheetahs-try.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+All `exit` actions in the machine will now be correctly resolved and executed when a machine gets stopped or reaches its final state. Previously the actions were not correctly resolved and that was leading to runtime errors.
```
davidkpiano, line 5:
```suggestion
All `exit` actions in the machine will now be correctly resolved and executed when a machine gets stopped or reaches its top-level final state. Previously, the actions were not correctly resolved and that was leading to runtime errors.
```

.changeset/slow-cheetahs-try.md
```diff
@@ -0,0 +1,7 @@
...
+
+To implement this fix in a reliable way a new special event has been introduced: `{ type: 'xstate.stop' }` and when machine stops its execution all exit handlers of its current state will be called with that event. You should always assume that an exit handler might be called with that event.
```
davidkpiano, line 7:
```suggestion
To implement this fix in a reliable way, a new internal event has been introduced: `{ type: 'xstate.stop' }` and when the machine stops its execution, all exit handlers of the current state (i.e. the active state nodes) will be called with that event. You should always assume that an exit handler might be called with that event.
```

.changeset/slow-cheetahs-try.md
davidkpiano: Some nits, LGTM

davidkpiano (approved):

dankreiger (approved):

davidkpiano:
@Andarist Can we merge this?

Andarist:
@davidkpiano just need to adjust the logic in how we generate typegen metadata so both could be released in tandem

Jimmy89:
@Andarist is there any ETA about when this change is being released?

PR #3133, opened 2022-03-08
fix: replace with getGlobal

fw6:
In some runtime environments(the Mini Program unique to China), self or global may be empty.

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -142,9 +142,9 @@ export class Interpreter<
         return clearTimeout(id);
       }
     } as Clock,
-    logger: global.console.log.bind(console),
+    logger: global?.console.log.bind(console),
     devTools: false
-  }))(typeof self !== 'undefined' ? self : global);
+  }))(getGlobal());
   /**
    * The current state of the interpreted machine.
    */
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -142,7 +142,7 @@ export class Interpreter<
         return clearTimeout(id);
       }
     } as Clock,
-    logger: global?.console.log.bind(console),
+    logger: global.console.log.bind(console),
     devTools: false
   }))(getGlobal());
   /**
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -131,7 +131,7 @@ export class Interpreter<
    * - `clock` uses the global `setTimeout` and `clearTimeout` functions
    * - `logger` uses the global `console.log()` method
    */
-  public static defaultOptions = ((global) => ({
+  public static defaultOptions = {
     execute: true,
     deferEvents: true,
     clock: {
@@ -142,9 +142,9 @@ export class Interpreter<
         return clearTimeout(id);
       }
     } as Clock,
-    logger: global?.console.log.bind(console),
+    logger: console.log.bind(console),
     devTools: false
-  }))(getGlobal());
+  };
   /**
    * The current state of the interpreted machine.
    */
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -142,7 +142,7 @@ export class Interpreter<
         return clearTimeout(id);
       }
     } as Clock,
-    logger: global.console.log.bind(console),
+    logger: global?.console.log.bind(console),
     devTools: false
   }))(getGlobal());
   /**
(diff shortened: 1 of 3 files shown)


davidkpiano:
This seems to be missing the `getGlobal()` implementation?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#search_for_the_global_across_environments

fw6:
> This seems to be missing the `getGlobal()` implementation?
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#search_for_the_global_across_environments

:construction: the Mini Program development framework don't provide `global/self`, only `window` can get [(framework app-service)](https://developers.weixin.qq.com/miniprogram/en/dev/framework/app-service/).  

🚑️  It should be better to replace that line of code with function `getGlobal`.

iShot2022-03-08 21 04 36

fw6:
> This seems to be missing the `getGlobal()` implementation?
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#search_for_the_global_across_environments

:construction: the Mini Program development framework don't provide `global/self`, only `window` can get [(framework app-service)](https://developers.weixin.qq.com/miniprogram/en/dev/framework/app-service/).  

🚑️  It should be better to replace that line of code with function `getGlobal`.

iShot2022-03-08 21 04 36

packages/core/src/interpreter.ts
```diff
@@ -142,9 +142,9 @@ export class Interpreter<
         return clearTimeout(id);
       }
     } as Clock,
-    logger: global.console.log.bind(console),
+    logger: global?.console.log.bind(console),
```
Andarist, line 145: Is this change required though? Are all standard~ global objects not available in this environment? Even `self` and `globalThis`?

fw6: my fault😢 . this change is not required. 
Only `self` and `globalThis` is undefined. `window` is available.

fw6: Restored that line. :fire:

fw6: `globalThis` may be `undefined`

image
1

Andarist: > window is available.

This is somewhat confusing as on the screenshot shared [here](https://github.com/statelyai/xstate/pull/3133#issuecomment-1061755747) it states that `window` shouldn't be available in this environment 🤔

fw6: replacing `global` with `window` or `globalThis` may be the right way, but why not use `getGlobal()` directly.

image

fw6: `global` isn't defined on this file, so it refers to `NodeJS.Global` here(that's the reason no error in TS)

Andarist: > replacing global with window or globalThis may be the right way, but why not use getGlobal() directly.

I think you might have misunderstood me - I'm totally on board with using `getGlobal` here. I'm just wondering what global object is gonna be available in your environment.

fw6: 😅  `global` don't have the `console` object
image

Andarist: That is... some bizarre env there 😅 The especially surprising bit is this:
```
window "[object Object]"
globalThis "[object Window]"
```

I'm still surprised that `window` is available here as that contradicts the documentation - but, of course, I believe the screenshot that you have shared just now. Is `globalThis.console` available there?

fw6: >  Is `globalThis.console` available there?

Yep.

Andarist: I've tweaked the fix and pushed it out to your branch. Could you check if this variant works for you?

packages/core/src/interpreter.ts
```diff
@@ -142,9 +142,9 @@ export class Interpreter<
         return clearTimeout(id);
       }
     } as Clock,
-    logger: global.console.log.bind(console),
```
Andarist, line 145: my rationale here is that the argument passed to this `bind` call wasn't looked up on the `global` variable anyway. This means that all our users were already accessing this in their respective global scopes automatically and it worked. it had to work as otherwise they wouldn't be able to run XState at all as it would throw `Uncaught ReferenceError: console is not defined` at the initialization time.

So I think that we don't have to resolve the global object here manually at all and we can just rely on the standard JS scope lookup.

packages/core/src/interpreter.ts
fw6: Everything works perfectly. Thank you so much :tada: :rocket:

PR #3136, opened 2022-03-08
Improved type safety for `spawn`ed actors

Andarist:
This is sort of a followup to https://github.com/statelyai/xstate/pull/3087 , I couldn't quite repro the problem back then with a bare `createMachine` call and it was haunting me cause I had to add this somewhat unusual test within that PR.

I've figured out **why** the `createMachine` variant was not erroring correctly and here is a PR to fix this.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -140,7 +140,7 @@ export type BaseAction<
   TAction extends BaseActionObject
 > =
   | SimpleActionsOf<TAction>['type']
-  | TAction
+  | (string extends TAction['type'] ? never : TAction)
   | RaiseAction<any>
   | SendAction<TContext, TEvent, any>
   | AssignAction<TContext, TEvent>
--- a/.changeset/lemon-penguins-relax.md
+++ b/.changeset/lemon-penguins-relax.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Improved type safety for `spawn`ed actors. Accidentally incompatible actors were being assignable to a context.
--- a/packages/core/test/types.test.ts
+++ b/packages/core/test/types.test.ts
@@ -531,4 +531,36 @@ describe('spawn', () => {
       spawnChild: () => spawn(createChild())
     });
   });
+
+  it('incompatible actor ref should not be accepted within assign actions', () => {
+    const child = createMachine({
+      schema: {
+        context: {} as {
+          counter: number;
+        }
+      }
+    });
+
+    const otherChild = createMachine({
+      schema: {
+        context: {} as {
+          title: string;
+        }
+      }
+    });
+
+    createMachine({
+      schema: {
+        context: {} as {
+          myChild: ActorRefFrom<typeof child>;
+        }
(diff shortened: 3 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -140,7 +140,7 @@ export type BaseAction<
   TAction extends BaseActionObject
 > =
   | SimpleActionsOf<TAction>['type']
-  | TAction
+  | (string extends TAction['type'] ? never : TAction)
```
Andarist, line 143: This is losing us some functionality when `TAction === BaseActionObject` (so... most of the time) as random **objects** won't be allowed anymore. Since this wasn't really type-safe anyway and that feature is little known - I'm OK with this.

davidkpiano: Can you show what kind of code would break (type-wise) with this change?

Andarist: Take a look here:
https://github.com/statelyai/xstate/blob/11d815ee92044e0fb8f4ce89a6e6b2f6cf8dd2b0/packages/core/test/actions.test.ts#L903-L907

In here, with this PR, we still allow `entry: 'definedAction'` but we don't allow `entry: { type: 'definedAction' }`. That is because this action is not really "defined" through `TAction`. Both flavors of defining actions are "unsafe" in the very same manner - but the string variant is very common and it doesn't collide with other builtin action objects (like `assign` or whatever) because a string and an object don't share any structure/property with each other. 

I believe that with typegen the "object variant" isn't even supported and we derive "TAction" only from the "string variant". Although we probably don't literally derive the `TAction`, it's just derived conceptually. This somewhat raises a question for the future - how one should define their actions? If actually defining `schema.actions` - should one be required to define there even "simple actions" (string only) or should they only need to define complex ones there and we should construct the final `TAction` based on `schema.actions` and the derived typegen data?

I think this issue here would also be fixed if we add "callable signatures" to our builtin action objects because then those "object variants" wouldn't match our bultin action objects. This is something I want to experiment with - but it's a big task and I'm not sure if I don't end up with some weird TS problems that might block this work. So given that... it's not certain when that other solution for the problem that is being fixed here can be delivered.

Another idea that just popped up to my mind... we could also try introducing brands to our action objects and maybe use something like `TAction & { __brand?: never }`. I wonder though - what's our end goal for managing/typing actions. I think that it makes sense to either:
- require everything to be declared up front
- only require "complex" ones to be declared and infer the rest

The first solution feels quite strict but OTOH this could be automated pretty neatly by the IDE extension and we wouldn't end up with magic rules that might be hard to grok for newcomers.

davidkpiano: It sounds like we should make an RFC for this. We still want to allow this:

```js
entry: [
  'definedAction', // simple action without parameters - allowed
  { type: 'greetAction', message: 'hello' }, // action with parameters - allowed
  'undefinedAction' // this should cause a type error - disallowed
]
```

Andarist:
superseded by https://github.com/statelyai/xstate/pull/4271

Issue #3138, opened 2022-03-09
Bug: useMachine actor and InterpreterFrom don't match with typegen


SimeonC:
### Description

The following code generates a typescript error when it shouldn't. (Typescript playground link has full TS example).

```tsx
  const Component = ({ actor }: { actor: InterpreterFrom<typeof machine>}) => null;

  const App = () => {
    const [state, send, service] = useMachine(() => machine.withConfig({ actions: {
      update: () => {}
    } }))
    return <Component actor={service} />
  }
```

### Expected result

No errors shown

### Actual result

Following typescript error was given;
```
Type 'Interpreter<{ expiry?: number | undefined; }, any, { type: "UPDATE"; }, { value: any; context: { expiry?: number | undefined; }; }, MarkAllImplementationsAsProvided<Typegen0 & { indexedActions: IndexByType<BaseActionObject>; indexedEvents: MergeWithInternalEvents<...>; }>>' is not assignable to type 'Interpreter<{ expiry?: number | undefined; }, any, { type: "UPDATE"; }, { value: any; context: { expiry?: number | undefined; }; }, Typegen0 & { indexedActions: IndexByType<BaseActionObject>; indexedEvents: MergeWithInternalEvents<...>; }>'.
  The types of 'machine.getStateNodes' are incompatible between these types.
    Type '(state: StateValue | State<{ expiry?: number | undefined; }, { type: "UPDATE"; }, any, { value: any; context: { expiry?: number | undefined; }; }, MarkAllImplementationsAsProvided<Typegen0 & { indexedActions: IndexByType<...>; indexedEvents: MergeWithInternalEvents<...>; }>>) => StateNode<...>[]' is not assignable to type '(state: StateValue | State<{ expiry?: number | undefined; }, { type: "UPDATE"; }, any, { value: any; context: { expiry?: number | undefined; }; }, Typegen0 & { indexedActions: IndexByType<...>; indexedEvents: MergeWithInternalEvents<...>; }>) => StateNode<...>[]'.
      Types of parameters 'state' and 'state' are incompatible.
[...]

PR #3139, opened 2022-03-09
`InterpreterFrom` and `ActorRefFrom` types used on machines with typegen data should now correctly return types with final/resolved typegen data

Andarist:
fixes https://github.com/statelyai/xstate/issues/3138

cc @SimeonC

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1690,7 +1690,7 @@ export type ActorRefFrom<T> = ReturnTypeOrValue<T> extends infer R
         TContext,
         TEvent,
         TTypestate,
-        TResolvedTypesMeta
+        MarkAllImplementationsAsProvided<TResolvedTypesMeta>
       >
     : R extends Promise<infer U>
     ? ActorRef<never, U>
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -1712,19 +1712,13 @@ export type InterpreterFrom<
...
-      infer TTypestate,
-      any,
-      any,
-      infer TResolvedTypesMeta
+  ? Interpreter<
+      TContext,
+      TStateSchema,
+      TEvent,
+      TTypestate,
+      MarkAllImplementationsAsProvided<TResolvedTypesMeta>
```
Andarist, line 1720: The crucial bit is to use `MarkAllImplementationsAsProvided` here.

It shouldn't be possible to "start"/create a service/actor without providing all implementations first. So this thing should be impossible to create as `interpret` won't allow it:
```ts
Interpreter<TContext, any, TEvent, any, TypegenMetaStillExpectingSomeImplementations>
```
Therefore, it makes sense to just mark those as provided here as wherever you might use `InterpreterFrom<typeof machine>` you should only be able to receive valid interpreters (and that is ones with all implementations marked as provided)

davidkpiano (approved):

PR #3140, opened 2022-03-09
Fixed an issue with interpreters started using a persisted state not being "resolved" in full

### LINKED ISSUE #3111: Bug: Delayed transitions stop working after state restoration.
### Description

When I reload the interpreter with a state, delayed transitions are no longer working.

It turns out that the machine instance gets modified/initialized when the interpreter starts (delayed transitions are transformed to entry/exit actions), but that does not hapen when it starts with the state argument.

Workaround so far is to call `machine._init()` which initializes the entry and exit actions.

### Expected result

After the interpreter was initialized to some arbitrary state, the delayed transitions should work as they do in normal conditions (started from an initial state).

### Actual result

The delayed transitions are ignored when the interpreter was started from some previously stored state.

### Reproduction

https://codesandbox.io/s/practical-river-qsthc6?file=/src/App.js

### Additional context

XState version 4.30.3

Andarist:
fixes #3111

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -519,6 +519,12 @@ export class Interpreter<
       return this;
     }
 
+    // yes, it's a hack but we need the related cache to be populated for some things to work (like delayed transitions)
+    // this is usually called by `machine.getInitialState` but if we rehydrate from a state we might bypass this call
+    // we also don't want to call this method here as it resolves the full initial state which might involve calling assign actions
+    // and that could potentially lead to some unwanted side-effects (even such as creating some rogue actors)
+    (this.machine as any)._init();
+
     registry.register(this.sessionId, this as Actor);
     this.initialized = true;
     this.status = InterpreterStatus.Running;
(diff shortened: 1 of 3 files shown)


packages/core/src/interpreter.ts
```diff
@@ -519,6 +519,12 @@ export class Interpreter<
       return this;
     }
 
+    // yes, it's a hack but we need the related cache to be populated for some things to work (like delayed transitions)
+    // this is usually called by `machine.getInitialState` but if we rehydrate from a state we might bypass this call
+    // we also don't want to call this method here as it resolves the full initial state which might involve calling assign actions
+    // and that could potentially lead to some unwanted side-effects (even such as creating some rogue actors)
+    (this.machine as any)._init();
```
Andarist, line 526: This should really only be called here: https://github.com/statelyai/xstate/blob/c3e2b75ea02622fbc788782ef96229ea90046d11/packages/core/src/StateNode.ts#L464-L468

but we are still blocked by supporting "getters" for targets so we can't call this in a constructor as the `machine` doesn't get yet returned to the caller at this point in time and thus a thing like this fails on reading from `undefined`:
https://github.com/statelyai/xstate/blob/c3e2b75ea02622fbc788782ef96229ea90046d11/packages/core/test/id.test.ts#L52-L54

davidkpiano (approved):

PR #3144, opened 2022-03-11
feat: @xstate/inspect support umd

lecepin:


packages/xstate-inspect/rollup.config.js
```diff
@@ -30,6 +34,34 @@ const createNpmConfig = ({ input, output }) => ({
...
+    input,
+    external: makeExternalPredicate(Object.keys(pkg.peerDependencies)),
+    output: {
+      file,
+      format: 'umd',
+      name,
+      globals: {
+        react: 'React',
+        xstate: 'XState',
+        '@xstate/fsm': 'XStateFSM'
```
Andarist, line 48: Does `@xstate/inspect` work with `@xstate`

Andarist, line 48: I don't think that `@xstate/inspect` is using `@xstate/fsm` anyhow - not the latter is compatible with the inspect package
```suggestion
```

lecepin: ok, fixed.

packages/xstate-inspect/package.json
```diff
@@ -38,23 +38,28 @@
...
+    "rollup-plugin-terser": "^5.1.2",
     "rollup-plugin-typescript2": "^0.30.0",
     "typescript": "^4.5.2",
     "ws": "^8.4.0",
     "xstate": "*"
   },
   "peerDependencies": {
     "@types/ws": "^8.0.0",
-    "ws": "^8.0.0"
+    "fast-safe-stringify": "^2.0.7",
```
Andarist, line 55: This should be moved back to dependencies as the `createUmdConfig` should "bundle" this in the final file.

PR #3148, opened 2022-03-12
[v5] Explicit spawn

davidkpiano:
This PR makes `spawn` more explicit and introduces the following changes:

- The external (magical) `spawn` function and related `CapturedState` code has been removed
- Spawning in initial context now comes from an object passed into initial context:

```diff
-import { spawn } from 'xstate';

// ...

-context: () => ({
+context: ({ spawn }) => ({
  ref: spawn(someBehavior)
}),
```

- Spawning in `assign` actions now comes from the 3rd "meta" argument:

```diff
-import { spawn } from 'xstate';

// ...

actions: assign({
- ref: (ctx, e) => {
+ ref: (ctx, e, { spawn }) => {
    return spawn(someBehavior);
  }
})
```

- Spawning actors from string references is now possible:

```js
actions: assign({
  ref: (ctx, e, { spawn }) => {
    return spawn('someBehavior');
  }
}),

// ...

{
  actors: {
    someBehavior: (ctx, e) => /* return a Behavior */
  }
}
```

- Event observables work as before, sending events to the parent:
```js
invoke: {
  src: () => fromEventObservable(() => value$.pipe(map(val => ({ type: 'VAL', val })))),
},
on: {
  VAL: {
    actions: (ctx, e) => { console.log(e.val) }
  }
}
```

- Observables with `fromObservable` deliver values rather than events:
```diff
 invoke: {
-  src: () => fromObservable(() => value$.pipe(map(val => ({ type: 'VAL', val })))),
+  src: () => fromObservable(() => value$),
+  onEmit: {
+    actions: (ctx, e) => { console.log(e.data) }
+  }
 },
-on: {
-  VAL: {
-    actions: (ctx, e) => { console.log(e.val) }
-  }
}
```

This PR closes STA-1269


christoph-fricke:
> * Spawning actors from string references is now possible:
> 
> ```js
> actions: assign({
>   ref: (ctx, e, { spawn }) => {
>     return spawn('someBehavior');
>   }
> }),
> 
> // ...
> 
> {
>   actors: {
>     someBehavior: (ctx, e) => /* return a Behavior */
>   }
> }
> ```

Is there a reason for naming the config object `actors` instead of `behaviors`? It only contains behaviors that are used to spawn actors, so I would expect it to be named `behaviors`. Storing behaviors under an actor key might lead to confusion around behaviors and actors?!

davidkpiano:
> > * Spawning actors from string references is now possible:
> > 
> > ```js
> > actions: assign({
> >   ref: (ctx, e, { spawn }) => {
> >     return spawn('someBehavior');
> >   }
> > }),
> > 
> > // ...
> > 
> > {
> >   actors: {
> >     someBehavior: (ctx, e) => /* return a Behavior */
> >   }
> > }
> > ```
> 
> Is there a reason for naming the config object `actors` instead of `behaviors`? It only contains behaviors that are used to spawn actors, so I would expect it to be named `behaviors`. Storing behaviors under an actor key might lead to confusion around behaviors and actors?!

Good feedback; open to renaming this to `behaviors`

packages/core/test/typegenTypes.test.ts
```diff
@@ -475,7 +475,7 @@ describe('typegen types', () => {
     interpret(machine);
   });
 
-  it('should not allow to create an actor out of a machine with missing implementations', () => {
+  it.skip('should not allow to create an actor out of a machine with missing implementations', () => {
```
Andarist, line 478: gonna take look into this

packages/core/test/assign.test.ts
```diff
@@ -324,7 +324,22 @@ describe('assign meta', () => {
     expect(nextState.context).toEqual({ count: 5 });
   });
 
-  it('should not provide the state from initial state', () => {
+  // TODO: determine what should be the correct behavior
+  it.skip('should not provide the state from initial state', () => {
```
Andarist, line 328: we need to double-check what this is about

packages/core/test/actor.test.ts
```diff
@@ -1151,10 +1254,12 @@ describe('actors', () => {
 
   it('should be able to spawn callback actors in (lazy) initial context', (done) => {
     const machine = createMachine<{ ref: ActorRef<any> }>({
-      context: () => ({
-        ref: spawnCallback((sendBack) => {
-          sendBack({ type: 'TEST' });
-        })
+      context: ({ spawn }) => ({
+        ref: spawn(
+          createDeferredBehavior(() => (sendBack) => {
```
Andarist, line 1259: maybe we can rethink the name? i guess "callback" was pretty broad, but the same could be said about deferred and i guess people might have got used to "callback actors"

packages/core/test/actions.test.ts
```diff
@@ -1765,29 +1764,47 @@ describe('sendTo', () => {
...
-      }),
-      entry: sendTo(
-        (ctx) => ctx.child as any,
-        (ctx) => ({ type: 'EVENT', count: ctx.count })
-      )
+    let createCount = 0;
+
+    const parentMachine = createMachine({
+      context: ({ spawn }) => {
+        // TODO: clean up
```
Andarist, line 1778: you said it...

packages/core/src/types.ts
```diff
@@ -1735,6 +1746,7 @@ export interface ActorRef<TEvent extends EventObject, TEmitted = any>
   getSnapshot: () => TEmitted | undefined;
   stop?: () => void;
   toJSON?: () => any;
+  parent?: ActorRef<any, any>;
```
Andarist, line 1749: I would probably only expose this within the actor's implementation - and not on the outside. So people shouldn't be able to do something like that:
```js
sendTo(ctx => ctx.actorRef.parent.parent.parent.parent)
```

packages/core/src/types.ts
```diff
@@ -1678,7 +1686,10 @@ export type Observer<T> =
       next?: (value: T) => void;
       error?: (err: any) => void;
       complete: () => void;
-    };
+    }
+) & {
+  done?: (doneData: T) => void;
```
Andarist, line 1691: to be discussed later

packages/core/src/stateUtils.ts
```diff
@@ -1582,7 +1582,7 @@ export function resolveMicroTransition<
     return inertState as any;
   }
 
-  const children = !currentState._initial ? { ...currentState.children } : {};
```
Andarist, line 1585: q: why this was removed? won't it result in mutability bugs?

packages/core/src/interpreter.ts
```diff
@@ -815,20 +803,14 @@ export class Interpreter<
...
-export function interpret<
-  TContext extends MachineContext,
-  TEvent extends EventObject = EventObject,
-  TResolvedTypesMeta = TypegenDisabled
->(
-  machine: AreAllImplementationsAssumedToBeProvided<TResolvedTypesMeta> extends true
-    ? StateMachine<TContext, TEvent, any, any, TResolvedTypesMeta>
-    : 'Some implementations missing',
+export function interpret<TMachine extends AnyStateMachine>(
+  // machine: AreAllImplementationsAssumedToBeProvided<TResolvedTypesMeta> extends true
```
Andarist, line 807: 😢

packages/core/src/interpreter.ts
```diff
@@ -402,7 +385,7 @@ export class Interpreter<
 
     const resolvedState =
       initialState === undefined
-        ? this.initialState
```
Andarist, line 405: any particular reason behind the change here?

Andarist: I think this was important because the `initialState` should be cached at the interpreter level

packages/core/src/behaviors.ts
```diff
@@ -462,7 +474,7 @@ export function createMachineBehavior<
     get initialState() {
       resolvedMachine =
         resolvedMachine || (isFunction(machine) ? machine() : machine);
-      return resolvedMachine.initialState; // TODO: this should get from machine.getInitialState(ref)
```
Andarist, line 465: it depends on how this is called by other things but it might be also important for this to be cached - need to recheck that

packages/core/src/behaviors.ts
```diff
@@ -295,28 +308,19 @@ export function createPromiseBehavior<T, TEvent extends EventObject>(
           resolvedPromise.then(
             (response) => {
               if (!canceled) {
-                parent?.send(
```
Andarist, line 298: we need a test to verify that it's possible to send an event to a parent from within that start signal handler

packages/core/src/StateMachine.ts
```diff
@@ -291,30 +312,22 @@ export class StateMachine<
...
-      this.preInitialState,
+      preInitialState,
       undefined
     );
-    return macrostep(nextState, null as any, this) as State<
-      TContext,
-      TEvent,
-      TResolvedTypesMeta
-    >;
+    nextState.actions.unshift(...preInitialState.actions);
```
Andarist, line 329: did we miss some actions before? do we have a new test verifying this change?

packages/core/src/ObservableActorRef.ts
```diff
@@ -1,6 +1,5 @@
 import { EventObject, ActorRef, Behavior, SCXML, ActorContext } from './types';
 import { Actor } from './actor';
-import { CapturedState } from './capturedState';
 import { symbolObservable, toSCXMLEvent } from './utils';
 
 export class ObservableActorRef<TEvent extends EventObject, TEmitted>
```
Andarist, line 5: I think we need to establish the vocabulary here - we need to clarify what's:
- behavior
- actor
- actor ref
- observable actor ref
- spawned actor ref

+ we need to establish what's the key on the implementations object, is it `implementations.actors`? or `implementations.behaviors`?

Andarist:
An additional thing that we've discussed at the meeting was that we could drop the `invokePromise` etc functions and only support this:
```js
{
  actors: {
    myActor: () => createPromiseBehavior()
  }
}
```
Let the value in the actor map be a regular, plain-function, factory for a behavior.

packages/core/test/typegenTypes.test.ts
davidkpiano: The test actually passes

packages/core/test/actor.test.ts
davidkpiano: Renamed to `createCallbackBehavior`

packages/core/src/ObservableActorRef.ts
davidkpiano: There should be only two concepts:

- behavior
- actor ref (created from a behavior)

packages/core/test/assign.test.ts
davidkpiano: The ~~new~~ improved behavior is that the `state` provided to initial actions is the _pre-initial_ state.

packages/core/src/stateUtils.ts
davidkpiano: f0de1acffe0c0471ebac622fdcebaaa2eb770e92

packages/core/src/interpreter.ts
davidkpiano: Reverted: eab43a59bab69666479e402ee9f4f599e80b1e36

packages/core/src/StateMachine.ts
davidkpiano: This is implicitly tested by numerous tests:

```
 FAIL  packages/core/test/actions.test.ts (14.654 s)
  ● sendTo › should be able to send an event to an actor

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1733 | 
      1734 | describe('sendTo', () => {
    > 1735 |   it('should be able to send an event to an actor', (done) => {
           |   ^
      1736 |     const childMachine = createMachine<any, { type: 'EVENT' }>({
      1737 |       initial: 'waiting',
      1738 |       states: {

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actions.test.ts:1735:3)
      at Object.<anonymous> (packages/core/test/actions.test.ts:1734:1)

  ● sendTo › should be able to send an event from expression to an actor

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1758 |   });
      1759 | 
    > 1760 |   it('should be able to send an event from expression to an actor', (done) => {
           |   ^
      1761 |     const childMachine = createMachine<any, { type: 'EVENT'; count: number }>({
      1762 |       initial: 'waiting',
      1763 |       states: {

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actions.test.ts:1760:3)
      at Object.<anonymous> (packages/core/test/actions.test.ts:1734:1)

 FAIL  packages/core/test/actor.test.ts (20.186 s)
  ● actors › should only spawn an initial actor once when it synchronously responds with an event

    expect(received).toEqual(expected) // deep equality

    Expected: "done"
    Received: "testing"

      715 | 
      716 |     const service = interpret(testMachine).start();
    > 717 |     expect(service.state.value).toEqual('done');
          |                                 ^
      718 |   });
      719 | 
      720 |   it('should spawn null actors if not used within a service', () => {

      at Object.<anonymous> (packages/core/test/actor.test.ts:717:33)

  ● actors › with behaviors › should work with a promise behavior (reject)

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1207 |     });
      1208 | 
    > 1209 |     it('should work with a promise behavior (reject)', (done) => {
           |     ^
      1210 |       const errorMessage = 'An error occurred';
      1211 |       const countMachine = createMachine<{
      1212 |         count: ActorRefFrom<Promise<number>>;

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actor.test.ts:1209:5)
      at Suite.<anonymous> (packages/core/test/actor.test.ts:1129:3)
      at Object.<anonymous> (packages/core/test/actor.test.ts:601:1)

  ● actors › should be able to spawn callback actors in (lazy) initial context

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1293 |   });
      1294 | 
    > 1295 |   it('should be able to spawn callback actors in (lazy) initial context', (done) => {
           |   ^
      1296 |     const machine = createMachine<{ ref: ActorRef<any> }>({
      1297 |       context: ({ spawn }) => ({
      1298 |         ref: spawn(

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actor.test.ts:1295:3)
      at Object.<anonymous> (packages/core/test/actor.test.ts:601:1)

  ● actors › should be able to spawn machines in (lazy) initial context

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1320 |   });
      1321 | 
    > 1322 |   it('should be able to spawn machines in (lazy) initial context', (done) => {
           |   ^
      1323 |     const childMachine = createMachine({
      1324 |       entry: sendParent('TEST')
      1325 |     });

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actor.test.ts:1322:3)
      at Object.<anonymous> (packages/core/test/actor.test.ts:601:1)

Summary of all failing tests
 FAIL  packages/core/test/actions.test.ts (14.654 s)
  ● sendTo › should be able to send an event to an actor

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1733 | 
      1734 | describe('sendTo', () => {
    > 1735 |   it('should be able to send an event to an actor', (done) => {
           |   ^
      1736 |     const childMachine = createMachine<any, { type: 'EVENT' }>({
      1737 |       initial: 'waiting',
      1738 |       states: {

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actions.test.ts:1735:3)
      at Object.<anonymous> (packages/core/test/actions.test.ts:1734:1)

  ● sendTo › should be able to send an event from expression to an actor

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1758 |   });
      1759 | 
    > 1760 |   it('should be able to send an event from expression to an actor', (done) => {
           |   ^
      1761 |     const childMachine = createMachine<any, { type: 'EVENT'; count: number }>({
      1762 |       initial: 'waiting',
      1763 |       states: {

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actions.test.ts:1760:3)
      at Object.<anonymous> (packages/core/test/actions.test.ts:1734:1)

 FAIL  packages/core/test/actor.test.ts (20.186 s)
  ● actors › should only spawn an initial actor once when it synchronously responds with an event

    expect(received).toEqual(expected) // deep equality

    Expected: "done"
    Received: "testing"

      715 | 
      716 |     const service = interpret(testMachine).start();
    > 717 |     expect(service.state.value).toEqual('done');
          |                                 ^
      718 |   });
      719 | 
      720 |   it('should spawn null actors if not used within a service', () => {

      at Object.<anonymous> (packages/core/test/actor.test.ts:717:33)

  ● actors › with behaviors › should work with a promise behavior (reject)

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1207 |     });
      1208 | 
    > 1209 |     it('should work with a promise behavior (reject)', (done) => {
           |     ^
      1210 |       const errorMessage = 'An error occurred';
      1211 |       const countMachine = createMachine<{
      1212 |         count: ActorRefFrom<Promise<number>>;

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actor.test.ts:1209:5)
      at Suite.<anonymous> (packages/core/test/actor.test.ts:1129:3)
      at Object.<anonymous> (packages/core/test/actor.test.ts:601:1)

  ● actors › should be able to spawn callback actors in (lazy) initial context

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1293 |   });
      1294 | 
    > 1295 |   it('should be able to spawn callback actors in (lazy) initial context', (done) => {
           |   ^
      1296 |     const machine = createMachine<{ ref: ActorRef<any> }>({
      1297 |       context: ({ spawn }) => ({
      1298 |         ref: spawn(

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actor.test.ts:1295:3)
      at Object.<anonymous> (packages/core/test/actor.test.ts:601:1)

  ● actors › should be able to spawn machines in (lazy) initial context

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      1320 |   });
      1321 | 
    > 1322 |   it('should be able to spawn machines in (lazy) initial context', (done) => {
           |   ^
      1323 |     const childMachine = createMachine({
      1324 |       entry: sendParent('TEST')
      1325 |     });

      at new Spec (node_modules/@jest/core/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (packages/core/test/actor.test.ts:1322:3)
      at Object.<anonymous> (packages/core/test/actor.test.ts:601:1)
```

packages/core/src/StateMachine.ts
```diff
@@ -275,13 +294,15 @@ export class StateMachine<
    * The initial state _before_ evaluating any microsteps.
```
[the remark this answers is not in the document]
davidkpiano: Added here: ecf98536a4d9c9b59b877a585623a19d9eddeca1

packages/core/src/behaviors.ts
davidkpiano: How would we test this?

packages/core/src/ObservableActorRef.ts
davidkpiano: The interpreted machine can be a "system actor ref" which _can_ be stopped from the outside, and we can have child actors only stoppable by a special "stop" signal from the parent, so basically:

```ts
interface SystemRef extends ActorRef {
  stop: () => void; // stops the system
}
```

Info on stopping actors: https://doc.akka.io/docs/akka/current/typed/actor-lifecycle.html

Info on terminating an actor system: https://doc.akka.io/docs/akka/current/general/actor-systems.html#terminating-actorsystem

packages/core/src/interpreter.ts
Andarist: fixed this one

packages/core/test/types.test.ts
```diff
@@ -384,17 +381,3 @@ describe('interpreter', () => {
     });
   });
 });
-
-describe('spawn', () => {
-  it('spawned actor ref should be compatible with the result of ActorRefFrom', () => {
```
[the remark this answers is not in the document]
Andarist: I brought it back

packages/core/src/types.ts
Andarist: Right, but that still means that the `parent`/`self` would be available for actors themselves but without the `.parent` property being available on whatever gets access to actor refs.

I've added a TODO comment about this.

packages/core/src/behaviors.ts
Andarist: i've just added caching for now plus a TODO comment to recheck this later

packages/core/test/actions.test.ts
```diff
@@ -1765,29 +1763,36 @@ describe('sendTo', () => {
...
+      .onTransition((state) => {
+        state.actions.forEach((action) => {
+          if (action.type === 'xstate.invoke') {
+            expect((action as InvokeActionObject).params.ref).toBe(
+              state.children['child']
+            );
+          }
+        });
+      })
+      .start();
```
Andarist, line 1795: I'm not sure how this assertion related to the intention of this test

```suggestion
    interpret(parentMachine).start();
```

packages/core/src/behaviors.ts
Andarist: I probably meant a more general situation - not specific to a promise behavior. But also, this has been refactored for now so let's not bother with it.

packages/core/src/actor.ts
```diff
@@ -83,63 +77,6 @@ export function fromMachine<
...
-}
-
-export function spawnCallback(callback: InvokeCallback, name?: string) {
-  return spawn(
-    createDeferredBehavior(() => callback),
-    name
-  );
-}
-
-export function spawnFrom<TEvent extends EventObject, TEmitted>(
```
Andarist, line 120: This comment applies to `createBehaviorFrom` but that isn't displayed anywhere in the diff so I'm just using this location for this comment :P

We need the argument to be lazy, some entity~ types are lazy by default, others - not so much. To make non-lazy entity lazy we need to wrap it in a factory function, like here:
```js
spawnFrom(() => promise)
```
The problem with this approach is that we can't introspect the final "type" in the `spawnFrom` because we can't invoke this function and check it.

So I'm unsure how we can provide a `*From` API for this.

Andarist (commented):
This needs a changeset and possibly we should revisit existing changesets to align them with the changes in here

packages/core/src/actor.ts
davidkpiano: We can omit this for now.

packages/core/package.json
```diff
@@ -12,9 +12,9 @@
   "files": [
     "dist",
     "actor",
+    "actors",
```
Andarist, line 15: it seems to be super confusing that we have `xstate/actor` and `xstate/actors` here, the `/actor` one should be removed and it's utils should be moved to the root entry or something (it only contains `isActorRef` and `toActorRef`). Alternatively, those can be moved to `xstate/actors` but I probably prefer the first suggestion

davidkpiano: Yes, agreed. Will make this change

packages/core/src/actors.ts
```diff
@@ -504,25 +391,98 @@ export function createBehaviorFrom<TEvent extends EventObject>(
 ): Behavior<TEvent, undefined>;
 export function createBehaviorFrom(entity: Spawnable): Behavior<any, any> {
```
Andarist, line 392: as mentioned somewhere - this should be removed for now because we should only accept lazy entities and with promises we can't detect what's enclosed in the function, an "aggregate" factory-like this would be nice but it would require some implementation adjustments to be made - in the current form this is just not meeting the requirements

packages/core/src/ObservableActorRef.ts
```diff
@@ -1,41 +1,47 @@
...
   }
   public stop() {
-    this.actor.stop();
+    this.mailbox.clear();
+    this.mailbox.enqueue(stopSignal);
     return this;
   }
   public subscribe(observer) {
-    return this.actor.subscribe(observer);
+    return this.behavior.subscribe?.(observer) ?? nullSubscription;
```
Andarist, line 41: Also, a potential refactor item - but I don't fully see why a behavior would have a `subscribe` method (especially that a behavior, at least in Akka terms, should be pure and `subscribe` requires keeping an internal state).

Note that even if `behavior` would have a `subscribe` method we can't truly just proxy to it here. This exact implementation made me to skip some `@xstate/react` tests because subscribers are called from within a behavior before we have a chance to assign the updated `this.current` in this very class here. This leads to "stale" notifications being emitted

packages/core/src/ObservableActorRef.ts
```diff
@@ -1,41 +1,47 @@
...
   implements ActorRef<TEvent, TEmitted> {
   private current: TEmitted;
   public deferred = true;
   private context: ActorContext<TEvent, TEmitted>;
-  private actor: Actor<TEvent, TEmitted>;
   public name: string;
+  private mailbox = new Mailbox(this._process.bind(this));
 
-  constructor(behavior: Behavior<TEvent, TEmitted>, name: string) {
+  constructor(public behavior: Behavior<TEvent, TEmitted>, name: string) {
```
Andarist, line 18: q: why `behavior` is `public`? shouldn't it be hidden from the consumer?

packages/core/src/actors.ts
```diff
@@ -504,25 +391,98 @@ export function createBehaviorFrom<TEvent extends EventObject>(
...
   throw new Error(`Unable to create behavior from entity`);
 }
 
-interface SpawnBehaviorOptions {
+interface CreateActorRefOptions {
   id?: string;
   parent?: ActorRef<any>;
 }
+
+export function createActorRef<TEvent extends EventObject, TEmitted>(
```
Andarist, line 417: q: what makes `createActorRef` special - I mean, why is this only used by `useSpawn`? Is there a potential on unifying this with something?

packages/core/src/actors.ts
davidkpiano: Removed

packages/core/src/ObservableActorRef.ts
davidkpiano: Will address this in a separate PR (and remove `behavior.subscribe`).

The tentative plan is to expose `actorContext.emit(data)` since the emitted data isn't necessarily tied to the `state`, which may be private for certain use-cases.

packages/core/src/ObservableActorRef.ts
davidkpiano: Fixed in [ac21c9f](https://github.com/statelyai/xstate/pull/3148/commits/ac21c9fe18e3ff3c77dc1c7ffe0979255fe5b4ad)

packages/core/src/actors.ts
davidkpiano: I'm thinking we should actually remove `useSpawn` in the v5-compatible versions of `@xstate/react` and `@xstate/vue` - it doesn't make sense to have them.

packages/core/src/interpreter.ts
```diff
@@ -679,15 +654,32 @@ export class Interpreter<
...
               },
               complete: () => {
+                this.send(
+                  toSCXMLEvent(doneInvoke(id) as any, {
+                    origin: ref
+                  })
+                );
                 /* ... */
+              },
+              done: (doneData) => {
```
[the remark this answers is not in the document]
davidkpiano: This has been removed.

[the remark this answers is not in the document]
davidkpiano: There are 2 main concepts: 
- **Passive:** observing an actor
- **Active:** an actor directly sending an event (to parent, or other actors)

Observed values should not be explicit events. Here's how I categorize the different actors in this PR:

- **Promise**: observed
- **Observable**: observed
- **Callback**: sends + receives explicit events
- **Machine**: sends + receives explicit events, can also be observed

There was an impedance mismatch in v4 with promises only emitting raw values (`Promise<any>`), whereas observables were expected to emit events (`Observable<EventObject>`). Now, promises and observables are treated similarly, where a promise emits at most 1 value and an observable can emit infinitely many values (not events). These emitted values are converted to `{ type: 'xstate.emit.[id]', data }`, which work with `onEmit: { ... }`.

packages/core/src/actors.ts
```diff
@@ -295,12 +191,6 @@ export function createPromiseBehavior<T, TEvent extends EventObject>(
...
               if (!canceled) {
-                parent?.send(
-                  toSCXMLEvent(doneInvoke(actorContext.name, response) as any, {
-                    origin: actorContext.self
-                  })
-                );
-
                 observers.forEach((observer) => {
                   observer.next?.(response);
                   observer.complete?.();
```
Andarist, line 196: In response to an event an observer might get unsubscribed and we shouldn't call `complete` then
```suggestion
                  if (observers.has(observer)) observer.complete?.();
```

packages/core/src/types.ts
```diff
@@ -1663,6 +1685,7 @@ export declare namespace SCXML {
 }
 
 // Taken from RxJS
+// TODO: fix complete types
```
Andarist, line 1688: Since `complete` has been removed this comment became redundant.
```suggestion
```

packages/core/src/actors.ts
```diff
@@ -253,7 +253,7 @@ export function fromObservable<
         observable = lazyObservable();
         subscription = observable.subscribe({
           next: (value) => {
-            sendNext(toSCXMLEvent(value, { origin: actorContext.self }));
+            sendNext(value);
```
Andarist, line 259: if `value` is not being boxed here in an event - why the same change has not been done to `sendError`?

packages/core/src/actors.ts
davidkpiano: Nice catch on the edge-case!

packages/core/src/actors.ts
davidkpiano: It should be; I will make that change (didn't get to it yet 😅)

packages/core/src/interpreter.ts
```diff
@@ -668,9 +668,28 @@ export class Interpreter<
...
+                  typeof data === 'object' &&
+                  data !== null &&
+                  'type' in data
+                ) {
+                  this.send(data as TODO);
+                } else {
+                  this.send(
+                    toSCXMLEvent(
+                      {
+                        type: `xstate.emit.${id}`,
```
Andarist: Could we say that this replaces `.sync` + `xstate.update`? It seems that with this change they are almost the same - while the `xstate.emit.*` is not sent conditionally so the parent will get "slammed" with events (I wouldn't bother optimizing this right now, just mentioning this fact as this will produce a lot of noisy events in the event log)

Andarist: on top of that - the emitted value is really unanimous to a snapshot, could we rename "emit" to snapshot in all the related things?

I know that `getSnapshot` might sometimes return `T | undefined` while the emitted value is always `T` but that seems to be a really small difference that might be handled differently without having more names for almost the same thing

davidkpiano: A snapshot is the latest emitted value:

1. emitted
2. emitted
3. emitted
4. emitted ← snapshot (latest)

The difference is that the snapshot is _taken_ and the emitted value is _given_, so it might be more confusing to combine these two concepts.

davidkpiano: Re: noisiness, we can do some optimizations in the future:

- Short-circuiting unknown events if there are no enabled wildcard transitions and completely ignoring them in the machine (no state transition occurs)
- Not subscribing to an actor if no `onEmit` nor `xstate.emit.*` transition is defined for that actor (if a tree falls in the forest...)

Andarist: > The difference is that the snapshot is taken and the emitted value is given, so it might be more confusing to combine these two concepts.

It's still the same concept/type though - it's a standard practice to have `on*` named after a thing that is being "emitted". If somebody needs to be more precise about the timing and the context of the usage they can always say something like "emitted snapshot", while `getSnapshot` just returns "last emitted snapshot".

IMHO calling those differently makes for a harder teach. WDYT @mattpocock ?

davidkpiano: If we name this transition `onSnapshot: { ... }`, it doesn't make sense, because it's not a snapshot; it's what the actor is emitting. A snapshot is something done by the reader, not the writer.

Andarist: > it's not a snapshot; it's what the actor is emitting. A snapshot is something done by the reader, not the writer.

I wouldn't describe it this way - mainly because the reader doesn't really manage it, it just gets it from the writer that maintains it internally. Otherwise `getSnapshot` would be conceptually an async request. It sounds to me that it's exactly the same as with BehaviorSubjects - they emit values and u can always just read it synchronously using `.getValue`. That is allowed by the fact that the behavior subject itself (and not its consumer) is responsible for managing this. 

I would agree with you IF what an actor is emitting wouldn't always be available as its snapshot - but it is.

davidkpiano: With XState it's inverted - the _reader_ (which creates the `ActorRef`) maintains the snapshot, not the writer. 

Conceptually, it should be thought of as this:

```js
// pseudocode
function parent() {
  const actorRef = spawnSomeActor();
  let snapshot;

  actorRef.subscribe({
    next: (emitted) => { snapshot = emitted; }
  });

  actorRef.getSnapshot = () => snapshot;

  // ...
}
```

The `ActorRef` is owned by the reader; it's the reader's "proxy" for communicating with a possibly remote actor that it doesn't own. That's why `.getSnapshot()` is always sync.

packages/core/src/actors.ts
davidkpiano: Fixed: [82fc286](https://github.com/statelyai/xstate/pull/3148/commits/82fc2860d836c3b90af7280913d1db8819d50a98)

packages/core/src/interpreter.ts
Andarist: I agree with everything said in your comment - I just don't strongly see the connection between this and the chosen naming. 

I understand that an actor can emit just anything, and in theory, that value can be adapted on the reader's side but that just complicates the things for a very little gain. Especially since this just doesn't happen in our own implementation. 

As long as the `ActorRef`'s declaration looks smth like this:
```ts
interface ActorRef<TEvent extends { type: string }, TEmitted> {
  // ...
  subscribe: (emitted: TEmitted) => (() => void);
  getSnapshot: () => TEmitted
}
```
There is a very, very strong relationship between those things. The same type appears at both of those positions. We are often also thinking about emitting events in the frontend, so using `TEmitted` might make think somebody that this is supposed to be an event, but there is no such restriction. So I think that "snapshot" makes more sense because it's more commonly associated with some kind of data/state (and that really is what we are dealing with here).

Andarist: I feel that it's especially important doubt because there were requests to allow listening for events coming from a specific actor and I think that `onEmit` could be easily mistaken for that.

packages/core/src/actors.ts
```diff
@@ -346,73 +228,82 @@ export function createPromiseBehavior<T, TEvent extends EventObject>(
...
 }
 
-export function createObservableBehavior<
-  T extends EventObject,
-  TEvent extends EventObject
->(lazyObservable: Lazy<Subscribable<T>>): Behavior<TEvent, T | undefined> {
-  const parent = CapturedState.current?.actorRef;
+export function fromObservable<T, TEvent extends EventObject>(
+  lazyObservable: Lazy<Subscribable<T>>
+): Behavior<TEvent, T | undefined> {
```
Andarist, line 233: so... in a way, this should be:
```suggestion
): Behavior<never, T | undefined> {
```
or with the latest changes:
```suggestion
): Behavior<{ type: string }, T | undefined> {
```
With this design, an observable is really closer to a promise than a machine/callback (they can send events directly to the parent)

packages/core/src/types.ts
```diff
@@ -589,6 +593,8 @@ export interface InvokeConfig<
   onError?:
     | string
     | SingleOrArray<TransitionConfig<TContext, DoneInvokeEvent<any>>>;
+
+  onEmit?: string | SingleOrArray<TransitionConfig<TContext, EmitEvent<any>>>;
```
Andarist, line 597: I have some concerns related to how we are going to type this - but we've seen worse 😉 so hopefully we'll be able to wrangle this.

packages/core/src/actors.ts
```diff
@@ -274,48 +158,46 @@ export function createDeferredBehavior<TEvent extends EventObject>(
...
-                    origin: actorContext.self
-                  })
-                );
-
                 observers.forEach((observer) => {
                   observer.next?.(response);
-                  observer.complete?.();
+                  if (observers.has(observer)) {
+                    observer.complete?.();
+                  }
```
Andarist, line 195: Somewhat weirdly this will both emit a snapshot and complete immediately - which is not incorrect, just maybe slightly redundant. It's more an observation on my part and not a strong concern about this.

packages/core/src/actors.ts
```diff
@@ -440,7 +331,7 @@ export function createMachineBehavior<
 
       if (event.type === stopSignalType) {
         service?.stop();
-        subscription && subscription.unsubscribe(); // TODO: might not be necessary
+        subscription?.unsubscribe(); // TODO: might not be necessary
```
Andarist, line 334: Ye, it probably isn't - although there is one scenario to be considered here. What about the state that is a result of the `service.stop()` call? Should it still be broadcasted to subscribers? Do we need to? 

I don't know how it works right now, nor I don't yet have an opinion on how this should behave.

packages/core/src/actors.ts
```diff
@@ -274,48 +158,46 @@ export function createDeferredBehavior<TEvent extends EventObject>(
 
       return undefined;
     },
+    subscribe: (observer) => {
```
Andarist, line 161: "stop handler" in here (a couple of lines above) should probably clear observers

huan:
Looking forward to seeing this PR being merged! 😃

packages/core/src/actions/invoke.ts
```diff
@@ -43,8 +43,12 @@ export function invoke<
         } as InvokeActionObject;
       }
 
+      const behaviorImpl = machine.options.actors[src.type];
       const behaviorCreator: BehaviorCreator<TContext, TEvent> | undefined =
-        machine.options.actors[src.type];
+        behaviorImpl &&
+        (typeof behaviorImpl === 'function'
+          ? behaviorImpl
+          : () => behaviorImpl);
```
Andarist, line 51: nit: in general it would be better here to just conditionally resolve the `behaviorImpl` at the line 60 instead of creating a temporary, throwaway, function here when the `behaviorImpl` is not a function (and not undefined)

packages/core/src/actors.ts
```diff
@@ -281,6 +281,66 @@ export function fromObservable<T, TEvent extends EventObject>(
...
 
+/**
+ * Creates an event observable behavior that listens to an observable
+ * that delivers event objects.
+ *
+ *
+ * @param lazyObservable A function that creates an observable
+ * @returns An event observable behavior
+ */
+export function fromEventObservable<T extends EventObject>(
```
Andarist, line 292: bikesheddable alternative name: `fromObservableEvents`

I'm OK with the current name, was just thinking if this could be named differently or not

packages/core/src/actions/invoke.ts
Andarist: I've created a PR to address this: https://github.com/statelyai/xstate/pull/3281

packages/core/src/actors.ts
davidkpiano: At first, I did call it `fromObservableEvents` but it made it awkward to answer the question "What does this behavior create?"

It's more succinct to say "This behavior creates an event observable" than "This behavior creates an observable that delivers events"

packages/core/src/actors.ts
```diff
@@ -274,48 +158,46 @@ export function createDeferredBehavior<TEvent extends EventObject>(
...
-  const parent = CapturedState.current?.actorRef;
   let canceled = false;
   const observers: Set<Observer<T>> = new Set();
 
   const behavior: Behavior<TEvent, T | undefined> = {
-    transition: (_, event, actorContext) => {
+    transition: (_, event) => {
       switch (event.type) {
         case startSignalType:
           const resolvedPromise = Promise.resolve(lazyPromise());
```
Andarist, line 186: just an observation for now - it's an interesting question if we should "adapt" this to a real promise through `Promise.resolve`, some other libraries don't do this: https://codesandbox.io/s/focused-brown-kojbgv?file=/src/index.js

The implementation in the `rxjs` can be found here: https://github.com/ReactiveX/rxjs/blob/47fa8d555754b18887baf15e22eb3dd91bf8bfea/src/internal/observable/innerFrom.ts#L84-L98

packages/core/src/ObservableActorRef.ts
```diff
@@ -46,9 +51,26 @@ export class ObservableActorRef<TEvent extends EventObject, TEmitted>
...
+  public receive(event: TEvent | LifecycleSignal) {
+    this.mailbox.enqueue(event);
+  }
+  private _process(event: TEvent | LifecycleSignal) {
+    this.context._event =
+      typeof event.type !== 'string'
+        ? (event as LifecycleSignal)
+        : toSCXMLEvent(event as TEvent);
+
+    this.current = this.behavior.transition(
```
Andarist, line 68: Let's not try to fix this here but I think there is something broken in how the pieces compose together here. It's unclear which piece is responsible for actually managing the snapshot's state, which piece is responsible for managing subscribers/observers, and also that there is a shared (& mutable) state in all of our existing behavior types (which has a nasty outcome - those behaviors are not truly reusable).

For example, if we remove the `this.current =` assignment from here then **only** 4 tests fail:
```
actors › with behaviors › should work with a reducer behavior
invoke › with behaviors › should work with a behavior
invoke › with reducers › should work with a reducer
invoke › with reducers › should schedule events in a FIFO queue
```

I have some thoughts related to this - let's maybe meet together and hash it out. I can do a followup refactor after we land this PR.

packages/core/src/actors.ts
```diff
@@ -274,48 +158,46 @@ export function createDeferredBehavior<TEvent extends EventObject>(
...
+      };
+    },
     initialState: undefined
   };
 
   return behavior;
 }
 
-export function createPromiseBehavior<T, TEvent extends EventObject>(
+export function fromPromise<T, TEvent extends EventObject>(
```
Andarist, line 176: Somewhat oddly we have a minor inconsistency in the signature (and in other functions here). Those functions here (`from*`) have their type params flipped if we take `Behavior`/`ActorRef` as a point of reference:
```ts
declare function fromReducer<TState, TEvent extends EventObject>(...): ...;
declare function fromCallback<TEvent extends EventObject>(...): ...;
declare function fromPromise<T, TEvent extends EventObject>(...): ...;
declare function fromObservable<T, TEvent extends EventObject>(...): ...;
declare function fromEventObservable<T extends EventObject>(...): ...;
declare function fromMachine<TMachine extends AnyStateMachine>(...): ...;

interface ActorRef<TEvent extends EventObject, TSnapshot = any> { ... }
interface Behavior<TEvent extends EventObject, TSnapshot = any> { ... }
```

Those that take a single type param shouldn't start getting a second one, but I find it somewhat confusing that those that take 2 type params have them flipped.

`fromMachine` is completely at odds here - but it is indeed special here (I don't know how to solve problems that this solves without making it special).

packages/core/test/actions.test.ts
Andarist: 👋

packages/xstate-svelte/src/useSelector.ts
```diff
@@ -6,10 +6,10 @@ const defaultCompare = (a, b) => a === b;
 export const useSelector = <
   TActor extends ActorRef<any, any>,
   T,
-  TEmitted = TActor extends Subscribable<infer Emitted> ? Emitted : never
+  TSnapshot = TActor extends Subscribable<infer Emitted> ? Emitted : never
```
Andarist, line 9:
```suggestion
  TSnapshot = TActor extends Subscribable<infer TActorSnapshot> ? TActorSnapshot : never
```

packages/xstate-svelte/src/useSelector.ts
```diff
@@ -6,10 +6,10 @@ const defaultCompare = (a, b) => a === b;
...
 >(
   actor: TActor,
-  selector: (emitted: TEmitted) => T,
+  selector: (emitted: TSnapshot) => T,
```
Andarist, line 12:
```suggestion
  selector: (snapshot: TSnapshot) => T,
```

packages/xstate-vue/src/useSelector.ts
```diff
@@ -7,12 +7,12 @@ const defaultCompare = (a, b) => a === b;
 export function useSelector<
   TActor extends ActorRef<any, any>,
   T,
-  TEmitted = TActor extends Subscribable<infer Emitted> ? Emitted : never
+  TSnapshot = TActor extends Subscribable<infer Emitted> ? Emitted : never
```
Andarist, line 10:
```suggestion
  TSnapshot = TActor extends Subscribable<infer S> ? S : never
```

packages/xstate-vue/src/useSelector.ts
```diff
@@ -7,12 +7,12 @@ const defaultCompare = (a, b) => a === b;
...
 >(
   actor: TActor,
-  selector: (emitted: TEmitted) => T,
+  selector: (emitted: TSnapshot) => T,
```
Andarist, line 13:
```suggestion
  selector: (snapshot: TSnapshot) => T,
```

packages/core/src/actors.ts
davidkpiano: We should definitely rework this in a separate PR. I don't anticipate anyone using the generics for these `from*` functions, though.

packages/core/src/ObservableActorRef.ts
davidkpiano: @Andarist This is actually a really good catch; the reason this is happening is because two different things are trying to manage observers (the ObservableActorRef and the behaviors). This should only be done in the ObservableActorRef, so the behaviors become even smaller in scope (good thing).

I have a fix for this in progress.

packages/core/src/actors.ts
```diff
@@ -128,12 +115,14 @@ export function fromCallback<TEvent extends EventObject>(
         if (isPromiseLike(dispose)) {
           dispose.then(
             (resolved) => {
-              sendNext(resolved);
-              sendComplete();
+              observers.forEach((o) => o.next?.(resolved));
```
Andarist, line 118: It IMHO would make sense to make `ObservableActorRef` responsible for emitting to observers. Ideally, the behavior would be responsible for just managing its own state and letting its direct owner (the `ObservableActorRef`) that something has changed. After all - one subscribes to an observable actor ref and not directly to behavior, so why behavior would be responsible for taking care of those multicast notifications?

Note that at the moment each behavior type has to implement this "broadcast" on its own, so there is a chance that a custom behavior could do it wrong etc, whereas in `ObservableActorRef` we can introduce a strong guarantee around this. 

With the proposed approach it will also be easier to fix the timing issue related to when observers are notified versus when the snapshot is being read by those subscribers.

At the moment we have a bug like this:
```
current = behavior.transition()
          -> observers.forEach(cb => cb(next))
            -> ref.getSnapshot() // oops, the `current` in the ObservableActorRef has not yet been updated
```

Whereas with this it wouldn't be an issue:
```
current = behavior.transition()
observers.forEach(cb => cb(next))
   -> ref.getSnapshot() // it's just fine as current has already been updated
```

davidkpiano: We have to think about what is "emittable". This is part of a bigger topic -  how do we keep state that should be private truly private if desired (not discoverable by anything)?

Andarist: In the spirit of keeping the discussion in public - we've quickly agreed on a call that we might think about "private" data for actors in the future. But right now even machines don't have such, so it's not worth worrying about.

I imagine that somebody could store such a private state in behaviors in a couple of ways - but there wouldn't be a clear strategy for persisting/rehydrating (and introspection) with any of those.

davidkpiano:
@Andarist I reworked the actors to do the original thing - send events directly to the parent. It turns out that this was a better approach that avoided the complexity of dealing with observers and implicit observer method calls.

This is also inspired by a relevant approach in Akka with dealing with futures (promises): https://blog.rockthejvm.com/pipe-pattern/

I think you'll appreciate this change 👍

packages/core/src/actors.ts
```diff
@@ -0,0 +1,516 @@
...
+            error: (err) => {
+              self.send({ type: errorEventType, data: err });
+            },
+            complete: () => {
+              self.send({ type: completeEventType });
+            }
+          });
+          return state;
+        case nextEventType:
+          return event.data;
```
Andarist, line 333: This is an unused code. It should either be removed or this event should be sent from here:
https://github.com/statelyai/xstate/blob/d3274d1f346c62e6702ecc58b5ff36181a1fbd0b/packages/core/src/actors.ts#L322
and the `self._parent.send(...)` should be performed here

packages/core/src/actors.ts
```diff
@@ -0,0 +1,516 @@
...
+        case rejectEventType:
+          const errorEvent = error(name, event.data);
+
+          self._parent?.send(
+            toSCXMLEvent(errorEvent, {
+              origin: self
+            })
+          );
+          return event.data;
```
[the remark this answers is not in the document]
davidkpiano: Curious on your thoughts here: since we want to represent an accurate promise snapshot, what do you think about representing promise snapshots as an "Either type" like other languages have?

packages/core/src/actors.ts
```diff
@@ -0,0 +1,516 @@
...
+              self._parent?.send(toSCXMLEvent(value, { origin: self }));
+            },
+            error: (err) => {
+              self.send({ type: errorEventType, data: err });
+            },
+            complete: () => {
+              self.send({ type: completeEventType });
+            }
+          });
+          return state;
```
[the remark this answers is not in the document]
davidkpiano: Should I rename `state` to `snapshot` to make this more clear?

packages/core/src/actors.ts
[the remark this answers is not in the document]
Andarist: So based on T we'd type the actual snapshot as `{ ok: true, value: T } | { ok: false, error: unknown }` (or smth like that)?

packages/core/src/interpreter.ts
```diff
@@ -589,21 +577,12 @@ export class Interpreter<
       isSCXMLErrorEvent(_event) &&
       !this.state.nextEvents.some((nextEvent) => nextEvent === _event.name)
     ) {
+      // Error event unhandled by machine
       this.handleErrorEvent(_event);
+      this.stop();
```
Andarist, line 582: @davidkpiano this is a somewhat surprising change in this PR, was there any specific reason for it to be included here? 🤔

huan:
Congratulations!

Could we expect the v5 to be published soon to the `xstate@next` NPM? :-D

Andarist:
It's already available as `@alpha`

huan:
Awesome, thank you very much for publishing it!

PR #3153, opened 2022-03-14
Made type displays more readable

Andarist:
**Before**
Screenshot 2022-03-14 at 17 52 43

**After**
Screenshot 2022-03-14 at 17 52 00

Arguably here the difference isn't that big because we still have a lot of types displayed here... but you might notice that `ResolveTypegenMeta` is no longer inlined in the display. This should also help with `.d.ts` generation as this type alias was often inlined in the output, see the issue [here](https://github.com/statelyai/xstate-tools/issues/91), cc @mdpratt

I'm facing a weird TS issue though that blocks this PR a little bit - I think this is fixable but I need to figure out a best solution for this. The problem is now with nested `MarkAllImplementationsAsProvided<MarkAllImplementationsAsProvided<ResolvedTypesMeta<ResolveTypegenMeta<...>>>>` "calls". Previously those were assignable to variants with less/more nesting like this (when I've been using type aliases) but now they are not. I'm yet unsure if this is by design or if this is a yet another weird TS edge case. The issue is depicted [here](https://www.typescriptlang.org/play?ssl=26&ssc=6&pln=26&pc=35#code/JYOwLgpgTgZghgYwgAgCoE8AOEDmEQCiIcARgDYQAmyA3gFDLIDkAAiwB4DOYckA9GCy58TAFzIwUAK4QA3HQC+dOqEixEKAIJkyASQC2mCvvw8wwAPYhOABSgWAbsEpVaDZPuCdOoHAaMQJuC8ltbi9IyMiOZWnOIgEA7Q8pHILmRw6HHICUlQKZE4UnBQlNm5ye6MnNBOSOWJlYwK8kp0gtjIdhaYADyoADTIANIAfMgAvCPIEOyQIGXIANYQ6BYwaMgA-GgA2sMAusjxjfnKqtDwSMgAsiVL2nqGxqYhsZq29k4ulLrglxp+gAlCCcCxkJKUDDYTg3CA8cYRZhsLhmCACIR4EBiLr2PqoEFgiFUaGguE8IasDjcXjojrCbGjApQUHgyHibr4wlsklCWHwuCUllEyFMcYAMmQj38L2CMWs3W+VFaynpUr+aiuKCmdygDx0MsCr3lnA+iucVA1AKQvV1+qeASCZlCps+jgtv3+6htpKxRFIFEoo1G7SEyAAQlbvdrbvdpc8jXKXWavh6o1r+pj8P7yFQQ6HOnZVMNVpxhqBKHB+jM5vhFtwoL5xlMaPtkKA0AdxABGNpqzQsgDCFkMIVz6Y0ADUSsA4OBJuqvVqa-NFpGlxptshu8dkAAGOh8PiRAB6WwLKBLWU0C6L4CvZYrVZWaw2mgnSBDR9P57VD-Dt6Nvepblgsz6rOsEYfhAX7HowZ4XsguicL6+BXsOo7mLmt7AAgSy9O+G5IJSKI0vw9JYmKK51pwXS4fh66ahoJHUmiGLYJR4w7Du4gHt+8G-mGyHcsSlDoSOmBjhQOF4QR0FCqyolUbMq60TY9G9Ix1oQApIpUFR3G7nxcHIAhiF2vGjrGsmbpKlCQjaLOnDAopkKkvyCILgSrm8jC5JwMgkqWbKzqxOaPyqmGmiko5cC0TqcYGgmTpvNYKbuj8MVkE5tqJQ6IWpa64W+RAsXOahhDELmQYhmq4ZZU5C4WUlVlJu8tkeg1cWZhx2ZVYG+b9kOElSRAXWcNOjZzmAC7RQ52VxdRCy0fV82NYZvGHiZZ5AA)

### DIFF
--- a/packages/core/src/typegenTypes.ts
+++ b/packages/core/src/typegenTypes.ts
@@ -118,16 +118,18 @@ export type AreAllImplementationsAssumedToBeProvided<
     : false
   : true;
 
+interface AllImplementationsProvided {
+  missingImplementations: {
+    actions: never;
+    delays: never;
+    guards: never;
+    services: never;
+  };
+}
+
 export interface MarkAllImplementationsAsProvided<TResolvedTypesMeta> {
   '@@xstate/typegen': Prop<TResolvedTypesMeta, '@@xstate/typegen'>;
-  resolved: Prop<TResolvedTypesMeta, 'resolved'> & {
-    missingImplementations: {
-      actions: never;
-      delays: never;
-      guards: never;
-      services: never;
-    };
-  };
+  resolved: Prop<TResolvedTypesMeta, 'resolved'> & AllImplementationsProvided;
 }
 
 type GenerateServiceEvent<
@@ -191,7 +193,8 @@ export interface ResolveTypegenMeta<
           Prop<TTypesMeta, 'internalEvents'>
         >;
       }
-    : MarkAllImplementationsAsProvided<TypegenDisabled> &
+    : TypegenDisabled &
+        AllImplementationsProvided &
         AllowAllEvents & {
           indexedActions: IndexByType<TAction>;
           indexedEvents: Record<string, TEvent> & {
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1708,7 +1708,9 @@ export type ActorRefFrom<T> = ReturnTypeOrValue<T> extends infer R
         TContext,
         TEvent,
         TTypestate,
-        MarkAllImplementationsAsProvided<TResolvedTypesMeta>
+        AreAllImplementationsAssumedToBeProvided<TResolvedTypesMeta> extends false
+          ? MarkAllImplementationsAsProvided<TResolvedTypesMeta>
+          : TResolvedTypesMeta
       >
     : R extends Promise<infer U>
     ? ActorRef<never, U>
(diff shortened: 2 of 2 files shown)


Andarist:
Ok, I've resolved the weird assignability problem with interfaces by wrapping in `MarkAllImplementationsAsProvided` conditionally 🤷‍♂️

packages/core/src/typegenTypes.ts
```diff
@@ -164,33 +172,37 @@ type AllowAllEvents = {
...
+          Prop<TTypesMeta, 'internalEvents'>
         >;
-      };
+      }
+    : TypegenDisabled &
+        AllImplementationsProvided &
+        AllowAllEvents & {
+          indexedActions: IndexByType<TAction>;
+          indexedEvents: Record<string, TEvent> & {
+            __XSTATE_ALLOW_ANY_INVOKE_DATA_HACK__: { data: any };
```
davidkpiano, line 201: Will this show up in intellisense ever?

Andarist: I dont think so - but if we catch this in any display then we can revisit this. Note that this PR didnt introduce this thing. 

My hope is to introduce tests for type displays, autocomplete etc. With such tests we could control this stuff with much better confidence

davidkpiano (approved):

PR #3154, opened 2022-03-14
implements server package

chlbri:
I add a package to handles simple function like state machine.
It's like a state machine without sending events.
The process of action and transitions is like eventless states.


chlbri:
Also, I added as small function to test a machine

chlbri (commented):
Reviewed

davidkpiano:
This was an accidental merge; I reverted the 3 commits.

Andarist:
@chlbri could you perhaps open a discussion thread about your ideas from this PR? We would like to discuss this further but it's not obvious what exact problems are you solving here and how.

PR #3157, opened 2022-03-16
@chlbri/feature-locale-fr

chlbri:
Starting


laurakalbag:
@chlbri Merci beaucoup… thank you so much for this work. 

Do you consider this translation ready to add to the documentation? 

Not all pages need to be translated for us to put the translation live, but we do need [navigation and sidebar translations added to the `config.js`  file in `/docs/.vuepress/`](https://vuepress.vuejs.org/guide/i18n.html#internationalization) so that folks can navigate to the translated pages.

Is this something you would like to add, or should we find somebody else to help with this work? Thank again!

chlbri:
Yes, I think the translations I made, it is ready for prod.

chlbri:
> @chlbri Merci beaucoup… thank you so much for this work.
> Welcome
> Do you consider this translation ready to add to the documentation?
> Yes
> Not all pages need to be translated for us to put the translation live, but we do need [navigation and sidebar translations added to the `config.js` file in `/docs/.vuepress/`](https://vuepress.vuejs.org/guide/i18n.html#internationalization) so that folks can navigate to the translated pages.
> I read the docs on vuepress. I never do any vuepress project, but I think I will use it. Thanks for the link.
> Is this something you would like to add, or should we find somebody else to help with this work? Thanks again!
 So the only thing I need is to add the path of the translation ?

laurakalbag:
> I read the docs on vuepress. I never do any vuepress project, but I think I will use it. Thanks for the link.
> So the only thing I need is to add the path of the translation ?

VuePress generates the top and sidebar navigation from the `config.js` file. If you add the path of the translation under `locales` it will create the Languages menu option in the top navigation as below:

Languages menu with options for English and Simplified Chinese.

If you add the sidebar navigation under `themeConfig: … locales:` then it will create the sidebar navigation for the translated pages. As below:

The XState docs sidebar navigation listing the docs pages in English.

The XState docs sidebar navigation listing the docs pages in Simplified Chinese.

chlbri:
Thanks, I will update config.js file and finish the translation

chlbri:
I updated the config.js file. You can check if everything ok.

PR #3167, opened 2022-03-22
Add basic Roadmap page

laurakalbag:
A basic Roadmap reflecting our internal Roadmap items. Later on we can make it fancy, but this is the basic content.

Some of them are reworded slightly (based on their issues) for clarity.

Please review considering:
- Is this everything we want to include?
- Should anything here not be included?
- Does grouping by “In Progress”, “Upcoming Q2 2022” and “Later” make sense?

### DIFF
--- a/docs/roadmap/README.md
+++ b/docs/roadmap/README.md
@@ -1,0 +1,89 @@
+---
+reviewed: 2022-03-22
+styleGuideVersion: 0
+---
+
+# Roadmap
+
+A roadmap for what Stately is working on for XState and Stately tools.
+
+## In Progress
+
+What we’re working on right now.
+
+### Editor
+
+- Bug fixes
+- Discover page
+- Inline implementations
+- Entity snapping
+
+## Upcoming Q2 2022
+
+Features planned for April - June 2022.
+
+### XState
+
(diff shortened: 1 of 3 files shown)


docs/roadmap/README.md
```diff
@@ -0,0 +1,89 @@
...
+- Bug fixes
+- Discover page
+- Inline implementations
+- Entity snapping
+
+## Upcoming Q2 2022
+
+Features planned for April - June 2022.
+
+### XState
```
davidkpiano, line 25: Should we make this more general? cc. @Andarist 

```suggestion
### Open Source
```

davidkpiano (approved):
One small suggestion (nit-level), LGTM!

docs/roadmap/README.md
laurakalbag: > Should we make this more general? cc. @Andarist

I was just thinking of using terms that the audience would be familiar with. But if it’s not all XState, then Open Source makes sense.

davidkpiano: I can go either way - actually, perhaps keeping it as XState makes sense 👍

Issue #3170, opened 2022-03-23
Inclusion of history states in a parallel state breaks the onDone transition


jens-ids:
**Description**

If a history state is included in a parallel state with an `onDone` transition, the `onDone` transition is not taken when the last final state in the parallel state is entered.

**Expected Result**

The `onDone` transition is taken

**Actual Result**

The `onDone` transition is not taken

**Reproduction**

```
{
  "id": "test",
  "initial": "pending",
  "states": {
   "pending": {
    "on": {
     "trigger": {
      "target": "parallelSteps"
     }
    }
   },
   "parallelSteps": {
     "type": "parallel",
     states: {
        hist: {
          type: 'history',
          history: 'deep' // optional; default is 'shallow'
        },
        one: {
          initial: 'waitingForConfirmation',
          states: {
            "waitingForConfirmation": {
              on: {
                "confirmation": {
                  "target": 'done'
                }
              }
            },
            "done": {
              type: 'final'
            }
          }
        },
        two: {
          initial: 'transferPending',
          states: {
            'transferPending': {
              on: {
                'transferComplete': {
                  target: 'done'
                }
              }
            },
            'done': {
              type: 'final',
            }
          }
        }
     },
     onDone: 'generationInitiated',
   },
   "generationInitiated": {
   }
  }
 }
```

**Additional context**

jens-ids:
@Andarist thanks for the extremely fast response and fix! Also, thanks to the entire team for great software. I'll make sure to provide runnable repro case if I find the need to report any issues in the future.

PR #3171, opened 2022-03-24
Fixed an issue with `onDone` on parallel states with a history state not being called correctly

Andarist:
fixes https://github.com/statelyai/xstate/issues/3170

### DIFF
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -18,7 +18,9 @@ export const isLeafNode = (
 export function getChildren<TC, TE extends EventObject>(
   stateNode: StateNode<TC, any, TE>
 ): Array<StateNode<TC, any, TE>> {
-  return Object.keys(stateNode.states).map((key) => stateNode.states[key]);
+  return Object.keys(stateNode.states)
+    .map((key) => stateNode.states[key])
+    .filter((sn) => sn.type !== 'history');
 }
 
 export function getAllStateNodes<TC, TE extends EventObject>(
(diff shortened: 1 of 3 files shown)


packages/core/src/stateUtils.ts
```diff
@@ -18,7 +18,9 @@ export const isLeafNode = (
 export function getChildren<TC, TE extends EventObject>(
   stateNode: StateNode<TC, any, TE>
 ): Array<StateNode<TC, any, TE>> {
-  return Object.keys(stateNode.states).map((key) => stateNode.states[key]);
+  return Object.keys(stateNode.states)
+    .map((key) => stateNode.states[key])
+    .filter((sn) => sn.type !== 'history');
```
Andarist, line 23: It's a pseudostate and thus it's never really in the configuration, so it should always be just ignored, and not treated as one. This matches now the [`getChildStates` procedure](https://www.w3.org/TR/scxml/#getChildStates) from SCXML

packages/core/src/stateUtils.ts
```diff
@@ -68,10 +70,6 @@ export function getConfiguration<TC, TE extends EventObject>(
     } else {
       if (s.type === 'parallel') {
         for (const child of getChildren(s)) {
-          if (child.type === 'history') {
```
Andarist, line 71: It's now already filtered out by `getChildren` so this could be safely removed from here.

packages/core/src/stateUtils.ts
```diff
@@ -18,7 +18,9 @@ export const isLeafNode = (
 export function getChildren<TC, TE extends EventObject>(
```
Andarist, line 18: The biggest change caused by this that I've noticed is that now history states won't have `.order` set on them. That should be fine though because that's only used for sorting actions and history states can't have actions anyway

davidkpiano (approved):

Issue #3169, opened 2022-03-23
Bug: useActor on spawned actor does not sync with component  [@xstate/vue]


jamus:
### Description

Using `useActor` in a child component does not sync with the changing state of a spawned actor as I would have expected from the doc description.

### Expected result

`useActor` subscribes to emitted changes from an existing actor as suggested from the docs - https://xstate.js.org/docs/packages/xstate-vue/#api

### Actual result

Instead it only shows the intial state of the actor.

However, if there is a change to the component that triggers a re-render (which Im assuming triggers the reactivity) syncing begins working as expected.

### Reproduction

https://codesandbox.io/s/xstate-vue-3-useactor-issue-forked-37bfl7

### Additional context

I'm new to Vue3 and spawning actors and also struggled to find a common convention / example, so it could be I'm missing something with my implimentation or expectation of `useActor`.

**A work-around seems to be** to use a `key` on the containing element that is the changing state - to activate a re-render. E.g. `:key="state.context.todos[0].ref"`

Andarist:
As described [here](https://vuejs.org/api/composition-api-setup.html#accessing-props) you need to wrap the argument with `toRef` (or similar). Take a look at [the fixed codesandbox](https://codesandbox.io/s/xstate-vue-3-useactor-issue-forked-mrl6k4?file=/src/ChildComponent.vue:570-637) where I use this:
```js
const { state, send } = useActor(toRef(props, "someSpawnedActor"));
```

This is somewhat annoying though, so I wonder if the API of `useActor` should look differently to accommodate for this use case better. I'm not sure what are the established conventions in the Vue world though. 

cc @ChrisShank @santicros

jamus:
Thank you @Andarist ! I Think that makes sense to rely on Vue to watch for changes I guess I was just mislead a little by the wording "subscribes to emitted changes from an existing". Perhaps "provides access to" woudl be better. With an added "Use in combination with `toRef` if subscription to changes from the actor are required"

Andarist:
Could you propose a documentation change around this in a form of PR?

santicros:
Hey guys,
You did a great debugging here!

Your solution is great @Andarist, tough I would recommend to better use `computed`, because `toRef` makes it a regular `ref`, which can then be mutated,... So [like this](https://codesandbox.io/s/xstate-vue-3-useactor-issue-computed-c30ykz):
```js
const { state, send } = useActor(computed(() => props.someSpawnedActor));
```

[...]

PR #3173, opened 2022-03-24
Xstate vue  update use actor documentation

jamus:
resolves https://github.com/statelyai/xstate/issues/3169

Updates the description and example code for `useActor` in [@xstate/vue](https://xstate.js.org/docs/packages/xstate-vue/)

### DIFF
--- a/docs/packages/xstate-vue/index.md
+++ b/docs/packages/xstate-vue/index.md
@@ -93,7 +93,10 @@ A [Vue composition function](https://v3.vuejs.org/guide/composition-api-introduc
 
 ### `useActor(actor, getSnapshot)`
 
-A [Vue composition function](https://v3.vuejs.org/guide/composition-api-introduction.html) that provides access to an existing [actor](https://xstate.js.org/docs/guides/actors.html). Can be in combination with Vue's [computed](https://vuex.vuejs.org/guide/composition-api.html#accessing-state-and-getters) to subscribe to changes from the actor. 
+A [Vue composition function](https://vuejs.org/guide/extras/composition-api-faq.html) that provides access to an existing [actor](https://xstate.js.org/docs/guides/actors.html).
+
+To subscribe to changes on the an actor whilst retaining reactivity from props or another reactive variable, Vue's [computed](https://vuejs.org/api/reactivity-core.html#computed) can be used.
+
 
 _Since 0.5.0_
 
--- a/docs/packages/xstate-vue/index.md
+++ b/docs/packages/xstate-vue/index.md
@@ -93,7 +93,7 @@ A [Vue composition function](https://v3.vuejs.org/guide/composition-api-introduc
 
 ### `useActor(actor, getSnapshot)`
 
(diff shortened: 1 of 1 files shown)


docs/packages/xstate-vue/index.md
```diff
@@ -93,7 +93,7 @@ A [Vue composition function](https://v3.vuejs.org/guide/composition-api-introduc
 
 ### `useActor(actor, getSnapshot)`
 
-A [Vue composition function](https://v3.vuejs.org/guide/composition-api-introduction.html) that subscribes to emitted changes from an existing [actor](https://xstate.js.org/docs/guides/actors.html).
+A [Vue composition function](https://v3.vuejs.org/guide/composition-api-introduction.html) that provides access to an existing [actor](https://xstate.js.org/docs/guides/actors.html). Can be in combination with Vue's [computed](https://vuex.vuejs.org/guide/composition-api.html#accessing-state-and-getters) to subscribe to changes from the actor.
```
Andarist, line 96: Please note that this is just a suggestion, feel free to adjust/reword if needed.
```suggestion
A [Vue composition function](https://vuejs.org/guide/extras/composition-api-faq.html) that provides access to an existing [actor](https://xstate.js.org/docs/guides/actors.html). If you need to listen for different actor instances throughout the lifecycle of this hook then you might need to retain reactivity when destructuring from reactive props or another reactive variable. This can be done using Vue's [computed](https://vuejs.org/api/reactivity-core.html#computed). 
```

1. I would like to put emphasis here that you only need to wrap this if you need to listen for the actor reference being changed - and not if you only want to subscribe to a **single** actor.
2. I would feel better if we'd link to Vue docs here instead of linking to Vuex
3. The note about `computed` could be moved to the bottom of this section, as it's additional info and not necessarily something that we need to start with

jamus: Good catch on the link. Also you are right about making computed an option rather than a requirement.

docs/packages/xstate-vue/index.md
```diff
@@ -104,12 +107,13 @@ _Since 0.5.0_
...
 
 ```js
+import { computed } from "vue";
 import { useActor } from '@xstate/vue';
 
 export default {
   props: ['someSpawnedActor'],
   setup(props) {
-    const { state, send } = useActor(props.someSpawnedActor);
+    const { state, send } = useActor(computed(() => props.someSpawnedActor));
```
Andarist, line 116: Since this is the main example here - this might still look like a requirement. I would just push the note about it + a single-line example to the end of this section (so basically below this example here).

jamus: ahhh I see. 👍

docs/packages/xstate-vue/index.md
```diff
@@ -114,6 +114,11 @@ export default {
   }
 };
 ```
+To subscribe to changes on the an actor whilst retaining reactivity from props or another reactive variable, Vue's [computed](https://vuejs.org/api/reactivity-core.html#computed) can be used.
+
+```js
+    const { state, send } = useActor(computed(() => props.someSpawnedActor));
```
Andarist, line 120:
```suggestion
const { state, send } = useActor(computed(() => props.someSpawnedActor));
```

Andarist:
Thanks!

PR #3178, opened 2022-03-26
[core] Add warning and stop forwarding when forwarding to undefined

davidkpiano:


packages/core/test/actions.test.ts
```diff
@@ -1246,6 +1246,19 @@ describe('forwardTo()', () => {
 
     service.send('EVENT', { value: 42 });
   });
+
+  it('should not cause an infinite loop when forwarding to undefined', () => {
+    const machine = createMachine({
+      on: {
+        '*': { cond: () => true, actions: forwardTo(undefined as any) }
```
davidkpiano, line 1253: This is what was causing the infinite loop...

packages/core/src/interpreter.ts
```diff
@@ -872,6 +872,13 @@ export class Interpreter<
...
             this.sendTo(sendAction._event, sendAction.to);
+          } else if (sendAction.sendType === 'forward') {
+            // Expected to forward to an actor
+            warn(
+              false,
+              `Attempted to forward event "${sendAction.event.type}" to undefined actor`
+            );
+            break;
           } else {
             this.send(
```
davidkpiano, line 876: And the infinite loop was being caused due to this line, which sends the event to self if `sendAction.to` is `undefined`.

packages/core/src/interpreter.ts
```diff
@@ -872,6 +872,13 @@ export class Interpreter<
         } else {
           if (sendAction.to) {
             this.sendTo(sendAction._event, sendAction.to);
+          } else if (sendAction.sendType === 'forward') {
```
davidkpiano, line 875: However, when we're _forwarding_ an event, it should never be forwarded to self! So we need a way of identifying that the send action is a "forwarding" action, without breaking any behavior (hence `sendAction.sendType === 'forward'`)

packages/core/src/actions.ts
```diff
@@ -563,7 +564,8 @@ export function forwardTo<TContext, TEvent extends EventObject>(
 ): SendAction<TContext, TEvent, AnyEventObject> {
   return send<TContext, TEvent>((_, event) => event, {
     ...options,
-    to: target
+    to: target,
+    sendType: 'forward'
```
davidkpiano, line 568: To do this, we add this little piece of `sendType: 'forward'` metadata to the action object created by `forwardTo(...)`

packages/core/src/actions.ts
```diff
@@ -205,7 +205,8 @@ export function send<
         ? options.id
         : isFunction(event)
         ? event.name
-        : (getEventType<TSentEvent>(event) as string)
+        : (getEventType<TSentEvent>(event) as string),
+    sendType: options?.sendType
```
davidkpiano, line 209: Make sure we're adding this on the action object...

packages/core/test/actions.test.ts
```diff
@@ -1246,6 +1246,19 @@ describe('forwardTo()', () => {
...
+      }
+    });
+
+    const service = interpret(machine).start();
+
+    // This is testing for an infinite loop
+    service.send('TEST');
```
davidkpiano, line 1260: And all should be well! Without adding that metadata, this would cause an infinite loop and this test would timeout (or worse, just hang).

This test completing is the success criterion.

davidkpiano:
Click "Tour" to see this code tour in CodeSee

CleanShot 2022-03-26 at 10 07 00@2x

PR #3180, opened 2022-03-28
Simplify the algorithm for computing reentry nodes during external transitions

Andarist:


packages/core/src/StateNode.ts
```diff
@@ -958,11 +958,7 @@ class StateNode<
 
     if (!isInternal) {
       allNextStateNodes.forEach((n1) => {
-        this.nodesFromChild(n1).forEach((n2) => {
-          if (n2 !== this.machine) {
-            reentryNodes.push(n2);
-          }
-        });
```
Andarist, line 965: Instead of "patching" the result of `nodesFromChild` I've reexamined the `nodesFromChild`. It was only used here (same for `escapes` that was used in it) so I took the liberty to just remove it (it was a private method) and reintroduce a different implementation, better suited for this particular task.

packages/core/src/StateNode.ts
```diff
@@ -976,13 +972,9 @@ class StateNode<
     };
   }
 
-  private nodesFromChild(
+  private getExternalReentryNodes(
     childStateNode: StateNode<TContext, any, TEvent, any, any, any>
   ): Array<StateNode<TContext, any, TEvent, any, any, any>> {
-    if (childStateNode.escapes(this)) {
-      return [];
-    }
```
Andarist, line 984: It was only used here, I've replaced this check with the conditional return [here](https://github.com/statelyai/xstate/pull/3180/files#diff-7f553f4d4c285f1c05cced2e979c5b957d8884e39e74cc6797b32c63a0be7e7aR987-R990). Thanks to that we only need 1 tree traversal instead of 2 to compute the result of `getExternalReentryNodes`

davidkpiano (approved):

Issue #3177, opened 2022-03-25
Footgun: `forwardTo` causes infinite loop when you pass `undefined`


Silverwolf90:
### Description

So I ran into this issue earlier this week, we had a bug where are servers were totally locking up and ultimately the culprit was code like the following:

```
forwardTo((context, event) => context.refs.get(event.id))
```

When the `event.id` was incorrect, then `context.refs.get(event.id)` returned `undefined` which then forwards the event to the machine's self, which forwards the event to the machine's self, which forwards the event to the machine's self, etc  😅

This feels like quite the footgun -- I'm surprised I haven't run into this issue sooner! We fixed it by passing through actor ids instead of an entire actor reference, so now we're properly getting the "Cannot send to child ...." error when the actor doesn't exist.

### Expected result

Not really sure. Maybe `forwardTo` should throw an error when it recieves `undefined`?  Is there a good usecase for it?

### Actual result

Infinite loop. Locks up the process.

### Reproduction

None

Andarist:
Could you prepare a repro case for this? I understand that the issue should be trivial to reproduce but a proper repro case always help the maintainers to get straight to the problem.

PR #3181, opened 2022-03-28
Alternative solution for #3177 that adds no runtime overhead

Andarist:
This is an alternative approach to the one taken in https://github.com/statelyai/xstate/pull/3178. This solution only validates that during the development and requires no extra code for the production (as the development-only branch gets stripped down).

This solution here tries to balance the improved DX with the bundlesize concerns

### DIFF
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -205,8 +205,7 @@ export function send<
         ? options.id
         : isFunction(event)
         ? event.name
-        : (getEventType<TSentEvent>(event) as string),
-    sendType: options?.sendType
+        : (getEventType<TSentEvent>(event) as string)
   };
 }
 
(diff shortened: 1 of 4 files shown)


davidkpiano:
One slight issue with this is that it's possible (but unwise) for this to never be caught in development, but somehow happen in production, causing that same infinite loop.

IMO we should prevent this from happening in production itself.

davidkpiano (approved):

Andarist:
> One slight issue with this is that it's possible (but unwise) for this to never be caught in development, but somehow happen in production, causing that same infinite loop.

I agree - we might think about it but we should tackle this from a more generic perspective. There are a lot of situations in which one could end up with an infinite loop between actors, we probably should introduce some threshold of the "sync event chain" or something and throw an error when it's exceeded. It would be somewhat similar to what React does with rerenders or what TypeScript does with type instantiations.

PR #3184, opened 2022-03-29
Update typescript.md

ScreamZ:
Update documentation to gives another technique of fixing errors with code compiler

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -569,6 +569,13 @@ entry: [
 ],
 ```
 
+ Or you can specify generics type by hand too, this allows to keep empty params args.
+```ts
+ assign<GameEngineContext, GameEngineEvents>({
+        color: () => _.sample(["bg-cyan-500", "bg-green-400", "bg-red-400"]),
+      })
+```
+
 This is a nasty bug to fix and involves moving our codebase to strict mode, but we're planning to do it in V5.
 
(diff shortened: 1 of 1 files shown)


docs/guides/typescript.md
```diff
@@ -569,6 +569,13 @@ entry: [
 ],
 ```
 
+ Or you can specify generics type by hand too, this allows to keep empty params args.
+```ts
+ assign<GameEngineContext, GameEngineEvents>({
+        color: () => _.sample(["bg-cyan-500", "bg-green-400", "bg-red-400"]),
+      })
```
Andarist, line 576: I'm not entirely sure that we want to "promote" this - inference is quite powerful and by providing explicit generic params you turn it off here. OTOH, I don't have an idea now how we could enhance the experience with inference here... so perhaps this ain't that bad.

However, if we are supposed to merge this - this should use the same "types" as the example above.

I would add some `MyEvent` type where the `interface Context` is defined and use `<Context, MyEvent>` here.

ScreamZ: You're in charge here, tell me what you prefer.

I guess it can be given as an alternative to the not used parameter. But also NodeJS and various engines optimise code depending on parameters defined or not.

davidkpiano:
Thank you for this contribution. All docs-related work is now in https://github.com/statelyai/docs

PR #3185, opened 2022-03-29
1276 add timestamp to delayed actions

gavination:
This change adds a timestamp to the SendAction object, allowing for better understanding and control of delayed Actions

### DIFF
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -251,7 +251,8 @@ export function resolveSend<
     to: resolvedTarget,
     _event: resolvedEvent,
     event: resolvedEvent.data,
-    delay: resolvedDelay
+    delay: resolvedDelay,
+    timestamp: Date.now()
   };
 }
 
--- a/.changeset/ninety-olives-arrive.md
+++ b/.changeset/ninety-olives-arrive.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+This change adds a timestamp to the SendAction object, allowing for better understanding and control of delayed Actions
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1268,6 +1268,7 @@ export interface SendActionObject<
   event: TSentEvent;
   delay?: number;
   id: string | number;
+  timestamp?: number;
 }
 
 export interface StopAction<TContext, TEvent extends EventObject>
--- a/packages/core/test/after.test.ts
+++ b/packages/core/test/after.test.ts
@@ -35,10 +35,10 @@ describe('delayed transitions', () => {
     expect(nextState.value).toEqual('yellow');
     expect(nextState.actions).toEqual([
       cancel(after(1000, 'light.green')),
-      {
+      expect.objectContaining({
         ...send(after(1000, 'light.yellow'), { delay: 1000 }),
         _event: toSCXMLEvent(after(1000, 'light.yellow'))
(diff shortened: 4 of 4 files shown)


packages/core/src/actions.ts
```diff
@@ -251,7 +251,8 @@ export function resolveSend<
     to: resolvedTarget,
     _event: resolvedEvent,
     event: resolvedEvent.data,
-    delay: resolvedDelay
+    delay: resolvedDelay,
+    timestamp: Date.now()
```
Andarist, line 255: I don't think this is a good change. It doesn't fit here conceptually for 2 reasons:
- it makes this function impure and as far as I can tell all "action resolvers" are currently pure(-ish)
- it's not clear what this actually represents. The problem is that, conceptually, `machine.transition` is pure and doesn't have to lead to the execution of the actions. This is up to the caller. So if we look at this from that point of view then this information might be incorrect because there is no constraint put on the system that this action has to be executed in "the same frame". 

I understand how this kind of information might be useful at times - but as mentioned, it's not up to the machine to manage this information. If anything this should be the responsibility of the interpreter (so basically, the caller of the `machine.transition`).

Even if we'd put a constraint on the system that this has to be consumed "asap" so this information would stay somewhat correct, it's not obvious to me **how** do you plan to utilize this information using this specific property **here**.

It seems to me, that if we want to rehydrate the state and have those timers acting as "consumed" from their respective starting times then this should be part of the state. Otherwise, this isn't really ergonomic for public consumption.

Imagine this situation:
1. machine enters a state and schedules a delayed send action
2. we continue processing, execute n arbitrary amount of transitions etc
3. now the "current state" doesn't hold this delayed send action, you can't easily record it (unless you track it on your own), nor you can persist the state of "active" timers
4. you've lost this delayed send action and you won't be able to rehydrate it

There is also an additional concern here - how do we "seed" the rehydrated interpreter with "active" timers?

Andarist, line 255: I don't think this is a good change. It doesn't fit here conceptually for 2 reasons:
- it makes this function impure and as far as I can tell all "action resolvers" are currently pure(-ish)
- it's not clear what this actually represents. The problem is that, conceptually, `machine.transition` is pure and doesn't have to lead to the execution of the actions. This is up to the caller. So if we look at this from that point of view then this information might be incorrect because there is no constraint put on the system that this action has to be executed in "the same frame". 

I understand how this kind of information might be useful at times - but as mentioned, it's not up to the machine to manage this information. If anything this should be the responsibility of the interpreter (so basically, the caller of the `machine.transition`).

Even if we'd put a constraint on the system that this has to be consumed "asap" so this information would stay somewhat correct, it's not obvious to me **how** do you plan to utilize this information using this specific property **here**.

It seems to me, that if we want to rehydrate the state and have those timers acting as "consumed" from their respective starting times then this should be part of the state. Otherwise, this isn't really ergonomic for public consumption.

Imagine this situation:
1. machine enters a state and schedules a delayed send action
2. we continue processing, execute n arbitrary amount of transitions etc
3. now the "current state" doesn't hold this delayed send action, you can't easily record it (unless you track it on your own), nor you can persist the state of "active" timers
4. you've lost this delayed send action and you won't be able to rehydrate it

There is also an additional concern here - how do we "seed" the rehydrated interpreter with "active" timers?

davidkpiano:
Will revisit/reconsider this

Issue #2951, opened 2022-01-19
Transient initial state with lazily initiated context causes two child actor spawns


tkaragianes:
## Bug or feature request?
Bug. 

## Description:
When a parent machine spawns a child actor in a lazily initialized context (`  context: () => ({ child: spawn(child, "child") })`), and there is a transient initial state, the parent spawns two child actors: one referenced in the context, and one referenced in it's children property. Without the transient initial state, a single actor is spawned as expected.

## (Bug) Expected result:
A single child actor is spawned, with references in the children and context properties pointing to the same instance.

## (Bug) Actual result:
Two child actors are spawned, one referenced in the children property and the other referenced in the context property.

## (Bug) Potential fix:
The issue appears to lie in the interpreter's handling of the transient initial state. On the initial service.start() call, in `var resolvedState = initialState === undefined ? this.initialState ...` this.initalState has a spawned child in context, but an empty children property. I'm not sure where this.initialState is defined.

## Link to reproduction or proof-of-concept:

Codesandbox reproduction: https://codesandbox.io/s/brave-chihiro-6p0jy

Andarist:
Would you be able to test out if this gets fixed with the patch from [this PR](https://github.com/statelyai/xstate/pull/2943/files)? 

Even if it doesn't get fixed with that - we were just talking today with David about refactoring the code around initial actors and this work has been planned for the next two weeks. This should resolve issues like this holistically.

tkaragianes:
This patch doesn't quite fix it. Initial state now has a child actor referenced in children, and one in context, but they are different instances still. This isn't a blocker for me, so I can wait for the refactor. 

The reason this popped up is because invoked actors can be referenced on the children property, but aren't in context. If you want to invoke a sibling child actor B and pass it the reference to child actor A through the data property, you don't have access to the children map, only context and event. So I tried spawning child actor A rather than invoking, so it could be referenced in context, and passed to child actor B.

davidkpiano:
This no longer happens in the latest version of XState: https://codesandbox.io/s/modern-shape-1hrgp2?file=/src/index.js

Issue #2851, opened 2021-12-01
v5 - avoid mutation of `state.children`


davidkpiano:
This probably should be refactored to avoid mutation of `this.state` - I think most people treat this as immutable. Maybe grabbing the actor ref to stop within machine APIs~ and embedding that in the action would help? This way we wouldn't have to access this here by ID

_Originally posted by @Andarist in https://github.com/statelyai/xstate/pull/2840#discussion_r758498981_

davidkpiano:
Fixed in `xstate@beta`

Issue #2152, opened 2021-04-28
Bug with useSelector and "children"


SimeonC:
**Description**
Using `useSelector` in combination with an "actor" (invoked service, referenced via state.children), for some reason these two lines of code do not work the same. Even though the custom comparator function is exactly the same as the default. (Please see attached codesandbox as it's kind of hard to describe)

```
const count1 = useSelector(actor, countSelector); // this one never updates
const count2 = useSelector(actor, countSelector, (a, b) => a === b); // this one works (it's the "workaround")
```

**Expected Result**
Both `count1` and `count2` should function identically.

**Actual Result**
`count1` using the default comparator function never updates.

**Reproduction**
https://codesandbox.io/s/xstatereact-bug-reproduction-utt80

davidkpiano:
Added a Replay for this: https://replay.io/view?id=902fa34c-5fa0-4f60-bce9-d96f6db55edf&point=12656223600751063219641806915371361&time=6917.484634404804&hasFrames=true

jasonLaster:
Oh hey @SimeonC! @davidkpiano and I were pairing this morning and found some pretty interesting things in the replay. All of this is in the comments, but passing it along here too:

1.  The fact that the second counter has a custom comparator matters
2. The subscribe call is never called for the default comparator  [link](https://replay.io/view?id=902fa34c-5fa0-4f60-bce9-d96f6db55edf&point=12656223600751063219641806915371361&time=6917.484634404804&hasFrames=true)


It'd be interesting to dig into the subscribe call and see what is going on there

SimeonC:
Hey @jasonLaster! Small world 😀 Xstate and replay.io together is like a superhero crossover event!

I had a look through the replay too and added a comment. My guess is that the useEffect is getting re-called on every render with a non-default comparator forcing the actor to be re-subscribed to. I wonder what would happen if we added the `actor` to the dependencies of the useEffect 🤔 . I think that'd "fix" the issue, but it might cause other issues. Maybe the recreation of `state.children` in the parent machine as outlined in https://github.com/davidkpiano/xstate/issues/1967 is at fault for this?

I've updated my codesandbox to include a couple of variations on referentially stable selector/comparator and non-stable inline functions.

Andarist:
Yep, the problem is that a deferred actor is created at first and it's based on a null actor that has a stub `subscribe` method. So we subscribe to that and we never re-subscribe to the correct actor because we only depend on `selector` and `compare`:
[...]

Issue #1567, opened 2020-10-18
Using `spawn()` in `onEntry` vs in a transition doesn't return the same result for the same configuration


Shiroy:
**Description**
When a machine is spawned in action triggered in the `onEntry` event, the resulting interpreter can't be used with `useService`. The `machine` property is missing.

When I run the same code during a transition, the resulting interpreter can be used with `useService`.

**Expected Result**
The resulting interpreter should be usable with `useService` in both scenarios.

**Actual Result**
Exception:
```
Attempted to use an actor-like object instead of a service in the useService() hook. Please use the useActor() hook instead.
```

**Reproduction**
https://codesandbox.io/s/bullet-journal-8v5rg

**Additional context**
The machine property is missing on the returned interpreter. On `src/machines/todoList.ts:60`, if you uncomment `ref.machine = ...`, `useService` accept the reference and everything works as expected.

It's still un-clear whether I should use `useService` or `useActor`. However the typing clearly indicates that I have to use `useService` as the returned typed of `spawn` is expected to be an `Interpreter` when passing a `StateNode` or a `StateMachine` as an argument.

On `src/machines/todoList.ts:104`, I perform **the exact same operation** and it works as expected without any workaround.

davidkpiano:
Briefly investigating this, it's a bit tricky because when an actor is spawned as the initial state (in a not-yet-started service), that actor needs to be a "deferred actor" instead of a live service, so it will be an `Actor` type.

Making the interpreter an `Actor` instead to simplify the types would work, but would be a breaking change.

Here's one workaround you can do (see the `Actor<any>` typings; you can use `Actor<SomeEvent>` as well):

```ts
export function TodoItem({ item }: { item: Actor<any> }) {
  console.log(item);
  const [machine, send] = useActor(item);

  // ...
}

// ...

export default function TodoList(props: TodoListProps) {
  const [machine, send] = useMachine(todoListMachine, { devTools: true });

  const items = machine.context.items.map(i => {
    return <TodoItem item={i.ref as Actor<any>} key={i.uuid} />
  });
```

Andarist:
In the recent versions of `@xstate/react` we've removed `useService` altogether and from what I've checked the `useActor` works for your use case here.

PR #3186, opened 2022-03-30
Attempt to fix the deferred children problem

Andarist:
Is it pretty? No
Do i like it? No
Do i plan to refactor this? Also no
Does it work? Mostly

It would fix https://github.com/statelyai/xstate/discussions/3161
It would fix https://github.com/statelyai/xstate/issues/2951
It would fix https://github.com/statelyai/xstate/issues/2851
It would fix https://github.com/statelyai/xstate/issues/2152
It would fix https://github.com/statelyai/xstate/issues/1251
It would fix https://github.com/statelyai/xstate/issues/1567


packages/core/src/Actor.ts
```diff
@@ -49,57 +56,327 @@ export function createNullActor(id: string): ActorRef<any> {
...
+  sync: boolean;
 }
 
+export const wrappedMachines = new WeakSet();
+
+type DisposableObserver = Observer<any> & {
+  unsubscribe?: () => void;
+};
+
 export function createDeferredActor(
```
Andarist, line 71: this whole function is gross but it tries to keep the existing semantics as much as possible. The idea is to create an actor object as soon as possible - so we can give it to the user and don't change it afterward, we only call `.start()` on this from the outside and that call happens in the interpreter.

We still allow for non-lazy actors but we wrap them with a function `() => nonLazy` and "resolve" it here. So the existing code should mainly still work in the same but perhaps we should implement a warning for non-lazy actors so people could start refactoring their machines.

packages/core/src/Actor.ts
```diff
@@ -49,57 +56,327 @@ export function createNullActor(id: string): ActorRef<any> {
...
+          };
+
+          entityOrStop = entity(callbackSend, callbackOnReceive);
+        } catch (err) {
+          parent.send(error(name, err));
+          return;
+        }
+      }
+
+      const startInterpreter = (service) => {
```
Andarist, line 184: the problem with machines is that we currently wrap them with that "deferred actor" object - this means that what we hand to the user is no longer an interpreter, it's that deferred actor object. At the type-level, nothing changes because those things (`state.children`, return type of `spawn`, etc) were already typed as ActorRefs - but at the runtime level we effectively "remove" things (properties, methods) from here.

packages/core/src/interpreter.ts
```diff
@@ -954,16 +958,18 @@ export class Interpreter<
...
-              autoForward
-            };
           }
 
-          this.spawn(source, id, options);
+          const isLazyEntity =
+            isMachine(source) || isFunction(source) || isBehavior(source);
+          activity.deferred.start({
+            parent: this,
+            entity: isLazyEntity ? source : () => source
```
Andarist, line 969: so invoke entities are still resolved here, in the interpreter, they get suboptimal events/context arguments but at least the timing of things doesn't change here that much

packages/core/src/Actor.ts
```diff
@@ -49,57 +56,327 @@ export function createNullActor(id: string): ActorRef<any> {
...
+            }
+            parent.send(toSCXMLEvent(e, { origin: name }));
+          };
+          const callbackOnReceive = (newListener) => {
+            receivers.add({
+              send: newListener
+            });
+          };
+
+          entityOrStop = entity(callbackSend, callbackOnReceive);
```
Andarist, line 177: every lazy function will receive now those 2 arguments, but that also seems like a minor change that could be ignored

Andarist:
@davidkpiano I put some comments about possible behavior changes, there is not a lot of them in fact and they are quite subtle, but also... because they are so subtle I'm a little bit worried about shipping this PR. Maybe we need to rip off that bandaid though? Let me know what you think.

PR #3187, opened 2022-03-30
[v5] Remove `createModel`

davidkpiano:
Closes STA-1275


packages/core/package.json
```diff
@@ -55,7 +55,6 @@
       "./actions.ts",
       "./invoke.ts",
       "./behaviors.ts",
-      "./model.ts",
```
Andarist, line 58: What's the migration strategy that we will offer to people? Should we think about a codemod? Should we hand the model API to the community so they could publish and maintain a drop-in replacement package? Or do we propose to use a replacement snippet akin to [this one](https://github.com/statelyai/xstate/discussions/2323#discussioncomment-1030245) (note that this doesn't actually contain runtime event factories, but that can be added to it) ?

Andarist (commented):
We need a changeset here 😉

mattpocock:
We should mark this as deprecated in v4, right?

davidkpiano:
> We should mark this as deprecated in v4, right?

Yes, good catch. Need to figure out a good migration strategy first though 🤔

packages/core/package.json
davidkpiano: The migration will be part of v4 and not this PR. We will first have to mark it as deprecated in v4, and point to using `schema` instead.

This RFC is relevant: https://github.com/statelyai/rfcs/pull/12

PR #3190, opened 2022-04-04
[core] Add `waitFor` helper

davidkpiano:
This PR adds the opt-in `waitFor(...)` helper function, that asynchronously waits for the emitted value of an actor to satisfy a predicate before returning that value.

See discussion here: https://discord.com/channels/795785288994652170/955085268677963776

Basic usage:

```js
import { waitFor } from 'xstate/lib/waitFor';

// ...
const loginService = interpret(loginMachine).start();

const loggedInState = await waitFor(loginService, state => state.hasTag('loggedIn'));

loggedInState.hasTag('loggedIn'); // true
```

cc. @mattpocock @gavination 

TODO: 
- [x] Add changeset


packages/core/src/waitFor.ts
```diff
@@ -0,0 +1,51 @@
+import { ActorRef, EmittedFrom } from '.';
+
+interface WaitForOptions<_TActorRef extends ActorRef<any, any>> {
```
Andarist, line 3: What do we expect to add here if we are "reserving" a generic slot for it?

packages/core/src/waitFor.ts
```diff
@@ -0,0 +1,51 @@
...
+    timeout: TIMEOUT, // 10 seconds,
+    ...options
+  };
+  return new Promise((res, rej) => {
+    const handle = setTimeout(() => {
+      sub.unsubscribe();
+      rej(new Error(`Timeout of ${resolvedOptions.timeout} ms exceeded`));
+    }, resolvedOptions.timeout);
+
+    const sub = actorRef.subscribe((emitted) => {
```
Andarist, line 43: q: what should happen if we reach a final state without meeting the predicate condition first? Perhaps we should reject as soon as we reach that final state?

davidkpiano: Good catch - yes, we should.

packages/core/src/waitFor.ts
davidkpiano: Nothing right now... I'll remove it since it's only to be used internally.

packages/core/src/interpreter.ts
```diff
@@ -1090,7 +1090,7 @@ export class Interpreter<
       })
       .start();
 
-    return actor;
+    return actor as any;
```
Andarist, line 1093: was this change needed?

davidkpiano: Tests were failing without it locally

Andarist (approved):

PR #3194, opened 2022-04-05
Fixed synchronous races in `waitFor`

Andarist:


packages/core/test/waitFor.test.ts
```diff
@@ -48,4 +48,45 @@ describe('waitFor', () => {
...
+      states: {
+        a: {}
+      }
+    });
+
+    const service = interpret(machine).start();
+
+    await expect(
+      waitFor(service, (state) => state.matches('a'))
+    ).resolves.toHaveProperty('value', 'a');
```
Andarist, line 64: this was throwing before the fix since internal `sub` wasn't yet returned when the predicate matched so it was just crashing on `sub.unsubscribe()`

davidkpiano (approved):

PR #3198, opened 2022-04-07
Fixed an issue that prevented some states, containing broken~ toJSON, from being sent correctly to the inspector

Andarist:


packages/xstate-inspect/test/inspect.test.ts
```diff
@@ -1,6 +1,50 @@
...
 
+afterEach(() => {
+  // this clears timers, removes global listeners etc
+  // I'm not sure if this is 100% safe to do
+  // it's not clear if the window object after this operation is still usable in the same way (is it recyclable?)
+  // it does seem to cover our needs so far though
+  window.close();
+});
+
+const createIframeMock = () => {
```
davidkpiano, line 12: This is very cool - simpler than I thought!

davidkpiano (approved):

PR #3199, opened 2022-04-07
Fixed an issue that caused sending the same event multiple times to the inspector for restarted services

Andarist:
This builds on top of https://github.com/statelyai/xstate/pull/3198 so it should be merged after it lands on the main branch

### DIFF
--- a/packages/xstate-inspect/src/browser.ts
+++ b/packages/xstate-inspect/src/browser.ts
@@ -1,5 +1,6 @@
 import {
   ActorRef,
+  AnyInterpreter,
   EventData,
   EventObject,
   interpret,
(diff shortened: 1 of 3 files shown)


packages/xstate-inspect/src/browser.ts
```diff
@@ -86,6 +87,8 @@ const getFinalOptions = (options?: Partial<InspectorOptions>) => {
   };
 };
 
+const patchedInterpreters = new Set<AnyInterpreter>();
```
Andarist, line 90: I'd like to revisit the `@xstate/inspect` API in the future but it's worth fixing this asap, especially given that React 18 restarts services in StrictMode by default.

davidkpiano (approved):

PR #3200, opened 2022-04-07
`waitFor` fixes for cases when starting to `waitFor` stopped interpreter

Andarist:


packages/core/src/interpreter.ts
```diff
@@ -429,12 +429,16 @@ export class Interpreter<
     this.listeners.add(listener);
 
     // Send current state to listener
-    if (this.status === InterpreterStatus.Running) {
+    if (this.status !== InterpreterStatus.NotStarted) {
```
Andarist, line 432: This is a small change in behavior. I was afraid to emit this at all times because `this.initialState` might be impure

packages/core/src/interpreter.ts
```diff
@@ -429,12 +429,16 @@ export class Interpreter<
...
       listener(this.state);
     }
 
     if (resolvedCompleteListener) {
-      this.onDone(resolvedCompleteListener);
+      if (this.status === InterpreterStatus.Stopped) {
+        resolvedCompleteListener();
```
Andarist, line 438: can't call this within `onDone` easily because there the callback expects to receive a `doneInvoke` event. Here we know that it's a parameter-less `observer.complete` though.

packages/core/test/waitFor.test.ts
```diff
@@ -57,30 +54,22 @@ describe('waitFor', () => {
...
     const service = interpret(machine).start();
 
-    setInterval(() => {
+    setTimeout(() => {
       service.send('NEXT');
     }, 10);
 
-    setTimeout(() => {
-      throw new Error('Should not reach here');
-    }, 50);
```
Andarist, line 76: similarly, this was never disposed and could be thrown when **other** test was executing and thus fail that other test. I don't think this was adding that much here and by using `await expect().rejects()` we make sure that this `waitFor` throws

davidkpiano (approved):

PR #3203, opened 2022-04-07
Add missing changeset about changes in the core from #3200

Andarist:


.changeset/red-walls-lay.md
```diff
@@ -0,0 +1,7 @@
+---
+'xstate': minor
+---
+
+pr: #3200
+
+Subscribing to a stopped interpreter will now always immediately emit it state and call a completion callback.
```
davidkpiano, line 7:
```suggestion
Subscribing to a stopped interpreter will now always immediately emit its state and call a completion callback.
```

davidkpiano (approved):

PR #3209, opened 2022-04-08
Add ESM build to @xstate/svelte package, fixes #2642

### LINKED ISSUE #2642: [@xstate/svelte] Missing ESM exports
**Description**
Right now, the distributed package of @xstate/svelte only exports CJS and this caused some problem after an update in SvelteKit (https://github.com/sveltejs/kit/issues/2418). 

**Expected Result**
In any case, Xstate core exports both CJS and ESM. Is there any reason why it isn't the same with @xstate/svelte? If there isn't any, I think it should have an ESM export.

schibrikov:
I think adding rollup is unnecessary here, so I just added a second tsc config in order to generate ESM version.

fixes #2642


schibrikov:
Not sure if it should be a major/minor/patch bump, since nothing is changed. It has some impact on build pipelines preferring ESM packages though.

PR #3210, opened 2022-04-08
Fixed issues with respawning the same actor within a macrotask

Andarist:
fixes #3161, this issue is also being fixed by https://github.com/statelyai/xstate/pull/3186

This is still a draft - I will comment on specific things in the code to explain why

### DIFF
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -55,11 +55,13 @@ import {
   toGuard,
   evaluateGuard,
   toArray,
-  isArray
+  isArray,
+  isActor
 } from './utils';
 import { State } from './State';
 import { StateNode } from './StateNode';
 import { IS_PRODUCTION } from './environment';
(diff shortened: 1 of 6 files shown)


packages/core/src/actions.ts
```diff
@@ -432,11 +452,30 @@ export function stop<TContext, TEvent extends EventObject>(
 
   return {
     type: ActionTypes.Stop,
+    subtype: 'spawn',
```
Andarist, line 455: some tests are still using `stop` over `invokeStop` when testing the `createMachine` API itself - and because of that, some snapshots see `'spawn'` from here. I could adjust the snapshot-like tests... or refactor them to use `invokeStop` and the problem would be gone. I think it's an acceptable solution - not many people are inspecting the shape of the action objects, this is usually handled for them by the interpreter. 

Thoughts?

PR #3213, opened 2022-04-11
Andarist/spawn stuff

Andarist:
Just a bunch of tweaks - can be reviewed one commit at a time


packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -997,7 +997,8 @@ describe('useMachine (strict mode)', () => {
     );
   });
 
-  it('custom data should be available right away for the invoked actor', (done) => {
+  // TODO: figure out the story for `invoke.data`
+  it.skip('custom data should be available right away for the invoked actor', (done) => {
```
davidkpiano, line 1001: Probably a separate PR, but maybe we can deprecate `data` in favor of `input` and put that on the `xstate.init` event's payload? This would be a deviation from SCXML though 🤔

davidkpiano (approved):

packages/xstate-react/test/useMachine.test.tsx
Andarist: I'm up for it - just wondering about exact semantics and the types. Following SCXML implies that this `data` (or `input`, or however we call it) gets **automatically** assigned to the context of the invoked "session". This makes expressing the correctness of those "assignments" at the type level quite difficult.

PR #3215, opened 2022-04-11
Allow passing `Infinity` to `waitFor` as timeout

### LINKED ISSUE #3214: Bug: Timeout of `Infinity` causes `waitFor` to reject immediately
### Description

This is caused by the use of setTimeout - `setTimeout(fn, Infinity)` invokes `fn` immediately.

### Expected result

Either:

- Throw an explicit error on infinite timeout values. It shouldn't reject instantly, this leaks implementation details of `waitFor`
- Remove the timeout

I think the second option is better as it allows for deferring the timeout to the user. I agree that a timeout is the right default but there are definitely scenarios where you want to handle that yourself.

### Actual result

Instant rejection of the promise returned by `waitFor`

### Reproduction

https://codesandbox.io/s/stoic-browser-ev9s97?file=/src/index.js

### Additional context

_No response_

tom-sherman:
Closes #3214

WIP for now as I need figure out how to write a test for this, any pointers?

### DIFF
--- a/.changeset/cool-ducks-pretend.md
+++ b/.changeset/cool-ducks-pretend.md
@@ -1,5 +1,5 @@
 ---
-'xstate': minor
+'xstate': patch
 ---
 
 Removing the timeout that's built in to `waitFor` is now supported by explicitly passing an `Infinity` value.
--- a/.changeset/cool-ducks-pretend.md
+++ b/.changeset/cool-ducks-pretend.md
@@ -1,0 +1,20 @@
+---
+'xstate': minor
+---
+
+Removing the timeout that's built in to `waitFor` is now supported by explicitly passing an `Infinity` value.
+
+Example usage:
+
+```js
+import { waitFor } from 'xstate/lib/waitFor';
+
+// This will
+const loggedInState = await waitFor(
+  loginService,
+  (state) => state.hasTag('loggedIn'),
+  { timeout: Infinity }
+);
+```
+
+This fixes a bug that causes `waitFor` to reject with an error immediately due to the behaviour of `setTimeout`.
--- a/packages/core/src/waitFor.ts
+++ b/packages/core/src/waitFor.ts
@@ -45,7 +45,9 @@ export function waitFor<TActorRef extends ActorRef<any, any>>(
   return new Promise((res, rej) => {
     let done = false;
     if (process.env.NODE_ENV !== 'production' && resolvedOptions.timeout < 0) {
-        console.error('`timeout` passed to `waitFor` is negative and it will reject its internal promise immediately.')
+      console.error(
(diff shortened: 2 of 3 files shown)


packages/core/src/waitFor.ts
```diff
@@ -44,10 +44,13 @@ export function waitFor<TActorRef extends ActorRef<any, any>>(
   };
   return new Promise((res, rej) => {
     let done = false;
-    const handle = setTimeout(() => {
-      sub.unsubscribe();
-      rej(new Error(`Timeout of ${resolvedOptions.timeout} ms exceeded`));
-    }, resolvedOptions.timeout);
+    const handle =
+      resolvedOptions.timeout === Infinity
```
Andarist, line 48: do you perhaps have a link that would mention the valid max value for this? It seems that `Number.MAX_SAFE_INTEGER` also calls this immediately

tom-sherman: I read that it's a 32bit unsigned int so assumed that the max value was un-specced.

tom-sherman: https://developer.mozilla.org/en-US/docs/web/api/settimeout#maximum_delay_value

packages/core/src/waitFor.ts
```diff
@@ -44,10 +44,13 @@ export function waitFor<TActorRef extends ActorRef<any, any>>(
   };
   return new Promise((res, rej) => {
     let done = false;
-    const handle = setTimeout(() => {
-      sub.unsubscribe();
-      rej(new Error(`Timeout of ${resolvedOptions.timeout} ms exceeded`));
-    }, resolvedOptions.timeout);
+    const handle =
```
Andarist, line 47:
```suggestion
    if (process.env.NODE_ENV !== 'production' && resolvedOptions.timeout < 0) {
        console.error('`timeout` passed to `waitFor` is negative and it will reject its internal promise immediately.')
    }
    const handle =
```

Andarist:
> WIP for now as I need figure out how to write a test for this, any pointers?

You could create a `Promise.race` between your `waitFor` and some short-lived `setTimeout` (wrapped in a promise). Just to ensure that the `waitFor` doesn't reject immediately

davidkpiano (approved):

packages/core/src/waitFor.ts
Andarist: I wonder if, in the next major version, the `Infinity` shouldn't be the default. cc @davidkpiano

.changeset/cool-ducks-pretend.md
```diff
@@ -0,0 +1,20 @@
+---
+'xstate': minor
```
Andarist, line 2:
```suggestion
'xstate': patch
```

PR #3217, opened 2022-04-12
Added a failing test for inference with inline actions

mattpocock:


packages/core/test/actions.test.ts
```diff
@@ -1851,3 +1851,30 @@ describe('assign action order', () => {
...
+        };
+
+    createMachine({
+      schema: {
+        context: {} as Context,
+        events: {} as Event
+      },
+      on: {
+        EVENT_WITH_FLAG: {
+          actions: (context, event) => {
```
Andarist, line 1872: I've checked that this gets fixed if we wrap [this](https://github.com/statelyai/xstate/blob/7a698d4e3e6c8d6ee55bde2febd4667166e0f92d/packages/core/src/types.ts#L451-L460) with `Compute`. I need to do further analysis to check if this is the correct way of handling this.

Andarist: Minimal repro case:
[TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggxsAlgewHYB4AqBRAbhVYKCAD2AIBMBnKAbylEgC4orgAnRVAcygF8AfFAC8UABQRcLHPkIBKEUNzJEFANwAoDV3LsAZgEM40ALJGAFlwgBhNHsTcseAkVLlU1Og3AQWbTjz8QrQaUKxw5hAAtgYsIWFhki5U0s6EmmF8GVBoAPxxoQkA2gDSUFxQMi5FAESMEDUAuo35sAgoGFWExGSUNPT1LGV8ULmVaUQsqEnsAtkjAGR0hWEA5ABUq63wSGhOssBzhVkafFoUEHAANgbs0HoArqjtaFBwdwbkZhFW+y497k8Ax8fg4XF4gjEhTgdgcLG+lmmtlQ9kcXUOGjkLGUqk0GneEE+EARVjE8XCkRiBQSxAOKToIwM-W8zCgNQAYgB5Tk1fhQAA+XkGbIAQjAAEq8vgAGmOsrCaGpCS5nJYElwCmEwRWCUkADp6tkaQB6Y1QAB6uR1MrlpzkmiAA)

PR #3222, opened 2022-04-14
Added docs for waitFor

mattpocock:


docs/guides/interpretation.md
```diff
@@ -184,6 +184,62 @@ service.onTransition((state) => {
 service.start();
 ```
 
+## `waitFor`
+
+Lots of backend code relies on short-running processes, such as backend functions. This is especially true in serverless contexts, where code needs to boot up and shut down as fast as possible.
```
Andarist, line 189: This is mentioned as a primary use case for `waitFor` - it's not super clear from here what will happen with the `actor` when we timeout. Perhaps this should be called out and some recommendation/strategies should be given.

mattpocock: Fixed

davidkpiano (approved):

PR #3228, opened 2022-04-16
Fixed an issue with inline functions in the config object used as transition actions not having their argument types inferred

Andarist:
fixes https://github.com/statelyai/xstate/pull/3217

[TS minimal repro](https://www.typescriptlang.org/play?ts=4.6.2#code/C4TwDgpgBAggxsAlgewHYB4AqBRAbhVYKCAD2AIBMBnKAbylEgC4orgAnRVAcygF8AfFAC8UABQRcLHPkIBKEUNzJEFANwAoDV3LsAZgEM40ALJGAFlwgBhNHsTcseAkVLlU1Og3AQWbTjz8QrQaUKxw5hAAtgYsIWFhki5U0s6EmmF8GVBoAPxxoQkA2gDSUFxQMi5FAESMEDUAuo35sAgoGFWExGSUNPT1LGV8ULmVaUQsqEnsAtkjAGR0hWEA5ABUq63wSGhOssBzhVkafFoUEHAANgbs0HoArqjtaFBwdwbkZhFW+y497k8Ax8fg4XF4gjEhTgdgcLG+lmmtlQ9kcXUOGjkLGUqk0GneEE+EARVjE8XCkRiBQSxAOKToIwM-W8zCgNQAYgB5Tk1fhQAA+XkGbIAQjAAEq8vgAGmOsrCaGpCS5nJYElwCmEwRWCUkADp6tkaQB6Y1QAB6uR1MrlpzkmiAA)
[Minimal repro fixed with Compute](https://www.typescriptlang.org/play?ts=4.6.2#code/C4TwDgpgBAwg9gWzAV2BAPAQShAHmgOwBMBnKAQwJAD4oBeKAbygG0BpKASwKgGsIQcAGZRMAXQBco9mKgBfKADIoyArwJwA7gQDcAKD2hIogMbBOcAugAqAUQBuEAsBz4npJlCMQpJYACduAHN5WgYACgh7KTtHZwBKelp7OE4ifT1uNH8hchNoAFk8gAtuCHgCIU4gmwcnFzxCD2ZvXwDg0KY9KCgSE2KIBHIpRm6enDjgEhi6530euXmoSwB+KXgkVAwxntHx8fYuHlj6lgAibzOxMTXTc0t0Hf2oE+dXJrIW8B8oDgUVl6zFxSAhRCD+J49ahLBZKLrPKBnABUZ1umDMFisr2A0KecjGuPxeiIEBMABtyP5oEJVBjLFATFTyGgiv0yrVJu93J8vN82oECCE5NRwmMTJYqkEpKzSqCKpKOfVqHp4lIUmkMoyIMyIDKyuE9r1+oNhvDxmDnNMmApyDzWoiAGIAeSdZ3kUAAPp57WcAEKYABKbrkABoxqGxpYRk9nU6pJF7Ik6LRDfsogA6bww8YAehzUAAeis8WGFmG5PF9EA)

The problem is in the fact that we are using an intersection here and TS doesn't the contextual type for a function in such a situation. I'm working a on a fix for this in the TypeScript itself, [here](https://github.com/microsoft/TypeScript/pull/48668), but until that lands... we must avoid an intersection here and by using `Compute` we make the type at this position to be computed earlier and thus TS can treat it an object type when needed.

EDIT:// actually... `Compute`-based solution caused our CI to got out of memory, so I've simplified the fix and just "merged" the intersection manually, into a single mapped type

### DIFF
--- a/packages/core/test/types.test.ts
+++ b/packages/core/test/types.test.ts
@@ -514,6 +514,57 @@ describe('events', () => {
 
     acceptMachine(toggleMachine);
   });
+
+  it('should infer inline function parameters when narrowing transition actions based on the event type', () => {
+    createMachine({
+      schema: {
+        context: {} as {
+          count: number;
+        },
+        events: {} as
+          | { type: 'EVENT_WITH_FLAG'; flag: boolean }
+          | {
+              type: 'EVENT_WITHOUT_FLAG';
+            }
+      },
+      on: {
+        EVENT_WITH_FLAG: {
+          actions: (_context, event) => {
+            ((_accept: 'EVENT_WITH_FLAG') => {})(event.type);
+            ((_accept: boolean) => {})(event.flag);
+            // @ts-expect-error
+            ((_accept: 'is not any') => {})(event);
+          }
+        }
+      }
+    });
+  });
+
+  it('should infer inline function parameters when for a wildcard transition', () => {
+    createMachine({
+      schema: {
+        context: {} as {
+          count: number;
+        },
+        events: {} as
+          | { type: 'EVENT_WITH_FLAG'; flag: boolean }
+          | {
+              type: 'EVENT_WITHOUT_FLAG';
+            }
+      },
+      on: {
+        '*': {
+          actions: (_context, event) => {
+            ((_accept: 'EVENT_WITH_FLAG' | 'EVENT_WITHOUT_FLAG') => {})(
+              event.type
+            );
+            // @ts-expect-error
+            ((_accept: 'is not any') => {})(event);
+          }
+        }
+      }
+    });
+  });
 });
 
 describe('interpreter', () => {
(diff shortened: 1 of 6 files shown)


packages/core/test/types.test.ts
```diff
@@ -514,6 +514,57 @@ describe('events', () => {
...
+        events: {} as
+          | { type: 'EVENT_WITH_FLAG'; flag: boolean }
+          | {
+              type: 'EVENT_WITHOUT_FLAG';
+            }
+      },
+      on: {
+        EVENT_WITH_FLAG: {
+          actions: (_context, event) => {
+            ((_accept: 'EVENT_WITH_FLAG') => {})(event.type);
```
davidkpiano, line 533: This is pretty clever!

davidkpiano (approved):

Issue #3102, opened 2022-03-02
Bug: typegen types resolving events to never in guards/actions


SimeonC:
### Description

I only have this occurring in my one machine and it's fairly difficult to reduce down to a easy reproduction as I couldn't figure out what was the same between the errors. I attached a Typescript playground illustrating the issue with the working actions/implementations commented out. (The typegen is amazing BTW, so much easier and I can remove a lot of checks, looking forward to a CLI tool ✨ )

### Expected result

Event type should be correctly inferred for all actions/guards

### Actual result

`event` is resolved to `never` in certain guards and actions.

### Reproduction

[...]

PR #3230, opened 2022-04-16
Fixed an issue with typegen types not being able to provide events that had a union of strings as their `type`

Andarist:
fixes https://github.com/statelyai/xstate/issues/3102

why and how? https://github.com/statelyai/eng-blog/pull/81/files

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -5,13 +5,6 @@
   "main": "lib/index.js",
   "module": "es/index.js",
   "types": "lib/index.d.ts",
-  "typesVersions": {
-    "<4.1": {
-      "lib/types.4.1.d.ts": [
-        "types.pre-4.1.d.ts"
-      ]
-    }
-  },
   "sideEffects": false,
   "funding": {
     "type": "opencollective",
@@ -22,8 +15,7 @@
     "lib/**/*.js",
     "lib/**/*.d.ts",
     "es/**/*.js",
-    "es/**/*.d.ts",
-    "*.d.ts"
+    "es/**/*.d.ts"
   ],
   "keywords": [
     "statechart",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -5,6 +5,13 @@
   "main": "lib/index.js",
   "module": "es/index.js",
   "types": "lib/index.d.ts",
+  "typesVersions": {
+    "<4.1": {
+      "lib/types.4.1.d.ts": [
+        "types.pre-4.1.d.ts"
+      ]
+    }
+  },
   "sideEffects": false,
   "funding": {
     "type": "opencollective",
@@ -15,7 +22,8 @@
     "lib/**/*.js",
     "lib/**/*.d.ts",
     "es/**/*.js",
-    "es/**/*.d.ts"
+    "es/**/*.d.ts",
+    "*.d.ts"
   ],
   "keywords": [
     "statechart",
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -9,6 +9,8 @@ import {
   TypegenConstraint,
   AreAllImplementationsAssumedToBeProvided
 } from './typegenTypes';
+import { IndexByType } from './types.4.1';
+export { IndexByType };
 
 export type AnyFunction = (...args: any[]) => any;
 type ReturnTypeOrValue<T> = T extends AnyFunction ? ReturnType<T> : T;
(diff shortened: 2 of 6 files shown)


packages/core/package.json
```diff
@@ -5,6 +5,13 @@
   "main": "lib/index.js",
   "module": "es/index.js",
   "types": "lib/index.d.ts",
+  "typesVersions": {
+    "<4.1": {
+      "lib/types.4.1.d.ts": [
+        "types.pre-4.1.d.ts"
+      ]
+    }
+  },
```
Andarist, line 14: I still need to verify in some external project that this will be resolved correctly

packages/core/package.json
```diff
@@ -15,7 +22,8 @@
     "lib/**/*.js",
     "lib/**/*.d.ts",
     "es/**/*.js",
-    "es/**/*.d.ts"
+    "es/**/*.d.ts",
+    "*.d.ts"
```
davidkpiano, line 26: Is this one needed?

Andarist: The plan was to keep types targeting lowe versions of TS in this file and that would have to be published. However, it seems that it's not possible to redirect internal file requests (see [here](https://github.com/microsoft/TypeScript-Website/pull/2344/files#r852949086)) so this strategy won't work.

To make it happen we would have to bundle all of our types, replace in it what we want, and point to that from `typesVersions`. It's way more work than what I've wanted to pull off here so I'm going to just revert this part of the change (optimization) and keep it in a TODO comment for v5.

packages/core/src/types.ts
```diff
@@ -20,9 +20,12 @@ export type Compute<A extends any> = { [K in keyof A]: A[K] } & unknown;
 export type Prop<T, K> = K extends keyof T ? T[K] : never;
 export type Values<T> = T[keyof T];
 export type Merge<M, N> = Omit<M, keyof N> & N;
+// TODO: replace in v5 with:
+// export type IndexByType<T extends { type: string }> = { [E in T as E['type']]: E; };
 export type IndexByType<T extends { type: string }> = {
-  [K in T['type']]: Extract<T, { type: K }>;
+  [K in T['type']]: T extends any ? (K extends T['type'] ? T : never) : never;
```
davidkpiano, line 26: I wonder why `T extends any` makes this work 🧐

davidkpiano (approved):

packages/core/src/types.ts
Andarist: https://github.com/statelyai/eng-blog/pull/81/files :P

This is just a way to make this conditional type distributive. Conditional types are distributive when they contain a "naked" type param:
```ts
type IsStringDistributive<T> = T extends string ? 1 : 0

// this will run the conditional type over each member of this union type
type Distributes = IsStringDistributive<string | number>
//   ^? 1 | 0

type IsStringNonDistributive<T> = [T] extends [string] ? 1 : 0

// this will run the conditional type on the union type "as is"
type DoesntDistribute = IsStringNonDistributive<string | number>
//   ^? 0
```

[TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgzgZWAJwJYDsDmARVcWoBGArsKgG4QA8AKgHxQC8UNUEAHsBOgCZxT40WKAH4oARigAuKAAYAUPNCQouQUVIR+zeEiE48BEmUpV1wgD5R0xALaEIyOvID0LqB4B6Ixcui6CLAA5AHt0NSNSCmp6JigAbRoAXTZObj4E80wUsUkZBT9VEK10YAi0Y2gdREDMUPDDCqjTLKgrG3tHZzcPKG8gA)
[TS playground with Omit variants](https://www.typescriptlang.org/play?&fbclid=IwAR3oPnf0KrLaCVZdkcZcnC4rxtz2WpV9SpsGa4XUNUHny5UwXE8BYISJDMw#code/C4TwDgpgBAsiDyBbAlsAPAFQDRQNID4oBeKAbygG0AFKZAOygFEAPAYwBsBXAEwjQGsIIAPYAzKNjz4AugC4J1aQG4oAXwBQ60JCgARZAGdgAJ2QAjTsGQA3CElSYcBYhKgRmwCHW4GoAQzoQKAB+WAQUdElneToIW2NNbWg4AFU6ZGEGEnIk+SNTOgBzFQCDAHcIYxjORDNKnGFgAAtK+TNhYXYIALUoAB8yKFyofPpi-zpy1qg6GrrjFVZhRDAu5nkAJQgl4240UaKcTjp+OmEyukINLXBoADlM-VGLK1stg052YBc4e3RU9KZHAAciSwPw6gA9JCoFAAHrBRK3PSGEzmSw2CDvT7fEhPNEvTF-NAAjJ0EFgiHQ2EIoA)

So this particular type here just checks if the current `K` extends any of the `T['type']` and it will match even if `T['type']` is a union. Whereas the previous version would fail because
```ts
'FOO' | 'BAR' extends 'FOO' ? 1 : 0 // 0
```

PR #3234, opened 2022-04-19
Added a `StateValueFrom` helper that can be used to extract valid state values from a machine

Andarist:
Just a helper type that was requested~ at least twice on Discord


mattpocock (approved):
Needs docs, but we can consider this for v5 docs.

davidkpiano (approved):

PR #3238, opened 2022-04-20
[core] Fix sendTo types

davidkpiano:
This PR fixes the wrong typings in `sendTo(...)`.

### DIFF
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -1653,7 +1653,7 @@ describe('sendTo', () => {
       context: () => ({
         child: spawn(childMachine)
       }),
-      entry: sendTo((ctx) => ctx.child as any, { type: 'EVENT' })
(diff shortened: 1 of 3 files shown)


packages/core/test/actions.test.ts
```diff
@@ -1653,7 +1653,7 @@ describe('sendTo', () => {
       context: () => ({
         child: spawn(childMachine)
       }),
-      entry: sendTo((ctx) => ctx.child as any, { type: 'EVENT' })
+      entry: sendTo((ctx) => ctx.child, { type: 'EVENT' })
```
Andarist, line 1656: we should add a test case validating that `sendTo((ctx) => ctx.child, { type: 'UNKNOWN' })` reports a type error

davidkpiano: Done in [b956a84](https://github.com/statelyai/xstate/pull/3238/commits/b956a8405e441bf75594852b68ef5f7e19eb1a24)

PR #3247, opened 2022-04-25
[xstate/fsm] v3 (WIP)

davidkpiano:

PR #3249, opened 2022-04-26
Add a failing test case for inline action objects inference

Andarist:
To be investigated...

### DIFF
--- a/packages/core/test/types.test.ts
+++ b/packages/core/test/types.test.ts
@@ -498,7 +498,7 @@ describe('events', () => {
     service.send({ type: 'UNKNOWN' });
   });
 
-  it('event type should be inferrable from a simple state machine typr', () => {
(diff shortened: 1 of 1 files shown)


davidkpiano:
> i've fixed this with a beautiful hack - adding a call signature to `AssignAction`, this makes TS to treat it more like a function during inference, and thus some stuff changes priority etc

Funny enough, I discovered this too while writing types for `@xstate/fsm`: https://github.com/statelyai/xstate/pull/3247/files#diff-84836580b55f88df972e7ad66292fb03d1ba4c5ba2af7d68ce7903a71ecb1432R258-R265

Andarist:
The mentioned "let's use a call signature to fix inference" strategy was implemented in https://github.com/statelyai/xstate/pull/4127

PR #3254, opened 2022-04-27
Add support for the typegen in `@xstate/svelte`

Andarist:


packages/xstate-svelte/src/useMachine.ts
```diff
@@ -1,40 +1,71 @@
...
+          TMachine['__TContext'],
+          TMachine['__TEvent'],
+          TMachine['__TResolvedTypesMeta'],
+          true
+        >
+    ]
+  : [
+      options?: InterpreterOptions &
+        UseMachineOptions<TMachine['__TContext'], TMachine['__TEvent']> &
+        InternalMachineOptions<
```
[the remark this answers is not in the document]
Andarist: Not right now because `InternalMachineOptions` is also used within `createMachine` and `TMachine` doesn't exist there (yet). Of course, a helper type could be introduced that would be tailored for this use case here - but it's probably better to clean it up later in v5. This pattern is already used both in `@xstate/react` and `@xstate/vue`

packages/xstate-svelte/src/useMachine.ts
```diff
@@ -75,5 +106,5 @@ export function useMachine<
     }).unsubscribe;
   });
 
-  return { state, send: service.send, service };
+  return { state, send: service.send, service } as any;
```
davidkpiano, line 109: Wonder why the typings are complaining here?

davidkpiano (approved):

packages/xstate-svelte/src/useMachine.ts
Andarist: This probably is just missing logic in TypeScript to "resolve" constraints of the conditional types in a generic context (or something like that). Although I've thought that there were some improvements in this area in the latest versions, perhaps they didn't cover a scenario like this

[Minimal repro](https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGwIB4AqBhA9gOyQD2AD4oBeKAbwCgooBjYPALigx32EoF9LKBLXCACcAZvFrQAsmIAW-FKwEFiVGgHMIwOIgjMAFAEpSxLUjRZFRADTUoAeltRgsgM5Qc0AO6ZBAa2eWoAFt4EAAjaAArAFdnYChw2ngY6CdoAAN4bBA0qHh+AHI4j0F4MEgAE0dpCGwoG3soXmB812coyEEoWJLeVWlgYW8PeEFK4Ew6atofRuEq6EENKMFa0EggxFpq1wysnPhXCAA3IRAnflVG1wPnXux4UIAbFInd7PqHdU0EJD1DEmICnYAS4PDW0AA4hpzOw0MQyKgoBACDVyq4pFs5Mh+MIhFAAErEAD8BKgzGwxyEPGEUWw9F4OCgXxM8gxsgpSJR2DRUDZWMyIEIhF0wUxFOYqD5FP0zChwBhyOAaClEGUNkWwGWtVF7IgADpmT8IAYuEA)

PR #3255, opened 2022-04-27
Fix part of the current type errors on the `next` branch

Andarist:
Still looking into fixing the rest


davidkpiano (approved):
I trust you here 😜

PR #3261, opened 2022-04-29
Improve example templates

davidkpiano:
This PR improves the templates to use a new `timeMachine` that fetches the current time and assigns it to `context`:

```ts
const timeMachine = createMachine(
  {
    id: 'time',
    initial: 'idle',
    context: {
      time: undefined
    },
    states: {
      idle: {
        on: {
          fetch: 'pending'
        }
      },
      pending: {
        invoke: {
          src: 'fetchTime',
          onDone: {
            actions: assign({
              time: (_, event) => event.data
            })
          }
        },
        on: {
          cancel: 'idle'
        }
      },
      success: {
        on: {
          refetch: 'pending'
        }
      }
    }
  },
  {
    services: {
      fetchTime: () =>
        new Promise((res) => {
          setTimeout(() => {
            res(new Date().toLocaleTimeString());
          }, 1000);
        })
    }
  }
);
```

The goal of this new template is to provide a statechart example that isn't as contrived as the original toggle example, has more states and transitions, and also shows the usefulness of invoking states.


examples/template-ts/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    initial: 'idle',
+    context: {
+      time: undefined
+    },
+    states: {
+      idle: {
+        on: {
+          fetch: 'pending'
+        }
+      },
```
KrisCoulson, line 20: I really like this example. One thing for clarity might be worth capitalizing states and actions or using some type of differentiation to say that those are user-defined and not special xstate keys. especially when things like fetch already exist in javascript.

examples/template-ts/package.json
```diff
@@ -1,15 +1,18 @@
...
+  "main": "index.html",
   "scripts": {
-    "dev": "vite",
-    "build": "tsc && vite build",
-    "serve": "vite preview"
+    "start": "parcel index.html --open",
+    "build": "parcel build index.html"
+  },
+  "dependencies": {
+    "parcel-bundler": "1.12.5",
```
Andarist, line 11: nit: this usually should be a dev dep

examples/template-ts/package.json
```diff
@@ -1,15 +1,18 @@
...
+    "xstate": "^4.32.1"
   },
   "devDependencies": {
     "typescript": "^4.7.4",
     "vite": "^2.9.13"
```
Andarist, line 16: you have switched to using parcel for this example so vite should be dropped from here

packages/core/CHANGELOG.md
```diff
@@ -503,10 +503,10 @@
 
   model.createMachine({
     // `ctx` was of type `any`
-    entry: ctx => {},
+    entry: (ctx) => {},
```
Andarist, line 506: We are using a different Prettier version than Changesets and this results in a lot of those reformatting changes. I have a PR open that is supposed to fix this: https://github.com/changesets/changesets/pull/905

PR #3265, opened 2022-05-02
removed asEffect and asLayoutEffect from docs because it no longer exists

SerkanSipahi:


davidkpiano (approved):

davidkpiano:
@laurakalbag @mattpocock Do we want to mark this as deprecated/"removed since" in the docs for historical purposes, or remove them outright?

Andarist (approved):

Andarist:
I think removing those is fine - they probably weren't used all that often anyway and we didn't get any complaints (yet) about those being removed.

PR #3282, opened 2022-05-10
Drop support for handling promises returned by `fromCallback`

Andarist:
My main rationale for this is that it's an easy mistake to do this:
```ts
fromCallback(async (sendBack) {
  const api = await getSomeApi()
  api.onFoo = (val) => sendBack({ type: 'EV', val })
})
```

The problem with this code is that this promise implicitly completes very quickly and thus the parent stops accepting events sent out from this callback. Here the `async` was used as a convenient mechanism to use `await`, the intention wasn't really to complete this callback actor with `undefined`. Note that this is not a made-up example - this has been reported a while back on Discord by Kent and the resulting situation was quite hard to debug for a user because XState was logging cryptic errors in this case.

This can easily be refactored to:
```ts
fromCallback(() => {
  (async (sendBack) {
    const api = await getSomeApi()
    api.onFoo = (val) => sendBack({ type: 'EV', val })
  })()
})
```

Note that supporting promises (what is being removed from here) doesn't also play that well with current types as the return value of `fromCallback` is `Behavior<TEvent, undefined>` whereas that `undefined` is not a strong guarantee with a returned promise.


Andarist:
Alternatively we could allow promises **but** ignore the completion value. In such a case though - what to do about errors? We probably should handle them but then out handling wouldnt be symmetric (and thus somewhat weird)

davidkpiano (approved):
I'm in favor of this; can you re-merge the `v5/improved-spawn-1` PR as there are new changes?

Andarist:
This was already approved before but it got stale and I just rebased it. @davidkpiano could you take a second look at this?

davidkpiano (approved):

PR #3288, opened 2022-05-12
nextEvents should be calculated properly when the transition is internal

mattpocock:


Andarist:
As mentioned on Discord - this probably is not something that we want to change as it makes sense for `nextEvents` to correspond to `state.can` and `state.can` returns `false` for forbidden transitions (and this one is such).

PR #3289, opened 2022-05-12
Implement `predictableActionArguments` feature flag

Andarist:
Although this PR's main goal wasn't that - it can be used to supersede https://github.com/statelyai/xstate/pull/3210 and thus it fixes the problem described [here](https://github.com/statelyai/xstate/discussions/3161)

The main goal of this work is to:
- always call/resolve actions with a source event - the event **directly** responsible for a given transition (currently builtin actions are called with that but custom actions are called with events responsible for triggering the current macro transition)
- ~do not call/resolve actions with an explicit NULL event (`{ type: '' }`). Instead, we are calling them with the source event as the always transitions are also called **eventless** transitions (this matches SCXML spec etc)~ 
- allow us to implement an improved & correct algorithm for generating typegen information (I think that once we land this PR and the stop event PR then the algorithm in the typegen should be adjusted to account for those changes and that typegen should enforce using this feature flag).


davidkpiano:
Bikeshed: `predictableActionArguments` is a pretty vague description of what this does, since it's mainly for preserving events in transient states. Maybe `preserveEvents`? Would be similar to `preserveActionOrder`

Andarist:
I've chosen that because it also enables the `preserveActionOrder` automatically - so it's both about context and event arguments. I'm OK with doing changes to that though.

Andarist:
@davidkpiano any thoughts about the core of this PR? I'd like to know how you feel about it before polishing it further

davidkpiano:
> @davidkpiano any thoughts about the core of this PR? I'd like to know how you feel about it before polishing it further

My primary concern is the spreading of `isTransient` throughout many methods; perhaps we can have a `.microTransition(...)` separate method instead or something like that?

Andarist:
> My primary concern is the spreading of isTransient throughout many methods; perhaps we can have a .microTransition(...) separate method instead or something like that?

"The first" method that receives this argument is already `resolveRaisedTransition`:
https://github.com/statelyai/xstate/blob/430362d1f6ef2482772a9bea28ea54a88f2b3798/packages/core/src/StateNode.ts#L1216

it just calls to some other functions (as the internal machinery is reused, in the same way, it was reused before), and those need to pass it around as the `event.type` alone is not sufficient for candidate transitions selection. IIRC we "only" need to get that information to the `getCandidates` call here:
https://github.com/statelyai/xstate/blob/430362d1f6ef2482772a9bea28ea54a88f2b3798/packages/core/src/StateNode.ts#L908

This is mostly internal and we might figure out a better algorithm in the future so I wouldn't be bothered about this that much.

packages/core/test/predictableExec.test.ts
```diff
@@ -0,0 +1,587 @@
...
+      }
+    });
+
+    const service = interpret(machine).start();
+    service.send({ type: 'NEXT' });
+
+    expect(eventArg.type).toBe('NEXT');
+  });
+
+  it('should call raised transition custom actions with source event', () => {
```
davidkpiano, line 133:
```suggestion
  it('should call raised transition custom actions with raised event', () => {
```

packages/core/test/predictableExec.test.ts
```diff
@@ -0,0 +1,587 @@
...
+      }
+    });
+
+    const service = interpret(machine).start();
+    service.send({ type: 'NEXT' });
+
+    expect(eventArg.type).toBe('RAISED');
+  });
+
+  it('should call raised transition builtin actions with source event', () => {
```
davidkpiano, line 163:
```suggestion
  it('should call raised transition builtin actions with raised event', () => {
```

packages/core/test/predictableExec.test.ts
```diff
@@ -0,0 +1,587 @@
...
+      }
+    });
+
+    const service = interpret(machine).start();
+    service.send({ type: 'NEXT' });
+
+    expect(eventArg.type).toBe('RAISED');
+  });
+
+  it('should call invoke creator with source event', () => {
```
davidkpiano, line 197:
```suggestion
  it('should call invoke creator with raised event', () => {
```

packages/core/test/predictableExec.test.ts
```diff
@@ -0,0 +1,587 @@
...
+      }
+    });
+
+    const service = interpret(machine).start();
+    service.send({ type: 'NEXT' });
+
+    expect(eventArg.type).toBe('RAISED');
+  });
+
+  it('invoked child should be available on the new state', () => {
```
davidkpiano, line 232: Was this not working before?

packages/core/test/predictableExec.test.ts
```diff
@@ -0,0 +1,587 @@
...
+    });
+
+    const service = interpret(machine).start();
+    service.send({ type: 'NEXT' });
+    service.send({ type: 'NEXT' });
+
+    expect(service.state.children.myChild).not.toBeDefined();
+  });
+
+  it('should automatically enable predictableActionArguments', () => {
```
davidkpiano, line 289: If I'm understanding this correctly...

```suggestion
  it('should automatically enable preserveActionOrder', () => {
```

packages/core/src/actions.ts
```diff
@@ -720,25 +731,28 @@ export function resolveActions<TContext, TEvent extends EventObject>(
...
             return resolvedActionsFromPure;
           }
           case actionTypes.stop: {
-            return resolveStop(
+            const resolved = resolveStop(
               actionObject as StopAction<TContext, TEvent>,
               updatedContext,
               _event
             );
+            predictableExec?.(resolved, updatedContext, _event);
```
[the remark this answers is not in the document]
Andarist: I think this is OK given the current state of things - `.children` are already mutated and somewhat unstable: https://codesandbox.io/s/keen-cookies-vfte0i?file=/src/index.js:798-972

I think though that this should be fixed in v5

packages/core/test/predictableExec.test.ts
```diff
@@ -584,4 +585,25 @@ describe('predictableExec', () => {
 
     expect(actual).toEqual(['stop 1', 'start 2']);
   });
+
+  // TODO: this might be tricky in v5 because this currently relies on `.exec` property being mutated to capture the context values in a closure
```
davidkpiano, line 589: I don't think this will be tricky, especially if we execute actions eagerly in v5, as you suggested?

davidkpiano: That's why currently there's the notion of `preInitialState` in v5, which represents the initial state and actions _without_ any assignments performed nor actions executed. 

In my view, this is what `machine.initialState` (the pure getter) should be - state without actions executed. Then when the machine starts (is interpreted), the actions should be executed.

So in this test, `machine.initialState` would have context `{ count: 0 }` since no actions were executed yet, but the read state from `interpreter.subscribe(state => ...)` would have `{ count: 2 }` since the state after the initial state has those actions executed.

This would also be what the input RFC would solve, since there would be no need for initial `assign` actions.

Half-baked thoughts, just thinking out loud

davidkpiano:
What remains left for this to be merged in?

Andarist:
I've reverted the change related to making a NULL event a pseudo-event with this flag. Because of that, all the entry actions following always transitions would have to be callable with `TEvent` and that takes away a lot of the typegen appeal (for those cases, but they are not that rare). Since we already have this logic in v5 - we should revisit it and make a decision if it's good or if we should revert it. To postpone this decision I got rid off it from here.

mellson (approved):
I think this looks great!

docs/guides/actions.md
```diff
@@ -1,5 +1,25 @@
 # Actions
 
+::: warning
+
+It is advised to configure `predictableActionArguments: true` at the top-level of your machine config, like this:
```
davidkpiano, line 5: Can you add the XState version this applies to?

packages/core/src/types.ts
```diff
@@ -662,6 +662,10 @@ export interface StateNodeConfig<
    * @default false
    */
   preserveActionOrder?: boolean;
+  /**
+   * @default false
```
davidkpiano, line 666:
```suggestion
   * Whether XState calls actions with the event directly responsible for the related transition.
   * @default false
```

davidkpiano (approved):

docs/guides/actions.md
```diff
@@ -1,5 +1,25 @@
...
+  // ...
+});
+```
+
+This flag is an opt into some fixed behaviors that will be the default in v5. With this flag:
+
+- XState will always call an action with the event directly responsible for the related transition,
+- you also automatically opt-into [`preserveActionOrder`](https://xstate.js.org/docs/guides/context.html#action-order).
+
+Please be aware that you might not able to use `state` from the `meta` argument when using this flag.
```
LekoArts, line 19: Can y'all add more information how to migrate? We had trouble finding a replacement for using this. Thanks!

Andarist: Could you open a discussion that would describe your use case for using this property?

LekoArts: Sure!

LekoArts: @Andarist Done :) https://github.com/statelyai/xstate/discussions/3511

PR #3292, opened 2022-05-13
Fixed an issue in the `EmittedFrom` type helper that could prevent it from inferring the desired type from some service

Andarist:
fixes the issue reported here: https://discord.com/channels/795785288994652170/809564589880639548/974402582393024562

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1799,7 +1799,7 @@ export interface Behavior<TEvent extends EventObject, TEmitted = any> {
 }
 
 export type EmittedFrom<T> = ReturnTypeOrValue<T> extends infer R
-  ? R extends ActorRef<infer _, infer TEmitted>
+  ? R extends ActorRef<any, infer TEmitted>
     ? TEmitted
     : R extends Behavior<infer _, infer TEmitted>
     ? TEmitted
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1799,7 +1799,9 @@ export interface Behavior<TEvent extends EventObject, TEmitted = any> {
 }
 
 export type EmittedFrom<T> = ReturnTypeOrValue<T> extends infer R
-  ? R extends ActorRef<any, infer TEmitted>
+  ? R extends Interpreter<infer _, infer __, infer ___, infer ____, infer _____>
+    ? R['initialState']
+    : R extends ActorRef<infer _, infer TEmitted>
     ? TEmitted
     : R extends Behavior<infer _, infer TEmitted>
     ? TEmitted
--- a/.changeset/popular-wolves-tap.md
+++ b/.changeset/popular-wolves-tap.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Fixed an issue in the `EmittedFrom` type helper that could prevent it from inferring the desired type from some services.
--- a/packages/core/test/typeHelpers.test.ts
+++ b/packages/core/test/typeHelpers.test.ts
@@ -2,6 +2,7 @@ import {
   assign,
   ContextFrom,
   createMachine,
+  EmittedFrom,
   EventFrom,
   interpret,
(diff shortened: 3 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -1799,7 +1799,7 @@ export interface Behavior<TEvent extends EventObject, TEmitted = any> {
 }
 
 export type EmittedFrom<T> = ReturnTypeOrValue<T> extends infer R
-  ? R extends ActorRef<infer _, infer TEmitted>
+  ? R extends ActorRef<any, infer TEmitted>
```
Andarist, line 1802: This is somewhat weird because my intuition was that it should not work when `R` has `TEvent` configured as `never`. 

And indeed it shouldn't work as per [this playground](https://www.typescriptlang.org/play?ts=4.6.2#code/JYOwLgpgTgZghgYwgAgIILAeygJQjAHgBUBRANwnGQgA9IQATAZ2QG9kwBPABwgC5kTMFFABzZAF8AfGwBQyQZQYCAFBDIDSFcAEpkAXhllMwBgG5ZE2bID0N5MDAOWtXhggMOACzhOwX4BYAd0wQAHInACMUOCYmYFEQOEiAGwhZLl5kIgghAEYDNAxsPEIQdWgZWnpmIqxcfAI4EE4ZAH5kAoEABlt7BQA9NutMlABlJWhickonaqUWdlGBIREQcWlCtW0wTRnwAG0InggwgF1kAB9s-bA9Q2RjUwtZUEhYRBR0etKAJmmdtQ6As2BwTithGJJDJWPJFIwBBNGFMtLMpJZrHZkM1PP5As5BGBgCkUsgQJgnLF4olkmkMidsrkwL9Ct8So1yhQoFVgYwWGyGoRmq1kB0usheljBm0gA). So it got me thinking... and I've realized that those becomes assignable when we turn off the `strictFunctionTypes` (see [here](https://www.typescriptlang.org/play?strictFunctionTypes=false&ts=4.6.2#code/JYOwLgpgTgZghgYwgAgIILAeygJQjAHgBUBRANwnGQgA9IQATAZ2QG9kwBPABwgC5kTMFFABzZAF8AfGwBQyQZQYCAFBDIDSFcAEpkAXhllMwBgG5ZE2bID0N5MDAOWtXhggMOACzhOwX4BYAd0wQAHInACMUOCYmYFEQOEiAGwhZLl5kIgghAEYDNAxsPEIQdWgZWnpmIqxcfAI4EE4ZAH5kAoEABlt7BQA9NutMlABlJWhickonaqUWdlGBIREQcWlCtW0wTRnwAG0InggwgF1kAB9s-bA9Q2RjUwtZUEhYRBR0etKAJmmdtQ6As2BwTithGJJDJWPJFIwBBNGFMtLMpJZrHZkM1PP5As5BGBgCkUsgQJgnLF4olkmkMidsrkwL9Ct8So1yhQoFVgYwWGyGoRmq1kB0usheljBsMgA)). 

The weird thing is that... our tests should be type-checked using strict mode but apparently this doesn't quite work 🤷‍♂️ 

Gonna fix this up in a sec.

packages/core/src/types.ts
```diff
@@ -1799,7 +1799,9 @@ export interface Behavior<TEvent extends EventObject, TEmitted = any> {
 }
 
 export type EmittedFrom<T> = ReturnTypeOrValue<T> extends infer R
-  ? R extends ActorRef<infer _, infer TEmitted>
+  ? R extends Interpreter<infer _, infer __, infer ___, infer ____, infer _____>
```
Andarist, line 1802: this might look a little bit strange because `Interpreter` implements `ActorRef` and `ActorRef` is already handled here - but, unfortunately, while it implements the `ActorRef`, its `send` method's signature is different and TS might not be able to actually treat it as `ActorRef` for the purpose of conditional types with `infer`

mattpocock (approved):

PR #3297, opened 2022-05-16
Added sync methods to xstate test

mattpocock:


packages/xstate-test/src/machine.ts
```diff
@@ -103,20 +103,6 @@ export function createTestModel<TMachine extends AnyStateMachine>(
             );
           })
         ),
-      testTransition: async (step) => {
```
mattpocock, line 106: Removed testTransition from options. According to our discussion it might be config overkill.

packages/xstate-test/src/TestModel.ts
```diff
@@ -194,6 +202,66 @@ export class TestModel<TState, TEvent extends EventObject> {
...
+  public testPlanSync(
+    plan: StatePlan<TState, TEvent>,
+    options?: Partial<TestModelOptions<TState, TEvent>>
+  ) {
+    for (const path of plan.paths) {
+      this.testPathSync(path, options);
+    }
+  }
+
+  public testPathSync(
```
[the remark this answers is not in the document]
davidkpiano: Will have to think further on this; can be refactored to some `baseTestPath()` private method that conditionally tests sync or async in some future PR

packages/xstate-test/src/testUtils.ts
```diff
@@ -0,0 +1,22 @@
+import { StatePlan } from '@xstate/graph';
```
davidkpiano, line 1: Refactor item: move this to the `/test/` directory instead (I will probably be confused again by this in the future 😂)

davidkpiano (approved):

PR #3298, opened 2022-05-16
Added defaults to getPlans and testPlans

mattpocock:


packages/xstate-test/src/TestModel.ts
```diff
@@ -83,9 +83,9 @@ export class TestModel<TState, TEvent extends EventObject> {
   }
 
   public getPlans(
-    planGenerator: PlanGenerator<TState, TEvent>,
-    options?: Partial<TraversalOptions<TState, TEvent>>
+    options?: GetPlansOptions<TState, TEvent>
```
mattpocock, line 86: `getPlans` now takes the `planGenerator` from options, instead of a required first argument.

packages/xstate-test/src/TestModel.ts
```diff
@@ -176,10 +171,8 @@ export class TestModel<TState, TEvent extends EventObject> {
     return Object.values(adj).map((x) => x.state);
   }
 
-  public async testPlans(
-    plans: Array<StatePlan<TState, TEvent>>,
-    options?: Partial<TestModelOptions<TState, TEvent>>
-  ) {
+  public async testPlans(options?: TestPlansOptions<TState, TEvent>) {
```
mattpocock, line 174: `testPlans` now takes plans from an options object instead of its first argument.

mattpocock: It means that the simplest API is:

```ts
const machine = createTestMachine();

const model = createTestModel(machine);

it('should work', model.testPlans);
```

Which I like.

davidkpiano: @mattpocock You can have that API with `testPlans(plans?, options?)` as well

mattpocock: Sure, you just can't pass options without passing plans, which seems unintuitive

davidkpiano (approved):
Approved (to be merged before overload PR #3307)

PR #3301, opened 2022-05-16
Added turborepo

mattpocock:
This PR:

1. Adds turborepo to run on `test` and `svelte-check` only (the slowest part of our setup by far)
2. Connects turborepo to Vercel via the `TURBO_TOKEN` to cache the results of our tests

This should speed up our CI significantly by only re-running tests that have already run.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -19,3 +19,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/ci-setup
       - uses: ./.github/actions/ci-checks
+        env:
+          TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+          TURBO_TEAM: statelyai
--- a/turbo.json
+++ b/turbo.json
@@ -1,0 +1,9 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "baseBranch": "origin/main",
+  "pipeline": {
+    "test": {
+      "outputs": []
+    }
+  }
+}
--- a/turbo.json
+++ b/turbo.json
@@ -4,6 +4,9 @@
   "pipeline": {
     "test": {
       "outputs": []
+    },
+    "svelte-check": {
+      "outputs": []
     }
   }
 }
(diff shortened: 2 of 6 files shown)


felixmosh:
Try this github action which replaces the need for vercel account
https://github.com/felixmosh/turborepo-gh-artifacts

.github/workflows/nodejs.yml
```diff
@@ -19,3 +19,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/ci-setup
       - uses: ./.github/actions/ci-checks
+        env:
+          TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
```
Andarist, line 23: I suspect that this might have a problem running on PRs as secrets shouldn't be exposed to the actions triggered by such event (it would be a security concern)

turbo.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "baseBranch": "origin/main",
+  "pipeline": {
+    "test": {
+      "outputs": []
+    },
+    "svelte-check": {
+      "outputs": []
+    }
```
Andarist, line 10: how does Turbo know when it should rerun those?

mattpocock: It re-runs them when they have changed against origin/main

.github/workflows/nodejs.yml
mattpocock: Ah, grim - true.

cc @jaredpalmer is there a solution here?

turbo.json
Andarist: But when **what** has changed?  🤔

mattpocock: When the individual packages have changed, or their dependents. It hashes the contents of the source files committed to git within the individual package.

Andarist: clever, would it recognize a dependency on the root `tsconfig.json`~? can it be added manually?

mattpocock: Yes, if anything changes in the root it'll blast everything

PR #3304, opened 2022-05-16
Added deduplication of generated paths

mattpocock:


packages/xstate-test/test/plans.test.ts
```diff
@@ -74,4 +98,51 @@ describe('testModel.testPlans(...)', () => {
       ]
     `);
   });
+
+  describe('When the machine only has one path', () => {
```
mattpocock, line 102: This test file feels a little thin, lmk if there's anything that you think might need adding.

packages/xstate-test/test/plans.test.ts
```diff
@@ -74,4 +98,51 @@ describe('testModel.testPlans(...)', () => {
...
+        }
+      });
+
+      const model = createTestModel(machine);
+
+      const plans = model.getPlans();
+
+      expect(plans).toHaveLength(1);
+    });
+  });
```
davidkpiano, line 127: Can you also add a test for multiple transitions between states (e.g. `on: { ANOTHER_EVENT: 'b' }`) to ensure that there exists a path with those events?

This will pass of course, just a good sanity check to have

davidkpiano (approved):
Approved with future optimizations (will implement in follow-up PR):

- Aggregated paths to be sorted by length descending
- Use `for (...)` instead of `.some(...)` (reason on next point)
- Short-circuit when `path === path` in `for (...)` loop

packages/xstate-test/src/dedupPathPlans.ts
```diff
@@ -0,0 +1,97 @@
+import { StatePath, StatePlan } from '@xstate/graph';
+import { EventObject } from 'xstate';
+import { PlanGenerator } from './types';
+
+/**
+ * Deduplicates your path plans so that A -> B
+ * is not executed separately to A -> B -> C
+ */
+export const planGeneratorWithDedup = <TState, TEvent extends EventObject>(
```
Andarist, line 9: With this in place, I think that the comment made [here](https://github.com/statelyai/xstate/pull/3036/files#r874904515) makes even more sense. If we'd have a better traversal algorithm then we wouldn't have to dedup anything - shouldn't it mostly boil down to **only** pushing to the generated plans/paths when we reach the end condition? If we wouldn't push the intermediate paths to the result data then we wouldn't have anything to dedupe.

davidkpiano: These are actually separate concerns. Graph traversal should provide _all_ of the paths, and then these are deduplicated after the fact. The reason is that we cannot know ahead of time if a single path also covers many other paths without generating the other paths first.

With that said, you're right about the `toPredicate` optimization that can be made in `@xstate/graph`; it's just that deduplication should _not_ happen in `@xstate/graph`.

Andarist: I've meant to add this comment to the main PR but accidentally added it here 🙈 Let's continue here though to keep this particular discussion in a single place.

> it's just that deduplication should not happen in @xstate/graph.

Isn't deduplication just another type of filtering? It feels that if `@xstate/graph` is already responsible for filtering (conditionally) then it could also be responsible for avoiding duplicates (conditionally). I'm not saying that `@xstate/graph` should **by default** remove those duplicates (although I'm not sure when I would like to have those duplicates generated, you probably would know more about use cases for that).

> The reason is that we cannot know ahead of time if a single path also covers many other paths without generating the other paths first.

I don't think that's true - but I might misinterpret the whole algorithm completely. From what I understand we are trying to explore/traverse the graph until we explore it all (we need to be able to confirm somehow that the traversal is complete) or until we explore all of the paths meeting some criterion (again, we need to be able to somehow stop the exploration/traversal, this is encoded in some kind of a predicate/test function).

Isn't it true that until we meet a "stop condition" then all the subpaths (from the initial state) leading to the current target states would be considered duplicates (they are literally just subpaths of this "final" path)? And isn't it true that we can't have duplicates **among** different "branches" of the traversal? I mean, in a case like this:
```
A -> B -> C
A -> B -> D
```
We can strongly guarantee that there are no duplicates in this set because those paths are "final" (a stop condition has been met) and to even generate those 2 we had to have some diverge point (the path diverged in B to C and D). Even though the `A -> B` subpath is shared between both "final" paths we'll still have to test both of those final paths from the beginning because we can't reliably test each of those paths individually without actually starting from the initial state (A).

So isn't it true that the "deduplication" would automatically work if we'd just avoid pushing to the generated plans when we decide to explore/traverse further and we'd only push to the plans when we meet the stop condition?

PR #3306, opened 2022-05-17
Optimize path deduplication algorithm

davidkpiano:
This PR optimizes the path deduplication algorithm by:

- Checking paths in path length order descending
- Checking event sequences from the start instead of `.includes()` - for state machines, the subpath must be considered from the initial state, since events "A, B, C" can result in a completely different state traversal in the middle of a state machine rather than at the start

This greatly reduces the number of comparisons made.


mattpocock:
Beautiful

PR #3308, opened 2022-05-17
Removed testPlans function

mattpocock:


packages/xstate-test/src/testUtils.ts
```diff
@@ -0,0 +1,22 @@
+import { StatePlan } from '@xstate/graph';
+import { TestModel } from './TestModel';
+
+const testModel = async (model: TestModel<any, any>) => {
+  for (const plan of model.getPlans()) {
+    await model.testPlan(plan);
+  }
+};
```
davidkpiano, line 8: Let's not include `testModel` as a function - this is the same as `testPlans()` which is more explicit, and `testModel` is (perhaps confusingly) used as a variable for "the test model"

```suggestion
```

packages/xstate-test/src/testUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+  plans: StatePlan<any, any>[]
+) => {
+  for (const plan of plans) {
+    await model.testPlan(plan);
+  }
+};
+
+export const testUtils = {
+  testPlans,
+  testModel
```
davidkpiano, line 21:
```suggestion
```

packages/xstate-test/test/plans.test.ts
```diff
@@ -67,7 +68,7 @@ describe('testModel.testPlans(...)', () => {
       }
     });
 
-    await testModel.testPlans(plans);
+    await testUtils.testModel(testModel);
```
davidkpiano, line 71: A model isn't what's tested; it's plans and paths that are tested. I feel this can be replaced in the future by:

```js
await testModel.getPlans().test();
```

davidkpiano (commented):
If we remove `.testPlans(...)`, it should be because we encourage this pattern _for now_:

```js
const plans = testModel.getPlans();

plans.forEach(plan => {
  // maybe add describe, it, etc.
  plan.test();
})
```

Instead of providing sugar for "test all of these plans". Or if we want sugar for getting the default plans and testing them all:

```js
await testModel.test();
```

mattpocock:
@davidkpiano Yes, correct. Do I need to change anything in the PR based on your observation?

packages/xstate-test/src/testUtils.ts
mattpocock: Let's bikeshed this outside of this PR.

PR #3319, opened 2022-05-20
Fixed `ObservableActorRef` subscriptions and remove some outdated/redundant code

Andarist:
*comment placeholder*


packages/core/src/Mailbox.ts
```diff
@@ -24,6 +24,22 @@ export class Mailbox<T> {
     }
   }
 
+  // TODO: rethink this design
+  public prepend(event: T): void {
```
Andarist, line 28: Why?
1. to process the start signal **first** in the `ObservableActorRef`: https://github.com/statelyai/xstate/blob/76d738f0f8df6f6782a9b6d9b9bd565b4f6e1627/packages/core/src/ObservableActorRef.ts#L32
2. the mailbox can receive and queue events before the actor gets started
3. but they were ignored here because the inner service was not yet created (because that requires a start signal): https://github.com/statelyai/xstate/blob/76d738f0f8df6f6782a9b6d9b9bd565b4f6e1627/packages/core/src/actors.ts#L431

packages/core/src/ObservableActorRef.ts
```diff
@@ -6,14 +6,13 @@ import {
...
 } from './types';
-import { symbolObservable, toSCXMLEvent } from './utils';
+import { symbolObservable, toObserver, toSCXMLEvent } from './utils';
 import { Mailbox } from './Mailbox';
 import { LifecycleSignal, startSignal, stopSignal } from './actors';
 
 export class ObservableActorRef<TEvent extends EventObject, TSnapshot>
   implements ActorRef<TEvent, TSnapshot> {
   private current: TSnapshot;
-  public deferred = true;
```
Andarist, line 16: I was able to remove this completely - this was mostly needed by the `useActor` but in v5 we have lazy actors so we don't have to recompute the initial state when starting, it's all cached. Actors are also capable of queueing events before they are started - so this flag is no longer needed for anything

packages/core/src/ObservableActorRef.ts
```diff
@@ -74,12 +73,19 @@ export class ObservableActorRef<TEvent extends EventObject, TSnapshot>
...
       typeof this.context._event.type !== 'string'
         ? (this.context._event as LifecycleSignal)
         : (this.context._event as SCXML.Event<TEvent>).data,
       this.context
     );
+
+    if (previous !== this.current) {
+      const current = this.current;
+      this._observers.forEach((observer) => observer.next?.(current));
+    }
```
Andarist, line 89: This was a crucial missing piece. You could have subscribed to the ref - but u wouldn't ever be notified about any change.

packages/core/src/StateMachine.ts
```diff
@@ -328,7 +328,12 @@ export class StateMachine<
       undefined
     );
     nextState.actions.unshift(...preInitialState.actions);
-    return macrostep(nextState, null as any, this) as StateFrom<typeof this>;
+
+    const macroState = macrostep(nextState, null as any, this) as StateFrom<
+      typeof this
+    >;
+    macroState.changed = undefined;
```
Andarist, line 335: `.changed === undefined` somewhat means the initial state but it wasn't delivering on that promise here - because `macrostep` was already computing a boolean for it as the `nextState` here already didn't have `._initial` set on it - that's only available on the `preInitialState` that was already "changed" by the `resolveMicroTransition` just 2 lines above this

sideways... i'd really like to explore removing `.changed` altogether

packages/core/src/actors.ts
```diff
@@ -366,63 +366,71 @@ export function fromMachine<TMachine extends AnyStateMachine>(
...
 ): Behavior<EventFrom<TMachine>, StateFrom<TMachine>> {
+  const snapshotEventType = Symbol('snapshot');
+
   const castedMachine = machine as TMachine;
   let service: InterpreterFrom<TMachine> | undefined;
-  let subscription: Subscription;
   let initialState: StateFrom<TMachine>;
 
-  const behavior: Behavior<EventFrom<TMachine>, StateFrom<TMachine>> = {
+  // TODO: use better type for `TEvent`
```
Andarist, line 377: The problem is that we are also handling signals here, so we cant simply use `TEvent` here. We also can't use smth like `TEvent | LifecycleEvent` because we also need to process `typeof snapshotEventType`.

I'm not yet sure what exact design to choose here - this works right now but I wonder if there is something cleaner~ out there. I think that it would be cool to ditch usage of symbols cause they don't give us that much and they are breaking the compatibility with the `EventObject` constraint (after all a symbol doesn't extend a string). At the same time we don't want those events/signals to "clash" with whatever the user can define for their `TEvent`

packages/xstate-react/src/useMachine.ts
```diff
@@ -74,36 +81,19 @@ export function useMachine<TMachine extends AnyStateMachine>(
...
-
-      // Only change the current state if:
-      // - the incoming state is the "live" initial state (since it might have new actors)
-      // - OR the incoming state actually changed.
-      //
-      // The "live" initial state will have .changed === undefined.
-      const initialStateChanged =
-        nextState.changed === undefined &&
-        (Object.keys(nextState.children).length > 0 ||
-          typeof prevState.changed === 'boolean');
```
Andarist, line 100: We no longer need this because actors are now lazy and we are not recomputing initial state when starting etc

packages/core/src/ObservableActorRef.ts
davidkpiano: Awesome!

packages/core/src/Mailbox.ts
davidkpiano: In the future, we can model this as a priority queue, with signals having the highest priority.

This will also help in situations where a mailbox has a lot of events in queue but it needs to be shut down ASAP. The "stop signal" should take priority.

packages/xstate-react/src/useActor.ts
```diff
@@ -55,9 +23,8 @@ export function useActor(
     [actorRef]
   );
 
-  const boundGetSnapshot = useCallback(() => getSnapshot(actorRef), [
-    actorRef,
-    getSnapshot
+  const boundGetSnapshot = useCallback(() => actorRef.getSnapshot(), [
```
davidkpiano, line 26: Would it be better to pre-bind `.getSnapshot()`?

packages/xstate-react/src/useActor.ts
```diff
@@ -66,38 +33,10 @@ export function useActor(
...
-    actorRefRef.current = actorRef;
-
-    // Dequeue deferred events from the previous deferred actorRef
-    while (deferredEventsRef.current.length > 0) {
-      const deferredEvent = deferredEventsRef.current.shift()!;
-
-      actorRef.send(deferredEvent);
-    }
-  }, [actorRef]);
+  const boundSend: typeof actorRef.send = useCallback(
```
davidkpiano, line 36: Would it be better to pre-bind `.send()`?

packages/core/src/interpreter.ts
```diff
@@ -400,6 +400,9 @@ export class Interpreter<
 
     this._state = resolvedState;
 
+    // TODO: this notifies all subscribers but usually this is redundant
+    // if we are using the initialState as `resolvedState` then there is no real change happening here
+    // we need to rethink if this needs to be refactored
```
davidkpiano, line 405: So is it only useful for when the `resolvedState` is different from the initial state?

Makes me wonder if `initialState` is being passed in too late, because it doesn't make sense for an observer to get:

1. The initial state of the machine
2. Immediately after, an arbitrary restored state

Ideally, it should never get 1. So maybe we should think about revising the API:

```js
interpret(machine, { state: restoredState })
  .start(); // never pass in restored state here
```

packages/core/src/actors.ts
davidkpiano: In the future, we can improve the `Behavior` interface:

```ts
interface Behavior {
  transition: (snapshot, event, ctx) => snapshot;
  onStart?: () => snapshot;
  onStop?: () => snapshot;
  // ...
}
```

And put the start and stop behavior in separate callbacks, similar to `Behaviors.setup(fn)` in Akka (see example here: https://doc.akka.io/docs/akka/current/typed/actors.html)

davidkpiano (approved):
Some comments/nits but in general this LGTM

My only concern is about the strict-mode rerenders - it seems odd that it is 3 or 4 vs. 1, but we can address that separately.

packages/core/src/Mailbox.ts
Andarist: > This will also help in situations where a mailbox has a lot of events in queue but it needs to be shut down ASAP. The "stop signal" should take priority.

This particular case is currently handled like this:
https://github.com/statelyai/xstate/blob/76d738f0f8df6f6782a9b6d9b9bd565b4f6e1627/packages/core/src/Mailbox.ts#L18-L25

and I feel like it's a good solution here as we want to drop all queued items when receiving a stop signal anyway.

The idea of priorities etc is definitely intriguing but I think other use cases would have to back up its implementation (and if we'd implement such a thing then stop signal maybe could be handled using this mechanism).

packages/core/src/actors.ts
Andarist: Ye, those could be perhaps handled like this - but the described problem here wouldn't totally go away with such an approach. Here, we are also handling some other, "internal", events - like `snapshotEventType`. This doesn't quite fit the idea of signals, it's not related to lifecycle etc.

In theory, we could wrap regular events (`TEvent`) in another custom event - call `transition` with it, handle that in the switch case, and "forward" the unpacked even in the dedicated case. This was the type of this `behavior` would become smth like `Behavior<LifecycleEvent | typeof snapshotEventType | EventWrapper<TEvent>, StateFrom<TMachine>>`. 

It would help us out with the types and we wouldn't be able to end up with any namespace clashes. OTOH it feels like juggling stuff a little bit too much, OYAH programming is about juggling stuff :sweat_smile:

packages/xstate-react/src/useActor.ts
Andarist: If we start assuming that `actorRef` has been created by XState then we could consider that. There was no such assumption up until now though. In theory, the only thing that we require here is for the interface to be satisfied and that doesn't guarantee us if the `this` is being used in a method or not.

I would be fine with requiring "XState actors" here though - people using it with other objects of the same interface is rather a far-fetched scenario.

PR #3327, opened 2022-05-24
Removed states, events from testmodel, removed cases from events

mattpocock:
We decided that having `states` and `events` in the test model means you have closure issues.

However, having `cases` only run on `paths` mean they were excluded from path generation.

### DIFF
--- a/packages/xstate-test/test/index.test.ts
+++ b/packages/xstate-test/test/index.test.ts
@@ -59,12 +59,12 @@ describe('events', () => {
     });
 
     const testModel = createTestModel(feedbackMachine, {
-      events: {
-        SUBMIT: { cases: [{ value: 'something' }, { value: '' }] }
+      eventCases: {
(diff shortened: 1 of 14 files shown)


packages/xstate-test/test/index.test.ts
```diff
@@ -59,12 +59,12 @@ describe('events', () => {
...
     const testModel = createTestModel(feedbackMachine, {
-      events: {
-        SUBMIT: { cases: [{ value: 'something' }, { value: '' }] }
+      eventCases: {
+        SUBMIT: [{ value: 'something' }, { value: '' }]
       }
     });
 
-    await testUtils.testModel(testModel);
+    await testUtils.testModel(testModel, {});
```
davidkpiano, line 67: Can this be optional?
```suggestion
    await testUtils.testModel(testModel);
```

davidkpiano (approved):

PR #3328, opened 2022-05-25
Adjust some types in `@xstate/graph` to be more strict

Andarist:


packages/xstate-graph/src/graph.ts
```diff
@@ -181,8 +181,8 @@ const defaultMachineStateOptions: TraversalOptions<State<any, any>, any> = {
 
 export function getShortestPlans<TMachine extends AnyStateMachine>(
   machine: TMachine,
-  options?: Partial<TraversalOptions<AnyState, EventObject>>
```
Andarist, line 184: got rid of anys from here 😜

packages/xstate-graph/src/types.ts
```diff
@@ -162,7 +162,10 @@ export interface TraversalOptions<TState, TEvent extends EventObject>
     vctx: VisitedContext<TState, TEvent>
   ) => boolean;
   filter?: (state: TState, event: TEvent) => boolean;
-  getEvents?: (state: TState, cases: EventCaseMap<TState, TEvent>) => TEvent[];
+  getEvents?: (
+    state: TState,
+    cases: EventCaseMap<TState, TEvent>
+  ) => ReadonlyArray<TEvent>;
```
Andarist, line 168: using `ReadonlyArray` is often better than using regular arrays in types because it accepts both readonly arrays and mutable arrays. In our context it's super convenient to use `[] as const` to narrow contained strings etc to literals and that produces readonly arrays

packages/xstate-graph/src/types.ts
```diff
@@ -173,9 +176,9 @@ export interface TraversalOptions<TState, TEvent extends EventObject>
 }
 
 export type EventCaseMap<TState, TEvent extends EventObject> = {
-  [TEventType in TEvent['type']]?:
-    | ((state: TState) => Array<EventCase<TEvent>>)
-    | Array<EventCase<TEvent>>;
+  [E in TEvent as E['type']]?:
```
Andarist, line 179: It requires TS 4.1 but I feel that for `@xstate/graph`/`@xstate/test` it's already OK - this version correctly "narrows down" the `TEvent` type per the property key in this mapped type

mattpocock (approved):

PR #3336, opened 2022-05-25
Add transition coverage by default

davidkpiano:
CleanShot 2022-05-25 at 07 36 45@2x


packages/xstate-graph/src/graph.ts
```diff
@@ -366,7 +366,7 @@ export function getPathFromEvents<
     defaultMachineStateOptions as any
   );
 
-  const { serializeState } = optionsWithDefaults;
```
Andarist, line 369: does it shadow a variable of the same name from the outer scope?

davidkpiano: Will double check

PR #3340, opened 2022-05-25
Make `Interpreter#nextState` pure again

Andarist:


packages/core/src/interpreter.ts
```diff
@@ -418,9 +406,31 @@ export class Interpreter<
     // TODO: handle errors
     this.forward(event);
 
+    let errored = false;
```
Andarist, line 409: I know this ain't really beautiful. It's just a quick juggle of code to satisfy current tests **and** make `nextState` pure (like it should be). 

This closely follows the logic hidden in those operations:
https://github.com/statelyai/xstate/blob/dc3eba3a9b9db62a5a5ee2b1f5e878cc64f166fa/packages/core/src/behaviors.ts#L299-L308

It's just a stopgap before cleaning up the whole story around error handling.

davidkpiano (approved):

packages/core/src/interpreter.ts
davidkpiano: Makes sense 🙈 🧹

PR #3348, opened 2022-05-27
Added a failing test to cover multiple transitions with the same target but different sources

mattpocock:


davidkpiano:
The problem is that the transition serialization logic is only considering the event and next state, not the prev state + event + next state, so it thinks that it's already traversed the `A via GO_TO_A` transition.

I'll need to modify this logic.

PR #3359, opened 2022-05-30
`@xstate/graph` - Consider `prevState` for state serialization

davidkpiano:
Fixes #3348

### DIFF
--- a/packages/xstate-graph/src/graph.ts
+++ b/packages/xstate-graph/src/graph.ts
@@ -217,7 +217,11 @@ export function traverseShortestPlans<TState, TEvent extends EventObject>(
     ]
   >();
   const stateMap = new Map<SerializedState, TState>();
-  const initialSerializedState = serializeState(behavior.initialState, null);
+  const initialSerializedState = serializeState(
+    behavior.initialState,
+    undefined,
+    undefined
+  );
   stateMap.set(initialSerializedState, behavior.initialState);
 
   weightMap.set(initialSerializedState, [0, undefined, undefined]);
@@ -234,7 +238,12 @@ export function traverseShortestPlans<TState, TEvent extends EventObject>(
         const { state: nextState, event: eventObject } = adjacency[
           serializedState
         ].transitions[event];
-        const nextSerializedState = serializeState(nextState, eventObject);
+        const prevState = stateMap.get(serializedState);
+        const nextSerializedState = serializeState(
+          nextState,
+          eventObject,
+          prevState
+        );
         stateMap.set(nextSerializedState, nextState);
         if (!weightMap.has(nextSerializedState)) {
           weightMap.set(nextSerializedState, [
@@ -369,10 +378,13 @@ export function getPathFromEvents<
 
   const adjacency = performDepthFirstTraversal(behavior, optionsWithDefaults);
 
-  const stateMap = new Map<string, TState>();
+  const stateMap = new Map<SerializedState, TState>();
   const path: Steps<TState, TEvent> = [];
 
-  const initialStateSerial = serializeState(behavior.initialState, null);
+  const initialStateSerial = serializeState(
+    behavior.initialState,
+    undefined
+  ) as SerializedState;
   stateMap.set(initialStateSerial, behavior.initialState);
 
   let stateSerial = initialStateSerial;
@@ -393,8 +405,12 @@ export function getPathFromEvents<
         `Invalid transition from ${stateSerial} with ${eventSerial}`
       );
     }
-
-    const nextStateSerial = serializeState(nextState, event);
+    const prevState = stateMap.get(stateSerial);
+    const nextStateSerial = serializeState(
+      nextState,
+      event,
+      prevState
+    ) as SerializedState;
     stateMap.set(nextStateSerial, nextState);
 
     stateSerial = nextStateSerial;
@@ -435,19 +451,31 @@ export function performDepthFirstTraversal<TState, TEvent extends EventObject>(
   const adj: AdjacencyMap<TState, TEvent> = {};
 
   let iterations = 0;
-  const queue: Array<[TState, TEvent | null]> = [[initialState, null]];
+  const queue: Array<
+    [
+      nextState: TState,
+      event: TEvent | undefined,
+      prevState: TState | undefined
+    ]
+  > = [[initialState, undefined, undefined]];
+  const stateMap = new Map<SerializedState, TState>();
 
   while (queue.length) {
-    const [state, event] = queue.shift()!;
+    const [state, event, prevState] = queue.shift()!;
 
     if (iterations++ > limit) {
       throw new Error('Traversal limit exceeded');
     }
 
-    const serializedState = serializeState(state, event);
+    const serializedState = serializeState(
+      state,
+      event,
+      prevState
+    ) as SerializedState;
     if (adj[serializedState]) {
       continue;
     }
+    stateMap.set(serializedState, state);
 
     adj[serializedState] = {
       state,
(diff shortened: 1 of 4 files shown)


packages/xstate-test/src/machine.ts
```diff
@@ -64,7 +65,11 @@ function serializeMachineTransition(
     return '';
   }
 
-  return ` via ${serializeEvent(state.event)}`;
+  const prevStateString = prevState
+    ? `from ${simpleStringify(prevState.value)}`
```
davidkpiano, line 69: The solution starts here - we want to distinguish this transition from other transitions that may have the same target state and event by saying where this one came from

packages/xstate-test/src/machine.ts
```diff
@@ -104,10 +109,15 @@ export function createTestModel<TMachine extends AnyStateMachine>(
   const testModel = new TestModel<StateFrom<TMachine>, EventFrom<TMachine>>(
     machine as SimpleBehavior<any, any>,
     {
-      serializeState: (state, event) => {
-        return `${serializeState(state)}${serializeTransition(state, event, {
-          serializeEvent
-        })}` as SerializedState;
+      serializeState: (state, event, prevState) => {
```
davidkpiano, line 112: To do that, we need `serializeState` in `@xstate/graph` to also consider/provide the previous state (`prevState`).

Fun fact: we can't use `state.history` for this because that's actually deleted for optimization reasons, and `state.history` is deprecated in XState v5 anyway.

packages/xstate-graph/src/types.ts
```diff
@@ -135,7 +135,11 @@ export interface VisitedContext<TState, TEvent> {
 
 export interface SerializationOptions<TState, TEvent extends EventObject> {
   eventCases: EventCaseMap<TState, TEvent>;
-  serializeState: (state: TState, event: TEvent | null) => string;
+  serializeState: (
+    state: TState,
+    event: TEvent | undefined,
+    prevState?: TState
```
davidkpiano, line 141: So, we add `prevState` to the type for `serializeState` and see where the red squiggly lines are in `graph.ts`

packages/xstate-graph/src/graph.ts
```diff
@@ -234,7 +238,12 @@ export function traverseShortestPlans<TState, TEvent extends EventObject>(
         const { state: nextState, event: eventObject } = adjacency[
           serializedState
         ].transitions[event];
-        const nextSerializedState = serializeState(nextState, eventObject);
+        const prevState = stateMap.get(serializedState);
```
davidkpiano, line 241: Thankfully, the `stateMap` contains a mapping of serialized states to their actual `TState` values, so we can reference them.

In the future, this should be a built-in feature of the adjacency map object (i.e., let's make the object smarter and more self-contained)

packages/xstate-graph/src/graph.ts
```diff
@@ -435,19 +451,31 @@ export function performDepthFirstTraversal<TState, TEvent extends EventObject>(
   const adj: AdjacencyMap<TState, TEvent> = {};
 
   let iterations = 0;
-  const queue: Array<[TState, TEvent | null]> = [[initialState, null]];
+  const queue: Array<
+    [
```
davidkpiano, line 455: This named tuple is awkward; should probably be an object. Will be a future refactor.

packages/xstate-graph/src/graph.ts
```diff
@@ -516,11 +544,10 @@ export function traverseSimplePlans<TState, TEvent extends EventObject>(
   > = {};
 
   function util(
-    fromState: TState,
-    toStateSerial: SerializedState,
-    event: TEvent | null
```
davidkpiano, line 521: We no longer need the `event` here because we have the `fromStateSerial` directly. This is a bit of an optimization too: we no longer need to serialize the state because it's already serialized from the adjacency map.

packages/xstate-graph/src/graph.ts
```diff
@@ -369,10 +378,13 @@ export function getPathFromEvents<
...
 
-  const stateMap = new Map<string, TState>();
+  const stateMap = new Map<SerializedState, TState>();
   const path: Steps<TState, TEvent> = [];
 
-  const initialStateSerial = serializeState(behavior.initialState, null);
+  const initialStateSerial = serializeState(
+    behavior.initialState,
+    undefined
+  ) as SerializedState;
```
Andarist, line 387: To match the style at the very similar call site in this file:
```suggestion
  const initialSerializedState = serializeState(
    behavior.initialState,
    undefined,
    undefined
  ) as SerializedState;
```

packages/xstate-test/src/types.ts
```diff
@@ -125,7 +125,11 @@ export interface TestModelOptions<TState, TEvent extends EventObject>
     log: (msg: string) => void;
     error: (msg: string) => void;
   };
-  serializeTransition: (state: TState, event: TEvent | null) => string;
+  serializeTransition: (
+    state: TState,
+    event: TEvent | undefined,
+    prevState?: TState
```
Andarist, line 131: The mixed style of `| undefined` vs `?` in those serializers is a little bit weird. I can see some reasons why one could use something like this at times - but I don't understand this code well enough to know why this has been done like that here. Could you elaborate on why you have chosen this approach here?

packages/xstate-test/src/machine.ts
```diff
@@ -104,10 +109,15 @@ export function createTestModel<TMachine extends AnyStateMachine>(
...
+        return `${serializeState(state)}${serializeTransition(
```
Andarist, line 113: If I understand correctly this one intentionally doesn't call `serializeState` with `event` and `prevState`. Could we add a comment that would explain this here?

packages/xstate-graph/src/graph.ts
Andarist: I guess the style is just different at various places here - I personally would prefer a full name over a "serial" abbreviation. Especially since serial is a meaningful word and not just an abbreviation so while initially reading through this I was briefly confused about its meaning

Andarist:
For context - my comments are mostly nit-level, although I would appreciate answers to the last 2 comments. They would help me to understand the implementation of `@xstate/test` better.

Feel free to merge at any time.

packages/xstate-test/src/types.ts
davidkpiano: It used to be `| null`, should I change it back?

packages/xstate-graph/src/graph.ts
davidkpiano: The serialization is not just an abbreviation; it is both a string representation of the traversed node (not necessarily the state, need to rename in the future) and is the main mechanism for the algorithms knowing whether a node has been visited or not.

packages/xstate-test/src/types.ts
Andarist: Nah, either is fine - i just specifically wonder why `event` is a **required** (although it can be `undefined`) when the `prevState` is **optional**. Why the difference?

packages/xstate-test/src/machine.ts
```diff
@@ -104,10 +109,16 @@ export function createTestModel<TMachine extends AnyStateMachine>(
   const testModel = new TestModel<StateFrom<TMachine>, EventFrom<TMachine>>(
     machine as SimpleBehavior<any, any>,
     {
-      serializeState: (state, event) => {
-        return `${serializeState(state)}${serializeTransition(state, event, {
-          serializeEvent
-        })}` as SerializedState;
+      serializeState: (state, event, prevState) => {
+        // Only consider the `state` if `serializeTransition()` is opted out (empty string)
```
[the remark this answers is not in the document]
davidkpiano: The user can opt out of considering transitions by providing `serializeTransition: () => ''`

PR #3363, opened 2022-05-31
Added failing test for self-events

mattpocock:
Here, the self event should be called _while in another path_ - it shouldn't require its own path in order to be run.

### DIFF
--- a/.changeset/flat-trees-suffer.md
+++ b/.changeset/flat-trees-suffer.md
@@ -1,0 +1,9 @@
+---
+'@xstate/test': patch
+---
+
+author: @Andarist
+pr: #3367
+commit: b8cf8cfdc5d0e4c0b8f864d55a0f3126de3357f2
+
+Fixed `getShortestPathsTo` issue that caused candidate paths to be incorrectly removed while deduplicating generated paths.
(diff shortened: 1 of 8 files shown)


Andarist:
Does it matter though? Isn't this a similar problem to the one reported [here](https://github.com/statelyai/xstate/pull/3361)? I mean, we've ensured here that all transitions were actually taken.

mattpocock:
Similar, but this one seems easier to resolve.

Andarist:
@davidkpiano this PR now contains some commits that shouldn't quite be here - at least not displaying as something that has changed here. This makes navigating & reviewing a PR like this much harder. Could we clean this up somehow? I think this is mainly related to merge commits that are **not** on the `main` branch.

I think there is a good chance that you might fix this branch with:
```
git checkout matt/added-failing-test-for-self-events && git reset --hard f78b268546ad4ad68a2bbf465d3dfaaf9e6ff086 && git merge --no-ff main && git push --force
```

Note that after the merge step you might need to resolve conflicts.

davidkpiano:
@Andarist I fixed it 🔧

Andarist:
Thanks ❤️

davidkpiano:
Doing some cleanup - this can be easily done and fully customized in `serializeState` option by including the transition in the string.

PR #3367, opened 2022-05-31
Fixed `getShortestPathsTo` issue that caused candidate paths to be incorrectly removed while deduplicating generated paths

Andarist:


packages/xstate-test/src/TestModel.ts
```diff
@@ -102,7 +108,9 @@ export class TestModel<TState, TEvent extends EventObject> {
     let minWeight = Infinity;
     let shortestPaths: Array<TestPath<TState, TEvent>> = [];
 
-    const paths = this.filterPathsTo(statePredicate, this.getShortestPaths());
```
Andarist, line 105: The problem was here - paths were deduplicated before the filter (deduplication was happening in `this.getPaths` that was called by `this.getShortestPaths`). So it was possible for the deduplication algorithm/step to remove something that was supposed to be matched by the `statePredicate`

packages/xstate-test/src/TestModel.ts
```diff
@@ -102,7 +108,9 @@ export class TestModel<TState, TEvent extends EventObject> {
...
+    const paths = deduplicatePaths(
+      this.filterPathsTo(statePredicate, this._getStatePaths())
```
Andarist, line 112: `this._getStatePaths` piggybacks on `TestModel.defaults.pathGenerator` so this is still using shortest paths - but perhaps I should make this more explicit

packages/xstate-test/src/TestModel.ts
```diff
@@ -85,15 +85,21 @@ export class TestModel<TState, TEvent extends EventObject> {
...
-    const paths = pathGenerator(this.behavior, this.resolveOptions(options));
-
-    return paths.map(this.toTestPath);
+    return deduplicatePaths(
+      this._getStatePaths(options),
+      options?.serializeEvent
+    ).map(this.toTestPath);
   }
 
   public getShortestPathsTo(
```
Andarist, line 105: q: I somewhat don't get why `getShortestPathsTo` does its own filtering here while `getSimplePathsTo` is able to do leverage filtering implemented in `traverseSimplePathsTo`. The distinction seems to be a little bit odd.

davidkpiano (approved):

PR #3368, opened 2022-05-31
Remove redundant `while` loop from `@xstate/graph`

Andarist:


packages/xstate-graph/src/graph.ts
```diff
@@ -233,45 +233,43 @@ export function traverseShortestPlans<TState, TEvent extends EventObject>(
   const visited = new Set<SerializedState>();
 
   unvisited.add(initialSerializedState);
-  while (unvisited.size > 0) {
```
Andarist, line 236: I've only removed this `while` as it was redundant and everything got de-indented.

changes made to the `unvisited` within for-of loop are "seen" by this for-of loop so we won't miss any leftover items after exiting that inner loop

davidkpiano (approved):

PR #3372, opened 2022-05-31
Naive jest console message

### LINKED ISSUE #3369: Description key should appear in the Jest console.
### Description

In [this thread](https://discord.com/channels/795785288994652170/809564635614150686/980963741036335114) I explained what I expected the behavoir to be. David recommended I post this here. 

Suppose I have a straightforward model:

```js
const machine = createTestMachine({
  id: "commandPaletteMachineTest",
  initial: "closed",
  states: {
    closed: {
      on: {
        "control p": {
          target: "open",
        },
      },
    },
    open: {
      on: {
        "keypress Escape": {
          target: "closed",
        },
      },
    },
  },
});
```

Jest console renders something like this:
```sh
    ✓ Reaches state "#commandPaletteMachineTest.closed": control p → keypress Escape (89 ms)
```

This makes sense. I have "description" keys on each state so I'd expect if I update the previous to:

```js
   open: {
      on: {
        "keypress Escape": {
          description: "cool message for the console!"
          target: "closed",
        },
      },
    },
```

I would get something like:
```sh
    ✓ Reaches state "#commandPaletteMachineTest.closed": control p → keypress Escape "cool message for the console!" (89 ms)
```

### Expected result

The expected result is descriptions appear in the jest console.

### Actual result

There was no description in the jest console.

### Reproduction

none

### Additional context

_No response_

aisflat439:
closes #3369 

Figured I'd take a look at this over lunch. This seems like a reasonable starting point. It works for jest. I'm not sure what the output would be for something like Cypress. I know you support other testing frameworks and it's possible there's more to be thought about here. This is perhaps a fine naive solution.

https://github.com/statelyai/xstate/issues/3369

image

### DIFF
--- a/packages/xstate-test/src/utils.ts
+++ b/packages/xstate-test/src/utils.ts
@@ -78,7 +78,7 @@ export function formatPathTestResult(
 export function getDescription<T, TContext>(state: AnyState): string {
   const contextString =
     state.context === undefined ? '' : `(${JSON.stringify(state.context)})`;
-  const machineDescriptionMessage = state.transitions[0].description
+  const transitionDescription = state.transitions[0].description
     ? `"${state.transitions[0].description}"`
     : '';
 
@@ -102,7 +102,7 @@ export function getDescription<T, TContext>(state: AnyState): string {
   return (
     `state${stateStrings.length === 1 ? '' : 's'} ` +
     stateStrings.join(', ') +
-    ` ${machineDescriptionMessage} ${contextString}`
+    ` ${transitionDescription} ${contextString}`
   ).trim();
 }
 
--- a/packages/xstate-test/src/utils.ts
+++ b/packages/xstate-test/src/utils.ts
@@ -78,6 +78,9 @@ export function formatPathTestResult(
 export function getDescription<T, TContext>(state: AnyState): string {
   const contextString =
     state.context === undefined ? '' : `(${JSON.stringify(state.context)})`;
+  const jestConsoleMessage = state.transitions[0].description
+    ? `"${state.transitions[0].description}"`
+    : '';
 
   const stateStrings = state.configuration
     .filter((sn) => sn.type === 'atomic' || sn.type === 'final')
@@ -99,8 +102,8 @@ export function getDescription<T, TContext>(state: AnyState): string {
   return (
     `state${stateStrings.length === 1 ? '' : 's'} ` +
     stateStrings.join(', ') +
-    ` ${contextString}`.trim()
-  );
+    ` ${jestConsoleMessage} ${contextString}`
+  ).trim();
 }
 
 export function flatten<T>(array: Array<T | T[]>): T[] {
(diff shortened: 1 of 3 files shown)


packages/xstate-test/src/utils.ts
```diff
@@ -78,6 +78,9 @@ export function formatPathTestResult(
 export function getDescription<T, TContext>(state: AnyState): string {
   const contextString =
     state.context === undefined ? '' : `(${JSON.stringify(state.context)})`;
```
Andarist, line 80: Let's call this in a more generic way - this is not bound to jest nor to console. The whole function just produces some text description

aisflat439: That makes sense. I switched to `transitionDescription` as this would be the description on the transition taken by the event. Happy to change it if you have a more appropriate name.

davidkpiano (approved):

packages/xstate-test/src/utils.ts
```diff
@@ -99,8 +102,8 @@ export function getDescription<T, TContext>(state: AnyState): string {
   return (
     `state${stateStrings.length === 1 ? '' : 's'} ` +
     stateStrings.join(', ') +
-    ` ${contextString}`.trim()
-  );
+    ` ${transitionDescription} ${contextString}`
```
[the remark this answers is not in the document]
aisflat439: I see. Sure! I added a few test description test cases ot handle the `""` and the ` `. 

Let me know if you'd like any changes.

packages/xstate-test/src/utils.ts
```diff
@@ -78,6 +78,12 @@ export function formatPathTestResult(
 export function getDescription<T, TContext>(state: AnyState): string {
   const contextString =
     state.context === undefined ? '' : `(${JSON.stringify(state.context)})`;
+  const transitionDescription = state.transitions[0].description
+    ? state.transitions[0].description.trim()
+    : '';
```
Andarist, line 83: I wonder if this is actually correct.

`getDescription` is actually kinda like `getStateDescription` and you are mixing descriptions of a transition with it. The problem I'm seeing here is that you will only ever print a description of **the very last** transition and that you will only print a description for a single transition but in reality, multiple transitions can be available on that `state` that you have access to here.

The first problem is especially important here because this description won't ever be printed anywhere and it seems off to me that only the last transition is ever taken into consideration:
```js
      const machine = createTestMachine({
        initial: "a",
        states: {
          a: {
            on: {
              EVENT: { description: description, target: "b" },
            },
          },
          b: {
            on: {
              EVENT: { target: "c" },
            },
          },
          c: {},
        },
      });
```

aisflat439: I'm not sure I follow what you're saying. I think what you are saying is

`getDescription` generating a description from the final state only. So in order to do this, what we'll need is to evaluate if this event has been called along the path to get here. As such, what we really want is something more like

```
// generate the test messge
// check if that path used our event
// if it did, append any events with a description onto the test message
```

That's a lot more than I considered 🤣 . I can try to look through this code and see if there's a way I can do that.

Issue #3353, opened 2022-05-29
v5 - Bug: Same invoke can't happen twice in a row


Valkendorm:
### Description

Making the same invoke twice (after state transitions), without any other invokes in-between, will see the machine stuck on the state of the 2nd occurrence.

> Not sure if or how it is related, but another way to make sure the machine gets stuck is to give the same id to two distinct invokes.

The example below is quite elaborate, but it's just to simulate how the machine will never make it through the `promise` state's invoke when it's not transitioning to the `promisePause` state, which has another invoke.

```js
import { assign, createMachine } from "xstate";
import { fromPromise } from "xstate/actors";
import { not } from "xstate/guards";

export const promiseMachine = createMachine(
  {
    id: "promise",
    initial: "promise",
    context: {
      promisedTimes: 0,
    },
    states: {
      promise: {
        invoke: {
          src: "doPromise",
          onDone: [
            {
             // Changing the target to "promisePause" here will make the machine to always behave properly
              target: "logging",
              actions: "incrementPromisedTimes",
              guard: not("hasPromisedEnough"),
            },
            { target: "hasPromisedEnough" },
          ],
        },
      },
      logging: {
        entry: "logPromisedTimes",
        always: [
          {
            target: "promisePause",
            guard: "hasPromiseOnce",
          },
          { target: "promise" },
        ],
      },
      promisePause: {
        invoke: {
          src: "noPauseAllowed",
          onDone: "promise",
        },
      },
      hasPromisedEnough: {},
    },
  },
  {
    actions: {
      incrementPromisedTimes: assign({
        promisedTimes: (ctx) => ctx.promisedTimes + 1,
      }),
      logPromisedTimes: (ctx) =>
        console.log(`Promises made: ${ctx.promisedTimes}`),
    },
    actors: {
      doPromise: (ctx, event) =>
        fromPromise(() => new Promise((resolve) => setTimeout(resolve, 1000))),
      noPauseAllowed: (ctx, event) =>
        fromPromise(() => new Promise((resolve) => resolve())),
    },
    guards: {
      hasPromiseOnce: (ctx) => ctx.promisedTimes === 1,
      hasPromisedEnough: (ctx) => ctx.promisedTimes >= 3,
    },
  }
);
```

### Expected result

Machine going through subsequent identical invokes.

### Actual result

Machine stuck on the 2nd immediate occurrence of the same invoke.

### Reproduction

[...]

PR #3374, opened 2022-06-01
Fixed an issue with actors not being reinstantiated correctly in a microstep

Andarist:
fixes https://github.com/statelyai/xstate/issues/3353

### DIFF
--- a/packages/core/src/actions/invoke.ts
+++ b/packages/core/src/actions/invoke.ts
@@ -62,10 +62,7 @@ export function invoke<
         type,
         params: {
           ...params,
-          id: params.id,
-          src: params.src,
-          ref: new ObservableActorRef(behavior, id),
-          meta
+          ref: new ObservableActorRef(behavior, id)
         }
       } as InvokeActionObject;
     }
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -632,7 +632,7 @@ export class Interpreter<
         [actionTypes.cancel]: (_ctx, _e, { action }) => {
           this.cancel((action as CancelActionObject).params.sendId);
         },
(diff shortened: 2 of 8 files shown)


packages/core/src/interpreter.ts
```diff
@@ -651,26 +651,16 @@ export class Interpreter<
...
           ref._parent = this; // TODO: fix
-          // If the actor will be stopped right after it's started
-          // (such as in transient states) don't bother starting the actor.
-          if (
-            state.actions.find((otherAction) => {
-              return (
-                otherAction.type === actionTypes.stop &&
-                (otherAction as StopActionObject).params.actor === id
-              );
-            })
```
Andarist, line 662: This was the main source of the problem because we had actions looking like this:
```
actions: [stop('a'), invoke('a')]
```

Stop action for `'a'` was simply found in the array and thus the actor was never reinstantiated - despite of the fact that the invoke was coming **after** it.

packages/core/src/actions/invoke.ts
```diff
@@ -62,10 +62,7 @@ export function invoke<
         type,
         params: {
           ...params,
-          id: params.id,
-          src: params.src,
-          ref: new ObservableActorRef(behavior, id),
-          meta
```
Andarist, line 68: all removed properties are already part of the `params` so they were "set" here twice

packages/core/src/interpreter.ts
```diff
@@ -651,26 +651,16 @@ export class Interpreter<
...
-          ) {
+          // If the actor didn't end up being in the state
+          // (eg. going through transient states could stop it) don't bother starting the actor.
+          if (!this.state.children[id]) {
```
Andarist, line 656: Since `state.children` are mutated in `setChildren` "in order" it should be enough to just check if the final result contains the given id or not.

PR #3377, opened 2022-06-02
Initial value as function

danielbrodin:
This PR makes it possible to set the initial state from a function. My use case for this is that I have a parallel machine where I want parts of it to be saved in localStorage, and then initiated with that value. There might be some better way to do this, but currently I am doing this

```js
{
  Volume: {
  initial: 'Initial',
    states: {
      Initial: {
        invoke: {
          src: () =>
            Promise.resolve(
              localStorage.getItem('volume-state') ?? 'Unmuted'
            ),
          onDone: [
            {
              cond: (c, e) => e.data === 'Muted',
              target: 'Muted',
            },
            { target: 'Unmuted' },
          ],
        },
      },
      Muted: {
        entry: () => localStorage.setItem('volume-state', 'Muted'),
        on: {
          Toggle: {
            target: 'Unmuted',
          },
        },
      },
      Unmuted: {
        entry: () => localStorage.setItem('volume-state', 'Unmuted'),
        on: {
          Toggle: {
            target: 'Muted',
          },
        },
      },
    },
  },
}
```

With this change I would instead be able to do this
```js
Volume: {
    initial: () => localStorage.getItem("volume-state") ?? "Unmuted",
    states: {
      Muted: {
        entry: () => localStorage.setItem("volume-state", "Muted"),
        on: {
          Toggle: {
            target: "Unmuted",
          },
        },
      },
      Unmuted: {
        entry: () => localStorage.setItem("volume-state", "Unmuted"),
        on: {
          Toggle: {
            target: "Muted",
          },
        },
      },
    },
  },
```

It makes it easier and cleaner to be able to restore parts of a parallel state, but also for saving the state without having to save the complete state tree and restore it.

### DIFF
--- a/.changeset/light-planes-sparkle.md
+++ b/.changeset/light-planes-sparkle.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+A machines initial state can now be set using a function
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -352,7 +352,9 @@ class StateNode<
       );
     }
 
-    this.initial = this.config.initial;
+    this.initial = isFunction(this.config.initial)
+      ? this.config.initial()
+      : this.config.initial;
 
     this.states = (this.config.states
       ? mapValues(
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -541,7 +541,10 @@ export interface StateNodeConfig<
   /**
    * The initial state node key.
    */
-  initial?: keyof TStateSchema['states'] | undefined;
+  initial?:
+    | keyof TStateSchema['states']
+    | (() => keyof TStateSchema['states'])
+    | undefined;
   /**
    * @deprecated
    */
--- a/packages/core/test/initial.test.ts
+++ b/packages/core/test/initial.test.ts
@@ -1,6 +1,6 @@
-import { Machine } from '../src';
+import { Machine, MachineConfig } from '../src';
 
-const config = {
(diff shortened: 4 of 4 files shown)


Andarist:
I like the idea and I wish we had support for smth like this. But I think that a better direction would be to allow using conditional transitions for `initial`. This could have better support for our tooling (studio, typegen, etc) as it's easier to statically analyze those. I would have to recheck but I think that this might even be already implemented on the `next` branch (or there is at least some groundwork laid out for it there)

davidkpiano:
> I like the idea and I wish we had support for smth like this. But I think that a better direction would be to allow using conditional transitions for `initial`. This could have better support for our tooling (studio, typegen, etc) as it's easier to statically analyze those. I would have to recheck but I think that this might even be already implemented on the `next` branch (or there is at least some groundwork laid out for it there)

Initial actions are, but not a conditional initial target. We could make an RFC for that idea because it's still viable, given that it is deterministic:

```js
initial: [
  { guard: 'something', target: 'first' },
  { target: 'second'
]
```

But with the original use-case, the more straightforward way to accomplish this is by persisting and restoring all state.

danielbrodin:
> the more straightforward way to accomplish this is by persisting and restoring all state.

Well not really, because we can't restore all state. All states except one has to start from scratch. Your idea however would probably solve this as well since you could have a guard with `localStorage.getItem('volume-state') === 'Muted'`.

Andarist:
I've created an internal ticket for v5 to work on this.

PR #3380, opened 2022-06-04
[xstate/test] Narrow down the `event` type passed to `EventExecutor` based on the key

Silverwolf90:
This narrows down the type of the `event` in the `EventExecutor` to the type that it's keyed by.

I think this is the intention of the code -- am I correct?

### DIFF
--- a/packages/xstate-test/src/types.ts
+++ b/packages/xstate-test/src/types.ts
@@ -86,7 +86,7 @@ export interface TestParam<TState, TEvent extends EventObject> {
     [key: string]: (state: TState) => void | Promise<void>;
   };
   events?: {
-    [TEventType in TEvent['type']]?: EventExecutor<TState, TEvent>;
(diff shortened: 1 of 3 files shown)


packages/xstate-test/src/types.ts
```diff
@@ -86,7 +86,7 @@ export interface TestParam<TState, TEvent extends EventObject> {
     [key: string]: (state: TState) => void | Promise<void>;
   };
   events?: {
-    [TEventType in TEvent['type']]?: EventExecutor<TState, TEvent>;
+    [TEventType in TEvent['type']]?: EventExecutor<TState, ExtractEvent<TEvent, TEventType>>;
```
Andarist, line 89: Could we add a test that would cover this change? We can fake type-level tests like here:
https://github.com/statelyai/xstate/blob/506686dbda27a6ee26ccda36eddcd89cb68fc2ab/packages/xstate-graph/test/types.test.ts

Silverwolf90: I've added a test -- is it adequate?

PR #3387, opened 2022-06-07
Add pure assignment warning

davidkpiano:
This PR adds a warning about pure assignments to the docs.

### DIFF
--- a/docs/guides/context.md
+++ b/docs/guides/context.md
@@ -189,6 +189,21 @@ The `meta` object contains:
 - `action` - the assign action
 
 ::: warning
+Assigners **must be pure**; they should not contain any side-effects.
+
+```js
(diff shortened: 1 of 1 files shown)


docs/guides/context.md
```diff
@@ -188,6 +188,21 @@ The `meta` object contains:
 - `state` - the current state in a normal transition (`undefined` for the initial state transition)
 - `action` - the assign action
 
+::: warning
```
Andarist, line 191: now we'll have two warnings stacked on top of each other, maybe worth combining them into one?

davidkpiano: Nah, they should be seen as separate warnings.

Andarist (approved):

PR #3393, opened 2022-06-08
Deprecate `send()` action creator

davidkpiano:
This PR deprecates the `send()` action creator in favor of `sendTo()` and `raise()`.


packages/core/src/actions.ts
```diff
@@ -292,7 +294,7 @@ export function sendTo<
   TEvent extends EventObject,
   TActor extends AnyActorRef
 >(
-  actor: string | TActor | ((ctx: TContext) => TActor),
+  actor: string | TActor | ((ctx: TContext, event: TEvent) => TActor),
```
Andarist, line 297: Should we add a test case that would check that `event` param can be used to get the `TActor`? I'm pretty sure that it just works but it would be good to have such test to validate the types

packages/core/src/actions.ts
```diff
@@ -181,6 +181,8 @@ export function resolveRaise<TEvent extends EventObject>(
  * Sends an event. This returns an action that will be read by an interpreter to
  * send the event in the next step, after the current step is finished executing.
  *
+ * @deprecated Use the `sendTo(...)` action creator instead.
```
Andarist, line 184: I don't think that we can deprecate this right now. `send` is still useful to create delayed events to "self".

davidkpiano: This is now possible

davidkpiano:
@Andarist Is this good to go?

packages/core/src/StateNode.ts
```diff
@@ -1307,11 +1307,16 @@ class StateNode<
...
+        const isSendAction = action.type === actionTypes.send;
+        const isRaisableSendAction =
+          isSendAction &&
           (action as SendActionObject<TContext, TEvent>).to ===
-            SpecialTargets.Internal)
+            SpecialTargets.Internal &&
+          !(action as SendActionObject<TContext, TEvent>).delay;
+
+        return action.type === actionTypes.raise || isRaisableSendAction;
+      }
```
Andarist, line 1319: nit: if this is a `raise` action then it's not worth checking upfront if it's a raisable send action
```ts
        if (action.type === actionTypes.raise) {
            return true;
        }
        if (action.type !== actionTypes.send) {
            return false;
        }
        const sendAction = action as SendActionObject<TContext, TEvent>;
        return sendAction.to === SpecialTargets.Internal || !sendAction.delay;
```

packages/core/src/StateNode.ts
```diff
@@ -1307,11 +1307,16 @@ class StateNode<
...
-        action.type === actionTypes.raise ||
-        (action.type === actionTypes.send &&
+        | SendActionObject<TContext, TEvent, TEvent> => {
+        const isSendAction = action.type === actionTypes.send;
+        const isRaisableSendAction =
+          isSendAction &&
           (action as SendActionObject<TContext, TEvent>).to ===
-            SpecialTargets.Internal)
+            SpecialTargets.Internal &&
+          !(action as SendActionObject<TContext, TEvent>).delay;
```
Andarist, line 1316: q: what should happen with `delay: 0`? 😉 is it raisable? or should it execute with `setTimeout(() => this.sendTo(event), 0)`?

packages/core/test/actions.test.ts
```diff
@@ -2314,6 +2315,72 @@ describe('sendTo', () => {
...
     interpret(parentMachine).start();
   });
+
+  it('should be able to send an event to itself', (done) => {
+    const machine = createMachine({
+      initial: 'a',
+      states: {
+        a: {
+          entry: sendTo(
+            SpecialTargets.Internal,
```
Andarist, line 2325: This pattern should get into the docs somewhere.

packages/core/src/StateNode.ts
davidkpiano: The latter; don't special-case it

packages/core/test/actions.test.ts
davidkpiano: This is only for tests; as discussed, we want to encourage raising over sending.

Andarist: IIRC we've discussed that `raise` should be able to raise delayed events. Is this already supported? I think I didn't see any tests like that.

mellson (approved):
It looks good to me, based on what little knowledge I have of this codebase so far 😊

docs/fr/guides/actions.md
```diff
@@ -193,6 +193,19 @@ entry: send({ type: 'SOME_EVENT' });
 
 ## Send action
 
+::: warning
```
Andarist, line 196: Those warnings should probably mention `raise` as alternative for sending to "self".

docs/guides/actions.md
```diff
@@ -214,6 +214,19 @@ entry: send({ type: 'SOME_EVENT' });
 
 ## Send action
 
+::: warning
```
Andarist, line 217: same as above

packages/core/src/actions.ts
```diff
@@ -157,11 +158,13 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
  * @param eventType The event to raise.
  */
 export function raise<TContext, TEvent extends EventObject>(
-  event: Event<TEvent>
-): RaiseAction<TEvent> | SendAction<TContext, AnyEventObject, TEvent> {
-  if (!isString(event)) {
-    return send(event, { to: SpecialTargets.Internal });
+  event: Event<TEvent>,
```
Andarist, line 161: This doesn't support expressions right now. I think that we should bring it to here - otherwise not all `send`s can be refactored to `raise`

packages/core/src/actions.ts
```diff
@@ -157,11 +158,13 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
...
+  options?: RaiseActionOptions<TContext, TEvent>
+): RaiseAction<TEvent> {
+  if (!isString(event) || options) {
+    return send(event, { ...options, to: SpecialTargets.Internal }) as any;
   }
```
Andarist, line 166: IMHO we should stop piggybacking on `send` here. It's especially important right now because you are suppressing the type checker to satisfy the declared return type. I wouldn't be surprised if we miss something that should be handled because of that, now or in the future. At times, we are checking `action.type` after all.

packages/core/test/types.test.ts
```diff
@@ -253,23 +253,23 @@ describe('Raise events', () => {
               {
                 actions: raise({
                   type: 'ALOHA'
-                }) as any /* TODO: FIX */,
+                }),
```
Andarist, line 256: I created a spike for type-related improvements around this here: https://github.com/statelyai/xstate/pull/3694

packages/core/test/actions.test.ts
```diff
@@ -2603,6 +2604,103 @@ describe('sendTo', () => {
...
+
+    const service = interpret(machine).start();
+
+    service.onDone(() => done());
+
+    // Ensures that the delayed self-event is sent when in the `b` state
+    service.send('TO_B');
+  });
+
+  it('should be able to send an event to itself', (done) => {
```
Andarist, line 2671: This test case has the very same title as the one above. This should be fixed as I'm not exactly sure what this test is testing here.

packages/core/test/actions.test.ts
```diff
@@ -2603,6 +2604,103 @@ describe('sendTo', () => {
...
+    });
+
+    const service = interpret(machine).start();
+
+    service.send({ type: 'EVENT', value: 'foo' });
+  });
+});
+
+describe('raise', () => {
+  it('should be able to send an event to itself', (done) => {
```
Andarist, line 2637:
```suggestion
  it('should be able to send a delayed event to itself', (done) => {
```

packages/core/test/actions.test.ts
```diff
@@ -2603,6 +2604,103 @@ describe('sendTo', () => {
...
+        }
+      }
+    });
+
+    const service = interpret(machine).start();
+
+    service.onDone(() => done());
+
+    // Ensures that the delayed self-event is sent when in the `b` state
+    service.send('TO_B');
```
Andarist, line 2668: This test isn't testing anything right now - it has 0 expectations ;p

packages/core/test/actions.test.ts
```diff
@@ -2603,6 +2604,103 @@ describe('sendTo', () => {
...
```
Andarist, line 2697: To avoid type errors when merging this into `next` we can do this:
```suggestion
    service.send({ type: 'TO_B' });
```

packages/core/test/actions.test.ts
Andarist, line 2668: To avoid type errors when merging this into `next` we can do this:
```suggestion
    service.send({ type: 'TO_B' });
```

packages/core/src/StateNode.ts
Andarist: it would be cool to have a test for that case

davidkpiano: Done here: [b583f05](https://github.com/statelyai/xstate/pull/3393/commits/b583f05b58076830041eb0d50eaaa88f9ca2dcab)

packages/core/test/actions.test.ts
davidkpiano: Yes this is done

packages/core/src/actions.ts
```diff
@@ -158,32 +159,57 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
...
+    _event: resolvedEvent,
+    delay: resolvedDelay
   } as any;
 }
 
 /**
  * Sends an event. This returns an action that will be read by an interpreter to
  * send the event in the next step, after the current step is finished executing.
  *
+ * @deprecated Use the `sendTo(...)` action creator instead.
```
Andarist, line 211:
```suggestion
 * @deprecated Use the `sendTo(...)` action creator when sending to a different actor or `raise(...)` when sending to the "self" actor.
```

Andarist (approved):

Issue #1251, opened 2020-06-19
Interpreter.nextState triggers spawn() calls


pipex:
**Description**

I saw this first using the XState visualizer. Here is some example gist for a ping pong machine with actors

https://xstate.js.org/viz/?gist=dc355bee0b368a5fc80e62141db4996d

<img width="601" alt="Screen Shot 2020-06-19 at 13 28 05" src="https://user-images.githubusercontent.com/1822826/85163386-bbf82e80-b230-11ea-9e55-c3b38988927e.png">

When passing the mouse over the `started` event, the visualizer calls the `Interpreter.nextState` method to hightlight the next state, however this results in the spawning of the child machine (it actually gets triggered multiple times because of the visualizer using `onMouseOver` instead of `onMouseEnter` but that is a different bug)

This can be seen when actually triggering the event, the state visualization shows multiple children, but only one event has been sent. 

<img width="285" alt="Screen Shot 2020-06-19 at 13 30 39" src="https://user-images.githubusercontent.com/1822826/85163868-8273f300-b231-11ea-97d5-58aad18f3bdd.png">

I can replicate the same behavior using an interpreter in code, if I set the ping machine to log on entry I can see the log message appearing on the console when running the following

```
const pingService = interpret(ping)
	.onTransition((state) => console.log(state.value))
	.start();
pingService.nextState('start');
```

**Expected Result**

A single child should get spawned when the following code is run (see linked gist for ping machine definition)

```
const pingService = interpret(ping)
	.onTransition((state) => console.log(state.value))
	.start();
console.log(pingService.nextState('start'));
pingService.send('next');
```

**Actual Result**

The `nextState` method triggers the `spawn` call and two children are spawned

**Reproduction**

https://xstate.js.org/viz/?gist=dc355bee0b368a5fc80e62141db4996d

**Additional context**

Happens with visualizer and also in v4.10.0

Andarist:
Hm, that's weird as `nextState` is just using `machine.transition` which is a pure API. Could you prepare a CodeSandbox representing a problem? I have a suspicion that this might be just a visualizer bug, not a runtime one.

pipex:
Hi @Andarist. Here is a codesandbox replicating the behavior. As you can see, the service in the end has 2 children where I would expect it to have only one. It may be possible that I'm misundertanding the purpose of the `nextState` method though.

https://codesandbox.io/s/wizardly-goodall-3g7no?file=/src/index.js

davidkpiano:
[...]

PR #3400, opened 2022-06-10
[v5] Add test to ensure `.nextState(...)` does not start spawned actors

davidkpiano:
Addresses #1251


davidkpiano:
Merging since this is just adding a useful passing test

PR #3401, opened 2022-06-14
[@xstate/react] Add `createHooks(...)`

davidkpiano:
For more info, see the RFC: https://github.com/statelyai/rfcs/pull/8

This is a bit of a revival of @mattpocock's PR: https://github.com/statelyai/xstate/pull/2783

### DIFF
--- a/.changeset/fifty-parrots-live.md
+++ b/.changeset/fifty-parrots-live.md
@@ -1,0 +1,31 @@
+---
+'@xstate/react': minor
+---
+
+The `createHooks(actorRef)` function is introduced, which creates useful global hooks for actors:
+
+```jsx
+import { createHooks } from '@xstate/react';
+import { authMachine } from './authMachine';
+import { interpret } from 'xstate';
+
+// Interpret the machine first (full interpretation control)
+const authService = interpret(authMachine).start();
+
+// Create a "hooks object" from the actor
+const auth = createHooks(authService);
+
+// ...
+
+const SomeForm = () => {
+  // Get full state snapshot and send() function
+  const [state, send] = auth.useActor();
+
+  // Get selected state
+  const isLoggedIn = auth.useSelector((state) => state.hasTag('authorized'));
+
+  // ...
+};
+```
+
+For more info, see the RFC: https://github.com/statelyai/rfcs/pull/8
--- a/packages/xstate-react/src/createHooks.ts
+++ b/packages/xstate-react/src/createHooks.ts
@@ -1,0 +1,104 @@
+import { createContext, useContext } from 'react';
+import { AnyActorRef, EmittedFrom } from 'xstate';
+import { useActor } from './useActor';
+import { useSelector } from './useSelector';
+
+/**
+ * Creates and returns React hooks that can be used in any component:
+ * - `obj.useActor()` - hook returns the same tuple as `useActor(actorRef)`
+ * - `obj.useSelector(selector, compare?)` - hook returns the same selected value as `useSelector(actorRef, selector, compare)`
+ *
+ * It also returns two useful values:
+ * - `obj.actorRef` - the original `actorRef`
+ * - `obj.Provider` - can be used to provide a different `actorRef` value
+ *
+ * @example
+ *
+ * ```jsx
+ * const authService = interpret(authMachine).start();
+ *
+ * const auth = createHooks(authService);
+ *
+ * // ...
+ *
+ * const Component = () => {
+ *   const [state, send] = auth.useActor();
+ *
+ *   const isLoggedIn = auth.useSelector(state => state.hasTag('authenticated'));
+ *
+ *   // ...
+ * }
+ * ```
+ * @param actorRef The actor ref (such as an invoked machine) to create hooks from
+ * @returns an object containing:
+ * - `.useActor()`
+ * - `.useSelector(selector, compare?)`
+ * - `.actorRef`
+ * - `.Provider`
+ */
+export function createHooks<TActorRef extends AnyActorRef>(
+  actorRef: TActorRef
+): {
+  actorRef: TActorRef;
+  /**
+   * The React context provider component, used for overriding the default `actorRef`.
+   *
+   * @example
+   * ```jsx
+   * const App = () => {
+   *   return <obj.Provider value={someOtherActorRef}>
+   *     // ...
+   *   </obj.Provider>
+   * };
+   */
+  Provider: React.Provider<TActorRef>;
+  /**
+   * Returns a tuple of `[snapshot, send]`:
+   * - The `snapshot` is the latest observed value from the `actorRef`
+   * - The `send(event)` function sends an event to the `actorRef`
+   *
+   * @example
+   * ```js
+   * const [state, send] = obj.useActor();
+   *
+   * return <button onClick={() => send({ type: 'EVENT' })} />
+   * ```
+   */
+  useActor(): [EmittedFrom<TActorRef>, TActorRef['send']];
+  /**
+   * Selects a value from the `actorRef` using the `selector` and an optional `comparator`
+   *
+   * @example
+   * ```js
+   * const values = obj.useSelector(state => {
+   *   return state.context.values;
+   * }, (a, b) => shallowCompare(a, b));
+   *
+   * const count = obj.useSelector(state => {
+   *   return state.context.count;
+   * });
+   * ```
+   */
+  useSelector: <T>(
+    selector: (snapshot: EmittedFrom<TActorRef>) => T,
+    comparator?: ((a: T, b: T) => boolean) | undefined
+  ) => T;
+} {
+  const actorContext = createContext(actorRef);
+
+  return {
+    actorRef,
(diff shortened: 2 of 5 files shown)


packages/xstate-react/src/createHooks.ts
```diff
@@ -0,0 +1,104 @@
...
+   *   return state.context.count;
+   * });
+   * ```
+   */
+  useSelector: <T>(
+    selector: (snapshot: EmittedFrom<TActorRef>) => T,
+    comparator?: ((a: T, b: T) => boolean) | undefined
+  ) => T;
+} {
+  const actorContext = createContext(actorRef);
```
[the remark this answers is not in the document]
farskid: Should this be a comment in the code too? Just to make sure we don't miss it later?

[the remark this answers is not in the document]
davidkpiano: It's here: https://github.com/statelyai/xstate/pull/3401/files/25dc187ad22b1a413359fab01b998fc779905889#diff-db64295cf7356bf0072dc45c279995dbe31b4e6b68cbaaaf4b9d681491e50028R44-R53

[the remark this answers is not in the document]
davidkpiano: [0155048](https://github.com/statelyai/xstate/pull/3401/commits/01550489310369b895a43226648f07a5c8e5b10e)

packages/xstate-react/src/createHooks.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @param actorRef The actor ref (such as an invoked machine) to create hooks from
+ * @returns an object containing:
+ * - `.useActor()`
+ * - `.useSelector(selector, compare?)`
+ * - `.actorRef`
+ * - `.Provider`
+ */
+export function createHooks<TActorRef extends AnyActorRef>(
+  actorRef: TActorRef
+): {
```
Andarist, line 41:
```suggestion
  actorRef: TActorRef,
  displayName?: string
): {
```

packages/xstate-react/src/createHooks.ts
```diff
@@ -0,0 +1,106 @@
...
+   *   return state.context.count;
+   * });
+   * ```
+   */
+  useSelector: <T>(
+    selector: (snapshot: EmittedFrom<TActorRef>) => T,
+    comparator?: ((a: T, b: T) => boolean) | undefined
+  ) => T;
+} {
+  const actorContext = createContext(actorRef);
```
Andarist, line 87:
```suggestion
  const actorContext = createContext(actorRef);
  if (displayName) {
    actorContext.displayName = displayName;
  }
```

packages/xstate-react/src/createHooks.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  return {
+    actorRef,
+    Provider: actorContext.Provider,
```
Andarist, line 91: bikeshedding - `createHooks` somewhat implies that I will only get hooks back but in here we are also getting back a `Provider` and an `actorRef`

packages/xstate-react/src/createHooks.ts
```diff
@@ -0,0 +1,106 @@
+import { createContext, useContext } from 'react';
+import { AnyActorRef, EmittedFrom } from 'xstate';
+import { useActor } from './useActor';
+import { useSelector } from './useSelector';
+
+/**
+ * Creates and returns React hooks that can be used in any component:
+ * - `obj.useActor()` - hook returns the same tuple as `useActor(actorRef)`
+ * - `obj.useSelector(selector, compare?)` - hook returns the same selected value as `useSelector(actorRef, selector, compare)`
```
Andarist, line 9: nit: in this file, you are calling the same argument both `compare` and `comparator`

packages/xstate-react/src/createHooks.ts
```diff
@@ -0,0 +1,106 @@
+import { createContext, useContext } from 'react';
+import { AnyActorRef, EmittedFrom } from 'xstate';
+import { useActor } from './useActor';
+import { useSelector } from './useSelector';
+
+/**
```
Andarist, line 6: This gets pretty lengthy and reiterates over what particular functions do - I wonder if it wouldn't be better to keep this succinct, only list what is being returned or smth and leave the rest for the descriptions of those functions that are being  returned

packages/xstate-react/test/createHooks.test.tsx
```diff
@@ -0,0 +1,177 @@
+import * as React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import { createMachine, interpret } from 'xstate';
+import { createHooks } from '../src/createHooks';
+
+describe('createHooks', () => {
```
Andarist, line 6: We could use `describeEachReactMode` here

.changeset/fifty-parrots-live.md
```diff
@@ -0,0 +1,31 @@
...
+  const [state, send] = auth.useActor();
+
+  // Get selected state
+  const isLoggedIn = auth.useSelector((state) => state.hasTag('authorized'));
+
+  // ...
+};
+```
+
+For more info, see the RFC: https://github.com/statelyai/rfcs/pull/8
```
Andarist, line 31:
```suggestion
For more info, see [the RFC](https://github.com/statelyai/rfcs/pull/8).
```

packages/xstate-react/src/createHooks.ts
```diff
@@ -0,0 +1,106 @@
...
+  ) => T;
+} {
+  const actorContext = createContext(actorRef);
+
+  return {
+    actorRef,
+    Provider: actorContext.Provider,
+    useActor: () => {
+      // Note: all methods should make use of the `actorRef` from the `actorContext`,
+      // not the initial one passed in.
```
Andarist, line 94: This feels more specific to the `createContext` call than to this line here. The same is true for `useSelector` but the comment is not repeated there.

packages/xstate-react/src/createHooks.ts
```diff
@@ -0,0 +1,106 @@
...
+      const actorRefFromContext = useContext(actorContext);
+      return useActor(actorRefFromContext);
+    },
+    useSelector: <T>(
```
Andarist, line 98: @mattpocock also implemented `createUseSelector` which was kinda nice

.changeset/fifty-parrots-live.md
```diff
@@ -0,0 +1,31 @@
+---
```
Andarist, line 1: I think that a must-have before merging this is to have some kind of SSR guidance.

packages/xstate-react/src/createHooks.ts
mattpocock: Was it `createUseSelector` or `createSelector`? Both are nice.

mattpocock:
## `useInterpret` in the root of your app

Having `createHooks` work from an already running service - `createHooks(service)` - kills a common use case:

1. Have a `useInterpret` at the root of your app
2. Add a context provider at the root, passing in the service

There are plenty of valid cases for this - especially since having a singleton outside React's component closure can often be painful when interacting with things inside the closure.

Another case that comes to mind is when you want a 'semi-global' service - i.e. a service that manages the context over a large portion of the app, but not the whole of it.

## Side effects in modules means they're not testable

Having a side-effect in a module also means that everything is less testable, especially when the `createHooks` relies on that running service in order to work. Imagine this case:

```ts
// myHooks.ts
const service = interpret(myMachine).start();

export const hooks = createHooks(service);

// myHooks.test.ts
import { hooks } from './myHooks';

// The service will actually be running in this file!
```

There's no way to access the hooks without starting the service. This means that the desired use case of 'overriding the singleton by wrapping components in a context provider' won't work.

## Solution

Instead, `createHooks` should work based off a machine definition, not a running service. This solves both problems, but will require some API changes.

Andarist:
> There are plenty of valid cases for this - especially since having a singleton outside React's component closure can often be painful when interacting with things inside the closure.

I assume that this feature should automate the common setup and stuff - so that would mean that the created Provider~ should accept `implementations` somehow as it would be responsible for calling `useInterpret` under the hood. Is this what you meant?

> Another case that comes to mind is when you want a 'semi-global' service - i.e. a service that manages the context over a large portion of the app, but not the whole of it.

I think that the current shape of this PR doesn't make this particular use case harder. You can always just apply the Provider around a portion of your app 🤔 

> There's no way to access the hooks without starting the service. This means that the desired use case of 'overriding the singleton by wrapping components in a context provider' won't work.

👍

Andarist:
When reconsidering after @mattpocock's comment - the ease of use of "global imports" seems to be a little bit overrated to me. The downsides of not having that are:
- a provider has to be put somewhere
- a different import has to be used and the returned variable (actor ref) has to be used to send an event or something (you can't directly use what you import)
- It becomes a little bit harder to interact with the code outside of React

It feels though that even the code that is not related directly to React is usually instantiated at the top level of the React app. This automatically solves the raised testing and SSR concerns and is, overall, close to React's philosophy of doing things.

PR #3419, opened 2022-06-23
[@xstate/react/useSpawn @xstate/vue/useSpawn] feat: spawn with option in useSpawn

cy-98:
Spawn behavior with options. 
Can find this actor with id.

```tsx
    const Test = () => {
      const actor = useSpawn(bahavior, {
        id: 'optionID'
      });

      return (
        <>
          <div data-testid="actor-id">{actor.id}</div>
        </>
      );
    };
```

### DIFF
--- a/packages/xstate-react/src/useSpawn.ts
+++ b/packages/xstate-react/src/useSpawn.ts
@@ -1,5 +1,10 @@
-import { ActorRef, Behavior, EventObject } from 'xstate';
-import { spawnBehavior } from 'xstate';
+import {
+  ActorRef,
+  Behavior,
+  EventObject,
+  spawnBehavior,
+  SpawnBehaviorOptions
+} from 'xstate';
 import useConstant from './useConstant';
 
 /**
@@ -10,10 +15,11 @@ import useConstant from './useConstant';
  * @returns An ActorRef with the specified `behavior`
  */
 export function useSpawn<TState, TEvent extends EventObject>(
-  behavior: Behavior<TEvent, TState>
+  behavior: Behavior<TEvent, TState>,
+  options?: SpawnBehaviorOptions
 ): ActorRef<TEvent, TState> {
   const actorRef = useConstant(() => {
-    return spawnBehavior(behavior);
+    return spawnBehavior(behavior, options);
   });
 
   return actorRef;
--- a/packages/core/src/behaviors.ts
+++ b/packages/core/src/behaviors.ts
@@ -100,7 +100,7 @@ export function fromPromise<T>(
   };
 }
 
-interface SpawnBehaviorOptions {
+export interface SpawnBehaviorOptions {
   id?: string;
   parent?: ActorRef<any>;
 }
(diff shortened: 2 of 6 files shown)


cy-98:
> Out of curiosity - could you describe what kind of use cases does `useSpawn` satisfy for you?

I have a machine which forward message to other actor. Some of actors are very simple spawn form simmple behavior.

Andarist:
Could you prepare a simple codesandbox demonstrating the pattern? I'm wondering how exactly you have wired things up. It's mostly interesting from my point of view cause `useSpawn` is sort of starting/managing/stopping an actor and I wonder why do you make this React's concern here and not let XState manage this using regular `spawn`.

cy-98:
> Could you prepare a simple codesandbox demonstrating the pattern? I'm wondering how exactly you have wired things up. It's mostly interesting from my point of view cause `useSpawn` is sort of starting/managing/stopping an actor and I wonder why do you make this React's concern here and not let XState manage this using regular `spawn`.

codesandbox: https://codesandbox.io/s/elegant-hodgkin-19vx3q?file=/src/index.tsx
you also can check this repo: https://github.com/cy-98/xstate-full-module

In this case, I want find a actor with special ID.

cy-98:
> > Could you prepare a simple codesandbox demonstrating the pattern? I'm wondering how exactly you have wired things up. It's mostly interesting from my point of view cause `useSpawn` is sort of starting/managing/stopping an actor and I wonder why do you make this React's concern here and not let XState manage this using regular `spawn`.
> 
> codesandbox: https://codesandbox.io/s/elegant-hodgkin-19vx3q?file=/src/index.tsx you also can check this repo: https://github.com/cy-98/xstate-full-module
> 
> In this case, I want find a actor with special ID.

Updated.
Actually I want make a actors system that, remote components can registry an actor and other find it easy. A little bit like RPC

cy-98:
@Andarist Is it can be merged ?

packages/xstate-react/src/useSpawn.ts
```diff
@@ -10,10 +15,11 @@ import useConstant from './useConstant';
  * @returns An ActorRef with the specified `behavior`
  */
 export function useSpawn<TState, TEvent extends EventObject>(
-  behavior: Behavior<TEvent, TState>
+  behavior: Behavior<TEvent, TState>,
+  options?: SpawnBehaviorOptions
```
Andarist, line 19: Currently, this allows `parent` to be passed as part of the `options` but I think that this shouldn't be allowed. Thoughts?

cy-98: Got it.
I disagree. `Spawn` has it own interface, but disabled by a simple hook.
Now I must make another custom hook.

Andarist: I'm hesitant about allowing `parent` here because within the current system all children should be linked to their parents. It's easy to achieve that when only the parent itself can spawn an "attached" child. Allowing `parent` here would allow bypassing that and potentially leading to bizarre/corrupted states.

I have no problem with allowing the `id` option here. I would just treat `useSpawn` as always spawning a "root" actor.

davidkpiano:
Thank you for the PR. `useSpawn` is removed in v5; instead, actors should be created directly.

PR #3421, opened 2022-06-23
Add a test for exiting exiting a deep descendant during a self-transition

Andarist:
I was testing some scenarios to see if I can break the existing algorithm - I couldn't make it happen so that's good. Since I've already written this test I've concluded that it might be worthwhile to add it either way.

### DIFF
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -485,6 +485,37 @@ describe('entry/exit actions', () => {
         newLightMachine.transition('red', 'NOTHING').actions.map((a) => a.type)
       ).toEqual(['exit_walk', 'exit_red', 'enter_red', 'enter_walk']);
     });
+
+    it('should exit deep descendant during a self-transition', () => {
+      const actual: string[] = [];
+      const m = createMachine({
+        initial: 'a',
+        states: {
+          a: {
+            on: {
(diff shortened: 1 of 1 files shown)


packages/core/test/actions.test.ts
```diff
@@ -485,6 +485,37 @@ describe('entry/exit actions', () => {
         newLightMachine.transition('red', 'NOTHING').actions.map((a) => a.type)
       ).toEqual(['exit_walk', 'exit_red', 'enter_red', 'enter_walk']);
     });
+
+    it('should exit deep descendant during a self-transition', () => {
```
davidkpiano, line 489:
```suggestion
    it('should exit deep descendant during an external self-transition', () => {
```

davidkpiano (approved):

packages/core/test/actions.test.ts
Andarist: This somewhat doesn't matter because:
> The behavior of transitions with 'type' of "internal" is identical, except in the case of a transition whose source state is a compound state and whose target(s) is a descendant of the source. In such a case, an internal transition will not exit and re-enter its source state, while an external one will, as shown in the example below.

In this case, the target is not a descendant of the source and thus the behavior is identical for both external and internal transitions.

PR #3424, opened 2022-06-24
Fixed an issue with targeted ancestor reentrancy

### LINKED ISSUE #3409: Bug: Entry actions do not fire on external transitions from child state to parent state
### Description

When marking a transition in a child state as "external", the entry actions on the parent state don't seem to be fired.

Given this example machine:
```
import {
  createMachine,
} from "xstate";

  initial: 'loaded',
  states: {
    loaded: {
      entry: ['logEntry'],
      initial: 'idle',
      states: {
        idle: {
          on: {
            UPDATE: {
              target: '../loaded',
              actions: ['logAction'],
              internal: false
            }
          }
        }
      }
    },
  }
}, {
  actions: {
    logEntry: () => console.log('entry'),
    logAction: () => console.log('action')
  }
});
```

### Expected result

I expected that after firing the `UPDATE` event, the `logEntry` and `logAction` actions would have both been fired.

```
entry # initial logEntry call when the machine starts
action # logAction called for the UPDATE event
entry # logEntry called when re-entering the loaded state
```

### Actual result

This is the actual sequence I get:

```
entry # initial logEntry call when the machine starts
action # logAction called for the UPDATE event
```

### Reproduction

https://codesandbox.io/s/mutable-microservice-m0ilfl?file=/src/index.js

### Additional context

_No response_

Andarist:
fixes #3409

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -956,8 +956,8 @@ class StateNode<
     const reentryNodes: StateNode<any, any, any, any, any>[] = [];
 
     if (!isInternal) {
-      allNextStateNodes.forEach((n1) => {
-        reentryNodes.push(...this.getExternalReentryNodes(n1));
+      nextStateNodes.forEach((targetNode) => {
+        reentryNodes.push(...this.getExternalReentryNodes(targetNode));
       });
     }
 
@@ -972,23 +972,24 @@ class StateNode<
   }
 
   private getExternalReentryNodes(
-    childStateNode: StateNode<TContext, any, TEvent, any, any, any>
+    targetNode: StateNode<TContext, any, TEvent, any, any, any>
   ): Array<StateNode<TContext, any, TEvent, any, any, any>> {
     const nodes: Array<StateNode<TContext, any, TEvent, any, any, any>> = [];
-    let marker:
-      | StateNode<TContext, any, TEvent, any, any, any>
-      | undefined = childStateNode;
+    let [marker, possibleAncestor]: [
+      StateNode<TContext, any, TEvent, any, any, any> | undefined,
+      StateNode<TContext, any, TEvent, any, any, any>
+    ] = targetNode.order > this.order ? [targetNode, this] : [this, targetNode];
 
-    while (marker && marker !== this) {
+    while (marker && marker !== possibleAncestor) {
       nodes.push(marker);
       marker = marker.parent;
     }
-    if (marker !== this) {
-      // we never got to `this`, therefore the `childStateNode` "escapes" it
+    if (marker !== possibleAncestor) {
+      // we never got to `possibleAncestor`, therefore the initial `marker` "escapes" it
       // it's in a different part of the tree so no states will be reentered for such an external transition
       return [];
     }
-    nodes.push(this);
+    nodes.push(possibleAncestor);
     return nodes;
   }
 
--- a/.changeset/unlucky-sloths-tell.md
+++ b/.changeset/unlucky-sloths-tell.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Fixed an issue with targeted ancestors not being correctly reentered during external transitions.
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -383,6 +383,34 @@ describe('entry/exit actions', () => {
       expect(stateA.actions.map((action) => action.type)).toEqual(['D2 Exit']);
     });
 
+    it("should reenter targeted ancestor (as it's a descendant of the transition domain)", () => {
+      const actual: string[] = [];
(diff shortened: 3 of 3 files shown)


packages/core/src/StateNode.ts
```diff
@@ -956,8 +956,8 @@ class StateNode<
     const reentryNodes: StateNode<any, any, any, any, any>[] = [];
 
     if (!isInternal) {
-      allNextStateNodes.forEach((n1) => {
-        reentryNodes.push(...this.getExternalReentryNodes(n1));
+      nextStateNodes.forEach((targetNode) => {
```
Andarist, line 959: Note the change here from `allNextStateNodes` to `nextStateNodes`.

`nextStateNodes` represent the target state nodes literally used in the transition definition whereas `allNextStateNodes` represent "resolved" targets (initial states etc already resolved). To compute correctly what has to be reentered we need to consider the target states before resolving them.

packages/core/src/StateNode.ts
```diff
@@ -972,23 +972,24 @@ class StateNode<
...
+    targetNode: StateNode<TContext, any, TEvent, any, any, any>
   ): Array<StateNode<TContext, any, TEvent, any, any, any>> {
     const nodes: Array<StateNode<TContext, any, TEvent, any, any, any>> = [];
-    let marker:
-      | StateNode<TContext, any, TEvent, any, any, any>
-      | undefined = childStateNode;
+    let [marker, other]: [
+      StateNode<TContext, any, TEvent, any, any, any> | undefined,
+      StateNode<TContext, any, TEvent, any, any, any>
+    ] = targetNode.order > this.order ? [targetNode, this] : [this, targetNode];
```
Andarist, line 981: somewhat neat trick with comparing those orders here:
- `targetNode.order > this.order` means that `targetNode` is either a descendant of `this` OR it's somewhere outside of it, within the latter siblings of one of its ancestors
- `targetNode.order < this.order` means that `targetNode` is either the ancestor of `this` OR it's somewhere outside of it, within the former siblings of one of its ancestors

Since this function returns `[]` when those two do not form an ancestor<->descendant relationship we can safely rely on this as we are not even concerned here with finding a common ancestor of the two, we only want to find out if one is the ancestor of the other one and collect all nodes in between if that's the case (including the detected ancestor node).

packages/core/src/StateNode.ts
```diff
@@ -972,23 +972,24 @@ class StateNode<
...
     }
-    if (marker !== this) {
-      // we never got to `this`, therefore the `childStateNode` "escapes" it
+    if (marker !== possibleAncestor) {
+      // we never got to `possibleAncestor`, therefore the initial `marker` "escapes" it
       // it's in a different part of the tree so no states will be reentered for such an external transition
       return [];
     }
-    nodes.push(this);
+    nodes.push(possibleAncestor);
```
Andarist, line 992: Observation - this sort of uses the parent of the found ancestor as the transition domain which means that we are using LCA algorithm here whereas SCXML is using LCCA. I'm not exactly sure why SCXML has switched to LCCA because I think that they have been using LCA at the beginning. It feels to me that LCA is an easier rule and LCCA is a little bit harder to explain so I wonder what made them switch those algorithms.

packages/core/src/StateNode.ts
davidkpiano: Very clever! To prevent any edge-cases, we should ensure that `order` is not configurable (I think it technically is right now)

packages/core/src/StateNode.ts
davidkpiano: Would this break any SCXML tests?

davidkpiano (approved):
The code + test makes sense. We just have to ensure this is working in v5, and maybe even copy the test over.

packages/core/src/StateNode.ts
Andarist: I don't think there is currently any risk here because `order` is a local variable initialized to a static `0` and only assigned within the `dfs` call here:
https://github.com/statelyai/xstate/blob/e35493f59d277ca57f0982417d5ba3bca0a352ed/packages/core/src/StateNode.ts#L374-L385

packages/core/src/StateNode.ts
Andarist: Using LCA over LCCA? Sure. Using LCCA can lead to [some weird behaviors](https://lists.w3.org/Archives/Public/www-voice/2013JanMar/0059.html) when parallel states are involved though and I'm not sure why those weird behaviors would be desirable in the first place. We should dig into this more to understand the motivation for using LCCA in UML.

Andarist:
> We just have to ensure this is working in v5,

I've checked - it already works OK there.

> and maybe even copy the test over.

That will just happen automatically~ when we get to merging main to next, so no specific additional actions are required here.

PR #3432, opened 2022-06-29
add every for periodic transitions

moh12594:
Hello 👋

This PR makes possible to add an `every` event that will trigger an event/transition recurrently after each period.

```
const lightPeriodsMachine = createMachine({
  id: 'lightPeriod',
  initial: 'green',
  context: {
    timeSpentInGreen: 0,
    timeSpentInYellow: 0
  },
  states: {
    green: {
      every: {
        // every 1 second, timeSpentInGreen will be increased by 1 second
        1000: {
          actions: assign({
            timeSpentInGreen: (context) => context.timeSpentInGreen + 1
          })
        }
      },
      on: {
        GO_TO_YELLOW: 'yellow'
      }
    },
    yellow: {
      every: {
        // every 0.5 seconds, timeSpentInYellow will be increased by 0.5 seconds
        500: {
          actions: assign({
            timeSpentInYellow: (context) => context.timeSpentInYellow + 0.5
          })
        }
      }
    }
  }
});
```

https://www.itemis.com/en/yakindu/state-machine/documentation/user-guide/sclang_reactions#sclang_reaction_trigger_every

### DIFF
--- a/docs/guides/periods.md
+++ b/docs/guides/periods.md
@@ -56,7 +56,7 @@ states: {
 // ...
 ```
 
-Periodic events can also be conditional with regard to a single period value:
+Periodic events can also be conditional with regard to a single interval value:
 
 ```js
 // ...
@@ -75,7 +75,7 @@ states: {
 
 ### Periodic expressions on transitions <Badge text="4.4+" />
 
-Periodic events specified on the `every: { ... }` property can have dynamic priods, specified either by a string period reference:
+Periodic events specified on the `every: { ... }` property can have dynamic priods, specified either by a string interval reference:
 
 ```js
 const lightDelayMachine = createMachine(
@@ -102,7 +102,7 @@ const lightDelayMachine = createMachine(
   },
   {
     // String delays configured here
-    periods: {
+    intervals: {
       LIGHT_DELAY: (context, event) => {
         return context.trafficLevel === 'low' ? 1000 : 3000;
       },
@@ -129,14 +129,14 @@ green: {
 // ...
 ```
 
-However, prefer using string period references, just like the first example, or in the `period` property:
+However, prefer using string interval references, just like the first example, or in the `interval` property:
 
 ```js
 // ...
 green: {
   after: [
     {
-      period: 'LIGHT_DELAY',
+      interval: 'LIGHT_DELAY',
       actions: doSomeAction
     }
   ]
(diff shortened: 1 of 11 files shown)


docs/guides/periods.md
```diff
@@ -0,0 +1,228 @@
...
+## Periodic events
+
+If you just want to send an event every period of time, you can specify the `period` as an option in the second argument of the `send(...)` action creator:
+
+```js
+import { actions } from 'xstate';
+const { send } = actions;
+
+// action to send the 'TIMER' event every 1 second
+const sendTimerEvery1Second = send({ type: 'TIMER' }, { period: 1000 });
```
davidkpiano, line 156: I'd probably call this `interval` rather than `period`

moh12594: Yes, I had a doubt about the naming :)) I'm going to edit it

moh12594: I changed it to `interval` @davidkpiano ! Thank you

docs/guides/periods.md
```diff
@@ -0,0 +1,228 @@
...
+        after: {
+          YELLOW_LIGHT_DELAY: { actions: doTheYellowThing }
+        }
+      }
+      // ...
+    }
+  },
+  {
+    // String delays configured here
+    intervals: {
```
Andarist, line 105: It feels to me that in general, the difference between `delays` and `intervals` is so small that we could just merge them and keep using `delays` for intervals.

docs/guides/periods.md
```diff
@@ -0,0 +1,228 @@
...
+  every: [
+    {
+      priod: (context, event) => {
+        return context.trafficLevel === 'low' ? 1000 : 3000;
+      },
+      target: 'yellow'
+    }
+  ]
+},
+// ...
```
Andarist, line 129: I don't understand this example - an `every` transition with a target doesn't make sense to me. What's the point? Especially when we consider external transitions then a source state (the one with `every` defined on it) is always exited, so the interval would get cleaned up (and sometimes set up again). Should we mandate internal transitions for those?

docs/guides/periods.md
```diff
@@ -0,0 +1,228 @@
...
+## Periodic events
+
+If you just want to send an event every period of time, you can specify the `interval` as an option in the second argument of the `send(...)` action creator:
+
+```js
+import { actions } from 'xstate';
+const { send } = actions;
+
+// action to send the 'TIMER' event every 1 second
+const sendTimerEvery1Second = send({ type: 'TIMER' }, { interval: 1000 });
```
Andarist, line 156: How useful this is if you can already leverage `every: { 1000: { actions: send('TIMER')  } }` combination?

Andarist (commented):
The idea is interesting but I'm rather opposed to introducing it right now. I think it would be best to revisit this once v5 gets released - XState has a lot of accompanying tools around it and introducing a feature is not just a matter of introducing it to the XState's core itself. We also need to think about it from the perspective of the Studio, typegen and more.

PR #3439, opened 2022-07-03
[@xstate/test] Add support for specific `initialState`

davidkpiano:
This PR adds support for generating paths from a specific `initialState` rather than the machine's default initial state.

The use-case for this is for custom path generation, where a dev might want to **join paths**. For example, if trying to generate a behavioral (given-when-then) test, they would want to:

1. Generate path(s) to the _given_ state
2. Generate paths starting at the _given_ state and then from the event(s) (the _when_ part)
3. Join the first and second set of paths

Another use-case is for E2E test flows where a predefined initial state is not needed; e.g. navigating directly to some page instead of the home page, and running the test model from there.

See here for more info: https://discord.com/channels/795785288994652170/799416911259369472/989132676986437673


packages/xstate-graph/test/graph.test.ts
```diff
@@ -438,106 +430,14 @@ describe('@xstate/graph', () => {
         ])
       ).toThrow();
     });
-  });
-
-  describe('getValueAdjacencyMap', () => {
```
davidkpiano, line 443: `getValueAdjacencyMap` was deleted as it is unused

Andarist:
@davidkpiano I've pushed out an alternative history to drop those unrelated commits from this branch (something like `git pull --rebase` should let you download this), also fixed some minor unused imports issue but the PR still doesn't build because it has some type errors 

---

The PR's description mentions something like:
> Join the first and second set of paths

But I'm not seeing that in the code and thus not sure if the described goal actually matches the implementation and how all of this fits together.

Could you write down an e2e usage example of this? The added tests don't really showcase how this is supposed to be used.

davidkpiano:
Redoing as #3446

PR #3444, opened 2022-07-05
Fixed an issue with reentrancy during external transition when targeting a history state

Andarist:
Fixes a regression from https://github.com/statelyai/xstate/pull/3424/files#diff-7f553f4d4c285f1c05cced2e979c5b957d8884e39e74cc6797b32c63a0be7e7aR981 where I've started using `.order` to compare source and target positions in the tree. This has not yet been published so we don't need to add a changeset.

I've removed `.order` from history nodes here: https://github.com/statelyai/xstate/pull/3171/files#r834036363 but back then it was OK because `.order` wasn't actually used in a meaningful way that would impact history states.

### DIFF
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -15,12 +15,16 @@ export const isLeafNode = (
   stateNode: StateNode<any, any, any, any, any, any>
 ) => stateNode.type === 'atomic' || stateNode.type === 'final';
 
+export function getAllChildren<TC, TE extends EventObject>(
+  stateNode: StateNode<TC, any, TE>
+): Array<StateNode<TC, any, TE>> {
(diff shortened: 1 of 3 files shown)


packages/core/src/stateUtils.ts
```diff
@@ -15,12 +15,16 @@ export const isLeafNode = (
   stateNode: StateNode<any, any, any, any, any, any>
 ) => stateNode.type === 'atomic' || stateNode.type === 'final';
 
+export function getAllChildren<TC, TE extends EventObject>(
```
Andarist, line 18: we probably could name those 2 in a better way - but the overall algorithm is overhauled in v5 so I didn't bother too much with figuring out better names right now

Andarist:
Fixing before bugs hit the production - that's just how we roll 🎲

PR #3446, opened 2022-07-05
[@xstate/graph][@xstate/test] Add support for `initialState`; clean up API

davidkpiano:
This PR adds support for generating paths from a specific initialState rather than the machine's default initial state.

The use-case for this is for custom path generation, where a dev might want to join paths. For example, if trying to generate a behavioral (given-when-then) test, they would want to:

- Generate path(s) to the given state
- Generate paths starting at the given state and then from the event(s) (the when part)
- Join the first and second set of paths

Another use-case is for E2E test flows where a predefined initial state is not needed; e.g. navigating directly to some page instead of the home page, and running the test model from there.

See here for more info: https://discord.com/channels/795785288994652170/799416911259369472/989132676986437673


packages/xstate-test/test/paths.test.ts
```diff
@@ -255,3 +255,41 @@ describe('getShortestPathsTo', () => {
...
+    // a (OTHER) -> b
+    expect(pathsToB).toHaveLength(2);
+
+    const shortestPaths = model.getShortestPathsFrom(pathsToB);
+
+    // a (NEXT) -> b (TO_C) -> c
+    // a (OTHER) -> b (TO_C) -> c
+    // a (NEXT) -> b (TO_D) -> d
+    // a (OTHER) -> b (TO_D) -> d
+    expect(shortestPaths).toHaveLength(4);
```
Andarist, line 291: From this alone it's not completely obvious how this works.

Most existing methods have various flavors and ways to limit path generation. This one only accepts a list of paths (and performs joining under the hood?) but doesn't allow us to do any filtering. It seems that there is a lack of a certain "symmetry" in those APIs

`*From` is not that descriptive here, perhaps `*FromPaths` would work better?

packages/xstate-test/src/TestModel.ts
```diff
@@ -409,6 +410,21 @@ export class TestModel<TState, TEvent extends EventObject> {
...
+  public getShortestPathsFrom(
+    paths: Array<TestPath<TState, TEvent>>
+  ): Array<TestPath<TState, TEvent>> {
+    const resultPaths: any[] = [];
+
+    for (const path of paths) {
+      const shortestPaths = this.getShortestPaths({ initialState: path.state });
+      for (const shortestPath of shortestPaths) {
+        resultPaths.push(joinPaths(path, shortestPath));
+      }
```
Andarist, line 423: I'm not sure about this but looking at the usage here... it looks like `this.getShortestPaths` won't consider the existing path. This means that the final path might look somewhat different than a path generated from the true initial state to the "destination" state (as the `getShortestPaths` will start with `visited: emptySet`, or similar). Is it a bug or a feature? :p

packages/xstate-graph/test/graph.test.ts
```diff
@@ -699,3 +603,99 @@ it('should provide previous state for serializeState()', () => {
...
+          on: { toC: 'c' }
+        },
+        c: {
+          on: { toA: 'a' }
+        }
+      }
+    });
+
+    const plans = getPlans(machine, {
+      initialState: machine.resolveState(State.from('b'))
```
Andarist, line 626: In general, I'm not super fond of specifying `initialState` like this. The problem itself is not that providing an initial state is bad per see but rather that we are able to use a somewhat arbitrary state as an initial state, one which might be invalid (not possible to achieve by transitioning from the true initial state) given the current config of the machine.

Maybe it's just something that the user has to be aware of. I just wonder if we can do any better? Nothing comes to my mind right now though because "proving" the initial state would be rather complex and not practical 🤔 

In the context of e2e tests though it should almost be impossible to do it. I assume that rarely one can "start" their model from an arbitrary point (joining paths is different since in there the provided `initialState` comes from the already generated paths that are starting in the true initial state.

packages/xstate-graph/src/graph.ts
```diff
@@ -675,3 +604,20 @@ export function traverseShortestPathsFromTo<TState, TEvent extends EventObject>(
...
+  path2: StatePath<TState, TEvent>
+): StatePath<TState, TEvent> {
+  const secondPathSource = path2.steps[0]?.state ?? path2.state;
+
+  if (secondPathSource !== path1.state) {
+    throw new Error(`Paths cannot be joined`);
+  }
+
+  return {
+    state: path1.state,
```
Andarist, line 619: Shouldn't this hold the "target"/destination state? If yes then we should use the `path2.state` as the final joined path is going to end up in that.
```suggestion
    state: path2.state,
```

packages/xstate-test/src/TestModel.ts
davidkpiano: This is a **feature**. The final path should look different from the path generated from the "true initial state" when you specify the "from state" that the algorithm should start at.

In the below example, the shortest path from the initial state to state B is shown in purple, but if we specify an initial state of **Y**, it _should_ take a different shortest path.

CleanShot 2022-07-21 at 08 40 47

packages/xstate-graph/test/graph.test.ts
davidkpiano: This is only meant to be "internal", specific to `@xstate/graph`. 

If you notice the test in `@xstate/test` (https://github.com/statelyai/xstate/pull/3446/files#diff-d7468424c510a8f0d4f4fcf0d2cbfe0ecda06e489f98982f501eb4275698a51eR259), you'll see that we're reaching the "custom initial state" by first getting the path to that desired state:

```js
const pathsToB = model.getShortestPathsTo((state) => state.matches('b'));
```

If we had `state.matches('someImpossibleState')`, the result will be an _empty array_, signaling that it is impossible to reach that state.

So in that sense, we already have safety for e2e tests.

packages/xstate-test/test/paths.test.ts
davidkpiano: What would make this method more symmetrical? Is it the options that are missing?

packages/xstate-graph/src/graph.ts
davidkpiano: Fixed + test added in [963b429](https://github.com/statelyai/xstate/pull/3446/commits/963b429d91a9dbfe3ac9f0a95a4ccef911a79cdf)

packages/xstate-test/src/TestModel.ts
davidkpiano: Does this make sense?

packages/xstate-test/test/paths.test.ts
Andarist: With this PR we have those on the `TestModel`:
```ts
getShortestPaths(options?: Partial<TraversalOptions<TState, TEvent>>): Array<TestPath<TState, TEvent>>;
getPaths(options?: Partial<GetPathsOptions<TState, TEvent>>): Array<TestPath<TState, TEvent>>;
getShortestPathsTo(statePredicate: StatePredicate<TState>): Array<TestPath<TState, TEvent>>;
getSimplePaths(options?: Partial<TraversalOptions<TState, any>>): Array<TestPath<TState, TEvent>>;
getSimplePathsTo(predicate: StatePredicate<TState>): Array<TestPath<TState, TEvent>>;
getPathFromEvents(events: TEvent[], statePredicate: StatePredicate<TState>): TestPath<TState, TEvent>;
getShortestPathsFrom(paths: Array<TestPath<TState, TEvent>>): Array<TestPath<TState, TEvent>>;
```

We can remove `getPathFromEvents` from this set as it seems to be slightly different. However, I'm not quite sure what it does **exactly** because it still calls `performDepthFirstTraversal` internally - so it could also be considered a part of this set of methods, even though its goal might be slightly different.

I see that this added method is not particularly unique in being different but at the very least it's not "symmetric" with others because it doesn't come with a "simple" variant (it only has the "shortest" variant). I think though that it might useful to consolidate stuff further anyway:
1. why does `getShortestPaths` allow options like `eventCases` but `getShortestPathsTo` doesn't? The goal of the latter is to allow for a custom `statePredicate` and not to limit other useful options
2. why do we have `getShortestPaths` and `getPaths`? They both do the same thing **by default**. I think this is a mistake that we should fix. Having aliases like this doesn't bring any real value beyond one of those being shorter and thus maybe it being more visually pleasing. I feel that such aliases come with a real tradeoffs:
- it's bikesheddable for teams which flavour do they want to use
- u need to document both, different articles online might starts using one or another and in the end you create more overall mental overhead for the consumers because of that
- a reader must realize that they are the same or draw the wrong conclusion that they are different "somehow". When I see A and B APIs I won't ever assume that they are merely aliases - I will always assume that there is a meaningful difference between them and I will look for it eventually
- it hurts searchability/"grepability" - it's not uncommon to search the whole codebase for occurrences of a certain API, trying to understand how they are used etc. If we provide aliases and they are both used in the codebase then it makes this job harder to do. 
- I know that `getPaths` actually allow custom `pathGenerator` but I think that in such a case it should **require** it and not default to the shortest paths algorithm (because of the mentioned reasons)

I think that perhaps there is a room for **merging** all of those into one method that would automatically support all of the supported options for all of those flavours~. Of course, some would be defaulted and in such a case it would be OK to default to the shortest paths.

Alternatively, if we want to simplify providing those "standard" path generators then we could have:
- `getShortestPaths(options)`
- `getSimplestPaths(options)`
- `getPaths(optionsWithRequiredPathGenerator)`

packages/xstate-test/src/TestModel.ts
Andarist: Yes, it makes sense - could we add a test case that would showcase exactly that? It could be associated with a comment above the `expect` call that would clarify this and make it clear that this is a notable difference from the standard usage that is usually not achievable without `joinPaths`

packages/xstate-graph/test/graph.test.ts
Andarist: > This is only meant to be "internal", specific to @xstate/graph.

Note: it's still available as the option to `TestModel#getShortestPaths`. Perhaps my main concern here is the name - since it's not exactly the "initial" state in the same sense that we use the same name in XState itself. Maybe a `startState` or `traversalStartState` would be a better fit here (or at least a fit that would differentiate it from other places where we use "initial state" differently)

davidkpiano:
@Andarist @farskid Ping on this review?

Andarist:
@davidkpiano u have some unaddressed comments from the previous round of review

packages/xstate-test/test/paths.test.ts
davidkpiano: I completely agree with you regarding `.getPaths(...)`, it should not be an alias and instead require a path generator.

packages/xstate-graph/test/graph.test.ts
davidkpiano: That makes sense. I'd opt to call it `fromState`, as the goal is to find the paths from a specific state rather than from the true initial state.

packages/xstate-test/test/paths.test.ts
davidkpiano: Made the path generator required in [49669d0](https://github.com/statelyai/xstate/pull/3446/commits/49669d052d1e2b5b6bcc63b064386c343599c663)

packages/xstate-graph/src/graph.ts
```diff
@@ -365,33 +286,38 @@ export function getPathFromEvents<
...
-  const { serializeState, serializeEvent } = optionsWithDefaults;
+  const { serializeState, serializeEvent } = resolvedOptions;
 
-  const adjacency = performDepthFirstTraversal(behavior, optionsWithDefaults);
+  const adjacency = performDepthFirstTraversal(behavior, resolvedOptions);
 
   const stateMap = new Map<SerializedState, TState>();
   const path: Steps<TState, TEvent> = [];
 
   const initialSerializedState = serializeState(
```
Andarist, line 312: this variable should be renamed now

packages/xstate-graph/src/graph.ts
```diff
@@ -443,14 +369,16 @@ export function performDepthFirstTraversal<TState, TEvent extends EventObject>(
...
-  const { transition, initialState } = behavior;
+  const { transition } = behavior;
   const {
     serializeEvent,
     serializeState,
     getEvents,
     eventCases,
-    traversalLimit: limit
+    traversalLimit: limit,
+    fromState: customInitialState
```
Andarist, line 379:
```suggestion
    fromState: customFromState
```

packages/xstate-test/src/TestModel.ts
```diff
@@ -82,58 +85,32 @@ export class TestModel<TState, TEvent extends EventObject> {
...
 
   public getSimplePaths(
     options?: Partial<TraversalOptions<TState, any>>
   ): Array<TestPath<TState, TEvent>> {
-    return this.getPaths({
-      ...options,
-      pathGenerator: getSimplePaths
+    return this.getPaths(getSimplePaths, {
+      ...options
     });
```
Andarist, line 114:
```suggestion
    return this.getPaths(getSimplePaths, options);
```

packages/xstate-test/src/TestModel.ts
```diff
@@ -82,58 +85,32 @@ export class TestModel<TState, TEvent extends EventObject> {
...
-    return this.getPaths({ ...options, pathGenerator: getShortestPaths });
-  }
-
-  private _getStatePaths(
-    options?: Partial<GetPathsOptions<TState, TEvent>>
-  ): Array<StatePath<TState, TEvent>> {
-    const pathGenerator =
-      options?.pathGenerator || TestModel.defaults.pathGenerator;
-    return pathGenerator(this.behavior, this.resolveOptions(options));
+    return this.getPaths(getShortestPaths, { ...options });
```
Andarist, line 88:
```suggestion
    return this.getPaths(getShortestPaths, options);
```

packages/xstate-test/test/paths.test.ts
Andarist: 1. what does `getPathFromEvents` do?
2. should we introduce `getSimplestPathsFrom` to "match" `getShortestPathsFrom` and bring back the balance to the universe? 😛

packages/xstate-graph/src/adjacency.ts
```diff
@@ -0,0 +1,79 @@
...
+    serializeEvent,
+    serializeState,
+    getEvents,
+    eventCases,
+    traversalLimit: limit,
+    fromState: customInitialState,
+    stopCondition
+  } = resolveTraversalOptions(options);
+  const fromState = customInitialState ?? behavior.initialState;
+  const adj: AdjacencyMap<TState, TEvent> = {};
```
Andarist, line 25:
```suggestion
    fromState: customFromState,
    stopCondition
  } = resolveTraversalOptions(options);
  const fromState = customFromState ?? behavior.initialState;
  const adj: AdjacencyMap<TState, TEvent> = {};
```

packages/xstate-graph/src/pathFromEvents.ts
```diff
@@ -0,0 +1,87 @@
...
+} from './graph';
+
+export function getPathFromEvents<
+  TState,
+  TEvent extends EventObject = EventObject
+>(
+  behavior: SimpleBehavior<TState, TEvent>,
+  events: TEvent[],
+  options?: TraversalOptions<TState, TEvent>
+): StatePath<TState, TEvent> {
```
Andarist, line 24: We have a method in the `TestModel` that is named exactly like this (in fact, that method calls this function here) **but** it has a different signature. This is kinda confusing - I would expect their APIs to be closer to each other:
https://github.com/statelyai/xstate/blob/eeea324a9fcbc2ac6771fb3b9df3fa97fc073860/packages/xstate-test/src/TestModel.ts#L147-L150

packages/xstate-graph/src/pathFromEvents.ts
```diff
@@ -0,0 +1,87 @@
...
+  const path: Steps<TState, TEvent> = [];
+
+  const initialSerializedState = serializeState(
+    fromState,
+    undefined,
+    undefined
+  ) as SerializedState;
+  stateMap.set(initialSerializedState, fromState);
+
+  let stateSerial = initialSerializedState;
```
Andarist, line 52:
```suggestion
  const serializedFromState = serializeState(
    fromState,
    undefined,
    undefined
  ) as SerializedState;
  stateMap.set(serializedFromState, fromState);

  let stateSerial = serializedFromState;
```

packages/xstate-graph/src/shortestPaths.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  weightMap.set(initialSerializedState, {
+    weight: 0,
+    state: undefined,
+    event: undefined
+  });
+  const unvisited = new Set<SerializedState>();
+  const visited = new Set<SerializedState>();
+
+  unvisited.add(initialSerializedState);
```
Andarist, line 64:
```suggestion
  const serializedFromState = serializeState(
    fromState,
    undefined,
    undefined
  );
  stateMap.set(serializedFromState, fromState);

  weightMap.set(serializedFromState, {
    weight: 0,
    state: undefined,
    event: undefined
  });
  const unvisited = new Set<SerializedState>();
  const visited = new Set<SerializedState>();

  unvisited.add(serializedFromState);
```

packages/xstate-graph/src/simplePaths.ts
```diff
@@ -0,0 +1,133 @@
...
+): Array<StatePath<TState, TEvent>> {
+  const resolvedOptions = resolveTraversalOptions(options);
+  const simplePaths = getSimplePaths(behavior, resolvedOptions);
+
+  return simplePaths.filter((path) => {
+    return (
+      toPredicate(path.state) &&
+      path.steps.some((step) => fromPredicate(step.state))
+    );
+  });
```
Andarist, line 132: I've noticed that this behaves differently from `getShortestPathsFromTo`. `getShortestPathsFromTo` first computes "from" paths and it only computes "to" paths from those. It's a better algorithm than the one here because here we have to generate **everything** and reject paths that don't match later on. 

How hard would it be to unify the approaches? Is there a reason why all simple paths have to be generated upfront for us to get an answer on the `FromTo` question?

packages/xstate-graph/src/types.ts
```diff
@@ -171,7 +197,14 @@ export interface TraversalOptions<TState, TEvent extends EventObject>
    *
    * @default `Infinity`
    */
-  traversalLimit?: number;
+  traversalLimit: number;
+  fromState: TState | undefined;
+  /**
+   * When true, traversal of the adjacency list will stop
```
Andarist, line 203: The concept of adjacencies could be explained. I'm not sure what is the difference between an adjacency map and an adjacency list. Are those relevant to the public API/to the user? Or are those implementation details? By mentioning them in the JS Doc we will surface them to the users as this will appear in the IDE tooltips etc

packages/xstate-graph/src/index.ts
```diff
@@ -1,15 +1,24 @@
...
+  getMachineSimplePaths,
+  getSimplePaths,
+  getSimplePathsFromTo
+} from './simplePaths';
+export {
+  getShortestPaths,
+  getMachineShortestPaths,
+  getShortestPathsFromTo,
+  getMachineShortestPathsFromTo
+} from './shortestPaths';
```
Andarist, line 20: The balance is missing here :P It seems that there is no `getMachineSimplePathsFromTo`

davidkpiano: 215bbbeb0572c8c4d178605ad4cf1be5c953b009

packages/xstate-graph/src/types.ts
davidkpiano: Tried to clarify it.

An adjacency list is the state -> event -> nextState relations. Example, given this machine:

```js
  initial: 'green',
  states: {
    green: { on: { TIMER: 'yellow' } },
    yellow: { on: { TIMER: 'red' } },
    red: { on: { TIMER: 'green' } }
  }
});
```

An adjacency list (array) would look like this (with states, not state values):
```
[
  { state: 'green', event: 'TIMER', nextState: 'yellow' }, // green → yellow
  { state: 'yellow', event: 'TIMER', nextState: 'red' }, // yellow → red
  { state: 'red', event: 'TIMER', nextState: 'green' } // red → green
]
```

packages/xstate-graph/src/simplePaths.ts
davidkpiano: Improved here: 286190b51d9a9629e85e5a5244c6189a82e8ba7b

packages/xstate-graph/src/pathFromEvents.ts
davidkpiano: This is now unified

Andarist (approved):

Issue #3452, opened 2022-07-10
Bug: completeListener on Subscription is not triggered on stop


pixtron:
### Description

When an interpreter is stoped, [all listeners get removed](https://github.com/statelyai/xstate/blob/6badd2ba3642391bee640aa4914003ad57f2e703/packages/core/src/interpreter.ts#L559-L562), therefore a potential subscription should `complete`. 

```
const machine = createMachine({...});
const service = interpret(machine)
const sub = service.subscribe(
  (state) => console.log('change', state),
  () => console.log('error'),
  () => console.log('complete')
);
service.start();
service.stop();
```

### Expected result

`complete` gets logged.

### Actual result

`complete` is not logged.

### Reproduction

https://codesandbox.io/s/competent-sunset-qmm9h8?file=/src/index.ts

### Additional context

As the [interpreter is stoped when entering a final state](https://github.com/statelyai/xstate/blob/6badd2ba3642391bee640aa4914003ad57f2e703/packages/core/src/interpreter.ts#L363) the [`resolvedCompleteListener`](https://github.com/statelyai/xstate/blob/6badd2ba3642391bee640aa4914003ad57f2e703/packages/core/src/interpreter.ts#L440) should probably be removed from `doneListeners` and `stopListeners` once it has been called, to avoid multiple calls to it. Eg in the repro `s2 complete` should only be logged once.

Andarist:
The question here is... if `.stop()` is like `.complete()` or like `.cancel()`? And if `complete` callback should be called in case of a `cancel` if we'd chose the latter semantics?

I think that I'm leaning towards your expected result here... but I need to rethink this.

pixtron:
@Andarist when i subscribe after the interpreter has been stopped, the [`resolvedCompleteListener` gets called immediately](https://github.com/statelyai/xstate/blob/6badd2ba3642391bee640aa4914003ad57f2e703/packages/core/src/interpreter.ts#L437-L439), therefore i would expect the same behavior when the interpreter enters the state `Stopped`.

Andarist:
Would you like to work on a fix for this?

PR #3453, opened 2022-07-10
Call subscription.complete on interpreter stop

pixtron:
When an interpreter is stoped, [all listeners get removed](https://github.com/statelyai/xstate/blob/6badd2ba3642391bee640aa4914003ad57f2e703/packages/core/src/interpreter.ts#L559-L562), therefore a potential subscription should complete.

Closes: #3452

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -112,6 +112,29 @@ export enum InterpreterStatus {
   Stopped
 }
 
+// onceWrapper and _onceWrap from nodejs EventEmitter
+// https://github.com/nodejs/node/blob/3350d9610864af3219de7dd20e3ac18b3c214c52/lib/events.js#L622-L638
+function onceWrapper() {
+  if (!this.fired) {
+    if (this.target.off) {
+      this.target.off(this.wrapFn);
+    }
+    this.fired = true;
+    if (arguments.length === 0) {
+      return this.listener.call(this.target);
+    }
+    return this.listener.apply(this.target, arguments);
+  }
+}
+
+function _onceWrap(target: any, listener: () => void) {
+  const state = { fired: false, wrapFn: undefined, target, listener };
+  const wrapped = onceWrapper.bind(state);
+  wrapped.listener = listener;
+  state.wrapFn = wrapped;
+  return wrapped;
+}
+
 export class Interpreter<
   // tslint:disable-next-line:max-classes-per-file
   TContext,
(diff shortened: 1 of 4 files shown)


packages/core/src/interpreter.ts
```diff
@@ -112,6 +112,29 @@ export enum InterpreterStatus {
...
+    }
+    this.fired = true;
+    if (arguments.length === 0) {
+      return this.listener.call(this.target);
+    }
+    return this.listener.apply(this.target, arguments);
+  }
+}
+
+function _onceWrap(target: any, listener: () => void) {
```
pixtron, line 130: not sure how to type `target` which should be an instance of `Interpreter`

packages/core/src/interpreter.ts
```diff
@@ -388,8 +388,8 @@ export class Interpreter<
     return this;
   }
   public subscribe(
-    observer: Observer<
-      State<TContext, TEvent, any, TTypestate, TResolvedTypesMeta>
+    observer: Partial<
```
Andarist, line 391: I've rechecked with RxJS and listener types are actually all optional (well, not exactly, but I'm not sure if we want to do [this](https://github.com/ReactiveX/rxjs/blob/9a85e01625a1ee7a23e4ab2c2c6d3089a58e2e54/src/internal/types.ts#L170)).

It was really annoying in tests for me to implement this:
```js
service.subscribe({
  next: noop,
  error: noop,
  complete: completeCb
})
```
and I think that it makes sense to make it easier to just provide a single callback type, so I've followed that instinct and wrapped `Observer` at the parameter position `Partial`

packages/core/src/interpreter.ts
```diff
@@ -404,48 +404,39 @@ export class Interpreter<
...
+      this.doneListeners.delete(completeOnce);
+      this.stopListeners.delete(completeOnce);
+      observer.complete();
+    };
+
+    if (this.status === InterpreterStatus.Stopped) {
+      observer.complete();
+    } else {
+      this.onDone(completeOnce);
+      this.onStop(completeOnce);
```
Andarist, line 432: this was a little bit annoying, but I think it's the cleanest solution possible right now. This probably will be completely refactored in v5 anyway.

packages/core/src/utils.ts
```diff
@@ -699,20 +699,20 @@ export function toInvokeSource(
...
 
   return {
-    next: nextHandler,
-    error: errorHandler || noop,
-    complete: completionHandler || noop
+    next: ((isObserver ? nextHandler.next : nextHandler) || noop).bind(self),
+    error: ((isObserver ? nextHandler.error : errorHandler) || noop).bind(self),
+    complete: (
+      (isObserver ? nextHandler.complete : completionHandler) || noop
+    ).bind(self)
```
Andarist, line 715: I don't think that we really need those `bind`s here but we were already binding `next` listener in the `Interpreter#subscribe`. So I didn't want to regress this accidentally and I've just used the same code pattern for all the callbacks here.

mellson (approved):
The code looks top quality 👍🏻

PR #3456, opened 2022-07-13
[@xstate/react] Add warning about `useSelector` and non-primitive values

davidkpiano:


docs/packages/xstate-react/index.md
```diff
@@ -208,6 +208,29 @@ const App = ({ service }) => {
 };
 ```
 
+::: warn
+The default comparison is a strict reference comparison (`===`). If your selector returns non-primitive values, such as objects or arrays, you should keep this in mind and either return the same reference, or provide a shallow or deep comparator.
```
Andarist, line 212: I was wondering what React Redux does - since `connect` was always using a `shallowEqual` under the hood - but it seems that they have "switched" to using `===` when they have introduced `useSelector`:
https://github.com/reduxjs/react-redux/blob/720f0ba79236cdc3e1115f4ef9a7760a21784b48/src/hooks/useSelector.ts#L35

docs/packages/xstate-react/index.md
```diff
@@ -208,6 +208,29 @@ const App = ({ service }) => {
...
+
+One option is to use the popular [`fast-deep-equal`](https://www.npmjs.com/package/fast-deep-equal) library:
+
+```js
+import isDeepEqual from 'fast-deep-equal';
```
Andarist, line 217: I think it would be better to recommend one of the libraries for `shallowEqual`. Deep equality check can be expensive sometimes so I'm worried about people using this example from the docs too blindly.

davidkpiano: Good call; will replace this in the morning 😴

laurakalbag (approved):

packages/xstate-react/test/useSelector.test.tsx
```diff
@@ -137,6 +137,89 @@ describeEachReactMode('useSelector (%s)', ({ suiteKey, render }) => {
     expect(nameEl.textContent).toEqual('DAVID');
   });
 
+  it.only('should work with the shallowEqual comparison function', () => {
```
Andarist, line 140:
```suggestion
  it('should work with the shallowEqual comparison function', () => {
```

packages/xstate-react/src/utils.ts
```diff
@@ -22,3 +22,41 @@ export function getServiceSnapshot<
 >(service: TService): TService['state'] {
   return service.status !== 0 ? service.state : service.machine.initialState;
 }
+
+// From https://github.com/reduxjs/react-redux/blob/master/src/utils/shallowEqual.ts
```
Andarist, line 26:
```suggestion
// From https://github.com/reduxjs/react-redux/blob/720f0ba79236cdc3e1115f4ef9a7760a21784b48/src/utils/shallowEqual.ts
```

docs/packages/xstate-react/index.md
davidkpiano: I'm now doing what React-Redux does 👍

PR #3487, opened 2022-08-01
Make it impossible to exit a root state

### LINKED ISSUE #3072: Bug: Root invocations should never be restarted

Andarist:
closes #3072


packages/core/src/stateUtils.ts
```diff
@@ -998,25 +1003,6 @@ export function removeConflictingTransitions<
   return Array.from(filteredTransitions);
 }
 
-function findLCCA<TContext extends MachineContext, TEvent extends EventObject>(
```
Andarist, line 1001: I've inlined this logic into `getTransitionDomain` because it now can return the root state. The LCCA stands for the Least Common Compound Ancestor. We were never limiting ourselves here to return a compound state - so the name was already incorrect. We could have changed it to `findLCA` where LCA stands for Least Common Ancestor. However, when returning the root state from here it's not always even an ancestor of the input `stateNodes` (for example, `findLCA([root])` would return `root` and the root is not an ancestor of itself). So I've concluded that it's best to inline this function altogether to avoid the confusion and debate.

packages/core/src/stateUtils.ts
```diff
@@ -1078,9 +1064,25 @@ function getTransitionDomain<
...
 
-  return lcca;
+  let current = getPathFromRootToNode(head);
+
+  for (let i = 1; i < involvedStates.length; i++) {
+    const path = getPathFromRootToNode(involvedStates[i]);
+    current = current.filter((sn) => path.includes(sn));
+  }
+
+  const domain = current[current.length - 1];
```
Andarist, line 1077: We sort of had a bug here previously. We could return `undefined` from here, whereas the annotated return type is `StateNode<TContext, TEvent> | null`. Everything was working OK as `.parent === domainButActuallyUndefined` (and similar) was also a good stop for most of our internal while loops etc.

packages/core/src/stateUtils.ts
```diff
@@ -929,15 +929,20 @@ function getPathFromRootToNode<
...
+  const path: Array<typeof stateNode> = [];
+
   while (marker) {
-    path.unshift(marker);
+    path.push(marker);
     marker = marker.parent;
   }
 
-  return path;
+  return path.reverse();
```
Andarist, line 945: Conceptually it's a little bit better to push to an array than to shift it all the time (each shift has to move all the existing items). So I've switched to pushing here but to correct the actual result I need to `.reverse()` at the end.

packages/core/src/stateUtils.ts
```diff
@@ -929,15 +929,20 @@ function getPathFromRootToNode<
   TC extends MachineContext,
   TE extends EventObject
 >(stateNode: StateNode<TC, TE>): Array<StateNode<TC, TE>> {
-  const path: Array<StateNode<TC, TE>> = [];
   let marker = stateNode.parent;
 
+  if (!marker) {
+    return [stateNode];
+  }
```
Andarist, line 936: By always including the root here we don't allow it to be ever exited.

packages/core/src/stateUtils.ts
```diff
@@ -1578,7 +1582,8 @@ export function resolveMicroTransition<
     currentState.context,
     new Set(prevConfig),
     machine,
-    _event
+    _event,
+    new Set(!currentState._initial ? [] : [machine.root])
```
Andarist, line 1586: By moving `mutStatesToEnter` initialization to here we make it easier to add the root as a state that has to be entered during the fake~ initial transition. 

The usual expectation is that in a transition similar to this the `addAncestorStatesToEnter` should add all ancestors of the target state to `mutStatesToEnter` (if needed, ofc). However, this relies on `getProperAncestors` and that algorithm doesn't include the `toStateNode` in its result, so by using this logic alone we were not able to add the root state to the initial `mutStatesToEnter` and the root's invokes and entry actions rely on being included in this set for that initial transition.

SCXML doesn't account for this case in their algorithms because `<scxml/>` itself can't have `<invoke/>`, `<onentry/>`, etc. We are extending the spec by allowing those on our root state and that raises the need to make adjustments to the algorithms.

.changeset/purple-buses-hug.md
```diff
@@ -0,0 +1,8 @@
+---
+'xstate': patch
+---
+
+author: @Andarist
+author: @davidkpiano
+
+Make it impossible to exit a root state. For example, yhis means that root-level transitions specified as external transitions will no longer restart root-level invocations. See [#3072](https://github.com/statelyai/xstate/issues/3072) for more details.
```
davidkpiano, line 8:
```suggestion
Make it impossible to exit a root state. For example, this means that root-level transitions specified as external transitions will no longer restart root-level invocations. See [#3072](https://github.com/statelyai/xstate/issues/3072) for more details.
```

packages/core/src/stateUtils.ts
davidkpiano: This was going to be my main question - where this differs from the SCXML spec etc. Thanks for this!

davidkpiano (approved):
This LGTM; just a comment about the robustness of the test.

packages/core/test/invoke.test.ts
```diff
@@ -3146,6 +3146,41 @@ describe('invoke', () => {
...
+
+  // https://github.com/statelyai/xstate/issues/3072
+  it('root invocations should not restart on root external transitions', () => {
+    let count = 0;
+
+    const machine = createMachine({
+      id: 'root',
+      invoke: {
+        src: () =>
+          fromPromise(() => {
```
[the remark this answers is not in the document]
Andarist: I will add a test for entry/exit actions too 👍 Until we change how invocations are handled it's good to have the existing test in place too.

PR #3500, opened 2022-08-05
Fixed an issue with `useSelector` always computing fresh snapshots internally for uninitialized services

### LINKED ISSUE #3497: Bug: Accessing children within `useSelector()`  throws an error
### Description

_Original discussion:_ https://github.com/statelyai/xstate/discussions/3379

___

I followed the XState doc, [using a React context provider](https://xstate.js.org/docs/recipes/react.html#global-state-react-context) to prevent unnecessary re-renders.

My main state machine is invoking several child machines.

Here is how I access them:
```ts
const main = useContext(MyMainMachineContext);
const child = useSelector(main, state => state.children['child-machine-id']);
```

### Expected result

It should work without any error.

### Actual result

On the first `useSelector` call, the following error is thrown:
```
react-dom.development.js:86 Warning: The result of getSnapshot should be cached to avoid an infinite loop
    at Initialization (http://localhost:5173/src/pages/Initialization.tsx:22:26)
    at App (http://localhost:5173/src/App.tsx:22:16)
    at MachineProvider (http://localhost:5173/src/hooks/useMachine.tsx:22:3)
```

### Reproduction

https://github.com/jzecca/xstate-getsnapshot-reproducer

### Additional context

````
"@xstate/react": "^3.0.0",
"react": "^18.2.0",
"react-dom": "^18.2.0",
"xstate": "^4.32.1"
````

### PR DESCRIPTION
**Pull request by Andarist**
closes #3497

### DIFF
--- a/packages/xstate-react/src/useSelector.ts
+++ b/packages/xstate-react/src/useSelector.ts
@@ -11,6 +11,7 @@ function isService(actor: any): actor is Interpreter<any, any, any, any> {
 const defaultCompare = (a, b) => a === b;
 const defaultGetSnapshot = (a, initialStateCacheRef) => {
   if (isService(a)) {
+    // A status of 0 = interpreter not started
     if (a.status === 0 && initialStateCacheRef.current) {
       return initialStateCacheRef.current;
     }
--- a/packages/xstate-react/src/useSelector.ts
+++ b/packages/xstate-react/src/useSelector.ts
@@ -1,6 +1,6 @@
-import { useCallback } from 'react';
+import { useCallback, useRef } from 'react';
 import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/shim/with-selector';
-import { ActorRef, Interpreter, Subscribable } from 'xstate';
+import { ActorRef, AnyState, Interpreter, Subscribable } from 'xstate';
 import { isActorWithState } from './useActor';
 import { getServiceSnapshot } from './utils';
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/xstate-react/src/useSelector.ts [hunk 1]
```diff
@@ -9,12 +9,17 @@ function isService(actor: any): actor is Interpreter<any, any, any, any> {
...
+  if (isService(a)) {
+    if (a.status === 0 && initialStateCacheRef.current) {
+      return initialStateCacheRef.current;
+    }
+    const snapshot = getServiceSnapshot(a);
+    initialStateCacheRef.current = a.status === 0 ? snapshot : null;
+    return snapshot;
+  }
+  return isActorWithState(a) ? a.state : undefined;
+};
```
Andarist, line 22: Is it pretty? Not at all. It's basically a hack to cache those results as they should represent the same thing, even if they are not preferentially transparent. 

We can't easily refactor this in v4 because we can't compute the initial "true" snapshot when the service isn't initialized as computing this might lead to side-effects. This problem doesn't exist in v5

packages/xstate-react/src/useSelector.ts
```diff
@@ -9,12 +9,17 @@ function isService(actor: any): actor is Interpreter<any, any, any, any> {
...
 const defaultCompare = (a, b) => a === b;
-const defaultGetSnapshot = (a) =>
-  isService(a)
-    ? getServiceSnapshot(a)
-    : isActorWithState(a)
-    ? a.state
-    : undefined;
+const defaultGetSnapshot = (a, initialStateCacheRef) => {
+  if (isService(a)) {
+    if (a.status === 0 && initialStateCacheRef.current) {
```
davidkpiano, line 14:
```suggestion
    // A status of 0 = interpreter not started
    if (a.status === 0 && initialStateCacheRef.current) {
```

packages/xstate-react/test/useSelector.test.tsx
```diff
@@ -564,4 +571,32 @@ describeEachReactMode('useSelector (%s)', ({ suiteKey, render }) => {
...
+      });
+      return null;
+    }
+
+    console.error = jest.fn();
+    render(<App />);
+
+    const [snapshot1] = snapshots;
+    expect(snapshots.every((s) => s === snapshot1));
+    expect(console.error).toHaveBeenCalledTimes(0);
```
davidkpiano, line 600: I'd love to find a more robust/flexible way of specifying the logging/error sinks for v5 (future note, not for this PR)

davidkpiano (approved):

packages/xstate-react/test/useSelector.test.tsx
Andarist: Note that this doesn't test a log that we control. This one comes from React

PR #3506, opened 2022-08-08
docs: show proper state paths for the shopping machine example

maciekmaciej:


docs/guides/final.md
```diff
@@ -129,7 +129,7 @@ const shoppingMachine = createMachine({
 });
 ```
 
-The `onDone` transition will only take place when all of the child states of `'cart'` (e.g., `'user'` and `'items'`) are in their final states. In the case of the shopping machine, once the `'shopping.cart.user.success'` and `'shopping.cart.items.success'` state nodes are reached, the machine will transition from the `'cart'` to the `'confirm'` state.
+The `onDone` transition will only take place when all of the child states of `'cart'` (e.g., `'user'` and `'items'`) are in their final states. In the case of the shopping machine, once the `'#shopping.cart.user.success'` and `'#shopping.cart.items.success'` state nodes are reached, the machine will transition from the `'cart'` to the `'confirm'` state.
```
Andarist, line 132: Possible alternative:
```suggestion
The `onDone` transition will only take place when all of the child states of `'cart'` (e.g., `'user'` and `'items'`) are in their final states. In the case of the shopping machine, once the `'cart.user.success'` and `'cart.items.success'` state nodes are reached, the machine will transition from the `'cart'` to the `'confirm'` state.
```

The question here is - what do we mention in this text? Paths (like in my example)? Or implied/auto-generated IDs (like in your example)?

maciekmaciej: It was confusing for a discord user, since it wasn't clear that these are absolute paths.
I think you're right, the relative ones are better here

davidkpiano (approved):

Issue #3519, opened 2022-08-13
Bug: Sending an empty array of events to an interpreter fails with "Cannot use 'in' operator to search for '$$type' in undefined"


Woodz:
### Description

According to the docs of `Interpreter.send` (https://paka.dev/npm/xstate@4.33.1/api#dcfdac634e7e7494):

> An array of events (batched) can be sent as well, which will send all batched events to the running interpreter

However, this throws an error if the array is empty, forcing users to wrap all calls to `send` that use arrays of dynamic length in an array length check to avoid this bug.

### Expected result

I expected the interpreter to do nothing

### Actual result

It threw error `Cannot use 'in' operator to search for '$$type' in undefined` with stack trace:

```
    at Object.toSCXMLEvent (/var/task/dist/handlers.js:159314:41)
    at /var/task/dist/handlers.js:161157:41
    at Scheduler2.process (/var/task/dist/handlers.js:160435:11)
    at Scheduler2.schedule (/var/task/dist/handlers.js:160419:14)
    at Interpreter2.batch (/var/task/dist/handlers.js:161121:24)
    at Interpreter2.send (/var/task/dist/handlers.js:160689:19)
```

### Reproduction

https://codesandbox.io/s/keen-bell-peys02

### Additional context

XState version: 4.30.5

Andarist:
It's an easy fix - I will prepare it soon. However, note that in the next major version we plan to remove `service.batch` (which gets called by `service.send([])`). For that reason, I recommend you to migrate towards using this:
```diff
-service.send(events)
+events.forEach(event => service.send(event))
```

Woodz:
Makes sense! Will take note of this planned breaking change

PR #3520, opened 2022-08-13
Fixed a runtime error when sending an empty array of events

Andarist:
fixes https://github.com/statelyai/xstate/issues/3519

I'm not super sure that it's worth fixing this but 🤷‍♂️

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -759,18 +759,6 @@ export class Interpreter<
     }
 
     if (!events.length) {
-      const currentState = this.state;
-      const nextState: AnyState = new State({
-        ...currentState,
-        history: currentState,
-        actions: [],
-        transitions: []
-      });
-      nextState.changed = false;
-      if (nextState.history) {
-        delete nextState.history.history;
-      }
-      this._state = nextState;
       return;
     }
 
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -758,6 +758,22 @@ export class Interpreter<
       );
     }
 
+    if (!events.length) {
+      const currentState = this.state;
+      const nextState: AnyState = new State({
+        ...currentState,
+        history: currentState,
+        actions: [],
+        transitions: []
+      });
+      nextState.changed = false;
+      if (nextState.history) {
+        delete nextState.history.history;
+      }
+      this._state = nextState;
+      return;
+    }
+
     this.scheduler.schedule(() => {
       let nextState = this.state;
       let batchChanged = false;
--- a/.changeset/strange-bottles-punch.md
+++ b/.changeset/strange-bottles-punch.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Fixed a runtime crash when sending multiple events as an array to a service. It is not recommended to use this feature though as it will be removed in the next major version.


packages/core/src/interpreter.ts
```diff
@@ -758,6 +758,22 @@ export class Interpreter<
...
     }
 
+    if (!events.length) {
+      const currentState = this.state;
+      const nextState: AnyState = new State({
+        ...currentState,
+        history: currentState,
+        actions: [],
+        transitions: []
+      });
```
Andarist, line 768: note that somewhat weirdly this `nextState` will hold a reference to the `currentState._event` because we don't really have any new value to assign to it.

packages/core/src/interpreter.ts
```diff
@@ -758,6 +758,22 @@ export class Interpreter<
...
+      nextState.changed = false;
+      if (nextState.history) {
+        delete nextState.history.history;
+      }
+      this._state = nextState;
+      return;
+    }
```
davidkpiano, line 775: We should treat sending an empty array of events as not sending any event at all:

```suggestion
    if (!events.length) {
      return;
    }
```

Andarist: This would have a somewhat weird effect though:
```ts
service.send(arr)
if (service.state.changed) {
  // oops, we can't truly trust this `.changed` value here
}
```

davidkpiano: > This would have a somewhat weird effect though:
> 
> ```ts
> service.send(arr)
> if (service.state.changed) {
>   // oops, we can't truly trust this `.changed` value here
> }
> ```

This makes sense though because `service.state` is still the _previous_ state, which indeed hasn't changed. The state machine should only change state in response to events (mathematically), so no transition should happen and the state should be exactly the same.

Andarist: I'm talking about a reversed scenario - where `.changed === true` even though nothing has changed:
```js
const sendAndDoSmth = (events) => {
  service.send(events)
  if (service.state.changed) {
    // do smth
  }
}

sendAndDoSmth([{ type: 'FOO' }]) // this gets handles by the machine and thus a new state has `.changed: true`
sendAndDoSmth([]) // we didn't change the internal state at all, it still has `.changed: true`

// oops, we have executed the internal conditional logic twice, although the machine's truly only has changed once
```

davidkpiano: Yes, that's what I mean. `service.send([])` would _not_ cause a transition, which means that it would never result in a new state anyway.

You'd be reading `.changed === true` from the _previous_ state (before the empty array was sent), which is valid and accurate.

davidkpiano:
@Andarist To resolve, what do you think about `service.send([])` instead being a no-op (invalid)?

packages/core/src/interpreter.ts
Andarist: > You'd be reading .changed === true from the previous state (before the empty array was sent), which is valid and accurate.

I'm not sure if a common expectation after using `.send()` would be that you might get a *previous* state "back". This looks counter-intuitive to me.

Andarist:
> @Andarist To resolve, what do you think about service.send([]) instead being a no-op (invalid)?

I'd be OK with this - better to fix a crash than do nothing and `.batch/send([])` is going away anyway.

Issue #3522, opened 2022-08-15
Regression: root-level invoked services not stopped after reaching `final` state


with-heart:
### Description

When an interpreted machine reaches a `final` state, services invoked at the root level are no longer stopped.

#3126 changed the way the interpreter is stopped when `this.state.done`, calling the new `_stop` method where previously `stop` was called. `stop` handles stopping all `children`:

https://github.com/statelyai/xstate/blob/f2bf4924c291069204baab8b2f156574c0bd085f/packages/core/src/interpreter.ts#L664-L668

### Expected result

When a service reaches a `final` state, all invoked services are stopped.

### Actual result

When a service reaches a `final` state, services invoked at the root are not stopped.

### Reproduction

https://codesandbox.io/s/kind-diffie-ewx8qk?file=/src/index.ts

### Additional context

Reproduction of previous behavior (`xstate@4.32.1`): https://codesandbox.io/s/modern-rain-4serhm

Andarist:
Thank you for the report, It got fixed by https://github.com/statelyai/xstate/pull/3523 and the new release is already out.

with-heart:
@Andarist You're the best! Ty 💖

PR #3523, opened 2022-08-15
Fixed a regression that caused child actors not being correctly stopped when their parent reached a final state

Andarist:
https://github.com/statelyai/xstate/issues/3522

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -369,6 +369,7 @@ export class Interpreter<
         listener(doneInvoke(this.id, doneData));
       }
       this._stop();
+      this._stopChildren();
     }
   }
   /*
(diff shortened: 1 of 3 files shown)


packages/core/src/interpreter.ts
```diff
@@ -369,6 +369,7 @@ export class Interpreter<
         listener(doneInvoke(this.id, doneData));
       }
       this._stop();
+      this._stopChildren();
```
Andarist, line 372: You might wonder why this can't be part of the `_stop` call as `_stop` gets called both:
- when reaching a final state
- when a machine gets stopped from the outside

The problem is that when `.stop()` (external call) gets called we first call `_stop` and then schedule an `xstate.stop` event. In this case, we should only stop children when we process that `xstate.stop` event and not immediately. This is mainly because exit actions should be able to send events to children. If we'd clear children in `_stop` then we would make this impossible.

davidkpiano (approved):
LGTM, not sure that the test fully tests the timing of the event though

PR #3526, opened 2022-08-16
Remove `@vue/vue3-jest` workaround for its `tsconfig` loading logic

Andarist:
⚠️ this currently doesn't work as it requires us to also update Jest itself. I don't want to do it right now, I'm just creating this PR as a draft and a reminder to get to this eventually.


Andarist:
I incorporated this change into https://github.com/statelyai/xstate/pull/3766

PR #3540, opened 2022-08-22
Fixed an issue that caused `invoke`d actors to be created before resolving `assign` actions from `entry` of the same state

Andarist:
Changes to the algorithm:
1. Instead of flattening the actions in `getActions` into an array I now group them in "blocks".
2. The main purpose for this is to have `invoke` actions and `entry` actions grouped together
3. This allows me to defer `send` actions to the end of the current block and flush them then

### DIFF
--- a/.changeset/gold-icons-protect.md
+++ b/.changeset/gold-icons-protect.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Fixed an issue that caused `invoke`d actors to be created before resolving `assign` actions from `entry` of the same state.
--- a/.changeset/gold-icons-protect.md
+++ b/.changeset/gold-icons-protect.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Fixed an issue that caused `invoke`d actors to be created before resolving `assign` actions from `entry` of the same state.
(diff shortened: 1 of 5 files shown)


.changeset/gold-icons-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Fixed an issue that caused `invoke`d actors to be created before resolving `assign` actions from `entry` of the same state.
```
Andarist, line 5:
```suggestion
Fixed an issue that caused `invoke`d actors to be created before resolving `assign` actions from `entry` of the same state when using `predictableActionArguments` flag.
```

davidkpiano (approved):
Can we document this behavior, if it differs from the SCXML algorithm (which I suspect it does)?

Andarist:
> Can we document this behavior, if it differs from the SCXML algorithm (which I suspect it does)?

Where I would document this? I'm not exactly sure if this differs from the SCXML. The spec itself doesn't mention those semantics at all as far as I can tell. The algorithm at the bottom is just an informative section (not normative) and it also doesn't mention those things explicitly. However, I certainly see how I could read from it different semantics. 

I'm gonna land this to get things going - but if we figure out where to document this I will happily follow up with a PR to add a note about this.

PR #3541, opened 2022-08-22
Fixed an issue with not being able to read the updated snapshot of a child when receiving and processing events from it and when using `predictableActionArguments` flag

### LINKED ISSUE #3533: Bug: `sendUpdate()` not updating parent actor with `predictableActionArguments`
### Description

Hello XState team 👋 

My team was working on updating our machine to use `predictableActionArguments`, but found `sendUpdate` to no longer work. This happened when child actor tried to update parent after promise invocation with `onError`. We could confirm this only happens when we set `predictableActionArguments` to `true`.

Please see the [minimal repro](https://codesandbox.io/s/xstate-4-33-2-sendupdate-error-cpjxod?file=/src/actors/child.js) for reference.

### Expected result

`sendUpdate()` should update the parent, and trigger a re-render on `useMachine(...)`.

### Actual result

No re-render happens after `childMachine` makes `onError` transition with `sendUpdate()` action.

### Reproduction

https://codesandbox.io/s/xstate-4-33-2-sendupdate-error-cpjxod?file=/src/actors/child.js

Child actor goes through `idle` -> `submit` -> `idle`, but the parent machine gets updated only for the first two states.

### Additional context

I could confirm that adding `{sync: true}` resolves the error, but we want to avoid this for optimization purposes.

Andarist:
fixes #3533

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -208,6 +208,10 @@ export class Interpreter<
   public children: Map<string | number, ActorRef<any>> = new Map();
   private forwardTo: Set<string> = new Set();
 
+  private _outgoingQueue: Array<
+    [{ send: (ev: unknown) => void }, unknown]
+  > = [];
+
   // Dev Tools
   private devTools?: any;
 
@@ -324,6 +328,11 @@ export class Interpreter<
       this.options.execute
     ) {
       this.execute(this.state);
+    } else {
+      let item: typeof this._outgoingQueue[number] | undefined;
+      while ((item = this._outgoingQueue.shift())) {
+        item[0].send(item[1]);
+      }
     }
 
     // Update children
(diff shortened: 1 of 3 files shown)


packages/core/src/interpreter.ts
```diff
@@ -208,6 +208,10 @@ export class Interpreter<
   public children: Map<string | number, ActorRef<any>> = new Map();
   private forwardTo: Set<string> = new Set();
 
+  private _outgoingQueue: Array<
+    [{ send: (ev: unknown) => void }, unknown]
```
davidkpiano, line 212: I'm confused about this tuple; what's the `unknown` for?

packages/core/test/predictableExec.test.ts
```diff
@@ -624,4 +624,43 @@ describe('predictableExec', () => {
...
+      invoke: {
+        id: 'myChild',
+        src: child
+      }
+    });
+
+    const service = interpret(machine).start();
+
+    service.subscribe((state) => {
+      if (shouldCheck) {
```
davidkpiano, line 660: Nit: I would add a `pass` and `fail` state to the `machine` and have a guarded transition on `CHILD_UPDATED` that goes to `pass` or `fail` depending on the snapshot value of the child, but that can be done as a refactor

davidkpiano (approved):
LGTM, I'm currently thinking about the best way to implement this in the v5 interpret branch (currently fails because sending to actors is eager, but we should treat it as "async" conceptually anyway)

packages/core/test/predictableExec.test.ts
Andarist: changed

packages/core/src/interpreter.ts
```diff
@@ -840,16 +849,25 @@ export class Interpreter<
...
-        (target as AnyInterpreter).send({
+        const scxmlEvent = {
           ...event,
           name:
             event.name === actionTypes.error ? `${error(this.id)}` : event.name,
           origin: this.sessionId
-        });
+        };
+        if (this.machine.config.predictableActionArguments) {
+          this._outgoingQueue.push([target, scxmlEvent]);
```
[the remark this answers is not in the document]
Andarist: ye, this could be simplified like that but I've wanted to avoid a temporary closure

PR #3545, opened 2022-08-25
allow action `type` strings in `pure` return array

with-heart:
This PR allows action `type` strings to be returned from the `getActions` argument of `pure`. `xstate` already handles action strings correctly but the types don't match that functionality.

This PR pairs well with statelyai/xstate-tools#207

### DIFF
--- a/.changeset/seven-ducks-look.md
+++ b/.changeset/seven-ducks-look.md
@@ -1,0 +1,24 @@
+---
+'xstate': patch
+---
+
+pr: #3545
+author: @with-heart
+
+Updated `pure` action types to allow action `type` strings to be returned in the array.
+
+```ts
+const machine = createMachine(
+  {
+    entry: ['doStuff']
+  },
+  {
+    actions: {
+      doStuff: pure(() => ['someAction']),
+      someAction: () => console.log('executed by doStuff')
+    }
+  }
+);
+```
+
+Returning action `type` strings were already handled by `xstate` and the types now correctly reflect that.
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -544,7 +544,9 @@ export function pure<TContext, TEvent extends EventObject>(
     context: TContext,
     event: TEvent
   ) =>
-    | SingleOrArray<ActionObject<TContext, TEvent> | TEvent['type']>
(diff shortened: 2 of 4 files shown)


packages/core/src/actions.ts
```diff
@@ -543,7 +543,9 @@ export function pure<TContext, TEvent extends EventObject>(
   getActions: (
     context: TContext,
     event: TEvent
-  ) => SingleOrArray<ActionObject<TContext, TEvent>> | undefined
+  ) =>
+    | SingleOrArray<ActionObject<TContext, TEvent> | TEvent['type']>
```
Andarist, line 547: This is incorrect - you are saying here that it's allowed to return the **event**'s type as an action type. This should allow returning `TAction['type']`

.changeset/seven-ducks-look.md
```diff
@@ -0,0 +1,24 @@
+---
+'xstate': patch
+---
+
+pr: #3545
+author: @with-heart
```
Andarist, line 6: just a note: those are only needed if the desired information doesn't match what can be automatically inferred. For example, I use those when I add changesets after landing a relevant PR.

packages/core/src/actions.ts
with-heart: Wow that's a major derp moment 😅 Good call

with-heart: Switched it to `ActionObject<TContext, TEvent>['type']` since we don't have `TAction` in this case

Andarist: In such a case - you could simplify this to just `string` but perhaps this gives more context behind what that `string` is.

.changeset/seven-ducks-look.md
with-heart: @Andarist is that a `@changesets/changelog-github` feature?

Andarist: > @Andarist is that a @changesets/changelog-github feature?

Yeah, it queries GitHub through the API to infer this info on its own. You only need to supply this manually if you want to override what would be inferred (for whatever reason)

davidkpiano (approved):
This should be good to go (cc. @Andarist)

Issue #3548, opened 2022-08-25
Bug: invoked callback not working with predictableActionArguments


customcommander:
### Description

The "ping pong" machine example doesn't work when `predictableActionArguments` is set to `true`:
https://xstate.js.org/docs/guides/communication.html#invoking-callbacks.

I copy/pasted (and instrumented the code):

```javascript
const { createMachine, interpret, send } = require('xstate');

const pingPongMachine = createMachine({
  predictableActionArguments: true,
  id: 'pinger',
  initial: 'active',
  states: {
    active: {
      invoke: {
        id: 'ponger',
        src: (context, event) => (callback, onReceive) => {
          console.log('ponger init')
          // Whenever parent sends 'PING',
          // send parent 'PONG' event
          onReceive((e) => {
            if (e.type === 'PING') {
              console.log('ponger done')
              callback('PONG');
            }
          });
        }
      },
      entry: send({ type: 'PING' }, { to: 'ponger' }),
      on: {
        PONG: { target: 'done' }
      }
    },
    done: {
      type: 'final'
    }
  }
});

interpret(pingPongMachine)
  .onDone(() => console.log('done!'))
  .start();
```

### Expected result

Expected output:

```
ponger init
ponger done
done!
```

### Actual result

Actual output:

```
ponger init
```

### Reproduction

https://codesandbox.io/s/lucid-gauss-hnuq18?file=/src/index.js

### Additional context

If you set `predictableActionArguments` to `false` then we get the expected output.

This was observed in v4.33.3.

In v4.33.2 we get the expected output regardless of the value of `predictableActionArguments`.

PR #3549, opened 2022-08-26
Fixed an issue with not being able to send events to initially started child actors when using `predictableActionArguments`

Andarist:
fixes https://github.com/statelyai/xstate/issues/3548

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -810,7 +810,8 @@ export class Interpreter<
 
   private sendTo = (
     event: SCXML.Event<AnyEventObject>,
-    to: string | number | ActorRef<any>
+    to: string | number | ActorRef<any>,
+    immediate: boolean
   ) => {
     const isParent =
       this.parent && (to === SpecialTargets.Parent || this.parent.id === to);
@@ -855,7 +856,7 @@ export class Interpreter<
             event.name === actionTypes.error ? `${error(this.id)}` : event.name,
           origin: this.sessionId
         };
-        if (this.machine.config.predictableActionArguments) {
+        if (!immediate && this.machine.config.predictableActionArguments) {
           this._outgoingQueue.push([target, scxmlEvent]);
         } else {
           (target as AnyInterpreter).send(scxmlEvent);
@@ -863,7 +864,7 @@ export class Interpreter<
       }
     } else {
       // Send normal events to other targets
-      if (this.machine.config.predictableActionArguments) {
+      if (!immediate && this.machine.config.predictableActionArguments) {
         this._outgoingQueue.push([target, event.data]);
       } else {
         target.send(event.data);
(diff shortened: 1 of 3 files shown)


packages/core/src/interpreter.ts
```diff
@@ -810,7 +810,8 @@ export class Interpreter<
 
   private sendTo = (
     event: SCXML.Event<AnyEventObject>,
-    to: string | number | ActorRef<any>
+    to: string | number | ActorRef<any>,
+    immediate: boolean
```
Andarist, line 814: The problem here was that executing initial actions is still done using `Interpreter#execute` - and doesn't happen from within `StateMachine#transition`:
https://github.com/statelyai/xstate/blob/0f1328c9c53c1409ee91ce00e99299643e054e6e/packages/core/src/interpreter.ts#L322-L336

So initial send actions were never sent out after my recent PR: https://github.com/statelyai/xstate/pull/3540

packages/core/src/interpreter.ts
```diff
@@ -991,7 +992,7 @@ export class Interpreter<
           return;
         } else {
           if (sendAction.to) {
-            this.sendTo(sendAction._event, sendAction.to);
+            this.sendTo(sendAction._event, sendAction.to, _event === initEvent);
```
davidkpiano, line 995: Was the problem that it was not sent immediately, or not sent at all?

packages/core/test/predictableExec.test.ts
```diff
@@ -688,4 +688,39 @@ describe('predictableExec', () => {
...
+        },
+        done: {
+          type: 'final'
+        }
+      }
+    });
+
+    const service = interpret(machine).start();
+
+    expect(service.getSnapshot().value).toBe('done');
```
davidkpiano, line 724: In real-life scenarios, we can't assume that this is synchronous. Will the `machine` still get the `PONG` event _after_ it settles on the initial state (it should)? 

Maybe we can test this by instead listening for `.onDone(done)` so we don't depend on it being sync 🤔

packages/core/src/interpreter.ts
Andarist: they would be flushed within the next `service.send(ev)` - which could happen at any time, or not at all.

packages/core/test/predictableExec.test.ts
Andarist: > In real-life scenarios, we can't assume that this is synchronous. Will the machine still get the PONG event after it settles on the initial state (it should)?

Yes, in this case, the test case is really more about the child receiving an event than about the parent receiving the response - I could slim it down.

> Maybe we can test this by instead listening for .onDone(done) so we don't depend on it being sync 🤔

I agree that those conceptually are async. However, I think that we also need to test synchronous logic as sync code often exposes more edge cases in scheduling etc. So it feels more beneficial to prefer sync tests in our test suite when possible - despite the mental model being async.

davidkpiano (approved):

PR #3551, opened 2022-08-28
[v5] Algorithm cleanup

davidkpiano:


packages/core/src/State.ts
```diff
@@ -96,39 +93,50 @@ export class State<
     TEvent extends EventObject = EventObject
   >(
     stateValue: State<TContext, TEvent, any> | StateValue,
-    context: TContext = {} as TContext
+    context: TContext = {} as TContext,
+    machine: AnyStateMachine
```
davidkpiano, line 97: States are now required to come from a `machine`.

packages/core/src/State.ts
```diff
@@ -137,70 +145,57 @@ export class State<
...
     this.transitions = config.transitions;
     this.children = config.children;
-    this.tags = config.configuration
-      ? getTagsFromConfiguration(config.configuration)
-      : config.tags
-      ? new Set(config.tags)
-      : new Set();
-    this.machine = config.machine;
+
+    this.value = getStateValue(machine.root, config.configuration);
```
davidkpiano, line 198: Value is derived from `configuration` to prevent impossible value/configuration mismatch.

packages/core/src/State.ts
```diff
@@ -316,6 +311,32 @@ export class State<
...
   public get nextEvents(): Array<TEvent['type']> {
-    return memo(this, 'nextEvents', () => nextEvents(this.configuration));
+    return memo(this, 'nextEvents', () => {
+      return [
+        ...new Set(flatten([...this.configuration.map((sn) => sn.ownEvents)]))
+      ];
+    });
+  }
+
+  public get meta() {
```
davidkpiano, line 321: `state.meta` is derived from the `configuration`

packages/core/src/State.ts
```diff
@@ -316,6 +311,32 @@ export class State<
...
+    return this.configuration.reduce((acc, stateNode) => {
+      if (stateNode.meta !== undefined) {
+        acc[stateNode.id] = stateNode.meta;
+      }
+      return acc;
+    }, {} as Record<string, any>);
+  }
+
+  public get tags(): Set<string> {
```
davidkpiano, line 330: `state.tags` is derived from the `configuration`

packages/core/src/StateNode.ts
```diff
@@ -337,7 +337,10 @@ export class StateNode<
     return this.events.includes(eventType);
   }
 
-  public next(state: State<TContext, TEvent>, _event: SCXML.Event<TEvent>) {
+  public next(
+    state: State<TContext, TEvent>,
+    _event: SCXML.Event<TEvent>
+  ): TransitionDefinition<TContext, TEvent>[] | undefined {
```
davidkpiano, line 343: Just made the return type explicit here

packages/core/src/actions/stop.ts
```diff
@@ -18,7 +18,10 @@ export function stop<
   TContext extends MachineContext,
   TEvent extends EventObject
 >(
-  actorRef: string | Expr<TContext, TEvent, ActorRef<any> | string>
+  actorRef:
+    | string
+    | ActorRef<any>
```
davidkpiano, line 23: Now you can directly specify a `stop(actorRef)`, which is useful in the internal algorithm

packages/core/src/index.ts
```diff
@@ -50,7 +49,6 @@ export {
   interpret,
   Interpreter,
   InterpreterStatus,
-  matchState,
```
davidkpiano, line 53: This should have been removed a long time ago

packages/core/src/types.ts
```diff
@@ -1581,6 +1587,7 @@ export interface StateConfig<
   done?: boolean;
   tags?: Set<string>;
   machine?: StateMachine<TContext, TEvent, any, any, any>;
+  _internalQueue?: Array<SCXML.Event<TEvent>>;
```
davidkpiano, line 1590: Only used to queue up an `error.execution` event if the `microstep` procedure fails

PR #3555, opened 2022-08-29
[core] Allow actor names in `sendTo(...)`

davidkpiano:
The `sendTo(actorName, event)` action creator now accepts a string `actorName`:

```js
actions: sendTo('someNamedActor', { type: 'someEvent' })
```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1314,7 +1314,10 @@ export enum SpecialTargets {
 export interface SendActionOptions<TContext, TEvent extends EventObject> {
   id?: string | number;
   delay?: number | string | DelayExpr<TContext, TEvent>;
-  to?: string | ExprWithMeta<TContext, TEvent, string | number | ActorRef<any>>;
+  to?:
+    | string
+    | ActorRef<any>
+    | ExprWithMeta<TContext, TEvent, string | number | ActorRef<any>>;
 }
 
 export interface CancelAction extends ActionObject<any, any> {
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -1314,7 +1314,10 @@ export enum SpecialTargets {
 export interface SendActionOptions<TContext, TEvent extends EventObject> {
   id?: string | number;
   delay?: number | string | DelayExpr<TContext, TEvent>;
-  to?: string | ExprWithMeta<TContext, TEvent, string | number | ActorRef<any>>;
+  to?:
+    | string
+    | ActorRef<any>
+    | ExprWithMeta<TContext, TEvent, string | number | ActorRef<any>>;
```
Andarist, line 1320:
```suggestion
    | string
    | ActorRef<any>
    | ExprWithMeta<TContext, TEvent, string | ActorRef<any>>;
```

packages/core/test/actions.test.ts
```diff
@@ -2254,6 +2254,66 @@ describe('sendTo', () => {
...
+      }
+    });
+
+    const parentMachine = createMachine<{
+      child: ActorRefFrom<typeof childMachine>;
+    }>({
+      context: () => ({
+        child: spawn(childMachine, 'child')
+      }),
+      // No type-safety for the event yet
```
Andarist, line 2278: 😢

Andarist (approved):

packages/core/test/actions.test.ts
davidkpiano: We'll get there!

Issue #3557, opened 2022-08-30
Bug: Compilation fails on TypeScript 4.8.2


irv075:
### Description

Compilation fails when compiling with TypeScript v4.8.2

Error message:

```
node_modules/xstate/lib/interpreter.d.ts:181:122 - error TS2344: Type 'TChildStateSchema' does not satisfy the constraint 'StateSchema<any>'.

181     spawnMachine<TChildContext, TChildStateSchema, TChildEvent extends EventObject>(machine: StateMachine<TChildContext, TChildStateSchema, TChildEvent>, options?: { 
                                                                                                                             ~~~~~~~~~~~~~~~~~

  node_modules/xstate/lib/interpreter.d.ts:181:33
    181     spawnMachine<TChildContext, TChildStateSchema, TChildEvent extends EventObject>(machine: StateMachine<TChildContext, TChildStateSchema, TChildEvent>, options?: {
                                        ~~~~~~~~~~~~~~~~~
    This type parameter might need an `extends StateSchema<any>` constraint.

node_modules/xstate/lib/types.d.ts:231:62 - error TS2344: Type 'TStateSchema["states"][K]' does not satisfy the constraint 'StateSchema<any>'.
  Type 'TStateSchema["states"][keyof TStateSchema["states"]]' is not assignable to type 'StateSchema<any>'.
    Type 'TStateSchema["states"][string] | TStateSchema["states"][number] | TStateSchema["states"][symbol]' is not assignable to type 'StateSchema<any>'.
      Type 'TStateSchema["states"][string]' is not assignable to type 'StateSchema<any>'.

231     [K in keyof TStateSchema['states']]: StateNode<TContext, TStateSchema['states'][K], TEvent>;
                                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/xstate/lib/types.d.ts:234:68 - error TS2344: Type 'TStateSchema["states"][K]' does not satisfy the constraint 'StateSchema<any>'.
  Type 'TStateSchema["states"][keyof TStateSchema["states"]]' is not assignable to type 'StateSchema<any>'.
    Type 'TStateSchema["states"][string] | TStateSchema["states"][number] | TStateSchema["states"][symbol]' is not assignable to type 'StateSchema<any>'.
      Type 'TStateSchema["states"][string]' is not assignable to type 'StateSchema<any>'.

[...]

PR #3559, opened 2022-08-30
Upgrade TypeScript version used in the repo

Andarist:
fixes https://github.com/statelyai/xstate/issues/3557

The applied `NonNullable` is at the wrong level (I think?) but, for some reason, it works - I'm still rechecking why. It should be applied on `TStateSchema['states']` and not on `TStateSchema['states'][K]`. However, this doesn't work because there is a bug In TypeScript. I've already created a fix for it but I'm not sure how long will it take to land it: https://github.com/microsoft/TypeScript/pull/50540

Let's wait a couple of days to see what happens with that PR. In the meantime, we can recommend using `skipLibCheck: true` as a workaround.


packages/core/src/types.ts
```diff
@@ -409,7 +409,7 @@ export type StateNodesConfig<
 > = {
   [K in keyof TStateSchema['states']]: StateNode<
     TContext,
-    TStateSchema['states'][K],
+    TStateSchema['states'][K] & {},
```
Andarist, line 412: Those `& {}` here ensure that TS understands that this type argument will be at least non-nullable (I didn't use `NonNullable` because in older versions of TS it was a conditional type and it could complicate assignability rules there). It also makes the whole intersection have `{}` or `StateSchema` constraint and both are assignable to `StateSchema` that is expected on the target side here (the required constraint within `StateNode`).

Perhaps a more clean solution would be to use `NonNullable` around `TStateSchema['states']` but that doesn't currently work in TS. I have prepared a bug fix for it.

However, all of this is related to `TStateSchema` so we won't even need that bug fix to land - we can go with this workaround here, n v4. In v5 we have already removed `TStateSchema` so this will just be removed when merging this to that branch

packages/core/src/StateNode.ts
```diff
@@ -1629,16 +1629,17 @@ class StateNode<
         (stateNode) => !(stateNode.type === 'history')
       );
     } else if (this.initial !== undefined) {
-      if (!this.states[this.initial]) {
+      if (!this.states[this.initial as string]) {
```
Andarist, line 1632: TS became more conservative when it comes to `keyof T` in generic contexts. After all `keyof T`'s constraint is `PropertyKey` and that is equivalent to `string | number | symbol`. We cast this to `string` as that is what it's going to be in practice anyway.

packages/core/src/interpreter.ts
```diff
@@ -1183,7 +1183,7 @@ export class Interpreter<
   }
   public spawnMachine<
     TChildContext,
-    TChildStateSchema,
+    TChildStateSchema extends StateSchema,
```
Andarist, line 1186: This was a bug in the code. Since TS became more conservative now, when it comes to such unconstrained generics, it started to complain about this.

packages/core/test/types.test.ts
```diff
@@ -440,7 +440,7 @@ describe('context', () => {
   });
 
   it('should work with generic context', () => {
-    function createMachineWithExtras<TContext>(
+    function createMachineWithExtras<TContext extends {}>(
```
Andarist, line 443: It probably was a slight bug in the past and it got uncovered by the new, more conservative, rules around unconstrained generics. To fix some inference problems we have introduced `LowInfer` here:
https://github.com/statelyai/xstate/blob/9f78588ead2361205e1c8318930d3eab392b33df/packages/core/src/types.ts#L983

We didn't really want to constrain the `TContext` to be a subtype of `{}` but that's the effect with the current version of TS. In practice, this shouldn't be a big problem because we only accept object contexts at runtime. Some people might be forced to add `extends {}` in their generic declarations, just like I've done here, but that's just about it.

PR #3561, opened 2022-08-30
add support for action groups

with-heart:
Per discussion in #3554, this PR adds support for action groups.

This allows us to execute named (and possibly nested) groups of actions together:

```ts
  {
    entry: 'group',
  },
  {
    actions: {
      group: ['action1', 'group2'],
      group2: ['action2']
      action1: () => console.log('action1'),
      action2: () => console.log('action2'),
    },
  },
)

interpret(machine).start()

// "action1", "action2"
```

### DIFF
--- a/.changeset/wicked-pandas-add.md
+++ b/.changeset/wicked-pandas-add.md
@@ -2,25 +2,33 @@
 'xstate': minor
 ---
 
-An action can now be defined as an "action group", which is an array of action `type` strings pointing to other actions that will be executed when the group is executed. This allows you to define named groups of actions that are executed together.
+An action can now be defined as an "action group" which is an array of actions that will be executed when the group action is executed. This allows you to define named groups of actions that are executed together.
 
 ```js
 const machine = createMachine(
   {
+    context: { value: '' },
     on: {
-      event: { actions: 'group' }
+      valueUpdated: { actions: 'updateValue' }
     }
   },
   {
     actions: {
-      group: ['action1', 'action2'],
-      action1: () => console.log('action 1'),
-      action2: () => console.log('action 2')
+      updateValue: [
+        assign({
+          value: (_context, event) => event.value
+        }),
+        'capitalizeValue',
+        (context) => console.log(`Value: ${context.value}`)
+      ],
+      capitalizeValue: assign({
+        value: (context) => context.value.toUpperCase()
+      })
     }
   }
 );
 
 const service = interpret(machine).start();
 
-service.send(service, 'event'); // "action 1", "action 2"
+service.send({ type: 'valueUpdated', value: 'hello' }); // logs "Value: HELLO"
 ```
--- a/.changeset/wicked-pandas-add.md
+++ b/.changeset/wicked-pandas-add.md
@@ -1,0 +1,26 @@
+---
+'xstate': minor
+---
+
+An action can now be defined as an "action group", which is an array of action `type` strings pointing to other actions that will be executed when the group is executed. This allows you to define named groups of actions that are executed together.
+
+```js
+const machine = createMachine(
+  {
+    on: {
+      event: { actions: 'group' }
+    }
+  },
+  {
+    actions: {
+      group: ['action1', 'action2'],
+      action1: () => console.log('action 1'),
+      action2: () => console.log('action 2')
+    }
+  }
+);
+
+const service = interpret(machine).start();
+
+service.send(service, 'event'); // "action 1", "action 2"
+```
--- a/docs/guides/actions.md
+++ b/docs/guides/actions.md
@@ -214,27 +214,35 @@ entry: send({ type: 'SOME_EVENT' });
 
 ## Action groups
 
-An action can be defined as an "action group", which is an array of action `type` strings pointing to other actions that will be executed when the group is executed. This allows you to define named groups of actions that are executed together.
+An action can be defined as an "action group" which is an array of actions that will be executed when the group action is executed. This allows you to define named groups of actions that are executed together.
 
 ```js
 const machine = createMachine(
   {
+    context: { value: '' },
     on: {
-      event: { actions: 'group' }
+      valueUpdated: { actions: 'updateValue' }
     }
   },
   {
     actions: {
-      group: ['action1', 'action2'],
-      action1: () => console.log('action 1'),
-      action2: () => console.log('action 2')
+      updateValue: [
+        assign({
+          value: (_context, event) => event.value
+        }),
+        'capitalizeValue',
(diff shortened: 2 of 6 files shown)


with-heart:
I should probably add the docs in a separate PR so we can add a version `Badge`

packages/core/test/actions.test.ts
```diff
@@ -1682,6 +1682,32 @@ describe('purely defined actions', () => {
   });
 });
 
+describe('action groups', () => {
+  it('should execute actions provided as type strings in a group', () => {
```
davidkpiano, line 1686: Awesome! Can you also add tests to support:

- Mixture of strings, objects, functions
- An action group that references another action that references that action group (edge-case)
- Some more edge-cases I can't think of right now

.changeset/wicked-pandas-add.md
```diff
@@ -0,0 +1,26 @@
...
+      group: ['action1', 'action2'],
+      action1: () => console.log('action 1'),
+      action2: () => console.log('action 2')
+    }
+  }
+);
+
+const service = interpret(machine).start();
+
+service.send(service, 'event'); // "action 1", "action 2"
```
Andarist, line 25:
```suggestion
service.send({ type: 'event' }); // logs: "action 1" and "action 2"
```

docs/guides/actions.md
```diff
@@ -212,6 +212,57 @@ entry: send({ type: 'SOME_EVENT' });
...
+      group: ['action1', 'action2'],
+      action1: () => console.log('action 1'),
+      action2: () => console.log('action 2')
+    }
+  }
+);
+
+const service = interpret(machine).start();
+
+service.send(service, 'event'); // "action 1", "action 2"
```
Andarist, line 237:
```suggestion
service.send({ type: 'event' }); // logs: "action 1" and "action 2"
```

packages/core/src/actionTypes.ts
```diff
@@ -19,3 +19,4 @@ export const error = ActionTypes.ErrorCustom;
 export const update = ActionTypes.Update;
 export const choose = ActionTypes.Choose;
 export const pure = ActionTypes.Pure;
+export const actionGroup = ActionTypes.ActionGroup;
```
Andarist, line 22: I wouldn't introduce this one. I feel like arrays are already pretty descriptive when it comes to such groups and we can use them directly.

packages/core/src/types.ts
```diff
@@ -788,7 +790,8 @@ type MachineOptionsActions<
         TContext,
         Cast<Prop<TIndexedEvents, TEventsCausingActions[K]>, EventObject>,
         Cast<Prop<TIndexedActions, K>, BaseActionObject>
-      >;
+      >
+    | Array<Cast<Prop<TIndexedActions, K>, BaseActionObject>['type']>;
```
Andarist, line 794: We probably should accept more things here. For instance those should be valid:
```
group: [assign(), 'another']
group: [assign(), sendParent()]
group: [assign(), { type: 'another' }]
```

with-heart: I'm working on shifting it to:

```ts
{
  [K in keyof TEventsCausingActions]?:
    | ActionObject<
        TContext,
        Cast<Prop<TIndexedEvents, TEventsCausingActions[K]>, EventObject>
      >
    | ActionFunction<
        TContext,
        Cast<Prop<TIndexedEvents, TEventsCausingActions[K]>, EventObject>,
        Cast<Prop<TIndexedActions, K>, BaseActionObject>
      >
    | Array<
        | ActionObject<
            TContext,
            Cast<Prop<TIndexedEvents, TEventsCausingActions[K]>, EventObject>
          >
        | ActionFunction<
            TContext,
            Cast<Prop<TIndexedEvents, TEventsCausingActions[K]>, EventObject>,
            Cast<Prop<TIndexedActions, K>, BaseActionObject>
          >
        | Cast<Prop<TIndexedActions, K>, BaseActionObject>['type']
      >;
}
```

It'll accept objects and functions though it still leaves us with the same type problems as `pure`. At least they're consistent 😆

with-heart: Also thinking I should change `Cast<Prop<TIndexedActions, K>, BaseActionObject>['type']` to `keyof TEventsCausingActions`

- autocompletion of `type` strings
- the typegen portion of this automatically adds `type` strings in groups to `eventsCausingActions`, so it wouldn't do anything for you on the first use of a `type`, but if you used it in later groups you'd get autocomplete

Think there'd be any downsides to that?

packages/core/src/actionTypes.ts
with-heart: I added this following @davidkpiano's comment about action groups being syntactic sugar for `xstate.actionGroup` over here: https://github.com/statelyai/xstate/discussions/3554#discussioncomment-3499861

If we didn't use an `ActionObject` to describe these, would we just flatten the group arrays in `toActionObjects`?

packages/core/test/actions.test.ts
with-heart: Added some tests for groups of each type as well as a mixture

> An action group that references another action that references that action group (edge-case)

Would we just test that this throws a maximum call stack exceeded error for the time being? 

Was playing around with a way to track and prevent cyclical action calls last night because `pure` and `choose` can suffer from the same problem, but it feels pretty thorny

with-heart:
Updated the example in docs/changeset to show a mix of `type` strings, objects, and functions

packages/core/test/actions.test.ts
davidkpiano: > Would we just test that this throws a maximum call stack exceeded error for the time being?

While creating the action, you can have some variable in closure (like a `Set`) that keeps track of actions already added to the group, and if it encounters it again (`set.has(someActionType)`), you can detect that cycle.

with-heart: Just want to make sure I'm understanding. Are you thinking this `Set` would be within a cycle test or in the actual codebase?

davidkpiano:
I still love the idea of this; what is needed to get it through?

with-heart:
@davidkpiano hoping to shift back to working on `xstate` stuff this week. This time with a digital garden so I can document all the things as I do them!

Need to spend some time going through this again and remembering where things left off 😆

davidkpiano:
@with-heart Can we re-target this for v5? (`next` branch)

with-heart:
> @with-heart Can we re-target this for v5? (`next` branch)

Yeeee, I'll get conflicts resolved and make sure it's working again

with-heart:
It was easier to just reimplement this from scratch than deal with the rebase lol. Will reopen from a new branch shortly

PR #3566, opened 2022-09-02
add failing test cases for invoked observable listeners

with-heart:
This is a draft PR demonstrating two possible bugs in the current `spawnObservable` implementation. I noticed both of these as a side effect of upgrading `jest` in #3565.

In that PR, the existing test for `state.children should reference invoked child actors (observable)` passes but leaves an open handle.

It seems like there are potentially two bugs here:

1. The invoked observable child and `service` are both stopped before a child listener receives its latest emitted value. This seems like a mismatch in behavior e.g. a listener on an invoked machine will have received notice of the child's final state by that point.
2. Listeners of an invoked observable child aren't automatically unsubscribed, which is the reason for the open handle. Listeners will continue to receive emitted values from the observable until the end of time, which seems like unexpected behavior.

The test in the `jest` PR can be made to pass pretty easily by just calling `unsubscribe` on the child listener before the test ends, but wanted to make sure we're not missing some edge cases here if we do that.


Andarist:
Thank you for those! This is an awesome contribution. I'm gonna get to fixing those soon.

with-heart:
> Thank you for those! This is an awesome contribution. I'm gonna get to fixing those soon.

I might take a stab at a fix as well, just to see how you'd do things differently 😉

Andarist:
This is a bigger problem with the current implementation. Unfortunately, we don't wrap observables - we just pass them around directly right now (this is already solved in v5, at least partially, as we wrap all the things there, but we should add those tests there and make sure that they pass there).

But why should we wrap them in the first place? Most RxJS observables are **cold** by default - which means that with this code:
```ts
const interval$ = interval(10);
interval$.subscribe(val => console.log(val))
interval$.subscribe(val => console.log(val))
```
we actually end up with **two** separate streams. And this is what essentially happens in those added tests - your additional subscriptions are not actually listening to the observable actor that XState has created, they create extra streams and those are not governed by XState. It's not a surprise that this leaks open handles etc.

PR #3568, opened 2022-09-04
[v5] Simplify `interpreter.send(...)` method

davidkpiano:
This PR:

- Removes the 2nd `payload?` argument from `interpreter.send(...)`
- Only allows event objects (normal & SCXML) in `interpreter.send(...)`
- Refactors tests

```diff
-actor.send('SOME_EVENT')
+actor.send({ type: 'SOME_EVENT' })

-actor.send('EVENT', { some: 'payload' })
+actor.send({ type: 'EVENT', some: 'payload' })
```

[Discord poll/discussion](https://discord.com/channels/795785288994652170/876090999426928681/897468871324356648)


packages/core/src/interpreter.ts
```diff
@@ -488,13 +490,8 @@ export class Interpreter<
...
    * @param event The event(s) to send
    */
-  public send: PayloadSender<TEvent> = (event, payload?): void => {
-    const eventObject = toEventObject(event, payload);
-    const _event = toSCXMLEvent(eventObject);
-    // console.log(
-    //   `${_event.origin?.sessionId} -> ${this.sessionId}`,
-    //   _event.name
-    // );
+  public send(event: TEvent | SCXML.Event<TEvent>) {
```
davidkpiano, line 493: This is the main change.

packages/core/src/interpreter.ts
```diff
@@ -477,13 +479,8 @@ export class Interpreter<
...
```
Andarist, line 482: Note: IMHO we should also consider removing `SCXML.Event` too - it breaks subtyping and stuff. That, of course, should be done in a separate PR and all.

.changeset/gold-buses-thank.md
```diff
@@ -0,0 +1,18 @@
+---
+'xstate': major
```
Andarist, line 2: We should add a similar changeset for `@xstate/fsm` now and also somehow document the same change for `transition`, `raise`, `send`, `respond` etc

Andarist:
@davidkpiano we only need to add more changesets here, but please don't merge this into `v5/interpret-all-behaviors` just yet - let's land that PR first and only then land this onto `next`

packages/core/src/interpreter.ts
davidkpiano: The slight problem with that is that other actors should send SCXML events to actors, since they contain `origin` data. There are probably other ways to deal with the typing issue

.changeset/gold-buses-thank.md
davidkpiano: I'll get on it

PR #3571, opened 2022-09-05
[core] Deprecate `service.state`

davidkpiano:
This PR deprecates `someService.state` in favor of `someService.getSnapshot()`.

### DIFF
--- a/packages/xstate-react/src/utils.ts
+++ b/packages/xstate-react/src/utils.ts
@@ -20,7 +20,7 @@ export function partition<T, A extends T, B extends T>(
 export function getServiceSnapshot<
   TService extends Interpreter<any, any, any, any>
 >(service: TService): TService['state'] {
-  return service.status !== 0 ? service.state : service.machine.initialState;
+  return service.getSnapshot();
 }
 
 // From https://github.com/reduxjs/react-redux/blob/master/src/utils/shallowEqual.ts
--- a/packages/xstate-react/src/utils.ts
+++ b/packages/xstate-react/src/utils.ts
@@ -20,7 +20,9 @@ export function partition<T, A extends T, B extends T>(
 export function getServiceSnapshot<
   TService extends Interpreter<any, any, any, any>
 >(service: TService): TService['state'] {
-  return service.getSnapshot();
+  return service.status !== 0
+    ? service.getSnapshot()
+    : service.machine.initialState;
 }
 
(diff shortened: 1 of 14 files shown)


mellson (approved):
LGTM 🚢

packages/xstate-react/src/utils.ts
```diff
@@ -20,7 +20,7 @@ export function partition<T, A extends T, B extends T>(
 export function getServiceSnapshot<
   TService extends Interpreter<any, any, any, any>
 >(service: TService): TService['state'] {
-  return service.status !== 0 ? service.state : service.machine.initialState;
+  return service.getSnapshot();
```
Andarist, line 23: If I read things correctly this is not a safe change because it might lead to reading `.initialState` [here](https://github.com/statelyai/xstate/blob/e1aafa632ce77a96074a2a137403682db313c76a/packages/core/src/interpreter.ts#L1538) and that in turn might accidentally spawn actors [here](https://github.com/statelyai/xstate/blob/e1aafa632ce77a96074a2a137403682db313c76a/packages/core/src/interpreter.ts#L269-L272) since the `serviceScope` is provided there

davidkpiano: We should have a test for this 🤔 tests pass currently

davidkpiano: Reverted part of this to use `.getSnapshot()` instead of `.state`

Andarist: > We should have a test for this 🤔 tests pass currently

I've pushed out a test that would fail with the initial change: https://github.com/statelyai/xstate/pull/3571/commits/220eb4500a9d4dddd71530312585700e475b338f

Note that I've expected it to fail [here](https://github.com/statelyai/xstate/pull/3571/commits/220eb4500a9d4dddd71530312585700e475b338f#diff-d397283c364620d17bac75fb488220ba5e79afea92fb8ccd8501d14669d4e4ceR617) but actually, it was failing [here](https://github.com/statelyai/xstate/pull/3571/commits/220eb4500a9d4dddd71530312585700e475b338f#diff-d397283c364620d17bac75fb488220ba5e79afea92fb8ccd8501d14669d4e4ceR623). The first expectation pass nowadays because of [this fairly recent addition to the codebase](https://github.com/statelyai/xstate/blob/220eb4500a9d4dddd71530312585700e475b338f/packages/core/src/interpreter.ts#L1179-L1181)

Andarist (approved):

PR #3581, opened 2022-09-09
Resolve action objects as lazily as possible to avoid stale closure problem in React

Andarist:
This might have not been a bug introduced in the `v5/interpret-all-behaviors` branch but I've noticed it when reviewing it so I'm just creating a branch targeting that.

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -31,7 +31,7 @@ import type {
 } from './types';
 import type { State } from './State';
 import * as actionTypes from './actionTypes';
-import { toActionObject } from './actions';
+import { toActionObjects } from './actions';
 import { formatInitialTransition, formatTransition } from './stateUtils';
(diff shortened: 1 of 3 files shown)


packages/core/src/StateNode.ts
```diff
@@ -182,13 +182,9 @@ export class StateNode<
     this.history =
       this.config.history === true ? 'shallow' : this.config.history || false;
 
-    this.entry = toArray(this.config.entry).map((action) =>
-      toActionObject(action, this.machine.options.actions)
-    );
+    this.entry = toActionObjects(this.config.entry);
+    this.exit = toActionObjects(this.config.exit);
```
Andarist, line 186: this is the core of the fix - we can't resolve those action objects here because their references in the `implementations` might change later on

davidkpiano (approved):
Great find!

PR #3582, opened 2022-09-09
Fixed looking up inlined action in the implementations objects

Andarist:


packages/core/src/actions.ts
```diff
@@ -50,9 +53,13 @@ export function getActionFunction<
 
 export function resolveActionObject(
   actionObject: BaseActionObject,
-  actionFunctionMap?: ActionFunctionMap<any, any>
+  actionFunctionMap: ActionFunctionMap<any, any>
 ): BaseActionObject {
+  if (isDynamicAction(actionObject) || isExecutableAction(actionObject)) {
+    return actionObject;
+  }
```
Andarist, line 60: This is the core of the fix

packages/core/test/actions.test.ts
```diff
@@ -415,6 +415,62 @@ describe('entry/exit actions', () => {
...
+      );
+
+      const actor = interpret(machine).start();
+      actor.send({ type: 'EV' });
+
+      expect(spy).not.toHaveBeenCalled();
+      expect(actor.getSnapshot().context.assigned).toBe(true);
+    });
+
+    it("shouldn't use referenced custom action over a builtin one when there is a naming conflict", () => {
```
davidkpiano, line 445:
```suggestion
    it("shouldn't use inline named custom action over a builtin one when there is a naming conflict", () => {
```

davidkpiano (commented):
This changes the behavior from v4, right? That inline named actions will _not_ be replaced with equally named referenced actions?

I'm okay with the change, just want to double check (and we should changeset this)

packages/core/test/actions.test.ts
Andarist: ah, copy-pasta from the other added test case went wrong 😅

packages/core/test/actions.test.ts
```diff
@@ -415,6 +415,62 @@ describe('entry/exit actions', () => {
       expect(actual).toEqual(['loaded entry']);
     });
 
+    it("shouldn't use referenced custom action over a builtin one when there is a naming conflict", () => {
```
Andarist, line 418:
```suggestion
    it("shouldn't use a referenced custom action over a builtin one when there is a naming conflict", () => {
```

Andarist:
> This changes the behavior from v4, right? That inline named actions will not be replaced with equally named referenced actions?

Yeah. I would consider this to be a plain bug fix though. I doubt somebody has been intentionally relying on this one.

davidkpiano (approved):
TODO changeset:

> Named inline action functions will not be replaced by equally named referenced actions.

Andarist:
@davidkpiano changeset added

davidkpiano (approved):

PR #3583, opened 2022-09-09
Andarist/more exec related cleanup

Andarist:


packages/core/test/interpreter.test.ts
```diff
@@ -110,7 +110,7 @@ describe('interpreter', () => {
     });
 
     // https://github.com/statelyai/xstate/issues/1174
-    it('executes actions from a restored state', (done) => {
+    it.skip('executes actions from a restored state', (done) => {
```
Andarist, line 113: So... I'm not sure if this is the desired behavior. I don't think we should execute all of the actions contained in the persisted state (I also don't think that `.actions` should be a part of the state, but that's a separate discussion).

I think that we should either:
- call no actions
- only call entry actions of all the nodes in the configuration, which is completely different from calling `state.actions`

davidkpiano, line 113: Do we want to support this use-case?

davidkpiano (approved):
Approved in general but 1 question

packages/core/test/interpreter.test.ts
davidkpiano: IMO the important thing is re-invoking. If a client-side state machine is mid-invocation (e.g. a promise) and we reload the browser window, the machine should be restarted and any invocations should be re-invoked.

For server-side use-cases, actions should not be re-executed.

Andarist: Yes, reinvoking has to happen - I just treat it as an orthogonal concern to actions. I don't think that reinvoking is working properly with the current version of the code, so it won't regress by a lot with this PR anyway.

davidkpiano: Makes sense; this is good to merge 🚢

PR #3591, opened 2022-09-12
Simplify `transitionParallelNode` a little bit

Andarist:
I've just noticed that there was no reason to use an object here if it ended up being converted to an array anyway.


packages/core/src/stateUtils.ts
```diff
@@ -806,17 +806,10 @@ export function transitionParallelNode<
       _event
     );
     if (nextStateNode) {
-      transitionMap[subStateKey] = nextStateNode;
+      transitions.push(...nextStateNode);
```
davidkpiano, line 809: Sigh, I didn't give this one the best name

```suggestion
      // TODO: rename nextStateNode to nextTransitions
      transitions.push(...nextStateNode);
```

PR #3455, opened 2022-07-13
[v5] Interpret all behaviors

davidkpiano:
The biggest changes:

- ~~`ObservableActorRef`~~ is removed
- **`interpret` now works with any `Behavior`: `interpret(behavior)`**
- The `Behavior` interface now has `Behavior.getSnapshot`, which is important for determining which "state" should be exposed, and which state is internal state
- `predictableActionArguments` flag is removed; `actorCtx.exec` is used instead
- _Most_ state-machine-specific code is removed from the `Interpreter`; still have some more work to do here

PR #3592, opened 2022-09-12
Handle events sent to `self` closer to events send to other targets

Andarist:
Implementing the suggestion from https://github.com/statelyai/xstate/pull/3455/files#r966943424

### DIFF
--- a/packages/core/src/exec.ts
+++ b/packages/core/src/exec.ts
@@ -87,12 +87,8 @@ function getActionFunction<TState extends AnyState>(
         interpreter.defer(sendAction);
         return;
       } else {
-        if (sendAction.params.to) {
-          const target = sendAction.params.to;
-          execSendTo(sendAction.params._event, target, actorCtx);
-        } else {
-          interpreter.send(sendAction.params._event as SCXML.Event<any>);
-        }
+        const target = sendAction.params.to!;
+        execSendTo(sendAction.params._event, target, actorCtx);
       }
     },
     [actionTypes.cancel]: (_ctx, _e, { action }) => {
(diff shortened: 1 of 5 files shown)


packages/core/src/exec.ts
```diff
@@ -87,12 +87,8 @@ function getActionFunction<TState extends AnyState>(
         interpreter.defer(sendAction);
         return;
       } else {
-        if (sendAction.params.to) {
-          const target = sendAction.params.to;
-          execSendTo(sendAction.params._event, target, actorCtx);
-        } else {
-          interpreter.send(sendAction.params._event as SCXML.Event<any>);
```
Andarist, line 94: The notable difference is now that events sent to self will also get `.origin`. I think this wasn't the case before this change

packages/core/src/types.ts
```diff
@@ -1254,6 +1254,7 @@ export interface SendActionObject<
     event: TSentEvent;
     delay?: number;
     id: string | number;
+    internal: boolean;
```
davidkpiano, line 1257: Is this needed or can it be derived from `to` being self?

There's a possibility for an impossible state if `to` is an external actor but `internal: true`

Andarist: > Is this needed or can it be derived from to being self?

It can't be derived right now because it's still valid to send an external event to `self`. Perhaps this should be changed and all events sent to itself should be raised but that would be a separate change.

> There's a possibility for an impossible state if to is an external actor but internal: true

Right, however, we are in full control when it comes to creating those objects and we just don't allow this invalid combination.

davidkpiano (approved):

PR #3593, opened 2022-09-12
Reflect the type for internal `Interpreter#observers` closer to the possible runtime values

Andarist:


packages/core/src/interpreter.ts
```diff
@@ -104,7 +104,9 @@ export class Interpreter<
 
   private delayedEventsMap: Record<string, unknown> = {};
 
-  private observers: Set<Observer<SnapshotFrom<TBehavior>>> = new Set();
+  private observers: Set<
+    Required<Observer<SnapshotFrom<TBehavior>>>
```
Andarist, line 108: the `toObserver` that we use always resolves all callbacks and provides noops for the missing ones

davidkpiano (approved):

PR #3596, opened 2022-09-12
Remove traces of the `StateNodeConfig["key"]`

Andarist:
Follow-up to https://github.com/statelyai/xstate/pull/3455/files#r967087849


packages/core/test/actions.test.ts
```diff
@@ -39,7 +39,6 @@ describe('entry/exit actions', () => {
   };
 
   const lightMachine = createMachine({
-    key: 'light',
```
davidkpiano, line 42: We should preserve these as `id: 'light',` etc.

Andarist: I've already done this in places where it mattered for the tests. Do you want me to adjust other places too?

davidkpiano:
Before merging this in, let me think about naming...

- `id` here should serve as only the **root state node ID**
- The `id` (or `name`?) that the parent gives the child is for identifying the entire child actor
- The root state node ID should _not_ be used to self-identify the actor; only the state node

With that said, is `id` still the best name? I think _probably_ yes, but we should document the difference in behavior from v4 that this:

```js
spawn(someMachine, 'some-id');
```

Is different from this:
```ts
const someMachine = createMachine({
  id: 'rootNodeId'
})
```

davidkpiano (approved):

PR #3598, opened 2022-09-13
Remove `TestModel["defaults"]`

Andarist:


packages/xstate-test/src/TestModel.ts
```diff
@@ -34,14 +34,6 @@ import {
 } from './utils';
 import { traverseShortestPathsTo } from '@xstate/graph/src/graph';
 
-export interface TestModelDefaults<TState, TEvent extends EventObject> {
-  pathGenerator: PathGenerator<TState, TEvent>;
-}
-
-export const testModelDefaults: TestModelDefaults<any, any> = {
-  pathGenerator: getShortestPaths
```
Andarist, line 42: this default became redundant after: https://github.com/statelyai/xstate/pull/3446/files#r967827274

packages/xstate-test/src/TestModel.ts
```diff
@@ -70,7 +62,6 @@ export class TestModel<TState, TEvent extends EventObject> {
       }
     };
   }
-  public static defaults: TestModelDefaults<any, any> = testModelDefaults;
```
Andarist, line 73: If I read things correctly - this wasn't actually used anywhere

packages/xstate-test/src/TestModel.ts
```diff
@@ -398,19 +389,6 @@ export class TestModel<TState, TEvent extends EventObject> {
   }
 }
 
-/**
- * Specifies default configuration for `TestModel` instances for path generation options
- *
- * @param testModelConfiguration The partial configuration for all subsequent `TestModel` instances
```
Andarist, line 404: IMHO this breaks the encapsulation - it also wasn't actually utilized as per: https://github.com/statelyai/xstate/pull/3598/files#r969327028

PR #3599, opened 2022-09-13
Fixed a regression that has caused `machine.getInitialState(value)` to not follow always transitions correctly

### LINKED ISSUE #3544: Bug: eventless transitions in parallel states not taken anymore when starting interpreter from given state
### Description

Eventless transitions in parallel state are not taken anymore when interpreter is started from a given state.

### Expected result

Eventless transitions should always be taken.

For example, in the linked reproduction sandbox, the expected result is:
```
state.value == {"b":{"x":"bar","y":{}}} 
```

This is the result obtained with xstate version < 4.33.0

### Actual result

The eventless transition in example is not taken anymore with xstate version >= 4.33.0.

Instead, we get:

```
state.value == {"b":{"x":"foo","y":{}}} 
```

### Reproduction

https://codesandbox.io/s/affectionate-thunder-esq8rn?file=/src/index.js

### Additional context

It looks like a regression introduced in version 4.33.0.

Andarist:
fixes #3544

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -1010,7 +1010,10 @@ class StateNode<
     );
 
     for (const sn of resolvedConfig) {
-      if (!has(prevConfig, sn) || has(transition.entrySet, sn.parent)) {
+      if (
+        !has(prevConfig, sn) ||
+        (has(transition.entrySet, sn.parent) && !has(transition.entrySet, sn))
+      ) {
         transition.entrySet.push(sn);
       }
     }
@@ -1661,7 +1664,7 @@ class StateNode<
     return this.resolveTransition(
       {
         configuration,
-        entrySet: [...configuration],
+        entrySet: configuration,
         exitSet: [],
         transitions: [],
         source: undefined,
--- a/packages/core/test/rehydration.test.ts
+++ b/packages/core/test/rehydration.test.ts
@@ -98,21 +98,4 @@ describe('rehydration', () => {
       expect(actual).toEqual(['active', 'root']);
     });
   });
-
-  describe('using resolved state value', () => {
-    it('should follow always transitions', () => {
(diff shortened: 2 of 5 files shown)


packages/core/src/StateNode.ts
```diff
@@ -1661,7 +1664,7 @@ class StateNode<
     return this.resolveTransition(
       {
         configuration,
-        entrySet: [...configuration],
+        entrySet: configuration,
```
Andarist, line 1667: It brings back a shared reference being used for both `entrySet` and for `configuration`. This is somewhat dangerous but I just hope that v5 algorithm works better in this regard.

It has been like that for a long time, I've "fixed" this problem [here](https://github.com/statelyai/xstate/pull/3422/files#r905127802) but that accidentally caused the mentioned regression. So now instead of copying the reference, I make sure that we don't push items to the `entrySet` that are already there:
https://github.com/statelyai/xstate/pull/3599/files#diff-7f553f4d4c285f1c05cced2e979c5b957d8884e39e74cc6797b32c63a0be7e7aR1015

packages/core/test/rehydration.test.ts
```diff
@@ -98,4 +98,21 @@ describe('rehydration', () => {
...
+        initial: 'a',
+        states: {
+          a: {
+            always: [{ target: 'b' }]
+          },
+          b: {}
+        }
+      });
+
+      const service = interpret(machine).start(machine.getInitialState('a'));
```
Andarist, line 114: Interestingly, this doesn't follow always transitions:
```diff
-      const service = interpret(machine).start(machine.getInitialState('a'));
+      const service = interpret(machine).start('a');
```

I would consider this to be a bug too but it just never worked correctly and I don't think it's necessary to fix this in v4.

davidkpiano (approved):

PR #3603, opened 2022-09-15
xstate-svelte: Avoid using svelte store get in useSelector.

mittinatten:
The Svelte docs say `get`  should [not be used in hot paths](https://svelte.dev/docs#run-time-svelte-store-get). It is used in `useSelector` to compare the updated value with the previous value.  This PR changes the function to store the previous value in a local variable instead.

### DIFF
--- a/.changeset/afraid-kids-happen.md
+++ b/.changeset/afraid-kids-happen.md
@@ -2,4 +2,4 @@
 '@xstate/svelte': patch
 ---
 
-Avoid using svelte store get in useSelector
+Improve performance of the `useSelector` by avoiding `get`
--- a/.changeset/afraid-kids-happen.md
+++ b/.changeset/afraid-kids-happen.md
@@ -1,0 +1,5 @@
+---
+'@xstate/svelte': patch
+---
+
+Avoid using svelte store get in useSelector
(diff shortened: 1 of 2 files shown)


davidkpiano (approved):

packages/xstate-svelte/src/useSelector.ts
```diff
@@ -13,11 +13,15 @@ export const useSelector = <
...
+
+  let prevSelected = selector(actor.getSnapshot());
+
+  const selected = readable(prevSelected, (set) => {
     sub = actor.subscribe((state) => {
       const nextSelected = selector(state);
-      if (!compare(get(selected), nextSelected)) {
+      if (!compare(prevSelected, nextSelected)) {
         set(nextSelected);
+        prevSelected = nextSelected;
```
Andarist, line 24:
```suggestion
        prevSelected = nextSelected;
        set(nextSelected);
```

.changeset/afraid-kids-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/svelte': patch
+---
+
+Avoid using svelte store get in useSelector
```
Andarist, line 5:
```suggestion
Improve performance of the `useSelector` by avoiding `get`
```

davidkpiano:
Is this good to go?

PR #3607, opened 2022-09-18
Deprecate `createModel(...)`

davidkpiano:
The `createModel(...)` function is now marked as deprecated, as it will be removed in XState version 5. It is recommended to use [Typegen](https://stately.ai/blog/introducing-typescript-typegen-for-xstate) instead.

### DIFF
--- a/.changeset/shiny-kangaroos-battle.md
+++ b/.changeset/shiny-kangaroos-battle.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+The `createModel(...)` function is now marked as deprecated, as it will be removed in XState version 5. It is recommended to use [Typegen](https://stately.ai/blog/introducing-typescript-typegen-for-xstate) instead.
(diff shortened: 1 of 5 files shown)


with-heart:
RIP `createModel`, you were a good experiment

farskid:
Oh no 🙊

gkiely:
@davidkpiano Typegen has never been reliable for me and sometimes it just doesn't work.
image

Is there any other way to allow the typing that createModel allowed, it's a lot nicer than passing generics to the assign method and doesn't require reliance on an extension.

davidkpiano:
> @davidkpiano Typegen has never been reliable for me, it can never cover all the userland requirements and sometimes it just doesn't work. image
> 
> Is there any other way to allow the typing that createModel allowed, it's a lot nicer than passing generics to the assign method and doesn't require reliance on an extension.

What specific things are you using `createModel` for besides `assign`?

gkiely:
Nothing outside of assign.

Usually actions with assigns, typically invoked.
image

image

gkiely:
@davidkpiano After giving this some more thought, I've moved to using typescript narrowing which achieves the same thing.
image

Andarist (approved):

PR #3622, opened 2022-09-30
Re-enter only the target when an ancestor during an external event

arromeo:
Found that when the target node was an ancestor of the current node, it would add all ancestor nodes to the list of nodes to re-enter as it climbed to the target.


arromeo:
Closing. Forgot to follow the steps in the contributing guide.

Issue #3618, opened 2022-09-27
Entry action gets triggered on exit


thomasaull:
### Description

Consider the following state machine:
```js
{
  initial: "update",
  id: "root",

  states: {
    update: {
      initial: "default",

      states: {
        default: {
          on: {
            START: "doSomething.pick"
          }
        },

        doSomething: {
          initial: "pick",

          states: {
            default: {},
            pick: {
              entry: () => console.log("ENTRY: doSomething.pick"),
              exit: () => console.log("EXIT: doSomething.pick"),

              on: {
                CANCEL: "#root.update"
              }
            }
          }
        }
      }
    }
  }
}
```

### Expected result

Exiting the state `doSomething.pick` should only trigger the `exit` action.

The console output should be:
```
EXIT: doSomething.pick
```

### Actual result

Exiting the state `doSomething.pick` triggers the `exit` and the `entry` action.

The console output is:
```
EXIT: doSomething.pick
ENTRY: doSomething.pick
```

### Reproduction

https://codesandbox.io/s/polished-moon-8gw0uu?file=/src/index.js

### Additional context

Apparently this is introduced with xstate `4.33.0`, downgrading to `4.32.1` does not show this behaviour

Edit: Updated the console outputs to make things a bit more clear

Andarist:
This was specifically fixed by: https://github.com/statelyai/xstate/pull/3424

Why would you like those actions not to be called?

thomasaull:
@Andarist When I transition from `update.doSomething.pick` to `update.default` why would I want the entry action of the state I'm transitioning away from be called?

Andarist:
Oh, sorry - I've misread the given example. You are totally right, the entry action shouldn't happen here - I've thought that you were routed to the `pick` state through the `.initial` property on its parent.

This feels highly related to https://github.com/statelyai/xstate/issues/3617 . I'm gonna have to fix both in one sweep.

thomasaull:
@Andarist No worries! I feel like I could have simplified the example a bit more, but it could not reproduce in every scenario, so I was happy I had one where it did :)

Thanks!

arromeo:
Opened a PR that I believe fixes this issue. https://github.com/statelyai/xstate/pull/3623

Andarist:
> This feels highly related to https://github.com/statelyai/xstate/issues/3617 . I'm gonna have to fix both in one sweep.

[...]

Issue #3613, opened 2022-09-23
Bug: Transitioning to a parent from a child runs incorrect entry actions in v4.33.x


qtiki:
### Description

Transitioning from a child to a parent state sets the initial state of the parent active. However it erroneously starts activities and runs entry actions for all children. This works correctly with v4.32.1 but is broken from v4.33.0 onwards.

### Expected result

The expected output from the reproduction case would be something like this:
```
[1663939314769] parent activity start 
[1663939314769] parent entry 
[1663939314769] a activity start 
[1663939314769] a entry 
[1663939314769] state: {parent: "a"}
[1663939315783] a exit 
[1663939315784] a activity stop 
[1663939315784] b activity start 
[1663939315784] b entry 
[1663939315784] state: {parent: "b"}
[1663939316783] b exit 
[1663939316783] b activity stop 
[1663939316783] a activity start 
[1663939316783] a entry 
[1663939316783] state: {parent: "a"}
```
Here we can see that in the last transition we only see entry actions from `a`. This is the actual output from the [reproduction case with v4.32.1](https://codesandbox.io/s/child-activity-start-v4-32-1-works-9pp7z3).

### Actual result

With v4.33.x the output is as follows:
```
[1663939373636] parent activity start 
[1663939373636] parent entry 
[1663939373636] a activity start 
[1663939373636] a entry 
[1663939373637] state: {parent: "a"}
[1663939374637] a exit 
[1663939374637] a activity stop 
[1663939374637] b entry 
[1663939374637] b activity start 
[1663939374637] state: {parent: "b"}
[1663939375646] b exit 
[1663939375646] b activity stop 
[1663939375646] parent entry 
[1663939375646] parent activity start 
[1663939375646] a entry 
[1663939375646] a activity start 
[1663939375646] b entry 
[1663939375646] b activity start 
[1663939375646] state: {parent: "a"}
```
Here we can see that even though the state shows `{parent: "a"}` we see the entry actions for `b` too.

### Reproduction

https://codesandbox.io/s/child-activity-start-v4-33-6-broken-wul5gq

qtiki:
Now that I look at the log and compare it to the old one it seems to be even more broken than my description says. It seems to run the parent entry actions too without actually running the exit actions. I wouldn't expect it to re-enter the parent state since it's already active.

Andarist:
1. entry actions for `b` **definitely** shouldn't happen - and I already see why the bug happens. 
[...]

PR #3623, opened 2022-09-30
During external events, only enter target and descendent nodes

arromeo:
Found that when the current node was a descendent of the target node, it was re-entering every node as it climbed to the target.

fixes https://github.com/statelyai/xstate/issues/3618
fixes https://github.com/statelyai/xstate/issues/3613

### DIFF
--- a/.changeset/mighty-phones-press.md
+++ b/.changeset/mighty-phones-press.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+During external events, only enter target plus descendent nodes.
--- a/.changeset/mighty-phones-press.md
+++ b/.changeset/mighty-phones-press.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-During external events, only enter target plus descendent nodes.
+Fixed an issue with external transitions targeting ancestor states. In such a case, `entry` actions were incorrectly called on the states between the source state and the target state for states that were not reentered within this transition.
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -976,11 +976,22 @@ class StateNode<
   private getExternalReentryNodes(
     targetNode: StateNode<TContext, any, TEvent, any, any, any>
   ): Array<StateNode<TContext, any, TEvent, any, any, any>> {
+    if (this.order > targetNode.order) {
+      return [targetNode];
+    }
+
     const nodes: Array<StateNode<TContext, any, TEvent, any, any, any>> = [];
-    let [marker, possibleAncestor]: [
-      StateNode<TContext, any, TEvent, any, any, any> | undefined,
-      StateNode<TContext, any, TEvent, any, any, any>
-    ] = targetNode.order > this.order ? [targetNode, this] : [this, targetNode];
+    let marker:
+      | StateNode<TContext, any, TEvent, any, any, any>
+      | undefined = targetNode;
+    let possibleAncestor: StateNode<
+      TContext,
+      any,
+      TEvent,
+      any,
+      any,
+      any
+    > = this;
 
     while (marker && marker !== possibleAncestor) {
       nodes.push(marker);
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -223,6 +223,46 @@ describe('entry/exit actions', () => {
     }
   });
 
+  const deepMachine2 = createMachine({
+    initial: 'A',
+    id: 'root',
+    predictableActionArguments: true,
+    states: {
+      A: {
+        initial: 'default',
+        entry: 'enter_A',
+        on: {
+          CHANGE_TO_A2: 'A.A2'
+        },
+        states: {
+          default: {
+            entry: 'enter_default'
+          },
+          A1: {
+            entry: 'enter_A1',
+            exit: 'exit_A1',
+            on: {
+              CHANGE_TO_A2A: 'A2.A2a'
+            }
+          },
+          A2: {
+            initial: 'A2a',
+            entry: 'enter_A2',
+            states: {
+              A2a: {
+                entry: 'enter_A2a',
+                exit: 'exit_A2a',
+                on: {
+                  CHANGE_TO_A: '#root.A'
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  });
+
   const parallelMachine2 = Machine({
     initial: 'A',
     states: {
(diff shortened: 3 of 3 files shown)


.changeset/mighty-phones-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+During external events, only enter target plus descendent nodes.
```
Andarist, line 5: Unfortunately, this is not correct. The external transition should always leave the source state. If the target is an ancestor of the source then everything up to that ancestor (including that ancestor) should be exited, target should be reentered and initial states etc should be recursively resolved and entered.

packages/core/src/StateNode.ts
```diff
@@ -976,11 +976,22 @@ class StateNode<
   private getExternalReentryNodes(
```
Andarist, line 976: A test case should be added with this change because I'm not sure what exactly is being fixed here.

arromeo: I will add one.

.changeset/mighty-phones-press.md
arromeo: It is exiting the source node correctly. The issue is that it's entering all the ancestor nodes until it reaches the target. In the example provided in the issue here https://github.com/statelyai/xstate/issues/3618, it was only mentioned that it re-entered the source node, but if you were to add entry actions in the other nodes, you would see that all of them are being triggered until it reached the source.

Andarist: Ah, yes - that sounds like the issue that i’ve identified in that linked issue. However, the fix seems to be incomplete (I might be wrong). Perhaps it works because other code “patches” the result of this.

Once you add a test case I will take a second look at this and gonna try to figure out a test case that would prove this being incomplete, or I will land this.

thank you for your contribution! It’s very much appreciated

packages/core/src/StateNode.ts
arromeo: I added three tests. The third one I'm not 100% is the correct behaviour. The documentation says when it's external it should exit, but it doesn't seem to be happening there.

packages/core/test/actions.test.ts
```diff
@@ -522,6 +562,28 @@ describe('entry/exit actions', () => {
...
+          .transition('A.A2.A2a', 'CHANGE_TO_A')
+          .actions.map((a) => a.type)
+      ).toEqual(['exit_A2a', 'enter_A', 'enter_default']);
+    });
+
+    it('should enter all descendents when target is a descendent of current', () => {
+      expect(
+        deepMachine2.transition('A', 'CHANGE_TO_A2').actions.map((a) => a.type)
+      ).toEqual(['enter_A', 'enter_A2', 'enter_A2a']);
+    });
```
Andarist, line 585: You have mentioned in [the other comment](https://github.com/statelyai/xstate/pull/3623#discussion_r984953334) that you are unsure about this test and you have mentioned something about exit actions there. Could you elaborate? I don't see exit actions anyhow related to this transition and I'm unsure what you have observed.

packages/core/test/actions.test.ts
```diff
@@ -223,6 +223,46 @@ describe('entry/exit actions', () => {
     }
   });
 
+  const deepMachine2 = createMachine({
```
Andarist, line 226: Could you use a single machine per test? I know that a lot of tests are written in this way here but we are trying to not add any new ones like this.

The preferred approach to write tests right now looks something like this: https://github.com/statelyai/xstate/pull/3424/files#diff-0f3376f69e4a3b17ec5757d5e7efd1d4438b4623627438bd1de139674b975ccbR386

arromeo: Yeah, no worries. I'll update it.

packages/core/test/actions.test.ts
arromeo: I was referring to this from the documentation:

"External transitions will exit and reenter the state node in which the transition is defined. "

That made me wonder if there was supposed to be an exit from A in this case. Maybe I'm misinterpreting it.

packages/core/test/actions.test.ts
arromeo: Done.

packages/core/test/actions.test.ts
Andarist: I would have to recheck to be double sure but I think that yes, it should be exited for the external transition. The [transition domain](https://www.w3.org/TR/scxml/#getTransitionDomain) is the common ancestor of the source and the target and all states below the transition domain are *exited* (some might also get reentered).

This particular test had no exit action on the `A` - and that's why I wasn't sure what kind of an effect you are observing here.

arromeo: Oh, you're right. No wonder. I have updated it to include the exits so it's clear it's doing the right thing.

packages/core/test/actions.test.ts
```diff
@@ -522,6 +522,154 @@ describe('entry/exit actions', () => {
       ).toEqual(['exit_walk', 'exit_red', 'enter_red', 'enter_walk']);
     });
 
+    it('should exit current node and enter target node when target is not a descendent or ancestor of current', () => {
```
Andarist, line 525: note: this test is already passing without the proposed change

packages/core/test/actions.test.ts
```diff
@@ -522,6 +522,154 @@ describe('entry/exit actions', () => {
...
+        }
+      });
+
+      const service = interpret(machine).start({ A: { A2: 'A2a' } });
+      service.send('CHANGE_TO_A');
+
+      expect(actual).toEqual(['exit_A2a', 'enter_A', 'enter_default']);
+    });
+
+    it('should enter all descendents when target is a descendent of current', () => {
```
Andarist, line 617: note: this test is already passing without the proposed change too

packages/core/src/StateNode.ts
Andarist: > The third one I'm not 100% is the correct behaviour. The documentation says when it's external it should exit, but it doesn't seem to be happening there.

Do you refer to the first test case and to the fact that this `A` state is not being exited during this transition?
https://github.com/statelyai/xstate/blob/111820100772bf5da515edc3d6fc40931a7d358f/packages/core/test/actions.test.ts#L532

Andarist: I see that we've discussed this in the other thread [here](https://github.com/statelyai/xstate/pull/3623#discussion_r985128206) but the `exit` has not been added to this particular state - this is perhaps to make the test pass. Would this be correct?

.changeset/mighty-phones-press.md
Andarist, line 5:
```suggestion
Fixed an issue with external transitions targeting ancestor states. In such a case, `entry` actions were incorrectly called on the states between the source state and the target state for states that were not reentered within this transition.
```

PR #3635, opened 2022-10-12
fix: onDone invocation order

### LINKED ISSUE #3416: Bug: onDone action is getting called after exit action
### Description

```js
import { createMachine } from "xstate";

export const machine = createMachine(
  {
    id: "action-order-bug",

    initial: "initial",
    states: {
      initial: {
        exit: () => {
          console.log("exit");
        },
        invoke: {
          id: "fakeService",
          src: "fakeService",
          onDone: {
            actions: () => {
              console.log("onDone");
            },
            target: "done"
          }
        }
      },
      done: {}
    }
  },
  {
    services: {
      fakeService: () => Promise.resolve("success")
    }
  }
);
```

In the above example, the `exit` action will be called before the `onDone` action. However when `onDone` is called the state machine is still in `initial` state.

### Expected result

I expect the `onDone` action to be called before the `exit` action.

### Actual result

`exit` action getting called before `onDone` action.

### Reproduction

https://codesandbox.io/s/broken-cherry-j1n230?file=/src/machine.js

### Additional context

_No response_

ilteoood:
Closes #3416


packages/core/src/StateNode.ts
```diff
@@ -1087,17 +1087,17 @@ class StateNode<
...
     );
 
-    const actions = exitActions
-      .concat([
-        toActionObjects(transition.actions, this.machine.options.actions as any)
-      ])
+    const actions = [
+      toActionObjects(transition.actions, this.machine.options.actions as any)
+    ]
+      .concat(exitActions)
```
Andarist, line 1100: This one is definitely a wrong change. All exit actions should be executed before transition actions and entry actions should be executed after transition actions:
1. exit actions
2. transition actions
3. entry actions

https://www.w3.org/TR/scxml/#microstepProcedure

Andarist:
Since this PR is trying to fix a reported issue that got classified as "working as intended" ([here](https://github.com/statelyai/xstate/issues/3416#issuecomment-1276008949)), I think it's fair to close this PR too.

A further discussion could be started [here](https://github.com/statelyai/xstate/discussions) to discuss with the community how intuitive the current behavior is. As it stands today - it's not a bug though.

PR #3666, opened 2022-11-03
[core] Remove predictable action arguments warning

davidkpiano:


packages/core/src/Machine.ts
```diff
@@ -142,11 +139,5 @@ export function createMachine<
   TServiceMap,
   TTypesMeta
 > {
-  if (!IS_PRODUCTION && !config.predictableActionArguments && !warned) {
```
Andarist, line 145: I think a better change here would be something along those lines:
```diff
-if (!IS_PRODUCTION && !config.predictableActionArguments && !warned) {
+if (!IS_PRODUCTION && !(predictableActionArguments in config) && !warned) {
```

davidkpiano: If it prevents the excess warnings, feel free to make the change on this branch (or a new one); I'm open for any improvement

Andarist: This change wouldn't prevent the excess warnings - it would allow people to silence the warning by using `predictableActionArguments: false`. 

We could also add a test-env detection (like [here](https://github.com/emotion-js/emotion/blob/92be52d894c7d81d013285e9dfe90820e6b178f8/packages/react/src/index.js#L20)) and skip the warnings in those envs. This should help with excessive warnings to some extent.

davidkpiano: @Andarist I made these changes

.changeset/silver-mayflies-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+The warning for the `predictableActionArguments` setting has been improved to only warn if it is absent.
```
Andarist, line 5:
```suggestion
The warning for the `predictableActionArguments` setting has been improved to only warn if it is absent. You can disable the warning by setting `predictableActionArguments: false`. It's still recommended to set it to `true` though.
```

Andarist (approved):

PR #3670, opened 2022-11-10
Export useMachines return type

bdombro:
...so that we can more easily pass it around as props

### DIFF
--- a/packages/xstate-react/src/useMachine.ts
+++ b/packages/xstate-react/src/useMachine.ts
@@ -56,7 +56,7 @@ type RestParams<
         >
     ];
 
-type UseMachineReturn<
+export type UseMachineReturn<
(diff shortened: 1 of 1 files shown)


packages/xstate-react/src/useMachine.ts
```diff
@@ -56,7 +56,7 @@ type RestParams<
         >
     ];
 
-type UseMachineReturn<
+export type UseMachineReturn<
```
Andarist, line 59: If we'd export this I think that we should try to think about a better name for this. However, I'm also not sure if we should export this - the builtin React hooks don't export their return types like this, so why we'd do that?

bdombro: yeah you prob right

PR #3675, opened 2022-11-14
[v5] Input

davidkpiano:


packages/core/test/input.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+    const machine = createMachine<typeof t>({
+      context: { count: 42 }
+    });
+
+    expect(() => {
+      machine.getInitialState();
+    }).not.toThrowError();
+  });
```
with-heart, line 100: I assume you're already aware of this but just wanted to make a note that the description and the test here don't match.

packages/core/test/input.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      context: ({ input }) => ({
+        count: input.startCount
+      }),
+      entry: (ctx) => {
+        expect(ctx.count).toBe(42);
+      }
+    });
+
+    const inputMachine = machine.withInput({ startCount: 42 });
+    const initialState = inputMachine.getInitialState();
```
with-heart, line 23: I love this so much!

`withInput` is going to be a nice mechanism for cleaning up a lot of boilerplate, and defining types externally with `createTypes` is a little more clear than `schema` + I like that creating a generic machine is now just `typeof t` instead of needing to remember which params it accepts and in which order.

with-heart (commented):
❤️ So much goodness coming with v5, excited to follow along with this draft in particular

packages/core/test/input.test.ts
davidkpiano: Ah thanks for the catch (was unaware!)

tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "resolveJsonModule": true,
     "jsx": "react",
     "lib": ["es2019", "dom"],
-    "strictPropertyInitialization": true
+    "strictPropertyInitialization": true,
+    "noErrorTruncation": true
```
Andarist, line 18: We don't want to commit this permanently
```suggestion
    "strictPropertyInitialization": true
```

packages/core/test/types.test.ts
```diff
@@ -480,3 +493,90 @@ describe('spawn', () => {
...
+          // @ts-expect-error
+          other: 42
+        }
+      }
+    );
+  });
+});
+
+describe('state.children', () => {
+  it('should type state children correctly', () => {
```
Andarist, line 554: All of the related functionality should kinda be removed from this PR. The PR's title is: "[v5] Input" but yet it messes with `.children`.

Of course, we want to improve the types for `children` and so on - but that should be a separate unit of work. Those PRs tend to be a little bit big and it's best to have a focused place for a discussion related to the goal of the PR.

packages/core/test/types.test.ts
```diff
@@ -229,16 +244,17 @@ describe('Raise events', () => {
...
       // @ts-expect-error
       context: 'string'
     });
   });
 });
 
 describe('context', () => {
-  it('should infer context type from `config.context` when there is no `schema.context`', () => {
-    createMachine(
+  // TODO: how would we do context inference here, if that is even desirable?
```
Andarist, line 255: Inference for a single slot is not possible with "generic maps". Only a type parameter can be inferred but we no longer have `TContext` here... we have `TTypes` and only that can be inferred (where `TTypes['context']` can't be inferred on its own)

Andarist: Agreed. That capability was built on top of TS tricks anyway - and I'm not super confident in them. It was here mainly to match closer how it used to work before `tsTypes` and typegen but it's better to just require this as part of the other "generics"

davidkpiano:
Superseded by #3795

@Andarist I split out only the input-specific logic into #3795, meaning it is currently _not_ strongly-typed. My thinking is we add input types after the "unified generic" type (separate PR) goes in.

PR #3677, opened 2022-11-14
Fixed issues around reentrancy during external transitions

Andarist:
We didn't have an explicit issue open about those fixed issues. Those are things that I've spotted when investigating some issue reports and when reviewing some other PR.

The issue with the targeted ancestor not being **exited** correctly when it was already being **entered** during reentering transition has been mentioned in the second point here:
https://github.com/statelyai/xstate/issues/3613#issuecomment-1256307617

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -956,14 +956,11 @@ class StateNode<
       transitions: [selectedTransition],
       exitSet: isInternal
         ? []
-        : [
-            this,
-            ...flatten(
-              nextStateNodes.map((targetNode) =>
-                this.getExternalReentryNodes(targetNode)
-              )
+        : flatten(
+            nextStateNodes.map((targetNode) =>
+              this.getExternalReentryNodes(targetNode)
             )
-          ],
+          ),
       configuration: allNextStateNodes,
       source: state,
       actions
@@ -977,7 +974,7 @@ class StateNode<
     targetNode: StateNode<TContext, any, TEvent, any, any, any>
   ): Array<StateNode<TContext, any, TEvent, any, any, any>> {
     if (this.order < targetNode.order) {
-      return [];
+      return [this];
     }
 
     const nodes: Array<StateNode<TContext, any, TEvent, any, any, any>> = [];
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -958,7 +958,7 @@ class StateNode<
         ? []
         : flatten(
             nextStateNodes.map((targetNode) =>
-              this.getExternalReentryNodes(targetNode)
+              this.getPotentiallyReenteringNodes(targetNode)
             )
           ),
       configuration: allNextStateNodes,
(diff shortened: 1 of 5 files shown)


packages/core/test/actions.test.ts
```diff
@@ -725,7 +727,8 @@ describe('entry/exit actions', () => {
         'exit: a.a1',
         'exit: a',
         'enter: a',
-        'enter: a.a1'
+        'enter: a.a1',
+        'enter: a.a1.a11'
```
Andarist, line 731: This has regressed in https://github.com/statelyai/xstate/pull/3424

packages/core/test/actions.test.ts
```diff
@@ -725,9 +727,115 @@ describe('entry/exit actions', () => {
         'exit: a.a1',
         'exit: a',
         'enter: a',
+        'enter: a.a1',
+        'enter: a.a1.a11'
+      ]);
+    });
+
+    it('should reenter leaf state during its self-transition', () => {
```
Andarist, line 735: This test and the other ones added below were already working. I was trying to "break" the existing implementation for those cases but they seem to work fine. They might overlap with some existing tests but I've figured out that it's better to have more than fewer tests so I've committed them 😜

packages/core/src/StateNode.ts
```diff
@@ -1015,17 +1012,20 @@ class StateNode<
...
         !has(prevConfig, sn) ||
-        (has(transition.entrySet, sn.parent) && !has(transition.entrySet, sn))
+        has(transition.exitSet, sn) ||
+        (sn.parent && entrySet.has(sn.parent))
       ) {
-        transition.entrySet.push(sn);
+        entrySet.add(sn);
       }
     }
     for (const sn of prevConfig) {
```
Andarist, line 1031: I think that the logic in this for-of loop is fragile but I couldn't make it break right now. Perhaps it's because nodes in the `prevConfig` are already sorted in a way that allows this for-of loop to do its job.

packages/core/test/actions.test.ts
```diff
@@ -647,6 +648,7 @@ describe('entry/exit actions', () => {
       expect(flushTracked()).toEqual([
         'exit: A.A2.A2_child',
         'exit: A.A2',
+        'exit: A',
```
davidkpiano, line 651: So strange; I wonder why these were expected to be omitted (not exited) originally. What was our stance on whether to exit the root node or not?

davidkpiano (approved):
Code LGTM if we're okay with exiting/entering the root node (only concern)

packages/core/test/actions.test.ts
Andarist: > So strange; I wonder why these were expected to be omitted (not exited) originally.

It wasn't expected. When working on https://github.com/statelyai/xstate/pull/3424 I've focused only on one side of the reentrancy (re-entering) but I've missed that I didn't actually fix the other half of the story - that the reentered state has to exit first. 

> What was our stance on whether to exit the root node or not?

In v4 we can exit the root state, in v5 this should become impossible (or at least that's what we've decided in the past months).

PR #3687, opened 2022-11-29
Mention timout in docstring of `waitFor`

ansgarm:
I somehow missed this in the [docs](https://xstate.js.org/docs/guides/interpretation.html#waitfor) and it took me quite a while to figure out where the timeout error came from. If you look at the source, it is obvious as the options mention the timout, but I only looked at the inline docs in my editor which only showed the `waitFor` docstring that didn't mention this timeout at all 😅


davidkpiano (approved):

davidkpiano:
Can you add a changeset please? Great change 👏

Andarist:
This is a nice improvement but probably doesn't require an explicit call-out in the generated changelog. Since we already have pending changesets on `main` - this will just be released in the next release anyway. I'm fine with landing this as-is

PR #3688, opened 2022-11-29
Convenient syntax for actor subscription in `useMachine` and `useInterpret`

farskid:
You can conveniently create a subscription to the interpreter inside `useInterpret` and `useMachine` hooks.

```js
  initial: 'active',
  states: {
    inactive: {
      on: {
        ACTIVATE: 'active'
      }
    },
    active: {
      on: {
        DEACTIVATE: 'inactive'
      }
    }
  }
});
```

**Before**:

```jsx
const [, , actor] = useMachine(machine);
// Or
const [, , actor] = useInterpret(machine);

React.useEffect(() => {
  return actor.subscribe((state) => {
    // do something with the state object
  }).unsubscribe;
}, []);
```

**After**:

```jsx
const [, , actor] = useMachine(machine, {
  subscribe: (state) => {
    // do something with the state object
  }
});
// Or
const [, , actor] = useInterpret(machine, {
  subscribe: (state) => {
    // do something with the state object
  }
});
```

### DIFF
--- a/.changeset/curvy-moles-applaud.md
+++ b/.changeset/curvy-moles-applaud.md
@@ -1,0 +1,53 @@
+---
+'@xstate/react': minor
+---
+
+You can conveniently create a subscription to the interpreter inside `useInterpret` and `useMachine` hooks.
+
+```js
+const machine = createMachine({
+  initial: 'active',
+  states: {
+    inactive: {
+      on: {
+        ACTIVATE: 'active'
+      }
+    },
+    active: {
+      on: {
+        DEACTIVATE: 'inactive'
+      }
+    }
+  }
+});
+```
+
+**Before**:
+
+```jsx
+const [, , actor] = useMachine(machine);
+// Or
+const [, , actor] = useInterpret(machine);
+
+React.useEffect(() => {
+  return actor.subscribe((state) => {
+    // do something with the state object
+  }).unsubscribe;
+}, []);
+```
+
+**After**:
+
+```jsx
+const [, , actor] = useMachine(machine, {
+  subscribe: (state) => {
+    // do something with the state object
+  }
+});
+// Or
+const [, , actor] = useInterpret(machine, {
+  subscribe: (state) => {
+    // do something with the state object
+  }
+});
+```
--- a/packages/xstate-react/src/useInterpret.ts
+++ b/packages/xstate-react/src/useInterpret.ts
@@ -82,6 +82,12 @@ export function useIdleInterpreter(
     Object.assign(service.machine.options.delays!, delays);
   }, [actions, guards, activities, services, delays]);
 
+  useIsomorphicLayoutEffect(() => {
+    if (options.subscribe) {
+      return service.subscribe(options.subscribe).unsubscribe;
+    }
+  }, []);
+
   return service as any;
 }
 
--- a/packages/xstate-react/src/useMachine.ts
+++ b/packages/xstate-react/src/useMachine.ts
@@ -29,6 +29,10 @@ export interface UseMachineOptions<TContext, TEvent extends EventObject> {
    * start at this state instead of its `initialState`.
    */
   state?: StateConfig<TContext, TEvent>;
+  /**
+   * Convenient syntax to create a subscription to the interpreter at the creation time.
+   */
+  subscribe?: (state: State<TContext, TEvent>) => void;
 }
 
 type RestParams<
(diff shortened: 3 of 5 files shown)


farskid:
@davidkpiano @Andarist I'm wondering if we could allow an array of subscriptions too?

.changeset/curvy-moles-applaud.md
```diff
@@ -0,0 +1,53 @@
+---
+'@xstate/react': minor
+---
+
+You can conveniently create a subscription to the interpreter inside `useInterpret` and `useMachine` hooks.
```
Andarist, line 5: I have some concerns related to this. A convenient solution might make it too easy for people to subscribe to the state like this - but that's usually something that should be avoided.

Could you mention what kind of real-life use cases motivate this proposal?

farskid: TLDR;
It allows for an easier way to avoid having `useInterpret|useMachine` + `useEffect` combo for cases such as persisting the state to local storage on every state change, for example.

Andarist: I'm a little bit more pro-composition and in this case, creating a service and subscribing to it are two separate concerns (IMHO). This is especially annoying in React where we can't even use hooks conditionally - so for the majority of `subscribe`-less cases, we would still clobber the internal hooks slot which would in turn clobber the React devtools UI with a redundant item (not that this list isn't clobbered today though 🤣 ).

Like, even `react-redux` doesn't come with `useSelector` that would use `shallowEqual` as the comparator (even though that was most likely this was an expectation of most of their users at first). Their official response is a copy-pastable recipe:
https://react-redux.js.org/api/hooks#recipe-useshallowequalselector

When it comes to `useMachine` this feels like a super straightforward (minute the types, types are complicated) recipe:
```ts
function useSubscribableMachine(getMachine, { subscribe, ...options } = {}) {
    const result = useMachine(getMachine, options);
        if (subscribe) {
            return service.subscribe(subscribe).unsubscribe;
        }
    }, [subscribe]);
    return result;
}
```

And when it comes to `useInterpret`... this is already supported 😅 :
```ts
useInterpret(machine, {}, state => { /* do what u want */ })
```

I don't strongly mind this extension but I think that the "slim core" has a certain beauty to it. Extending low-level bits is what writing software is - if we need to add something on top of an existing block we can just extend it and use it throughout the app.

packages/xstate-react/src/useInterpret.ts
```diff
@@ -82,6 +82,12 @@ export function useIdleInterpreter(
     Object.assign(service.machine.options.delays!, delays);
   }, [actions, guards, activities, services, delays]);
 
+  useIsomorphicLayoutEffect(() => {
+    if (options.subscribe) {
+      return service.subscribe(options.subscribe).unsubscribe;
+    }
+  }, []);
```
Andarist, line 89: this suffers from the stale closure problem, it should be:
```suggestion
  }, [options.subscribe]);
```

packages/xstate-react/src/useMachine.ts
```diff
@@ -29,6 +30,10 @@ export interface UseMachineOptions<TContext, TEvent extends EventObject> {
    * start at this state instead of its `initialState`.
    */
   state?: StateConfig<TContext, TEvent>;
+  /**
+   * Convenient syntax to create a subscription to the interpreter at the creation time.
+   */
+  subscribe?: (state: AnyState) => void;
```
Andarist, line 36: This type is definitely too loose - if we gonna decide to move with this feature it should provide a stricter type to the one who has provided the callback.

packages/xstate-react/src/useInterpret.ts
Andarist, line 89: This has been applied in the internal `useIdleInterpret` and now `useInterpret` has two ways t do the same thing, as it already supports such subscribers as its third parameter, see here:
https://github.com/statelyai/xstate/blob/cd79700880890675af7d0e96516a2622630e2905/packages/xstate-react/src/useInterpret.ts#L131-L141

.changeset/curvy-moles-applaud.md
farskid: I had no idea there's a 3rd argument and I believe most users don't know that either :D

packages/xstate-react/src/useMachine.ts
farskid: How do you tighten the types here? Reuse the types of `interpreter.subscribe` argument? Because I tried that, but that subscribe method doesn't infer types.

packages/xstate-react/src/useInterpret.ts
farskid: Relates to https://github.com/statelyai/xstate/pull/3688#discussion_r1035087432

farskid:
@Andarist I talked to @davidkpiano, and he believes this can wait til XState V5. So closing this.

PR #3693, opened 2022-12-07
Sort imports on save in VS Code, and sort existing files

mellson:
I kept getting file changes whenever I saved a file because my vscode setup tries to organize imports on save. This PR sets `organizeImports` to `false` in the shared vs code settings file. Setting it to false means fewer file changes now, but my preference would be to set it to `true` so we get coherent imports over time.


.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "editor.codeActionsOnSave": {
+    "source.organizeImports": false
```
Andarist, line 4: What about this one?
```suggestion
    "source.sortImports": true
```

Personally, I don't mind that imports are getting shuffled in PRs. They are always on the top of the file and I just skip that part when reviewing. I also wouldn't mind if we'd just make one big PR that would sort all of them in all of the files. Once that would land this would become non-issue for future PRs, right?

mellson: I can run the formatting against all the files in this PR if you'd like to land it together with this change?

mellson: Done; it wasn't even that many files that needed sorting 👍🏻

Andarist: Oh, I didn't quite realize that this is a PR to the XState itself 🤦 I would like to enable this option... but perhaps in the future. Right now, we'll end up with conflicts with the `next` branch - they won't be that hard to fix though so maybe we can just roll with it. 

At the end of the day - I'm fine with merging this but personally, I would probably delay doing this after we get back to having a single release line. I'll leave the decision about this PR to @davidkpiano

Andarist:
@mellson with the current shape of this PR - the PR's title should be adjusted to reflect the current set of changes

.vscode/settings.json
davidkpiano: Let's tackle this in this order:

0. Merge `next` into #3455 (if needed - it's out of sync right now)
1. Merge #3455 when ready into `next`
2. "Sync" (merge?) `main` into `next
3. Merge this PR into `main`
4. Merge `main` into `next` again with the latest changes

davidkpiano:
🧹 Closing as stale - this is a nice-to-have but would need to be redone

Issue #2994, opened 2022-01-31
Bug: send() with specific TContext/TEvent/TSentEvent not assignable to `on.<event>.actions`


hedgepigdaniel:
### Description

The return type of `send<TContext, TEvent, TSentEvent>()` is `SendAction<TContext, TEvent, TSentEvent>`. As I understand, the generics have the following meaning:
 * `TContext`: the type of the context of the machine where the action is sent from
 * `TEvent`: the type of the events of the machine where the action is sent from
 * `TSentEvent`: the type of the event that is sent (possibly to a different machine than where the event is send from, as added in https://github.com/statelyai/xstate/pull/1178)

However when using `send` in the `on.<state>.actions` key, the `TEvent` type is inferred as `never`. Manually narrowing the `TEvent` type to refer to the source machine's event type results in the returned `SendAction` not being assignable to the `actions` array of the machine's transitions.

The expected type of the actions in the `actions` array is:
```typescript
export declare type Action<
  TContext,
  TEvent extends EventObject
> =
  | ActionType
  | ActionObject<TContext, TEvent>
  | ActionFunction<TContext, TEvent>;
```

but this apparently isn't satisfied by the `SendEvent` returned from `send`:
```typescript
export interface SendAction<
  TContext,
  TEvent extends EventObject,
  TSentEvent extends EventObject
> extends ActionObject<TContext, TEvent> {
    to:
      | string
      | number
      | ActorRef<any>
      | ExprWithMeta<TContext, TEvent, string | number | ActorRef<any>>
      | undefined;
    event: TSentEvent | SendExpr<TContext, TEvent, TSentEvent>;
    delay?: number | string | DelayExpr<TContext, TEvent>;
    id: string | number;
}
```

### Expected result

It should be possible to specify a type for `TEvent` compatible with a machine, and in that case the return type of `send()` should be assignable to the `actions` on that machine.

### Actual result

There is a type error:
```
Type '{ target: string; actions: SendAction<ToggleContext, ToggleEvent, { type: "LOG"; }>[]; }' is not assignable to type 'TransitionConfigOrTarget<ToggleContext, { type: "TOGGLE"; }>'.
  Types of property 'actions' are incompatible.
    Type 'SendAction<ToggleContext, ToggleEvent, { type: "LOG"; }>[]' is not assignable to type 'Actions<ToggleContext, { type: "TOGGLE"; }>'.
[...]

Issue #2897, opened 2021-12-28
Cannot use raise action in model-based machine


wKovacs64:
**Description**

TypeScript gets mad when you try to use the `raise` action inside a model-based machine definition.

**Expected Result**

I would expect it to work as it does with a machine not created from a model.

**Actual Result**

```
Type '{ STEP: { target: string; }; RAISE: { target: string; actions: (RaiseAction<EventObject> | SendAction<unknown, AnyEventObject, EventObject>)[]; }; }' is not assignable to type 'TransitionsConfig<{}, UnionFromCreatorsReturnTypes<FinalEventCreators<{ STEP: () => {}; RAISE: () => {}; NEXT: () => {}; RESET: () => {}; TOGGLE: () => {}; }>>> | undefined'.
  Types of property 'RAISE' are incompatible.
    Type '{ target: string; actions: (RaiseAction<EventObject> | SendAction<unknown, AnyEventObject, EventObject>)[]; }' is not assignable to type 'TransitionConfigOrTarget<{}, { type: "RAISE"; }>'.
      Types of property 'actions' are incompatible.
        Type '(RaiseAction<EventObject> | SendAction<unknown, AnyEventObject, EventObject>)[]' is not assignable to type 'Actions<{}, { type: "RAISE"; }> | undefined'.
          Type '(RaiseAction<EventObject> | SendAction<unknown, AnyEventObject, EventObject>)[]' is not assignable to type 'Action<{}, { type: "RAISE"; }>[]'.
            Type 'RaiseAction<EventObject> | SendAction<unknown, AnyEventObject, EventObject>' is not assignable to type 'Action<{}, { type: "RAISE"; }>'.
              Type 'SendAction<unknown, AnyEventObject, EventObject>' is not assignable to type 'Action<{}, { type: "RAISE"; }>'.
                Type 'SendAction<unknown, AnyEventObject, EventObject>' is not assignable to type 'string'.(2322)
```

**Reproduction**

https://stately.ai/viz/293fb8f2-2ede-4e1f-965f-9e1288097cb3

**Additional context**

I created this minimal reproduction because it wasn't working in my real app. In the real app, it's more than a TS error though - it doesn't seem to raise the event at all. I'm assuming there must be more to the story in the real app though, as the reproduction here _does_ actually work, it just makes TS mad.

Andarist:
I'm not yet sure what happens here, but this is what I've gathered:
[...]

Issue #1414, opened 2020-08-26
raise actions cause TypeScript errors


AlCalzone:
**Description**
I was asked to post this here ([original comment](https://github.com/microsoft/TypeScript/issues/39826#issuecomment-680761736))
After I turned on `strictFunctionTypes` in my repo, TypeScript started complaining about `raise` actions in my transition definitions.
In https://github.com/AlCalzone/node-zwave-js/blob/e10cb2c6d706bfc9dda0b5039ec38c651a2fbbd7/packages/zwave-js/src/lib/driver/SendThreadMachine.ts, you'll see I had to `any` some of the `raise` actions, e.g. line 394 or 405 ([relevant commit](https://github.com/AlCalzone/node-zwave-js/commit/e10cb2c6d706bfc9dda0b5039ec38c651a2fbbd7))

**Expected Result**
Since `raise` actions don't convey any context, I wouldn't expect them to cause incompatible types. 

**Actual Result**
If I remove the `as any` in line 405, this is the output from `lerna run build`:
```
zwave-js: src\lib\driver\SendThreadMachine.ts(383,4): error TS2322: Type '{ add: ({ cond: string; actions: any[]; } | { actions: (AssignAction<SendThreadContext, { type: "add"; transaction: Transaction; }> | RaiseAction<EventObject> | SendAction<unknown, AnyEventObject, EventObject>)[]; })[]; message: ({ cond: string; actions: any; } | { actions: (_: any, evt: any) => void; })[]; serialAPIUnexpected: ({ cond: string; actions: any; } | { actions: (_: any, evt: any) => void; })[]; sortQueue: { actions: any[]; }; }' is not assignable to type 'TransitionsConfigMap<SendThreadContext, SendThreadEvent> | TransitionsConfigArray<SendThreadContext, SendThreadEvent> | 
undefined'.
zwave-js:   Types of property 'add' are incompatible.
zwave-js:     Type '({ cond: string; actions: any[]; } | { actions: (AssignAction<SendThreadContext, { type: "add"; transaction: Transaction; }> | RaiseAction<EventObject> | SendAction<unknown, AnyEventObject, EventObject>)[]; })[]' is not assignable to type 'SingleOrArray<string | StateNode<SendThreadContext, any, { type: "add"; transaction: Transaction; }, { value: any; context: SendThreadContext; }> | TransitionConfig<SendThreadContext, 
{ type: "add"; transaction: Transaction; }> | undefined>'.
[...]

PR #3694, opened 2022-12-07
Introduce `TExpressionEvent` generic

Andarist:
With this change, the inferred types for `raise` are being improved. We can see here that both `TExpressionEvent` and `TEvent` are inferred correctly here - to `{ type: 'DECIDE'; aloha?: boolean | undefined; }` and `GreetingEvent`, respectively.
Screenshot 2022-12-07 at 12 15 34
In here we can see that this improves autocomplete capabilities in the IDE because the contextual type is restricted to what can actually be sent to this machine:
Screenshot 2022-12-07 at 12 15 50
And in here we can see that a type error is raised (huehue) correctly when we try to use an invalid event:
Screenshot 2022-12-07 at 12 16 12

⚠️ there are some type failures here as I didn't yet adjust everything around the implementation etc. This only touches public-facing APIs etc - as only those had to be adjusted to validate the idea.

I'm not super keen on touching types in v4 as I'd really like to completely rewrite them in v5 (IMHO it's the easiest way to clean them up). OTOH, this brings some very nice improvements to the table.

Our types are so complex that it's quite hard to predict all potential problems with this but, dunno, perhaps it's worth shipping this?

fixes https://github.com/statelyai/xstate/issues/2897
fixes https://github.com/statelyai/xstate/issues/1414
fixes https://github.com/statelyai/xstate/issues/2994

### DIFF
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -185,7 +185,7 @@ export function resolveRaise<
   ctx: TContext,
   _event: SCXML.Event<TExpressionEvent>,
   delaysMap?: DelayFunctionMap<TContext, TEvent>
-): RaiseActionObject<TContext, TEvent> {
+): RaiseActionObject<TContext, TExpressionEvent, TEvent> {
   const meta = {
     _event
   };
@@ -231,13 +231,13 @@ export function send<
   TEvent extends EventObject,
   TSentEvent extends EventObject = AnyEventObject
 >(
-  event: Event<TSentEvent> | SendExpr<TContext, TEvent, TSentEvent>,
+  event: Event<AnyEventObject> | SendExpr<TContext, TEvent, AnyEventObject>,
   options?: SendActionOptions<TContext, TEvent>
 ): SendAction<TContext, TEvent, TSentEvent> {
   return {
     to: options ? options.to : undefined,
     type: actionTypes.send,
-    event: isFunction(event) ? event : toEventObject<TSentEvent>(event),
+    event: isFunction(event) ? event : toEventObject(event),
     delay: options ? options.delay : undefined,
     // TODO: don't auto-generate IDs here like that
     // there is too big chance of the ID collision
@@ -246,7 +246,7 @@ export function send<
         ? options.id
         : isFunction(event)
         ? event.name
-        : (getEventType<TSentEvent>(event) as string)
+        : (getEventType(event) as string)
   } as any;
 }
 
@@ -307,7 +307,7 @@ export function sendParent<
   TEvent extends EventObject,
   TSentEvent extends EventObject = AnyEventObject
 >(
-  event: Event<TSentEvent> | SendExpr<TContext, TEvent, TSentEvent>,
+  event: Event<AnyEventObject> | SendExpr<TContext, TEvent, AnyEventObject>,
   options?: SendActionOptions<TContext, TEvent>
 ): SendAction<TContext, TEvent, TSentEvent> {
   return send<TContext, TEvent, TSentEvent>(event, {
@@ -355,11 +355,9 @@ export function sendTo<
 export function sendUpdate<TContext, TEvent extends EventObject>(): SendAction<
   TContext,
   TEvent,
-  { type: ActionTypes.Update }
+  AnyEventObject
 > {
-  return sendParent<TContext, TEvent, { type: ActionTypes.Update }>(
-    actionTypes.update
-  );
+  return sendParent(actionTypes.update);
 }
 
 /**
@@ -735,7 +733,7 @@ export function resolveActions<TContext, TEvent extends EventObject>(
           machine.options.delays as any
         );
         if (predictableExec && typeof raisedAction.delay === 'number') {
-          predictableExec(raisedAction, updatedContext, _event);
+          predictableExec(raisedAction as any, updatedContext, _event);
         }
         return raisedAction;
       }
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -42,6 +42,7 @@ import {
   AnyActorRef,
   PredictableActionArgumentsExec,
   RaiseActionOptions,
+  NoInfer,
   BaseActionObject,
   LowInfer
 } from './types';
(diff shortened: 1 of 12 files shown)


packages/core/src/actions.ts
```diff
@@ -157,10 +158,14 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
...
-export function raise<TContext, TEvent extends EventObject>(
-  event: Event<TEvent>,
-  options?: RaiseActionOptions<TContext, TEvent>
-): RaiseAction<TEvent> {
+export function raise<
+  TContext,
+  TExpressionEvent extends EventObject,
+  TEvent extends EventObject = TExpressionEvent
+>(
+  event: NoInfer<TEvent>,
```
Andarist, line 166: previously this was typed as `Event<TEvent>`:
https://github.com/statelyai/xstate/blob/70f426a024616be2f0a77bcfb9ad9a74c87844d7/packages/core/src/types.ts#L90

This wasn't correct for events with payload. IIRC we've decided to ditch the idea of extracting "simple" events and allowing them to be used as strings. I think it would be OK-ish to introduce such a "breaking change" (type-level only) in a minor version. Let me know what do you think.

packages/core/src/types.ts
```diff
@@ -1171,9 +1192,12 @@ export enum ActionTypes {
   Choose = 'xstate.choose'
 }
 
-export interface RaiseAction<TEvent extends EventObject> {
+export interface RaiseAction<
+  TContext,
+  TExpressionEvent extends EventObject,
+  TEvent extends EventObject = TExpressionEvent
+> extends ActionObject<TContext, TExpressionEvent, TEvent> {
```
Andarist, line 1199: Adding this `extends` was one of the crucial changes here.

packages/core/src/actions.ts
davidkpiano: Yeah, that's fine. Worst-case scenario, it would only "break" code at the type-level (ignorable) but still work exactly the same.

davidkpiano:
Yeah, this might be too big of a change to introduce to v4. Is there a more limited change that we can make to improve the `raise` typings for now, like I introduced in the linked PR?

Andarist:
> Yeah, this might be too big of a change to introduce to v4.

Note that I've tried to ensure maximum (and reasonable) backward compatibility here. That's why I didn't reorder some type parameters and why I put default type arguments in all places that I touched.

The only unknown, from my PoV, is how this might impact projects "in the wild". Our types are so complex that it's super hard to predict this. I plan to add a lot of type-related tests in v5 but that's a future talk. 

The shape of the PR in itself - I think it's good. That doesn't mean that we have to ship this though. I'm really tempted though, just take a look at those DX improvements and correctness! 😅 

> Is there a more limited change that we can make to improve the raise typings for now, like I introduced in the linked PR?

The question here is - what is our aim with the type-related changes? If I'd know how to change this in a simpler way in v4 then I would do it differently right from the start. The changes in the linked PR are not correct - they just allow things in an unsafe way, so "accidentally" the errors go away. Arguably... the old types weren't that much better (I'd have to take another look to confirm that) so it's not necessarily like those changes introduce a regression (but again, I'd have to recheck that), maybe they just trade one incorrect/unsound thing for another incorrect/unsound thing.

davidkpiano:
@Andarist I'll take a deeper look at this. My main goal is to make the normal usage of `raise(...)` work instead of incorrectly causing type errors.

davidkpiano (approved):
Looks good, but can we add a `// @ts-expect-error` test?

packages/core/src/interpreter.ts
```diff
@@ -960,7 +961,7 @@ export class Interpreter<
       | RaiseActionObject<TContext, TEvent>
   ): void {
     const timerId = this.clock.setTimeout(() => {
-      if (sendAction.to) {
+      if ('to' in sendAction && sendAction.to) {
```
davidkpiano, line 964: Is this just to make TS happy?

packages/core/test/types.test.ts
```diff
@@ -211,87 +211,111 @@ describe('Nested parallel stateSchema', () => {
...
-              }
-            ]
-          }
-        },
-        morning: {},
-        lunchTime: {},
-        afternoon: {},
-        evening: {},
-        night: {}
+  // TODO: will be rmeoved in v5
```
davidkpiano, line 225:
```suggestion
  // TODO: will be removed in v5
```

packages/core/test/types.test.ts
```diff
@@ -472,7 +496,7 @@ describe('events', () => {
           type: 'FOO';
         }
       },
-      entry: raise('FOO')
+      entry: raise<any, any, any>('FOO')
```
davidkpiano, line 499: Can we get this to work with just `entry: raise('FOO')`? We shouldn't expect to write `<any, any, any>`

packages/xstate-react/test/useActor.test.tsx
```diff
@@ -127,14 +127,12 @@ describeEachReactMode('useActor (%s)', ({ render }) => {
...
-      actorRef?: ActorRefFrom<typeof childMachine>;
+      actorRef: ActorRefFrom<typeof childMachine>;
     }
 
     const machine = createMachine<Ctx>({
       initial: 'active',
-      context: {
-        actorRef: undefined
-      },
+      context: {} as Ctx,
```
davidkpiano, line 135: Why wouldn't `actorRef: undefined` work?

packages/xstate-react/test/useSelector.test.tsx
```diff
@@ -303,8 +304,8 @@ describeEachReactMode('useSelector (%s)', ({ suiteKey, render }) => {
     });
 
     const identitySelector = (value: any) => value;
-    const getSnapshot = (actor: ReturnType<typeof createActor>) =>
-      actor.latestValue;
+    const getSnapshot = (actor: ActorRef<any, any>) =>
+      (actor as any).latestValue;
```
davidkpiano, line 308: Why is `as any` needed here (and below)?

packages/core/test/types.test.ts
Andarist: It works without `any`s - it's just now what this test is about. The test title is this:
> should not use actions as possible inference sites 1

This test is specifically testing that we end up inferring `{ type: 'FOO' }` for the `TEvent` despite those `any`s here

packages/xstate-react/test/useSelector.test.tsx
Andarist: It's because this `.latestValue` is not actually part of the `ActorRef<any, any>`. `useSelector` actually works OK with such a thing but the problem is that this test grabs the reference to this actor from `state.context.childActor` and that is typed as `ActorRef<any, any>` and not as "`CustomActor`". So I really had to introduce casting **somewhere** here - and it made sense to me to just use it here. Ultimately, it doesn't really matter all that much

PR #3702, opened 2022-12-12
Refactor to improve readability

mellanslag-de:
Initially wanted to try if i can fix an issue, but ended up refactoring a bit to improve the readability at some places.
All methods still do the same, just rewrote a bit of the inner workings, to reduce the mental overhead to think about the issue in the future, and make the changes a bit cleaner.

Also deleted some seemingly unused methodes. I don't know if that was okay, i'll be happy to undo those changes if you tell me so.

### DIFF
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -154,7 +154,7 @@ export function toActivityDefinition<TContext, TEvent extends EventObject>(
  * Raises an event. This places the event in the internal event queue, so that
  * the event is immediately consumed by the machine in the current step.
  *
- * @param eventType The event to raise.
+ * @param event The event to raise.
  */
 export function raise<TContext, TEvent extends EventObject>(
   event: Event<TEvent>
@@ -450,12 +450,10 @@ export function resolveStop<TContext, TEvent extends EventObject>(
       ? { id: actorRefOrString }
       : actorRefOrString;
 
-  const actionObject = {
+  return {
     type: ActionTypes.Stop as const,
     activity: resolvedActorRef
   };
-
-  return actionObject;
 }
 
 /**
@@ -471,12 +469,6 @@ export const assign = <TContext, TEvent extends EventObject = EventObject>(
     assignment
   };
 };
-
-export function isActionObject<TContext, TEvent extends EventObject>(
-  action: Action<TContext, TEvent>
-): action is ActionObject<TContext, TEvent> {
-  return typeof action === 'object' && 'type' in action;
-}
 
 /**
  * Returns an event type that represents an implicit event that
@@ -665,14 +657,16 @@ export function resolveActions<TContext, TEvent extends EventObject>(
     ? [currentContext]
     : undefined;
 
-  const deferredToBlockEnd: Array<ActionObject<TContext, TEvent>> = [];
+  let deferredToBlockEnd: Array<ActionObject<TContext, TEvent>> = [];
 
-  function handleAction(actionObject: ActionObject<TContext, TEvent>) {
+  function resolveAction(
+    actionObject: ActionObject<TContext, TEvent>
+  ): Array<ActionObject<TContext, TEvent>> {
     switch (actionObject.type) {
       case actionTypes.raise: {
-        return resolveRaise(actionObject as RaiseAction<TEvent>);
+        return [resolveRaise(actionObject as RaiseAction<TEvent>)];
       }
-      case actionTypes.send:
+      case actionTypes.send: {
         const sendAction = resolveSend(
           actionObject as SendAction<TContext, TEvent, AnyEventObject>,
           updatedContext,
@@ -694,7 +688,8 @@ export function resolveActions<TContext, TEvent extends EventObject>(
           deferredToBlockEnd.push(sendAction);
         }
 
-        return sendAction;
+        return [sendAction];
+      }
       case actionTypes.log: {
         const resolved = resolveLog(
           actionObject as LogAction<TContext, TEvent>,
@@ -702,7 +697,7 @@ export function resolveActions<TContext, TEvent extends EventObject>(
           _event
         );
         predictableExec?.(resolved, updatedContext, _event);
-        return resolved;
+        return [resolved];
       }
       case actionTypes.choose: {
         const chooseAction = actionObject as ChooseAction<TContext, TEvent>;
@@ -779,7 +774,7 @@ export function resolveActions<TContext, TEvent extends EventObject>(
         );
 
         predictableExec?.(resolved, currentContext, _event);
-        return resolved;
+        return [resolved];
       }
       case actionTypes.assign: {
         updatedContext = updateContext(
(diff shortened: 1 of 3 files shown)


packages/core/src/actions.ts
```diff
@@ -665,14 +657,16 @@ export function resolveActions<TContext, TEvent extends EventObject>(
     ? [currentContext]
     : undefined;
 
-  const deferredToBlockEnd: Array<ActionObject<TContext, TEvent>> = [];
+  let deferredToBlockEnd: Array<ActionObject<TContext, TEvent>> = [];
```
mellanslag-de, line 660: Made non-const to allow resetting the array further down without hacks

packages/core/src/actions.ts
```diff
@@ -665,14 +657,16 @@ export function resolveActions<TContext, TEvent extends EventObject>(
...
 
-  function handleAction(actionObject: ActionObject<TContext, TEvent>) {
+  function resolveAction(
+    actionObject: ActionObject<TContext, TEvent>
+  ): Array<ActionObject<TContext, TEvent>> {
     switch (actionObject.type) {
       case actionTypes.raise: {
-        return resolveRaise(actionObject as RaiseAction<TEvent>);
+        return [resolveRaise(actionObject as RaiseAction<TEvent>)];
```
mellanslag-de, line 667: consistently return an Array<ActionObject> now to simplify the method handling

packages/core/src/actions.ts
```diff
@@ -789,9 +784,9 @@ export function resolveActions<TContext, TEvent extends EventObject>(
           !predictableExec ? currentState : undefined
         );
         preservedContexts?.push(updatedContext);
-        break;
+        return [];
```
mellanslag-de, line 787: Return empty array to not return undefined only in this case

packages/core/src/actions.ts
```diff
@@ -808,28 +803,34 @@ export function resolveActions<TContext, TEvent extends EventObject>(
...
-    let resolvedActions: Array<ActionObject<TContext, TEvent>> = [];
-
-    for (const action of block) {
-      const resolved = handleAction(action);
-      if (resolved) {
-        resolvedActions = resolvedActions.concat(resolved);
-      }
-    }
+  function resolveActionBlocks(
+    _actionBlocks: ActionObject<TContext, TEvent>[]
```
mellanslag-de, line 812: renamed because theres a variable with an equal name in the outer function. The overshadowing was noted in ESLint.
I probably would have preferred to move this method to the topmost level. But seems to not be the pattern elsewhere, so i decided to just rename the parameter.

packages/core/src/actions.ts
```diff
@@ -808,28 +803,34 @@ export function resolveActions<TContext, TEvent extends EventObject>(
...
+  ) {
+    const resolvedActions = flatten(_actionBlocks.map(resolveAction));
 
-    deferredToBlockEnd.forEach((action) => {
-      predictableExec!(action, updatedContext, _event);
-    });
-    deferredToBlockEnd.length = 0;
+    // predictableExec can not be undefined if deferredToBlockEnd has items
+    deferredToBlockEnd.forEach((action) =>
+      predictableExec?.(action, updatedContext, _event)
```
mellanslag-de, line 818: previously an exception would have been thrown if predictableExec was undefined.
That can never be the case, but maybe it was intended, in case code changes above would introduce a bug?

Then we should probably change it back. But disliked the 'type casting', thats why i wanted to change it

Andarist:
Thanks for taking the interest in the internal implementation! It's very much appreciated 👍 

I'm inclined to decline this PR at this moment though. The PR only touches some code structure and we currently have a second active branch (`next`) to which we merge back `main`. So while this improves some nit-level things in the codebase, it also adds work to resolve guaranteed conflicts with the `next` branch.

packages/core/src/utils.ts
```diff
@@ -74,28 +74,13 @@ export function getEventType<TEvent extends EventObject = EventObject>(
   event: Event<TEvent>
 ): TEvent['type'] {
   try {
-    return isString(event) || typeof event === 'number'
```
mellanslag-de, line 77: According to the types, it can never be a number anyway. ESLint had a warning with this.

packages/core/src/utils.ts
```diff
@@ -251,7 +237,7 @@ export function toStatePaths(stateValue: StateValue | undefined): string[][] {
     return [[stateValue]];
   }
 
-  const result = flatten(
```
mellanslag-de, line 254: ESLint warning fixed by returning directly

packages/core/src/utils.ts
```diff
@@ -530,12 +516,13 @@ export const symbolObservable: typeof Symbol.observable = (() =>
   (typeof Symbol === 'function' && Symbol.observable) ||
   '@@observable')() as any;
 
-// TODO: to be removed in v5, left it out just to minimize the scope of the change and maintain compatibility with older versions of integration paackages
+// TODO: to be removed in v5, left it out just to minimize the scope of the change and maintain compatibility with older
+// versions of integration packages
 export const interopSymbols = {
-  [symbolObservable]: function () {
```
mellanslag-de, line 535: also ESLint

tslint.json
```diff
@@ -6,14 +6,15 @@
     "trailing-comma": [false],
     "object-literal-sort-keys": false,
     "arrow-parens": false,
-    "quotemark": ["single"],
+    "quotemark": [true, "single"],
```
mellanslag-de, line 9: fixed to have the correct type according to https://palantir.github.io/tslint/rules/quotemark/

tslint.json
```diff
@@ -6,14 +6,15 @@
...
     "ordered-imports": [false],
     "variable-name": false,
     "object-literal-key-quotes": false,
     "interface-name": false,
     "member-ordering": false,
     "no-namespace": false,
-    "no-unused-expression": [true, "allow-fast-null-checks"]
+    "no-unused-expression": [true, "allow-fast-null-checks"],
+    "array-type": false
```
mellanslag-de, line 17: Saw that we use both (Array<> and []), so wanted to silence the warnings by explicitly allowing it.

davidkpiano:
> > Thanks for taking the interest in the internal implementation! It's very much appreciated 👍
> > I'm inclined to decline this PR at this moment though. The PR only touches some code structure and we currently have a second active branch (`next`) to which we merge back `main`. So while this improves some nit-level things in the codebase, it also adds work to resolve guaranteed conflicts with the `next` branch.
> 
> Oh, I see a lot changed there already. I can barely recognize the old code again. Then it indeed doesnt make sense at all to merge this, probably never. I guess i should have asked first, before spending my time on this^^
> 
> Thanks for the quick reply though!

Really happy that you took a look at it, though! 😅 

We can definitely use some of these refactor techniques/readability improvements in v5 when it becomes more ready for outside contributions, if you would like to help with that; we'd greatly appreciate it.

Andarist:
Yeah - I'd also like to emphasize that this is really appreciated. There are not many people contributing to the codebase and each contribution is very much valued. It's just a slightly wrong time for this particular PR.

I'm going to close this but if you'd like to contribute to XState - let's talk! I'm sure that we could find some things for the community to work on :)

PR #3708, opened 2022-12-16
Convert `State["done"]` and `State["output"]` to simple properties

Andarist:
Related to https://github.com/statelyai/xstate/pull/3455/files#r967738767


packages/core/src/State.ts
```diff
@@ -43,6 +43,22 @@ export function isStateConfig<
   return 'value' in state && '_event' in state;
 }
 
+function getOutput<TContext, TEvent extends EventObject>(
```
Andarist, line 46: I'm not exactly sure if this should be computed here - but for the time being let's keep it here.

packages/core/src/State.ts
```diff
@@ -163,6 +190,14 @@ export class State<
...
 
     this.value = getStateValue(machine.root, this.configuration);
+    this.done = this.configuration.length
+      ? isInFinalState(this.configuration)
+      : false;
+    this.output =
+      config.output ??
+      (this.done
+        ? getOutput(this.configuration, this.context, this._event)
+        : undefined);
```
Andarist, line 200: This actually can't be computed in the constructor right now because it might lead to the `data` functions being called multiple times and thus this test fails:
https://github.com/statelyai/xstate/blob/10e07cf4c27913a1b9e49916fb76143fad8bb5e7/packages/core/test/final.test.ts#L198-L219

This happens because we `.clone()` the state a lot while transitioning.

This further tells me that perhaps we shouldn't be computing this stuff here, in this class, but rather just provide this as an argument.

PR #3710, opened 2022-12-16
Provide precomputed `State["done"]` and `State["output"]`

Andarist:
supersedes https://github.com/statelyai/xstate/pull/3708

This way:
- we avoid using getter for those things
- they are automatically memoized (since they are not recomputed with getters)
- they become regular enumerable properties
- we move them out of State to keep it as "dumb" as possible when it comes to those two (IMHO it's fine for State to call into machine to return a value for `state.can()` but other than that it should just be a simple state container, without much of its own logic)


davidkpiano (approved):
Tests pass 🚢

Issue #3617, opened 2022-09-26
Bug: Sending an event to a service on `exit` triggers an error


Mxime:
### Description

Since I use `predictableActionArguments: true` as advised in the Actions [documentation](https://xstate.js.org/docs/guides/actions.html), I get an error when I try to send an event from `exit`.

```error
Error: Unable to send event to child 'my-service' from service 'toggle'.
```

### Expected result

I expect `my-event` to have been received by `my-service` upon exiting a state

### Actual result

```error
Error: Unable to send event to child 'my-service' from service 'toggle'.
```

### Reproduction

https://codesandbox.io/s/thirsty-ritchie-yfpzis

### Additional context

xstate 4.33.6

Andarist:
Thank you for the report and the repro case. I think that I already see what the problem is, I need to figure out a fix for this though.

GuillaumeAmat:
@Andarist did you figure a possible solution here? :pray:

Andarist:
Sorry for the delay, I'll take care of this next week

Mxime:
It seems [we still face](https://codesandbox.io/s/thirsty-ritchie-yfpzis) the issue with `4.35.0`

davidkpiano:
The problem has been identified and we will fix it soon.

PR #3712, opened 2022-12-19
Add failing test for #3617

davidkpiano:
This PR adds a failing test for #3617 

@Andarist The main issue is that stopping invoked actors should occur _after_ exit actions are executed, according to the SCXML algorithm. This order is correct up until the point that `resolveActions` is called, at which point the `send(...)` action is deferred for some reason, which causes it to execute after stopping the actor.

SCXML algorithm for reference:
```
procedure exitStates(enabledTransitions):
    statesToExit = computeExitSet(enabledTransitions)           
    for s in statesToExit:
        statesToInvoke.delete(s)
    statesToExit = statesToExit.toList().sort(exitOrder)
    for s in statesToExit:
        for h in s.history:
            if h.type == "deep":
                f = lambda s0: isAtomicState(s0) and isDescendant(s0,s) 
            else:
                f = lambda s0: s0.parent == s
            historyValue[h.id] = configuration.toList().filter(f)
    for s in statesToExit:
        for content in s.onexit.sort(documentOrder):
            executeContent(content)
        for inv in s.invoke:
            cancelInvoke(inv)
        configuration.delete(s)
```

### DIFF
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -2764,3 +2764,48 @@ describe('assign action order', () => {
     }
   );
 });
+
+// https://github.com/statelyai/xstate/issues/3617
+it('should only stop invoked services after exit actions have been executed', (done) => {
+  const machine = createMachine(
+    {
+      id: 'toggle',
+      initial: 'active',
+      predictableActionArguments: true,
+      states: {
+        inactive: {
+          on: { TOGGLE: 'active' }
+        },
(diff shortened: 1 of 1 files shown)


Andarist:
I prepared a PR fixing this: https://github.com/statelyai/xstate/pull/3713 . This PR includes this commit (thanks for creating the test case!) and targets `main` directly so I'm gonna close this PR

PR #3717, opened 2022-12-20
Remove `sync` option

Andarist:
This basically just removes the associated tests as the target branch already removed the implementation


mellson (approved):
Looks good from my perspective.

PR #3719, opened 2022-12-20
Defer stopping actors

Andarist:
As per comment [here](https://github.com/statelyai/xstate/pull/3455/files#r968358269):
> This tests a very specific situation - with no other machines in the picture etc so a grandchild can't be just stopped for any other reason than because its parent here (child) gets stopped.

> No race condition is possible here even if we assume that sending events is asynchronous because this test tests that events sent from child to grandchild are sent "in order". It also tests that a sent event is sent before an implicit event that informs the grandchild to stop. In other words, it tests that we can send events from a machine to its children before notifying them about them being stopped and it tests that they can process that event (because the stop notification has not been delivered to them yet, or because it stays queued after that other event).

---

The issue here was that even though actions were resolved and executed correctly as per the algorithm quoted recently by @davidkpiano [here](https://github.com/statelyai/xstate/pull/3712#issue-1502992414) - we were first executing the executable content (exit actions) and then stopping children.

The problem is that sending events is currently "deferred" to enable `.getSnapshot()` reads in the receiver (the receiver can read the updated state of the origin in its transitions for the received event). However, stop actions were not deferred like this - and thus the sent events could never reach their destination (as by the time they were **actually** executed the destination/target actor was already stopped).

### DIFF
--- a/packages/core/src/exec.ts
+++ b/packages/core/src/exec.ts
@@ -157,8 +157,9 @@ function getActionFunction<TState extends AnyState>(
       const { actor } = (action as StopActionObject).params;
 
       if (actor) {
-        actor.stop?.();
-        delete state.children[actor.id];
+        actorCtx.defer?.(() => {
+          actor.stop?.();
+        });
       }
     },
     [actionTypes.log]: (_ctx, _e, { action }) => {
--- a/packages/core/test/actor.test.ts
+++ b/packages/core/test/actor.test.ts
@@ -1287,7 +1287,7 @@ describe('actors', () => {
             fromPromise(
               () =>
                 new Promise<number>((_, rej) => {
-                  setTimeout(() => rej(errorMessage), 1000);
+                  setTimeout(() => rej(errorMessage), 1);
                 })
             ),
             'test'
--- a/packages/core/test/predictableExec.test.ts
+++ b/packages/core/test/predictableExec.test.ts
@@ -259,11 +259,11 @@ describe('predictableExec', () => {
       initial: 'active',
       context: ({ spawn }) => {
         const localId = ++invokeCounter;
-        actual.push(`start ${localId}`);
 
         return {
           actorRef: spawn(
             fromCallback(() => {
+              actual.push(`start ${localId}`);
               return () => {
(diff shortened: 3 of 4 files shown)


packages/core/src/exec.ts
```diff
@@ -157,8 +157,9 @@ function getActionFunction<TState extends AnyState>(
       const { actor } = (action as StopActionObject).params;
 
       if (actor) {
-        actor.stop?.();
-        delete state.children[actor.id];
```
Andarist, line 161: I'm not super confident about removing this `delete` but the current tests pass

packages/core/test/actor.test.ts
```diff
@@ -1287,7 +1287,7 @@ describe('actors', () => {
             fromPromise(
               () =>
                 new Promise<number>((_, rej) => {
-                  setTimeout(() => rej(errorMessage), 1000);
+                  setTimeout(() => rej(errorMessage), 1);
```
Andarist, line 1290: just a small speed boost - there was no reason here to wait for the whole second

packages/core/test/predictableExec.test.ts
```diff
@@ -259,11 +259,11 @@ describe('predictableExec', () => {
       initial: 'active',
       context: ({ spawn }) => {
         const localId = ++invokeCounter;
-        actual.push(`start ${localId}`);
```
Andarist, line 262: as actors are created at a different time than they are started, I moved those pushes to the start's site to fix how the expectations were recorded in this file. Those tests were introduced in v4 where there was no distinction between those two times, so the previous version was OK there.

packages/core/src/exec.ts
davidkpiano: This might cause a memory leak

davidkpiano (approved):
Approved with one small concern

packages/core/src/exec.ts
Andarist: In such a case, we should be able to verify this with a test - it's not a leak that could happen through closure or anything like that. If a reference to a child stays on the `state` then we can write an assertion for this thing

PR #3720, opened 2022-12-20
Remove a custom `done` callback from the `Observer`

Andarist:
This didn't end up being used by anything


davidkpiano (changes requested):
This is still necessary. There's a difference between `complete()` (an observable is no longer producing values) and `done()` (an observable-ish has provided a specific output value).

davidkpiano:
Also @Andarist the tests fail when this is removed.

davidkpiano:
@Andarist Tests still fail 😢

davidkpiano (approved):

PR #3722, opened 2022-12-20
Remove `ActorContext['observers']`

Andarist:


packages/core/src/interpreter.ts
```diff
@@ -154,8 +154,7 @@ export class Interpreter<
       },
       defer: (fn) => {
         this._deferred.push(fn);
-      },
-      observers: this.observers
```
Andarist, line 158: this is no longer managed by behaviors so they don't need to receive it

davidkpiano (approved):

PR #3723, opened 2022-12-20
Add tests for history state's default target when the history state is the initial state

Andarist:


packages/core/src/stateUtils.ts
```diff
@@ -649,7 +649,7 @@ function resolveHistoryTarget<
   }
   return normalizedTarget.map((t) =>
     typeof t === 'string'
-      ? getStateNodeByPath<TContext, TEvent>(stateNode, t)
+      ? getStateNodeByPath<TContext, TEvent>(stateNode.parent!, t)
```
Andarist, line 652: `stateNode` here is the history state itself - since history states are pseudo-states and can't contain any children states, it doesn't make sense to `getStateNodeByPath` within them. As per [examples in the docs](https://xstate.js.org/docs/guides/history.html#history-state-configuration) the `.target` value targets states within siblings of the history state

davidkpiano (approved):

PR #3726, opened 2022-12-26
[v5] Refactor actions

davidkpiano:
This PR does a bunch of things:

- Removes side-effects from `getInitialState` 
- Refactors actions so that their implementations/executions are within the dynamic action object instead of `stateUtils.ts`
- Adds tests and fixes some tests
- Deletes `ExecutableAction.ts`
- Deletes `exec.ts`
- Changes the `start()` function for behaviors to return a state and be the place where side-effects are executed (as suggested by @andarist)
- Breaking: changes the `type` of function actions to be `’xstate.action’` instead of the function name
- Refactors some tests to be more robust (avoid testing deep implementation details)
- Somehow fixes React strict mode execution to call things the expected number of times (e.g. 2 instead of 4)


packages/xstate-test/test/index.test.ts
```diff
@@ -202,7 +202,7 @@ it('prevents infinite recursion based on a provided limit', () => {
 });
 
 // TODO: have this as an opt-in
-it('executes actions', async () => {
+it.skip('executes actions', async () => {
```
Andarist, line 205: what changed here and why?

davidkpiano: See your comment:

```ts
// packages/xstate-test/src/machine.ts

export function executeAction(
  actionObject: BaseActionObject,
  state: AnyState
): void {
  // TODO: this is a hack, it doesn't correctly resolve actions
  // what gets passed as `action` here is probably also not correct
  if (typeof (actionObject as any)._exec === 'function') {
    (actionObject as any)._exec(state.context, state.event, {
      _event: state._event,
      action: actionObject,
      state
    });
  }
}
```

davidkpiano: We need to determine if executing actions is actually valuable in a testing scenario; I don't think it is. Tests verify; the only execution they should be responsible for is in the execution of external events.

Andarist: > Tests verify; the only execution they should be responsible for is in the execution of external events.

Makes sense, although then we should somehow just disallow actions in those test machine configs.

But in such a case, we should just remove this test and the mentioned hacky code. WDYT?

davidkpiano: Yes, let's remove it.

packages/xstate-react/src/useMachine.ts
```diff
@@ -118,7 +119,7 @@ export function useMachine<TMachine extends AnyStateMachine>(
         ? ((service.behavior as AnyStateMachine).createState(
             rehydratedState
           ) as any)
-        : storeSnapshot
+        : undefined
```
Andarist, line 122: review note: `storeSnapshot` is what is internally returned by `this.getInitialState()` (as internally that initial state is cached). The difference is now in what `.start(arg)` might do based on the explicit argument - as the rehydrated state would go through `this.behavior.restoreState`. This in turn might create a copy of the state argument - maybe change its values somehow etc.

packages/core/test/predictableExec.test.ts
```diff
@@ -8,7 +8,9 @@ describe('predictableExec', () => {
 
     const machine = createMachine({
       initial: 'a',
-      context: {},
+      context: {
+        assigned: false
+      },
```
Andarist, line 13: This isn't used by the test anyhow
```suggestion
      context: {},
```

packages/core/test/predictableExec.test.ts
```diff
@@ -20,7 +22,9 @@ describe('predictableExec', () => {
             },
             assign(() => {
               actual.push('assign');
-              return {};
+              return {
+                assigned: true
+              };
```
Andarist, line 27: This isn't used by the test anyhow
```suggestion
              return {};
```

packages/core/test/invoke.test.ts
```diff
@@ -715,15 +715,15 @@ describe('invoke', () => {
         }
       });
 
-      const expectedStateValue = 'two';
       let currentState: AnyState;
```
Andarist, line 718: This would be even easier to follow:
```ts
      const actor = interpret(mainMachine).start();

      expect(actor.getSnapshot().value).toBe('two');
```

Note that this also requires removing the `done` callback argument

packages/core/test/interpreter.test.ts
```diff
@@ -1843,3 +1843,49 @@ it(`should execute entry actions when starting the actor after reading its snaps
 
   expect(spy).toHaveBeenCalled();
 });
+
+it('ref', (done) => {
```
Andarist, line 1847: those two added test cases could use meaningful titles 😉 we should also move them to the main `describe` block - the auto-sorting of the test summary is super confusing. Even though those are at the end of the file, they are on the top in the test summary:
```
 PASS  packages/core/test/interpreter.test.ts
  ✓ should execute entry actions when starting the actor after reading its snapshot first
  ✓ ref (1 ms)
  ✓ ref2
  interpreter
    ✓ can cancel a delayed event (1 ms)
    ✓ can cancel a delayed event using expression to resolve send id (202 ms)
    ✓ should throw an error if an event is sent to an uninitialized interpreter if { deferEvents: false } (5 ms)
    ✓ should not throw an error if an event is sent to an uninitialized interpreter if { deferEvents: true }
    ✓ should not throw an error if an event is sent to an uninitialized interpreter (default options) (1 ms)
    ✓ should defer events sent to an uninitialized service
    ✓ should throw an error if initial state sent to interpreter is invalid (1 ms)
    ✓ should not update when stopped (23 ms)
    ✓ should be able to log (log action) (1 ms)
    ✓ should be able to log event origin (log action) (2 ms)
    ✓ should receive correct _event (log action)
    ✓ shouldn't execute actions when reading a snapshot of not started actor
    initial state
      ✓ immediately notifies the listener with the initial state and event (3 ms)
      ✓ .initialState returns the initial state (1 ms)
      ✓ initially spawned actors should not be spawned when reading initial state (102 ms)
      ✓ executes actions from a restored state (2 ms)
      ✓ should not execute actions that are not part of the actual persisted state (1 ms)
[...]
```

packages/xstate-test/test/index.test.ts
davidkpiano: @Andarist I will remove the test, but the executing action changes need to be made upstream on the `main` branch.

packages/core/test/interpreter.test.ts
davidkpiano: Sorry about that - these tests were useful to help me fix some bugs in `@xstate/react` without going through the React plumbing 😂 I can remove them if you want me to.

packages/core/test/actions.test.ts
```diff
@@ -2371,7 +2372,11 @@ describe('sendTo', () => {
       )
     });
 
-    interpret(parentMachine).start();
+    interpret(parentMachine)
+      .onTransition((state) => {
+        state;
+      })
+      .start();
```
Andarist, line 2379:
```suggestion
    interpret(parentMachine).start();
```

packages/core/test/actions.test.ts
```diff
@@ -2341,7 +2338,11 @@ describe('sendTo', () => {
       entry: sendTo((ctx) => ctx.child, { type: 'EVENT' })
     });
 
-    interpret(parentMachine).start();
+    interpret(parentMachine)
+      .onTransition((state) => {
+        state;
+      })
+      .start();
```
Andarist, line 2345:
```suggestion
    interpret(parentMachine).start();      
```

packages/core/test/actions.test.ts
```diff
@@ -1663,25 +1674,11 @@ describe('actions config', () => {
...
-    const inactiveState = anonMachine.transition(initialState, 'EVENT');
-
-    expect(inactiveState.actions.length).toBe(2);
-
-    inactiveState.actions.forEach((action) => {
-      if ('execute' in action) {
-        (action as any).execute(inactiveState);
-      }
-    });
+    actor.send('EVENT');
```
Andarist, line 1681:
```suggestion
    actor.send({ type: 'EVENT' });
```

packages/core/test/actions.test.ts
```diff
@@ -1474,7 +1485,7 @@ describe('initial actions', () => {
     }
   });
 
-  it.skip('should support initial actions', () => {
+  it('should support initial actions', () => {
     // TODO: fix initial state actions on root node
```
Andarist, line 1489: is this comment still relevant? I don't see anything in the `machine` that would be missing from this expectation
```suggestion
```

packages/core/test/interpreter.test.ts
Andarist: It might be worth keeping the first one (with a proper title etc) - the second one is based on a somewhat hacky implementation within `@xstate/react` since it changes the `.status` "manually". That probably shouldn't be something that people should do. I hope to refactor this in the `@xstate/react` too. So I'd say that the second one can be removed.

packages/xstate-test/test/index.test.ts
davidkpiano: For reference, this is removed here: https://github.com/statelyai/xstate/pull/3735

packages/core/test/interpreter.test.ts
```diff
@@ -1846,20 +1846,35 @@ describe('interpreter', () => {
...
-    })
-  );
+    const actorRef = interpret(
+      createMachine({
+        entry: spy
+      })
+    );
 
-  actorRef.getSnapshot();
+    actorRef.getSnapshot();
```
Andarist, line 1859: To make it more obvious when this is supposed to be called here:
```suggestion
    actorRef.getSnapshot();
    expect(spy).not.toHaveBeenCalled();
```

packages/core/src/stateUtils.ts
```diff
@@ -1497,15 +1503,20 @@ export function resolveActionsAndContext<
...
   const raiseActions: Array<RaiseActionObject<TEvent>> = [];
-  let { context } = currentState;
-  let updatedContext = context;
+  let intermediateState = currentState;
+
+  function handleAction(action: BaseActionObject): void {
+    resolvedActions.push(action);
+    if (actorCtx?.self.status === ActorStatus.Running) {
+      action.execute?.(actorCtx!);
+      delete action.execute;
```
Andarist, line 1517: Hmm, kinda hacky... but let's roll with it for now

davidkpiano: Yeah. This really only applies to restoring a state & starting from an initial state where actions need to be executed.

davidkpiano, line 1517:
```suggestion
      // TODO: this is hacky; re-evaluate
      delete action.execute;
```

packages/core/src/actions/stop.ts
```diff
@@ -31,21 +36,40 @@ export function stop<
...
   return createDynamicAction(
-    stopActionType,
     {
-      actor
+      type: stopActionType,
+      params: {
+        actor
+      }
     },
-    ({ params, type }, context, _event, { state }) => {
```
Andarist, line 38: removal of `params` from here feels a little bit clunky at times since we need to reach into the closure and at times we end up duplicating things like here:
https://github.com/statelyai/xstate/blob/7106e9daa44cf32891a34a2cdce75881c3048349/packages/core/src/actions/send.ts#L58-L81

If we'd bring params back as a parameter here then we could even move those final "action handlers" out of the factories.
```ts
// pre
const assign = (assignment) => {
  return createDynamicAction(
    { type: "assign", params: { assignment } },
    (event, { state }) => {
      //...
    }
  );
};

// post
const assignHandler = (event, { state }) => {
  //...
};

const assign = (assignment) => {
  return createDynamicAction(
    { type: "assign", params: { assignment } },
    assignHandler
  );
};
```

packages/core/src/StateMachine.ts
```diff
@@ -404,19 +428,12 @@ export class StateMachine<
...
   ): State<TContext, TEvent, TResolvedTypesMeta> {
     const restoredState = isStateConfig(state)
       ? this.resolveState(state as any)
       : this.resolveState(State.from(state as any, this.context, this));
 
-    if (actorCtx) {
-      for (const action of restoredState.actions) {
-        execAction(action, restoredState, actorCtx);
-      }
-    }
```
Andarist, line 418: Wha's the current status of executing with restored states? What's the default expected result? Should entry actions be called? Should actions contained in `restoredState.actions` be restored?

davidkpiano: It depends, but we shouldn't _disallow_ actions from being re-executed, since this is a common expectation (client-side, not server-side).

Andarist: I'm OK with allowing that - it's just not clear to me what we are picking as default and how one can opt-into the non-default behavior.

Andarist (approved):

Issue #3350, opened 2022-05-28
Bug: xstate/actors in v5 alpha can't be imported from .mjs


tivac:
### Description

Trying to import `xstate/actors` from an ESM context in node leads to an error. Using `require("xstate/actors")` works fine but is obviously weird in ESM mode.

Here's an example file t

### Expected result

I expect to be able to load all the xstate entrypoints no matter if I'm using `require()` or `import`.

### Actual result

Here's the `.mjs` file I created to test (also available as a [gist](https://gist.github.com/tivac/b6f58329ba3a90db30dd9e30873a221a))

```js
import * as actors from "xstate/actors";

console.log(actors);
```

and the resulting error from node

```bash
$ node xstate-actors.mjs
node:internal/process/esm_loader:94
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import 'C:\Users\pcavi\Documents\GitHub\xstate-component-tree\node_modules\xstate\actors' is not supported resolving ES modules imported from C:\Users\pcavi\Documents\GitHub\xstate-component-tree\xstate-actors.mjs
Did you mean to import xstate-component-tree/node_modules/xstate/actors/dist/xstate-actors.cjs.js?
    at new NodeError (node:internal/errors:372:5)
    at finalizeResolution (node:internal/modules/esm/resolve:433:17)
    at moduleResolve (node:internal/modules/esm/resolve:1009:10)
    at defaultResolve (node:internal/modules/esm/resolve:1218:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:580:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:294:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_UNSUPPORTED_DIR_IMPORT',
  url: 'file:///C:/Users/pcavi/Documents/GitHub/xstate-component-tree/node_modules/xstate/actors'
}
```

### Reproduction

https://gist.github.com/tivac/b6f58329ba3a90db30dd9e30873a221a

### Additional context

`npx envinfo --system --npmPackages xstate --binaries --markdown`

## System:
 - OS: Windows 10 10.0.22000
 - CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
 - Memory: 20.48 GB / 31.92 GB
[...]

PR #3727, opened 2022-12-27
Added `package.json#exports` to all packages

Andarist:
fixes https://github.com/statelyai/xstate/issues/3350

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,10 @@
     ],
     "globals": {
       "react": "React"
+    },
+    "exports": true,
+    "___experimentalFlags_WILL_CHANGE_IN_PATCH": {
+      "exports": true
     }
   },
   "scripts": {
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -13,6 +13,10 @@
     ],
     "globals": {
       "react": "React"
+    },
+    "exports": true,
+    "___experimentalFlags_WILL_CHANGE_IN_PATCH": {
```
davidkpiano, line 18: What's this for?

davidkpiano (approved):

package.json
Andarist: It's an opt-in flag that makes it possible to use the `exports: true` setting. `package.json#exports` can be convoluted and can be easily screwed up. We've done our best with Emma to implement something that works OK (while not allowing some things that people expected from us since they were more tricky and can't be safely enabled by all of the existing projects) but we couldn't be 100% sure that it would work with all of the available bundlers, node, etc.

So we've shipped this under an opt-in flag so it's more apparent that something could potentially break for the package consumers. I think that we did a pretty good job here though - considering that Emotion is using this for several months already and we didn't get any issue reports about stuff being broken (lately I found a bug in Vite's resolver but that's definitely their bug and not ours).

PR #3629, opened 2022-10-04
feat(test): add tag support

tivac:
https://statelyai.canny.io/xstate/p/xstatetest-support-testing-against-tags

Based off of the branch for #3446 per @davidkpiano [on Discord](https://discord.com/channels/795785288994652170/801127053092716596/1026925662843457546)

Seems to be working fine, except that I can't figure out the correct typing for the `TestModel.ts` changes ಠ_ಠ

No matter what I try it refused to believe that `tags` is a property on `TState`, so I think I'm missing something around how that should be used/set-up.

PR #3730, opened 2022-12-28
feat(test): add tag support

tivac:
(re-creating #3629 since it was based on a PR branch and got auto-closed)

Fixed the typing issue in that one by using an `any`, would love a better fix if anybody sees it.


davidkpiano:
We can now use `getShortestPaths({ toState })` more flexibly for this.

Doing some cleanup 🧹

PR #3733, opened 2022-12-30
Fixed initial state cache eviction in the `Interpreter`

Andarist:


packages/core/src/interpreter.ts
```diff
@@ -169,7 +169,7 @@ export class Interpreter<
   private _deferred: Array<(state: any) => void> = [];
 
   public getInitialState(): InternalStateFrom<TBehavior> {
-    return memo(this, 'init', () => {
+    return memo(this, 'initial', () => {
```
Andarist, line 172: this has to match the key that we evict here:
https://github.com/statelyai/xstate/blob/61e873448c45a2eb8d53b928c69962f18d086fe3/packages/core/src/interpreter.ts#L363

PR #3735, opened 2022-12-31
Remove test action execution

davidkpiano:
This PR removes the ability to execute actions in the upcoming `@xstate/test`. Executing side-effects should only be done either within the event execution or asserting the state (if needed).

@Andarist Not sure how to handle the alpha changeset for this one


Andarist:
> @Andarist Not sure how to handle the alpha changeset for this one

It's a tricky one. It would be great to get `@xstate/test` out of the alpha phase. For the time being... hm... I would probably add a "hidden" changeset here - just create one and prepend the file name with a dot.

davidkpiano:
@Andarist Since it's prerelease, I'll just omit the changeset.

Andarist (approved):

PR #3736, opened 2023-01-02
[Documentation] - Service Schema Typescript

JayeshVP24:
Added a note about how service schema accepts only type with data key. But while writing the service function, we have to return the type assigned to data key. This misinterpretation took me hours to realize.

having this specifically mentioned in the documentation would help


davidkpiano:
Thanks for your contribution; this has since changed.

PR #3737, opened 2023-01-02
Tweak logic around stopping actors

Andarist:


packages/core/src/stateUtils.ts
```diff
@@ -1665,7 +1665,7 @@ function stopStep(
   for (const stateNode of nextState.configuration.sort(
     (a, b) => b.order - a.order
   )) {
-    actions.push(...stateNode.definition.exit);
```
Andarist, line 1668: This was quite a 🦋 effect. 

1. `definition` is a getter: https://github.com/statelyai/xstate/blob/68b24d7c4ce7fa8920edab0a85c8d6c1e46c83ea/packages/core/src/StateNode.ts#L211
2. each `StateNode` was reading there `context` of the machine: https://github.com/statelyai/xstate/blob/68b24d7c4ce7fa8920edab0a85c8d6c1e46c83ea/packages/core/src/StateNode.ts#L216
3. `StateMachine['context']` is a getter as well: https://github.com/statelyai/xstate/blob/68b24d7c4ce7fa8920edab0a85c8d6c1e46c83ea/packages/core/src/StateMachine.ts#L90-L92

This means that we were calling "lazy context" functions like crazy~

packages/core/src/actions/invoke.ts
```diff
@@ -94,17 +94,16 @@ export function invoke<
           return;
         }
         ref._parent = interpreter; // TODO: fix
-        actorCtx.defer?.((state: AnyState) => {
+        actorCtx.defer(() => {
+          if (actorRef.status === ActorStatus.Stopped) {
+            return;
+          }
```
Andarist, line 100: IMHO it's much better to use the `actorRef` from the outer scope here instead of looking this up in the `state`. I don't currently have a failing test case for the old code (and maybe there is none to be created) but, on principle, it's way more error-prone to look this up by the key/id. 

The intention of this function is to start this concrete `actorRef` and not an `actorRef` that happens to have the same `id` at the time this deferred function gets finally called.

davidkpiano (approved):
Nice!

packages/core/src/actions/invoke.ts
davidkpiano: I was looking it up on the `state` before to "emulate" `invokeStates` from `statesToInvoke` in the SCXML algorithm, but this is cleaner.

PR #3740, opened 2023-01-02
Small cleanup of internal implementations of actors

Andarist:


packages/core/src/actors.ts
```diff
@@ -225,7 +225,7 @@ export function fromPromise<T>(
     getInitialState: () => {
       return {
         canceled: false,
-        status: 'pending',
+        status: 'active',
```
Andarist, line 228: this just unified the name of the status with the name used by observable actors

packages/core/src/actors.ts
```diff
@@ -244,25 +251,18 @@ export function fromObservable<T, TEvent extends EventObject>(
...
+
       if (state.canceled) {
         return state;
       }
 
       switch (_event.name) {
         case nextEventType:
+          state.data = event.data.data;
+
           self._parent?.send(
```
Andarist, line 266: Those `self._parent.send` calls in this file should be wrapped with `actorCtx.defer` - at the same time we need to add tests for this. Those tests should do smth like:
```js
onDone: (ctx, ev) => {
  expect(child.getSnapshot()).toEqual(ev.data)
}
```

packages/core/src/actors.ts
```diff
@@ -404,9 +405,11 @@ export function fromEventObservable<T extends EventObject>(
           self.send({ type: completeEventType });
         }
       });
+
       return state;
     },
-    getSnapshot: (_) => undefined
+    getSnapshot: (_) => undefined,
+    getStatus: (state) => state
```
Andarist, line 412: This is likely a bug fix for `fromEventObservable`. I need to write a test for this case

packages/core/src/actors.ts
```diff
@@ -244,68 +251,56 @@ export function fromObservable<T, TEvent extends EventObject>(
...
-                data: _event.data.data
-              },
-              { origin: self }
-            )
-          );
-          return { ...state, data: event.data.data };
+          state.data = event.data.data;
+          // match the exact timing of events sent by machines
+          // send actions are not executed immediately
+          defer(() => {
```
Andarist, line 267: Technically this isn't needed right now. Things work OK with this PR without this `defer` - that's because in this PR we start mutating `state` directly. If the `state` would be updated in an immutable way (like before this PR) then this `defer` would be very much needed to make the added test pass:
https://github.com/statelyai/xstate/pull/3740/files#diff-8ecfc4c99852fea354bfd5cf3c0a171ff48556ef07a9fd9d588aceaae25969f5R440

I still think though that it's better to have this `defer` here in place - to match the timing of things here as close to the timing within machines

davidkpiano (approved):

PR #3743, opened 2023-01-03
[v5] Recursive persistence

davidkpiano:


packages/xstate-react/src/useMachine.ts
```diff
@@ -27,17 +25,13 @@ const isEqual = (prevState: AnyState, nextState: AnyState) => {
...
   /**
    * If provided, will be merged with machine's `context`.
    */
+  // TODO: deprecate & remove context as option
   context?: Partial<TContext>;
-  /**
-   * The state to rehydrate the machine to. The machine will
-   * start at this state instead of its `initialState`.
-   */
-  state?: StateConfig<TContext, TEvent>;
```
Andarist, line 40: I feel like this had much better DX - perhaps the option should be renamed though (`persistedState`?). Having to call `.at` (or any other method used for this) should be hidden from the consuming sites.

I feel like this is a similar story to `input`. The rehydrated state is the `interpret`'s argument and it's Interpreter's job to correctly resolve it internally through ActorBehavior's APIs. And since `useMachine`/`useInterpret` correlates closely to the `interpret` call - I feel that providing this argument to both of them should look very similar. And it does look similar now - it's just that I propose making it the `interpret`'s argument.

packages/core/src/interpreter.ts
```diff
@@ -281,13 +282,9 @@ export class Interpreter<
   }
 
   /**
-   * Starts the interpreter from the given state, or the initial state.
-   * @param initialState The state to start the statechart from
+   * Starts the interpreter from the initial state
    */
-  public start(
-    // TODO: remove this argument
-    initialState?: InternalStateFrom<TBehavior> | StateValue
```
Andarist, line 289: This change to be documented in a changeset

packages/xstate-react/src/useMachine.ts
Andarist: It's especially important in React since we shouldn't pass new machines to the `useMachine` on each render and that's exactly what happens with code like this:
```ts
useMachine(machine.at(state))
```

packages/core/src/StateMachine.ts
```diff
@@ -307,13 +313,21 @@ export class StateMachine<
     return transitionNode(this.root, state.value, state, _event) || [];
   }
 
+  public preInitialState:
```
Andarist, line 316:
```suggestion
  prviate _preInitialState:
```

packages/core/src/StateMachine.ts
```diff
@@ -307,13 +313,21 @@ export class StateMachine<
...
   /**
    * The initial state _before_ evaluating any microsteps.
    * This "pre-initial" state is provided to initial actions executed in the initial state.
    */
   private getPreInitialState(
     actorCtx: ActorContext<any, any> | undefined
   ): State<TContext, TEvent, TResolvedTypesMeta> {
+    if (this.preInitialState) {
+      return this.preInitialState;
+    }
```
Andarist, line 329: why do we have to cache it now?

packages/core/src/State.ts
```diff
@@ -288,3 +289,28 @@ export function cloneState<TState extends AnyState>(
...
+  const childrenJson: any = {};
+
+  for (const key in children) {
+    childrenJson[key] = children[key].getPersistedState?.();
+  }
+
+  return {
+    ...jsonValues,
+    children: childrenJson,
+    persisted: true
```
Andarist, line 314: q: what's the purpose of this flag? I couldn't find any code that would use this

packages/core/src/types.ts
```diff
@@ -1814,7 +1817,8 @@ export interface ActorContext<TEvent extends EventObject, TSnapshot> {
 export interface ActorBehavior<
   TEvent extends EventObject,
   TSnapshot = any,
-  TInternalState = any
+  TInternalState = any,
+  TPersisted = TInternalState
```
Andarist, line 1821: Do we currently have any need for `TPersisted` that is different from `TInternalState`?

packages/core/src/interpreter.ts
```diff
@@ -475,6 +470,14 @@ export class Interpreter<
     };
   }
 
+  public getPersistedState(): PersistedFrom<TBehavior> | undefined {
+    if (!this._state) {
+      return undefined;
+    }
+
+    return this.behavior.getPersistedState?.(this._state);
```
Andarist, line 478: It feels like this could be read like this:
```suggestion
    return this.behavior.getPersistedState?.(this._state || this._getInitialState());
```

Any reason why we don't grab the persisted state for not-started behaviors?

packages/core/src/interpreter.ts
```diff
@@ -296,12 +293,10 @@ export class Interpreter<
...
     this.status = ActorStatus.Running;
 
-    let resolvedState = initialState
-      ? this.behavior.restoreState?.(initialState, this._actorContext)
-      : this._getInitialState() ?? undefined;
+    let resolvedState = this._getInitialState();
 
     if (this.behavior.start) {
-      resolvedState = this.behavior.start(resolvedState, this._actorContext);
+      this.behavior.start(resolvedState, this._actorContext);
```
Andarist, line 299: is it a change in the contract between the `interpret` and the `behavior`? The `ActorBehavior` type still mentions `TInternalState` as the return type from `start`:
https://github.com/statelyai/xstate/blob/3769caea411bf7a4390ae34a4457c6bc7a464afe/packages/core/src/types.ts#L1837-L1840

packages/core/src/State.ts
davidkpiano: I think while I was working on this PR, it was hard to distinguish between a normal "state config" and a persisted "state config"; the latter would need to rehydrate the `children`

packages/core/src/types.ts
davidkpiano: Yes; in theory, the persisted state may e.g. omit data from the internal state (which you may not want to reveal), or add additional data that the actor behavior knows how to handle when restoring it, like... a checksum or version number or timestamp or something like that.

packages/core/src/interpreter.ts
davidkpiano: This should be changed; `.start(...)` is too late to provide any state.

packages/core/src/actors/reducer.ts
```diff
@@ -9,22 +14,33 @@ import { isSCXMLEvent } from '../utils';
...
+export function fromReducer<TState extends {}, TEvent extends EventObject>(
   transition: (
     state: TState,
     event: TEvent,
     actorContext: ActorContext<TEvent, TState>
   ) => TState,
   initialState: TState
-): ActorBehavior<TEvent, TState> {
-  return {
+): WithRequired<ActorBehavior<TEvent, TState, TState>, 'at'> {
```
Andarist, line 24: Do we need this `WithRequired`? It doesn't display that well in the IDe tooltips, cause I see this alias there and it's a lot of noise for the reader.

Fun fact, you can achieve this kind of `WithRequired` through intersections:
```ts
type ActorBehavior = {
  getInitialState: () => void,
  start?: () => void,
  at?: () => void
}

const behavior = {
  getInitialState() {},
} satisfies ActorBehavior & { at: unknown } // error, since `at` has not been provided
```

This doesn't really solve the readability concern. I have some ideas how to solve this, gonna play around with them and post later

davidkpiano: Yeah, my main goal is to make `at` required, but also, if we make it part of `interpret`, then we're going to remove `at` anyway.

Andarist: Re `WithRequired`:

to display the return type consistently we probably only have 2 options:
```ts
// duplicate
interface ReducerBehavior<TState, TEvent extends EventObject> extends ActorBehavior<TEvent, TState, TState> {
  at: (state: TInternalState) => this;
}

// type gymnastics
type _ReducerBehavior2<TState, TEvent extends EventObject> = ActorBehavior<TEvent, TState, TState> & { at: {} }
interface ReducerBehavior2<TState, TEvent extends EventObject> extends _ReducerBehavior2<TState, TEvent> {}
```

Both preserve JS docs from the original which is good.

> if we make it part of interpret, then we're going to remove at anyway.

Yeah, that's likely true

packages/core/src/interpreter.ts
Andarist: It could still return a new state - based on the provided one. But I agree, it's best if the whole initial state is known upfront. We don't want to create cascading updates on mount etc (even if that's only a UI concern). 

So ye - this signature should be probably changed

packages/core/src/types.ts
Andarist: The internal state is hidden from `getSnapshot` but it's likely always (almost?) important for the persisted state. If there is any internal state that can be lost when you persist+rehydrate then u won't get "the same" behavior back

packages/core/src/State.ts
Andarist: how they are different though? why we'd like to ignore the children when rehydrating from a "state config"?

packages/core/src/types.ts
Andarist: OTOH, while I think that we don't need this to "hide" things - we might want this to "serialize" state

packages/core/src/StateNode.ts
```diff
@@ -271,6 +271,7 @@ export class StateNode<
         ) {
           this.machine.options.actors = {
             ...this.machine.options.actors,
+            // TODO: this should accept `src` as-is
```
Andarist, line 274: It would make sense to remove this completely. We shouldn't have to put those on `this.machine.options.actors`. They were not provided there so why we'd keep them there?

This might impact the rehydration story for inline actors but to fix this we can use some extra information to understand how to look those up. We can either encode it in the auto-generated `id` somehow. Or we might have to use some extra `.type`-like thing somewhere.

I would also consider completely disallowing `getPersistedState` when the state has inline actors. How do we rehydrate this?
```ts
createMachine({
  entry: assign((_, __, { spawn }) => ({
    childRef: spawn(fromPromise(() => Promise.resolve(42)))
  }))
})
```

The problem here is that when serializing the persisted state we don't have any way to "point" to a stable behavior implementation of this actor. When rehydrating we won't have a clue how to access the original behavior of that child

packages/core/src/actors/callback.ts
```diff
@@ -12,7 +12,7 @@ import {
   isFunction
 } from '../utils';
 import { doneInvoke, error } from '../actions';
-import { startSignalType, stopSignalType, isSignal } from '../actors';
+import { startSignalType, stopSignalType, isSignal } from '../actors/index';
```
Andarist, line 15: Those should use `.js` at the end

packages/core/src/actors/callback.ts
```diff
@@ -24,10 +24,10 @@ export function fromCallback<TEvent extends EventObject>(
   invokeCallback: InvokeCallback
 ): ActorBehavior<TEvent, undefined> {
   const behavior: ActorBehavior<TEvent, undefined, CallbackInternalState> = {
-    start: (state, { self }) => {
-      self.send({ type: startSignalType } as TEvent);
+    start: (_state, actorCtx) => {
+      actorCtx.self.send({ type: startSignalType } as TEvent);
 
-      return state;
+      return behavior.getInitialState(actorCtx);
```
Andarist, line 30: as per the discussion in other comments - this return value should likely be removed

packages/core/src/types.ts
```diff
@@ -784,6 +784,7 @@ export interface MachineImplementationsSimplified<
   >;
   delays: DelayFunctionMap<TContext, TEvent>;
   context: Partial<TContext> | ContextFactory<Partial<TContext>>;
+  state: State<TContext, TEvent, any> | undefined;
```
Andarist, line 787: This doesn't belong to a machine - state is the interpreter's concern
```suggestion
```

packages/core/src/types.ts
```diff
@@ -1717,6 +1718,7 @@ export interface ActorRef<TEvent extends EventObject, TSnapshot = any>
   // TODO: should this be optional?
   start?: () => void;
   getSnapshot: () => TSnapshot | undefined;
+  getPersistedState?: () => any;
```
Andarist, line 1721: `any` definitely shouldn't be used here, I'm not yet sure though at which level we should care about this. Perhaps only the root `actorRef` could expose this? But then... Its own `TPersistedState` should probably encode the persisted state types of all of their children (recursively). 

But maybe this is not a concern for now. We can reexamine this when revamping the types as a whole

packages/core/src/types.ts
```diff
@@ -1834,9 +1838,17 @@ export interface ActorBehavior<
     state: TInternalState,
     actorCtx: ActorContext<TEvent, TSnapshot>
   ) => TInternalState;
+  /**
+   * Clones behavior that starts at specified state
+   */
+  at?: (state: TInternalState) => this;
+  /**
+   * @returns Persisted state as JSON
```
Andarist, line 1846: We don't completely mandate the nature of the persisted state. We could mention this as a recommendation here but we shouldn't imply here what doesn't have to be always true.
```suggestion
   * @returns Persisted state
```

packages/core/src/StateNode.ts
Andarist: We also have the same problem with invoke, despite having a "stable" reference to it (object path). And more than that - even referenced implementations have this problem when they are declared using factories.

```ts
createMachine({
  invoke: {
    src: 'dynamic'
  }
}, {
  actors: {
    dynamic: (ctx) => fromPromise(() => Promise.resolve(42))
  }
)
```

How do we rehydrate this? It looks a little bit like maybe `src` should just **never** be a factory. We need to have the ability to use the same actor behavior with different arguments and stuff... but that seems to be solved by `input`! The only thing is that the JS API might look a little bit quirky for this, because we'd have to accept two things for any given actor: 1. a static `src` 2. an input resolver~ (with the "standard" `(ctx, ev) => input` signature)

packages/core/src/StateMachine.ts
```diff
@@ -428,6 +448,74 @@ export class StateMachine<
...
+
+        if (!impl) return;
+
+        if (typeof impl === 'function') {
+          const behavior = impl(state.context, state.event, {
+            id: key,
+            src: {} as any,
+            _event: state._event,
+            meta: undefined
+          });
```
Andarist, line 483: This is a no-go - those arguments here are completely different from what usually can be used to create actors. This effectively makes typegen useless for actors because we'd have to annotate each actor factory with `TEvent` and not `TNarrowedEvent`. The context value is also way different here - it's used from a completely different moment in time. 

We try to re-instantiate "old" actors but we are using "new" values here. This is IMHO a conceptual clash

packages/core/src/StateMachine.ts
```diff
@@ -428,6 +448,74 @@ export class StateMachine<
...
+
+          state.actions.unshift(
+            invoke({
+              id: key,
+              // @ts-ignore TODO: fix types
+              src: actorRef, // TODO
+              ref: actorRef,
+              meta: undefined
+            })
+          );
```
Andarist, line 498: is this needed here for anything?

Andarist (commented):
After thinking about it more - I'm only more certain that the implementation of this should be closer to the interpreter. A machine that starts in a rehydrated state is still the same machine, not a fork of the original.

packages/core/src/State.ts
davidkpiano: The `children` here is a key-value pairing where the value is the persisted state. In a live state config, the `children` values would be the actual actorrefs

packages/core/src/StateMachine.ts
davidkpiano: I don't remember but tests fail if I remove this part

packages/core/src/StateMachine.ts
davidkpiano: What do you suggest we do instead? The actor implementations still expect those arguments, even if they're not used when restoring state.

packages/core/src/actors/reducer.ts
davidkpiano: `WithRequired` removed (as well as `at`)

packages/core/src/StateMachine.ts
davidkpiano: Removed

packages/core/src/types.ts
davidkpiano: Made a clarifying comment

packages/core/src/types.ts
davidkpiano: Added a todo for this

jlarmstrongiv:
Will recursive persistence work with the [receptionist](https://github.com/statelyai/xstate/pull/3837) for a fully serializable and persistable system?  That would open a lot of possibilities.

packages/core/src/StateMachine.ts
davidkpiano: Ping @Andarist

Andarist:
> Will recursive persistence work with the https://github.com/statelyai/xstate/pull/3837 for a fully serializable and persistable system? That would open a lot of possibilities.

The goal is definitely to create a cohesive architecture with which different pieces work together seamlessly. It's hard to answer this with more detail without knowing more context behind your question though.

packages/core/src/StateMachine.ts
```diff
@@ -374,16 +379,15 @@ export class StateMachine<
 
     return macroState;
   }
+
   public start(
     state: State<TContext, TEvent, TResolvedTypesMeta>,
     actorCtx: ActorContext<TEvent, State<TContext, TEvent, TResolvedTypesMeta>>
-  ): State<TContext, TEvent, TResolvedTypesMeta> {
+  ): void {
```
Andarist, line 386: This change should be documented with a changeset

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -160,33 +162,34 @@ describeEachReactMode('useMachine (%s)', ({ suiteKey, render }) => {
...
+  //     expect(state.context).toEqual({
+  //       foo: 'bar',
+  //       test: true
+  //     });
+
+  //     return null;
+  //   };
+
+  //   render(<Test />);
+  // });
```
Andarist, line 192: That feature will just have a completely new shape - so this particular test isn't that applicable to the post-input world, right? Perhaps this can just be removed from here already in this PR.

packages/core/test/rehydration.test.ts
```diff
@@ -73,9 +73,11 @@ describe('rehydration', () => {
         }
       });
 
-      const service = interpret(machine);
+      const activeState = machine.transition(undefined, { type: 'NEXT' });
```
Andarist, line 76: Shouldn't this be rewritten like this?
```suggestion
      const activeState = machine.resolveStateValue('active');
```

packages/core/src/types.ts
```diff
@@ -1960,3 +1989,13 @@ export type StateValueFrom<TMachine extends AnyStateMachine> = Parameters<
...
+
+export interface PersistedMachineState<TState extends AnyState> {
+  [key: string]: any;
+  children: {
+    [key in keyof TState['children']]: any;
+  };
+  persisted: true;
+}
+
+export type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
```
Andarist, line 2001: Do we use it anywhere?

davidkpiano: Not anymore, will remove

packages/core/src/StateMachine.ts
davidkpiano: Don't know if it makes much sense (this isn't a change from v4 since it's a new method) but I'll add one.

.changeset/shy-cobras-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': major
+---
+
+The return type of the `.start(...)` method of actor behaviors now returns `void`.
```
Andarist, line 5: I reported the need for this changeset [here](https://github.com/statelyai/xstate/pull/3743#discussion_r1121893328) but I thought that I was looking at `Interpreter` while this was just in `StateMachine`. So I actually think that it's OK to skip it. It's also OK to keep it but we will have to remove/clean it up after releasing the new alpha release - since this one won't be relevant for the final v5 release.

```suggestion
---
---
```

packages/core/test/interpreter.test.ts
```diff
@@ -184,14 +184,14 @@ describe('interpreter', () => {
...
 
-      const s = interpret(machine);
+      const bState = machine.initialState;
+
+      expect(bState.value).toEqual('b');
 
-      // would have deferred the entry action
-      s.getSnapshot();
+      // Ensure that there are no actions attached to the state
+      bState.actions.length = 0;
```
Andarist, line 192: Couldn't this be just replaced with `machine.resolveStateValue('b')`? It seems more straightforward than resetting actions on an object that should be treated as immutable

packages/core/src/actors/reducer.ts
```diff
@@ -9,22 +9,26 @@ import { isSCXMLEvent } from '../utils';
  * @returns An actor behavior
  */
 
-export function fromReducer<TState, TEvent extends EventObject>(
+export function fromReducer<TState extends {}, TEvent extends EventObject>(
```
Andarist, line 12: I don't think that we have to prescribe the constraint of `TState` here. The user should have complete freedom here
```suggestion
export function fromReducer<TState, TEvent extends EventObject>(
```

packages/core/src/State.ts
```diff
@@ -287,3 +288,22 @@ export function cloneState<TState extends AnyState>(
...
+  state: TState
+): PersistedMachineState<TState> {
+  const { configuration, transitions, tags, machine, children, ...jsonValues } =
+    state;
+
+  const childrenJson: any = {};
+
+  for (const key in children) {
+    childrenJson[key] = children[key].getPersistedState?.();
+  }
```
Andarist, line 302:
```suggestion
  for (const id in children) {
    childrenJson[id] = children[id].getPersistedState?.();
  }
```

packages/core/src/State.ts
Andarist: > In a live state config, the children values would be the actual actorrefs

That is the distinction then - maybe that would allow us to get rid of `persisted: true`.

packages/core/src/types.ts
```diff
@@ -1960,3 +1989,11 @@ export type StateValueFrom<TMachine extends AnyStateMachine> = Parameters<
 
 export type StateFromMachine<TMachine extends AnyStateMachine> =
   TMachine['initialState'];
+
+export interface PersistedMachineState<TState extends AnyState> {
+  [key: string]: any;
```
Andarist, line 1994: can we do better here? :p

packages/core/src/types.ts
```diff
@@ -1635,6 +1641,8 @@ export interface InterpreterOptions {
   autoForward?: boolean;
 
   sync?: boolean;
+
+  state?: any; // TODO: type this
```
Andarist, line 1645: this should be `TPersisted`, right? 🤔

packages/core/test/behaviors.test.ts
```diff
@@ -82,6 +85,34 @@ describe('promise behavior (fromPromise)', () => {
...
+        })
+      );
+
+      const restoredActor = interpret(promiseBehavior, {
+        state: resolvedPersistedState
+      }).start();
+      expect(restoredActor.getSnapshot()).toBe(42);
+      done();
+    }, 5);
+  });
```
Andarist, line 115: I think that this is an improved version of this test and it doesn't pass:
```suggestion
  it('should persist a promise', (done) => {
    let createdPromises = 0;
    const promiseBehavior = fromPromise(() => {
      createdPromises++;
      return Promise.resolve(createdPromises);
    });

    const actor = interpret(promiseBehavior);
    actor.start();

    setTimeout(() => {
      const resolvedPersistedState = actor.getPersistedState();

      expect(resolvedPersistedState).toEqual(
        expect.objectContaining({
          data: 1
        })
      );

      const restoredActor = interpret(promiseBehavior, {
        state: resolvedPersistedState
      }).start();
      expect(restoredActor.getSnapshot()).toBe(1);
      expect(createdPromises).toBe(1);
      done();
    }, 5);
  });
```

Andarist: Additionally - this test prints this in the console now:
```
    Warning: Event "$$xstate.resolve" was sent to stopped actor "x:1 (x:1)". This actor has already reached its final state, and will not transition.
    Event: {"type":"$$xstate.resolve","data":42}
```

packages/core/src/interpreter.ts
```diff
@@ -484,6 +490,12 @@ export class Interpreter<
     };
   }
 
+  public getPersistedState(): PersistedFrom<TBehavior> | undefined {
```
Andarist, line 493: When this can return `undefined` if the snapshot is always available synchronously right after creation?

packages/core/src/interpreter.ts
```diff
@@ -484,6 +490,12 @@ export class Interpreter<
...
+    return this.behavior.getPersistedState?.(
+      this._state ?? this._getInitialState()
```
Andarist, line 495: Couldn't this be simplified to this?
```suggestion
      this.getSnapshot()
```

packages/core/src/interpreter.ts
```diff
@@ -303,12 +311,10 @@ export class Interpreter<
...
     this.status = ActorStatus.Running;
 
-    let resolvedState = initialState
-      ? this.behavior.restoreState?.(initialState, this._actorContext)
-      : this._getInitialState() ?? undefined;
+    let resolvedState = this._getInitialState();
 
     if (this.behavior.start) {
-      resolvedState = this.behavior.start(resolvedState, this._actorContext);
+      this.behavior.start(resolvedState, this._actorContext);
```
Andarist, line 317: I wonder if there are any unwanted things that might happen if we call `.start()` with a rehydrated state. How do we describe the exact role of `.start()`?

packages/core/src/interpreter.ts
```diff
@@ -132,6 +134,10 @@ export class Interpreter<
       ...options
     };
 
+    if (resolvedOptions.state) {
+      this._initialState = this._state = resolvedOptions.state;
```
Andarist, line 138: q: why do we need both? why can't we just use `this._state`?

packages/core/src/StateMachine.ts
```diff
@@ -442,9 +452,53 @@ export class StateMachine<
...
   ): State<TContext, TEvent, TResolvedTypesMeta> {
-    const restoredState = isStateConfig(state)
-      ? this.resolveState(state as any)
-      : this.resolveState(State.from(state as any, this.context, this));
+    let restoredState: State<TContext, TEvent, TResolvedTypesMeta>;
+
+    if ('persisted' in state) {
+      const restoredChildren: Record<string, AnyActorRef> = {};
+
+      Object.keys(state.children).forEach((key) => {
```
Andarist, line 460: let's rename `key` here to `id` since that's how it's used later on - at least in some places

packages/core/src/StateMachine.ts
```diff
@@ -442,9 +452,53 @@ export class StateMachine<
...
+        const persistedState = state.children[key];
+        const impl = this.options.actors[key];
```
Andarist, line 462: this seems to be incorrect - `this.options.actors` is keyed using behavior keys~ but you are using an actor's id to index it here

packages/core/src/StateMachine.ts
```diff
@@ -442,9 +452,53 @@ export class StateMachine<
       State<TContext, TEvent, TResolvedTypesMeta>
     >
   ): State<TContext, TEvent, TResolvedTypesMeta> {
-    const restoredState = isStateConfig(state)
-      ? this.resolveState(state as any)
-      : this.resolveState(State.from(state as any, this.context, this));
+    let restoredState: State<TContext, TEvent, TResolvedTypesMeta>;
```
Andarist, line 455: q: what should happen if we rehydrate a machine with a state that is missing some child while the machine configuration expects that child to exist (like if we'd rehydrate in `b` that has `invoke: 'fooActor'` without a persisted state for that `fooActor`)?

davidkpiano: I think by default, we would just start (invoke) the `'fooActor'`.

packages/core/src/StateMachine.ts
davidkpiano: I'm still not sure how this can be fixed in this PR, since the solution depends on `input` which is in the linked PR. Open to any suggestions.

Also, I am open to disallowing `() => fromPromise()` completely 👍

packages/core/src/interpreter.ts
davidkpiano: Tests fail (try changing `this._initialState` in `_getInitialState()`) for `@xstate/react`. My guess is something relies on the `_initialState` (unchanged) being different from the `_state` (can change).

packages/core/src/StateMachine.ts
davidkpiano: Yes, this ensures that the `actorRef` actually gets invoked. Test `packages/core/test/behaviors.test.ts:386:71` fails without this.

packages/core/src/interpreter.ts
davidkpiano: The role of `.start(state)` is to _attempt_ to put an actor at the specified `state`.

packages/core/src/interpreter.ts
davidkpiano: It cannot because the snapshot is not always equal to the internal state (this is only true for state machines and reducers)

packages/core/src/interpreter.ts
davidkpiano: When `.getPersistedState` is not defined

jlarmstrongiv:
> The goal is definitely to create a cohesive architecture with which different pieces work together seamlessly. It's hard to answer this with more detail without knowing more context behind your question though.

Mainly for server-side and multiplayer use-cases.  

For example, this server-side demo:
- https://www.youtube.com/watch?v=r4xI6LPf6iQ
- https://github.com/erikras/remix-conf-2022
Serializes a state machine in a cookie and sends events for the server to load and calculate.

For multi-user/multiplayer, the idea is similar but with a database instead of a cookie in this demo:
- https://www.youtube.com/watch?v=sAtonzSZ8HM
- https://github.com/learnwithjason/state-machine-edge-function-workflow

Having a fully serializable system will make these patterns very powerful.

packages/core/src/StateMachine.ts
Andarist: we need to add a test case for this

packages/core/src/interpreter.ts
Andarist: So perhaps let's just initialize `this._state` in the `constructor` and use `this._state` here. Without the fallback to the `this._getInitialState()`. Do we get anything from this information being computed lazily?

packages/core/src/interpreter.ts
Andarist: It would be great to have a test case for this. I wonder though... we have quite a bit of different concepts here:
- snapshots -> public~ state
- internal state
- persisted state (a serializable internal state)

It's quite a lot, I wonder if we could simplify this somehow. Perhaps serializing could be an enhancement and if there is no `getPersistedState` then maybe we should just return the internal state

packages/core/src/interpreter.ts
Andarist: I strongly suspect that those are just symptoms of something else. I prepare a PR [here](https://github.com/statelyai/xstate/pull/3874) with some changes that I think are correct. Tests fail there - but that's just because the implementation is yet incomplete. 

One of the things that React's test suite catches here:
```ts
const actor = interpreter(machine).start()
actor.stop()
actor.start()
```

If this actor has any children then they should be rehydrated when we `start` it again.

packages/core/src/interpreter.ts
Andarist: In other words - it's used both for starting a fresh actor and for starting a rehydrated one. I think that overall this is OK - but the implementers need to be aware of this.

packages/core/src/interpreter.ts
davidkpiano: Tried this quickly; not easy to do this (tests fail). Help wanted

Andarist: It's likely "just" a part of [this](https://github.com/statelyai/xstate/pull/3743#discussion_r1122973161). I can play more with this PR here and that proposed patch to understand better how the final implementation could look like

packages/core/src/types.ts
davidkpiano: 5dc9f4668c6c99085d19f7e578f6480b460dd5ed

packages/core/src/types.ts
davidkpiano: 739d8cbc8920ef5f2a76b16316a82b059a53c1da

packages/core/test/behaviors.test.ts
davidkpiano: 8c9ae224dffa7a14c5f1e74c1a7ac9bb873fd504

packages/core/src/StateMachine.ts
davidkpiano: Added a skipped test case; may be best handled after this PR

davidkpiano:
Closes https://linear.app/statelyai/issue/STA-3444

packages/core/src/StateMachine.ts
```diff
@@ -436,15 +453,54 @@ export class StateMachine<
   }
 
   public restoreState(
-    state: State<TContext, TEvent, TResolvedTypesMeta>,
+    state: PersistedMachineState<any>,
```
Andarist, line 456:
```suggestion
    state: PersistedMachineState<State<TContext, TEvent, TResolvedTypesMeta>>,
```

packages/core/src/StateMachine.ts
```diff
@@ -446,58 +446,64 @@ export class StateMachine<
...
     _actorCtx?: ActorContext<
       TEvent,
       State<TContext, TEvent, TResolvedTypesMeta>
     >
   ): State<TContext, TEvent, TResolvedTypesMeta> {
-    let restoredState: State<TContext, TEvent, TResolvedTypesMeta>;
+    const self = this;
```
Andarist, line 462: oh wow, this is a real blast from the past 😅 you are already using an arrow function below - you don't need this old school trick, just use `this`

packages/core/src/types.ts
```diff
@@ -1904,6 +1932,26 @@ export type SnapshotFrom<T> = ReturnTypeOrValue<T> extends infer R
...
 
+export type PersistedStateFrom<TBehavior extends ActorBehavior<any, any>> =
+  TBehavior extends ActorBehavior<
+    infer _TEvent,
+    infer _TSnapshot,
+    infer _TInternalState,
+    infer TPersisted
+  >
+    ? TPersisted
+    : 'nope';
```
Andarist, line 1943: nope, this won't pass the review, nope

girl shaking her head and saying 'nope'

packages/core/src/types.ts
```diff
@@ -1960,3 +2008,13 @@ export type StateValueFrom<TMachine extends AnyStateMachine> = Parameters<
 
 export type StateFromMachine<TMachine extends AnyStateMachine> =
   TMachine['initialState'];
+
+export type PersistedMachineState<TState extends AnyState> = TState & {
```
Andarist, line 2012: This intersection is definitely lying to us :P it should look closer to this one:
```suggestion
export type PersistedMachineState<TState extends AnyState> = Pick<
  TState,
  'value' | 'output' | 'context' | '_event' | 'done' | 'historyValue'
> & {
```

packages/core/src/StateMachine.ts
```diff
@@ -446,58 +446,64 @@ export class StateMachine<
...
+    let restoredState;
 
-    if ('persisted' in state) {
-      const restoredChildren: Record<string, AnyActorRef> = {};
+    const children = {};
 
-      Object.keys(state.children).forEach((id) => {
-        const persistedState = state.children[id];
-        const impl = this.options.actors[id];
+    if (_actorCtx && state.persisted) {
```
Andarist, line 467: this `state.persisted` check is redundant now since you now require `PersistedMachineState` as the argument

packages/core/src/types.ts
```diff
@@ -1960,3 +2008,13 @@ export type StateValueFrom<TMachine extends AnyStateMachine> = Parameters<
...
+  children: {
+    [K in keyof TState['children']]: {
+      state: any; // TODO: fix (should be state from actorref)
+      src?: InvokeSourceDefinition;
```
Andarist, line 2016: what do we get from using `InvokeSourceDefinition` here instead of a simple `string`?

packages/core/src/State.ts
Andarist: Ye, I think this should be enough to test those things out. `persisted: true` is a little bit redundant + it would be a little bit silly if we'd fail to rehydrate a persisted state that doesn't have this property (when the rest of it is valid, ofc). We already have a lot of validation here at the type-level since `PersistedState` isn't compatible with `State` (and vice-versa)

packages/core/test/interpreter.test.ts
```diff
@@ -639,10 +636,10 @@ describe('interpreter', () => {
       });
       const bState = toggleMachine.transition(activeState, { type: 'SWITCH' });
 
-      interpret(toggleMachine).start(bState);
+      interpret(toggleMachine, { state: bState }).start();
 
       setTimeout(() => {
-        expect(activityActive).toBeFalsy();
+        expect(activityActive).toBeTruthy();
```
Andarist, line 642: this change is suspicious - what's the reason behind it? 🤔

packages/core/test/behaviors.test.ts
```diff
@@ -186,3 +298,189 @@ describe('observable behavior (fromObservable)', () => {
...
+
+    // Ensure that the child actor is started
+    // LAST is sent to parent which sends LAST to child
+    newActor.send({ type: 'LAST' });
+    // child is at 'c'
+
+    expect(newActor.getSnapshot().children.child.getSnapshot().value).toBe('c');
+  });
+
+  it('should return the initial persisted state of a non-started actor', () => {
```
Andarist, line 445: could we also add a test case that would check that the initial state of a child is available in the persisted state even before we start its parent?

packages/core/src/types.ts
```diff
@@ -1876,18 +1900,22 @@ export interface ActorBehavior<
     actorCtx: ActorContext<TEvent, TSnapshot>
   ) => TInternalState;
   restoreState?: (
-    restoredState: any,
+    restoredState: TPersisted,
```
Andarist, line 1903:
```suggestion
    persistedState: TPersisted,
```

packages/core/src/StateMachine.ts
davidkpiano: Yep, there was a point during development where I need to do this, accidentally left it in hah

packages/core/src/types.ts
davidkpiano: I need help with this 🥺

packages/core/src/types.ts
davidkpiano: I'm actually going to move this to a simple `string` soon

packages/core/test/interpreter.test.ts
davidkpiano: As discussed, actions should not restart by default (still need to do that) but state invocations should be restarted.

packages/core/src/StateMachine.ts
Andarist: this also means that the `else` branch could be removed from here

packages/core/src/StateNode.ts
```diff
@@ -254,6 +254,35 @@ export class StateNode<
...
+    // //   !this.machine.options.actors[resolvedId] &&
+    // //   typeof src !== 'string' &&
+    // //   !('type' in src)
+    // // ) {
+    // //   this.machine.options.actors = {
+    // //     ...this.machine.options.actors,
+    // //     // TODO: this should accept `src` as-is
+    // //     [resolvedId]: typeof src === 'function' ? src : () => src
+    // //   };
+    // // }
```
Andarist, line 285:
```suggestion
```

packages/core/test/behaviors.test.ts
davidkpiano: b75d480508b38f5b29b4749c9deec3478f8c1508

packages/core/src/actors/promise.ts
```diff
@@ -42,6 +54,12 @@ export function fromPromise<T>(
       }
     },
     start: (state, { self }) => {
+      // TODO: determine how to allow customizing this so that promises
+      // can be restarted if necessary
+      if (state.status !== 'active') {
+        return;
+      }
```
Andarist, line 61: In a way, that's achievable today with:
```ts
const fromRestartablePromise: typeof fromPromise = (...args) => {
  const behavior = fromPromise(...args)
  behavior.getPersistedState = behavior.getInitialState
  return behavior
}
```

I agree that we might want to evaluate making this configurable but that can be done post-release.

packages/core/src/actors/promise.ts
```diff
@@ -42,6 +54,12 @@ export function fromPromise<T>(
       }
     },
     start: (state, { self }) => {
+      // TODO: determine how to allow customizing this so that promises
+      // can be restarted if necessary
+      if (state.status !== 'active') {
```
Andarist, line 59: the same should be added to both observable types (+tests)

packages/core/src/types.ts
Andarist: all of the other types like this in our repo just use `never` here
```suggestion
    : never;
```

packages/core/test/interpreter.test.ts
Andarist: hm, quite interesting that it didn't work like that before - when it comes to invocations

packages/core/src/actors/promise.ts
davidkpiano: https://github.com/statelyai/xstate/pull/3743/commits/a08743226d1c009ca922dc684fb44b44d6af87dc

packages/core/src/actors/observable.ts
```diff
@@ -61,7 +72,7 @@ export function fromObservable<T, TEvent extends EventObject>(
           return state;
         case stopSignalType:
           state.canceled = true;
-          state.subscription!.unsubscribe();
+          state.subscription?.unsubscribe();
```
Andarist, line 75: when the subscription can be missing when receiving `stopSignalType`? This sounds like an impossible state

Andarist: Ah, `stopSignalType` is coming from an external `.stop()`, hmmm. It might be needed to have some guard in place here for this case - but I wonder about the exact scenario related to this because it seems that we don't have a test for it.

packages/core/src/actors/observable.ts
```diff
@@ -76,6 +87,11 @@ export function fromObservable<T, TEvent extends EventObject>(
       };
     },
     start: (state, { self }) => {
+      if (state.status === 'done') {
+        // Do not restart a completed observable
```
Andarist, line 91: what about a canceled actor? a canceled actor is pretty much "completed" (even if not successfully). OTOH, a canceled actor might not get persisted at all, right? Since it should be cleaned up from `state.children`.

This also brings an important issue here - spawned actors (including stopped ones) might be part of the `context`. We are not doing anything about this and likely we should. They should be rehydrated in the same way as the invoked ones - the fact that they live in `context` complicates the matter though :s

packages/core/test/actor.test.ts
```diff
@@ -1177,4 +1177,49 @@ describe('actors', () => {
...
+        src: fromObservable(() => {
+          const obs = of(42);
+          const sub = obs.subscribe.bind(obs);
+
+          obs.subscribe = (...args) => {
+            subscriptionCount++;
+            return sub(...(args as any));
+          };
+
+          return obs;
```
Andarist, line 1195:
```suggestion
          subscriptionCount++;
          return of(42);
```

Andarist: it's pretty much the same thing but way more simplified

packages/core/test/actor.test.ts
```diff
@@ -1177,4 +1177,49 @@ describe('actors', () => {
...
+    const obs = actor
+      .getSnapshot()
+      .children['observable'].getPersistedState?.();
+
+    expect(obs).toEqual(
+      expect.objectContaining({
+        data: 42,
+        status: 'done'
+      })
+    );
```
Andarist, line 1211: This kinda ain't part of this test, I don't mind it that much but I find shorter tests to be easier to read
```suggestion
```

packages/core/test/actor.test.ts
```diff
@@ -1177,4 +1177,49 @@ describe('actors', () => {
...
+
+    interpret(machine, {
+      state: persistedState
+    }).start();
+
+    setTimeout(() => {
+      // Will be 2 if the observable is resubscribed
+      expect(subscriptionCount).toBe(1);
+      done();
+    });
```
Andarist, line 1223: everything here is sync, there is no need to complicate this with async jobs:
```suggestion
    // Will be 2 if the observable is resubscribed
    expect(subscriptionCount).toBe(1);
```

packages/core/src/actors/observable.ts
```diff
@@ -76,6 +87,11 @@ export function fromObservable<T, TEvent extends EventObject>(
       };
     },
     start: (state, { self }) => {
+      if (state.status === 'done') {
```
Andarist, line 90: the same fix has to be applied to `fromEventObservable`

packages/core/src/actors/observable.ts
davidkpiano: If you restore the observable actor and it's already completed, `subscription` will be undefined.

packages/core/src/actors/observable.ts
davidkpiano: For spawned actors, follow-up PR?

Andarist: 👍

packages/core/src/actors/observable.ts
Andarist: Yes, but this line of code is specifically called at some point - and we should have a test case that would hit this line of the code when the subscription is absent (and we dont have such rn)

davidkpiano: I can but isn't `?.` safer? I'm not sure what should be tested here since the code won't fail regardless of if it's defined or not. We would be testing the change from `!` to `?` which is sorta meaningless.

Do we need to test all the places where we have `?.`?

Andarist: > I can but isn't ?. safer? 

It is. But it's also potentially a little bit of redundant code, runtime overhead etc. I find defensive programming often obscures the meaning of the code and covers up potential bugs.

For example, from my PoV the important question here is - can the `stopSignalType` even be received when `state.subscription === undefined`? If the answer is "no" but it happens anyway then this `?.` will "swallow" the logical/integration bug and we won't even learn about it. If the answer is "yes" then we can write a test case that would exercise this scenario and we can verify there that nothing throws (and then we will be sure that `?.` is actually needed!)

packages/core/src/StateNode.ts
davidkpiano: This is removed in `v5/input-3` #3795

packages/core/src/actors/observable.ts
davidkpiano: I reverted this

packages/core/src/types.ts
davidkpiano: This is done in #3815

Issue #3744, opened 2023-01-04
Bug: Invoke configuration does not support string type when given array


viglucci:
### Description

There are type errors when attempting to use a string as the name of a state in the `onDone` or `onError` portion for an invoked service, even though this should be considered a perfectly valid configuration, and without typescript functions just fine.

### Expected result

The following simple machine should not cause type errors:

```ts
import { createMachine } from "xstate";

createMachine({
    invoke: {
        src: () => () => { },
        onDone: [{ target: "a" }, "b"],
    },
    states: {
        a: {},
        b: {},
    },
});
```

### Actual result

Typescript is unhappy:

```
Type '{ src: () => () => void; onDone: (string | { target: string; })[]; }' is not assignable to type 'SingleOrArray<AnyStateMachine | InvokeConfig<unknown, AnyEventObject>>'.
  Types of property 'onDone' are incompatible.
    Type '(string | { target: string; })[]' is not assignable to type 'string | SingleOrArray<TransitionConfig<unknown, DoneInvokeEvent<any>>>'.
      Type '(string | { target: string; })[]' is not assignable to type 'TransitionConfig<unknown, DoneInvokeEvent<any>>[]'.
        Type 'string | { target: string; }' is not assignable to type 'TransitionConfig<unknown, DoneInvokeEvent<any>>'.
          Type 'string' has no properties in common with type 'TransitionConfig<unknown, DoneInvokeEvent<any>>'.ts(2322)
```

### Reproduction

https://codesandbox.io/s/currying-shadow-8tnshi?file=/src/index.ts

### Additional context

Originally reported here: https://github.com/statelyai/xstate-tools/issues/275

Andarist:
This was fixed by https://github.com/statelyai/xstate/pull/3745

PR #3745, opened 2023-01-04
Fix: update types for invoked service target to allow string

viglucci:
Addresses #3744 https://github.com/statelyai/xstate-tools/issues/275

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -514,13 +514,13 @@ export interface InvokeConfig<TContext, TEvent extends EventObject> {
    */
   onDone?:
     | string
-    | SingleOrArray<string | TransitionConfig<TContext, DoneInvokeEvent<any>>>;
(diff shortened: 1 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -514,13 +514,13 @@ export interface InvokeConfig<TContext, TEvent extends EventObject> {
    */
   onDone?:
     | string
-    | SingleOrArray<TransitionConfig<TContext, DoneInvokeEvent<any>>>;
+    | SingleOrArray<string | TransitionConfig<TContext, DoneInvokeEvent<any>>>;
```
Andarist, line 517: Other places like this use `TransitionConfigOrTarget`. Could you use this to keep things consistent?

viglucci: Sure thing! I've made this change.

davidkpiano (approved):

PR #3752, opened 2023-01-08
Fixed how the initial state of a service is read by the `useActor` hook

davidkpiano:
This PR adds a failing test and proposes a "fix" for an issue [reported in Discord](https://discord.com/channels/795785288994652170/1061308540490367006/1061308540490367006):

> I've been putting off updating the xstate package in my company's codebase for some time now as some of our machines break when upgrading from 4.32.1 -> 4.33.0. 
>
> I finally got some time to dedicate to it and have replicated the issue in a simplified code sandbox.  The sandbox has a very basic react app, which uses a parentMachine, which in turn spawns a childMachine. There is a button that sends an event to the parentMachine. The parentMachine should then send an event to the childMachine. The UI should then update to say that the child received the event. Simple.
>
> The issue seems to be that the parentMachine does not send the event to the childMachine after the upgrade. The childMachine also does not appear in the inspector following the upgrade, possibly another symptom of whatever the underlying issue is.
> 
> Refer to the two code sandboxes below, the only difference between them is the xstate version:
> 4.32.1 https://codesandbox.io/s/xstate-issue-4-32-1-91uthm?file=/src/App.tsx
> 4.33.0 https://codesandbox.io/s/xstate-issue-4-33-0-w1c3zu?file=/src/App.tsx
> 
> Any help on this would be greatly appreciated, have been stuck on it for awhile!

### DIFF
--- a/.changeset/quiet-swans-divide.md
+++ b/.changeset/quiet-swans-divide.md
@@ -2,4 +2,4 @@
 '@xstate/react': patch
 ---
 
-Computing the initial state is now consistent with `useMachine` and `useInterpret`, avoiding stale initial state problems with nested machines
+Computing the initial state is now consistent with `useMachine` and `useActor`, avoiding stale initial state problems with nested machines
--- a/.changeset/quiet-swans-divide.md
+++ b/.changeset/quiet-swans-divide.md
@@ -1,0 +1,5 @@
+---
+'@xstate/react': patch
+---
+
+Computing the initial state is now consistent with `useMachine` and `useInterpret`, avoiding stale initial state problems with nested machines
--- a/packages/xstate-react/src/useInterpret.ts
+++ b/packages/xstate-react/src/useInterpret.ts
@@ -136,10 +136,6 @@ export function useInterpret<TMachine extends AnyStateMachine>(
 
   useEffect(() => {
     const rehydratedState = options.state;
-    // @ts-ignore
-    // This ensures that the interpreter recomputes the initial state
-    // TODO: figure out a better way to do this
-    delete service._initialState;
     service.start(
       rehydratedState ? (State.create(rehydratedState) as any) : undefined
     );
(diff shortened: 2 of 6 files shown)


packages/xstate-react/src/useInterpret.ts
```diff
@@ -136,6 +136,10 @@ export function useInterpret<TMachine extends AnyStateMachine>(
 
   useEffect(() => {
     const rehydratedState = options.state;
+    // @ts-ignore
+    // This ensures that the interpreter recomputes the initial state
+    // TODO: figure out a better way to do this
+    delete service._initialState;
```
davidkpiano, line 142: @Andarist There's probably a better way to handle this for v4, wdyt?

Andarist: Actually, I'd say that this is an acceptable solution in v4. However, I wonder why the same thing here isn't enough.
https://github.com/statelyai/xstate/blob/40514d22ccaf7bd99318de1bdd9c931231fc8019/packages/core/src/interpreter.ts#L598

After all, we already call `.stop()` in the effect's cleanup.

davidkpiano: Because of this: https://github.com/statelyai/xstate/blob/40514d22ccaf7bd99318de1bdd9c931231fc8019/packages/core/src/interpreter.ts#L591-L594

That line never gets called if the interpreter was never even started, which is the case here. 🕵️

.changeset/quiet-swans-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/react': patch
+---
+
+Computing the initial state is now consistent with `useMachine` and `useInterpret`, avoiding stale initial state problems with nested machines
```
Andarist, line 5:
```suggestion
Computing the initial state is now consistent with `useMachine` and `useActor`, avoiding stale initial state problems with nested machines
```

PR #3756, opened 2023-01-09
Make all transitions internal by default

Andarist:
This is just a WIP to open/continue the discussion about this change. I expect that many tests might fail since we definitely rely on "inferred" transition types in many of them - it will be easy to fix those tests up, I can do it after we agree on the change itself.

### DIFF
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -363,17 +363,7 @@ export function formatTransition<
   }
 ): AnyTransitionDefinition {
   const normalizedTarget = normalizeTarget(transitionConfig.target);
-  const internal =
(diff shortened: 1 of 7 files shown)


davidkpiano:
Let's go from `internal: true`/`external: true` to `type: 'internal'` or `type: 'external'`

davidkpiano:
Are we renaming to `reenter` here or in a separate PR?

Andarist:
In light of https://discord.com/channels/795785288994652170/800812262473728000/1086277244147748936 , I would prefer to just do that rename in a quick follow-up - because it will touch a little bit more than just the name of the property

davidkpiano (approved):

PR #3762, opened 2023-01-15
[v5] Behavior consistency

davidkpiano:


packages/xstate-react/test/typegenTypes.test.typegen.ts
```diff
@@ -0,0 +1,301 @@
+// This file was automatically generated. Edits will be overwritten
```
Andarist, line 1: this file shouldn't be committed - we probably need some option in the extension to disallow typegen for the whole workspace

packages/core/test/typegenTypes.test.typegen.ts
```diff
@@ -0,0 +1,761 @@
+// This file was automatically generated. Edits will be overwritten
```
Andarist, line 1: this also shouldn't be committed

packages/core/test/typeHelpers.test.typegen.ts
```diff
@@ -0,0 +1,202 @@
+// This file was automatically generated. Edits will be overwritten
```
Andarist, line 1: this also shouldn't be committed

packages/core/test/typeHelpers.test.ts
```diff
@@ -85,7 +85,7 @@ describe('EventFrom', () => {
 
   it('should return events for a typegened machine', () => {
     const machine = createMachine({
-      tsTypes: {} as TypegenMeta,
+      tsTypes: {} as import('./typeHelpers.test.typegen').Typegen3,
```
Andarist, line 88:
```suggestion
      tsTypes: {} as TypegenMeta,
```

packages/core/src/types.ts
```diff
@@ -850,20 +850,20 @@ type MachineImplementationsGuards<
 type MachineImplementationsActors<
```
Andarist, line 850: this type should also be renamed

packages/core/src/types.ts
```diff
@@ -600,7 +600,7 @@ export interface StateNodeConfig<
    */
   states?: StatesConfig<TContext, TEvent, TAction> | undefined;
   /**
-   * The services to invoke upon entering this state node. These services will be stopped upon exiting this state node.
+   * The actors to invoke upon entering this state node. These actors will be stopped upon exiting this state node.
```
Andarist, line 603: are those actors to invoke though? 🤔 actors are definitely invoked as the end result, but the `invoke` property holds behaviors/behavior factories

packages/core/src/typegenTypes.ts
```diff
@@ -167,7 +167,7 @@ type MergeWithInternalEvents<TIndexedEvents, TInternalEvents> = TIndexedEvents &
 
 type AllowAllEvents = {
   eventsCausingActions: Record<string, string>;
-  eventsCausingActors: Record<string, string>;
+  eventsCausingBehaviors: Record<string, string>;
```
Andarist, line 170: I'm glad that the renamed property stays at the same position when we consider the alphabetical order 🤣

packages/core/src/dev/index.ts
```diff
@@ -2,7 +2,7 @@ import { IS_PRODUCTION } from '../environment';
 import { AnyInterpreter, DevToolsAdapter } from '../types';
 
 interface DevInterface {
-  services: Set<AnyInterpreter>;
+  behaviors: Set<AnyInterpreter>;
```
Andarist, line 5: this property looks suspicious - `AnyInterpreter` is already an actor and not a behavior

packages/core/src/behaviors/index.ts
```diff
@@ -0,0 +1,89 @@
...
+  | typeof startSignalType
+  | typeof stopSignalType;
+
+/**
+ * An object that expresses the behavior of an actor in reaction to received events,
+ * as well as an optionally emitted stream of values.
+ *
+ * @template TReceived The received event
+ * @template TSnapshot The emitted value
+ */
```
Andarist, line 40: this comment is definitely misplaced

packages/core/src/behaviors/eventObservable.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Creates an event observable behavior that listens to an observable
+ * that delivers event objects.
+ *
+ *
+ * @param lazyObservable A function that creates an observable
+ * @returns An event observable behavior
+ */
+
```
Andarist, line 13:
```suggestion
```

Andarist:
Overall, I'm OK with the change. However, I still find it not so beginner-friendly to have both actors and behaviors. Perhaps this is not a big deal though. I think though that we should pull into the loop the rest of the team here, especially Laura and Nick. This stuff will have to be explained in the docs and the elements of the UI will refer to those things. If we target more than devs with our product - we also need to think about how easily those terms can be adopted by non-technical folks. Theoretically, both groups could use a different vocabulary (different terms could be used in the code/API and different names could be displayed on the UI) but I feel like that is a recipe for a disaster and a communication barrier that we shouldn't introduce.

davidkpiano:
Superseded by #3770

Issue #3764, opened 2023-01-18
Bug: No types on Deno


guy-borderless:
### Description

Hello!
when importing xstate in deno like so: 
`import * as xstate from "https://deno.land/x/xstate@xstate@4.35.2/src/index.ts";`

loaded code has no types, probably since xstate doesn't use the _.ts_ file extensions when importing files. 
`import { createMachine, Machine } from './Machine'; `
=>
`import { createMachine, Machine } from './Machine.ts';`

### Expected result

typescript types should be loaded

### Actual result

everything is typed as any

### Reproduction

import * as xstate from "https://deno.land/x/xstate@xstate@4.35.2/src/index.ts";

Andarist:
This will be fixed in the next major version of XState.

Andarist:
We landed the appropriate change on the `next` branch: https://github.com/statelyai/xstate/pull/3765

Andarist:
Actually, while that PR switches the repo to `moduleResolution: node16` and thus to using explicit extensions - it doesn't fix the issue at hand here. The repo is now using `.js` extensions in import sources - but Deno won't understand it as it's actually using `.ts`.

We gonna fix this and provide a way to import XState correctly in Deno but I need to rethink our options.

davidkpiano:
Can you check `xstate@beta` and see if it now has types?

guy-borderless:
importing [xstate](https://deno.land/x/xstate) 5 latest- alpha 6 I see types!
`import * as xstate from "https://deno.land/x/xstate@xstate@5.0.0-alpha.6/src/index.ts"`

Andarist:
How does `deno.land` work today with the recently added support for `node_modules` in deno? It feels to me that, ideally, you shouldn't reach into `/src` sirectory when importing `xstate`.

guy-borderless:
Personally I like that the import is not abstracted, like it is with local imports. Easier to debug and understand imported code IMO. For the main module in deno there is a [mod.ts convention](https://deno.land/manual@v1.32.4/references/contributing/style_guide#do-not-use-the-filename-indextsindexjs). Also, [This is an interesting read](https://deno.com/blog/package-json-support). if I understand correctly the next version of deno will formalize the mod.ts pattern to encourage npm style import with the "deno:" URL scheme. I think the more ESM the better, great ergonomics and less power to npm.

davidkpiano:
Closing as it seems that `xstate@beta` in Deno how provides types 🎉 

[...]

PR #3765, opened 2023-01-18
Use `moduleResolution: node16`

Andarist:
closes https://github.com/statelyai/xstate/issues/3764


packages/core/package.json
```diff
@@ -10,6 +10,10 @@
       "module": "./dist/xstate.esm.js",
       "default": "./dist/xstate.cjs.js"
     },
+    "./scxml": {
+      "module": "./scxml/dist/xstate-scxml.esm.js",
+      "default": "./scxml/dist/xstate-scxml.cjs.js"
+    },
```
Andarist, line 16: we might want to remove this - but since it was imported by another package, I had to add it for the time being

Andarist: Hm, we might want to juggle this away from the core before landing this PR since this is causing a *legit* CI failure:
> error xstate "xml-js" is imported by "src/scxml.ts" but the package is not specified in dependencies or peerDependencies

And we don't want to add `xml-js` to `xstate`'s deps

davidkpiano: Yes, feel free to remove it

davidkpiano (approved):

davidkpiano (approved):

PR #3767, opened 2023-01-18
[v5] Add test from #3752

davidkpiano:
This PR adds the v4 test from #3752


Andarist:
Hm, this should just come to the `next` branch through a regular `main`->`next` merge. Why do we need a separate PR for this?

PR #3770, opened 2023-01-20
[v5] Organize actors

davidkpiano:
This PR organizes reducer, promise, observable, and callback actors in the codebase


packages/core/package.json
```diff
@@ -26,6 +22,26 @@
...
+      "default": "./actors/reducer/dist/xstate-actors-reducer.cjs.js"
+    },
+    "./actors/callback": {
+      "module": "./actors/callback/dist/xstate-actors-callback.esm.js",
+      "default": "./actors/callback/dist/xstate-actors-callback.cjs.js"
+    },
+    "./actors/observable": {
+      "module": "./actors/observable/dist/xstate-actors-observable.esm.js",
+      "default": "./actors/observable/dist/xstate-actors-observable.cjs.js"
+    },
```
Andarist, line 44: What is the goal of those entries though? They are not even suggested by VS Code as a valid autoimport target:
auto-import suggestion in the VS Code

I've done some testing and VS Code always filters out "duplicate" entries like this and prefers paths that are the shortest/closest to the root.

My initial fear was those duplicate entries would be displayed - which would be worse than the current situation. But even if they are not displayed there, I don't think this brings a lot of value to the table. Would those be documented anyhow? What we'd use in the docs examples? When one would like to use the longer ones?

davidkpiano: My motivation was for "import only what you need". Maybe there's a better way to accomplish that?

Andarist: > My motivation was for "import only what you need". Maybe there's a better way to accomplish that?

In a way, this doesn't accomplish this in full for actors since `fromObservable` and `fromEventObservable` live in the same entry :p

> That is, if I only invoke promises, I don't want to include the code for invoking observables, reducers, callbacks, etc. etc. (there might be more in the future)

This stuff is quite easily tree-shakeable. If you are not using a bundler then sure, you might pull some extra code into your runtime - but if that's a concern for somebody then they should use tooling. It's rather an impossible task for us to chunk everything into the smallest bits possible - we don't aim for 1 file-1 export structure.

davidkpiano: We can remove the entries but leave them in these files for organization, what do you think?

Andarist: Works for me

Andarist: I pushed out a commit with just that

Andarist (approved):

PR #3778, opened 2023-01-22
[xstate/react] Add `createActorContext()`

davidkpiano:
This PR adds the `createActorContext()` helper for creating global actors:

```tsx
const someMachine = createMachine({ ... });

const SomeContext = createActorContext(someMachine);

function App() {
  return (
    <SomeContext.Provider>
      <Component />
    </SomeContext.Provider>
  );
);

function Component() {
  // getting the actor instance
  const actor = SomeContext.useActorRef();

  // getting the actor state + send
  const [state, send] = SomeContext.useActor();

  // getting an actor snapshot value
  const count = SomeContext.useSelector(state => state.context.count);
  
  // ...
}
```

You can also provide a machine value:

```tsx
// ...

function App() {
  return (
    <SomeContext.Provider value={machine.withConfig({ ... })}>
      // ...
    </SomeContext.Provider>
  );
}
```

### DIFF
--- a/.changeset/tender-dragons-leave.md
+++ b/.changeset/tender-dragons-leave.md
@@ -1,0 +1,41 @@
+---
+'@xstate/react': minor
+---
+
+The `createActorContext(...)` helper has been introduced to make global actors easier to use with React. It outputs a React Context object with the following properties:
+
+- `.Provider` - The React Context provider
+- `.useActor(...)` - A hook that can be used to get the current state and send events to the actor
+- `.useSelector(...)` - A hook that can be used to select some derived state from the actor's state
+- `.useActorRef()` - A hook that can be used to get a reference to the actor that can be passed to other components
+
+Usage:
+
+```jsx
+import { createActorContext } from '@xstate/react';
+import { someMachine } from './someMachine';
+
+// Create a React Context object that will interpret the machine
+const SomeContext = createActorContext(someMachine);
+
+function SomeComponent() {
+  // Get the current state and `send` function
+  const [state, send] = SomeContext.useActor();
+
+  // Or select some derived state
+  const someValue = SomeContext.useSelector((state) => state.context.someValue);
+
+  // Or get a reference to the actor
+  const actorRef = SomeContext.useActorRef();
+
+  return (/* ... */);
+}
+
+function App() {
+  return (
+    <SomeContext.Provider>
+      <SomeComponent />
+    </SomeContext.Provider>
+  );
+}
+```
--- a/packages/xstate-react/src/createActorContext.tsx
+++ b/packages/xstate-react/src/createActorContext.tsx
@@ -1,13 +1,12 @@
 import * as React from 'react';
-import { createContext, useContext } from 'react';
 import { useInterpret } from './useInterpret';
-import { useActor } from './useActor';
-import { useSelector } from './useSelector';
+import { useActor as useActorUnbound } from './useActor';
+import { useSelector as useSelectorUnbound } from './useSelector';
 import { ActorRefFrom, AnyStateMachine, EmittedFrom, EventFrom } from 'xstate';
 
 export function createActorContext<TMachine extends AnyStateMachine>(
   machine: TMachine
-): React.Context<ActorRefFrom<TMachine>> & {
+): {
   useActor: () => [
     EmittedFrom<ActorRefFrom<TMachine>>,
     (event: EventFrom<TMachine>) => void
@@ -21,9 +20,7 @@ export function createActorContext<TMachine extends AnyStateMachine>(
     children: React.ReactNode;
   }) => React.ReactElement<any, any>;
 } {
-  const ReactContext = createContext<ActorRefFrom<TMachine> | null>(
-    null
-  ) as any; // TODO: fix types
+  const ReactContext = React.createContext<ActorRefFrom<TMachine> | null>(null);
 
   const OriginalProvider = ReactContext.Provider;
 
@@ -35,8 +32,8 @@ export function createActorContext<TMachine extends AnyStateMachine>(
 
   Provider.displayName = `Provider(${machine.id})`;
 
-  ReactContext.useActorRef = () => {
-    const actor = useContext(ReactContext) as any;
+  function useActorRef() {
+    const actor = React.useContext(ReactContext);
 
     if (!actor) {
       throw new Error(
@@ -45,24 +42,25 @@ export function createActorContext<TMachine extends AnyStateMachine>(
     }
 
     return actor;
-  };
+  }
 
-  ReactContext.useActor = () => {
-    const actor = useContext(ReactContext) as any;
+  function useActor() {
+    const actor = useActorRef();
+    return useActorUnbound(actor);
+  }
 
-    return useActor(actor);
-  };
-
-  ReactContext.useSelector = <T,>(
+  function useSelector<T>(
     selector: (snapshot: EmittedFrom<TMachine>) => T,
     compare?: (a: T, b: T) => boolean
-  ): T => {
-    const actor = useContext(ReactContext) as any;
+  ): T {
+    const actor = useActorRef();
+    return useSelectorUnbound(actor, selector, compare);
+  }
 
-    return useSelector(actor, selector, compare);
+  return {
+    Provider,
+    useActorRef,
+    useActor,
+    useSelector
   };
-
-  ReactContext.Provider = Provider;
-
-  return ReactContext;
 }
--- a/packages/xstate-react/src/createActorContext.tsx
+++ b/packages/xstate-react/src/createActorContext.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { useInterpret } from './useInterpret';
 import { useActor as useActorUnbound } from './useActor';
 import { useSelector as useSelectorUnbound } from './useSelector';
-import { ActorRefFrom, AnyStateMachine, EmittedFrom, EventFrom } from 'xstate';
+import { ActorRefFrom, AnyStateMachine, EmittedFrom } from 'xstate';
 
 export function createActorContext<TMachine extends AnyStateMachine>(
   machine: TMachine
@@ -12,10 +12,10 @@ export function createActorContext<TMachine extends AnyStateMachine>(
     selector: (snapshot: EmittedFrom<TMachine>) => T,
     compare?: (a: T, b: T) => boolean
   ) => T;
-  useContext: () => ActorRefFrom<TMachine>;
+  useActorRef: () => ActorRefFrom<TMachine>;
   Provider: (props: {
     children: React.ReactNode;
-    value?: TMachine | (() => TMachine);
+    machine?: TMachine | (() => TMachine);
   }) => React.ReactElement<any, any>;
 } {
   const ReactContext = React.createContext<ActorRefFrom<TMachine> | null>(null);
@@ -24,12 +24,12 @@ export function createActorContext<TMachine extends AnyStateMachine>(
 
   function Provider({
     children,
-    value = machine
+    machine: providedMachine = machine
   }: {
     children: React.ReactNode;
-    value: TMachine;
+    machine: TMachine | (() => TMachine);
   }) {
-    const actor = useInterpret(value) as ActorRefFrom<TMachine>;
+    const actor = useInterpret(providedMachine) as ActorRefFrom<TMachine>;
 
     return <OriginalProvider value={actor}>{children}</OriginalProvider>;
   }
@@ -63,7 +63,7 @@ export function createActorContext<TMachine extends AnyStateMachine>(
 
   return {
     Provider,
-    useContext,
+    useActorRef: useContext,
     useActor,
     useSelector
   };
--- a/packages/xstate-react/src/createActorContext.tsx
+++ b/packages/xstate-react/src/createActorContext.tsx
@@ -16,7 +16,7 @@ export function createActorContext<TMachine extends AnyStateMachine>(
     selector: (snapshot: EmittedFrom<TMachine>) => T,
     compare?: (a: T, b: T) => boolean
   ) => T;
-  useContext: () => ActorRefFrom<TMachine>;
+  useActorRef: () => ActorRefFrom<TMachine>;
   Provider: (props: {
     children: React.ReactNode;
   }) => React.ReactElement<any, any>;
@@ -35,7 +35,7 @@ export function createActorContext<TMachine extends AnyStateMachine>(
 
   Provider.displayName = `Provider(${machine.id})`;
 
-  ReactContext.useContext = () => {
+  ReactContext.useActorRef = () => {
     const actor = useContext(ReactContext) as any;
 
     if (!actor) {
--- a/packages/xstate-react/src/createActorContext.tsx
+++ b/packages/xstate-react/src/createActorContext.tsx
@@ -7,10 +7,7 @@ import { ActorRefFrom, AnyStateMachine, EmittedFrom, EventFrom } from 'xstate';
 export function createActorContext<TMachine extends AnyStateMachine>(
   machine: TMachine
 ): {
-  useActor: () => [
-    EmittedFrom<ActorRefFrom<TMachine>>,
-    (event: EventFrom<TMachine>) => void
-  ];
+  useActor: () => ReturnType<typeof useActorUnbound<ActorRefFrom<TMachine>>>;
   useSelector: <T>(
     selector: (snapshot: EmittedFrom<TMachine>) => T,
     compare?: (a: T, b: T) => boolean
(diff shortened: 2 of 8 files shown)


matthewdavi:
Why not make a contextless version using a singleton hook similar to how zustand works?

davidkpiano:
> Why not make a contextless version using a singleton hook similar to how zustand works?

Good question. We've explored that idea with @Andarist and came to the conclusion that explicitly providing through context is more robust, especially since the actors are also responsible for executing effects.

But if you wanted to use a singleton hook like Zustand, it's already really easy to do so:

```js
const actor = interpret(machine).start();

function Component() {
  const [state, send] = useActor(actor);

  // ...
}
```

judicaelandria:
> This PR adds the `createProvider()` helper for creating global actors:
> 
> ```tsx
> const someMachine = createMachine({ ... });
> 
> const SomeProvider = createProvider(someMachine);
> 
> function App() (
>   <SomeProvider>
>     <Component />
>   </SomeProvider>
> );
> 
> function Component() {
>   // getting the actor instance
>   const actor = SomeProvider.useContext();
> 
>   // getting the actor state + send
>   const [state, send] = SomeProvider.useActor();
> 
>   // getting an actor snapshot value
>   const count = SomeProvider.useSelector(state => state.context.count);
>   
>   // ...
> }
> ```

maybe a cool feature to have, we can pass an array of machine?
```ts
const SomeProvider = createProvider([someMachine, anotherMachine]);
```

packages/xstate-react/src/createProvider.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import { createContext, useContext } from 'react';
+import { useInterpret } from './useInterpret';
+import { useActor } from './useActor';
+import { useSelector } from './useSelector';
+import { ActorRefFrom, AnyStateMachine, EmittedFrom } from 'xstate';
+import { EventFrom } from '@xstate/fsm';
```
Andarist, line 7:
```suggestion
import { ActorRefFrom, AnyStateMachine, EmittedFrom, EventFrom } from 'xstate';
```

davidkpiano:
> maybe a cool feature to have, we can pass an array of machine?
> 
> ```ts
> const SomeProvider = createProvider([someMachine, anotherMachine]);
> ```

@judicaelandria Hmm interesting idea but I would recommend using multiple providers instead (for now).

packages/xstate-react/src/createProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export function createProvider<TMachine extends AnyStateMachine>(
+  machine: TMachine
+): {
+  (props: { children: React.ReactNode }): JSX.Element;
```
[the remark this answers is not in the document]
judicaelandria: I'm curious! why ```React.ReactElement<any,any>``` is better option here that ```JSX.Element```

judicaelandria:
> > maybe a cool feature to have, we can pass an array of machine?
> > ```ts
> > const SomeProvider = createProvider([someMachine, anotherMachine]);
> > ```
> 
> @judicaelandria Hmm interesting idea but I would recommend using multiple providers instead (for now).

okay @davidkpiano ! but this can avoid us to have many providers tho

packages/xstate-react/src/createProvider.tsx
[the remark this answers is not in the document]
Andarist: They are pretty much the same - `JSX.Element` is actually defined like this:
```ts
declare global {
    namespace JSX {
        interface Element extends React.ReactElement<any, any> { }
    }
}
```

The main difference for me is that... we have stronger guarantees that we'll resolve the correct thing. Support for JSX in TS is weirdly implemented, so I just make sure that we don't try to resolve to a global `JSX` since that could be somehow tampered with. 

It also matched how the builtin `React.FC` (and more) are typed - so I want to follow their suite.

packages/xstate-react/src/createProvider.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import { createContext, useContext } from 'react';
+import { useInterpret } from './useInterpret';
+import { useActor } from './useActor';
+import { useSelector } from './useSelector';
+import { ActorRefFrom, AnyStateMachine, EmittedFrom, EventFrom } from 'xstate';
+
+export function createProvider<TMachine extends AnyStateMachine>(
```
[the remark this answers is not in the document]
davidkpiano: You mean like this?

```jsx
const actorContext = createReactContext(someMachine);

// ...

<actorContext.Provider>
  // ...
</actorContext.Provider>

// ...

const [state, send] = actorContext.useActor();

const data = actorContext.useSelector(...);
```

I'm okay with that; I just thought that "context" would be potentially confusing with XState context, but we don't have any similarly-named hooks (`context` is more-or-less internal to the machine/state in XState) so maybe it's not a big deal.

packages/xstate-react/src/createProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+  useSelector: <T>(selector: (snapshot: EmittedFrom<TMachine>) => T) => T;
+  useContext: () => ActorRefFrom<TMachine>;
+} {
+  const ReactContext = createContext(null as any);
+
+  function Provider({ children }: { children: React.ReactNode }) {
+    const actor = useInterpret(machine);
+
+    return (
+      <ReactContext.Provider value={actor}>{children}</ReactContext.Provider>
```
YagamiNewLight, line 25: I think here instead of coupling with the `actor`, it should be open for the user to pass in their own dependencies here.
basically `value={actor}` => `value={value?? actor}`

judicaelandria: but the dependency is the actor! what other value can we pass here? 
and also, in this case, it's exactly like react context

davidkpiano: > 

So you're thinking of this usage?

```jsx
// another actor of same type
<SomeProvider value={anotherActor}>
  // ...
</SomeProvider>
```

And also:
```jsx
// another machine of same type
<SomeProvider value={anotherMachine}>
  // ...
</SomeProvider>
```

packages/xstate-react/src/createProvider.tsx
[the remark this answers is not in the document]
Andarist: > You mean like this?

Ye.

> I'm okay with that; I just thought that "context" would be potentially confusing with XState context, but we don't have any similarly-named hooks (context is more-or-less internal to the machine/state in XState) so maybe it's not a big deal.

This is a valid concern. OTOH, we already have this naming clash with React - so it's not an entirely new problem

horacioh:
what would be de benefit on having this in the library instead of documentation on how to do it ourselves?

I do have my abstractions for this on my apps, as [others](https://twitter.com/erikras/status/1617538399438790656) too, and maybe this is enough? I see the potential for sure!, not sure why have it in the actual library is my point

with-heart:
> what would be de benefit on having this in the library instead of documentation on how to do it ourselves?

It's a common pattern React pattern that many devs are familiar with but the knowledge required to create it yourself with `@xstate/react` isn't immediately available to you when you first start using xstate. 

So there's this disconnect where a pattern that should be easy for you because you do it often in React isn't because you aren't familiar with how to do it with this particular library.

David mentioned this after the Stately hackathon was over on Friday and it was definitely a lesson learned directly from that experience. A lot of teams were able to move quickly on a lot of fronts but they found that they weren't able to figure out how to make their machines global. 

This allows the library to solve that particular problem up front rather than keeping it behind a knowledge gate so people familiar with React can move faster using familiar tools

davidkpiano:
> IMHO, people should embrace copy-pasting code more 😉

Somewhat agree, but admittedly this is a problem for two audiences: junior developers who might not have as much experience with React Context (which is not an intuitive API and has a slight learning curve) as well as senior developers who would struggle with the typings for doing this manually.

I strongly believe that it is much more helpful than not to provide a "blessed" and optional way to do something than to not provide it at all, especially if the maintenance cost is low (as it is here - this is not a complex implementation).

davidkpiano:
@horacioh Even if they understand both, it's wiring them together that can get confusing. Traditionally, React Context is used to share _state_ that seldom changes, whereas we're using it to share a subscribable _actor_.

The temptation is to do this:

```jsx

return <SomeContext.Provider value={state} />
// or...
return <SomeContext.Provider value={{state, send}} />
```

which can introduce bad performance.

And even if you know to provide the actor itself, this is not intuitive:

```ts
const SomeContext = React.createContext(/* what goes here?? */);

// and worse in TypeScript:
const SomeContext = React.createContext</* ??? */>(/* ??? */);
```

Andarist:
> Traditionally, React Context is used to share state that seldom changes, whereas we're using it to share a subscribable actor.

I would say that traditionally context was used to inject all sorts of things (like a Redux store, history object for routers etc) - it feel like a newer "trend" that people started to using it as state management solution ;p

judicaelandria:
> I think that most of them can usually be addressed by docs and recipes

Somewhat I completely agree with @Andarist. but @davidkpiano got a point on this too 
> senior developers who would struggle with the typings for doing this manually.

with-heart:
> IMHO, people should embrace copy-pasting code more 😉

Just look at it like this PR being the copy-paste to end all copy-pastes? 😆

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { createContext, useContext } from 'react';
+import { useInterpret } from './useInterpret';
+import { useActor } from './useActor';
+import { useSelector } from './useSelector';
+import { ActorRefFrom, AnyStateMachine, EmittedFrom, EventFrom } from 'xstate';
+
+export function createActorContext<TMachine extends AnyStateMachine>(
+  machine: TMachine
+): React.Context<ActorRefFrom<TMachine>> & {
+  // (props: { children: React.ReactNode }): React.ReactElement<any, any>;
```
Andarist, line 11:
```suggestion
```

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  function Provider({ children }: { children: React.ReactNode }) {
+    const actor = useInterpret(machine) as ActorRefFrom<TMachine>;
+
+    return <OriginalProvider value={actor}>{children}</OriginalProvider>;
+  }
+
+  Provider.displayName = `Provider(${machine.id})`;
+
+  ReactContext.useContext = () => {
```
Andarist, line 39: The name of this one is super generic - I wonder if we couldn't rename it to something like `useActorRef`? That would make us have both `useActorRef` and `useActor`.

I'm not sure if that particular proposal is better - but it feels like a better name for this one would be appreciated. Let's not bother with it too much if we can't figure out a better one though

davidkpiano: We can also omit this in favor of the very idiomatic:

```js
const actorRef = useContext(SomeContext);

// versus:
// const actorRef = SomeContext.useContext();
```

Andarist: We could, but then we'd have to also return that `SomeContext` to be usable with `React.useContext` (so we'd have to name it, even if that would be `.context`). 

I think I would prefer to keep this extra helper - especially since it "validates" if the thing was actually provided.

davidkpiano: > We could, but then we'd have to also return that `SomeContext` to be usable with `React.useContext` (so we'd have to name it, even if that would be `.context`).
> 
> I think I would prefer to keep this extra helper - especially since it "validates" if the thing was actually provided.

Yep, I was just about to implement this and realized we don't get any helpful errors if the Provider is omitted.

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -15,24 +15,31 @@ export function createActorContext<TMachine extends AnyStateMachine>(
...
 } {
   const ReactContext = React.createContext<ActorRefFrom<TMachine> | null>(null);
 
   const OriginalProvider = ReactContext.Provider;
 
-  function Provider({ children }: { children: React.ReactNode }) {
-    const actor = useInterpret(machine) as ActorRefFrom<TMachine>;
+  function Provider({
+    children,
+    value = machine
```
Andarist, line 30: When it comes to the barebones React context `value` is both what you provide and what you consume. We break that parity here because we provide an actor ref under the hood but the `value` here is the machine. 

I'd probably name this prop a `machine` for this reason.

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -15,24 +15,31 @@ export function createActorContext<TMachine extends AnyStateMachine>(
     selector: (snapshot: EmittedFrom<TMachine>) => T,
     compare?: (a: T, b: T) => boolean
   ) => T;
-  useActorRef: () => ActorRefFrom<TMachine>;
+  useContext: () => ActorRefFrom<TMachine>;
   Provider: (props: {
     children: React.ReactNode;
+    value?: TMachine | ActorRefFrom<TMachine>;
```
Andarist, line 21: This type is incorrect and it was only allowed because we still don't use `strictFunctionTypes: true`.

This `value` is passed to the `useInterpret` and that doesn't accept actor refs - it only accepts machines (and machine factories):

It should be:
```suggestion
    value?: TMachine | (() => TMachine);
```

packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -211,4 +211,31 @@ describe('createActorContext', () => {
...
+    const Component = () => {
+      const actor = SomeContext.useContext();
+      const count = useSelector(actor, (state) => state.context.count);
+
+      return <div data-testid="value">{count}</div>;
+    };
+
+    const App = () => {
+      return (
+        <SomeContext.Provider value={someMachine.withContext({ count: 42 })}>
```
Andarist, line 231:
```suggestion
        <SomeContext.Provider value={() => someMachine.withContext({ count: 42 })}>
```

.changeset/little-frogs-fly.md
```diff
@@ -0,0 +1,5 @@
+---
```
Andarist, line 1: This change doesn't belong to this branch

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { useSelector as useSelectorUnbound } from './useSelector';
+import { ActorRefFrom, AnyStateMachine, EmittedFrom, EventFrom } from 'xstate';
+
+export function createActorContext<TMachine extends AnyStateMachine>(
+  machine: TMachine
+): {
+  useActor: () => [
+    EmittedFrom<ActorRefFrom<TMachine>>,
+    (event: EventFrom<TMachine>) => void
+  ];
```
Andarist, line 13: 😃 
```suggestion
useActor: typeof useActorUnbound<ActorRefFrom<TMachine>>;
```

packages/xstate-react/src/createActorContext.tsx
Andarist: OTOH, we already "break" the parity with the barebones React context as the "initial value" used to create this is different from what we end up providing.

But even considering this - I feel like a more descriptive prop name (`machine`) is better. React has to use a generic `value` because their whole signature is super generic and has to accept just any value.

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }: {
+    children: React.ReactNode;
+    value: TMachine;
+  }) {
+    const actor = useInterpret(value) as ActorRefFrom<TMachine>;
+
+    return <OriginalProvider value={actor}>{children}</OriginalProvider>;
+  }
+
+  Provider.displayName = `Provider(${machine.id})`;
```
Andarist, line 40:
```suggestion
  Provider.displayName = `ActorProvider(${machine.id})`;
```

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+): {
+  useActor: () => [
+    EmittedFrom<ActorRefFrom<TMachine>>,
+    (event: EventFrom<TMachine>) => void
+  ];
+  useSelector: <T>(
+    selector: (snapshot: EmittedFrom<TMachine>) => T,
+    compare?: (a: T, b: T) => boolean
+  ) => T;
+  useContext: () => ActorRefFrom<TMachine>;
```
Andarist, line 18: I think that I preferred `useActorRef` for this - but it's up to you.

.changeset/tender-dragons-leave.md
```diff
@@ -0,0 +1,41 @@
+---
+'@xstate/react': minor
+---
+
+The `createActorContext(...)` helper has been introduced to make global actors easier to use with React. It outputs a React Context object with the following properties:
```
Andarist, line 5: This could also get to the README

Andarist (approved):

.changeset/tender-dragons-leave.md
davidkpiano: Will do after release 👍

PR #3780, opened 2023-01-23
Fix external transition examples

davidkpiano:


docs/fr/guides/transitions.md
```diff
@@ -240,14 +240,17 @@ External transitions _will_ exit and reenter the state node in which the transit
...
-  RIGHT_CLICK: 'word.right',
-  CENTER_CLICK: { target: '.center', internal: false }, // same as 'word.center'
-  JUSTIFY_CLICK: { target: 'word.justify', internal: false } // same as 'word.justify'
+states: {
+  loading: {
+    entry: 'someEntryAction',
+    exit: 'someExitAction',
+    on: {
+      RETRY: { target: 'loading' },
+      RETRY2: { target: 'loading', internal: false }, // same as above
```
Andarist, line 251: I think that the example doesn't quite fit the text above it:
> By default, transitions are external, but any transition can be made external by explicitly setting `{ internal: false }` on the transition.

It feels that this example miss an example of an internal transition that is being converted to an external one using `internal: false`

PR #3785, opened 2023-01-24
Code juggle `createActorContext` a little bit

Andarist:


packages/xstate-react/src/createActorContext.tsx
```diff
@@ -1,14 +1,12 @@
...
 import { useInterpret } from './useInterpret';
-import { useActor } from './useActor';
-import { useSelector } from './useSelector';
+import { useActor as useActorUnbound } from './useActor';
+import { useSelector as useSelectorUnbound } from './useSelector';
 import { ActorRefFrom, AnyStateMachine, EmittedFrom, EventFrom } from 'xstate';
 
 export function createActorContext<TMachine extends AnyStateMachine>(
   machine: TMachine
-): React.Context<ActorRefFrom<TMachine>> & {
```
Andarist, line 10: I don't think that we need to intersect with `React.Context` here, we can slightly adjust the return type's shape (we extend its structure anyway).

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -46,24 +42,25 @@ export function createActorContext<TMachine extends AnyStateMachine>(
...
+    const actor = useContext();
+    return useActorUnbound(actor);
+  }
 
-  ReactContext.useSelector = <T,>(
+  function useSelector<T>(
     selector: (snapshot: EmittedFrom<TMachine>) => T,
     compare?: (a: T, b: T) => boolean
-  ): T => {
-    const actor = useContext(ReactContext) as any;
```
Andarist, line 61: Fixed the usage of `useContext` here. We should grab it from the "validated" `useContext` and not directly from the created React context - this way we will throw if the user forgets to use the provider.

davidkpiano (approved):

PR #3788, opened 2023-01-26
Add Angular template

aldovelco:
Adds an Angular template. It could be simpler with the new standalone components API but I suppose most projects are still using NgModules in the wild.

Should I remove stuff that was automatically generated by the Angular CLI like the `.editorconfig` file?
I'd also like to make a copy of the TodoMVC examples for React and Vue but this time implemented with Angular.

image


Andarist:
> Should I remove stuff that was automatically generated by the Angular CLI like the .editorconfig file?

I'd say yes - although that's probably the only file that should be removed.

examples/template-angular-ts/src/app/app.component.ts
```diff
@@ -0,0 +1,14 @@
+import { Component } from '@angular/core';
```
Andarist, line 1: > Adds an Angular template. It could be simpler with the new standalone components API but I suppose most projects are still using NgModules in the wild.

How this file would look like with the new standalone components API?

examples/template-angular-ts/src/app/app.component.ts
```diff
@@ -0,0 +1,14 @@
...
+import { toggleMachine } from 'src/toggle.machine';
+import { interpret } from 'xstate';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: './app.component.html',
+  styleUrls: ['./app.component.css']
+})
+export class AppComponent {
+  actor = interpret(toggleMachine).start();
```
Andarist, line 12: shouldn't this happen in `ngOnInit`?

examples/template-angular-ts/src/app/app.component.html
```diff
@@ -0,0 +1,13 @@
+<ng-container *ngIf="state$ | async as state">
```
Andarist, line 1: can `state` ever be undefined here? i might not understand the syntax but I wonder why do we need `ngIf` here. Ideally, the state should be provided synchronously to the template

examples/template-angular-ts/src/app/app.component.ts
aldovelco: Not necessarily. I would move this logic to `ngOnInit` if I had a component input that I wanted to use for the machine context for example. 

Angular TS defaults have been getting more strict. We could put the initialization logic in `ngOnInit` but then we would have to either use the non-null assertion operator, use optional chaining while accessing the actor property or initializing in the constructor which would put us where we started.

e.g.
```ts
export class AppComponent implements OnInit {
  actor!: InterpreterFrom<typeof toggleMachine>;

  ngOnInit(): void {
    this.actor = interpret(toggleMachine).start();
  }
}
```

examples/template-angular-ts/src/app/app.component.html
aldovelco: Yes, you're right. I like using observables on my apps because I use `ChangeDetectionStrategy.OnPush` on all of my components and the `async` pipe marks the view for check behind the scenes whenever the observable emits a value.

I will change it to use the sync version as this is a basic example.

examples/template-angular-ts/src/app/app.component.ts
Andarist: https://angular.io/guide/lifecycle-hooks#initializing-a-component-or-directive

> Components should be cheap and safe to construct. You should not, for example, fetch data in a component constructor. You shouldn't worry that a new component will try to contact a remote server when created under test or before you decide to display it.
> An ngOnInit() is a good place for a component to fetch its initial data.

It's not unlikely that some async jobs could start at the init time of a machine. In such a case, isn't the preferred approach to use `ngOnInit`? 

> Not necessarily. I would move this logic to ngOnInit if I had a component input that I wanted to use for the machine context for example.

It seems that `ngOnInit` is something that fits more use cases, shouldn't it be preferred here even if only for "easy to extend" argument? What I mean is that it's somewhat annoying to move the code around only because you have a new requirement - it's somewhat better to start in a place that is more flexible in the first place.

Note that I'm only thinking out loud/ask questions here - I don't necessarily mean that you have to change this if you believe that this is the right approach.

examples/template-angular-ts/src/app/app.component.html
Andarist: > <ng-container *ngIf="state$ | async as state">

So this is just a standard way of unwrapping a non-nullable observable?

examples/template-angular-ts/src/app/app.component.ts
Andarist: Maybe we should have two templates then? If the new way is fairly new. 

Sideways q: do you know how a reusable `@xstate/angular` could look like?

examples/template-angular-ts/src/app/app.component.html
aldovelco: That's right

Andarist: Roger that, doesn't it get annoying when u have a couple of observables in a component? Or do you `combineLatest` them then, to unwrap and destructure in the template with a single `ng-container`? 🤔

examples/template-angular-ts/src/app/app.component.ts
aldovelco: It's just personal preference, I don't really have a strong opinion about this. I can get behind your "easy to extend" argument. Will change accordingly.

For completeness sake, from I'm modifying an example from https://angular.io/api/common/AsyncPipe:
This would be a valid use case for defining an observable property and then handling it directly with the async pipe.
This wouldn't go against the advice of making an async request on the constructor because the observable is lazy and will only be executed after the async pipe subscribes to it once the component's view has rendered.

```ts
@Component({
  selector: 'async-observable-pipe',
  template: '<div><code>observable | async</code>: Data: {{ data$ | async }}</div>'
})
export class AsyncObservablePipeComponent {
  data$ = this.http.get<{ count: number; }>('...');

  constructor(private http: HttpClient) {}
}
```

examples/template-angular-ts/src/app/app.component.html
aldovelco: I do just that, `combineLatest` and single `ng-container`

examples/template-angular-ts/src/app/app.component.ts
aldovelco: I usually just create a service that will be `providedIn: root` in the case of singletons or inside a component's providers array so I don't think much is really needed because we can just interact with xstate via observables. 

Tim Deschryver has a gist https://gist.github.com/timdeschryver/dcdeefdba43bef7134c8249e8257dae7 with some interesting examples.

The following is the most important part in my opinion:
It would be nice to have a custom rxjs creation operator like `fromInterpreter` or similar to encapsulate this logic. I had a hard time working with the typegen generics though. 
```ts
const state$ = from(service).pipe(
		filter(p => p.changed),
		startWith(service.state),
		shareReplay(1),
		finalize(() => {
			service.stop()
			input.unsubscribe()
		}),
	)
```

Andarist: > It would be nice to have a custom rxjs creation operator like fromInterpreter or similar to encapsulate this logic. I had a hard time working with the typegen generics though.

This makes sense. Should we then aim for creating `@xstate/rxjs` and not `@xstate/angular`? 😅 

The main advantage of `@xstate/angular` would be if that could encapsulate the `.start()`+`.stop()` calls - but I'm not sure if Angular allows for this kind of thing. Can we anyhow automatically slap a custom setup/teardown onto an angular component with a single thing? Perhaps a decorator would be capable of doing that?

Andarist: Tim Deschryver's gist sends events to the machine using an "input observable". This is fine, but at the same time, it's perhaps a little bit redundant. I mean, is there really any practical difference here between a simple `actor.send(ev)` and `events$.next(ev)`?

examples/template-angular-ts/src/app/app.component.ts
Andarist: > This wouldn't go against the advice of making an async request on the constructor because the observable is lazy and will only be executed after the async pipe subscribes to it once the component's view has rendered.

Yeah, this makes sense - it's just that in our case the `.start()` is not that lazy :p

examples/template-angular-ts/src/app/app.component.ts
aldovelco: A popular way of unsubscribing from observables is to use the `takeUntil` operator. For example, there is this [until-destroy library](https://github.com/ngneat/until-destroy) that does it to help with the boilerplate from creating a `destroy$` observable to use inside the `takeUntil` operator.
I suppose we could do something similar to create an observable that emits upon `ngOnInit`.

`ngOnDestroy` is a bit easier to build on user-land because we can take advantage of the fact that Angular Services may implement that lifecycle hook. We can then provide them at the component scope and it will be called once the component is destroyed.

e.g.
`providers: [DestroyService]` is required.
```ts
// xstate-angular.ts
import { inject, Injectable, OnDestroy } from '@angular/core';
import {
  filter,
  finalize,
  from,
  Observable,
  shareReplay,
  startWith,
  Subject,
  takeUntil
} from 'rxjs';
import { AnyStateMachine, interpret, InterpreterFrom, StateFrom } from 'xstate';

@Injectable()
export class DestroyService implements OnDestroy {
  private destroySubject = new Subject<void>();
  readonly destroy$ = this.destroySubject.asObservable();

  ngOnDestroy() {
    this.destroySubject.next();
    this.destroySubject.complete();
  }
}

// I don't know how to type a generic function with the result of `typeof createMachine<???>`
interface InjectMachineReturnType<T extends AnyStateMachine> {
  state$: Observable<StateFrom<T>>;
  send: InterpreterFrom<T>['send'];
  service: InterpreterFrom<T>;
}

export function injectMachine<T extends AnyStateMachine>(machine: T) {
  const destroy$ = inject(DestroyService).destroy$;

  const service = interpret(machine, { devTools: true }).start();
  const state$ = from(service).pipe(
    filter((state) => !!state.changed),
    startWith(service.getSnapshot()),
    shareReplay(1),
    takeUntil(destroy$),
    finalize(() => {
      service.stop();
    })
  );

  return { state$, send: service.send, service } as InjectMachineReturnType<T>;
}
``` 

```ts
// child.component.ts
import { AsyncPipe, NgIf } from '@angular/common';
import { Component } from '@angular/core';
import { map } from 'rxjs';
import { toggleMachine } from 'src/toggle.machine';
import { DestroyService, injectMachine } from 'src/xstate-angular';

@Component({
  selector: 'app-child',
  standalone: true,
  imports: [NgIf, AsyncPipe],
  template: `
    <ng-container *ngIf="state$ | async as state">
      <button type="button" (click)="send({ type: 'TOGGLE' })">
        Click me ({{ state.matches('active') ? '✅' : '❌' }})
      </button>
      <code>
        Toggled
        <strong>{{ state.context.count }}</strong> times
      </code>
    </ng-container>
  `,
  providers: [DestroyService]
})
export class ChildComponent {
  private actor = injectMachine(toggleMachine);
  state$ = this.actor.state$;
  send = this.actor.send;
}
````

Andarist: Hmm, I somewhat understand what is happening here. But then I guess there is no way to combine `injectMachine` and `DestroyService` in a single "thing". Nor there is a way to specify the name of the field "dynamically" (through an argument or something). What should always be possible is to use multiple machines in a given component - this is actually solved by your `injectMachine` example because the developer controls the assignment site. It's just a bummer that this doesn't encapsulate the `providers: [DestroyService]` bit 😢

Andarist: When I browse some different examples on the internet - it seems that it's not uncommon that integrating a thing requires more than a single place to be touched. So perhaps this is just the angular-way and I'm overthinking it? WDYT?

aldovelco: I made an xstate-rxjs version based on xstate's react package so I can use it on my projects while I test it: https://github.com/aldovelco/xstate-libs/tree/main/packages/xstate-rxjs. I'd love to be able to contribute back to this repo though.

It's not uncommon to have to add something to the component providers when you want to tie your logic to the component's lifecycle.

We might be able to encapsulate the service from the providers array but we would still have to use a decorator. I'm still experimenting with the syntax for what an xstate angular package would be like in the same repo: https://github.com/aldovelco/xstate-libs/blob/main/apps/angular-example/src/app/app.component.ts

Andarist: All of this looks pretty neat! I think that landing those 2 in the repo is better than landing the proposed Angular template. The Angular template could still land but ideally, it would already be using our packages.

One feedback regarding `@xstate/rxjs` - I would probably avoid using `use*` naming there. It feels reserved for hooks/composables/etc. It's better to use some names that will feel more natural in the context of RxJS (`from*`?)

aldovelco: Yes, I actually already have a `fromInterpreter` internal function that returns the Observable<`state$`>. I'll continue to iterate over on that repo. Would you prefer it if I close this PR and instead open a proposal for a `@xstate/rxjs` and accompanying `template-rxjs-ts` example?

I'd like to get this going with the rxjs part and then submit a different PR for Angular.

aldovelco:
Closing in favor of https://github.com/statelyai/xstate/pull/3844

PR #3790, opened 2023-01-26
Stop reusing `send` in `raise`

Andarist:


packages/core/test/actions.test.ts
```diff
@@ -2734,6 +2733,67 @@ describe('raise', () => {
       expect(service.getSnapshot().value).toEqual('a');
     }, 50);
   });
+
+  it('should accept event expression', () => {
```
Andarist, line 2737: TODO: add more tests

packages/core/test/types.test.ts
```diff
@@ -253,23 +253,23 @@ describe('Raise events', () => {
               {
                 actions: raise({
                   type: 'ALOHA'
-                }),
+                } as any),
```
Andarist, line 256: improvements to `raise` typings will be in a separate PR

davidkpiano (approved):
Code LGTM

PR #3793, opened 2023-01-27
[xstate/inspect] Add support for custom targetWindow

mdpratt:
This will allow develops to reuse existing tools that they might have already developed, and allow their own custom messages to the same window.
    
Canny:
* https://statelyai.canny.io/devtools/p/inspect-add-targetwindow-to-returned-inspector
* https://statelyai.canny.io/devtools/p/inspect-add-targetwindow-to-inspectoroptions


packages/xstate-inspect/test/inspect.test.ts
```diff
@@ -340,4 +340,40 @@ describe('@xstate/inspect', () => {
         .filter((message: any) => message.type === 'service.event')
     ).toHaveLength(1);
   });
+
+  describe.only('browser inspector', () => {
```
mdpratt, line 344: 🤦‍♂️

packages/xstate-inspect/src/browser.ts
```diff
@@ -216,11 +218,12 @@ export function inspect(options?: InspectorOptions): Inspector | undefined {
     });
 
     iframe.setAttribute('src', String(url));
-  } else {
+  } else if (!targetWindow) {
     targetWindow = window.open(String(url), 'xstateinspector');
   }
 
   return {
+    targetWindow,
```
Andarist, line 226: q: why do we want to return it here?

packages/xstate-inspect/test/inspect.test.ts
```diff
@@ -340,4 +340,40 @@ describe('@xstate/inspect', () => {
         .filter((message: any) => message.type === 'service.event')
     ).toHaveLength(1);
   });
+
+  describe('browser inspector', () => {
```
Andarist, line 344: I would unnest this block - just add the new tests to the existing block. I find `describe` blocks to be hard to work with because people often accidentally insert new tests in a "random" block (well, usually in the last one) without realizing that they are actually in a block.

packages/xstate-inspect/test/inspect.test.ts
```diff
@@ -340,4 +340,40 @@ describe('@xstate/inspect', () => {
...
+    let windowSpy;
+
+    beforeEach(() => {
+      windowSpy = jest.spyOn(window, 'open');
+      windowSpy.mockImplementation(() => windowMock);
+    });
+
+    afterEach(() => {
+      windowSpy.mockRestore();
+    });
```
Andarist, line 355: similarly, i would move this stuff to the tests that are using it - we can wrap repetitive setup in a helper but i find it much easier to reason about when the whole implementation of the test is explicitly stated in the test body itself. Some truly global teardowns can sometimes be put in `afterEach` and stuff like that (I'm not rly dogmatic about it) but mocks, spies etc are (IMHO) better to be initialized in the test themselves

packages/xstate-inspect/test/inspect.test.ts
mdpratt: Sounds good. I'll make this change.

packages/xstate-inspect/test/inspect.test.ts
mdpratt: Sounds good to me. I'll make this change as well.

packages/xstate-inspect/src/browser.ts
mdpratt: By returning the `targetWindow`, it will allow customers to `postMessage` to it. This is useful for custom implementations.

There are two use cases here:
1. If a user does a `window.open` locally, and passes it as an `InspectorOptions` parameter. This is useful if the user wants a "just-in-time" implementation of the XState inspector after they've already opened a window.
2. If a user specifies only the `url`. This is useful if the customer wants an eager implementation of the XState inspector.

 In the first scenario it's simply a pass-through, while in the second it's returning the newly created window.

mdpratt:
@davidkpiano I did some additional testing to iron out exactly how this will behave. The default behavior is exactly the same, so I'll only cover new functionality. Please will your team let me know if these are acceptable.

The reciever logic is quite simple:



### Scenario 1: User provides an open targetWindow

The client logic, as follows:

```typescript
const url = "http://127.0.0.1:5174/"
const targetWindow = window.open(url);
  
const inspector = inspect({
  // The URL must still be provided. This is used by postMessage, as it's
  // not possible to do targetWindow.location for security reasons
  url,
  targetWindow
});

// Post a custom message
targetWindow.postMessage("Hello world", url);
```

The receiver logic is as follows:

```typescript
// Enable XState handling
createWindowReceiver({}).subscribe(console.log);
// Add a generic listener to validate that custom messages are being sent
window.addEventListener("message", console.log);
```

In this scenario, the custom "Hello world" message is being received as expected.

### Scenario 2: User provides window as targetWindow

In this scenario, the user provides the parent `window` as the `targetWindow`. This is actually pretty useful, as it's very easy to add a listener to the same window, pass `{ devTools: true }` to your services, and instantly see what's going on in the same window. Saves from having to add `console.log` into all your machines. This works because a `window` can `postMessage` to itself.

```typescript
inspect({
  url, // This can be constructed from window.location.origin
  window
});

// In the same window, simply listen for messages 
createWindowReceiver({}).subscribe(console.log);

// Start your machine, and all events generated are logged to the console
interpret(machine, { devTools: true }).start({});
```

### Scenario 3: User does not provide a targetWindow

This will work as it does today, except with the returned `targetWindow` value. The client logic, as follows:

```typescript
const url = "http://127.0.0.1:5174/"
  
const inspector = inspect({ 
  url,
  iframe: undefined
});

// Post a custom message
inspector.targetWindow.postMessage("Hello world", url);
```

The receiver logic is as follows:

```typescript
// Enable XState handling
createWindowReceiver({}).subscribe(console.log);
// Add a generic listener to validate that custom messages are being sent
window.addEventListener("message", console.log);
```

In this scenario, the custom "Hello world" message is being received as expected.

packages/xstate-inspect/src/browser.ts
mdpratt: I've outlined behaviors in my comment below: https://github.com/statelyai/xstate/pull/3793#issuecomment-1407377281

davidkpiano (approved):

mdpratt:
@Andarist, let me know if there are any other blocking items for this.

Andarist:
@mdpratt gonna review this tomorrow, together with a couple of other PRs

mdpratt:
> @mdpratt gonna review this tomorrow, together with a couple of other PRs

Great, thank you!

packages/xstate-inspect/src/browser.ts
```diff
@@ -89,10 +89,12 @@ const getFinalOptions = (options?: Partial<InspectorOptions>) => {
 
 const patchedInterpreters = new Set<AnyInterpreter>();
 
-export function inspect(options?: InspectorOptions): Inspector | undefined {
+export function inspect(
+  options?: InspectorOptions
+): BrowserInspector | undefined {
   const { iframe, url, devTools } = getFinalOptions(options);
```
Andarist, line 95: I would move the resolving of the potential `targetWindow` option to this `getFinalOptions` call. Given the existing structure of the code, `options` should not be used anywhere in this function - except for being used as an argument for this very call here.

packages/xstate-inspect/src/browser.ts
Andarist: I think that I would prefer to remove this from the return type altogether. If you want to provide a custom `url` to a custom iframe-based inspector and you need to communicate with it - it doesn't sound too bad if you'd have to create the iframe on your own and pass that as `targetWindow`.

The problem with the current solution in this PR is that `targetWindow` is **not** available synchronously when we are made responsible for creating the iframe. So it won't always be here which makes it harder to document when you can actually rely on this property. We could expose a getter function - but then... how do we communicate to the consumer the current state of the returned value? It feels that it would complicate more than it solves - when creating the window on your own doesn't involve a lot of steps anyway (and you only do that in a single place in your app)

packages/xstate-inspect/src/browser.ts
mdpratt: Sure, I can make this change.

packages/xstate-inspect/src/browser.ts
mdpratt: The iframe is not available synchronously, but the `targetWindow` is. However I understand your comment that sometimes giving a customer too much choice can add additional complexity. If a custom cares to control the window, provide it. Otherwise, let XState do it's thing.

I can make this change as well.

Andarist: > The iframe is not available synchronously, but the targetWindow is.

It might be an illusion - IIRC it's not available synchronously, at least not the one that we are interested in. Even if you set `src` of an iframe before attaching it to the document, it "starts" in something akin to `about:blank` and "redirects" to your actual page when the main thread yields or something. Each navigation creates a new `contentWindow` - so whatever we could grab synchronously would actually be pretty useless for the consumer

mdpratt: This is a good learning for me. I need to do some more deep diving reading into this! Thanks for the insights.

mdpratt:
@Andarist Made the changes to align with your vision. I've also updated the README to give some implementation examples.

packages/xstate-inspect/test/inspect.test.ts
```diff
@@ -340,4 +340,23 @@ describe('@xstate/inspect', () => {
...
+      devTools,
+      iframe: undefined,
+      targetWindow: localWindowMock
+    });
+
+    expect(windowSpy).not.toHaveBeenCalled();
+
+    windowSpy.mockRestore();
+  });
+});
```
Andarist, line 362:
```suggestion
});

```

packages/xstate-inspect/src/types.ts
```diff
@@ -19,6 +19,7 @@ export interface InspectorOptions {
   iframe?: MaybeLazy<HTMLIFrameElement | null | false>;
   devTools?: MaybeLazy<XStateDevInterface>;
   serialize?: Replacer | undefined;
+  targetWindow?: Window | undefined | null;
```
Andarist, line 22: Any reason to include `null` here?
```suggestion
  targetWindow?: Window | undefined;
```

packages/xstate-inspect/README.md
```diff
@@ -55,3 +55,103 @@ import { inspect } from '@xstate/inspect';
...
 service.start();
 ```
+
+## Configuration
+
+* `url` *(optional)* - The endpoint that the Inspector sends events to. Default: https://stately.ai/viz?inspect
+* `iframe` *(optional)* - The iframe that loads the provided URL. If iframe is set to `false`, then a new tab is opened instead.
+* `devTools` *(optional)* - Allows custom implementation for lifecycle hooks.
+* `serialize` *(optional)* - A custom serializer for messages sent to the URL endpoint. Useful for sanitizing sensitive information, such as credentials, from leaving your application.
+* `targetWindow` *(optional)* - Provide a preferred `window.open` location that will be used instead. When using this option, you must still provide the `url` value due to security checks in browser APIs.
```
Andarist, line 65: The note about `window.open` here makes it sound as if this argument would internally be somehow used with it but it's used instead of it. 

I would also include information here that this makes `iframe` option to be ignored.

Maybe something like this?
```suggestion
* `targetWindow` *(optional)* - Provide a pre-existing window location that will be used instead of opening a new window etc. When using this option, you must still provide the `url` value due to security checks in browser APIs, and the `iframe` option is ignored in such a case.
```

packages/xstate-inspect/src/types.ts
mdpratt: TypeScript intelligence reports that `window.open` can return `Window | null`.

```typescript
inspect({
  url,
  // This will throw a type error if targetWindow doesn't support `null` type
  targetWindow: window.open(url)
});
```

packages/xstate-inspect/README.md
mdpratt: Yeah, I like that more as well. Thank you!

packages/xstate-inspect/src/browser.ts
```diff
@@ -89,10 +90,15 @@ const getFinalOptions = (options?: Partial<InspectorOptions>) => {
...
 const patchedInterpreters = new Set<AnyInterpreter>();
 
-export function inspect(options?: InspectorOptions): Inspector | undefined {
-  const { iframe, url, devTools } = getFinalOptions(options);
+export function inspect(
+  options?: InspectorOptions
+): Inspector | undefined {
+  const finalOptions = getFinalOptions(options);
+  const { iframe, url, devTools } = finalOptions;
+  let targetWindow: Window | null | undefined = finalOptions?.targetWindow;
```
Andarist, line 98:
```suggestion
  if ('targetWindow' in finalOptions && !finalOptions.targetWindow) {
    throw new Error('Received a nullable `targetWindow`.')
  }
  let targetWindow: Window | null | undefined = finalOptions.targetWindow;
```

packages/xstate-inspect/src/browser.ts
```diff
@@ -89,10 +90,18 @@ const getFinalOptions = (options?: Partial<InspectorOptions>) => {
...
 const patchedInterpreters = new Set<AnyInterpreter>();
 
-export function inspect(options?: InspectorOptions): Inspector | undefined {
-  const { iframe, url, devTools } = getFinalOptions(options);
+export function inspect(
+  options?: InspectorOptions
+): Inspector | undefined {
+  const finalOptions = getFinalOptions(options);
+  const { iframe, url, devTools } = finalOptions;
+  if ('targetWindow' in finalOptions && !finalOptions.targetWindow) {
```
Andarist, line 98:
```suggestion
  if ('targetWindow' in options && !finalOptions.targetWindow) {
```

mdpratt:
Sorry it took so long to get that fix in. I was away for a few days.

mdpratt:
@Andarist This is ready (hopefully) for a merge now. Let me know if you have any outstanding questions or comments.

mdpratt:
Added changeset

.changeset/sour-keys-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/inspect': minor
+---
+
+Add support for a custom targetWindow
```
Andarist, line 5:
```suggestion
Add support for a custom `targetWindow`
```

PR #3795, opened 2023-01-29
[v5] Input

davidkpiano:


Andarist (commented):
Support for this should also be added to our integration packages. Something like this would probably work OK:
```ts
useMachine(config, { input })
```

PR #3801, opened 2023-01-30
Fixed an issue with not clearing registered interpreters when their machines reached final states

### LINKED ISSUE #3616: Bug: Memory leak appeared between xstate 4.32.1 and xstate 4.33.4
### Description

We've recently had a memory leak in a Home Assistant addon which uses xstate under the hood (as a transitive dependency). It took a while to narrow this down to xstate (or using it), since there are two similar addons, using the exact same library with the exact same version, and only one of those exhibited a quite drastic memory leak (in the order of 100MB/h), while the other one is perfectly stable:
https://github.com/home-assistant/core/issues/77767

It turns out that the difference between those installations was how the dependencies got installed. The non-leaking one uses `yarn install --immutable`, resulting in an install of xstate@4.32.1 from the lockfile - while the leaking one uses `npm install`, which installed the latest version at the time. The leak appeared starting with xstate@4.33.4 and persists with 4.33.6 - unfortunately I don't know which versions between those two are also affected.
After pinning the xstate version in the leaking addon, the leak went away.

Sadly, I cannot provide a reproduction, since only a handful of users are affected and I haven't been able to reproduce myself. I haven't seen anything in the changelog that would suggest a change in behavior - like actor references not getting cleaned up or anything like that.
I would be grateful for some help with tracking this down though, since I'm afraid this is going to prevent us from upgrading xstate indefinitely.

### Expected result

No memory leak

### Actual result

Memory leak

### Reproduction

Not possible unfortunately, see description

### Additional context

_No response_

Andarist:
fixes #3616

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -130,7 +130,8 @@ export class Interpreter<
     ActorRef<
       TEvent,
       State<TContext, TEvent, TStateSchema, TTypestate, TResolvedTypesMeta>
-    > {
+    >
+{
   /**
    * The default interpreter options:
    *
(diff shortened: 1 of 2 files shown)


packages/core/src/interpreter.ts
```diff
@@ -382,6 +382,7 @@ export class Interpreter<
       }
       this._stop();
       this._stopChildren();
+      registry.free(this.sessionId);
```
Andarist, line 385: this is the fix

davidkpiano (approved):
LGTM

PR #3804, opened 2023-02-01
Ability to pass interpreter options and observers to actors created via `createActorContext`

farskid:
Currently, it's not possible to pass interpreter options to the underlying `useInterpret` in `createActorContext`. One of the use cases for passing such options is to add observers to the actor on the creating time or inspect an actor in the runtime using `xstate/inspect`.

### DIFF
--- a/packages/xstate-react/test/createActorContext.test.tsx
+++ b/packages/xstate-react/test/createActorContext.test.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { createMachine, assign } from 'xstate';
 import { fireEvent, screen, render } from '@testing-library/react';
-import { shallowEqual, useSelector, createActorContext } from '../src';
+import { useSelector, createActorContext } from '../src';
 
 const originalConsoleError = console.error;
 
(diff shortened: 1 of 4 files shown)


davidkpiano (commented):
Looks good (cc. @Andarist), don't forget a changeset

davidkpiano:
Oh yeah, @farskid can you add a test or two for this?

farskid:
> Oh yeah, @farskid can you add a test or two for this?

In order to test interpreter options, we'd need to access `actor.options` which is available in the runtime but not allowed by `ActorRefFrom<TMachine>` type. What's the way to go?

davidkpiano (approved):

davidkpiano:
And a changeset, like:

> Interpreter options can now be specified in the second argument of `createActorContext(machine, options)`

packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -122,7 +122,7 @@ describe('createActorContext', () => {
     expect(screen.getByTestId('value').textContent).toBe('b');
   });
 
-  it('should work with useSelector and a custom comparator', async () => {
+  it.skip('should work with useSelector and a custom comparator', async () => {
```
Andarist, line 125: Some accidental changes got merged here. This test shouldn't be skipped - nor the `shallowEqual` should be removed from the `useSelector` call

farskid: I can revert the commit and re-open the PR.

farskid: Agreed to address separately in https://github.com/statelyai/xstate/pull/3808

PR #3805, opened 2023-02-01
[xstate/solid] Change prepublish to prepare

GoldingAustin:
Change `prepublish` to `prepare` in package.json to fix build not running on npm publish.

### DIFF
--- a/.changeset/sharp-weeks-end.md
+++ b/.changeset/sharp-weeks-end.md
@@ -2,4 +2,4 @@
 '@xstate/solid': patch
 ---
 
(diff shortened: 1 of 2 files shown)


.changeset/sharp-weeks-end.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/solid': patch
+---
+
+Change prepublish to prepare in package.json to fix build not running on npm publish for @xstate/solid
```
Andarist, line 5:
```suggestion
Fixed an issue with npm lifecycle scripts that caused only a single set of files to be published.
```

Issue #3811, opened 2023-02-05
Bug: useActorContext type issues


mwarger:
### Description

I have a basic machine that will do a search and report results.  The machine has a service, where at runtime I'm going to supply a TRPC client.

Normally, at runtime I would provide a service:

```
function ClassRevenueBreakdownContainer() {
    const facilityId = useFacilityIdParam();
    const client = useTrpcClient();

    return (
        <RevenueContext.Provider
            machine={() =>
                machine
                    .withContext({
                        start_date: new Date(),
                        end_date: new Date(),
                        facilityId,
                    })
                    .withConfig({
                        services: {
                            searchRevenue: async (context) => {
                                const { start_date, end_date, facilityId } = context;
                                const results =
                                    await client.classes.getClassTypesRevenueBreakdown.query({
                                        facility_id: facilityId,
                                    });
                                return results;
                            },
                        },
                    })
            }
        >
            <div>stuff</div>
        </RevenueContext.Provider>
    );
}
```

However, I end up with two issues.  First is that my default machine config is still requiring this service.

![image](https://user-images.githubusercontent.com/686823/216796892-48e90ca7-25b2-45be-839f-395cf265508e.png)

I won't have this service until I'm within the component, so I don't want to supply it to the initial createActorContext call.

Second, during runtime, I am still seeing a type error for the machine.

![image](https://user-images.githubusercontent.com/686823/216796903-1fe5dd46-b291-4803-9cef-dd4f1e4daf55.png)

I'm working on a repro of this, but I thought I'd ask here in case I'm missing something obvious.  Thanks in advance for anyone that can help me think about this.

### Expected result

I expect to not have to pass the service implementation to the useActorContext creation.  

I also wouldn't expect a type error during the use of `withConfig`.

### Actual result

See reproduction.

### Reproduction

https://github.com/mwarger/trpc-prisma-starter-useActorcontext

### Additional context

xstate - 4.35.4
@xstate/react - 3.1.2

Andarist:
1. The `Provider` should conditionally accept `options` prop - that would allow it to resolve the "final" machine
2. The `createActorContext` shouldn't require all options upfront

PR #3814, opened 2023-02-07
Tweak what options are accepted by `createActorContext` and where

Andarist:
fixes https://github.com/statelyai/xstate/issues/3811


packages/xstate-react/src/createActorContext.tsx
```diff
@@ -1,38 +1,80 @@
...
+} from 'xstate';
+import { UseMachineOptions } from './useMachine';
 
 export function createActorContext<TMachine extends AnyStateMachine>(
   machine: TMachine,
-  ...[options = {}, observerOrListener]: RestParams<TMachine>
+  interpreterOptions?: InterpreterOptions,
+  observerOrListener?:
+    | Observer<StateFrom<TMachine>>
+    | ((value: StateFrom<TMachine>) => void)
```
Andarist, line 22: this is a little bit of a breaking change for this test case:
https://github.com/statelyai/xstate/blob/6f15b2c0cf33c70f97e91ecaba5a250f0b68011a/packages/xstate-react/test/createActorContext.test.tsx#L307-L338

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -1,38 +1,80 @@
...
 
   function Provider({
     children,
-    machine: providedMachine = machine
+    machine: providedMachine = machine,
+    options
   }: {
     children: React.ReactNode;
     machine: TMachine | (() => TMachine);
+    options?: any;
```
Andarist, line 73: I used `any` for simplicity here since the return type of `createActorContext` is explicitly typed and I didn't want to include the typegen-related monster in both places here. I think that we could drop the explicit return type from there, annotate things in the body and let TS to infer things here. This way each signature would only have to live in one place

packages/xstate-react/src/useInterpret.ts
```diff
@@ -85,36 +85,35 @@ export function useIdleInterpreter(
   return service as any;
 }
 
-export type RestParams<
```
Andarist, line 88: I just removed `export` here

packages/xstate-react/src/createActorContext.tsx
```diff
@@ -1,38 +1,80 @@
...
   useActorRef: () => ActorRefFrom<TMachine>;
-  Provider: (props: {
-    children: React.ReactNode;
-    machine?: TMachine | (() => TMachine);
-  }) => React.ReactElement<any, any>;
+  Provider: (
+    props: {
+      children: React.ReactNode;
+      machine?: TMachine | (() => TMachine);
+    } & (AreAllImplementationsAssumedToBeProvided<
```
davidkpiano, line 34: Why can't we have the simpler type here? At least without `AreAllImplementationsAssumedToBeProvided`

davidkpiano (commented):
Spooky types:

CleanShot 2023-02-08 at 21 32 53@2x

davidkpiano (approved):
The types are a little questionable, but LGTM 😅

Andarist:
> Spooky types:

I agree that those types don't have the best display in the IDE. It's not a new problem with our types though. I will keep thinking about how this can be fixed - it is quite hard with those kinds of types.

packages/xstate-react/src/createActorContext.tsx
Andarist: The added type-oriented tests wouldn't pass. It's quite nice that we validate if the user has configured all implementations or not (the cost for that is the readability of those types though 😬)

packages/xstate-react/src/createActorContext.tsx
Andarist: just rechecking - are we OK with this breaking change?

davidkpiano: Yes; options was in the wrong place

PR #3815, opened 2023-02-07
[v5] Input

davidkpiano:
This PR adds `input` to the second argument of the `interpret(...)` function. This passes input data in the `"xstate.init"` event:

```js
const greetMachine = createMachine({
  context: ({ input }) => ({
    greeting: `Hello ${input.name}!`
  })
  // ...
});

const actor = interpret(greetMachine, {
  // Pass input data to the machine
  input: { name: 'David' }
}).start();
```

Addresses STA-3438


packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -163,34 +163,33 @@ describeEachReactMode('useMachine (%s)', ({ suiteKey, render }) => {
...
-  //       foo: 'bar',
-  //       test: true
-  //     });
-
-  //     return null;
-  //   };
-
-  //   render(<Test />);
-  // });
+  it('should merge machine context with options.context', () => {
```
Andarist, line 166:
```suggestion
  it('should accept input and provide it to the context factory', () => {
```

packages/core/src/types.ts
```diff
@@ -107,7 +107,10 @@ export interface ActionMeta<
 // or just make machine a behavior
 export type Spawner = <T extends ActorBehavior<any, any> | string>( // TODO: read string from machine behavior keys
   behavior: T,
-  name?: string
+  options?: Partial<{
+    name: string;
```
Andarist, line 111: shouldn't this be named `id`? it's passed as `id` here: 
https://github.com/statelyai/xstate/blob/686565da4ff9084b2492c59300d6f34db0b9744c/packages/core/src/spawn.ts#L40

packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -163,34 +163,33 @@ describeEachReactMode('useMachine (%s)', ({ suiteKey, render }) => {
...
+      }),
+      initial: 'idle',
+      states: {
+        idle: {}
+      }
+    });
+
+    const Test = () => {
+      const [state] = useMachine(testMachine, {
+        input: { test: true }
```
Andarist, line 180: we should add the same test to all other integration packages

packages/core/src/StateMachine.ts
```diff
@@ -95,16 +95,20 @@ export class StateMachine<
...
-  private _contextFactory: (stuff: { spawn: Spawner }) => TContext;
+  private _contextFactory: (stuff: {
+    spawn: Spawner;
+    input: any; // TODO: fix types
+  }) => TContext;
   // TODO: this getter should be removed
   public get context(): TContext {
-    return this.getContextAndActions()[0];
+    return this.getContextAndActions({} as any)[0];
   }
```
Andarist, line 105: Let's just drop it already :P

packages/core/test/input.test.ts
```diff
@@ -0,0 +1,106 @@
+import { assign, interpret } from '../src';
+import { createMachine } from '../src/Machine';
+
+describe('input', () => {
```
Andarist, line 4: should we also support `input` in other actor types? `fromPromise` & co could also accept this

davidkpiano: Yes

with-heart:
Are you thinking of adding a way to get/set the type of `input` for a machine?

Andarist:
As in with something like `withInput`? I think that's not necessary. A machine is like a class and input is information specific to its instance.

with-heart:
More in the sense of making it part of a machine's public api so you can expect a certain type of input data that's consumed

Andarist:
Ah, ye - the types for this will come in a followup PR

packages/core/test/input.test.ts
davidkpiano: Supported everywhere!

packages/xstate-react/test/useMachine.test.tsx
davidkpiano: https://linear.app/statelyai/issue/STA-3445

packages/core/src/StateMachine.ts
```diff
@@ -94,24 +94,22 @@ export class StateMachine<
...
-  private _contextFactory: (stuff: { spawn: Spawner }) => TContext;
+  private _contextFactory: (stuff: {
+    spawn: Spawner;
+    input: any; // TODO: fix types
+  }) => TContext;
   // TODO: this getter should be removed
-  public get context(): TContext {
-    return this.getContextAndActions()[0];
+  public getContext(input?: any): TContext {
+    return this.getContextAndActions(input)[0];
```
Andarist, line 103: thoughts about removing this getter altogether?

packages/core/src/actions/invoke.ts
```diff
@@ -44,29 +43,23 @@ export function invoke<
...
-              : behaviorImpl;
+          const behavior = behaviorImpl;
 
           const ref = interpret(behavior, {
             id,
             src,
-            parent: actorContext?.self
+            parent: actorContext?.self,
+            input: input
+              ? mapContext(input, state.context, _event as any) // TODO: fix type
```
Andarist, line 61: The `mapContext` is typed incorrectly right now and could have a better name. Previously it was only used to map `Partial<LazyValues<TContext, TContext>>` to `Partial<TContext>` (or smth like that), and here it can map `LazyValues<TInput, TContext>` to `TInput`

packages/core/src/types.ts
```diff
@@ -804,10 +799,7 @@ export interface MachineImplementationsSimplified<
 > {
   guards: Record<string, GuardPredicate<TContext, TEvent>>;
   actions: ActionFunctionMap<TContext, TEvent, TAction>;
-  actors: Record<
-    string,
-    ActorBehaviorCreator<TContext, TEvent> | AnyActorBehavior
-  >;
+  actors: Record<string, AnyActorBehavior>;
```
Andarist, line 802: As we might see in the diff here - now actors won't have access to `TContext` and `TEvent`. So it's not possible to "configure" their dynamic part (input) in the implementations object. That's only possible within the config itself, through the `invoke.input` property. However, at the moment, we are not able to provide typegen-based information there so this might get annoying to people. 

WDYT about allowing this to be:
```ts
// it's not a complete final type, just an illustration for this accepting `{ src, input }` 
actors: Record<string, AnyActorBehavior | { src: AnyActorBehavior, input: InputCreator<TContext, TEvent> }>; 
```

packages/core/src/types.ts
```diff
@@ -1792,7 +1778,10 @@ export interface ActorRef<TEvent extends EventObject, TSnapshot = any>
   // TODO: figure out how to hide this externally as `sendTo(ctx => ctx.actorRef._parent._parent._parent._parent)` shouldn't be allowed
   _parent?: ActorRef<any, any>;
   status: ActorStatus;
-  src?: InvokeSourceDefinition;
+
+  // Meta data about the actor
+  src?: string;
+  input?: any;
```
Andarist, line 1784: I think that `input` should be removed from here. Once we pass it to the actor when computing the initial state it should become "protected". It's up to the actor itself to expose it on its snapshot if needed.

I also don't see why `src` should be part of the public information about the actor's instance.

packages/core/src/actions/invoke.ts
davidkpiano: I'm open to renaming this and fixing the types.

packages/core/test/invoke.test.ts
```diff
@@ -2924,6 +2888,47 @@ describe('invoke', () => {
       .start();
   });
 
+  it.each([(ctx) => ({ endpoint: ctx.url }), { endpoint: (ctx) => ctx.url }])(
```
Andarist, line 2891: imho it would be better to just split this into 2 explicit test cases, separate cases are easier to read and debug if needed

packages/core/test/invoke.test.ts
```diff
@@ -3009,15 +3013,6 @@ describe('invoke', () => {
...
       }
-    ],
-    [
-      'src containing a parametrized invokee with id parameter',
-      {
-        src: {
-          type: 'someSrc',
-          id: 'h4sh'
-        }
-      }
```
Andarist, line 3020: any particular reason why this one was removed? it doesn't look like something that became not applicable with the change 🤔

packages/core/test/predictableExec.test.ts
```diff
@@ -150,10 +150,13 @@ describe('predictableExec', () => {
           entry: raise({ type: 'RAISED' })
         },
         c: {
+          // TODO: fix this
```
Andarist, line 153: fix what? 😉

packages/core/src/types.ts
davidkpiano: The problem is that `input` might be different depending on where you come from:
```js
state1: {
  invoke: {
    src: 'something',
    input: (ctx) => /* get from context */
  }
}

state2: {
  invoke: {
    src: 'something',
    input: (_ctx, ev) => /* get from event */
  }
}
```

I don't understand the typegen information; `invoke.input` is only relevant to the invoked actor, not to the machine (which is the scope of typegen); can you clarify?

packages/core/test/invoke.test.ts
davidkpiano: Sources are always strings now. There is no "invoke source definition object".

packages/core/test/predictableExec.test.ts
davidkpiano: `this`

Andarist: this review is executed in strict mode:
> Cannot read properties of undefined (reading 'answer')

packages/core/test/invoke.test.ts
Andarist: ah, ok - makes sense now, thanks for clarifying!

packages/core/src/StateMachine.ts
Andarist: let's do this in a follow up work as this is just a symptom of a bigger story

packages/core/src/actions/invoke.ts
Andarist: We probably won't find a good specific name, so let's just use a vague one:
```ts
type ResolvableObject<
  TObject extends {},
  TEvent extends EventObject
> = TObject | ((context: TContext, event: TEvent) => TObject);

type ResolvableProperties<
  TObject extends {},
  TEvent extends EventObject
> = {
  [K in keyof TObject]:
    | TObject[K]
    | ((context: TContext, event: TEvent) => TObject[K]);
};

declare function resolveObject<
  TObject extends {},
  TEvent extends EventObject
>(
  obj:
    | ResolvableObject<TObject, TContext, TEvent>
    | ResolvableProperties<TObject, TContext, TEvent>,
  context: TContext,
  event: TEvent
): TObject;
```

This is very close to the existing definition. It's just that:
1. `resolveObject` accept more generic params - as the argument used for resolving (`TContext`) isn't necessarily the same as the returned value (`TObject`)
2. `ResolvableObject`/`ResolvableProperties` don't contain the optional modifier within the mapped type. The optionality should be provided by the "caller" (`assign` should provide that optionality as it handles "slices" of `TContext`). It shouldn't be assumed by that inner mapped type. After all, when we need to resolve the input - we need all of it, not its subset.

This was just a quick experiment that turned out to be very, very similar to the existing functions (just with different names and very minor tweaks). Feel free to adjust this to your liking, however, you see fit.

packages/core/src/types.ts
Andarist: The conceptual `TInput` always comes from a given actor type and is the same regardless of where and how it is instantiated. What might be different between instantiation sites, are the "arguments" used to resolve that same type. But that's exactly the same as with any other implementation type:
```ts
state1: {
  entry: 'something'
}

state2: {
  entry: 'something'
}
```
The final implementation of `something` just has to deal with all possible argument types. It's free to ignore `context` completely, and it's free to ignore `event` - but when it starts using the `event`, it just has to assume that it's a union of all possible event types that can lead to triggering this specific action.

davidkpiano: What should be the resolution for this?

packages/core/src/types.ts
```diff
@@ -391,20 +395,15 @@ export interface InvokeDefinition<
...
   autoForward?: boolean;
-  /**
-   * Data from the parent machine's context to set as the (partial or full) context
-   * for the invoked child machine.
-   *
-   * Data should be mapped to match the child machine's context shape.
-   */
-  data?: Mapper<TContext, TEvent, any> | PropertyMapper<TContext, TEvent, any>;
+
+  input?: Mapper<TContext, TEvent, any>;
```
Andarist, line 406:
```suggestion
  input?: Mapper<TContext, TEvent, any> | any;
```

packages/core/src/types.ts
Andarist: I propose this: https://github.com/statelyai/xstate/pull/3897

.changeset/heavy-jeans-arrive.md
```diff
@@ -0,0 +1,19 @@
...
+---
+
+The `interpret(...)` function now accepts `input` in the second argument, which passes input data in the `"xstate.init"` event:
+
+```js
+const greetMachine = createMachine({
+  context: ({ input }) => ({
+    greeting: `Hello ${input.name}!`
+  })
+  // ...
```
Andarist, line 12:
```suggestion
  context: ({ input }) => ({
    greeting: `Hello ${input.name}!`
  }),
  entry: (_, event) => {
    event.type // 'xstate.init'
    event.input // { name: 'David' }
  }
  // ...
```

Issue #3151, opened 2022-03-14
Bug: Typescript issue `assign` with function and static value


bartenra:
### Description

Say we have this context signature:

```ts
interface ToggleContext {
  count: number;
  skip: boolean;
}
```
Then this works

```ts
entry: assign({
  count: (context) => context.count + 1,
  skip: (context) => true
}),
```

but this doesn't

```ts
entry: assign({
  count: (context) => context.count + 1,
  skip: true
}),
```

### Expected result

I expect it to work like the docs: https://xstate.js.org/docs/guides/context.html#assign-action

Without Typescript issues.

### Actual result

`count` know has type `unknown`. So Typescript complains.

### Reproduction

https://codesandbox.io/s/zealous-hopper-gncn3e

Andarist:
I plan to make some significant changes to how built-in action factories are currently typed and **maybe** this will help this case. The main goal of the planned refactor is to improve inference of the those action factories nested within `createMachine` call but I'm not yet sure if supporting "mixed" variant for `assign` will be possible.

johtso:
I stumbled upon what I think is the same issue here: https://github.com/statelyai/xstate/issues/3267

It doesn't even seem to be an issue with mixing static and dynamic assignment values.. it seems to be an issue with assigning static values full stop.

https://stately.ai/viz/da4f046e-b98b-467b-9ff0-18aaf3a8274d

What is the workaround here?

Andarist:
It's likely that this would get fixed if we'd wrap `TContext` occurrences on [this line](https://github.com/statelyai/xstate/blob/3bb4a00e7e4eb66a3fabfc20e811bbadfa8eb7a4/packages/core/src/actions.ts#L467) with `LowInfer` (or `NoInfer` but that would prevent inference here altogether and we don't exactly want that)

PR #3818, opened 2023-02-07
Fixed inference for `assign` using `PropertyAssigner`

Andarist:
fixes https://github.com/statelyai/xstate/issues/3151

### DIFF
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -41,7 +41,8 @@ import {
   EventFrom,
   AnyActorRef,
   PredictableActionArgumentsExec,
-  BaseActionObject
+  BaseActionObject,
+  LowInfer
 } from './types';
 import * as actionTypes from './actionTypes';
 import {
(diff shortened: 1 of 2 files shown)


packages/core/src/actions.ts
```diff
@@ -467,7 +468,9 @@ export function resolveStop<TContext, TEvent extends EventObject>(
  * @param assignment An object that represents the partial context to update.
  */
 export const assign = <TContext, TEvent extends EventObject = EventObject>(
-  assignment: Assigner<TContext, TEvent> | PropertyAssigner<TContext, TEvent>
+  assignment:
+    | Assigner<LowInfer<TContext>, TEvent>
+    | PropertyAssigner<LowInfer<TContext>, TEvent>
```
Andarist, line 473: This has some drawbacks when it comes to inferring standalone `assign`s, like here:
```ts
const a1 = assign(/*...*/)
```

When `Assigner` is used then it works just fine:
```ts
const a1 = assign((ctx: { counter: number }) => ({})) // TContext inferred to `{ counter: number }`
```
but when it comes to `PropertyAssigner` then it doesn't infer `TContext` and it leaves it as `unknown` (the internal partial assigned function stays types correctly, it's about what the `typeof a1` is):
```ts
const a1 = assign({
  counter: (context: { counter: number }) => context.counter + 1,
})
```
It also doesn't work, in this context, with static values:
```ts
const a1 = assign({
  counter: 100,
})
```

I think that `LowInfer` combined with reverse mapped types is broken in TS, a minimal repro case of this can be found here: [TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAMg9gdwJIDsBmEBOAeAKgPigF4pcoAyKAbwF8BuAKAdEigCUIBbAQzD0JJUGUKAG0A0lACWKKAGsIIOGlIBdAFykJqxvSYATCAGMANt0zQjcFAGdgUIxe7AIAeQBGAK038AFHC9NDh4+eGR0LH58AEpiQlxGBitbewsbYgcnFw9PXyoHOABXFBdMTQAGABooblsELE0AIjQ4OEaoGmjGAHpukREAPQB+BiA)

davidkpiano (approved):

PR #3819, opened 2023-02-08
[core] Add self to action meta

davidkpiano:
This PR adds the `self` interpreter reference to action meta arguments:

```js
  entry: (ctx, ev, { self }) => {
    // can pass self around
  }
});
```

### DIFF
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -12,7 +12,8 @@ import {
   Typestate,
   ActorRef,
   StateMachine,
-  SimpleEventsOf
+  SimpleEventsOf,
+  AnyInterpreter
 } from './types';
 import { EMPTY_ACTIVITY_MAP } from './constants';
 import { matchesState, isString, warn } from './utils';
@@ -65,7 +66,8 @@ export const isState = isStateConfig;
 
 export function bindActionToState<TC, TE extends EventObject>(
   action: ActionObject<TC, TE>,
-  state: State<TC, TE, any, any, any>
+  state: State<TC, TE, any, any, any>,
+  interpreter: AnyInterpreter
 ): ActionObject<TC, TE> {
   const { exec } = action;
   const boundAction: ActionObject<TC, TE> = {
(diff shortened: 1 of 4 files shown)


Andarist (commented):
You should also:
- include it in the [`InvokeMeta` type](https://github.com/statelyai/xstate/blob/a7d79c678049d47964fb5017fadd989da086e7df/packages/core/src/types.ts#L325)
- call the `serviceCreator` with it [here](https://github.com/statelyai/xstate/blob/a7d79c678049d47964fb5017fadd989da086e7df/packages/core/src/interpreter.ts#L1094)
- include it in the [`AssignMeta` type](https://github.com/statelyai/xstate/blob/a7d79c678049d47964fb5017fadd989da086e7df/packages/core/src/types.ts#L109)
- add it to this `assign`'s `meta` [here](https://github.com/statelyai/xstate/blob/a7d79c678049d47964fb5017fadd989da086e7df/packages/core/src/utils.ts#L432) 
- write tests for both :P

Those are the most important ones - but even with those changes (where adding that to the `assign`'s meta will already require some changes to the code) it still won't be supported by all builtin action types: `send`, `raise`, `log`, `choose`, `pure`, `stop`

davidkpiano:
Actually, closing this for a few reasons:

- Adding `self` needs to be done in multiple areas (as listed above) and this currently touches a _lot_ of the API, including how initial state is calculated, which has ripple effects in the libraries that consume XState
- There's still the question of if `self` should exist or not when using `createMachine()` by itself (not interpreted)
- This is not the exact API for v5, so I want to avoid confusion
- This is mostly to support a userland "receptionist pattern", but this is possible today without any changes: https://stackblitz.com/edit/js-vcmeg3?file=index.js

br-matt:
> * This is mostly to support a userland "receptionist pattern", but this is possible today without any changes: https://stackblitz.com/edit/js-vcmeg3?file=index.js

Thank you, its extremely hard to find recommended examples for this.. a lot suggest raise but does not work.

PR #3822, opened 2023-02-09
Scope `svelte-check` properly and make it use the correct `tsconfig.json`

Andarist:
While investigating something else I noticed that this might be configured slightly incorrectly


davidkpiano (approved):
Not sure what this does but your call ✅

Andarist:
> Not sure what this does but your call ✅

It's typechecking our `.svelte` files (the ones used in tests). `tsc` itself is not capable of that

PR #3835, opened 2023-02-15
add `TagsFrom` helper type

with-heart:
Hi yall! I saw a helper type request on Discord today that seemed easy enough to implement: `TagsFrom`

`TagsFrom` allows us to access the typegen `tags` type of a `Machine` ~~`State`, or `Interpreter`~~:

```ts
import { TagsFrom } from "xstate";

  tsTypes: {} as {
    "@@xstate/typegen": true;
    tags: "a" | "b" | "c";
  }
});

type Tags = TagsFrom<typeof machine>; // 'a' | 'b' | 'c'
```

If typegen isn't enabled, `TagsFrom` returns `string` (which matches `hasTag`):

```ts
import { TagsFrom } from "xstate";

type Tags = TagsFrom<typeof machine>; // string
```

### DIFF
--- a/.changeset/mean-oranges-dress.md
+++ b/.changeset/mean-oranges-dress.md
@@ -6,12 +6,12 @@ The new `TagsFrom` helper type extracts the type of `tags` from a state machine 
 
 ```ts
 const machine = createMachine({
+  // `tags` attached to machine via typegen
+  tsTypes: {} as import('./machine.typegen').Typegen0,
   tags: ['a', 'b'],
   states: {
-    c: { tags: 'c' }
+    idle: { tags: 'c' }
   },
-  // `tags` attached to machine via typegen
-  tsTypes: {} as import('./machine.typegen').Typegen0
 });
 
 type Tags = TagsFrom<typeof machine>; // 'a' | 'b' | 'c'
@@ -23,7 +23,7 @@ If typegen is not enabled, `TagsFrom` returns `string`:
 const machine = createMachine({
   tags: ['a', 'b'],
   states: {
-    c: { tags: 'c' }
+    idle: { tags: 'c' }
   }
 });
 
--- a/.changeset/mean-oranges-dress.md
+++ b/.changeset/mean-oranges-dress.md
@@ -2,7 +2,7 @@
 'xstate': minor
 ---
 
-The new `TagsFrom` helper type extracts the type of `tags` from any type which uses tags (when typegen is enabled):
+The new `TagsFrom` helper type extracts the type of `tags` from a state machine when typegen is enabled:
 
 ```ts
 const machine = createMachine({
--- a/.changeset/mean-oranges-dress.md
+++ b/.changeset/mean-oranges-dress.md
@@ -1,0 +1,31 @@
+---
+'xstate': minor
+---
+
+The new `TagsFrom` helper type extracts the type of `tags` from any type which uses tags (when typegen is enabled):
+
+```ts
+const machine = createMachine({
+  tags: ['a', 'b'],
+  states: {
+    c: { tags: 'c' }
+  },
+  // `tags` attached to machine via typegen
+  tsTypes: {} as import('./machine.typegen').Typegen0
+});
+
+type Tags = TagsFrom<typeof machine>; // 'a' | 'b' | 'c'
+```
+
+If typegen is not enabled, `TagsFrom` returns `string`:
+
+```ts
+const machine = createMachine({
+  tags: ['a', 'b'],
+  states: {
+    c: { tags: 'c' }
+  }
+});
+
+type Tags = TagsFrom<typeof machine>; // string
+```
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -1984,18 +1985,53 @@ type Matches<TypegenEnabledArg, TypegenDisabledArg> = {
...
   action: ActionObject<unknown, EventObject>,
   context: unknown,
   _event: SCXML.Event<EventObject>
 ) => void;
+
+type _TagsFrom<TResolvedTypesMeta> = TResolvedTypesMeta extends TypegenEnabled
+  ? Prop<Prop<TResolvedTypesMeta, 'resolved'>, 'tags'>
+  : string;
+
+export type TagsFrom<T> = ReturnTypeOrValue<T> extends infer R
```
Andarist, line 2009: I would follow the structure of `StateValueFrom` here. Your version might be a little bit easier to use at times - as it handles more input types but it feels like grabbing this stuff from the machine type is the most common use case. At the very least - both of those types should accept the same inputs and for now, I would just downsize the scope of this new helper to the existing scope of `StateValueFrom`.

On top of that, I would infer this from the `hasTag` argument and not from the `TResolvedTypesMeta`.

with-heart: That makes perfect sense! I'll get that updated.

with-heart:
How do you feel about the tests for this @Andarist?

packages/core/src/types.ts
```diff
@@ -1,4 +1,5 @@
 import { Clock, Interpreter } from './interpreter';
+import { createMachine } from './Machine';
```
Andarist, line 2:
```suggestion
```

packages/core/test/types.test.ts
```diff
@@ -955,3 +957,38 @@ describe('actions', () => {
     });
   });
 });
+
+describe('tags', () => {
```
Andarist, line 961: I would move those to [`typeHelpers.test.ts`](https://github.com/statelyai/xstate/blob/4e6b54ad50770afa125d3cc51fae66b63c29c31a/packages/core/test/typeHelpers.test.ts#L346) since this file has the most tests related to those helpers

with-heart:
Ugh, we should achieve formatting parity with `xstate-tools` bc I keep forgetting that imports aren't managed automatically like they are over there 😭

Andarist:
> Ugh, we should achieve formatting parity with xstate-tools bc I keep forgetting that imports aren't managed automatically like they are over there 😭

yeah, the situation is unfortunate - but we don't want to enable things like this at the moment here to avoid unnecessary merge conflicts when merging `main` to `next`

packages/core/test/typeHelpers.test.ts
```diff
@@ -428,3 +429,43 @@ describe('EmittedFrom', () => {
...
+    acceptTag('d');
+  });
+
+  it('derives string from StateMachine without typegen', () => {
+    const machine = createMachine({
+      tsTypes: ({} as unknown) as {
+        '@@xstate/typegen': false;
+        tags: 'a' | 'b' | 'c';
+      }
+    });
```
Andarist, line 459:
```suggestion
    const machine = createMachine({});
```

packages/core/test/typeHelpers.test.ts
```diff
@@ -428,3 +429,43 @@ describe('EmittedFrom', () => {
...
 });
+
+describe('tags', () => {
+  it('derives tags from StateMachine when typegen is enabled', () => {
+    const machine = createMachine({
+      tsTypes: ({} as unknown) as {
+        '@@xstate/typegen': true;
+        tags: 'a' | 'b' | 'c';
+      }
+    });
```
Andarist, line 440:
```suggestion
    interface TypesMeta extends TypegenMeta {
      tags: 'a' | 'b' | 'c';
    }
    const machine = createMachine({
      tsTypes: {} as TypesMeta
    });
```

.changeset/mean-oranges-dress.md
```diff
@@ -0,0 +1,31 @@
...
+The new `TagsFrom` helper type extracts the type of `tags` from a state machine when typegen is enabled:
+
+```ts
+const machine = createMachine({
+  tags: ['a', 'b'],
+  states: {
+    c: { tags: 'c' }
+  },
+  // `tags` attached to machine via typegen
+  tsTypes: {} as import('./machine.typegen').Typegen0
```
Andarist, line 14:
```suggestion
  // `tags` attached to machine via typegen
  tsTypes: {} as import('./machine.typegen').Typegen0,
  tags: ['a', 'b'],
  states: {
    idle: { tags: 'c' }
  },
```

.changeset/mean-oranges-dress.md
```diff
@@ -0,0 +1,31 @@
...
+type Tags = TagsFrom<typeof machine>; // 'a' | 'b' | 'c'
+```
+
+If typegen is not enabled, `TagsFrom` returns `string`:
+
+```ts
+const machine = createMachine({
+  tags: ['a', 'b'],
+  states: {
+    c: { tags: 'c' }
```
Andarist, line 26:
```suggestion
    idle: { tags: 'c' }
```

with-heart:
@Andarist makes perfect sense. nice changes, appreciate it!

PR #3837, opened 2023-02-15
[v5] System (receptionist)

davidkpiano:
This PR creates an implicit system and provides it in the machine implementations:

```ts
  id: 'parent',
  initial: 'a',
  states: {
    a: {
      invoke: [
        {
          src: fromCallback((_, receive) => {
            receive((event) => {
              expect(event.type).toBe('HELLO');
            });
          }),
          key: 'receiver'
        },
        {
          src: createMachine({
            id: 'childmachine',
            entry: (_ctx, _ev, { system }) => {
              const receiver = system.get('receiver');

              if (receiver) {
                receiver.send({ type: 'HELLO' });
              }
            }
          })
        }
      ]
    }
  }
});

```


packages/core/src/actors/callback.ts
```diff
@@ -29,7 +29,7 @@ export function fromCallback<TEvent extends EventObject>(
 
       return state;
     },
-    transition: (state, event, { self, id }) => {
+    transition: function callbackTransition(state, event, { self, id }) {
```
davidkpiano, line 32: I just did this as a test because it was an anonymous function in debugging; @Andarist any better ways to make this non-anonymous? Use a class?

tivac:
(Sorry if this is an obvious question, can't look through the changes ATM atm)

Is the system accessible from the interpreter for outside use as well?

with-heart:
Love this so much. It's a testament to the system yall have built that such a powerful feature requires so little to code implement. Super excited for this!

davidkpiano:
> (Sorry if this is an obvious question, can't look through the changes atm)
> 
> Is the system accessible from the interpreter for outside use as well?

It is "not" (at least not naturally)... but of course, once you have access to the `{ system }` grabbed from an action, we can't stop you from passing that reference around 🤷

davidkpiano:
> Is there a previous discussion on why that isn't allowed?
> 
> With how we use child machines being able to send events to specific ones instead of blasting everybody or manually plumbing through the parent seems really valuable.

Maybe I'm misreading the question, but from any descendant machine within the same system, you have access to the `system`, so yes, you can send events to specific actors (`'blah'` in this case) instead of "blasting".

CleanShot 2023-02-20 at 14 25 14@2x

tivac:
@davidkpiano I'm talking about something like this:

```js
import { createMachine, interpret } from 'xstate';

const rootMachine = createMachine({
  /* ... */
});

const service = interpret(rootMachine);

service.start();

const receiver = service.system.get("receiver");

if(receiver) {
  receiver.send({ type: 'HELLO' });
}
```

Since we're using [`xstate-component-tree`](https://github.com/tivac/xstate-component-tree) to manage components that are being rendered, and sometimes components want to fire an event back into the statechart that spawned them. Currently `xstate-component-tree` offers a [`.broadcast(<event>)`](https://github.com/tivac/xstate-component-tree#broadcasteventname--eventobject-payload) API that literally sends the event into **every running statechart** but that's a little coarse and dangerous for my tastes.

I guess since `xstate-component-tree` is actively tracking the running children at any given time it could provide a service-style API, maybe, but I was hoping that this would let me rip all that logic out of my code and use something built-in and more standard.

davidkpiano:
@tivac Ah okay, I understand; you want to access the service completely outside the machine.

I agree that this should be made possible by the API you've described:

```ts
const receiver = service.system.get('receiver');
```

I'll add tests for this.

packages/core/src/actors/callback.ts
Andarist: it should have an inferred name here:
```ts
var obj = {
  someFn: () => { throw new Error('test') }
}
obj.someFn()
// 17:12:58.428 VM328:2 Uncaught Error: test
//    at Object.someFn (<anonymous>:2:25)
//    at <anonymous>:4:5
```
which of course wouldn't exactly be entirely helpful here - the inferred name would be `transition` and not something like `callbackTransition`.

I don't think classes would help to solve this particular problem, the method name would still be pretty generic (`transition`). I think what you did here is OK - we had an anonymous function, and now it's named. Although I would probably try to stick to arrows... since arrows might play better with minifiers.

We have a lot of anonymous arrows like this in the codebase and this has been raised as an issue for the first time, so I wonder how big of a problem this truly is. 🤷‍♂️ 

Naming an arrow function would require hoisting it and assigning to a local variable.

packages/core/src/actions/invoke.ts
```diff
@@ -67,7 +67,7 @@ export function invoke<
             type,
             params: {
               ...invokeDef,
-              ref: interpret(behavior, { id })
+              ref: interpret(behavior, { id, parent: actorContext?.self })
```
Andarist, line 70: q: why this is needed now and why it wasn't needed before? I think this is the correct change - but I wonder what changed that prompted it now

packages/core/src/interpreter.ts
```diff
@@ -118,6 +119,7 @@ export class Interpreter<
   // TODO: remove
   public _forwardTo: Set<AnyActorRef> = new Set();
 
+  public system: ActorSystem<any>;
```
Andarist, line 122: should the system be accessible publicly from just any ref belonging to it?

packages/core/src/interpreter.ts
```diff
@@ -135,8 +137,9 @@ export class Interpreter<
     const { clock, logger, parent, id } = resolvedOptions;
     const self = this;
 
-    // TODO: this should come from a "system"
-    this.sessionId = registry.bookId();
+    this.system = parent?.system ?? options?.system ?? createSystem();
```
Andarist, line 140: why do we allow `system` to be provided through options? Shouldn't it always be given by the parent or instantiated here?

packages/core/src/system.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  return {
+    register: (actorRef) => {
+      const id = `x:${sessionIdIndex++}`;
+      children.set(id, actorRef);
+      return id;
+    },
+    unregister: (actorRef) => {
+      children.delete(actorRef.id);
+    },
```
Andarist, line 42: If I'm not mistaken - those methods are leaked publicly but they should only be accessible internally.

packages/core/src/system.ts
```diff
@@ -0,0 +1,50 @@
...
+    get: (key) => {
+      return keyedActors.get(key) as T['actors'][any];
+    },
+    set: (key, actorRef) => {
+      keyedActors.set(key, actorRef);
+    }
```
Andarist, line 48: what should happen when an actor with a given key is already registered?

packages/core/src/types.ts
```diff
@@ -1854,23 +1863,29 @@ export type __ResolvedTypesMetaFrom<T> = T extends StateMachine<
 export type EventOfMachine<TMachine extends AnyStateMachine> =
   TMachine extends StateMachine<any, infer E, any, any, any> ? E : never;
 
-export interface ActorContext<TEvent extends EventObject, TSnapshot> {
+export interface ActorContext<
+  TEvent extends EventObject,
+  TSnapshot,
+  TSystem extends ActorSystem<any> = any
```
Andarist, line 1869: That's a bad habit 😉  we never should provide `any` as a default type param
```suggestion
  TSystem extends ActorSystem<any>
```

packages/core/src/system.ts
```diff
@@ -0,0 +1,50 @@
+import { ActorRef, ActorSystem, ActorSystemInfo } from './types.js';
+
+const children = new Map<string, ActorRef<any>>();
+
+let sessionIdIndex = 0;
+
+export interface Registry {
```
Andarist, line 7: What is Registry used for now? shouldn't we just remove this concept in favor of the system?

packages/core/src/interpreter.ts
davidkpiano: According to early feedback, yes.

packages/core/src/interpreter.ts
davidkpiano: Yes it should - it was originally provided through options for this use-case:

```ts
const system = createSystem(...);

const actor = interpret(machine, { system }).start();
```

But I don't think that `createSystem(...)` is in scope for this PR anymore.

packages/core/src/system.ts
davidkpiano: It overwrites the previous actor

packages/core/src/system.ts
davidkpiano: Yes, this can now be removed

packages/core/src/system.ts
Andarist: It feels to me that this might be an easy mistake to make and one that might be quite hard to debug if somebody accidentally overwrites the existing registered actor. 

On top of that - there is now no way to **clear** a registered actor. This can lead to memory leaks

davidkpiano: > On top of that - there is now no way to **clear** a registered actor. This can lead to memory leaks

Registered actors are unregistered when they are stopped. If an actor is not stopped, it won't be unregistered, but the actor not being stopped is a memory leak anyway.

packages/core/src/system.ts
davidkpiano: What do you recommend to do?

packages/core/src/actions/invoke.ts
davidkpiano: Before, this was directly added to `actorRef.parent = this` or something like that.

packages/core/src/types.ts
```diff
@@ -388,6 +388,8 @@ export interface InvokeDefinition<
   TEvent extends EventObject
 > {
   id: string;
```
tivac, line 390:
```suggestion
  id: string | undefined;
```

If I'm reading [the code in this PR](https://github.com/statelyai/xstate/pull/3837/files#diff-fc91146eff717b2bcd6adbc0955ca0549346ad7bef9aa2278e0a75a9793a66b9R147-R148) right looks like `id` is optional now? Looks like if it's not defined it'll be set to the returned ID from registering the actor with the system.

davidkpiano: The `id` is "local" to the actor, so it can reference its child by `id`. The `key` makes it discoverable throughout the system.

The `id` is still necessary and autogenerated if not specified, and different from the `key`.

CleanShot 2023-03-04 at 14 10 01@2x

tivac: Right, I get that. It just seems like passing an `id` is optional, since it'll be based off the `sessionId` returned from registering with the system if not specified. I was wondering about it because the example in the changeset didn't have an `id` and I thought that was required for all invoked actors.

davidkpiano: Ah okay I get what you're saying. The `id` is indeed optional; this is the `InvokeDefinition` type used internally with the resolved ID (either the given one or a generated one), which is always provided

tivac: Hah, oops. What I get for trying to find type definitions in the diff view 😅

packages/core/src/system.ts
davidkpiano: deb302ae9210614a0ee89e61c79daec0f7e55e90

packages/core/src/system.ts
```diff
@@ -0,0 +1,34 @@
...
+  const reverseKeyedActors = new WeakMap<AnyActorRef, keyof T['actors']>();
+
+  const system: ActorSystem<T> = {
+    _register: (actorRef) => {
+      const id = `x:${sessionIdIndex++}`;
+      children.set(id, actorRef);
+      return id;
+    },
+    _unregister: (actorRef) => {
+      children.delete(actorRef.id);
```
Andarist, line 16: I'm almost certain that this is a bug/mem leak. It should be:
```suggestion
      children.delete(actorRef.sessionId);
```

packages/core/src/system.ts
```diff
@@ -0,0 +1,34 @@
...
+      const key = reverseKeyedActors.get(actorRef);
+
+      if (key !== undefined) {
+        keyedActors.delete(key);
+        reverseKeyedActors.delete(actorRef);
```
Andarist, line 21: this one could be called unconditionally, otoh - it kinda doesn't matter cause you are using `WeakMap`, in a way... there is no point in this `.delete` call at all, the engine itself should just clean this up - that's the whole point of the Weak part in the WeakMap

packages/core/src/types.ts
```diff
@@ -1608,6 +1613,10 @@ export interface InterpreterOptions<_TActorBehavior extends AnyActorBehavior> {
 
   sync?: boolean;
 
+  /**
+   * The receptionist key to register this actor under
+   */
+  key?: string;
```
Andarist, line 1619: As we discussed today - it might be worth renaming this to `systemId`

packages/core/src/types.ts
```diff
@@ -1608,6 +1613,10 @@ export interface InterpreterOptions<_TActorBehavior extends AnyActorBehavior> {
 
   sync?: boolean;
 
+  /**
+   * The receptionist key to register this actor under
```
Andarist, line 1617: This feature has been added as a "system" but we refer here to a "receptionist". The latter term doesn't appear anywhere in the API. Could we stick to a single name for this?

packages/core/src/interpreter.ts
```diff
@@ -133,11 +136,16 @@ export class Interpreter<
...
-    const { clock, logger, parent, id } = resolvedOptions;
+    const { clock, logger, parent, id, key } = resolvedOptions;
     const self = this;
 
-    // TODO: this should come from a "system"
-    this.sessionId = registry.bookId();
+    this.system = parent?.system ?? createSystem();
+
+    if (key) {
+      this.system.set(key, this);
```
Andarist, line 145: This should definitely be called in the `start` method. Without it `@xstate/react` might break when it performs its Fast Refresh.

I'm not entirely sure if we should remove this call from here though. What use cases would break without this call in the constructor?

packages/core/src/system.ts
```diff
@@ -0,0 +1,34 @@
...
+      }
+    },
+    get: (key) => {
+      return keyedActors.get(key) as T['actors'][any];
+    },
+    set: (key, actorRef) => {
+      keyedActors.set(key, actorRef);
+      reverseKeyedActors.set(actorRef, key);
+    }
```
Andarist, line 30: What do you think about throwing if the `key` is already set?

davidkpiano: We should allow overriding the actor

davidkpiano: > Why though? It seems like an easy mistake to make. What is practical use case for a hard overwrite like this?

Because, for example, if a _spawned_ actor is assigned a `systemId` and we want to instead spawn a different actor for that `systemId` (replacing it), then it is not easy to replace it.

We can disallow it for now though.

Andarist: But what would be a practical use case for such replacing?

davidkpiano: > But what would be a practical use case for such replacing?

Not sure; I can disable this, we can always add it back later if we find a valid use-case.

davidkpiano: ba996b642612b51df2e7e45723da0cfb57e9f81d

packages/core/src/types.ts
```diff
@@ -1846,30 +1856,38 @@ export type __ResolvedTypesMetaFrom<T> = T extends StateMachine<
 export type EventOfMachine<TMachine extends AnyStateMachine> =
   TMachine extends StateMachine<any, infer E, any, any, any> ? E : never;
 
-export interface ActorContext<TEvent extends EventObject, TSnapshot> {
+export interface ActorContext<
+  TEvent extends EventObject,
+  TSnapshot,
+  TSystem extends ActorSystem<any> = ActorSystem<any>
```
Andarist, line 1862: I wonder if this shouldn't accept the same generic as `ActorSystem` itself accepts. Not a big deal though and I don't have a strong opinion about this right now.

packages/core/src/types.ts
```diff
@@ -1608,6 +1613,10 @@ export interface InterpreterOptions<_TActorBehavior extends AnyActorBehavior> {
 
   sync?: boolean;
 
+  /**
+   * The system ID to register this actor under
+   */
```
Andarist, line 1618: #yoda

packages/core/src/types.ts
```diff
@@ -1982,6 +2000,17 @@ export type StateValueFrom<TMachine extends AnyStateMachine> = Parameters<
...
+  actors: Record<string, AnyActorRef>;
+}
+
+export interface ActorSystem<T extends ActorSystemInfo> {
+  _bookId: () => string;
+  _register: (sessionId: string, actorRef: AnyActorRef) => string;
+  _unregister: (actorRef: AnyActorRef) => void;
+  _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
+  get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+}
```
Andarist, line 2013: Don't worry about this right now but I find defining a type like this here extremely annoying to work with. It's not exactly an "abstract" type that comes with no direct implementation. It has a single implementation in the `system.ts` and whenever I cmd+click on `ActorSystem` I'm being brought to here by the IDE which isn't that helpful (going to the implementation in the `system.ts` would be way more useful in this instance)

davidkpiano: Want a class instead?

packages/core/src/types.ts
davidkpiano: Can change it later; if for whatever reason a dev needs to manually construct this, this type can be `typeof someSystem` rather than `ActorSystemInfoFrom<typeof someSystem>`

packages/core/src/types.ts
Andarist: the current runtime code is fine - but the type definition could be derived from the implementation or just declared next to the implementation then jumping to it would at least bring us to the correct file

Andarist (approved):

packages/core/src/types.ts
davidkpiano: > the current runtime code is fine - but the type definition could be derived from the implementation or just declared next to the implementation then jumping to it would at least bring us to the correct file

I'm open to having this be refactored, if you want to (not immediately sure how to do this)

PR #3844, opened 2023-02-19
xstate-rxjs

aldovelco:
I finally got some time to put a PR together for xstate-rxjs. I wanted to get feedback before starting to work on the rxjs template so I have only implemented `fromMachine`, `fromInterpret` and `fromSelector` for now.

### DIFF
--- a/packages/xstate-rxjs/src/fromInterpret.ts
+++ b/packages/xstate-rxjs/src/fromInterpret.ts
@@ -1,0 +1,114 @@
+import { from, Observable } from 'rxjs';
+import { distinctUntilChanged, shareReplay } from 'rxjs/operators';
+import {
+  AnyStateMachine,
+  AreAllImplementationsAssumedToBeProvided,
+  InternalMachineOptions,
+  interpret,
+  InterpreterFrom,
+  InterpreterOptions,
+  InterpreterStatus,
+  State,
+  StateFrom
+} from 'xstate';
+import { FromMachineOptions, MaybeLazy, Prop } from './types';
+
+type RestParams<
+  TMachine extends AnyStateMachine
+> = AreAllImplementationsAssumedToBeProvided<
+  TMachine['__TResolvedTypesMeta']
+> extends false
+  ? [
+      options: InterpreterOptions &
+        FromMachineOptions<TMachine['__TContext'], TMachine['__TEvent']> &
+        InternalMachineOptions<
+          TMachine['__TContext'],
+          TMachine['__TEvent'],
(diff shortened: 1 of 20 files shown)


packages/xstate-rxjs/src/fromInterpret.ts
```diff
@@ -0,0 +1,114 @@
...
+    ];
+
+type FromInterpretReturn<
+  TMachine extends AnyStateMachine,
+  TInterpreter = InterpreterFrom<TMachine>
+> = {
+  state$: Observable<StateFrom<TMachine>>;
+  send: Prop<TInterpreter, 'send'>;
+  service: TInterpreter;
+};
```
Andarist, line 48: is it a popular pattern to return a stream and some other things?

aldovelco: I haven't seen that in other rxjs operators. They generally return the stream directly like the internal [`fromService` function](https://github.com/statelyai/xstate/pull/3844/files/01443aaeaaf5ee7aa8aaf666a261f65c4f031bfb#diff-1a2a8ac3aeb18b3c2697e2107d2fb478682bb71d011fbf4c5f743815ac22e863R91) from that file.

Couldn't think think of a better name and if we just exposed the stream directly we would either lose the reference to the service or we wouldn't be able to call `.start()` on it automatically.

Maybe we don't really need to do that and it's just fine to expose a `fromInterpret` that returns the stream?
e.g.
```ts
import { interpret, createMachine } from 'xstate';
import { fromInterpret } from '@xstate/rxjs';

const service = interpret(createMachine(...)).start();
const state$ = fromInterpret(service);
```

Andarist: > Maybe we don't really need to do that and it's just fine to expose a fromService that returns the stream?

I like that in terms of composition and a single responsibility principle... OTOH, the DX seems to suffer a little bit here. I need to think about it

aldovelco: I think it would be wise to wait for Angular 16 before starting to work on an Angular package.
The Angular team is going to introduce a [DestroyRef](https://next.angular.io/api/core/DestroyRef) that we'll be able to use for stopping the machine without having to provide an additional service in the providers array.

Also, with the introduction of Signals as a new reactive primitive things may look a lot different in the Angular scene.

davidkpiano:
Doing some cleanup 🧹 

FWIW XState actors are already fully observable and interop with RxJS, but if you have further ideas on what other kind of support we could add I'm open to suggestions.

PR #3856, opened 2023-02-25
[xstate/solid] Fix array deletions/additions causing duplicate entries

### LINKED ISSUE #3853: Bug: @xstate/solid not updating arrays properly (createImmutable.ts)
### Description

I writed this reproducible example:
https://codesandbox.io/p/github/russiann/vite-solid/main?file=%2Fsrc%2Fmain.tsx

Scenario:

When you delete Item 1.
This happens:

CleanShot 2023-02-23 at 16 23 56@2x

Note that Item 3 is duplicated.
If you delete from top to down this duplication happens with other item.

The only way to delete all without bug is from down to top starting from last hehe

I tried to figure out whats happening and I my guess is that `updateStore` function at `createImmutable.ts` is not updating arrays correctly. During debugging I noticed that the loop at line 76 is not applying diff in all indexes it should do. I even tried to fix it somehow but I did not understand the `updateStore` properly (I started to try solidjs these days so I'm newbie) so I create this issue =)

When I return `setStore` instead `setImmutable` at line 109 everything works.

### Expected result

Arrays should be updated correctly

### Actual result

Some indexes are being ignored when updateStore update arrays

### Reproduction

https://codesandbox.io/p/github/russiann/vite-solid/main?file=%2Fsrc%2Fmain.tsx

### Additional context

_No response_

GoldingAustin:
Add test to cover affected array updates
Fixes #3853

### DIFF
--- a/packages/xstate-solid/src/createImmutable.ts
+++ b/packages/xstate-solid/src/createImmutable.ts
@@ -53,7 +53,7 @@ const updateStore = <Path extends unknown[]>(
       const largestSize = Math.max(next.length, prev.length);
 
       // Diff array
-      for (let start = 0, end = smallestSize - 1; start < end; start++, end--) {
+      for (let start = 0, end = largestSize - 1; start < end; start++, end--) {
         diff(next[start], prev[start], [...path, start] as Path);
         diff(next[end], prev[end], [...path, end] as Path);
       }
--- a/.changeset/quick-ads-compete.md
+++ b/.changeset/quick-ads-compete.md
@@ -1,0 +1,5 @@
+---
+'@xstate/solid': patch
+---
+
+Fix array deletions/additions causing duplicate entries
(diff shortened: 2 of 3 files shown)


packages/xstate-solid/src/createImmutable.ts
```diff
@@ -53,7 +53,7 @@ const updateStore = <Path extends unknown[]>(
       const largestSize = Math.max(next.length, prev.length);
 
       // Diff array
-      for (let start = 0, end = smallestSize - 1; start < end; start++, end--) {
+      for (let start = 0, end = largestSize - 1; start < end; start++, end--) {
```
Andarist, line 56: could you explain what happens here and what was the meaning of the old code + what is the meaning of the new one?

GoldingAustin: This loops through and diffs each index of the old and new array. The old code only diffed between index 0 and the length of the smallest array, because in the next if statement, we handle any new or now undefined indexes in the new array.

The new code diffs between index 0 and the length of the largest array. This will always diff and update the entire array in the loop; then the if statement will ensure the length and new indexes are correct. The new code may hit some indexes twice in diffing, but it won't affect performance in any meaningful way, and we shouldn't see any duplicate array indexes anymore.

PR #3864, opened 2023-02-28
[xstate/test] Simpler graph events

davidkpiano:
This PR simplifies the options for getting events for graph traversal by introducing a new `events` property that can either be:

- an array of event objects (most common usage)
- a function that accepts the current `state` as an argument and returns an array of event objects

This consolidates the existing `getEvents` and `eventCases` properties:

```diff
const shortestPaths = getShortestPaths(someMachine, {
- getEvents: () => { ... },
- eventCases: { ... },
+ events: [ ... ]
});
```

Event cases should now be specified as an array of events:
```diff
const shortestPaths = getShortestPaths(someMachine, {
- eventCases: {
-   click: [{ x: 10, y: 10 }, { x: 20, y: 20 }]
- }
+ events: [
+   { type: 'click', x: 10, y: 10 },
+   { type: 'click', x: 20, y: 20 }
+ ]
});
```


packages/xstate-test/test/paths.test.ts
```diff
@@ -172,15 +175,17 @@ describe('transition coverage', () => {
...
+        { type: 'NEXT', value: 100 },
+        { type: 'NEXT', value: 1000 }
+      ]
     });
 
     // { value: 1000 } already covered by first guarded transition
     expect(paths.map((path) => path.description)).toMatchInlineSnapshot(`
       Array [
-        "Reaches state \\"b\\": NEXT ({\\"value\\":0})",
+        "Reaches state \\"b\\": NEXT ({\\"value\\":0}) → NEXT ({\\"value\\":0})",
```
Andarist, line 188: is this an intentional change? a bug that was fixed perhaps?

packages/xstate-test/test/index.test.ts
```diff
@@ -114,9 +115,10 @@ describe('events', () => {
     const testedEvents: any[] = [];
 
     const testModel = createTestModel(testMachine, {
-      eventCases: {
-        EVENT: (state) => state.context.values.map((value) => ({ value }))
-      }
+      events: (state) =>
+        state.context.values.map(
+          (value) => ({ type: 'EVENT', value } as EventFrom<typeof testMachine>)
+        )
```
Andarist, line 121: This is closer to what I would expect from an average user:
```suggestion
        state.context.values.map((value) => ({ type: 'EVENT', value } as const))
```

packages/xstate-test/test/index.test.ts
```diff
@@ -1,4 +1,4 @@
-import { assign, createMachine } from 'xstate';
+import { EventFrom, assign, createMachine } from 'xstate';
```
Andarist, line 1: With [the proposed change](https://github.com/statelyai/xstate/pull/3864/files#r1123172455) this import becomes obsolete
```suggestion
import { assign, createMachine } from 'xstate';
```

packages/xstate-graph/test/types.test.ts
```diff
@@ -52,28 +50,20 @@ describe('types', () => {
...
-            value: 100
-          }
-        ]
-      }
+      events: [
+        {
+          type: 'UNKNOWN',
+          value: 100
+        } as any
+      ]
```
Andarist, line 66: This is a type test - so it's whole point was to check for the error ;p
```suggestion
      events: [
        {
          // @ts-expect-error
          type: 'UNKNOWN',
          value: 100
        }
      ]
```

packages/xstate-graph/src/types.ts
```diff
@@ -187,10 +183,7 @@ export interface TraversalConfig<TState, TEvent extends EventObject>
    * `event` when building the adjacency map.
    */
   filter: (state: TState, event: TEvent) => boolean;
-  getEvents: (
-    state: TState,
-    cases: EventCaseMap<TState, TEvent>
-  ) => ReadonlyArray<TEvent>;
+  events: TEvent[] | ((state: TState) => TEvent[]);
```
Andarist, line 186:
```suggestion
  events: readonly TEvent[] | ((state: TState) => readonly TEvent[]);
```

this change should come with 2 tests that would exercise `as const` on the whole array (and not on the array elements)

packages/xstate-test/test/paths.test.ts
davidkpiano: Updated snapshots

packages/xstate-test/test/index.test.ts
davidkpiano: @Andarist Ideally we shouldn't have to specify `as const` or anything in there; do you know why inference is lacking here?

packages/xstate-graph/src/types.ts
davidkpiano: This doesn't work:

```ts
  it('`events` should not require all event types (function)', () => {
    const machine = createMachine<
      unknown,
      { type: 'FOO'; value: number } | { type: 'BAR'; value: number }
    >({});

    getMachineShortestPaths(machine, {
      events: () => [{ type: 'FOO', value: 100 }] as const
    });
  });
```

But this does:
```ts
  it('`events` should not require all event types (function)', () => {
    const machine = createMachine<
      unknown,
      { type: 'FOO'; value: number } | { type: 'BAR'; value: number }
    >({});

    getMachineShortestPaths(machine, {
      events: () => [{ type: 'FOO', value: 100 } as const]
    });
  });
```

Andarist: I think that you didn't rebuild the project or something. I had the same problem but after running `yarn build` it went away.

I pushed out this change and the tests, let's wait for the CI to be extra sure

packages/xstate-test/test/index.test.ts
Andarist: Not (yet). We suffer from this problem in multiple places. The problem arises when the return type of the expression should be constrained by what we want to infer from other sources: [TS playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAeAFQFEA3EVDeEAD0NWCPgG8CBPABxAAueCRhZUAc3gBfAHwAKHACMAVsM4AoePAkgMNOhiLD5ASngBeWfDhRgeCD3hVa9ANoBdANwbpp4dQ4WMA+GgD0YQQAFlgcAO44MADWHPKoOPDiiCBw6CCmGoQkpNwY-ELwAEQAYgDytZUyCprauvquRibmVvBupeXCNfWN0h4ANL6moaCQsAizRBwAslBgMaggFAb01i0EMFCoRFjYeCYg1MIuht3WgcE+0hrhkRgxHPbEqADkDAnJ8HkWRydDA+Q0M2gcCQaEwuHwRQwACYth1GCw6OwuLwBMJROIpHJ5Fp4ABbVbrCorNbiTbXHYTbTKNRcEmtPTbTqA242EB2BxOekYTw+bRPfzwe4hZ6IpHyTjSeBQZYU2klHEVIYNJpjVns9qGYzcyzWPoawZ1bWjCZ+HxAA)

PR #3874, opened 2023-03-02
Initialize the `Interpreter`'s state eagerly

Andarist:


davidkpiano (approved):
Let's get this in. We still need to think about hot reloading btw; `.initState()` might not play well with that.

PR #3877, opened 2023-03-04
[v5] Remove observable replay

davidkpiano:
This PR makes `actor.subscribe(...)` a hot observable; it will no longer emit the current snapshot, which can be retrieved synchronously via `actor.getSnapshot()`.


davidkpiano:
@GoldingAustin Do you know what can be done here to make `@xstate/solid` work with the actor (service) as a _hot_ observable? We can still get the initial state from `actor.getSnapshot()`.

tivac:
I can't look at the Linear ticket but would love to understand the reasoning behind this change. I'm used to svelte store style observables so `.subscribe()` immediately sending the current value was exactly how I expected it to work.

Andarist:
Often u might be interested to listen for a **change**. This might be different from UI-related needs where you might want to **render** a retrieved value immediately though. The point is that if you want to listen for a change then it's really quirky to ignore the first emitted value but when you want to combine change+current value... that is more straightforward to do.

tivac:
Thanks for explaining the reasoning @Andarist, I really appreciate the insight. I'm firmly over here in UI-land so this isn't a change I'm excited about but at least I understand it now.

davidkpiano:
> Thanks for explaining the reasoning @Andarist, I really appreciate the insight. I'm firmly over here in UI-land so this isn't a change I'm excited about but at least I understand it now.

@tivac You're in luck; since `@xstate/svelte` implements stores like this:

```ts
  //  initial snapshot   vvvvvvvvvvvvvvvvvvvvv
  const state = readable(service.getSnapshot(), (set) => {
    return service.subscribe((state) => {
      if (state.changed) {
        set(state);
      }
    }).unsubscribe;
  });
```

I believe everything will still work the same, and you will have the initial snapshot available immediately regardless of these changes. It might even work better because you won't get a redundant update (two initial snapshots - one with `.getSnapshot()` and one upon subscribing).

tivac:
I'm mostly subscribing to interpreters via `xstate-component-tree` these days, but that library will certainly be updated to work with v5 so it's not a huge deal.

GoldingAustin:
I'll take a look at this today!

GoldingAustin:
> It seems like the only failing tests are for `@xstate/solid` - @GoldingAustin can you help with this?
> 
> The context is that `.subscribe` will no longer receive the current snapshot; that needs to be read from `actor.getSnapshot()`.

This will be a quick fix; I'll remove the `from` test suite. I can update the current tests to not use `from` or create a new function export from xstate-solid that emulates the past `from` functionality. Do you have any preferences on the direction?

davidkpiano:
> > It seems like the only failing tests are for `@xstate/solid` - @GoldingAustin can you help with this?
> > The context is that `.subscribe` will no longer receive the current snapshot; that needs to be read from `actor.getSnapshot()`.
> 
> This will be a quick fix; I'll remove the `from` test suite. I can update the current tests to not use `from` or create a new function export from xstate-solid that emulates the past `from` functionality. Do you have any preferences on the direction?

I think the new function export would be the best direction (cc. @Andarist)

packages/core/test/input.test.ts
```diff
@@ -141,12 +141,16 @@ describe('input', () => {
 
     const observableActor = interpret(observableBehavior, {
       input: { count: 42 }
-    }).start();
+    });
 
-    observableActor.subscribe((state) => {
+    const sub = observableActor.subscribe((state) => {
+      if (state?.count !== 42) return;
```
Andarist, line 147: is this one needed?
```suggestion
```

davidkpiano: Yes, because it is initially called with `undefined`

.changeset/lemon-mails-smoke.md
```diff
@@ -0,0 +1,19 @@
+---
+'xstate': major
+---
+
+Observing an actor via `actor.subscribe(...)` no longer immediately receives the current snapshot. Instead, the current snapshot can be read from `actor.getSnapshot()`, and observers will receive snapshots only when a transition in the actor occurs.
```
Andarist, line 5: note for me - this might require a small tweak in the wording if we decide to go with https://github.com/statelyai/xstate/pull/3912

Andarist (approved):

PR #3878, opened 2023-03-04
[v5] Restrict stopping actors

davidkpiano:
An actor cannot be stopped directly. It can only be stopped by its parent internally, or by stopping the root actor

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -141,7 +141,7 @@ export class Interpreter<
     const { clock, logger, parent, id, systemId } = resolvedOptions;
     const self = this;
 
-    this.system = parent?.system ?? createSystem();
+    this.system = parent?.system ?? createSystem(this);
 
     if (systemId) {
       this._systemId = systemId;
@@ -164,7 +164,14 @@ export class Interpreter<
       defer: (fn) => {
         this._deferred.push(fn);
       },
-      system: this.system
+      system: this.system,
+      stop: (child) => {
+        if (child._parent === this) {
+          child._stop?.();
+        } else {
+          throw new Error('no');
+        }
+      }
     };
 
     // Ensure that the send method is bound to this interpreter instance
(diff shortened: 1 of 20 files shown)


packages/core/src/interpreter.ts
```diff
@@ -165,7 +166,14 @@ export class Interpreter<
...
         this._deferred.push(fn);
       },
-      system: this.system
+      system: this.system,
+      stop: (child) => {
+        if (child._parent === this) {
+          child._stop?.();
+        } else {
+          throw new Error('no');
+        }
```
Andarist, line 175: No. 🤣

packages/core/src/system.ts
```diff
@@ -27,6 +29,9 @@ export function createSystem<T extends ActorSystemInfo>(): ActorSystem<T> {
     set: (key, actorRef) => {
       keyedActors.set(key, actorRef);
       reverseKeyedActors.set(actorRef, key);
+    },
+    stop: () => {
+      rootActor._stop?.();
     }
```
Andarist, line 35: `system` is available everywhere, so we could still stop the whole thing from an actor deep in the "tree" - this feels off

packages/core/src/interpreter.ts
```diff
@@ -365,7 +373,8 @@ export class Interpreter<
   /**
    * Stops the interpreter and unsubscribe all listeners.
    */
-  public stop(): this {
+  public _stop(): this {
+    evict(this, 'initial');
```
Andarist, line 377: As far as I was able to analyze this, this is redundant and basically just a noop because we never `memo(actor, 'initial', ...)` 
```suggestion
```

packages/core/src/system.ts
davidkpiano: Any other suggestions? Nothing's stopping a dev (no pun intended) from doing _anything_ in an action, so no matter where we put stop, we still have this "problem"

PR #3879, opened 2023-03-04
Deprecate `cond` → `guard`

davidkpiano:
This PR deprecates the `cond` property in favor of the `guards` property. The `cond` property will be renamed in the next major version.

```diff
const machine = createMachine({
  // ...
  on: {
    EVENT: {
-     cond: (context, event) => { ... },
+     guard: (context, event) => { ... },
    }
  }
})
```

Note: `cond` will continue to work (no breaking changes); however, you will see a deprecation warning:

CleanShot 2023-03-04 at 14 41 10@2x
CleanShot 2023-03-04 at 14 41 23@2x

### DIFF
--- a/.changeset/rude-geckos-admire.md
+++ b/.changeset/rude-geckos-admire.md
@@ -1,0 +1,17 @@
+---
+'xstate': minor
+---
+
+The `cond` property has been deprecated in favor of the `guards` property. The `cond` property will be renamed in the next major version.
+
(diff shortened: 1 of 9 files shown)


Andarist:
This change is likely not worth it at the moment. We would have to first add support for the `guard` property in the tooling (from typegen through bidi edits to code<->diagram conversions)

PR #3887, opened 2023-03-13
[v5] Add `self` to meta objects

davidkpiano:
This PR makes `self` available for action and guard meta to make it easier to reference the `self` actor ref so that actions such as `sendTo` can include it in the event payload:

```ts
// Sender
actions: sendTo('somewhere', (ctx, ev, { self }) => ({
  type: 'EVENT',
  ref: self
}),

// ...

// Responder
actions: sendTo((ctx, ev) => ev.ref, ...)
```


davidkpiano:
Superseded (really renamed) by https://github.com/statelyai/xstate/pull/3911

PR #3888, opened 2023-03-13
[v5] Do not replay actions from a persisted state

davidkpiano:
This PR changes the behavior of restarting a machine from a persisted state, where actions are _not_ replayed (nor recorded in the state object).


davidkpiano:
Moving this to draft as the logic is still in question.

- Built-in actions like `assign()`, `raise()`, etc. will be "executed" regardless, so there is an impedance mismatch here
- It may be desirable to re-execute actions (opt-in or opt-out?) https://github.com/statelyai/xstate/discussions/3690

In fact, if we have the default behavior of _always_ re-executing actions, we would have more consistent behavior, and allow for opt-out logic:

```js
// pseudocode
createMachine({
  entry: () => {
    if (thisActionWasAlreadyExecuted) {
      return;
    }
    // execute action
  },
  // ...
});
```

davidkpiano:
Superseded by https://github.com/statelyai/xstate/pull/3926

PR #3889, opened 2023-03-13
[v5] Remove `autoForward`

davidkpiano:
The recommended pattern moving forward would be to explicitly `sendTo(...)` events to actors.

~~There are still some SCXML tests that depend on `autoForward`.~~ XState will no longer support `autoForward` in SCXML.


packages/core/test/invoke.test.ts
```diff
@@ -232,7 +235,7 @@ describe('invoke', () => {
     service.send({ type: 'FORWARD_DEC' });
   });
 
-  it('should forward events to services if autoForward: true before processing them', (done) => {
+  it('another forwarding pattern', (done) => {
```
Andarist, line 238: IMHO this test could just be removed, it was testing a specific thing which is no longer relevant and the new test title is meaningless

packages/core/test/invoke.test.ts
```diff
@@ -162,7 +164,7 @@ describe('invoke', () => {
       .start();
   });
 
-  it('should forward events to services if autoForward: true', () => {
+  it('can forward events to services', () => {
```
Andarist, line 167: This test title is also kinda irrelevant now, we are not even forwarding anything here, just sending an event to a child from within a transition action. Perhaps we should rename this to something like:
```suggestion
  it('child can immediately respond to the parent with multiple events', () => {
```

packages/core/test/invoke.test.ts
```diff
@@ -118,16 +118,18 @@ describe('invoke', () => {
           start: {
             invoke: {
               src: 'child',
-              id: 'someService',
-              autoForward: true
+              id: 'someService'
```
Andarist, line 121: similarly, this test doesn't quite bring any value, I think - it's almost identical to the one below it

davidkpiano: Should we delete it?

Andarist: i think so

packages/core/test/invoke.test.ts
Andarist: 👋

davidkpiano: Will delete

PR #3890, opened 2023-03-13
[v5] Rename `state._sessionid` to `state.sessionId`

davidkpiano:
This Pr renames `state._sessionid` to `state.sessionId` for consistency.


.changeset/slow-lizards-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': major
+---
+
+The `state._sessionid` property has been removed. It should be obtained directly from the actor: `actor.sessionId`.
```
Andarist, line 5: I'm not sure if we even need to expose this anyhow on the `actor` - but we can bikeshed this later on.

Andarist (approved):

PR #3893, opened 2023-03-15
Remove `withContext` and `.provide({ context })`

Andarist:


packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -230,11 +230,12 @@ describe('createActorContext', () => {
...
-    const someMachine = createMachine({
-      context: { count: 0 }
-    });
+    const createSomeMachine = (context: { count: number }) =>
+      createMachine({
+        context
+      });
 
-    const SomeContext = createActorContext(someMachine);
+    const SomeContext = createActorContext(createSomeMachine({ count: 0 }));
```
davidkpiano, line 238: Hmm, now I feel that either:

- `createActorContext(machine, options?)` should take a 2nd arg for interpret options
- `createActorContext(actor)` should accept an actual (started or not) actor

Andarist: > createActorContext(machine, options?) should take a 2nd arg for interpret options

This is actually already the case.

> createActorContext(actor) should accept an actual (started or not) actor

a good chunk of the problem here is that many of those `options` are meant to be provided inside a component in the React world (or rather: people want to provide them there, they might be coming from props etc). So we can't expect all of them to be given upfront to us and thus we can't really require the actor to be created outside of the component. That being said... `Provider` doesn't accept `state` and `input` right now (at type-level) and likely it should accept them.

PR #3897, opened 2023-03-17
Allow input to be referenced in the implementations

Andarist:


packages/core/src/StateMachine.ts
```diff
@@ -90,7 +94,13 @@ export class StateMachine<
     const resolvedContext =
       typeof context === 'function'
         ? context({
-            spawn: createSpawner(actorCtx?.self, this, actions),
+            spawn: createSpawner(
+              actorCtx?.self,
+              this,
+              undefined as any, // TODO: this requires `| undefined` for all referenced inputs that are spawnable in the context factory,
```
Andarist, line 100:
```suggestion
              undefined as any, // TODO: this requires `| undefined` for all referenced dynamic inputs that are spawnable in the context factory,
```
This is the only problematic part of this PR from my PoV. Alternatively, we could only allow input-less actors to be spawned within the context factory... but that's perhaps even worse. A good thing is that you can always use `spawn('child', { input })` in the context factory.

packages/core/test/input.test.ts
```diff
@@ -159,4 +159,204 @@ describe('input', () => {
...
+            }),
+            input: 42
+          }
+        }
+      }
+    );
+
+    interpret(machine).start();
+
+    expect(spy).toHaveBeenCalledWith(69);
```
davidkpiano, line 360: Just to keep it clean 😅 
```suggestion
          childRef: spawn('child', { input: 100 })
        }))
      },
      {
        actors: {
          child: {
            src: createMachine({
              context: ({ input }) => {
                spy(input);
                return {};
              }
            }),
            input: 42
          }
        }
      }
    );

    interpret(machine).start();

    expect(spy).toHaveBeenCalledWith(100);
```

packages/core/test/input.test.ts
```diff
@@ -159,4 +159,204 @@ describe('input', () => {
...
```
davidkpiano, line 331:
```suggestion
          input: 100
        }
      },
      {
        actors: {
          child: {
            src: createMachine({
              context: ({ input }) => {
                spy(input);
                return {};
              }
            }),
            input: 42
          }
        }
      }
    );

    interpret(machine).start();

    expect(spy).toHaveBeenCalledWith(100);
```

PR #3907, opened 2023-03-22
Fix warnings and TS errors for `xstate/graph` tests

kevinmaes:
Fixes warnings and TypeScript errors in tests for xstate/graph

To suppress console warnings due to a missing `predictableActionArguments: true` in XState machine configs, just use `--silent` when running tests.

### DIFF
--- a/.changeset/weak-kangaroos-roll.md
+++ b/.changeset/weak-kangaroos-roll.md
@@ -1,0 +1,5 @@
+---
+'@xstate/graph': patch
+---
+
+Fix warnings and TypeScript errors for tests in xstate/graph
--- a/.changeset/weak-kangaroos-roll.md
+++ b/.changeset/weak-kangaroos-roll.md
@@ -1,5 +1,0 @@
----
-'@xstate/graph': patch
----
-
-Fix warnings and TypeScript errors for tests in xstate/graph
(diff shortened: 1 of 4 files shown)


packages/xstate-graph/test/types.test.ts
```diff
@@ -1,11 +1,11 @@
 import { createMachine } from 'xstate';
-import { getMachineShortestPaths } from '../';
+import { getMachineShortestPaths } from '../src/shortestPaths';
```
kevinmaes, line 2: Fixes the TS errors when passing args to `getMachineShortestPaths`

packages/xstate-graph/test/graph.test.ts
```diff
@@ -61,7 +60,8 @@ describe('@xstate/graph', () => {
     }
   };
 
-  const lightMachine = Machine({
+  const lightMachine = createMachine({
+    predictableActionArguments: true,
```
Andarist, line 64: Those warnings in the test output are annoying but I think it's better to avoid using this flag in tests right now. The reason is that this flag doesn't exist in v5 (it's just the default) and we don't want to end up cleaning up those flags when merging to `next` (+we avoid conflicts by not using this flag)

.changeset/weak-kangaroos-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/graph': patch
+---
+
+Fix warnings and TypeScript errors for tests in xstate/graph
```
Andarist, line 5: This PR only touches test files so it doesn't need a public-facing changeset file
```suggestion
---
---
```

packages/xstate-graph/test/types.test.ts
kevinmaes: TS errors seem to sometimes be reported when running tests in watch mode
CleanShot 2023-03-22 at 10 58 55

PR #3911, opened 2023-03-22
[v5] Add `self` to meta objects

davidkpiano:
This PR makes `self` available for action and guard meta to make it easier to reference the `self` actor ref so that actions such as `sendTo` can include it in the event payload:

```ts
// Sender
actions: sendTo('somewhere', (ctx, ev, { self }) => ({
  type: 'EVENT',
  ref: self
}),

// ...

// Responder
actions: sendTo((ctx, ev) => ev.ref, ...)
```


packages/core/src/guards.ts
```diff
@@ -104,14 +92,19 @@ export function evaluateGuard<
...
   const { machine } = state;
   const guardMeta: GuardMeta<TContext, TEvent> = {
     state,
     guard,
     _event,
-    evaluate: evaluateGuard
+    evaluate: evaluateGuard,
+    // TODO: `self` should always be "defined", and developers should expect it to exist
+    // only when they're actually interpreting the machine.
+    // Trying to access `self` with an uninterpreted machine should fail correctly.
```
Andarist, line 106: it's nice to have the exact same shape of "meta" stuff in all implementation types. I wonder though - what are practical use cases for having access to `self` in guards? Remember that having it here impacts purity/idempotency and that has negative consequences for APIs like `state.can`

packages/core/src/actions/send.ts
```diff
@@ -83,8 +84,10 @@ export function send<
             ? eventOrExpr.name
             : eventOrExpr.type
       };
-      const meta = {
-        _event
+      const meta: SCXMLEventMeta<TEvent> = {
+        _event,
+        state: state as any, // TODO: fix
```
Andarist, line 89: fix what and how? we have two hacks~ line by line here :P

packages/core/src/types.ts
```diff
@@ -1544,16 +1542,15 @@ export interface ValueAdjacencyMap<
   [stateId: string]: Record<string, State<TContext, TEvent>>;
 }
 
-export interface SCXMLEventMeta<TEvent extends EventObject> {
-  _event: SCXML.Event<TEvent>;
-}
+export type SCXMLEventMeta<TEvent extends EventObject> = StateMeta<any, TEvent>;
```
Andarist, line 1545: This is somewhat weirdly named - what does it have to do with SCXML now? I'd probably remove this one altogether and rename `StateMeta` to something like `ExecutionMeta`

.changeset/eleven-forks-sleep.md
```diff
@@ -0,0 +1,18 @@
+---
+'xstate': major
+---
+
+The `self` actor reference is now available in all action and guard meta. This makes it easier to reference the "self" `ActorRef` so that actions such as `sendTo` can include it in the event payload:
```
Andarist, line 5: It's also passed to `delay` creators - this can be done to keep things the same/similar although I really can't find a reason to use `self` in there (and perhaps we could extend the argument against adding `self` to guards here, `delay`s could be treated as pure/idempotent).

It's not passed to `input` factories though - and that one should likely be possible. The parent should be able to pass its own reference to the child through `input`

Andarist (approved):

PR #3912, opened 2023-03-23
Avoid notifying subscribers when snapshot didn't change

Andarist:


.vscode/launch.json
```diff
@@ -19,7 +19,13 @@
       "request": "launch",
       "name": "Jest Current File",
       "program": "${workspaceFolder}/node_modules/.bin/jest",
-      "args": ["${file}", "--config", "jest.config.js", "--no-cache"],
+      "args": [
+        "${file}",
+        "--config",
+        "jest.config.js",
+        "--no-cache",
+        "--testTimeout=2147483647"
```
Andarist, line 27: makes it easier to debug tests in VS Code since Jest won't kill the process after the default timeout when we are just looking around while being paused in the debugger

packages/core/src/interpreter.ts
```diff
@@ -204,8 +205,10 @@ export class Interpreter<
       deferredFn(state);
     }
 
-    for (const observer of this.observers) {
-      observer.next?.(snapshot);
+    if (previousSnapshot !== snapshot) {
+      for (const observer of this.observers) {
+        observer.next?.(snapshot);
+      }
     }
```
Andarist, line 212: that's the core of the change really but I had to remove mutations from actors to make this check work since with mutations... the snapshots **were** the same, they were just mutated ;p

davidkpiano:
I'm not sure if this change is desired. IMO it should be up to the consumer to decide what to do if the snapshot is the same; this is why operators like `distinctUntilChanged` exist in RxJS, for example.

To me, if I send N events to a state machine actor, I expect at least N observed snapshots, even if some of them didn't change. Not emitting a snapshot is ambiguous; is the actor stopped, or is the actor working but the value didn't change, or did some other error occur?

Andarist:
Hm, yeee - I think I agree that maybe it's not our job *here* to deduplicate those notifications. I feel though that maybe there is something funky going around notifications coming from `.start()`. OTOH, that is kinda sugar over `send({ type:'xstate.init' })`, so maybe it's just fine to emit that synchronous notification on `.start()`.

PR #3915, opened 2023-03-23
[v5] Remove `.onTransition()`

davidkpiano:
This PR removes `.onTransition(...)` in favor of `.subscribe(...)`

```diff
 const actor = interpret(machine)
-  .onTransition(...)
-  .start();
+actor.subscribe(...);
+actor.start();
```

### DIFF
--- a/packages/core/test/interpreter.test.ts
+++ b/packages/core/test/interpreter.test.ts
@@ -54,18 +54,6 @@ const lightMachine = createMachine({
 
 describe('interpreter', () => {
   describe('initial state', () => {
-    it('immediately notifies the listener with the initial state and event', (done) => {
-      const service = interpret(idMachine);
-      service.subscribe((initialState) => {
-        expect(initialState).toBeInstanceOf(State);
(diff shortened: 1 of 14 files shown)


packages/core/test/interpreter.test.ts
```diff
@@ -55,7 +55,8 @@ const lightMachine = createMachine({
 describe('interpreter', () => {
   describe('initial state', () => {
     it('immediately notifies the listener with the initial state and event', (done) => {
```
Andarist, line 57: this test should be removed

packages/core/test/interpreter.test.ts
```diff
@@ -212,13 +213,15 @@ describe('interpreter', () => {
       setTimeout(() => done(), 10);
     });
 
-    it('should notify subscribers of the current state upon subscription (onTransition)', (done) => {
-      const service = interpret(machine).start();
+    it('should notify subscribers of the current state upon subscription', (done) => {
```
Andarist, line 216: this test also should be removed

Andarist (approved):

PR #3919, opened 2023-03-24
Fix warning about unmet peer dependencies

c-w:
When installing `@xstate/react` on a React 18 project, we get a warning that `use-isomorphic-layout-effect@1.0.0` doesn't support this version of React. As such, this pull request bumps the required minimum version of the dependency to fix the warning. There were no breaking changes in the dependency between 1.0.0 and 1.1.2 so from what I can see, this can be released as a minor version bump.

### DIFF
--- a/.changeset/stupid-buttons-brake.md
+++ b/.changeset/stupid-buttons-brake.md
@@ -1,0 +1,5 @@
+---
+'@xstate/react': minor
+---
+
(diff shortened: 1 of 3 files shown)


Andarist:
FYI, this has almost no effect on consuming projects. Dependencies are using ranges for a reason - 1.2.0 is compatible with ^1.0.0

PR #3921, opened 2023-03-25
[v5] Persist & restore spawned actors

davidkpiano:


packages/core/src/State.ts
```diff
@@ -291,6 +291,8 @@ export function getPersistedState<TState extends AnyState>(
   const { configuration, transitions, tags, machine, children, ...jsonValues } =
     state;
 
+  const contextJSON = JSON.stringify(state.context);
```
Andarist, line 294: I don't think that we should be using `JSON.stringify` here. It's super lossy and doesn't handle some basic things - like `undefined`, `Map` etc.

When a user is using those kind of data structures they should be aware of the serializability issues and they need to implement their own serialize/deserialize logic (as in: we can't possibly handle that for them). 

The only thing that we should replace here are actor references as that's something the user can't do on their own.

packages/core/src/State.ts
```diff
@@ -291,6 +291,8 @@ export function getPersistedState<TState extends AnyState>(
...
+
   const childrenJson: Partial<PersistedMachineState<any>['children']> = {};
 
   for (const id in children) {
```
Andarist, line 298: we should throw when an inline actor is being persisted since we won't be able to restore it later

packages/core/src/State.ts
davidkpiano: No longer being used

packages/core/src/State.ts
davidkpiano: How do we detect these again?

davidkpiano:
@Andarist This is ready for final review

packages/core/src/StateMachine.ts
```diff
@@ -450,6 +450,17 @@ export class StateMachine<
...
     });
 
+    const context = JSON.parse(JSON.stringify(state.context), (_, value) => {
+      // if value is an object with $$type property, it is a reference to an actor
+      if (typeof value === 'object' && value && value.$$type === 'ActorRef') {
+        return children[value.id];
+      }
+
+      return value;
+    });
```
Andarist, line 460: I think this "hydration" has to be system-wide to cover for cases like this:
```ts
const child = createMachine({
  context: (input) => ({ parent: input.parent })
})
const parent = createMachine({
  invoke: {
    src: child,
    input: ({ self }) => ({ parent: self })
  }
})
interpret(parent).getPersistedState()
```

packages/core/src/StateMachine.ts
```diff
@@ -450,6 +450,17 @@ export class StateMachine<
       }
     });
 
+    const context = JSON.parse(JSON.stringify(state.context), (_, value) => {
```
Andarist, line 453: This should be implemented as a custom traversal thing (that clones the affected object paths). Piggy-backing like this on `JSON.parse`+`JSON.stringify` like this is too simple of a solution that won't cover for a lot of situations. 

The persisted state should be serializable - but that doesn't necessarily mean that it has to be JSON-serializable. What if users use a different format (like even `superjson`)?

packages/core/src/StateMachine.ts
```diff
@@ -450,6 +450,17 @@ export class StateMachine<
...
+      // if value is an object with $$type property, it is a reference to an actor
+      if (typeof value === 'object' && value && value.$$type === 'ActorRef') {
+        return children[value.id];
+      }
+
+      return value;
+    });
+
+    restoredState.context = context;
```
Andarist, line 462: nit-level but I think that it would just be better to create this rehydrated context first and create `State` right away with it instead of re-assigning it after the fact

packages/core/src/State.ts
Andarist: don't we keep `src` on the `ActorRef`? I think that it makes to only keep there strings that reference actors in `implementations`. If we'd do that then differentiating an inline and a referenced actor would be super simple

packages/core/src/StateMachine.ts
```diff
@@ -653,6 +653,20 @@ export class StateMachine<
       }
     });
 
+    const context = JSON.parse(
+      JSON.stringify(restoredSnapshot.context),
```
Andarist, line 657: I don't think we should be using JSON.stringify+JSON.parse dance here. This can legitimately drop values from the `restoredSnapshot.context`. We should write a custom "reviver" here

packages/core/src/State.ts
```diff
@@ -342,3 +347,30 @@ export function getPersistedState<
...
+  let copy: typeof contextPart | undefined;
+  for (const key in contextPart) {
+    const value = contextPart[key];
+    if (value && typeof value === 'object') {
+      if ('sessionId' in value && 'send' in value && 'ref' in value) {
+        copy ??= Array.isArray(contextPart)
+          ? (contextPart.slice() as typeof contextPart)
+          : { ...contextPart };
+        copy[key] = {
+          xstate$$type: $$ACTOR_TYPE,
```
davidkpiano, line 361: We need to standardize these kinds of keys 😂

packages/core/test/actorLogic.test.ts
```diff
@@ -781,6 +781,76 @@ describe('machine logic', () => {
...
+
+    expect(persistedState.children.child.state.context).toEqual({ count: 42 });
+
+    const newActor = createActor(machine, {
+      state: persistedState
+    }).start();
+
+    const snapshot = newActor.getSnapshot();
+
+    // TODO: should this be a "dummy actor" so that the types are accurate?
```
davidkpiano, line 850: For now, no (can re-evaluate later)

packages/core/src/State.ts
Andarist: i hope we don't have to add any more special values but with this in place, a key like `xstate$$type` is already pretty unique and its values can be extended if any need arises

Andarist (approved):

PR #3924, opened 2023-03-26
[v5] Rename reducer -> transition function

davidkpiano:
This PR renames `fromReducer` to `fromTransition`.

### DIFF
--- a/packages/core/src/actors/transition.ts
+++ b/packages/core/src/actors/transition.ts
@@ -2,14 +2,16 @@ import { ActorBehavior, ActorContext, EventObject } from '../types';
 import { isSCXMLEvent } from '../utils';
 
 /**
- * Returns an actor behavior from a reducer and its initial state.
+ * Returns an actor behavior from a transition function and its initial state.
  *
- * @param transition The pure reducer that returns the next state given the current state and event.
- * @param initialState The initial state of the reducer.
+ * A transition function is a function that takes the current state and an event and returns the next state.
+ *
+ * @param transition The pure transition function that returns the next state given the current state and event.
+ * @param initialState The initial state of the transition function.
  * @returns An actor behavior
  */
 
-export function fromReducer<TState, TEvent extends EventObject>(
+export function fromTransition<TState, TEvent extends EventObject>(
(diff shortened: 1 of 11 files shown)


packages/xstate-graph/test/__snapshots__/graph.test.ts.snap
```diff
@@ -899,6 +899,323 @@ exports[`shortest paths for reducers 1`] = `
 ]
 `;
 
+exports[`shortest paths for transition functions 1`] = `
```
Andarist, line 902: hm, it feels that perhaps there are some outdated snapshots to be removed in this file

packages/core/src/actors/transition.ts
```diff
@@ -2,14 +2,16 @@ import { ActorBehavior, ActorContext, EventObject } from '../types';
...
 
 /**
- * Returns an actor behavior from a reducer and its initial state.
+ * Returns an actor behavior from a transition function and its initial state.
  *
- * @param transition The pure reducer that returns the next state given the current state and event.
- * @param initialState The initial state of the reducer.
+ * A transition function is a function that takes the current state and an event and returns the next state.
+ *
+ * @param transition The pure transition function that returns the next state given the current state and event.
```
Andarist, line 9: i would refrain from calling this a "pure" function - since it can utilize `actorContext` to execute side-effects

packages/core/src/actors/transition.ts
```diff
@@ -2,14 +2,16 @@ import { ActorBehavior, ActorContext, EventObject } from '../types';
...
+ * Returns an actor behavior from a transition function and its initial state.
  *
- * @param transition The pure reducer that returns the next state given the current state and event.
- * @param initialState The initial state of the reducer.
+ * A transition function is a function that takes the current state and an event and returns the next state.
+ *
+ * @param transition The transition function that returns the next state given the current state and event.
+ * @param initialState The initial state of the transition function.
  * @returns An actor behavior
  */
```
Andarist, line 13:
```suggestion
```

Andarist (approved):

PR #3926, opened 2023-03-27
[v5] Prevent action replay

davidkpiano:
This PR removes actions from persisted state, which prevents them from being replayed.


packages/core/src/State.ts
```diff
@@ -302,6 +302,7 @@ export function getPersistedState<TState extends AnyState>(
 
   return {
     ...jsonValues,
-    children: childrenJson
+    children: childrenJson,
+    actions: []
```
Andarist, line 306: shouldn't we also discard those in `StateMachine['restoreState']`? This would require us to change those two tests:
- [`interpreter.test.ts`: executes actions from a restored state](https://github.com/statelyai/xstate/blob/297cd61cef31d19dda3e58e2c76ba7250461d5b0/packages/core/test/interpreter.test.ts#L110-L158)
- [`after.test.ts`: should execute an after transition after starting from a state resolved using `machine.getInitialState`](https://github.com/statelyai/xstate/blob/297cd61cef31d19dda3e58e2c76ba7250461d5b0/packages/core/test/after.test.ts#L140-L159)

davidkpiano: Yeah, we can do that

Andarist: 👋

davidkpiano: I did that

.changeset/hip-melons-clean.md
```diff
@@ -0,0 +1,7 @@
+---
+'xstate': major
+---
+
+Persisted state will no longer contain actions, since they have already been executed. Actions will not be replayed.
```
Andarist, line 5: This is no longer true since you changed the implementation from discarding those actions when persisting to ignoring them when rehydrating. Perhaps for the time being we should do both or adjust this sentence? 

I'll explore tomorrow how feasible it is to remove `State['actions']`

packages/core/test/interpreter.test.ts
```diff
@@ -106,42 +106,34 @@ describe('interpreter', () => {
     });
 
     // https://github.com/statelyai/xstate/issues/1174
```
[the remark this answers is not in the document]
Andarist: 👋

.changeset/hip-melons-clean.md
Andarist: 👋

davidkpiano: Adjusted

packages/core/test/interpreter.test.ts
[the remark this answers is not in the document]
davidkpiano: Removed

PR #3932, opened 2023-03-29
Re-register actors in the system when they start to account for restarted actors

Andarist:


packages/xstate-react/test/useMachine.test.tsx
```diff
@@ -945,4 +945,40 @@ describeEachReactMode('useMachine (%s)', ({ suiteKey, render }) => {
 
     expect(container.textContent).toBe('2');
   });
+
+  it('should deliver messages sent from an effect to an actor registered in the system', () => {
```
Andarist, line 949: this one was already working because we don't rehydrate deep children, so since the child was recreated anyway - it had a chance to register itself in the system

packages/xstate-react/test/useInterpret.test.tsx
```diff
@@ -260,4 +260,29 @@ describeEachReactMode('useInterpret (%s)', ({ suiteKey, render }) => {
 
     expect(childState.textContent).toBe('received');
   });
+
+  it('should deliver messages sent from an effect to the root actor registered in the system', () => {
```
Andarist, line 264: this one wasn't working because `useEffect(() => () => actor.stop())` was deleting the registered `systemId` and it was never re-registered

packages/core/src/interpreter.ts
```diff
@@ -287,6 +290,9 @@ export class Interpreter<
     }
 
     this.system._register(this.sessionId, this);
+    if (this._systemId) {
+      this.system._set(this._systemId, this);
+    }
```
Andarist, line 295: This happens in `.start()` now. I didn't remove the call from the constructor though. it creates a little bit of a weird effect but I'm not sure what we can do about it:
```ts
const root = interpret(m)
// actors in the system are available here
root.start()
// actors in the system are still available
root.stop()
// actors in the system are *not* available here
root.start()
// actors in the system are available again
```

I don't think this is a big issue but it's likely that our users will prove me wrong. Maybe we could consider a hybrid approach - similar to the one that we mentioned yesterday. Registered actors could always be cleared when they are stopped *except* for the top-level manual `.stop()` on the root.

davidkpiano: Let's leave it as-is for now; can always address if it comes up.

davidkpiano (approved):

PR #3934, opened 2023-03-30
[v5] Make `system` accessible in all custom actions

Andarist:


packages/core/test/system.test.ts
```diff
@@ -127,4 +127,39 @@ describe('system', () => {
...
+      },
+      entry: (_, __, { system }) => {
+        expect(system!.get('test')).toBeDefined();
+      }
+    });
+
+    interpret(machine).start();
+  });
+
+  it('should be accessible in referenced custom actions', () => {
```
Andarist, line 145: I only fixed this for "custom actions" aka functions. Previously it was only provided to inline ones but references ones didn't have that privilege. 

It's still not accessible to `assign`, `sendTo` (!), and more. Do we want to add it to every builtin action type?

davidkpiano: I'd say so 👍

davidkpiano (approved):
Approved, but feel free to add it to the other built-in actions

packages/core/test/system.test.ts
davidkpiano: Added here: 4adcba4f648042f249269eb3441c88ccb4836136

PR #3939, opened 2023-04-01
[v5] Unified args

davidkpiano:
This PR unifies arguments. Everywhere that you used to write `(context, event, other)` is now `({ context, event, other })`:

```diff
-actions: (_ctx, event) => { ... }
+actions: ({ event }) => { ... }
```

### DIFF
--- a/.changeset/hot-zoos-destroy.md
+++ b/.changeset/hot-zoos-destroy.md
@@ -1,0 +1,14 @@
+---
+'xstate': major
+---
+
+Action arguments are now consolidated into a single object argument. This is a breaking change for all actions that are called with arguments.
+
+```diff
+assign({
+- count: (context, event) => {
++ count: ({ context, event }) => {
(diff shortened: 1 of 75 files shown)


davidkpiano:
@Andarist Tests and types are passing except for a couple snapshots in `@xstate/inspect` that I'm having a hard time with 😵

Andarist (commented):
It needs a changeset :)

.changeset/hot-zoos-destroy.md
```diff
@@ -0,0 +1,14 @@
+---
+'xstate': major
+---
+
+Action arguments are now consolidated into a single object argument. This is a breaking change for all actions that are called with arguments.
```
Andarist, line 5:
```suggestion
Action/actor/delay/guard arguments are now consolidated into a single object argument. This is a breaking change for all of those things that are called with arguments.
```

PR #3943, opened 2023-04-05
[v5] Remove strict mode

davidkpiano:
This PR removes `strict: true`.


packages/core/test/invoke.test.ts
```diff
@@ -812,13 +812,12 @@ describe('invoke', () => {
       });
 
       // tslint:disable-next-line:max-line-length
-      it('should be invoked with a promise factory and stop on unhandled onError target when on strict mode', (done) => {
+      it('should be invoked with a promise factory and stop on unhandled onError target', (done) => {
```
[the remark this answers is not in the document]
Andarist: hm, after a second thought - the title is correct, it's just that the part that you removed from it was already stale

.changeset/unlucky-months-sort.md
```diff
@@ -0,0 +1,24 @@
+---
+'xstate': major
+---
+
+Strict mode is no longer available. You can no longer set `strict: true` to control:
```
Andarist, line 5: This was already removed [here](https://github.com/statelyai/xstate/pull/3455/files#diff-3f1cb738974f23da45b9ccda1df60076a239709f9354e23ad8bae87c44aef2c1). I think that the changeset added in this PR should be removed (as this PR only clean up some leftover from the other ones). The existing changeset doesn't have a lot of information so if you feel like the information added here is important u can add it to that other one.

davidkpiano: Removed

PR #3947, opened 2023-04-06
[v5] React `.provide` support

davidkpiano:


packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -335,6 +334,9 @@ describe('createActorContext', () => {
 
     render(<App />);
 
-    expect(stubFn).toHaveBeenCalledTimes(1);
+    setTimeout(() => {
+      expect(stubFn).toBeCalled();
+      done();
+    }, 1000);
```
davidkpiano, line 340: Disregard this/change it, I was trying things out and failed

packages/core/src/StateMachine.ts
```diff
@@ -355,6 +355,7 @@ export class StateMachine<
     state: State<TContext, TEvent, TResolvedTypesMeta>,
     actorCtx: ActorContext<TEvent, State<TContext, TEvent, TResolvedTypesMeta>>
   ): void {
+    // TODO: do actions need to be resolved?
```
davidkpiano, line 358:
```suggestion
```

packages/xstate-react/src/createActorContext.ts
```diff
@@ -26,46 +24,25 @@ export function createActorContext<TMachine extends AnyStateMachine>(
...
   useActorRef: () => ActorRefFrom<TMachine>;
-  Provider: (
-    props: {
-      children: React.ReactNode;
-      machine?: TMachine | (() => TMachine);
-    } & (AreAllImplementationsAssumedToBeProvided<
-      TMachine['__TResolvedTypesMeta']
-    > extends false
-      ? {
-          options: InternalMachineImplementations<
```
Andarist, line 37: `<Provider machine={someMachine.provide()} />` is workable but i think that it makes it a little bit awkward because we are repeating `someMachine` here - we already given it to create the `Provider` , it's available internally as a closure, types are bound to it === it looks like a chore that we still require it to be used directly here. 

The best idea I have so far to solve this is for the behavior to somehow declare it's available options. This way the interpreter-level stuff could accept those options in a generic way.

packages/xstate-react/src/useInterpret.ts
```diff
@@ -32,40 +29,21 @@ export function useIdleInterpreter(
   ) {
     const [initialMachine] = useState(machine);
 
-    if (getMachine !== initialMachine) {
+    if (getMachine.config !== initialMachine.config) {
```
Andarist, line 32: Similarly, `.config` is super machine-specific thing. If we want to make `useInterpret` to accept just any actor then it should be possible to check smth like this on just any actor type.

davidkpiano: Good call, I'll think about how we can generalize this.

davidkpiano: Added to 4659f4511539b6c5c6ed4b585796aed0eef0f8e8 - let me know what you think. Might need help on this

packages/xstate-react/src/createActorContext.ts
Andarist: Smth like this could work
```ts
<Context.Provider machine={machine => machine.provide({})}>{null}</Context.Provider>
```

That's quite a bit of `machine` in here though 🤔

PR #3950, opened 2023-04-07
Do not call actions with `state`

Andarist:


packages/core/test/assign.test.ts
```diff
@@ -261,94 +261,62 @@ describe('assign', () => {
...
+var value = globalThis.fetch;
+Object.defineProperty(globalThis, 'fetch', {
+  set(newValue) {
+    debugger;
+    value = newValue;
+  },
+  get() {
+    return value;
+  }
+});
```
davidkpiano, line 273: Hm?

davidkpiano (commented):
LGTM, but I think there's some leftover debugging code

packages/core/test/assign.test.ts
Andarist: XDDDD haha, ye - let's pretend that this didn't happen :D

Andarist:
@davidkpiano removed this leftover

davidkpiano (approved):

PR #3952, opened 2023-04-08
[v5] Final `.data` to `.output`

davidkpiano:
This PR renames `.data` to `.output` on final states:

```diff
const machine = createMachine({
  // ...
  states: {
    // ...
    success: {
-     data: { message: 'Success!' }
+     output: { message: 'Success!' }
    }
  }
})
```

### DIFF
--- a/.changeset/five-mirrors-join.md
+++ b/.changeset/five-mirrors-join.md
@@ -1,0 +1,21 @@
+---
+'xstate': major
+'@xstate/fsm': major
+'@xstate/react': major
+'@xstate/vue': major
+---
+
+The output data on final states is now specified as `.output` instead of `.data`:
+
+```diff
+const machine = createMachine({
+  // ...
+  states: {
+    // ...
+    success: {
+-     data: { message: 'Success!' }
++     output: { message: 'Success!' }
+    }
+  }
+})
+```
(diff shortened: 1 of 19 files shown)


.changeset/five-mirrors-join.md
```diff
@@ -0,0 +1,21 @@
+---
+'xstate': major
+'@xstate/fsm': major
+'@xstate/react': major
+'@xstate/vue': major
```
Andarist, line 5: I don't think that this change affects any of those libraries.
```suggestion
```

packages/core/src/StateNode.ts
```diff
@@ -187,7 +187,7 @@ export class StateNode<
     this.meta = this.config.meta;
     this.doneData =
```
Andarist, line 188: this one could be renamed as well but we can also do it later

.changeset/five-mirrors-join.md
```diff
@@ -0,0 +1,21 @@
...
+The output data on final states is now specified as `.output` instead of `.data`:
+
+```diff
+const machine = createMachine({
+  // ...
+  states: {
+    // ...
+    success: {
+-     data: { message: 'Success!' }
++     output: { message: 'Success!' }
```
Andarist, line 17: this later becomes part of the `done.state.*` and `done.invoe.*` events, should we rename the properties in those events as well?

davidkpiano: I'm not sure. I think there should be a difference between "this is the output of this machine" and "this is the data in this event". Not sure yet.

PR #3957, opened 2023-04-10
[v5] `.types` and `.typegen`

davidkpiano:
This PR:

- renames `config.schema` to `config.types`
- puts `config.tsTypes` inside of `config.types.typegen`

```diff
createMachine({
- tsTypes: {},
- schema: {
+ types: {
+   typegen: {}
  }
});
```


packages/core/test/schema.test.ts
```diff
@@ -1,4 +1,4 @@
-import { createMachine, createSchema } from '../src/index.ts';
+import { createMachine } from '../src/index.ts';
```
Andarist, line 1: It feels like this test file could be removed. It doesn't quite test anything new/useful. Other tests are already testing inference based on `.types` - the `fromJSONSchema` thing here is completely separate from XState, it doesn't matter for inference anyhow (only its return type matters but we already test that elsewhere). 

The "schema should be present in the machine definition" test is also not testing something that we should be guarantee. The `.types` purpose is only type-level and runtime-wise its quite irrelevant. In fact, in this test - its runtime type doesnt match the type-level one, so this is pretty much unusable without casting to `any` or something.

packages/xstate-test/src/types.ts
```diff
@@ -32,7 +32,7 @@ export interface TestMachineConfig<
   TTypesMeta = TypegenDisabled
 > extends TestStateNodeConfig<TContext, TEvent> {
   context?: MachineConfig<TContext, TEvent>['context'];
-  schema?: MachineSchema<TContext, TEvent, ActorMap>;
+  schema?: MachineTypes<TContext, TEvent, ActorMap>;
```
Andarist, line 35:
```suggestion
  types?: MachineTypes<TContext, TEvent, ActorMap, TTypesMeta>;
```

packages/xstate-test/src/types.ts
```diff
@@ -32,7 +32,7 @@ export interface TestMachineConfig<
...
   tsTypes?: TTypesMeta;
```
Andarist, line 36:
```suggestion
```

Issue #1476, opened 2020-09-22
machine ignores invocation when starting from arbitrary state


kristofkalocsai:
**Description**

When explicitly starting the machine via the interpreter from a state which defines an invocation, the machine does not transition from that state, regardless of the outcome -resolve or reject-. Starting from the baked-in initial state, the interpreter and machine working as expected.

```javascript
const machine = Machine({
  id: "machine",
  initial: "S1",
  states: {
    S1: {
      invoke: {
        src: () => {
          return new Promise((resolve) => setTimeout(resolve, 1000));
        },
        onDone: "S2"
      }
    },
    S2: {
      invoke: {
        src: () => {
          return new Promise((resolve) => setTimeout(resolve, 1000));
        },
        onDone: "S1"
      }
    }
  }
});

const service = interpret(machine).onTransition((state) => {
  console.log(state.value);
});

service.start(); // working as expected, 'S1, S2, S1, S2, ...'
// service.start("S1"); // it stays in 'S1'

```

**Expected Result**

The interpreter and machine should be indifferent to the _initialState._

**Actual Result**

However, this is not the case, according to these examples. I hope I'm not messing something up, which causes this unexpected behaviour.

**Reproduction**

- [CodeSandbox (starting without argument) (JavaScript) :heavy_check_mark: ](https://codesandbox.io/s/dank-leaf-nyhk6?file=/src/index.js)
- [CodeSandbox (starting with _initalState_ defined) (JavaScript) :x:](https://codesandbox.io/s/icy-mountain-kb6sv?file=/src/index.js)

**Additional context**

XState version: 4.13.0

Andarist:
This is a delicate matter - starting from a state is mainly supposed to be used when rehydrating machines. And in this context there are legitimate reasons why a service should not be started - it could, for example, reach its final state before and is not supposed to be activated again.

kristofkalocsai:
So is this the intended behaviour?

I was trying to start the machine from a state to be able to test each transition separately, without having to 'play' through all the previous states and conditions, and introducing cross-dependencies.

Andarist:
> So is this the intended behaviour?

I can't be 100% sure - that's a question for @davidkpiano . I only have noticed what kind of problems changing the current behavior could cause.

> I was trying to start the machine from a state to be able to test each transition separately, without having to 'play' through all the previous states and conditions, and introducing cross-dependencies.

[...]

PR #3959, opened 2023-04-10
[v5] Fix persisting unresolved promises

davidkpiano:


davidkpiano:
Addresses #1476

packages/core/test/behaviors.test.ts
```diff
@@ -86,7 +86,31 @@ describe('promise behavior (fromPromise)', () => {
...
+        new Promise<number>((res) => {
+          setTimeout(() => res(42), 10);
+        })
+    );
+
+    const actor = interpret(promiseBehavior);
+    actor.start();
+
+    const resolvedPersistedState = actor.getPersistedState();
+    actor.stop();
```
Andarist, line 101: The problem here was that `.stop()` mutated referenced contained in the `getPersistedState`, right?

davidkpiano: Yes

PR #3966, opened 2023-04-14
[v5] Some export cleanup

davidkpiano:
- `send` is un-exported (use `raise` or `sendTo`)
- `pure` is exported
- `waitFor` is exported

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -14,6 +14,7 @@ import { StateNode } from './StateNode.ts';
 export { SimulatedClock } from './SimulatedClock.ts';
 export { StateMachine } from './StateMachine.ts';
 export { getStateNodes } from './stateUtils.ts';
+export { waitFor } from './waitFor.ts';
 export * from './typegenTypes.ts';
 export * from './types.ts';
 // TODO: decide from where those should be exported
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,21 +1,11 @@
-import {
-  after,
-  done,
-  doneInvoke,
-  escalate,
-  forwardTo,
-  respond,
-  sendParent,
-  sendTo
-} from './actions.ts';
-import { assign } from './actions/assign.ts';
-import { cancel } from './actions/cancel.ts';
-import { choose } from './actions/choose.ts';
-import { log } from './actions/log.ts';
-import { pure } from './actions/pure.ts';
-import { raise } from './actions/raise.ts';
-import { send } from './actions/send.ts';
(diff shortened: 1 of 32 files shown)


packages/core/src/index.ts
```diff
@@ -1,21 +1,11 @@
...
-import { send } from './actions/send.ts';
-import { stop } from './actions/stop.ts';
+import { doneInvoke, forwardTo, sendParent, sendTo } from './actions.ts';
+export { assign } from './actions/assign.ts';
+export { cancel } from './actions/cancel.ts';
+export { choose } from './actions/choose.ts';
+export { log } from './actions/log.ts';
+export { pure } from './actions/pure.ts';
+export { raise } from './actions/raise.ts';
+export { stop } from './actions/stop.ts';
```
Andarist, line 8: With those root exports here we kinda don't need the separate entry point:
https://github.com/statelyai/xstate/blob/d46e0399cecc7c4a4973686a4ef1d74530fe0e1d/packages/core/package.json#L17-L20

I think that there is no value in keeping those exports available in both places, we should just stick to one of them. Which one? I don't have strong opinion about this really, a separate entrypoint might make it a little bit easier to treeshake at times - but I know how to do just anything treehskaeable so it's not a big issue anyway. Let's just think about what is better for the consumer.

packages/core/test/actionCreators.test.ts
```diff
@@ -17,10 +16,13 @@ describe('action creators', () => {
     });
 
     it('should accept an event object', () => {
-      const action = send({ type: 'foo', bar: 'baz' });
+      const action = raise({
+        type: 'foo',
+        // @ts-ignore
+        bar: 'baz'
+      });
```
Andarist, line 23: It could be changed like this:
```suggestion
      const event = {
        type: 'foo' as const,
        bar: 'baz'
      };
      const action = raise<{}, typeof event>(event);
```

Right now we can't avoid explicit type params here, the types are more fine-tuned for nested inference than for creating those action objects standalone. We can explore fixing this in the future but this was the only way to fix the current inference that I was able to find.

packages/core/src/index.ts
davidkpiano: Pros for index:

- Easier migration (devs don't need to do anything from v4 -> v5)

Pros for /actions:

- ~~More tree-shakeable? (maybe not)~~

davidkpiano: I removed this; can you check to see if I've done it correctly?

Andarist:
Could you also reexport everything in the root from `/actors` and `/guards`?

Andarist (approved):

PR #3967, opened 2023-04-14
[v5] Throw when sending strings

davidkpiano:
Event objects should be sent; sending strings is no longer allowed.

### DIFF
--- a/packages/core/src/actions/raise.ts
+++ b/packages/core/src/actions/raise.ts
@@ -74,6 +74,11 @@ export function raise<
       const delaysMap = state.machine.options.delays;
 
       // TODO: helper function for resolving Expr
+      if (typeof eventOrExpr === 'string') {
+        throw new Error(
+          `Only event objects may be sent to actors; use .send({ type: "${eventOrExpr}" }) instead`
+        );
+      }
       const resolvedEvent = toSCXMLEvent(
         typeof eventOrExpr === 'function' ? eventOrExpr(args) : eventOrExpr
       );
--- a/packages/core/src/actions/raise.ts
+++ b/packages/core/src/actions/raise.ts
@@ -76,7 +76,7 @@ export function raise<
       // TODO: helper function for resolving Expr
       if (typeof eventOrExpr === 'string') {
         throw new Error(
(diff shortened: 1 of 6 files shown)


davidkpiano:
@Andarist Should we throw or warn?

packages/core/src/interpreter.ts
```diff
@@ -409,6 +409,12 @@ export class Interpreter<
    * @param event The event to send
    */
   public send(event: TEvent | SCXML.Event<TEvent>) {
+    if (typeof event === 'string') {
+      throw new Error(
+        `Only event objects may be sent to actors; use .send({ type: "${event}" }) instead`
+      );
+    }
```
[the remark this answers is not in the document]
davidkpiano: Sure

packages/core/src/actions/raise.ts
```diff
@@ -74,6 +74,11 @@ export function raise<
       const delaysMap = state.machine.options.delays;
 
       // TODO: helper function for resolving Expr
+      if (typeof eventOrExpr === 'string') {
+        throw new Error(
+          `Only event objects may be sent to actors; use .send({ type: "${eventOrExpr}" }) instead`
+        );
```
Andarist, line 80:
```suggestion
        throw new Error(
          `Only event objects may be used with raise; use raise({ type: "${eventOrExpr}" }) instead`
        );
```

packages/core/src/actions/send.ts
```diff
@@ -95,6 +95,11 @@ export function send<
       const delaysMap = state.machine.options.delays;
 
       // TODO: helper function for resolving Expr
+      if (typeof eventOrExpr === 'string') {
+        throw new Error(
+          `Only event objects may be sent to actors; use .send({ type: "${eventOrExpr}" }) instead`
+        );
```
Andarist, line 101:
```suggestion
        throw new Error(
          `Only event objects may be used with sendTo; use sendTo({ type: "${eventOrExpr}" }) instead`
        );
```

Andarist: Note that I used here `sendTo` instead of `send` - despite the fact that this is in the `send`'s implementation. In a follow-up PR we should remove `send` altogether and just inline its implementation into `sendTo`.

Issue #1416, opened 2020-08-26
TypeError: Cannot read property 'state' of undefined if interpreter passed to useService is undefined


ivorpad:
**Description**
If an interpreter at any point of the message passing dance between parent and child machines is `undefined` then the application crashes. I believe this is a bug because you can't use conditional with Hooks.

**Expected Result**
```
const [state] = useService(undefined)
// state should be undefined
```

This is a problem specially when you're passing down the `service` to a child component:

```jsx
import React from "react";
import { useService } from "@xstate/react";

function Child({ service }) {
  const [state] = useService(service);
  return <div>{state.event.type}</div>;
}

export default Child;
```

Once the child machine reaches its final state the `service` will be `undefined`, thus the app crashes.

**Actual Result**

<img width="490" alt="Screen Shot 2020-08-27 at 1 45 01 AM" src="https://user-images.githubusercontent.com/52452/91367504-f2d95a80-e806-11ea-9ebf-52211fd5bdc6.png">

**Reproduction**
https://codesandbox.io/s/sweet-heisenberg-o9jqx?file=/src/App.js
https://codesandbox.io/s/heuristic-dew-b2h23?file=/src/child.machine.js:477-491 (comment L25 → uncomment L26 → click the button)

davidkpiano:
Not sure if this makes sense. You get the same type of error if you pass `undefined` as a reducer to the built-in `useReducer(undefined, ...)` hook.

In that case, you'd want to _prevent rendering_ that component until it has what's necessary for it to run. This is more of a React design concern than an XState problem.

ivorpad:
@davidkpiano That's not exactly true:

<img width="516" alt="Screen Shot 2020-08-27 at 8 05 27 AM" src="https://user-images.githubusercontent.com/52452/91398303-22ef2080-e83c-11ea-83e3-22d7bf46614f.png">

That's how `useService` should behave IMHO.

I _could_ prevent rendering that component but what if, according to my app design, that's not what it's intended?

mattpocock:
This feels like a 'I should be able to conditionally call hooks' discussion. If there are UI elements you want to render, you can pass down the relevant information via props instead of through a service. I think it's reasonable to expect, when you use a hook, it has all the information it needs.

If you need this behaviour, you can write a custom wrapper for useService which checks to see if the service is defined?

Andarist:
[...]

PR #3968, opened 2023-04-15
[v5] `createEmptyActor()`

davidkpiano:
This PR makes it easy to create null actors which only emit `undefined`.

Especially in `@xstate/react`, this makes working with maybe-undefined actors easier:

```jsx
import { createEmptyActor } from 'xstate/actors';

// ...
function Comp({ maybeActor }) {
  const [state, send] = useActor(maybeActor ?? createEmptyActor());

  state?.count; // strongly-typed; state can be undefined
}
```

### DIFF
--- a/.changeset/odd-pigs-brush.md
+++ b/.changeset/odd-pigs-brush.md
@@ -1,0 +1,16 @@
+---
+'xstate': major
+---
+
+The `createNullActor()` function has been added to make it easier to create actors that do nothing ("null" actors). This is useful for testing, or for some integrations such as `useActor(actor)` in `@xstate/react` that require an actor:
+
+```jsx
+import { createNullActor } from 'xstate';
+
+const SomeComponent = (props) => {
+  // props.actor may be undefined
+  const [state, send] = useActor(props.actor ?? createNullActor());
+
+  // ...
+};
+```
--- a/.changeset/odd-pigs-brush.md
+++ b/.changeset/odd-pigs-brush.md
@@ -5,7 +5,7 @@
 The `createEmptyActor()` function has been added to make it easier to create actors that do nothing ("empty" actors). This is useful for testing, or for some integrations such as `useActor(actor)` in `@xstate/react` that require an actor:
 
 ```jsx
-import { createNullActor } from 'xstate';
+import { createEmptyActor } from 'xstate';
 
 const SomeComponent = (props) => {
   // props.actor may be undefined
--- a/.changeset/odd-pigs-brush.md
+++ b/.changeset/odd-pigs-brush.md
@@ -2,7 +2,7 @@
 'xstate': major
 ---
 
-The `createNullActor()` function has been added to make it easier to create actors that do nothing ("null" actors). This is useful for testing, or for some integrations such as `useActor(actor)` in `@xstate/react` that require an actor:
+The `createEmptyActor()` function has been added to make it easier to create actors that do nothing ("empty" actors). This is useful for testing, or for some integrations such as `useActor(actor)` in `@xstate/react` that require an actor:
 
 ```jsx
 import { createNullActor } from 'xstate';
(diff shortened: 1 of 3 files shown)


packages/xstate-react/test/useActor.test.tsx
```diff
@@ -511,4 +512,53 @@ describeEachReactMode('useActor (%s)', ({ render, suiteKey }) => {
...
+
+      try {
+        // @ts-expect-error (state can be undefined)
+        state.count;
+      } catch (_) {}
+
+      if (state) {
+        // another type test... probably a better way to do this
+        ((_count: number) => void 0)(state.count);
+      }
```
davidkpiano, line 530: I already know this is not exactly @Andarist-approved 🙈

davidkpiano:
Addresses #1416

packages/core/src/actors/index.ts
```diff
@@ -63,3 +69,9 @@ export function toActorRef<
     ...actorRefLike
   };
 }
+
+const nullBehavior = fromTransition((_) => undefined, undefined);
```
Andarist, line 73: I see that you took a page from the JavaScript language design :D a null actor that returns undefined? 😉 
```suggestion
const nullBehavior = fromTransition(() => null, null);
```

packages/xstate-react/test/useActor.test.tsx
Andarist, line 530: I'd love to integrate a better way for writing type tests but so far we have been using things like this:
```suggestion
      // @ts-expect-error
      ((_accept: { count: number }) => {})(state);
      ((_accept: { count: number } | undefined) => {})(state);
```

packages/core/src/actors/index.ts
```diff
@@ -63,3 +69,9 @@ export function toActorRef<
...
+
+export function createNullActor(): ActorRef<AnyEventObject, undefined> {
```
Andarist, line 75: q: why do we even need a null actor factory? can't we just export a null actor? 

```ts
export const nullActor = /* #__PURE__ */ interpret(nullBehavior);
```

davidkpiano: Equality checks. We don't want `someNullActor === anotherNullActor`

Andarist: Hm, I don't think that it matters too much - but perhaps to avoid weird issues with people trying to stop this actor etc it's best to create unique "instances".

Andarist, line 75:
```suggestion
export function createEmptyActor(): ActorRef<AnyEventObject, undefined> {
```

.changeset/odd-pigs-brush.md
```diff
@@ -0,0 +1,16 @@
+---
+'xstate': major
+---
+
+The `createNullActor()` function has been added to make it easier to create actors that do nothing ("null" actors). This is useful for testing, or for some integrations such as `useActor(actor)` in `@xstate/react` that require an actor:
```
Andarist, line 5:
```suggestion
The `createEmptyActor()` function has been added to make it easier to create actors that do nothing ("empty" actors). This is useful for testing, or for some integrations such as `useActor(actor)` in `@xstate/react` that require an actor:
```

.changeset/odd-pigs-brush.md
```diff
@@ -0,0 +1,16 @@
...
+
+```jsx
+import { createNullActor } from 'xstate';
```
Andarist, line 8:
```suggestion
import { createEmptyActor } from 'xstate';
```

.changeset/odd-pigs-brush.md
```diff
@@ -0,0 +1,16 @@
...
+
+const SomeComponent = (props) => {
+  // props.actor may be undefined
+  const [state, send] = useActor(props.actor ?? createNullActor());
```
Andarist, line 12:
```suggestion
  const [state, send] = useActor(props.actor ?? createEmptyActor());
```

PR #3969, opened 2023-04-17
[v5] `actor.inspect()`

davidkpiano:
This PR adds the `actor.inspect(observer)` method to actors. This is meant for devTools/inspection (as the name implies) and observes the _internal state_ of an actor, as opposed to `.subscribe(...)`, which observes the _snapshot_ of an actor.

Additionally, inspectors are _not_ automatically removed when a machine stops; not sure if this should be the desired behavior yet.

It also adds some tests for devTools using `.inspect` and ensuring that devTools are part of the `system` instead of the individual actors.


packages/core/src/dev/index.ts
```diff
@@ -47,26 +47,10 @@ function getDevTools(): DevInterface | undefined {
   return undefined;
 }
 
-export function registerService(service: AnyInterpreter) {
```
davidkpiano, line 50: This was unused

packages/core/src/interpreter.ts
```diff
@@ -141,7 +142,11 @@ export class Interpreter<
     const { clock, logger, parent, id, systemId } = resolvedOptions;
     const self = this;
 
-    this.system = parent?.system ?? createSystem();
+    this.system =
+      parent?.system ??
+      createSystem({
+        devTools: resolvedOptions.devTools
```
davidkpiano, line 148: This was added; `devTools` are relevant to the system, not just the actor

packages/core/src/interpreter.ts
```diff
@@ -208,6 +213,10 @@ export class Interpreter<
       observer.next?.(snapshot);
     }
 
+    for (const inspector of this.inspectors) {
+      inspector.next?.(this._state);
```
davidkpiano, line 217: Biggest difference: `this._state` is received, not `snapshot`

packages/core/src/interpreter.ts
```diff
@@ -270,6 +279,36 @@ export class Interpreter<
     };
   }
 
+  public inspect(
```
Andarist, line 282: Hm, shouldn't this be part of the system and not the actor? 🤔

packages/core/src/interpreter.ts
```diff
@@ -270,6 +279,36 @@ export class Interpreter<
...
+    observer: Observer<InternalStateFrom<TBehavior>>
+  ): Subscription;
+  public inspect(
+    nextListener?: (state: InternalStateFrom<TBehavior>) => void,
+    errorListener?: (error: any) => void,
+    completeListener?: () => void
+  ): Subscription;
+  public inspect(
```
Andarist, line 290: Like we discussed before - I think this should not listen for (internal) states but rather for tailored events that would inform the inspect about various things that happen to the actor. I assume this is a draft right now, and we are waiting on Gavin's to finalize the "protocol" part for this?

davidkpiano:
Will redo this system-first

PR #3977, opened 2023-04-23
[v5] Awaitable actors

davidkpiano:
This PR makes actors thenable; awaiting an actor will wait for output.

```ts
    initial: 'a',
    states: {
      a: {
        after: {
          10: 'b'
        }
      },
      b: {
        type: 'final',
        output: { foo: 'bar' }
      }
    }
  });

  const output = await interpret(machine).start();

  expect(output).toEqual({ foo: 'bar' });
```


packages/core/src/interpreter.ts
```diff
@@ -288,6 +288,31 @@ export class Interpreter<
     return this;
   }
 
+  public getOutput() {
+    return this.behavior.getOutput?.(this._state);
+  }
+
+  // thenable
+  public then(
```
Andarist, line 296: I don't "feel" this feature. If anything, I'd probably prefer smth like `toPromise` over this. Sure, we can make things "thenable" - should we though? The actor is not a promise and despite the fact that JS allows us to do this, it feels rather weird. 

Since we already have `waitFor` I also don't necessarily feel the need for `toPromise` either. It's already simple to do this. Yet a different option would be to just have `actor.waitFor`. 

Our `ActorLogic` interface starts to be fatter than a turkey on Thanksgiving 😬

davidkpiano: The current problem is there's no easy way to have "wait for actor to complete" for _any_ actor logic, unless you're okay with an optional selector here:

```ts
// single argument - wait for actor to be done
const output = await waitFor(someActor);

// current behavior
// PROBLEM: not sure what should be returned
const output = await waitFor(someActor, state => ???);
```

What do you think would be a good workaround with `waitFor`?

davidkpiano: Ping @Andarist - need a good solution to this use-case:

```js
// get actor output when the actor is completed
const output = await fromPromise(actor);
```

Can't do this with `waitFor()`.

davidkpiano:
Superseded by #4198

PR #3985, opened 2023-04-26
Rewrite a few tests

Andarist:
This just ports test changes from https://github.com/statelyai/xstate/pull/3912 . I feel like `done` often makes it harder to read and debug tests, if something doesnt get called then the test runner has to first wait for the timeout before failing the test case - faster fails -> faster dev cycle -> 💰

### DIFF
--- a/packages/core/test/interpreter.test.ts
+++ b/packages/core/test/interpreter.test.ts
@@ -1097,87 +1097,98 @@ describe('interpreter', () => {
   });
 
   describe('.start()', () => {
-    const startMachine = createMachine({
-      id: 'start',
-      initial: 'foo',
-      states: {
-        foo: {
-          initial: 'one',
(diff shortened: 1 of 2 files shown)


packages/core/test/interpreter.test.ts
```diff
@@ -1097,87 +1097,98 @@ describe('interpreter', () => {
...
+          foo: {},
+          bar: {}
+        }
       });
-
-      startService.subscribe((state) => {
-        expect(state.matches('bar')).toBeTruthy();
-        done();
+      const actor = interpret(machine, {
+        state: State.from('bar', undefined, machine)
```
davidkpiano, line 1146: Note to self: this needs to be rewritten with `machine.createState`

davidkpiano (approved):

PR #3986, opened 2023-04-26
Improve control over console logs in tests

Andarist:
I find our test output to be spammy and at times I feel like it's likely that some of the logged messages indicate logical errors in tests. I yet have to verify that - this PR lays out a foundation for deliberate console outputs.

With this PR we will be **forced** to either use `expect(console.warn).toMatchMockCallsInlineSnapshot()` (and similar) or `clearConsoleMocks()` (although that should be done *rarely*, I only used this in our SCXML tests). This should force us to be thoughtful about the logged things and to take appropriate action for each logged information

### DIFF
--- a/packages/core/src/actors/promise.ts
+++ b/packages/core/src/actors/promise.ts
@@ -76,12 +76,14 @@ export function fromPromise<T>(
 
       resolvedPromise.then(
         (response) => {
+          // TODO: remove this condition once dead letter queue lands
           if ((self as any)._state.status !== 'active') {
             return;
           }
           self.send({ type: resolveEventType, data: response });
         },
         (errorData) => {
+          // TODO: remove this condition once dead letter queue lands
           if ((self as any)._state.status !== 'active') {
             return;
           }
--- a/packages/core/src/actors/promise.ts
+++ b/packages/core/src/actors/promise.ts
@@ -76,9 +76,15 @@ export function fromPromise<T>(
 
       resolvedPromise.then(
         (response) => {
+          if ((self as any)._state.status !== 'active') {
+            return;
+          }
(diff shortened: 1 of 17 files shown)


packages/core/test/eventDescriptors.test.ts
```diff
@@ -251,11 +251,43 @@ describe('event descriptors', () => {
         .transition(undefined, { type: 'event.foo.bar.first.second' })
         .matches('success')
     ).toBeFalsy();
+
+    expect(console.warn).toMatchMockCallsInlineSnapshot(`
```
Andarist, line 255: For example, this output shows that we log **three** things for a single (!) `machine.transition` call. This should likely be improved, it seems to be too excessive.

davidkpiano (approved):

packages/core/src/actors/promise.ts
```diff
@@ -76,9 +76,15 @@ export function fromPromise<T>(
 
       resolvedPromise.then(
         (response) => {
+          if ((self as any)._state.status !== 'active') {
+            return;
+          }
```
Andarist, line 81: This is obviously a hack but also... it's the only way, right now, to access your own private state. One could *abuse* `getPersistedState` to do that... but I don't think it's a semantically good idea :p

Alternative approach would be to rely on the dead letter queue but it's not a thing (at least not yet)

Andarist: For context, it avoid this warning:
https://github.com/statelyai/xstate/blob/ed4ed9fc9a72de453dcde60bcd1b17dfd1410be3/packages/core/src/interpreter.ts#L420-L435

davidkpiano (approved):

PR #3987, opened 2023-04-27
[v5] Examples

davidkpiano:
This PR converts some examples to XState v5 beta.


examples/react-7guis-counter/package.json
```diff
@@ -6,10 +6,10 @@
     "serve": "vite preview"
   },
   "dependencies": {
-    "@xstate/react": "^1.6.3",
+    "@xstate/react": "^4.0.0-beta.3",
```
Andarist, line 9: That's likely a mistake done other packages as well - all of those ranges should use *exact* version while in the prerelease mode
```suggestion
    "@xstate/react": "4.0.0-beta.3",
```

PR #3988, opened 2023-04-27
[v5] Dead-letter queue

davidkpiano:


packages/core/src/system.ts
```diff
@@ -34,6 +37,17 @@ export function createSystem<T extends ActorSystemInfo>(): ActorSystem<T> {
 
       keyedActors.set(systemId, actorRef);
       reverseKeyedActors.set(actorRef, systemId);
+    },
+    get deadLetters() {
```
Andarist, line 41: I would prefer not to have getters with side-effects - and for that matter, i think it's best to avoid getters (they obscure the nature of a thing, when i debug and read code like `smth.foo` I don't anticipate that behind this banana is a gorilla holding the banana and the entire jungle)

packages/core/src/deadLetter.ts
```diff
@@ -0,0 +1,19 @@
...
+    type: 'deadLetter';
+    event: AnyEventObject;
+  },
+  any
+>((state, event) => {
+  if (event.type === 'deadLetter') {
+    state.push({
+      event: event.event
+    });
+  }
```
Andarist, line 16: This is basically a memory leak. Those dead letters are never cleared.

packages/core/src/system.ts
```diff
@@ -34,6 +37,17 @@ export function createSystem<T extends ActorSystemInfo>(): ActorSystem<T> {
...
+      if (!deadLetters) {
+        deadLetters = interpret(deadLettersBehavior, {
+          _system: system
+        });
+
+        deadLetters.start();
```
Andarist, line 47: It's started but isn't ever stopped. Shouldn't it become a regular child of the root to behave in the same way as its other child actors?

packages/core/src/deadLetter.ts
Andarist: in Akka dead letters are not part of the state, but there is a concept of an event stream that is available on the system, so u can subscribe to it to be notified about all events (dead letters included)

packages/core/src/actions/send.ts
```diff
@@ -134,12 +134,32 @@ export function send<
...
+                internal: false,
+                to: undefined,
+                delay: undefined
+              },
+              execute: (actorCtx) => {
+                actorCtx.system.deadLetters.send({
+                  type: 'deadLetter',
+                  event: resolvedEvent
+                });
+              }
```
Andarist, line 157: could you specify the exact role of this dead letter thing? for instance, i don't think it should be send action's role to send it there. Why don't we put it in the `actor.send` method?

davidkpiano:
🧹 Going to redo this one

PR #3990, opened 2023-04-29
[v5] `systemId` for `spawn()`

davidkpiano:
This PR adds `systemId` to the 2nd argument of `spawn(logic, options?)`.


.vscode/launch.json
```diff
@@ -6,13 +6,7 @@
       "request": "launch",
       "name": "Jest Current File",
       "program": "${workspaceFolder}/node_modules/.bin/jest",
-      "args": [
-        "${file}",
-        "--config",
-        "jest.config.js",
-        "--no-cache",
-        "--testTimeout=2147483647"
```
davidkpiano, line 14: Sorry @Andarist, I think this was a bad idea. With this test timeout, any failing test will now hang for a long time instead of timing out, making it very difficult to determine which test is failing.

.changeset/great-kangaroos-confess.md
```diff
@@ -0,0 +1,19 @@
+---
+'xstate': major
```
Andarist, line 2:
```suggestion
'xstate': patch
```

.vscode/launch.json
Andarist: Let's revert this change from here. I opened a PR [here](https://github.com/statelyai/xstate/pull/4024) to add a second launch script - there will be one with the timeout and one without it.

Andarist (approved):

.changeset/great-kangaroos-confess.md
with-heart: Wouldn't this technically be `minor`?

Also I've been wondering about how this works with the `next` branch. When I ran `changesets add` for #3997, it just assumed it was a `major` change for some reason.

Do the different versions have any effect when it's in prerelease mode?

Andarist: > Wouldn't this technically be minor?

That works too. I suggest `patch` here since this was an oversight on our part. This kinda should be supported from the start. I don't mind using minor though - we'll have to clean up this changeset before releasing final v5 anyway (this particular changeset is only relevant for the prerelease)

> When I ran changesets add for https://github.com/statelyai/xstate/pull/3997, it just assumed it was a major change for some reason.

That's weird, it should allow u to select the type.

> Do the different versions have any effect when it's in prerelease mode?

It's just for grouping them together. And all pending changesets are used to determine the next version - if we didn't have any major changesets then we would end up with `xstate@4.38.0-beta.11` (or similar).

with-heart: Thanks for the explanation! Makes sense 👍

.changeset/great-kangaroos-confess.md
```diff
@@ -0,0 +1,19 @@
...
+  context: ({ spawn }) => ({
+    actorRef: spawn(
+      createMachine({
+        // ...
+      }),
+      { systemId: 'actorRef' }
+    )
+  })
+});
+```
```
with-heart, line 19: Showing how to reference the actor by `systemId` would add a lot of value here

davidkpiano: Will do this in the docs

PR #3991, opened 2023-04-29
[v5] Remove `actor.onDone(...)`

davidkpiano:
This PR removes `actor.onDone(...)` in favor of `actor.subscribe({ complete: () => { ... })`


packages/core/src/interpreter.ts
```diff
@@ -99,6 +99,7 @@ export class Interpreter<
   private delayedEventsMap: Record<string, unknown> = {};
 
   private observers: Set<Observer<SnapshotFrom<TBehavior>>> = new Set();
+  private stopListeners: Set<() => void> = new Set();
```
Andarist, line 102: was adding this intentional? currently, there is no way to add anything to this set

.vscode/launch.json
```diff
@@ -10,8 +10,8 @@
         "${file}",
         "--config",
         "jest.config.js",
-        "--no-cache",
-        "--testTimeout=2147483647"
+        "--no-cache"
+        // "--testTimeout=2147483647"
```
Andarist, line 14: this has to be reverted, which i guess just means that u need to merge `next` to this branch

Andarist (commented):
this PR needs a changeset

packages/core/src/interpreter.ts
davidkpiano: It was not IIRC

Andarist (approved):

PR #3992, opened 2023-05-02
[@xstate/graph] [@xstate/solid] Fix `xstate/src` imports

davidkpiano:


.changeset/ninety-paws-roll.md
```diff
@@ -0,0 +1,6 @@
+---
+'@xstate/graph': patch
+'@xstate/solid': patch
```
Andarist, line 3: This one was only in the test file so it doesn't need to be released
```suggestion
```

Issue #3996, opened 2023-05-03
Action Groups


with-heart:
### Discussed in https://github.com/statelyai/xstate/discussions/3554

<div type='discussions-op-text'>

<sup>Originally posted by **with-heart** August 28, 2022</sup>
Hi yall! 👋 

*This discussion stems from a conversation with @Andarist regarding https://github.com/statelyai/xstate-tools/pull/207*

One major frustration I've faced when using `xstate` is the lack of type-safety when calling action functions from within `pure`. 

`pure` takes a function (`getActions`) that returns an array of actions to execute. Unfortunately, there's no way to inject the generic types `TContext`, `TEvent`, etc. that `pure` receives in the returned array, which means action function calls inside `pure` use their default generic types:

```ts
{
  actions: {
    // pure<TContext, TEvent> (or an event type derived from `eventsCausingActions` 
    // if using typegen)
    doSomething: pure(() => [
      // assign<unknown, EventObject>
      assign(),
      // send<unknown, EventObject>
      send(),
    ]);
  }
}
```

Since this makes `pure` difficult to work with, I'd like to propose an alternative that I've been thinking of as "action bundles". 

Instead of calling action functions within `pure`, we define the actions individually and then "bundle" them together by referring to them from another action:

```ts
{
  actions: {
    doSomething: ['assignThing', 'sendSomething'],
    // assign<TContext, TEvent>
    assignThing: assign(),
    // send<TContext, TEvent>
    sendSomething: send(),
  }
}
```

Much like the previous example using `pure`, executing `doSomething` will then execute `assignThing` and `sendSomething`. We get the same behavior but we retain type-safe awesomeness!

This also makes it much easier to typegen. https://github.com/statelyai/xstate-tools/pull/207 is an implementation for detecting array `type` strings returned in `pure` arrays, but since `getActions` is a function and a function return can take many shapes, it'd be very difficult for us to analyze all of the possible AST structures.

If actions can be defined as simple arrays of action `type` strings, this becomes relatively easy.

---

If you're familiar with `xstate`, you might be thinking to yourself "but @with-heart, you can already define actions as arrays of `type` strings using `entry`/`exit`/`actions` in the machine config, and those are all type-safe". And you'd be correct!

The unfortunate thing about that method is that sometimes you end up in situations where you have actions that should always be paired (executing one before or after the other, etc.), so you're stuck with the mental overhead of needing to remember those types of details. It works, but with enough actions, it's pretty frustrating to maintain.</div>

[...]

PR #3997, opened 2023-05-03
[v5] implement action groups

with-heart:
This PR implements #3996, enabling actions to be implemented as an _action group_. (replaces #3561)

An _action group_ is a named array of actions. When the action group is executed, each action in the group is executed in order:

```js
import { createMachine, interpret, log } from 'xstate';

  {
    // reference the action group by name (`someGroup`)
    entry: 'someGroup'
  },
  {
    actions: {
      // executes `action1`, then `action2`
      someGroup: ['action1', 'action2'],
      action1: log('action1'),
      action2: log('action2')
    }
  }
);

interpret(machine).start(); // logs "action1", then "action2"
```

Action groups allow us to avoid error-prone repetition of actions, instead defining the group once and reusing it anywhere—like a single source of truth for the algorithm the group represents:

```diff
 const machine = createMachine(
   {
     context: { count: 0 },
     on: {
-      // increment count, then print it
-      incrementClick: { actions: ['incrementCount', 'printCount'] },
+       incrementClick: { actions: 'increment' },
-      // increment count, then print it
-      // oops! we accidentally put the actions in the wrong order
-      tick: { actions: ['printCount', 'incrementCount'] },
+      tick: { actions: 'increment' }
     }
   },
   {
     actions: {
+      // increment count, then print it. single source of truth for the `increment` algorithm
+      increment: ['incrementCount', 'printCount'],
       incrementCount: assign({
         count: ({ context }) => context.count + 1
       }),
       printCount: log(({ context }) => `Count: ${context.count}`)
     }
   }
 );
```

Action groups can reference other action groups by name. The referenced group's actions will be executed in order from the point of reference—like spreading the referenced group's actions in the group:

```js
  {
    entry: 'initialize'
  },
  {
    actions: {
      // executes `load` group actions, then `listen` group actions
      initialize: ['load', 'listen'],
      load: ['loadConfig', 'loadData'],
      listen: ['startApp', 'listenOnPort']
    }
  }
);

interpret(machine).start();
// actions: (load) loadConfig -> loadData -> (listen) startApp -> listenOnPort
```

With a mix of actions, action groups, and action group references, we can compose our algorithms in flexible and reusable ways:

```js
import { assign, createMachine, log } from 'xstate';

  {
    entry: 'initialize',
    exit: 'terminate',
    on: {
      timeout: { actions: 'reconnect' },
      reload: { actions: 'reload' }
    }
  },
  {
    actions: {
      initialize: ['load', 'connect'],
      terminate: ['disconnect', 'save', 'exitProgram'],
      reconnect: ['disconnect', 'connect'],
      reload: ['disconnect', 'save', 'initialize']
    }
  }
);
```

I feel like there's something powerful to thinking about action groups as composable algorithms, though I'm not entirely sure what that is yet, so I'm excited for this to land so more people can play with it!

### DIFF
--- a/.changeset/wild-roses-sort.md
+++ b/.changeset/wild-roses-sort.md
@@ -1,0 +1,103 @@
+---
+'xstate': major
+---
+
+Actions can now be implemented as an _action group_.
+
+An _action group_ is a named array of actions. When the action group is executed, each action in the group is executed in order:
+
+```js
+import { createMachine, interpret, log } from 'xstate';
+
+const machine = createMachine(
+  {
+    // reference the action group by name (`someGroup`)
+    entry: 'someGroup'
+  },
+  {
+    actions: {
+      // executes `action1`, then `action2`
+      someGroup: ['action1', 'action2'],
+      action1: log('action1'),
+      action2: log('action2')
+    }
+  }
+);
+
+interpret(machine).start(); // logs "action1", then "action2"
+```
+
+Action groups allow us to avoid error-prone repetition of actions, instead defining the group once and reusing it anywhere—like a single source of truth for the algorithm the group represents:
+
+```diff
+ const machine = createMachine(
+   {
+     context: { count: 0 },
+     on: {
+-      // increment count, then print it
+-      incrementClick: { actions: ['incrementCount', 'printCount'] },
++       incrementClick: { actions: 'increment' },
+-      // increment count, then print it
+-      // oops! we accidentally put the actions in the wrong order
+-      tick: { actions: ['printCount', 'incrementCount'] },
++      tick: { actions: 'increment' }
+     }
+   },
+   {
+     actions: {
++      // increment count, then print it. single source of truth for the `increment` algorithm
++      increment: ['incrementCount', 'printCount'],
+       incrementCount: assign({
+         count: ({ context }) => context.count + 1
+       }),
+       printCount: log(({ context }) => `Count: ${context.count}`)
+     }
+   }
+ );
+```
+
+Action groups can reference other action groups by name. The referenced group's actions will be executed in order from the point of reference—like spreading the referenced group's actions in the group:
+
+```js
+const machine = createMachine(
+  {
+    entry: 'initialize'
+  },
+  {
+    actions: {
+      // executes `load` group actions, then `listen` group actions
+      initialize: ['load', 'listen'],
+      load: ['loadConfig', 'loadData'],
+      listen: ['startApp', 'listenOnPort']
+    }
+  }
+);
+
+interpret(machine).start();
+// actions: (load) loadConfig -> loadData -> (listen) startApp -> listenOnPort
+```
+
+With a mix of actions, action groups, and action group references, we can compose our algorithms in flexible and reusable ways:
+
+```js
+import { assign, createMachine, log } from 'xstate';
+
+const machine = createMachine(
+  {
+    entry: 'initialize',
+    exit: 'terminate',
+    on: {
+      timeout: { actions: 'reconnect' },
+      reload: { actions: 'reload' }
+    }
+  },
+  {
+    actions: {
+      initialize: ['load', 'connect'],
+      terminate: ['disconnect', 'save', 'exitProgram'],
+      reconnect: ['disconnect', 'connect'],
+      reload: ['disconnect', 'save', 'initialize']
+    }
+  }
+);
+```
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -69,6 +69,17 @@ export function resolveActionObject(
         return [state, a];
       }
     );
+  } else if (isArray(dereferencedAction)) {
+    return createDynamicAction(
+      { type: ActionTypes.ActionGroup, params: actionObject.params ?? {} },
+      (_event, { state }) => {
+        const action: BaseActionObject = {
+          type: actionObject.type,
+          params: { actions: dereferencedAction }
+        };
+        return [state, action];
+      }
+    );
   } else if (dereferencedAction) {
     return dereferencedAction;
   } else {
--- a/packages/core/src/actions.ts
+++ b/packages/core/src/actions.ts
@@ -70,12 +70,21 @@ export function resolveActionObject(
       }
     );
   } else if (isArray(dereferencedAction)) {
+    const params = actionObject.params ?? {};
     return createDynamicAction(
-      { type: ActionTypes.ActionGroup, params: actionObject.params ?? {} },
+      { type: ActionTypes.ActionGroup, params },
       (_event, { state }) => {
         const action: BaseActionObject = {
           type: actionObject.type,
-          params: { actions: dereferencedAction }
+          params: {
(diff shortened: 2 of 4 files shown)


.changeset/wild-roses-sort.md
with-heart, line 1: I was worried this might be too long but after thinking about it, I think this is probably the minimum explanation I can come up with to really get to the heart of what's cool about this feature

packages/core/src/actions.ts
```diff
@@ -69,6 +69,17 @@ export function resolveActionObject(
         return [state, a];
       }
     );
+  } else if (isArray(dereferencedAction)) {
+    return createDynamicAction(
+      { type: ActionTypes.ActionGroup, params: actionObject.params ?? {} },
```
with-heart, line 74: Should we even bother with `actionObject.params` here? I had trouble wrapping my mind around this part. Would there be a reason we might want them available?

packages/core/src/types.ts
```diff
@@ -1086,7 +1097,8 @@ export enum ActionTypes {
   ErrorPlatform = 'error.platform',
   ErrorCustom = 'xstate.error',
   Pure = 'xstate.pure',
-  Choose = 'xstate.choose'
+  Choose = 'xstate.choose',
+  ActionGroup = 'xstate.group'
```
with-heart, line 1101: I don't really understand the difference between `ActionTypes` and `actionTypes`. Is this the right place to add this?

packages/core/src/actions.ts
davidkpiano: Hmm... I'd say so. Example:

```js
entry: {
  type: 'sendMessage',
  message: 'hello'
}
```
And...
```js
actions: {
  sendMessage: [
    'sendEmail',
    'notifySent'
  ],
  sendEmail: ({ action }) => {
    sendEmail(action.params.message);
  },
  notifySent: ({ action }) => {
    toast.send(`Message sent: ${action.params.message}`);
  }
}
```

with-heart: OH wow okay! Thanks for the example!

I didn't realize there was now an `action` key on action functions. That's beautiful and it makes sense why we'd want params here.

I added a (failing) test to make sure we're handling this and I'll work on getting that resolved.

packages/core/src/types.ts
with-heart: I'll leave it alone for now and if it turns out it's in the wrong spot, we can move it!

packages/core/src/types.ts
```diff
@@ -212,6 +212,58 @@ export type Actions<
...
+      any, // TODO: this should receive something like `Cast<Prop<TIndexedActions, K>, ParameterizedObject>`, but at the moment builtin actions expect Resolved*Action here and this should be simplified somehow
+      any
+    >
+  | ActionFunction<
+      TContext,
+      TExpressionEvent,
+      ParameterizedObject, // TODO: when bringing back parametrized actions this should accept something like `Cast<Prop<TIndexedActions, K>, ParameterizedObject>`. At the moment we need to keep this type argument consistent with what is provided to the fake callable signature within `BaseDynamicActionObject`
+      TEvent
+    >
+>;
```
with-heart, line 265: Adding action groups to `MachineImplementationsActions` was getting pretty messy so I pulled this out as a separate type.

packages/core/src/actions.ts
```diff
@@ -69,6 +69,26 @@ export function resolveActionObject(
...
+          type: actionObject.type,
+          params: {
+            actions: toActionObjects(dereferencedAction).map((object) => ({
+              ...object,
+              params: {
+                ...object.params,
+                ...params
+              }
+            }))
+          }
```
with-heart, line 87: Since each action needs to receive `params`, I just resolved the action objects for each here and then mapped them to include the params.

Is that fine?

packages/core/src/actions.ts
```diff
@@ -69,6 +69,26 @@ export function resolveActionObject(
...
+      (_event, { state }) => {
+        const action: BaseActionObject = {
+          type: actionObject.type,
+          params: {
+            actions: toActionObjects(dereferencedAction).map((object) => ({
+              ...object,
+              params: {
+                ...object.params,
+                ...params
+              }
```
with-heart, line 85: Is this fine? It would allow outer `params` to override inner `params`:

```ts
  {
    entry: {
      type: "sayHello",
      params: { location: "entry" }
    },
    initial: "state1",
    states: {
      state1: {
        always: {
          target: "state2",
          actions: { type: "sayHello", params: { location: "state1" } }
        }
      },
      state2: {
        entry: "sayHello"
      }
    }
  },
  {
    actions: {
      sayHello: [{ type: "greet", params: { location: "somewhere" } }],
      greet: log(({ action }) => `Hello from ${action.params!.location}!`)
    }
  }
);

interpret(machine).start()

// Hello from entry!
// Hello from state1!
// Hello from somewhere!
```

Doubt this will be a common scenario but does this feel right?

packages/core/test/actions.test.ts
```diff
@@ -3664,3 +3664,224 @@ it('should call a referenced action responding to an initial raise with updated
...
+      }
+    );
+
+    interpret(machine).start();
+
+    expect(handler1).toHaveBeenCalledWith({ message: 'Hello world!' });
+    expect(handler2).toHaveBeenCalledWith({
+      message: 'Hello world!',
+      level: 'info'
+    });
```
with-heart, line 3837: This has an implicit test for `params` overriding. Should we be more explicit with a separate test?

packages/core/src/actions.ts
davidkpiano: This makes sense to me 👍

packages/core/test/actions.test.ts
davidkpiano: Might not be a need because if the behavior changes, this test will fail.

cc. @Andarist for his thoughts

packages/core/src/actions.ts
davidkpiano: Seems fine to me 👍

davidkpiano:
🧹 I think that with `enqueueActions(…)` and v6 ideas for single-function actions (basically `enqueueActions(…)` but by default), action groups will be made redundant, since it would be easier to use functions as "groups" instead:

```ts
// TENTATIVE API
const machine = setup({
  actions: {
    actionGroup: (_, x) => {
      x.action({ type: 'action1' });
      x.action({ type: 'action2' });
    },
    action1: () => { ... },
    action2: () => { ... },
  }
}).createMachine({
  entry: 'someGroup'
});
```

with-heart:
@davidkpiano that sounds lovely! Do you have v6 ideas documented anywhere publicly? I'd love to see what you're thinking

davidkpiano:
> @davidkpiano that sounds lovely! Do you have v6 ideas documented anywhere publicly? I'd love to see what you're thinking

Not yet, but once I set up a branch/draft PR for it I'll let you know.

with-heart:
Thank you! ❤️

PR #3999, opened 2023-05-04
[v5] Data actor

davidkpiano:
This PR adds `fromData(...)`, which makes it easy to work with plain data:

```js
const dataActor = interpret(fromData({ count: 0 }));

dataActor.send({
  type: 'set',
  data: { count: dataActor.getSnapshot().count + 1 }
});

dataActor.getSnapshot();
// { count: 1 }
```

### DIFF
--- a/packages/core/src/actors/data.ts
+++ b/packages/core/src/actors/data.ts
@@ -1,0 +1,43 @@
+import { toSCXMLEvent } from '..';
+import { ActorBehavior } from '../types';
+
+export function fromData<T>(initialData: T): ActorBehavior<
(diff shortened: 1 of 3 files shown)


packages/core/src/actors/data.ts
```diff
@@ -0,0 +1,39 @@
+import { ActorBehavior } from '../types';
```
Andarist, line 1:
```suggestion
import { ActorBehavior } from '../types.ts';
```

davidkpiano:
This is more of a user-land abstraction

PR #4000, opened 2023-05-06
[v5] `useMachine()` and `useInterpret()` for any actor

davidkpiano:


davidkpiano:
@Andarist I'll need some help fixing the types here 🥺

Andarist:
@davidkpiano is this still relevant? I think that lately some of those changes landed on `next` and this PR now has a lot of conflicts. Could you sync it with `next` and assess what's left to do for this unification?

davidkpiano:
This is no longer relevant, as it is mostly done (just some cleanup/renaming to do in `next`)

PR #4002, opened 2023-05-08
[v5] Private events

davidkpiano:


packages/core/src/types.ts
```diff
@@ -1826,6 +1826,7 @@ export interface ActorContext<
   TSystem extends ActorSystem<any> = ActorSystem<any>
 > {
   self: ActorRef<TEvent, TSnapshot>;
+  enqueue: (event: TEvent) => void;
```
Andarist, line 1829: I think I like the idea of a privileged send function like this one. I also think that ideally, we should be able to route `done.invoke.*` events etc through a privileged channel like this.

davidkpiano: Great; I'll continue on with this

davidkpiano:
Doing some cleanup 🧹

PR #4006, opened 2023-05-10
[v5] Hook renaming

davidkpiano:
This PR renames/repurposes the following hooks:

- `useActorRef` is introduced, which returns an `ActorRef` from actor logic:
  ```ts
  const actorRef = useActorRef(machine, { ... });
  const anotherActorRef = useActorRef(fromPromise(...));
  ```
- ~~`useMachine`~~ is deprecated in favor of `useActor`, which works with machines and any other kind of logic
  ```diff
  -const [state, send] = useMachine(machine);
  +const [state, send] = useActor(machine);
  const [state, send] = useActor(fromTransition(...));
  ```
- ~~`useSpawn`~~ is removed in favor of `useActorRef`
  ```diff
  -const actorRef = useSpawn(machine);
  +const actorRef = useActorRef(machine);
- The previous use of `useActor(actorRef)` is now replaced with just using the `actorRef` directly, and with `useSelector`:
  ```diff
  -const [state, send] = useActor(actorRef);
  +const state = useSelector(actorRef, s => s);
  // actorRef.send(...)
  ```


packages/xstate-react/src/useActor.ts
```diff
@@ -1,16 +1,37 @@
...
+} from 'xstate';
+import { useIdleInterpreter } from './useActorRef.ts';
+
+function identity<T>(a: T): T {
+  return a;
+}
+
+const isEqual = (prevState: AnyState, nextState: AnyState) => {
+  return prevState === nextState || nextState.changed === false;
+};
```
Andarist, line 19: Note that this one doesn't belong here - yet another reason why we need to get rid of `.changed` 😉

Andarist: (that will be done in a separate batch of work though, let's leave it as is here)

packages/xstate-react/test/useSelector2.test.tsx
```diff
@@ -0,0 +1,556 @@
+import { act, fireEvent, screen } from '@testing-library/react';
```
Andarist, line 1: This file can now be removed, right? Or does it require "migration"?

packages/xstate-react/src/createActorContext.ts
```diff
@@ -96,7 +95,8 @@ export function createActorContext<TMachine extends AnyStateMachine>(
 
   function useActor() {
```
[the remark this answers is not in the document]
davidkpiano: Agree with removing it. Also, would this be feasible?

```ts
const value = SomeContext.useSelector(state => state.context.obj);

const { send } = SomeContext.actorRef; // no hook
```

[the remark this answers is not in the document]
Andarist: > Agree with removing it. Also, would this be feasible?

No, context is a singleton but we expect to get access to the instance scoped to the wrapped subtree. There is no single `SomeContext.actorRef`, that's created within the Provider.

I don't think this is that big of a problem though, it looks pretty much the same:
```diff
-const { send } = SomeContext.actorRef; // no hook
-const { send } = SomeContext.useActorRef(); // hook
```

Note that you might not want to destructure `send` like this because it might rely on `this`.

packages/xstate-react/test/useSelector2.test.tsx
Andarist: i moved the worthwhile ones to other files and removed this one

packages/xstate-react/src/useActor.ts
```diff
@@ -1,16 +1,37 @@
...
+
+export function useActor<TBehavior extends AnyActorBehavior>(
+  behavior: TBehavior,
+  options: InterpreterOptions<TBehavior> = {}
+): [
+  SnapshotFrom<TBehavior>,
+  ActorRefFrom<TBehavior>['send'],
+  ActorRefFrom<TBehavior>
+] {
+  const actorRef = useIdleInterpreter(behavior, options as any);
```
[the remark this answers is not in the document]
davidkpiano: Done

Andarist (approved):

PR #4007, opened 2023-05-10
Removed the ability to pass a factory function as argument to `useMachine` and `useInterpret`

Andarist:


.changeset/fifty-dingos-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/react': patch
```
Andarist, line 2:
```suggestion
'@xstate/react': major
```

davidkpiano (approved):

PR #4009, opened 2023-05-11
Update Vue 2 recipe

amypellegrini:
Added context on XState usage depending on the version of Vue.

### DIFF
--- a/docs/recipes/vue.md
+++ b/docs/recipes/vue.md
@@ -1,5 +1,12 @@
 # Usage with Vue
 
+Usage of Xstate in a Vue application may vary depeding on which version of Vue your application is running. This page focuses on Vue 2 only. To see how to use Xstate with Vue 3 check the documentation for the Xstate [`@xstate/vue` package](../packages/xstate-vue).
+
+There are two ways you can use Xstate with Vue 2:
+
+1. Using the `useMachine` hook provided by the [`xstate-vue2` package](https://github.com/ChrisShank/xstate-vue2) (3rd-party) plugin;
+2. Using Xstate `interpret` utility to create a service and inject it into your app.
+
 ::: tip
 If you want to use the Vue Composition API, we recommend using the following packages:
 
--- a/docs/recipes/vue.md
+++ b/docs/recipes/vue.md
@@ -1,11 +1,11 @@
 # Usage with Vue
 
-Usage of Xstate in a Vue application may vary depeding on which version of Vue your application is running. This page focuses on Vue 2 only. To see how to use Xstate with Vue 3 check the documentation for the Xstate [`@xstate/vue` package](../packages/xstate-vue).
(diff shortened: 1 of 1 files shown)


docs/recipes/vue.md
```diff
@@ -1,5 +1,12 @@
 # Usage with Vue
 
+Usage of Xstate in a Vue application may vary depeding on which version of Vue your application is running. This page focuses on Vue 2 only. To see how to use Xstate with Vue 3 check the documentation for the Xstate [`@xstate/vue` package](../packages/xstate-vue).
```
Andarist, line 3:
```suggestion
Usage of XState in a Vue application may vary depending on which version of Vue your application is running. This page focuses on Vue 2 only. To see how to use XState with Vue 3 check the documentation for the XState [`@xstate/vue` package](../packages/xstate-vue).
```

docs/recipes/vue.md
```diff
@@ -1,5 +1,12 @@
...
+
+There are two ways you can use Xstate with Vue 2:
```
Andarist, line 5:
```suggestion
There are two ways you can use XState with Vue 2:
```

docs/recipes/vue.md
```diff
@@ -1,5 +1,12 @@
...
+
+1. Using the `useMachine` hook provided by the [`xstate-vue2` package](https://github.com/ChrisShank/xstate-vue2) (3rd-party) plugin;
+2. Using Xstate `interpret` utility to create a service and inject it into your app.
```
Andarist, line 8:
```suggestion
2. Using XState `interpret` utility to create a service and inject it into your app.
```

Andarist:
I've noticed that it's still in the draft. Is there any other information that you plan to add to this PR?

amypellegrini:
Hi @Andarist yes, I intend to work on the whole page so I think it makes sense to wait until the whole editing is done before merging, instead of having multiple PR's. Unless you think it's better otherwise, just let me know.

docs/recipes/vue.md
```diff
@@ -1,5 +1,12 @@
 # Usage with Vue
 
+Usage of XState in a Vue application may vary depending on which version of Vue your application is running. This page focuses on Vue 2 only. To see how to use XState with Vue 3 check the documentation for the XState [`@xstate/vue` package](../packages/xstate-vue).
+
+There are two ways you can use XState with Vue 2:
+
+1. Using the `useMachine` hook provided by the [`xstate-vue2` package](https://github.com/ChrisShank/xstate-vue2) (3rd-party) plugin;
```
ChrisShank, line 7: Might be worth specifying this is `^2.7`?

amypellegrini:
I made the distinction between using the external plugin or using `interpret` more explicit. I'm not sure if it is possible to make use of the `setup` feature from Vue for the `interpret` example:

https://vuejs.org/api/sfc-script-setup.html

amypellegrini:
I'm not sure if I should add a change set for this, but my best guess is that docs are not versioned since they are not inside the packages folder. Please let me know if I should add a change set, I try to learn more about how the docs are built so I can make PR's easier to merge.

davidkpiano (approved):

PR #4013, opened 2023-05-12
[v5] Rename hooks in other packages

davidkpiano:


packages/xstate-solid/src/createActorRef.ts
```diff
@@ -0,0 +1,21 @@
...
+import { onCleanup } from 'solid-js';
+import { isServer } from 'solid-js/web';
+
+export function createActorRef<TBehavior extends AnyActorBehavior>(
+  behavior: TBehavior,
+  ...[options = {}]: RestParams<TBehavior>
+): ActorRefFrom<TBehavior> {
+  const machineWithConfig = behavior;
+
+  const service = interpret(machineWithConfig, options);
```
Andarist, line 13:
```suggestion
  const actorRef = interpret(behavior, options);
```

packages/xstate-solid/src/useMachine.ts
```diff
@@ -1,6 +1,6 @@
 import type { AnyStateMachine, StateFrom, InterpreterFrom, Prop } from 'xstate';
```
Andarist, line 1: likely this file should be removed (or rather renamed to `useActor`)

davidkpiano: I did and worked on making this more generic, but there are some type issues I need help with

davidkpiano:
@Andarist Might need help with changesets here since I'm not getting the option to add changesets for the changed packages

packages/xstate-svelte/test/UseMachineNonPersistentSubcription.svelte
```diff
@@ -21,13 +21,13 @@
...
 </script>
 
 <div>
   <button type="button" on:click={() => (visible = !visible)}>Toggle</button>
   {#if visible}
     <!-- inlined version of this doesn't unsubscribe from the store when the content gets hidden, so we need to keep this in a separate component  -->
-    <UseMachineNonPersistentSubcriptionChild {send} {state} />
-  {/if}
+    <UseMachineNonPersistentSubcriptionChild {send} {snapshot} />
+  {/if}useActor
```
Andarist, line 32:
```suggestion
  {/if}
```

packages/xstate-vue/src/useMachine.ts
```diff
@@ -1,59 +1,29 @@
...
-  send: Prop<TInterpreter, 'send'>;
-  service: TInterpreter;
-};
+import { useActor } from './useActor.ts';
+import { Ref } from 'vue';
 
+/**
+ * @deprecated Use `useActor(...)` instead.
+ */
 export function useMachine<TMachine extends AnyStateMachine>(
```
Andarist, line 16: I wonder how useful it really is to keep this around if its signature changed *anyway*. Right now, the second argument only accepts `InterpreterOptions` whereas previously it also accepted `InternalMachineImplementations`.

Andarist (changes requested):
I juggled files a little bit more to clean up more places, rename them properly etc. I think this is now almost in a good state but I think that READMEs should also get adjusted here.

There are also signature inconsistencies between `useActor`, `useMachine` and `useActorRef` across packages but let's tackle that unification in a follow up PR.

Andarist:
> @Andarist Might need help with changesets here since I'm not getting the option to add changesets for the changed packages

I can't reproduce this problem locally but remember that changeset files are just simple `.md` files. The CLI is a nice convenience (and it helps a lot!) but it's not a strict requirement to create those files. You can always just copy-paste some existing one and tweak it however you like.

davidkpiano:
> I juggled files a little bit more to clean up more places, rename them properly etc. I think this is now almost in a good state but I think that READMEs should also get adjusted here.
> 
> There are also signature inconsistencies between `useActor`, `useMachine` and `useActorRef` across packages but let's tackle that unification in a follow up PR.

The source of truth for this documentation will actually be in the docs repo (https://github.com/statelyai/docs) - let's do this in a separate PR, since I want to make a good way of syncing those documentation pages with the README pages (maybe as simple as providing a link once the docs are updated).

davidkpiano:
@Andarist I need help merging `next` into this one

Andarist:
> The source of truth for this documentation will actually be in the docs repo (https://github.com/statelyai/docs) - let's do this in a separate PR, since I want to make a good way of syncing those documentation pages with the README pages (maybe as simple as providing a link once the docs are updated).

I feel like the syncing problem is separate. If you come up with a syncing solution, that PR should be focused on syncing - not on the docs updates. I feel like updating the code here without adjusting the user-facing API reference in the README when the API changes so much is something that shouldn't land. We don't have to make it perfect right now - just update the signatures and mention what arguments are required by them.

davidkpiano:
Going to redo this since this is stale

PR #4014, opened 2023-05-14
[v5] Improve errors

davidkpiano:


packages/core/src/StateMachine.ts
```diff
@@ -352,16 +352,17 @@ export class StateMachine<
...
     });
-    Object.values(state.children).forEach((child) => {
+    for (const child of Object.values(state.children)) {
       if (child.status === 0) {
         try {
           child.start?.();
         } catch (err) {
           // TODO: unify error handling when child starts
           actorCtx.self.send(error(child.id, err) as unknown as TEvent);
+          throw err;
```
Andarist, line 362: hm, that likely should test if there is a transition defined for handling errors and only rethrow if there is no such a transition. I suspect that this wouldn't pass (or would introduce some unwanted behaviors) a test case like this:
```ts
const actorRef = interpreter(
  createMachine({
    initial: "a",
    states: {
      a: {
        on: {
          NEXT: "b",
        },
      },
      b: {
        invoke: {
          src: fromCallback(() => {
            throw new Error("error");
          }),
          onError: "c",
        },
      },
      c: {},
    },
  })
);

actorRef.start();
actorRef.send({ type: "NEXT" });

expect(actorRef.status).toBe(ActorStatus.Running);

```

davidkpiano: You're right, will change

davidkpiano:
Superseded by #4145

PR #4016, opened 2023-05-15
Migrate away from the `process.env.NODE_ENV` convention

Andarist:
fixes https://github.com/statelyai/xstate/issues/787
supersedes https://github.com/statelyai/xstate/pull/2820


packages/core/src/State.ts
```diff
@@ -1,5 +1,5 @@
+import isDevelopment from '#is-development';
```
davidkpiano, line 1: Whoa, is there prior art to this? Never seen it before 😮

Andarist: It's a standard thing - it's based on the lesser-known sibling of the `package.json#exports`. 

we define this specific **internal** import in `package.json` here:
https://github.com/statelyai/xstate/blob/e7946a1b84f1d548c8256e5f36aa7af9377c923e/packages/core/package.json#L81-L86 

The things there have to start with `#`. I assume that this is modeled after "private class fields" (as they also use leading `#`). And so, with different conditions we point to files that resolve to `true`/`false` and our builder ([preconstruct](https://github.com/preconstruct/preconstruct/)) does it magic and produces all combinations of the used `imports` entries~. The neat thing is that it doesn't really know what conditions there might be, it doesn't dictate how you are supposed to name those things etc. The only thing that matters is that those are defined in `package.json#imports` so preconstruct can group them to the least amount of combinations and stuff.

davidkpiano (approved):

PR #4017, opened 2023-05-15
[v5] `useSnapshot`

davidkpiano:
The Context object returned from `createActorContext(...)` now provides a `.useSnapshot()` hook, which replaces `useActor()`:

```ts
// instead of this:
// const [state, send] = SomeContext.useActor();
const [state, send] = SomeContext.useSnapshot();
```

### DIFF
--- a/.changeset/fifty-flies-count.md
+++ b/.changeset/fifty-flies-count.md
@@ -2,7 +2,7 @@
 '@xstate/react': major
 ---
 
-The Context object returned from `createActorContext(...)` now provides a `.useSnapshot()` hook, which replaces `useActor()`:
+The Context object returned from `createActorContext(...)` now provides a `.useSnapshot()` hook, which replaces what was previously provided as `.useActor()`:
 
 ```ts
 // instead of this:
--- a/.changeset/fifty-flies-count.md
+++ b/.changeset/fifty-flies-count.md
@@ -1,0 +1,11 @@
+---
+'@xstate/react': major
+---
+
+The Context object returned from `createActorContext(...)` now provides a `.useSnapshot()` hook, which replaces `useActor()`:
+
+```ts
+// instead of this:
+// const [state, send] = SomeContext.useActor();
+const [state, send] = SomeContext.useSnapshot();
+```
--- a/packages/xstate-react/src/createActorContext.ts
+++ b/packages/xstate-react/src/createActorContext.ts
@@ -11,7 +11,8 @@ import {
   MarkAllImplementationsAsProvided,
   StateMachine,
   EventFromBehavior,
-  AnyActorBehavior
+  AnyActorBehavior,
+  AnyActorRef
 } from 'xstate';
 
 type ToMachinesWithProvidedImplementations<TMachine extends AnyStateMachine> =
@@ -91,7 +92,7 @@ export function createActorContext<TBehavior extends AnyActorBehavior>(
   }
 
   Provider.displayName = `ActorProvider(${
-    machine.id ?? machine.transition.name
+    (machine as unknown as AnyStateMachine).id ?? machine.transition.name
   })`;
 
   function useContext(): ActorRefFrom<TBehavior> {
@@ -122,7 +123,7 @@ export function createActorContext<TBehavior extends AnyActorBehavior>(
     const actor = useContext();
     const state = useSelector((s) => s);
 
-    return [state, actor.send, actor];
+    return [state, (actor as AnyActorRef).send, actor];
   }
 
   return {
--- a/packages/xstate-react/src/createActorContext.ts
+++ b/packages/xstate-react/src/createActorContext.ts
@@ -35,7 +35,7 @@ type ToMachinesWithProvidedImplementations<TMachine extends AnyStateMachine> =
     : never;
 
 export function createActorContext<TBehavior extends AnyActorBehavior>(
-  machine: TBehavior,
+  behavior: TBehavior,
   interpreterOptions?: InterpreterOptions<TBehavior>,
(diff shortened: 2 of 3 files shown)


.changeset/fifty-flies-count.md
```diff
@@ -0,0 +1,11 @@
+---
+'@xstate/react': major
+---
+
+The Context object returned from `createActorContext(...)` now provides a `.useSnapshot()` hook, which replaces `useActor()`:
```
Andarist, line 5:
```suggestion
The Context object returned from `createActorContext(...)` now provides a `.useSnapshot()` hook, which replaces what was previously provided as `.useActor()`:
```

.changeset/fifty-flies-count.md
```diff
@@ -0,0 +1,11 @@
...
+
+```ts
+// instead of this:
+// const [state, send] = SomeContext.useActor();
+const [state, send] = SomeContext.useSnapshot();
```
Andarist, line 10: what do you think about exporting the same thing~ from the root of the package as well?
```ts
import { useSnapshot } from '@xstate/react'

const actorRef = useActorRef(machine)
const [snapshot, send] = useSnapshot(actorRef)
```

Andarist: like, i'd probably never use it... but then things put on the actor context would match better what we export from `@xstate/react`

packages/xstate-react/src/createActorContext.ts
```diff
@@ -67,20 +80,22 @@ export function createActorContext<TMachine extends AnyStateMachine>(
     machine: providedMachine = machine
   }: {
     children: React.ReactNode;
-    machine: TMachine;
+    machine: TBehavior;
```
Andarist, line 83:
```suggestion
    behavior: TBehavior;
```

packages/xstate-react/src/createActorContext.ts
```diff
@@ -67,20 +80,22 @@ export function createActorContext<TMachine extends AnyStateMachine>(
     machine: providedMachine = machine
```
Andarist, line 80:
```suggestion
    machine: providedBehavior = behavior
```

packages/xstate-react/src/createActorContext.ts
```diff
@@ -67,20 +80,22 @@ export function createActorContext<TMachine extends AnyStateMachine>(
...
       observerOrListener
-    ) as ActorRefFrom<TMachine>;
+    ) as ActorRefFrom<TBehavior>;
 
     return React.createElement(OriginalProvider, { value: actor, children });
   }
 
-  Provider.displayName = `ActorProvider(${machine.id})`;
+  Provider.displayName = `ActorProvider(${
+    (machine as unknown as AnyStateMachine).id ?? machine.transition.name
```
Andarist, line 95: naughty

Andarist: I'm not sure if `?? machine.transition.name` is really that great here, that usually would result in just `'transition'`

.changeset/fifty-flies-count.md
davidkpiano: Sure, good symmetry.

packages/xstate-react/src/createActorContext.ts
```diff
@@ -93,16 +108,28 @@ export function createActorContext<TMachine extends AnyStateMachine>(
...
 
+  function useSnapshot(): [
+    SnapshotFrom<TBehavior>,
+    (event: EventFromBehavior<TBehavior>) => void,
+    ActorRefFrom<TBehavior>
+  ] {
+    const actor = useContext();
+    const state = useSelector((s) => s);
+
+    return [state, (actor as AnyActorRef).send, actor];
```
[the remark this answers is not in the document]
davidkpiano: Yeah, same argument can be made against `useState` though

packages/xstate-react/src/createActorContext.ts
```diff
@@ -32,33 +34,44 @@ type ToMachinesWithProvidedImplementations<TMachine extends AnyStateMachine> =
...
+        > extends true
+        ? {
+            machine?: TBehavior;
+          }
+        : {
+            machine: ToMachinesWithProvidedImplementations<TBehavior>;
+          }
       : {
-          machine: ToMachinesWithProvidedImplementations<TMachine>;
+          machine?: TBehavior;
```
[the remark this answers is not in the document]
davidkpiano: What about other actor logic?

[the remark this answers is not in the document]
Andarist: hm, `machine` prop is specific to machines - so it only should be accepted here if `TBehavior extends AnyStateMachine`. If something is unclear - could you give a code example of what situation are you thinking about here?

packages/xstate-react/src/createActorContext.ts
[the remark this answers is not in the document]
Andarist: hm, yee - but it creates and own the state so it makes slightly more sense there that they have to give us the setter *somehow* (through the return value). This code only consumes the `actorRef` and the "setter" (`send`) already exists... but ye, since we don't have any alternative ideas here... this seems to be good enough.

packages/xstate-react/src/createActorContext.ts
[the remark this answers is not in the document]
Andarist: Ah, OK - i see what you mean here. Yeah, it might make sense to actually always accept this optionally here. Disregard my comment

davidkpiano:
Decided not to add `.useSnapshot()`

PR #4028, opened 2023-05-18
Update xstate.js.org

laurakalbag:
As #4026 but targeting `main` this time. Thanks for the help, @Andarist.


index.html
```diff
@@ -0,0 +1,147 @@
+<!DOCTYPE html>
```
Andarist, line 1: The file structure deployed to `gh-pages` is vastly different from the equivalent sources. 

It seems that this is the file that describes the structure of the page: https://github.com/statelyai/xstate/blob/595964aac347a0fef2db5c41064ceb562f83a92d/docs/.vuepress/config.js . Do you only want to change the styles? Or do you also need to do some other changes?

laurakalbag: > The file structure deployed to `gh-pages` is vastly different from the equivalent sources.

Yeah, I think it might contain a really old version of the repo… not sure how that would happen.

>Do you only want to change the styles? Or do you also need to do some other changes?

It’s just a few changes to the styles and content. And also updating the logo-white.svg image.

davidkpiano (approved):

PR #4036, opened 2023-05-22
[v5] Actor types

davidkpiano:
```ts
  types: {} as {
    actors: {
      src: 'fetchData'; // src name (inline behaviors ideally inferred)
      id: 'fetch1' | 'fetch2'; // possible ids
      input: { foo: string };
      output: { result: string };
    }
  },
  invoke: {
    src: 'fetchData', // strongly typed
    id: 'fetch2', // strongly typed
    onDone: {
      actions: ({ event }) => {
        event.output; // strongly typed as { result: string }
      }
    },
    input: { foo: 'hello' } // strongly typed
  }
});
```


.changeset/seven-balloons-attack.md
```diff
@@ -0,0 +1,41 @@
...
+Actor types can now be specified in the `.types` property of `createMachine`:
+
+```ts
+const machine = createMachine({
+  types: {} as {
+    actors: {
+      src: 'fetchData'; // src name (inline behaviors ideally inferred)
+      id: 'fetch1' | 'fetch2'; // possible ids
+      input: { foo: string };
+      output: { result: string };
```
Andarist, line 14: Before adding `input` and `output` here we should first add `TInput` and `TOutput` to the actor types.

packages/xstate-solid/test/useMachine.test.tsx
```diff
@@ -32,13 +32,18 @@ afterEach(() => {
...
   };
-  const fetchMachine = createMachine<
-    typeof context,
-    { type: 'FETCH' } | DoneEventObject
-  >({
+  const fetchMachine = createMachine({
     id: 'fetch',
+    types: {} as {
+      context: typeof context;
+      events: { type: 'FETCH' } | DoneEventObject;
```
Andarist, line 41: I think that users shouldn't have to include done events in what they provide for `TEvent`. I've checked this locally and this works without it.
```suggestion
      events: { type: 'FETCH' };
```

packages/xstate-solid/test/useMachine.test.tsx
```diff
@@ -54,7 +59,7 @@ describe('useMachine hook', () => {
             actions: assign({
               data: ({ event }) => event.output
             }),
-            guard: ({ event }) => event.output.length
+            guard: ({ event }) => !!event.output.length
```
Andarist, line 62: I noticed that the narrowed down event type is provided to `guard` here but it's not provided to `actions`. It's quite surprising since `TExpressionEvent` is used in both.

packages/xstate-react/test/useActor.test.tsx
```diff
@@ -48,6 +56,7 @@ describeEachReactMode('useActor (%s)', ({ suiteKey, render }) => {
                 return event.output;
               }
             }),
+            // TODO: this should be typed as { output: string }
```
Andarist, line 59: If you'd remove explicit generics from this `createMachine` call then this would be typed correctly. This happens because your actor-related types on `config.types` are just ignored (since you provided explicit generics and didn't pass actor-related type there)

packages/core/test/invoke.test.ts
```diff
@@ -1185,6 +1224,10 @@ describe('invoke', () => {
                           src: 'getRandomNumber',
                           onDone: {
                             target: 'success',
+                            // TODO: investigate why event is DoneInvoke<any> instead of
+                            // DoneInvoke<{result: ...}>
+                            // The type is properly inferred for:
+                            // actions: ({ event }) => event.output.result
```
Andarist, line 1230: nested inference contexts strike again 😬 

I think the conditional type that distributes over "non-fixed" (aka not yet set in stone) `TActors` is at fault here. I prepared a somewhat minimal repro of this problem: [TS playground](https://www.typescriptlang.org/play?ts=5.1.3#code/JYOwLgpgTgZghgYwgAgIILAeygSQLYAOANsgN4BQyyAzlAgFw1hSgDmA3JcpgK5gF9GcEAE9OVAPQSq3PgLCMeIANYhMAdxCcAvuXKhIsRCgAimEBBwgAbpmUQAotYjgAPABUTcMHAB8ZLjARAghGAAMAE3MIADpQW3sYgBJSamY2bTDxWX5BZE9vOB09IJC0ampgVhB0MGBzAHkAIwArCAwPJxcwZAgAD0gQCOoyZFLQphYQVmRtfwBeAKpxxgAiPrTvWLgKqpBV7IhrRncu8GyACiOTs7AASkZbYAidTnJxtAx6kAAxJS-zJ1nOBegMXMNRitJmxZgtkFdjvlbndkPN-E8XnoIu0iHAoCgYP86uZkDtKtUgd1QYMIaQxsEJmkpjM5hcuGS9tBGAibsD7qj0ZhnuQHuVyTUASBmm0Oqc+b43lJkEqohBqCAAOQ9dTYZTIdTAMAACzGRuAI2seOAwh6EENRug5BVOLxBKJ31Juwpcqp-RpIzpUKZMNZTukVA51S58OuSL5KLRyAxYeQotq3z+IEllPACpKDM+xJAOdtYKGAfpIUYwemsNRXAAPmK9unGq12mAS75G4WM+7AT7c29sQhcfjkISs0XkAh8VsALKIM0WDy1bAjP3gkZr3CEIi+C4IcwwKqMCjLBnUM9cCMYdcnHfUbLabKbSBX5AAJXa2Airhv0LTAANAB54yFQ8R2BM7iPtSW6jLQDDIKAMDQPkADKdCzAB4EAPxLOBhFUIhJyYQg2REeB5hmBYjCtsWNGWDYUG3Kud5QNQADaqy8LkYCrAAur4eaUTIuiiYwFjOFAFHYVQebaKKGJvLOEALkuoAQBcYHjB+YG3lgHFntonoEeBJHIKsrAQGAn7CFEeAAHI8HgTTQAcOE5PIZ7IPi1A8EQCjICALluVAsyydoIFUFFXBvmq17gXAiWEZB9gpURFlWTZdlDJgTmhe50WUdR0Tckc3QJv4+mURV5yeeBSoAHq4Q1szFeBsWdcVsWKSpc6QIuCDLlpOmXhlpLsXpJk7GZMhZdZtn2flzmue5slULx3mjH5AVBSFa3hS+AFde1cU+O+E3JXNEHMelN3zXQayLblDmrWFqwdYRpW0Z64oXAilUCg94F1WAG2Ec1rWiYpX0xV9p29XcnBAA)

Note that the type for that action object is pulled directly from the constraint, if we do this:
```diff
interface ActorImpl {
  src: string;
-  output: any;
+  output: any;
}
```
then we'll see `DoneInvokeEvent<unknown>` instead of `DoneInvokeEvent<any>`.

packages/core/src/types.ts
```diff
@@ -1009,19 +1037,29 @@ export interface MachineConfig<
...
   version?: string;
-  types?: MachineTypes<TContext, TEvent, TActorMap, TTypesMeta>;
+  types?: MachineTypes<TContext, TEvent, TActors, TTypesMeta>;
+}
+
+export interface ActorImpl {
+  src: string;
+  id?: string;
+  logic?: AnyActorLogic;
+  // TODO: delete all the below and replace with logic
```
Andarist, line 1047: this feels like a hard requirement before landing this PR, right?

.changeset/seven-balloons-attack.md
davidkpiano: Done

packages/core/test/invoke.test.ts
davidkpiano: Is there any workaround to this problem?

Andarist: Convince TS team to merge my https://github.com/microsoft/TypeScript/pull/48838 😆 see how it works just alright in the playground that uses that PR: [TS playground](https://www.staging-typescript.org/play?lib=lib.esnext.d.ts&ts=5.2.0-pr-48838-5#code/JYOwLgpgTgZghgYwgAgIILAeygSQLYAOANsgN4BQyyAzlAgFw1hSgDmA3JcpgK5gF9GcEAE9OAX3LlQkWIhQARTCAg4QAN0wBrCAFF1EcAB4AKgrhg4APjJcwIghEYADACbKIAOlCadngCSk1Mxs4s6cVLz8gshmFnASUvaOaNTUwKwg6GDAygDyAEYAVhAYpvqGYMgQAB6QIK7UZMjJTkwsIKzI4jYAvLZUrYwARDXBFl5waRkgwxHV6owmFeDzABQQi7ErYACUjJrArhKc5K1oGLkgAGI8IJfK5Qbg1XWGjc1D7WzdfcgbW2Wzz2yF6NkOxykrlKRDgUBQMDuDxAyCm6UyT0qr3qH1ILQcbWCHS6PTWXDRM2gjABSx2u1B4MwR3I+1S6KyyMKJTKQMqVlO52yV0xL1qOKaeK+RJ+PVBXAAPmyZkL8sVSmARWArAqLjllLd7nqQJr+VCYXCEUijcgEPCJgBZRAAC1AEFM2WwTTF7yaHtwhCIVjWCGUMAyjAogwJ1AjXCoiCwUBjsT91Hm4nm40gyYASqVsK4jHHvp0ADTFyNUKs+bRtEyp7E+5q0BjIUAwaCxADKdG6xaryAA-AMB6OqC2lj2EPMxwPlEoVIwVcaF6oNLWdu6MJ6ANrDKICMDDAC6VlNs6rkgvjBUBigM77VFN4lZENOtogDudrrWldaycreNtyTCNxFRCV+xoOgRlYCAwBzYR3DwAA5Hg8AKaA5kgg8YjxeFqB4IgwBvNCMKgboH3EcsqCorgswgADizgWMxxrHQWNnCdkGGWD4MQzAUNIzDqNnecPGpTZKnpMERwvSTVkggcAHolOQAA9QdFO6ESB1o3SRNol93ztSBHQQF0VF-Oxow4oDEwAsCplkqsuJ4uCEIaATUPQzCH0iPhDwjZB8MI4jkBAITyIzYs9O0ujLGzWzUSSqg2LaQDR1c3iPKQ7yyOGHTRzExdwPZNYASkhlnLHeSwD80cVPUzSLxfQqaMK2LDN2TggA)

On a serious note, I'd have to spend some time to figure this out for the current TS capabilities (if it's even possible)

.changeset/seven-balloons-attack.md
```diff
@@ -0,0 +1,39 @@
+---
```
Andarist, line 1: TODO: revamp this changeset since `ActorImpl` is no longer a thing and we require `logic` (`{ 
src: string; id?: string; logic: AnyActorLogic }`)

packages/core/src/State.ts
```diff
@@ -39,7 +43,9 @@ export function isStateConfig<
 export const isState = isStateConfig;
 export class State<
   TContext extends MachineContext,
-  TEvent extends EventObject = EventObject,
+  TEvent extends EventObject,
+  _TActions extends ParameterizedObject,
+  TActor extends ProvidedActor,
```
Andarist, line 48: I wonder if this shouldn't receive `TChildren` already (preprocessed type). I'm still on the fence about this though

packages/core/src/State.ts
```diff
@@ -63,7 +69,14 @@ export class State<
   /**
    * An object mapping actor names to spawned/invoked actors.
    */
-  public children: Record<string, ActorRef<any>>;
+  public children: TActor['id'] extends string
```
Andarist, line 72: TODO: this could be improved. The fallback branch here kicks in way too early. For example: when we **provide** `TActor` without any IDs then we should create an index signature of all provided actor types. In fact, this should also happen for a "mixed" bag of actors (when some have IDs and when some don't). In other words, the index signature containing possible actor types should always be available (unless all actors have explicit IDs assigned to them?).

packages/core/test/invoke.test.ts
Andarist: This one still falls back to the constraint - despite the fact that they have callable signatures already. I doubt that this is something that we can control rn without landing the mentioned fix and all I can do is to ping the TS team periodically to give it another look 🤷‍♂️

Andarist: That PR of mine mentions that the position of the conditional type can change the inferred result - so I guess I still have to try this out.

packages/core/test/input.test.ts
```diff
@@ -283,8 +300,14 @@ describe('input', () => {
...
       {
+        types: {} as {
+          actors: {
+            src: 'child';
+            logic: AnyActorLogic;
+          };
+        },
         entry: assign(({ spawn }) => ({
-          childRef: spawn('child')
+          childRef: spawn('child') // TODO: type-check for spawn
```
Andarist, line 310: TODO: type `spawn`

packages/core/src/typegenTypes.ts
```diff
@@ -208,11 +202,13 @@ export interface ResolveTypegenMeta<
       AllImplementationsProvided &
       AllowAllEvents & {
         indexedActions: IndexByType<TAction>;
+        indexedActors: IndexByProp<TActor, 'src'>;
         indexedEvents: Record<string, TEvent> & {
           __XSTATE_ALLOW_ANY_INVOKE_OUTPUT_HACK__: { output: any };
         };
         invokeSrcNameMap: Record<
           string,
+          // TODO: remove this hack
```
Andarist, line 211: TODO: recheck if this can already be removed

packages/core/src/typegenTypes.ts
```diff
@@ -146,32 +149,16 @@ export interface MarkAllImplementationsAsProvided<TResolvedTypesMeta> {
...
+      type: `done.invoke.${K & string}`;
+      output: OutputFrom<TActor['logic']>;
+    }
   : never;
 
-type GenerateActorEvents<
-  TActorMap extends ActorMap,
-  TInvokeSrcNameMap
-> = string extends keyof TActorMap
-  ? never
```
Andarist, line 163: TODO: this had a potentially important check here, need to re-evaluate its role and if we are losing anything by removing this today

packages/core/src/actors/transition.ts
```diff
@@ -12,16 +12,17 @@ import { ActorLogic, ActorContext, ActorSystem, EventObject } from '../types';
...
   transition: (
     state: TState,
     event: TEvent,
     actorContext: ActorContext<TEvent, TState, TSystem>
   ) => TState,
-  initialState: TState | (({ input }: { input: any }) => TState) // TODO: type
+  initialState: TState | (({ input }: { input: TInput }) => TState) // TODO: type
 ): ActorLogic<TEvent, TState, TState> {
-  const logic: ActorLogic<TEvent, TState, TState, TState> = {
+  const logic: ActorLogic<TEvent, TState, TState> = {
```
Andarist, line 25: TODO: Since `TInput` was introduced to this actor type - it should be passed to this `ActorLogic` here as well

packages/core/src/types.ts
```diff
@@ -84,25 +84,40 @@ export interface ActionArgs<
...
+export type InputFrom<T extends AnyActorLogic> = T extends ActorLogic<
+  infer _TEvent,
+  infer _TSnapshot,
+  infer _TInternalState,
+  infer _TPersisted,
+  infer _TSystem,
+  infer TInput
+>
+  ? TInput
+  : any; // TODO: never?
```
Andarist, line 96: hm, at first I was thinking about `undefined` but probably `unknown` would fit this even better (let's tinker with that in a separate PR though)

packages/core/src/types.ts
```diff
@@ -423,55 +440,105 @@ export type TransitionsConfig<
...
-}
+  TEvent extends EventObject,
+  TActor extends ProvidedActor
+> = TActor extends { src: infer TSrc }
+  ? {
+      /**
+       * The unique identifier for the invoked machine. If not specified, this
+       * will be the machine's own `id`, or the URL (from `src`).
+       */
+      id?: TActor['id'];
```
Andarist, line 453: hm, when the `id` is provided I feel like it's better to require, thoughts?

packages/core/src/types.ts
```diff
@@ -499,11 +566,13 @@ export interface StateNodeConfig<
...
    * The mapping of state node keys to their state node configurations (recursive).
    */
-  states?: StatesConfig<TContext, TEvent, TAction> | undefined;
+  states?: StatesConfig<TContext, TEvent, TAction, TActor> | undefined;
   /**
    * The services to invoke upon entering this state node. These services will be stopped upon exiting this state node.
    */
-  invoke?: SingleOrArray<string | InvokeConfig<TContext, TEvent>>;
+  invoke?: SingleOrArray<
+    TActor['src'] | InvokeConfig<TContext, TEvent, TActor>
```
Andarist, line 574: Passing a string to `invoke` is still supported - but it seems that we only have 4~ tests using this "flavor" and none of them is specifically testing this flavor. Should we continue supporting this?

packages/core/src/State.ts
```diff
@@ -39,7 +43,9 @@ export function isStateConfig<
 export const isState = isStateConfig;
 export class State<
   TContext extends MachineContext,
-  TEvent extends EventObject = EventObject,
+  TEvent extends EventObject,
+  _TAction extends ParameterizedObject,
```
Andarist, line 47: This should be removed from here. I think that `State` won't actually depend on `TAction` (it doesn't right now and some of the recent changes were specifically decoupling `State` from actions).

packages/core/src/types.ts
davidkpiano: You mean when it's provided to the `{ actors: ... }` type?

packages/core/src/types.ts
davidkpiano: Especially if it's simpler to type, I don't think we should continue supporting it. Quick reasons:

- Only ~one~ two ways to invoke: `invoke: {}` and `invoke: [{}, {}, ...]` = good for consistency (besides `SingleOrArray`)
- Having `src: ...` subtly encourages adding `onDone` and `onError`, which are good things to add
- Simplifies things for future `machine-extractor`

packages/core/src/State.ts
davidkpiano: I removed them

packages/core/test/invoke.test.ts
Andarist: I tried moving this conditional type in a variety of ways and couldn't make it work. I think it's best to assume, for now, that it's not possible to improve this and we can revisit this later.

packages/core/src/types.ts
Andarist: Yeah, if any of the given `src` has an associated `id` with it then it feels like we should *require* it here.

packages/core/src/types.ts
Andarist: 👍 let's do this in a followup PR

packages/core/src/types.ts
davidkpiano: Okay so to clarify:
```ts
createMachine({
  type: {} as {
    actors: {
      src: 'someSrc';
      logic: SomeLogic;
      id: 'foo' | 'bar';
    }
  },
  invoke: {
    src: 'someSrc',
    // no id = type error
  }
})
```

I agree, if that makes things simpler

packages/core/test/typegenTypes.test.ts
```diff
@@ -1187,9 +1191,10 @@ describe('typegen types', () => {
         }
       },
       {
-        // @ts-expect-error
+        // TODO: determine the exact behavior here and how eventsCausingActors + TActor should interact with each other
```
Andarist, line 1194: TODO: let's discuss this

packages/core/test/typegenTypes.test.ts
```diff
@@ -947,7 +951,7 @@ describe('typegen types', () => {
           fooActor: fromCallback((_send, onReceive) => {
             onReceive((event) => {
               ((_accept: string) => {})(event.type);
-              // @x-ts-expect-error TODO: determine how to get parent event type here
+              // @ts-expect-error TODO: determine how to get parent event type here
```
Andarist, line 954: TODO: I can try to do that in a follow up PR

packages/core/src/types.ts
Andarist: done in https://github.com/statelyai/xstate/pull/4036/commits/68609041e469e239fac67874eb27cef5ab024901

packages/core/src/types.ts
```diff
@@ -85,25 +85,40 @@ export interface ActionArgs<
...
+  infer TOutput
+>
+  ? TOutput
+  : any; // TODO: never?
+
 // TODO: do not accept machines without all implementations
 // we should also accept a raw machine as actor logic here
 // or just make machine actor logic
-export type Spawner = <T extends ActorLogic<any, any> | string>( // TODO: read string from machine logic keys
+export type Spawner = <T extends AnyActorLogic | string>( // TODO: read string from machine logic keys
```
Andarist, line 114: TODO: improve `spawn`

packages/core/src/types.ts
Andarist: actually, this type constraints the input already so this branch should just use `never` here (it's a dead branch anyway)

packages/core/src/actors/transition.ts
Andarist: already fixed this

datner:
@Andarist @davidkpiano 
Hey, sorry to pop in unannounced like this. I saw that a lot of pain is coming from handling variance..
I don't imagine I can just drop in here and deliver the solution like I have all (or any) of the details, but maybe I'll bring a spark of inspiration

I don't know if you already know the following or not, I'm not assuming and I don't mean to offend, I'm just trying to help 😄 

There are two ways that I know of of covariance and contravariance being properly expressed within typescript.
1. the `in` and `out` variance modifiers on generics. Extremely niche but very much existing feature of typescript (not even documented on their website for some reason)
```ts
type Provider<out T> = () => T;
type Consumer<in T> = (x: T) => void;
type Mapper<in T, out U> = (x: T) => U;
type Processor<in out T> = (x: T) => T;
```
2. "hack" typescript to respect variance
```ts
declare const Variance: unique symbol 
interface Provider<T> {
  [Variance]: {
    _T: (_: never) => T
  }
}

interface Consumer<T> {
  [Variance]: {
    _T: (_:T) => never
  }
}

interface Mapper<T,U> {
  [Variance]: {
    _T: (_:T) => never
    _U: (_: never) => U
  }
}

interface Processor<T> {
  [Variance]: {
    _T: (_:T) => T
  }
}
```
The latter being especially useful to keep track of types generically without widening types or smashing into bivariance bugs
there are other hacks in regards to tracking kind of typeinfo but I am not savvy enough to explain why they work, though I can try to connect you with those who can if it's useful.

packages/core/src/typegenTypes.ts
Andarist: This was still needed for some things that we do. I brought this in https://github.com/statelyai/xstate/pull/4036/commits/d5b63e33a4a3a903c572ac20458a1884af2b734a and added comments explaining this whole type in https://github.com/statelyai/xstate/pull/4036/commits/ca81cc66e2a4d3bf3a6ba4f44fa19a4f0dbdd07f

Andarist:
> the in and out variance modifiers on generics. Extremely niche but very much existing feature of typescript (not even documented on their website for some reason)

TS team considers announcement blog posts to be part of the documentation 🤷‍♂️ So this is "documented" in [TS 4.7 blog post](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#optional-variance-annotations-for-type-parameters). 

> "hack" typescript to respect variance

We already utilize a similar technique, for example here:
https://github.com/statelyai/xstate/blob/4df574405f8fa412acbd2f5ee6eb87825d9ef4f7/packages/core/src/types.ts#L122

Although this one is more about helping the inference engine to recognize this as a potential inference source than about actually enforcing variance.

> Hey, sorry to pop in unannounced like this.

Don't be sorry! Every help is very much appreciated.

> I saw that a lot of pain is coming from handling variance..

If this particular PR spurred this thought, then let me clarify what I'm doing in those tests. Type parameters have their own variance as long as they are actually used by structures. We rarely want to use runtime values as type containers (although at times that comes in handy and we might be using such tricks at some point. Those extensive~ tests were just added to ensure that the variance is correct. To some extent, we could just annotate type parameters with `in`/`out` and be done with it but:
1. that requires TS 4.7 and we might still want to support older versions. I'm OK with breaking compat with older TS versions in general, but I don't feel that those annotations have good enough ROI to be an argument for that.
2. it's easier for me to reason about how those variances play out in concrete examples than to reason about the implications of a local annotation. It's easy to reason about the annotated types but I have a hard time reasoning if they are doing what I want for structures references those annotated types. This just increases coverage and serves as an internal documentation, a more robust information for other maintainers about what we want to support etc

packages/core/src/State.ts
Andarist: fixed https://github.com/statelyai/xstate/pull/4036/commits/041b40b8bcddbec420471855f7cd5b1bc47c64d2 , https://github.com/statelyai/xstate/pull/4036/commits/2e7a58bc1428185ebd522b3d7abbf3739d267db6

datner:
> We already utilize a similar technique, for example here:

As I said I'm not totally in the details, but I know there is some significant difference between
```ts
// this:
interface Foo<A> {
  _?: A
}

// and this:
interface Bar<A> {
  _?: (_: never) => A
}
```
The function signature notation is doing more than just signaling that it's covariant. I didn't mean to use properties as type containers, their shape is coincidental. I should have used only type lambas instead of indicating covariant/contravariant association. that was confusing of me. I've seen many phantom types around the repo. And I know you are well aware of functions changing how the engine determines type, as this quirk is utilized here:
https://github.com/statelyai/xstate/blob/9911850389601f66cbd322a4b7800849a1eb3df2/packages/core/src/types.ts#L30

They can be both used together for greater good 😄 

There is a way to avoid runtime representation using a symbol as the key for an optional field and not exporting it (basically expunging it from existence for everything besides types)

I do think it's worth some exploration where this are applicable within xstate. A much more mature and comprehensive example than I am capable of producing can be found here:
https://github.com/Effect-TS/io/blob/8fa68a9539c5fe379bee8835128181a9a6104f39/src/Effect.ts#L204

Theres example of extraction, unification (turning `Foo<A, C, Z> | Foo<B, D, Z>` to `Foo<A | B, C | D, Z>` in cases where this is the runtime behavior), blacklisting, and less relevantly so things like HKT and type lambdas.

xstate is insanely smart, and I really think it straddles the edge of typescript in some cases. I don't think it should eschew more esoteric usage if users get more free stuff, especially if theres a compat concern

Andarist:
> The function signature notation is doing more than just signaling that it's covariant.

I'm not aware of any difference. Not saying that it isn't there - TS sometimes acts differently based on details like this and the heuristics that it implements. So far though I didn't notice any drawbacks of the approach that we are using so I'm fine with sticking to it. 

> And I know you are well aware of functions changing how the engine determines type, as this quirk is utilized here:

Well, that one is just copy-pasted from SO 🤣 but overall I'm pretty familiar with how TS works under the hood (learning more and more every day! I certainly don't have a perfect knowledge about those things).

> There is a way to avoid runtime representation using a symbol as the key for an optional field and not exporting it (basically expunging it from existence for everything besides types)

I'm slightly worried about using symbols because they are nominal and staying away from them allows us to potentially avoid some problems with mismatched package versions etc.

> I do think it's worth some exploration where this are applicable within xstate.

Sure thing, I'm just not sure what does it solve for us today. If it solves some concrete problems, I would love to add tests showcasing those problems and using the alternative that would make them pass.

> xstate is insanely smart, and I really think it straddles the edge of typescript in some cases. I don't think it should eschew more esoteric usage if users get more free stuff, especially if theres a compat concern

As I mentioned - I'm not strongly trying to avoid newer TS features. It's just that variance annotations aren't really strictly needed. You have to use those type parameters somewhere *anyway* because unused type parameters have unpredictable behaviors. And once type parameters are used they have their own variance already - annotations can't change them (well, they can make covariant/contravariant things invariant but that's not something that I'm looking for). In a way, those annotations should only be used as some kind of extra validation, like a lint rule.

datner:
Thanks for the detailed responses @Andarist, I appreciate it greatly.

I don't have too much free time, but if you can point me to some tests you'd like me to try and take a crack at, I'll be totally open to give it a go.

PR #4038, opened 2023-05-23
Links

davidkpiano:


examples/fetch/src/index.ts
```diff
@@ -8,7 +22,18 @@ const fetchMachine = createMachine({
         FETCH: 'loading'
       }
     },
-    loading: {}
+    loading: {
+      invoke: {
+        src: fromPromise(({ input }) => getGreeting(input.name)),
+        input: ({ context }) => ({ name: context.name })
```
kevinmaes, line 28: We're stuck in the loading state. Should this invoke lead to `success` when done?

PR #4043, opened 2023-05-26
feat: state value added to transition type

ksv90:


.changeset/rotten-tigers-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/fsm': patch
+---
+
+state value added to transition type
```
with-heart, line 5: Hi @ksv90!

Changesets generate changelog entries for users so you want to _think like a user_ when writing your changesets.

If you're a user reading a changelog entry, you'd want to know things like:

- **What** is the change?
- **Why** is the change important?
- **How** do I make use of the change?

Your changeset isn't _technically wrong_, it just doesn't help users understand the change that you've made here!

packages/xstate-fsm/src/types.ts
```diff
@@ -86,10 +86,14 @@ export namespace StateMachine {
...
-    | string
+  export type Transition<
+    TContext extends object,
+    TEvent extends EventObject,
+    TStateValue extends string
+  > =
+    | TStateValue
     | {
-        target?: string;
+        target?: TStateValue;
```
Andarist, line 96: What exactly does this solve for you? Could you share some code that is affected by this?

ksv90: @Andarist 
I decided that it would be better to show in dynamics. This file describes the state machine. The creation takes place elsewhere, here I wanted to show the development process. Please ask more questions if you need to explain in more detail what is happening in the video.

https://github.com/statelyai/xstate/assets/68888927/5c216593-56d0-43e8-9bf0-cd862b4f4039

```
import { StateMachine } from '@xstate/fsm';

export type TestContext = {
  readonly value: number;
};

export type TestStateValue = 'init' | 'loading' | 'idle' | 'process';

export type TestStateMachine = StateMachine.Config<
  TestContext,
  { type: 'NEXT' | 'STOP' | 'COMPLETE' },
  { value: TestStateValue; context: TestContext }
>;

export const createStateMachine = (): TestStateMachine => ({
  context: { value: 42 },
  initial: 'init',
  states: {
    init: {
      on: {
        COMPLETE: 'loading',
      },
    },
    loading: {
      on: {
        COMPLETE: { target: 'idle' },
      },
    },
    idle: {
      on: {
        NEXT: 'process',
      },
    },
    process: {
      on: {
        STOP: { target: 'idle' },
      },
    },
  },
});

```

ksv90:
Please write if these changes cannot be added at this time.

Andarist:
I'm OK with the change for the current major version of `@xstate/fsm` but I'm not yet sure if this will stay here in the next major version.

We are exploring ways to have stricter config types both for `xstate` and `@xstate/fsm`. I think though that you shouldn't have to declare your state names upfront - that's redundant information if the same could be inferred from the passed-in config. To rephrase - we'd like to get you the same auto-completion benefits and strict types for `initial`/`target` but without you passing us any `Typestate`

ksv90:
Thank you for the clarification. I want to add that in our case, we know all the names of the states before writing the code, since we design the state machine in advance. And the description of all states in a separate place (in this case, in the type) is an additional check so that we do not forget anything and describe all relationships. It becomes difficult to track if there are many states.

Perhaps in other situations it is really redundant. But then we will not have auto-completion until we have described the next state (otherwise the typescript will not know about this state) and we will have to return to writing each state again after that.

ksv90:
Is there any certainty on this issue?

Andarist:
I'll land this tomorrow.

Andarist:
> I want to add that in our case, we know all the names of the states before writing the code, since we design the state machine in advance. And the description of all states in a separate place (in this case, in the type) is an additional check so that we do not forget anything and describe all relationships. It becomes difficult to track if there are many states.

I don't consider this to be the primary way of authoring our state machines. Note that I'm not saying that it's bad or anything like that, it's just that usually state names come from the state machine itself and are not known ahead of time. For that reason, we plan to optimize for this workflow and use case.

I think though that there are ways to support your workflow as well. There are multiple ways to achieve this kind of extra validation that you are looking for. It's even possible to achieve this at type-level to get autocompletions and all. A short demo of this could look like this ([TS playground](https://www.typescriptlang.org/play?ts=5.1.3#code/C4TwDgpgBAysCGwIDl4FsIGcoF4oCIBLAEwBsJ8oAfA0ge3mMIDsBzSm-Yu5igbgBQA0JCgBBAPIAVGITRhyMAMYALCGngAeKVAgAPJM2LZMwAE4tWAPlxQA3gKhQWhYIXikAXFCmCnpxCxvACUIJTozYm0AGnsoHhCwiKjTCzZYqRsAXytBLMEBYjDSeDNoADMAV2YlNx4oJTLAgFl4VRYITXDmUx8AYR5ywmsACm6h1m8pAeYJgEpvADc6EgLGiBa2lQ6RhycXNw9vIjIKaMcoAKRMbz2nZ1Pbi-v45ieXl4BRADVj7l58OcPlAskD7qCLhCspdEIRMEMsOJpLJ5IpVOotHBAqgMJgrHNBEA)):
```ts
type StateNames = "idle" | "loading" | "done";

type AOTSimpleSchema<T extends string> = {
  initial: T;
  states: Record<T, { on: Record<string, T> }>;
};

declare function createMachine<const TConfig>(config: TConfig): void;

createMachine({
  initial: "idle",
  states: {
    idle: {
      on: {
        EV: "done",
      },
    },
  },
} satisfies AOTSimpleSchema<StateNames>);
```

packages/xstate-fsm/src/types.ts
```diff
@@ -104,9 +108,7 @@ export namespace StateMachine {
     changed?: boolean | undefined;
     matches: <TSV extends TState['value']>(
       value: TSV
-    ) => this is TState extends { value: TSV }
-      ? TState & { value: TSV }
-      : never;
+    ) => this is { value: TSV; context: TContext };
```
Andarist, line 111: what motivated this change? could you give an example of what didn't work before this change?

ksv90: Since TSV was extended with TSstate['value'] this worked well for a true expression. But with a false expression, the context became type "never". So it was not possible to check multiple states: 
if(state.matches('idle')) {}
else if(state.matches('latest')) {}

Andarist: could you add a test case for this [here](https://github.com/statelyai/xstate/blob/3d3a385a655ef524f2c90d908c04ed9855c42b71/packages/xstate-fsm/test/types.test.ts#L3)?

ksv90: There is no need to add new examples there. This test passes in the current implementation because the value is of type any. You can pass the number 12 to the match function and it will work. But I narrowed the value down to a string, so the typescript started swearing at the second branch of the condition. Then I changed the types for the correct behavior.

packages/xstate-fsm/src/types.ts
```diff
@@ -86,10 +86,14 @@ export namespace StateMachine {
     assignment: Assigner<TContext, TEvent> | PropertyAssigner<TContext, TEvent>;
   }
 
-  export type Transition<TContext extends object, TEvent extends EventObject> =
-    | string
+  export type Transition<
+    TContext extends object,
+    TEvent extends EventObject,
+    TStateValue extends string
```
Andarist, line 92:
```suggestion
    TStateValue extends string = string
```

.changeset/modern-spies-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/fsm': patch
+---
+
+Fixed state types when initializing the machine, as well as when switching to another state. Previously, the "string" type was used for this, which allowed any value to be passed and led to a subtle error in a running application. After this change, the error will be seen at the development stage.
```
Andarist, line 5:
```suggestion
Pass around `TState['value']` type to `Transition` and `.initial` property of the machine configuration.
```

PR #4049, opened 2023-05-31
[v5] Require `context` if specified

### LINKED ISSUE #583: [TS] Context not required on MachineConfig

davidkpiano:
Fixes #583


packages/core/test/types.test.ts
```diff
@@ -291,6 +299,35 @@ describe('types', () => {
...
+
+    createMachine({
+      types: {} as {
+        context: { count: number };
+      },
+      context: () => ({
+        count: 0
+      })
+    });
+  });
```
davidkpiano, line 330: This is the main test

packages/core/src/types.ts
```diff
@@ -1005,23 +1005,26 @@ export type ContextFactory<TContext extends MachineContext> = ({
...
-  context?: InitialContext<LowInfer<TContext>>;
   /**
    * The machine's own version.
    */
   version?: string;
   types?: MachineTypes<TContext, TEvent, TActorMap, TTypesMeta>;
-}
+  getContext?: (ctx: TContext) => void;
+}) &
+  (Equals<TContext, MachineContext> extends true
```
davidkpiano, line 1025: This probably needs to be tweaked @Andarist - I'm comparing to `MachineContext` since it's the default type if none other is specified. Maybe we should allow `never` here?

packages/core/src/types.ts
```diff
@@ -1005,23 +1005,26 @@ export type ContextFactory<TContext extends MachineContext> = ({
...
+    ? { context?: InitialContext<LowInfer<TContext>> }
```
Andarist, line 1025: In this branch, we can skip the other wrappers:
```suggestion
    ? { context?: {} }
```
alternatively we can do this since at this point it's guaranteed to be equal to the constraint:
```suggestion
    ? { context?: TContext }
```

packages/core/src/types.ts
Andarist: I'm a little bit unsure what kind of effect this intersection might have on inference, gonna have to recheck this. Other than that, this looks pretty OK.

davidkpiano:
Merging this in since the relevant parts of the change were reviewed/resolved

PR #4050, opened 2023-05-31
[v5] `@xstate/react` add `options` to `createActorContext`

davidkpiano:


packages/xstate-react/src/useActorRef.ts
```diff
@@ -84,27 +84,27 @@ export function useActorRef<TLogic extends AnyActorLogic>(
   machine: TLogic,
   ...[options = {}, observerOrListener]: RestParams<TLogic>
 ): ActorRefFrom<TLogic> {
-  const service = useIdleInterpreter(machine, options);
+  const interpretedActor = useIdleInterpreter(machine, options);
```
Andarist, line 87: In lots of places we use `actorRef`/`actor` for this already:
```suggestion
  const actorRef = useIdleInterpreter(machine, options);
```

packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -326,4 +326,114 @@ describe('createActorContext', () => {
...
+
+    expect(screen.getByTestId('value').textContent).toBe('b');
+
+    // unrender app
+    unmount();
+
+    // re-render app
+    render(<App />);
+
+    expect(screen.getByTestId('value').textContent).toBe('b');
```
Andarist, line 437: it would be cool to have a comment here to clarify what is even being tested here

Andarist: it feels a little indirect - we can create a persisted state out of the actor, before even rendering a component and then pass that persisted state in the very first render and immediately assert what we want

packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -326,4 +326,114 @@ describe('createActorContext', () => {
...
+    fireEvent.click(screen.getByTestId('button'));
+
+    expect(screen.getByTestId('count').textContent).toBe('1');
+
+    fireEvent.click(screen.getByTestId('button'));
+
+    expect(screen.getByTestId('count').textContent).toBe('2');
+  });
+
+  it('options can be passed to the provider', () => {
```
Andarist, line 381:
```suggestion
  it('persisted state can be passed to the provider', () => {
```

and I'd add a test that would check that we can pass `input` in the same way

packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -326,4 +326,114 @@ describe('createActorContext', () => {
       expect(screen.getByTestId('value').textContent).toBe('42');
     });
   });
+
+  it('should expect a constant machine reference', () => {
```
Andarist, line 330: this seems to be tested indirectly, could we either adjust the test case (by using more targeted assertions) or adjust this test title somehow?

packages/xstate-react/test/createActorContext.test.tsx
davidkpiano: I like doing it as real-life as possible; chances are developers will _not_ construct a state out of the actor manually.

packages/xstate-react/test/createActorContext.test.tsx
davidkpiano: Open to suggestions here

packages/xstate-react/test/createActorContext.test.tsx
Andarist: I like the "as real-life as possible" approach but I'm not sure that this is it. The user is unlikely to use this kind of remounting in their app. In addition to that... this test gives a false impression to the reader that it's OK to feed the snapshot state like this as persisted state. That's not OK, or at least not OK in some cases. 

You are typing `state` as `any` here (the option is also typed as `any` right now, see [here](https://github.com/statelyai/xstate/blob/69dc9c1a5d14ddc6be9263edc17f4811ddbc1e5e/packages/core/src/types.ts#L1562) so this looks alright-ish. But in fact what you assign to it is of `TSnapshot` type~ whereas what is expected to be provided to the `SomeContext.Provider` is of `TPersistedState` type~. Those are not synonymous and can't be mixed-and-matched - an explicit conversion step should always be required here (and this test should take this into account).

packages/xstate-react/test/createActorContext.test.tsx
Andarist:
```suggestion
  it("should preserve machine's identity when swapping options using in-render `.provide`", () => {
```

packages/xstate-react/src/createActorContext.ts
```diff
@@ -32,49 +32,52 @@ type ToMachinesWithProvidedImplementations<TMachine extends AnyStateMachine> =
...
-  actorLogic: TMachine,
-  interpreterOptions?: InterpreterOptions<TMachine>,
+export function createActorContext<TLogic extends AnyActorLogic>(
+  actorLogic: TLogic,
+  interpreterOptions?: InterpreterOptions<TLogic>,
   observerOrListener?:
-    | Observer<SnapshotFrom<TMachine>>
-    | ((value: SnapshotFrom<TMachine>) => void)
+    | Observer<SnapshotFrom<TLogic>>
+    | ((value: SnapshotFrom<TLogic>) => void)
```
Andarist, line 40: We have no tests utilizing those 2 arguments. I wonder if we need `observerOrListener` here - it feels weird here really, since this such a listener feels instance-specific and `createActorContext` doesn't actually create an instance of an actor.

packages/xstate-react/test/createActorContext.test.tsx
davidkpiano: Fixed

davidkpiano:
@Andarist This is ready for final review

davidkpiano:
@Andarist Merging this once tests pass

PR #4053, opened 2023-06-01
[v5] Serverless workflow examples

davidkpiano:
This PR implements serverless workflow examples found here: https://github.com/serverlessworkflow/specification/tree/main/examples

To run these examples, in any `-workflow` directory:

1. `pnpm i`
2. `pnpm start`
3. That's it!


examples/event-based-service-workflow/main.ts
```diff
@@ -0,0 +1,140 @@
...
+//       eventRef:
+//        produceEventRef: MakeVetAppointment
+//        data: "${ .patientInfo }"
+//        consumeEventRef: VetAppointmentInfo
+//       actionDataFilter:
+//        results: "${ .appointmentInfo }"
+//    timeouts:
+//     actionExecTimeout: PT15M
+//    end: true
+
```
Andarist, line 41: looks like a leftover
```suggestion
```

examples/reusing-functions-workflow/main.ts
```diff
@@ -0,0 +1,257 @@
...
+//   actions:
+//   - functionRef:
+//       refName: sendInsufficientFundsEmail
+//       arguments:
+//         applicant: "${ .customer }"
+//   end:
+//     produceEvents:
+//     - eventRef: ConfirmationCompletedEvent
+//       data: "${ .payment }"
+
```
Andarist, line 71: looks like a leftover
```suggestion
```

examples/send-cloudevent-workflow/main.ts
```diff
@@ -0,0 +1,123 @@
...
+//   actions:
+//   - functionRef:
+//       refName: provisionOrderFunction
+//       arguments:
+//         order: "${ .singleorder }"
+//   end:
+//     produceEvents:
+//     - eventRef: provisioningCompleteEvent
+//       data: "${ .provisionedOrders }"
+
```
Andarist, line 31: looks like a leftover
```suggestion
```

Andarist:
Ye, it's just that it's likely hard to categorize thinks in those Next examples whereas this "group" of examples that is being added here is pretty focused on one thing~. I'm fine either way but at the very least i would propose renaming those directories to have some kind of the same prefix, like those examples could start with `workflow-*` or something like that.

PR #4054, opened 2023-06-03
[v5] Input types

davidkpiano:
This PR strongly types `input` values:

```ts
  types: {} as {
    input: { count: number }
  },
  context: ({ input }) => ({
    value: input.count // strongly-typed
  })
});

interpret(machine, {
  input: { count: 42 } // strongly-typed
}).start();
```

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -125,9 +125,8 @@ export class StateNode<
   /**
    * The output data sent with the "done.state._id_" event if this is a final state node.
    */
-  public output?:
-    | Mapper<TContext, TEvent, any>
-    | PropertyMapper<TContext, TEvent, any>;
+  public output?: Mapper<TContext, TEvent, any>;
+
   /**
    * The order this state node appears. Corresponds to the implicit document order.
    */
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -143,7 +143,7 @@ export class StateNode<
     /**
      * The raw config used to create the machine.
      */
-    public config: StateNodeConfig<TContext, TEvent, TODO, TODO, TODO>,
+    public config: StateNodeConfig<TContext, TEvent, TODO, TODO>,
     options: StateNodeOptions<TContext, TEvent>
   ) {
     this.parent = options._parent;
@@ -169,7 +169,7 @@ export class StateNode<
         ? mapValues(
             this.config.states,
             (
-              stateConfig: StateNodeConfig<TContext, TEvent, TODO, TODO, TODO>,
+              stateConfig: StateNodeConfig<TContext, TEvent, TODO, TODO>,
               key
             ) => {
               const stateNode = new StateNode(stateConfig, {
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -117,7 +117,7 @@ export class StateNode<
   /**
    * The root machine node.
    */
-  public machine: StateMachine<TContext, TEvent, any, any>;
+  public machine: StateMachine<TContext, TEvent, any, any, TODO, TODO>;
   /**
    * The meta data associated with this state node, which will be returned in State instances.
    */
@@ -143,7 +143,7 @@ export class StateNode<
     /**
      * The raw config used to create the machine.
      */
-    public config: StateNodeConfig<TContext, TEvent, TODO, TODO>,
+    public config: StateNodeConfig<TContext, TEvent, TODO, TODO, TODO>,
     options: StateNodeOptions<TContext, TEvent>
   ) {
     this.parent = options._parent;
@@ -169,7 +169,7 @@ export class StateNode<
         ? mapValues(
             this.config.states,
             (
-              stateConfig: StateNodeConfig<TContext, TEvent, TODO, TODO>,
+              stateConfig: StateNodeConfig<TContext, TEvent, TODO, TODO, TODO>,
               key
             ) => {
               const stateNode = new StateNode(stateConfig, {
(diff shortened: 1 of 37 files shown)


packages/core/src/StateNode.ts
```diff
@@ -104,7 +104,7 @@ export class StateNode<
   /**
    * The root machine node.
    */
-  public machine: StateMachine<TContext, TEvent, any, any>;
+  public machine: StateMachine<TContext, TEvent, any, any, TODO, TODO>;
```
mellson, line 107: First time I've seen a `<TODO>` generic type, brilliant 😆

packages/core/test/input.test.ts
```diff
@@ -94,8 +135,15 @@ describe('input', () => {
...
-        expect(event.input.greeting).toBe('hello');
+      types: {} as {
+        input: { greeting: string };
+        context: { greeting: string };
+      },
+      context({ input }) {
+        return input;
+      },
+      entry: ({ context }) => {
+        expect(context.greeting).toBe('hello');
```
Andarist, line 146:
```suggestion
        expect(context.greeting).toBe('hello');
        expect(event.input.greeting).toBe('hello');
```

packages/core/test/input.test.ts
```diff
@@ -76,8 +112,13 @@ describe('input', () => {
...
     const invokedMachine = createMachine({
-      entry: ({ event }) => {
-        expect(event.input.greeting).toBe('hello');
+      types: {} as {
+        input: { greeting: string };
+        context: { greeting: string };
+      },
+      context: ({ input }) => input,
+      entry: ({ context }) => {
+        expect(context.greeting).toBe('hello');
```
Andarist, line 121:
```suggestion
        expect(context.greeting).toBe('hello');
        expect(event.input.greeting).toBe('hello');
```

packages/core/test/input.test.ts
```diff
@@ -21,29 +31,55 @@ describe('input', () => {
...
+        wrongCountProperty: 42
+      }
+    });
+
     interpret(machine, { input: { startCount: 42 } }).start();
 
     expect(spy).toHaveBeenCalledWith(42);
   });
 
   it('initial event should have input property', (done) => {
```
Andarist, line 46: this test case no longer tests what it should :wink:

davidkpiano: Yeah, we shouldn't rely on the event having the input data. That should be saved to `context` and read there as a best practice

Andarist: As a best practice - sure, that's probably for the better. As it stands though, this is also part of the init event (creating a nice symmetry with "done" event that contains `output` and we can't easily remove it from there). We should be careful when adjusting tests - we don't want to lose valuable tests and assertions. Even if we'd like to remove `input` from the init event then this whole test should be gone (or adjusted to document that there is only `type` property on the init event).

Andarist: Keeping `input` on the init event has also one pretty nice advantage - it will be easier to keep this information around in the inspector tools etc

packages/core/src/types.ts
```diff
@@ -1186,7 +1202,11 @@ export type Mapper<
   TContext extends MachineContext,
   TEvent extends EventObject,
   TParams extends {}
-> = (args: { context: TContext; event: TEvent }) => TParams;
+> = (args: {
+  context: TContext;
+  event: TEvent;
+  self: ActorRef<TEvent>;
```
Andarist, line 1208: this has been added here without any test coverage 😢

davidkpiano: Sorry! I'll add tests

packages/core/src/utils.ts
```diff
@@ -231,10 +232,11 @@ export function mapContext<
 >(
   mapper: Mapper<TContext, TEvent, any> | PropertyMapper<TContext, TEvent, any>,
   context: TContext,
-  event: TEvent
+  event: TEvent,
+  self: AnyActorRef
 ): any {
   if (typeof mapper === 'function') {
-    return mapper({ context, event });
+    return mapper({ context, event, self });
```
Andarist, line 239: `self` is not passed here to the property mapper (`subMapper`) that is called below

davidkpiano:
@Andarist I removed the property mapper type from `output` (was never on `input`) and added a dev-time warning:

```
  console.warn
    Dynamically mapping values to individual properties is deprecated. Use a single function that returns the mapped object instead.
    Found object containing properties whose values are possibly mapping functions: 
     - user: ({event}) => event.user
```

davidkpiano:
@Andarist Feel free to approve or merge when ready

packages/core/test/final.test.ts
```diff
@@ -113,7 +113,7 @@ describe('final states', () => {
             reveal: {
               type: 'final',
               output: {
-                secret: () => 'the secret'
+                secret: 'the secret'
```
Andarist, line 116: This test no longer tests what it should (`'should call data expressions on nested final nodes'`). To fix this we could make the whole `output` an expression here

packages/core/src/utils.ts
```diff
@@ -239,20 +239,27 @@ export function mapContext<
...
-  for (const key of Object.keys(mapper)) {
-    const subMapper = mapper[key];
-
-    if (typeof subMapper === 'function') {
-      result[key] = subMapper(args);
-    } else {
-      result[key] = subMapper;
+  if (isDevelopment) {
+    if (typeof mapper === 'object') {
+      if (Object.values(mapper).some((val) => typeof val === 'function')) {
```
Andarist, line 244: nit: this could be a single if statement: `a && b && c`

PR #4055, opened 2023-06-03
[v5] System for all actor logic

davidkpiano:
This PR provides access to `system` for all actor logic.

```ts
fromPromise(({ system }) => { ... });

fromTransition((state, event, { system }) => { ... });

fromObservable(({ system }) => { ... });

fromEventObservable(({ system }) => { ... });

fromCallback((sendBack, receive, { system }) => { ... });
```

### DIFF
--- a/packages/core/src/actors/observable.ts
+++ b/packages/core/src/actors/observable.ts
@@ -1,4 +1,10 @@
-import { Subscribable, ActorLogic, EventObject, Subscription } from '../types';
+import {
+  Subscribable,
+  ActorLogic,
+  EventObject,
+  Subscription,
+  AnyActorSystem
+} from '../types';
 import { stopSignalType } from '../actors';
 
 export interface ObservableInternalState<T> {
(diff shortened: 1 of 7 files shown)


packages/core/test/system.test.ts
```diff
@@ -269,4 +274,53 @@ describe('system', () => {
...
+        }
+      ]
+    });
+
+    const actor = interpret(machine).start();
+
+    expect(actor.system.get('test')).toBeDefined();
+
+    // The assertion won't be checked until the transition function gets an event
+    actor.system.get('reducer')!.send({ type: 'a' });
```
Andarist, line 324: why is that? 🤔

packages/core/test/system.test.ts
```diff
@@ -269,4 +274,53 @@ describe('system', () => {
 
     interpret(machine).start();
   });
+
+  it('should be accessible in other actors', () => {
+    expect.assertions(6);
```
Andarist, line 279: could u split this test into 6 separate tests?

mellson (approved):
I guess my comment is just a type-level nit, so there is no reason why you shouldn't get an approval

packages/core/src/actors/observable.ts
```diff
@@ -131,7 +146,12 @@ export function fromObservable<T, TEvent extends EventObject>(
  */
 
 export function fromEventObservable<T extends EventObject>(
-  lazyObservable: ({ input }: { input: any }) => Subscribable<T>
+  lazyObservable: ({
+    input
```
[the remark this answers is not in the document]
davidkpiano: Will do this in a follow-up PR, good idea

PR #4056, opened 2023-06-03
[v5] Simulate

davidkpiano:
This PR introduces a `simulate` function that creates a dummy actor context for safe simulation.

```ts
import { simulate } from 'xstate';

// instead of:
// logic.transition(state, event, someActorContext);
simulate(logic).transition(state, event);

simulate(logic).getInitialState();

simulate(logic).microstep(state, event);
```


packages/core/src/simulate.ts
```diff
@@ -0,0 +1,50 @@
+import {
+  AnyActorLogic,
+  AnyStateMachine,
+  EventFromLogic,
+  InternalStateFrom,
+  SnapshotFrom
```
mellson, line 6:
```suggestion
```

packages/core/src/simulate.ts
```diff
@@ -0,0 +1,50 @@
...
+} from '.';
+
+export function simulate<T extends AnyActorLogic>(
+  actorLogic: T,
+  options: { input?: any } = {}
+) {
+  const dummyActorContext: any = {
+    self: {}
+  };
```
Andarist, line 15: This definitely needs more to be put on it - we need to satisfy the real `ActorContext` type to avoid mistakes further down the road. While working on the other PR I had something like this for it:
```ts
function createMockActorContext(): AnyActorContext {
  const nullActor = createEmptyActor();
  return {
    self: nullActor,
    logger: console.log,
    id: 'root_test',
    sessionId: Math.random().toString(32).slice(2),
    defer: () => {},
    system: nullActor,
    stopChild: () => {}
  };
}
```

packages/xstate-graph/src/graph.ts
```diff
@@ -93,9 +94,7 @@ export function createDefaultMachineOptions<TMachine extends AnyStateMachine>(
         })
       ) as any[];
     },
-    fromState: machine.getInitialState(
-      undefined as any // TODO: figure out the simulation API
-    ) as StateFrom<TMachine>,
+    fromState: simulate(machine).getInitialState() as StateFrom<TMachine>,
```
mellson, line 97:
```suggestion
    fromState: simulate(machine).getInitialState(),
```

mellson (approved):
Great work, it looks good to me. But I'm still new to this codebase, so take it with a pinch of salt 🤏🏻 🧂

packages/core/src/simulate.ts
```diff
@@ -0,0 +1,62 @@
...
+        options.input
+      ),
+      event: EventFromLogic<T>
+    ): InternalStateFrom<T> => {
+      return actorLogic.transition(state, event, dummyActorContext);
+    },
+    getInitialState: () => {
+      return actorLogic.getInitialState(dummyActorContext, options.input);
+    },
+    microstep: (
```
Andarist, line 42: Im not sure if this should be here. It’s specific to machines and it’s not even expressed at type level correctly~ right now. Maybe we can just squint are eyes and accept this duck typing but it would be cool if we wouldnt have to do it

packages/core/test/deterministic.test.ts
```diff
@@ -97,69 +97,56 @@ describe('deterministic machine', () => {
...
     });
 
     it('should transition to nested states as target', () => {
       expect(
-        testMachine.transition(
-          testMachine.resolveStateValue('a'),
-          { type: 'T' },
-          undefined as any // TODO: figure out the simulation API
-        ).value
+        simulate(testMachine).transition(testMachine.resolveStateValue('a'), {
```
Andarist, line 109: Do we have any ideas on how to put this kind of state resolution in the simulation API? Or do we just want to keep API and require people to resolve state values outside of the simulation API?

packages/core/test/deterministic.test.ts
```diff
@@ -97,69 +97,56 @@ describe('deterministic machine', () => {
...
     it('should use the machine.initialState when an undefined state is given', () => {
       expect(
-        lightMachine.transition(
-          lightMachine.getInitialState(
-            undefined as any // TODO: figure out the simulation API
-          ),
-          { type: 'TIMER' },
-          undefined as any // TODO: figure out the simulation API
-        ).value
+        simulate(lightMachine).transition(undefined, {
```
Andarist, line 139: This `undefined` here looks really suspicious. Im reviewing on mobile so i have a hard time rechecking this across files but it feels like this shouldnt be allowed

davidkpiano: "Sugar" for initial state

packages/core/test/deterministic.test.ts
davidkpiano: Ideally `.resolveStateValue` is seldom used in real-world applications, in favor of real state. So, the latter is fine.

packages/core/src/simulate.ts
davidkpiano: Removed

packages/core/test/deterministic.test.ts
Andarist: I'm not super fond of this - partially because this moves the `input` argument from the `simulate.getInitialState` to the `simulate` itself. This feels weird/out of place because it diverges from the "core" API in an unexpected way.

If this sugar wouldn't be here then `input` could be removed from the `simulate` and there would be a nice "symmetry" between those:
```ts
machine.getInitialState(actorContext, input)
simulate(machine).getInitialState(input)
```

The only difference here would be the implied/provided actor context but other than that the API would stay almost the same.

packages/xstate-graph/src/graph.ts
```diff
@@ -93,9 +94,7 @@ export function createDefaultMachineOptions<TMachine extends AnyStateMachine>(
         })
       ) as any[];
     },
-    fromState: machine.getInitialState(
-      undefined as any // TODO: figure out the simulation API
-    ) as StateFrom<TMachine>,
+    fromState: simulate(machine).getInitialState(),
```
Andarist, line 97: I this this is somewhat wrong because it creates a fresh (and disposable) simulation object. 

Consider those lines:
https://github.com/statelyai/xstate/blob/ede42d88c521a5c8d1acd4ccdbf4032eefb2f023/packages/xstate-graph/src/pathFromEvents.ts#L40-L45

Both `fromState = ` and `getAdjacencyMap` might create the simulation object and thus 2 different dummy actor contexts. If things would, anyhow, rely on the (widely understood) identity of the actor context then things could break in some subtle ways because the same actor context/simulation object is not reused consistently throughout `getPathsFromEvents` (the function containing the quoted lines of code).

I think that a good part of the reason behind introducing `simulate` is the fact that we can create the actor context on behalf of the user and pass it around correctly from the very start to the very end of the "simulation".

packages/core/src/simulate.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const sim = {
+    transition: (
+      state: InternalStateFrom<T> = actorLogic.getInitialState(
+        dummyActorContext,
+        options.input
+      ),
+      event: EventFromLogic<T>
+    ): InternalStateFrom<T> => {
+      return actorLogic.transition(state, event, dummyActorContext);
```
Andarist, line 36: Through `@xstate/test` we are kinda branching off different potential paths while simulating. `actorContext` provides access to a mutable resource that is being shared among many "timelines" created by branching those paths etc. This sounds a little bit disastrous... but I'm also not quite sure yet what we should do about it here etc. It would be easier if people just... wouldnt use anything from this context when using `@xstate/test` 😅 but that's hard to forbid as well.

For the time being, I'm just mentioning this to raise awareness that this is a little bit dangerous - but I don't have any proposals at all as to what we should do here instead.

packages/core/src/index.ts
```diff
@@ -41,7 +41,7 @@ export {
 } from './actors/index.ts';
 
 export { stateIn, not, and, or } from './guards.ts';
-
+export { simulate } from './simulate.ts';
```
Andarist, line 44: I'm not super sure about the current shape of this API. I also don't have any concrete feedback about what could be improved here though. I like how we use it internally right now (for the most part), since we can hide some tedious stuff in this lil helper. I'm not sold on exposing it publicly as a stable API though, it feels somewhat 0.1/beta-ish. It would be cool if it could do **more** but at the moment I don't know what that more is. Take it with a grain of salt, it's mostly a feedback based on a vague gut feeling and not on concrete technical aspects. 

It also feels that really this API is mostly useful to machines. Yes, I know that one could maybe implement other logics that would switch their internal implementation to some kind of a "pure variant" when a flag/option/whatever would be passed to that actor logic but that doesn't really sound appealing to me at all. If one wants to simulate a different type of an actor logic then swapping its reference or something to a completely different pure/mocked variant sounds pretty reasonable to me. 

We have a luxury of machines being pure-ish by default and it's super tempting to just reuse the whole thing publicly for both pure and impure use cases but I feel like we are hitting some conceptual limitations with that approach because it's hard to satisfy **both** worlds at the same time. It almost feels to me that `createTestMachine`, with its own types, simplified API, slightly limited but yet operating on the same core concepts... is just a much cleaner solution. 

That doesn't exactly preclude existence of/need for a `simulate`/`simulateMachine` API. Perhaps the use cases of `@xstate/test` and simulation are slightly different? They certainly feel like the same thing but for instance `simulate` should be able to do something with invoked children (even if that thing is just to "record" that they have been created) whereas in `@xstate/test` it doesn't make a lot of sense to have any invokes at all (current `createTestMachine` actually forbids `invoke`).

davidkpiano: The problem is that if we don't export this, importing them in other package tests gets a lot trickier.

I think we should export this while in beta but not document it. I actually want to get rid of `createTestMachine` because `@xstate/graph` and `@xstate/test` should operate with any behavior, not just a special kind of machine.

Otherwise, what's the best way to move forward with this PR?

davidkpiano: Ping @Andarist

packages/core/test/deterministic.test.ts
davidkpiano: Removed the sugar

davidkpiano:
Superseded by https://github.com/statelyai/xstate/pull/4596

PR #4058, opened 2023-06-06
Update `state.children` directly when resolving `stop` action

Andarist:


packages/core/src/actions/stop.ts
```diff
@@ -53,8 +54,16 @@ export function stop<
...
+      if (actorRef) {
+        children = { ...children };
+        delete children[actorRef.id];
+      }
+
       return [
-        state,
+        cloneState(state, {
+          children
+        }),
```
Andarist, line 66: making it more consistent with how new `children` are added by `invoke`'s resolver:
https://github.com/statelyai/xstate/blob/ceae8227ab19a3b6a0c1c895f541246e0f33395c/packages/core/src/actions/invoke.ts#L84-L89

davidkpiano (approved):

PR #4062, opened 2023-06-07
Remove `State['event']`

Andarist:


packages/xstate-test/test/dieHard.test.ts
```diff
@@ -260,53 +260,54 @@ describe('error path trace', () => {
       initial: 'first',
       states: {
         first: {
-          on: { NEXT: 'second' }
+          on: { NEXT_1: 'second' }
         },
         second: {
-          on: { NEXT: 'third' }
+          on: { NEXT_2: 'third' }
```
davidkpiano, line 266: Will the test work without these being renamed?

davidkpiano (approved):
Just needs a changeset

PR #4064, opened 2023-06-08
[v5] Allow guard objects to reference other guard objects

davidkpiano:


packages/core/test/stateIn.test.ts
```diff
@@ -424,4 +424,44 @@ describe('transition "in" check', () => {
     actorRef.send({ type: 'TIMER' });
     expect(actorRef.getSnapshot().value).toEqual('green');
   });
+
+  it('should allow a guard object to reference another guard object', () => {
```
Andarist, line 428: I'd move this test to a new test dedicated to `stateIn`, something like:
```suggestion
  it('should be possible to use a referenced `stateIn` guard', () => {
```

It's an important, basic~, case that needs to work and I think it makes sense to have a specific test for it.

The current test title is good as well, I'd just move it to a different test suite (since this one is in `stateIn.tes.ts`) and use some random~ custom guards in it.

packages/core/src/guards.ts
```diff
@@ -108,14 +108,28 @@ export function toGuardDefinition<
   TEvent extends EventObject
 >(
   guardConfig: GuardConfig<TContext, TEvent>,
-  getPredicate?: (guardType: string) => GuardPredicate<TContext, TEvent>
+  getPredicate?: (
+    guardType: string
+  ) => GuardPredicate<TContext, TEvent> | GuardDefinition<TContext, TEvent>
 ): GuardDefinition<TContext, TEvent> {
+  // TODO: do this recursively and check for cycles
   if (isString(guardConfig)) {
```
Andarist, line 116: All of this feels a tad complicated - but it's not exactly a unique problem to this function (similar things happen when "resolving" actions).

I think that the main mistake that we do here is that we try to resolve (through `toGuardDefinition`) too eagerly. We always need to have this resolution~ step when finally checking guards since we need to be inside the execution context of a machine. It's kinda pointless to resolve this twice.

Let's roll with this for now but I think that this should be flagged as something to refactor.

Andarist (commented):
Don't forget about a changeset! 😉

packages/core/src/guards.ts
```diff
@@ -108,14 +108,28 @@ export function toGuardDefinition<
   TEvent extends EventObject
 >(
   guardConfig: GuardConfig<TContext, TEvent>,
-  getPredicate?: (guardType: string) => GuardPredicate<TContext, TEvent>
+  getPredicate?: (
+    guardType: string
+  ) => GuardPredicate<TContext, TEvent> | GuardDefinition<TContext, TEvent>
 ): GuardDefinition<TContext, TEvent> {
+  // TODO: do this recursively and check for cycles
```
davidkpiano, line 115:
```suggestion
  // TODO: do this recursively and check for cycles - REFACTOR
```

.changeset/red-berries-occur.md
```diff
@@ -0,0 +1,30 @@
+---
+'xstate': major
```
Andarist, line 2:
```suggestion
'xstate': patch
```

packages/core/test/guards.test.ts
```diff
@@ -555,6 +555,70 @@ describe('referencing guards', () => {
...
+      }
+    );
+
+    const actorRef = interpret(machine).start();
+    actorRef.send({ type: 'EVENT' });
+
+    expect(actorRef.getSnapshot().matches('b')).toBeTruthy();
+  });
+
+  it('should be possible to reference a composite guard that references other guards recursively', () => {
```
Andarist, line 588: @davidkpiano you noted down in a code comment that we should handle things recursively but I pushed out some new tests here and this things seems to work. Am I missing some cases that you were thinking about?

davidkpiano: This still only goes one level deep; try something like:

```ts
guards: {
  ref1: { type: 'ref2' },
  ref2: { type: 'ref3' },
  ref3: () => true
}
```

Andarist (approved):

PR #4065, opened 2023-06-08
Add support for wildcard transitions to xstate-fsm

benblank:
After a recent discussion regarding how to handle unexpected events (#4057) brought wildcard transitions to my attention, I wanted to see what it would take to add them to the @xstate/fsm package. Fortunately, it turned out to be pretty straightforward.

This PR introduces three behavioral changes, in addition to the supporting tests and documentation:

1. If a machine is created with transitions whose event type is a single asterisk, those transitions are added to the list of transitions attempted regardless of event type. This has the effect that, if no other transition has occurred, the wildcard transitions are checked before giving up and returning an unchanged state.
2. To prevent the above from causing typing problems, a single asterisk is considered a valid key in the transitions objects in the machine config.
3. If the transition method is called with an event whose type is a single asterisk, it will throw in non-production mode. (This matches the behavior of the full XState library.)

(Also, while updating "docs/packages/xstate-fsm/index.md", I noticed that it had gotten out of sync with "packages/xstate-fsm/README.md"; I corrected that while I was in there, but as a separate commit, so that it can be easily dropped if you'd prefer to keep the PR clean.)

### DIFF
--- a/.changeset/tall-flies-occur.md
+++ b/.changeset/tall-flies-occur.md
@@ -1,5 +1,5 @@
 ---
-'@xstate/fsm': major
+'@xstate/fsm': minor
 ---
 
 This change adds support for using "\*" as a wildcard event type in machine configs.
--- a/.changeset/tall-flies-occur.md
+++ b/.changeset/tall-flies-occur.md
@@ -1,0 +1,7 @@
+---
+'@xstate/fsm': major
+---
+
+This change adds support for using "\*" as a wildcard event type in machine configs.
+
+Because that event type previously held no special meaning, it was allowed as an event type both in configs and when transitioning and matched as any other would. As a result of changing it to be a wildcard, any code which uses "\*" as an ordinary event type will break, making this a major change.
--- a/packages/xstate-fsm/src/types.ts
+++ b/packages/xstate-fsm/src/types.ts
@@ -136,7 +136,7 @@ export namespace StateMachine {
     states: {
       [key in TState['value']]: {
         on?: {
-          [K in TEvent['type']]?: SingleOrArray<
+          [K in TEvent['type'] | '*']?: SingleOrArray<
             Transition<
               TContext,
               TEvent extends { type: K } ? TEvent : never,
--- a/packages/xstate-fsm/src/types.ts
+++ b/packages/xstate-fsm/src/types.ts
@@ -139,7 +139,11 @@ export namespace StateMachine {
           [K in TEvent['type'] | '*']?: SingleOrArray<
             Transition<
               TContext,
-              TEvent extends { type: K } ? TEvent : never,
+              K extends '*'
+                ? TEvent
+                : TEvent extends { type: K }
+                ? TEvent
+                : never,
               TState['value']
             >
           >;
--- a/packages/xstate-fsm/src/types.ts
+++ b/packages/xstate-fsm/src/types.ts
@@ -137,15 +137,13 @@ export namespace StateMachine {
       [key in TState['value']]: {
         on?: {
           [K in TEvent['type'] | '*']?: SingleOrArray<
-            Transition<
-              TContext,
-              K extends '*'
-                ? TEvent
-                : TEvent extends { type: K }
-                ? TEvent
-                : never,
-              TState['value']
-            >
+            K extends '*'
+              ? Transition<TContext, TEvent, TState['value']>
+              : Transition<
+                  TContext,
+                  TEvent extends { type: K } ? TEvent : never,
+                  TState['value']
+                >
           >;
         };
         exit?: SingleOrArray<Action<TContext, TEvent>>;
--- a/packages/xstate-fsm/test/fsm.test.ts
+++ b/packages/xstate-fsm/test/fsm.test.ts
@@ -131,6 +131,88 @@ describe('@xstate/fsm', () => {
     expect(nextState.actions).toEqual([]);
   });
 
+  describe('when a wildcard transition is defined', () => {
+    type ParserContext = { entries: Array<Record<string, string>> };
+
+    type ParserEvent =
+      | { type: 'toggle' }
+      | { type: 'data' }
+      | { type: 'finish' };
+
+    type ParserState =
+      | { value: 'waiting'; context: ParserContext }
+      | { value: 'reading'; context: ParserContext }
+      | { value: 'done'; context: ParserContext }
+      | { value: 'error'; context: ParserContext };
+
+    const parserConfig: StateMachine.Config<
+      ParserContext,
+      ParserEvent,
+      ParserState
+    > = {
+      id: 'parser',
+      initial: 'reading',
+      context: { entries: [] },
+      states: {
+        waiting: {
+          on: {
+            toggle: 'reading',
+            finish: 'done',
+            '*': 'error'
+          }
+        },
+        reading: {
+          on: {
+            toggle: 'waiting',
+            data: 'reading',
+            finish: {
+              target: 'done',
+              cond: (context) => Object.keys(context.entries ?? {}).length > 0
+            },
+            '*': 'error'
+          }
+        },
+        done: { on: { '*': 'error' } },
+        error: {}
+      }
+    };
+
+    const parserFSM = createMachine(parserConfig);
+
+    it('should not use a wildcard when an unguarded transition matches', () => {
+      const nextState = parserFSM.transition(parserFSM.initialState, 'toggle');
+      expect(nextState.value).toBe('waiting');
+    });
+
+    it('should not use a wildcard when a guarded transition matches', () => {
+      const currentState = {
+        ...parserFSM.initialState,
+        context: { entries: [{ foo: 'bar' }] }
+      };
+      const nextState = parserFSM.transition(currentState, 'finish');
+      expect(nextState.value).toBe('done');
+    });
+
+    it('should use a wildcard when no guarded transition matches', () => {
+      const nextState = parserFSM.transition(parserFSM.initialState, 'finish');
+      expect(nextState.value).toBe('error');
+    });
+
+    it('should use a wildcard when no transition matches', () => {
+      const nextState = parserFSM.transition(
+        parserFSM.initialState,
+        'FAKE' as any
+      );
+      expect(nextState.value).toBe('error');
+    });
+
+    it("should throw an error when an event's type is the wildcard", () => {
+      expect(() =>
+        parserFSM.transition(parserFSM.initialState, '*' as any)
+      ).toThrow();
+    });
+  });
+
   it('should throw an error for undefined states', () => {
     expect(() => {
       lightFSM.transition('unknown', 'TIMER');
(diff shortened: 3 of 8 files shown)


packages/xstate-fsm/src/types.ts
```diff
@@ -136,7 +136,7 @@ export namespace StateMachine {
     states: {
       [key in TState['value']]: {
         on?: {
-          [K in TEvent['type']]?: SingleOrArray<
+          [K in TEvent['type'] | '*']?: SingleOrArray<
             Transition<
               TContext,
               TEvent extends { type: K } ? TEvent : never,
```
Andarist, line 142: This bit here wont be correct anymore with this change. With `'*'` u dont want to "filter" from `TEvent` based on `{ 
type: K }`, you just want to grab `TEvent` as a whole

benblank: I'm afraid you've lost me. 🙃

If I understand the typings correctly, this change is from "valid keys are of type `K`, where `K` is any type assignable to the value of property `"type"` on type `TEvent`" to "…where `K` is *the union of* any type assignable to the value of property `"type"` on type `TEvent` *and `'\*'`*". The change is made here rather than on `TEvent["type"]` because `"\*"` has a valid meaning when describing a transition, but is **not** valid as an event type in general, such as when calling `Machine.transition(…)`.

If that doesn't address your concern, let me know!

Andarist: > If I understand the typings correctly, this change is from "valid keys are of type K, where K is any type assignable to the value of property "type" on type TEvent" to "…where K is the union of any type assignable to the value of property "type" on type TEvent and "*""

This change is correct. The problem was created here though because the "value position" of this mapped type has not been adjusted anyhow and it was defined (and still is) as:
```ts
Transition<
  TContext,
  TEvent extends { type: K } ? TEvent : never,
  TState['value']
>
```

If we substitute `K` for `'*'` here then we end up with:
```ts
Transition<
  TContext,
  TEvent extends { type: '*' } ? TEvent : never,
  TState['value']
>
```

So in this case we'll try to filter `TEvent` members that extend `{ type: '*' }` and such should never exist (we can't truly disallow them at type-level but we already disallow them at runtime with "If the transition method is called with an event whose type is a single asterisk, it will throw in non-production mode. (This matches the behavior of the full XState library.)")

What I'm saying is that we should end up here with something closer to this:
```ts
Transition<
  TContext,
  K extends '*' ? TEvent : TEvent extends { type: K } ? TEvent : never,
  TState['value']
>
```

Take a look at how we handle this in the XState's types:
https://github.com/statelyai/xstate/blob/754afa022047518ef4813f7aa85398218b39f960/packages/core/src/types.ts#L490-L494

benblank: Got it! I hadn't noticed earlier that `K` was being used deeper in the type. 🤦

I spent some time playing around with the types and in the end, I'm not sure how this should be addressed. If `K` were being used to specify the type for an event, it would certainly cause trouble, but because it's sole use is in an `extends` clause, the typing of `Transition<…>` doesn't actually change.

In fact, I wasn't able to convince myself that `TEvent extends { type: K } ? TEvent : never` can actually result in `never` here, with or without adding the wildcard. `TEvent` is constrained to `{ type: string }` via extending `EventObject`, which means that `TEvent["type"]` always extends `string`; `"*"` also extends `string`, which means that `K` does, as well. A subtype of `string` will always extend the union of that same subtype with any subtype of `string`. In other words, `("a" | "b" | "c") extends ("a" | "b" | "c" | "*") ? true : false` is always `true`.

Of course, knowing that requires more than just a glance at the types; having `K` show up in the type passed to `Transition<…>` after explicitly including `"*"` in `K` certainly suggests that it has some effect there, even though it doesn't. I can think of three ways to address that:

1. Remove `"*"` from the type in the conditional, e.g. `TEvent extends { type: Exclude<K, "*"> } ? TEvent : never`.
2. Separate the types of `on[K in TEvent['type']]` and `on["*"]`, though if I'm correct, they'd simply be identical.
3. Get rid of the whole conditional; change it to just `Transition<TContext, TEvent, TState["value"]>`, which I believe it's already equivalent to. Unsurprisingly, this is the one I'm most partial to. 🙂

I also put together a [CodePen sandbox](https://codesandbox.io/s/crazy-wing-yqhy2z) to demonstrate a bunch of this. Let me know what you think!

Andarist: The problem is that `K` is distributed through instantiations of a mapped type's template. That conditional type is not testing things like `("a" | "b" | "c") extends ("a" | "b" | "c" | "*")`, it just tests a **concrete** `K`, a single member of the union over which we are iterating there. Without that you wouldn't be able to write the simplest:
```ts
type KeysAsValues<T> = {
  [K in keyof T]: K
}

type Mapped = KeysAsValues<{ a: 1, b: 2 }> // actual: { a: "a", b: "b" }, without instantiated/distributed `K`: { a: "a" | "b", b: "a" | "b" }
```

> Of course, knowing that requires more than just a glance at the types; having K show up in the type passed to Transition<…> after explicitly including "*" in K certainly suggests that it has some effect there, even though it doesn't. 

The effect, and the fact that it is narrowed to `never` for wildcards, can be easily observed on this [TS playground](https://www.typescriptlang.org/play?ts=5.1.3#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzDigxAFkowALLVEAHgBUBRANxFQ3hAA8TVgAzvADe8DAE8ADiABc8ARhg0A5vAC+APgAUYPIizK5wgFDx4eI6bPwA2gGl4NeMzYcbAIgnT3AXXgAfeHcAKl8AfjktK2suFjkXdk4ePkERMSlZeAc1eDDnVkT4OVo2GGj4AEp4AF4NeBYcLGAAbis1VrUKuQam1uNCEGIyCmpaOlEvTPcAMQB5Wfd1ALTJuXcAIQBBACVFzS0TMwsRcrnZyJBXDCrak5izS8TW+-gAelf4AD0w8rUAGnKIXcFyuNzqh3ujw4z3u7y+Pxi-zaAM6rSAA)

>  I can think of three ways to address that

We should just route this through a conditional type to a branch that narrows down the event type or to the one that skips the narrowing and just forwards the whole `TEvent`, just like in the quoted types from the core module:
https://github.com/statelyai/xstate/blob/754afa022047518ef4813f7aa85398218b39f960/packages/core/src/types.ts#L490-L494

benblank: Bah; it's very frustrating that I wasn't able to come up with an example of adding `"*"` having an effect on `Transition<…>`. Thanks for providing one! I'll need to spend some time studying how TS handles concrete type variables.

In the meantime, I've updated the PR to fix the types; I was also finally able to add that related test I was looking for.

Thanks for helping me through this type stuff! I'm learning a lot. 🙂

benblank: For example, I just learned that VS Code wasn't fully honoring the package's TypeScript configuration. 🤦

Because the `TEvent` parameter to `Transition<…>` is only used in function signatures, the test doesn't work unless `"strictFunctionTypes"` is enabled in the TypeScript config. Unsurprisingly, enabling it for the entire repo blows up, but enabling it for just the @xstate/fsm package works just fine, so that's what I've done.

Let me know if that isn't a desirable change.

packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -131,6 +131,88 @@ describe('@xstate/fsm', () => {
     expect(nextState.actions).toEqual([]);
   });
 
+  describe('when a wildcard transition is defined', () => {
+    type ParserContext = { entries: Array<Record<string, string>> };
```
Andarist, line 135: I'd copy-paste this stuff into each added test and trim down redundant bits for each particular test. Don't be afraid of some repetition, embrace it! 😉

packages/xstate-fsm/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "include": ["src/**/*"],
   "compilerOptions": {
+    "strictFunctionTypes": true,
```
Andarist, line 5: To avoid merge conflicts with the `next` branch, I'd like to leave this off - even though that in general this is a good change. On the `next` branch we'll work on enabling this setting globally soon.

packages/xstate-fsm/test/types.test.ts
```diff
@@ -58,4 +59,31 @@ describe('matches', () => {
...
+      context: {},
+      initial: 'one',
+      states: {
+        one: {
+          on: {
+            foo: {
+              target: 'one',
+              actions: (_context: Context, _event: InitEvent | FooEvent) => {}
+            },
+            // @ts-expect-error
```
Andarist, line 79: to accompany [the other suggested change](https://github.com/statelyai/xstate/pull/4065/files#r1230935353) I'd change this to smth like:
```suggestion
            // @x-ts-expect-error
```

We'll "turn this on" at later point in time.

packages/xstate-fsm/src/types.ts
```diff
@@ -136,10 +136,14 @@ export namespace StateMachine {
     states: {
       [key in TState['value']]: {
         on?: {
-          [K in TEvent['type']]?: SingleOrArray<
+          [K in TEvent['type'] | '*']?: SingleOrArray<
             Transition<
               TContext,
-              TEvent extends { type: K } ? TEvent : never,
+              K extends '*'
```
Andarist, line 142: i think that by routing this earlier it becomes a little bit more readable, so I'd use this variant:
```ts
      on?: {
        [K in TEvent["type"] | "*"]?: SingleOrArray<
          K extends "*"
            ? Transition<TContext, TEvent, TState["value"]>
            : Transition<
                TContext,
                TEvent extends { type: K } ? TEvent : never,
                TState["value"]
              >
        >;
      };
```

packages/xstate-fsm/src/types.ts
Andarist: > For example, I just learned that VS Code wasn't fully honoring the package's TypeScript configuration. 🤦

Yeah, this is pretty annoying. I execute "Select TypeScript version..." command right after opening just any new project on my machine. There are ways to configure VS Code to prompt you to do this - they don't want to run the installed TS version without explicit consent for security reasons.

packages/xstate-fsm/tsconfig.json
benblank: Okay. I've reverted this change, disabled the `@ts-expect-error` comment as you suggested, and added a comment to the test saying that it isn't actually working at the moment.

packages/xstate-fsm/src/types.ts
benblank: 👍

.changeset/tall-flies-occur.md
```diff
@@ -0,0 +1,7 @@
+---
+'@xstate/fsm': major
```
Andarist, line 2:
```suggestion
'@xstate/fsm': minor
```

PR #4069, opened 2023-06-12
Add a reference to the root actor on the system object

kevinmaes:
This would allow any actor within the system to send events to the system's root actor without needing to know the `systemId` of of the root actor that was passed into `interpret(rootActor, { systemId: 'root-id' })`

`sendTo(({ system } ) => system.root, { type: 'to root', message: 'Message to root' });`

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -123,6 +123,8 @@ export class Interpreter<
 
   public src?: string;
 
+  public root: any;
+
   /**
    * Creates a new Interpreter instance (i.e., service) for the given logic with the provided options, if any.
    *
@@ -139,6 +141,7 @@ export class Interpreter<
     const self = this;
 
     this.system = parent?.system ?? createSystem();
+    this.system.root = this;
 
     if (systemId) {
       this._systemId = systemId;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1893,6 +1893,7 @@ export interface ActorSystem<T extends ActorSystemInfo> {
   _unregister: (actorRef: AnyActorRef) => void;
   _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
   get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+  root?: T['actors'][keyof T['actors']];
 }
 
 export type AnyActorSystem = ActorSystem<any>;
--- a/packages/core/test/system.test.ts
+++ b/packages/core/test/system.test.ts
@@ -126,6 +126,12 @@ describe('system', () => {
     expect(actor.system.get('test')).toBe(actor);
   });
 
+  it('root actor can be directly referenced by a root propoerty on the system', () => {
+    const machine = createMachine({});
+    const actor = interpret(machine, { systemId: 'test' });
+    expect(actor.system.root).toBe(actor);
+  });
+
   it('should remove invoked actor from receptionist if stopped', () => {
     const machine = createMachine({
       initial: 'active',


packages/core/src/types.ts
```diff
@@ -1893,6 +1893,7 @@ export interface ActorSystem<T extends ActorSystemInfo> {
   _unregister: (actorRef: AnyActorRef) => void;
   _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
   get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+  root?: T['actors'][keyof T['actors']];
```
kevinmaes, line 1896: Is there a way to restrict this to only be the type of the root actor?

packages/core/src/interpreter.ts
```diff
@@ -123,6 +123,8 @@ export class Interpreter<
 
   public src?: string;
 
+  public root: any;
```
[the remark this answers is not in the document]
with-heart: I don't think we'd want `root` here, right?:

- we're only assigning to `this.system.root`
- since `this` _is_ the root, referencing `interpreter.root` would be the same thing as just referencing `interpreter`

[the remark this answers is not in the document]
davidkpiano: `AnyActorRef` for now

packages/core/src/types.ts
davidkpiano: Not yet (to my knowledge); best to have this as required and `AnyActorRef` for now

with-heart: Looks like there are a couple options here:

1. add a `root` property + generic param for it to `ActorSystemInfo` so that we can reference it directly on `T`
2. add a new generic param to `ActorSystem` that `extends AnyActorRef` so we can specify the root type without adding a generic param to `ActorSystem`. would also need to add a generic param to `Interpreter` so it can be passed in to `ActorSystem`

kevinmaes: @with-heart Thank you for these other suggestions! For simplicity, I made the type required and `AnyActorRef` but I'd be interested in taking more time to explore some alternatives as I dig deeper into the code and types.

packages/core/src/interpreter.ts
```diff
@@ -123,7 +124,7 @@ export class Interpreter<
 
   public src?: string;
 
-  public root: any;
+  public root?: AnyActorRef;
```
davidkpiano:
```suggestion
  public root: AnyActorRef;
```

An actor system will always have a root actor ref

kevinmaes: @davidkpiano I'm not getting any errors if I omit this entirely.

.changeset/fresh-terms-jam.md
```diff
@@ -0,0 +1,9 @@
+---
+'xstate': minor
+'@xstate/fsm': minor
+'@xstate/react': minor
+'@xstate/vue': minor
+'@xstate-repo/jest-utils': minor
```
Andarist, line 6:
```suggestion
```

packages/core/src/types.ts
Andarist: @with-heart has good ideas about what it would take to support this at type-level here. And with those changes in mind - I'm not quite convinced that this is a feature that we'd like to land.

To fully support this we **need** type-level support. Types nowadays are not a nice-to-have, they are the baseline. 

So with that in mind, I'm not sure if this will make the overall usage that much simpler once we demand from people to type this. At the definition site you will have to remember how to type this - is this a second param? first? something else? Treating this just like every other registerable actor is pretty neat because the user doesn't have to learn any new concept, nor API.

I see one advantage of this though: this particular actor wouldn't have to be nullable because it's guaranteed to exist. 

It would be nice if the PR's description would motivate this addition beyond not having to use the explicit ID at the call site. I feel like this alone is really a tiny improvement. 

Either way, I would like to get this on hold. I think that we should first land improved types for the system in its current shape. Only with that base in place, we'll actually be able to tell what's the cost of this feature and if it's worth it. The change itself is purely additive so it can land at just any time in the future.

kevinmaes: @Andarist Thanks for thinking this one through a bit more and thanks again @with-heart again for your type suggestions. I'll hold off on working on this until we find a way to type the root actor. In the meantime, I'll add mention in the Docs about how to set and obtain the root actor's `systemId` when sending events.

davidkpiano:
Doing some cleanup 🧹 

In v6, a system can have many "root" actors

PR #4081, opened 2023-06-18
[v5] Fix higher-level guard types

davidkpiano:


packages/core/test/guards.test.ts
```diff
@@ -808,6 +808,29 @@ describe('not() guard', () => {
...
+      {
+        id: 'b',
+        types: {} as {
+          events: { type: 'EVENT' };
+        },
+        on: {
+          EVENT: {
+            target: '#b',
+            // would have been a type error
+            guard: not('falsy')
```
davidkpiano, line 823: @Andarist I just need help with one type here. It seems like TS can't properly "generalize" `not('falsy')` to be a `BooleanGuardDefinition` that can be used anywhere.

Andarist: It's quite related to the "actions as functions" work since we suffer from the same problem here. It starts to work with this big ugly lie:
```diff
diff --git a/packages/core/src/types.ts b/packages/core/src/types.ts
index 73836098c..9765b0f68 100644
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -283,6 +283,7 @@ export interface GuardDefinition<
   children?: Array<GuardDefinition<TContext, TEvent>>;
   predicate?: GuardPredicate<TContext, TEvent>;
   params: { [key: string]: any };
+  (): never;
 }
 
 export interface BooleanGuardObject<

```

>  It seems like TS can't properly "generalize" not('falsy') to be a BooleanGuardDefinition that can be used anywhere.

In a general sense, this can't/shouldn't be done. We don't know what `not('falsy')` resolves to and what events it is compatible with - so we can't make sure that this alone is compatible with just anything. As showcased above we can kinda try to make sure that it can be used anywhere in the context of any given machine - but then we only can type the actual implementation of the `falsy` guard with a typegen.

davidkpiano: Is there any good workaround for now? I think it's acceptable that any boolean guard with serialized children like `not('whatever')` should be allowed if no typegen data is present.

Andarist: I came to the conclusion that the responsibility of keeping types aligned here is on the machine's side of things and its implementations. `not` that references a string doesn't specify any requirements for the events that this guard will be called with. It simply declares/uses a guard's type/name - the even requirements should be derived from the machine's config (its graph/transitions) and enforced within the `implementations` argument.

PR #4082, opened 2023-06-18
[v5] Inspect API

davidkpiano:


packages/core/test/inspect.test.ts
```diff
@@ -0,0 +1,281 @@
...
+          type: event.type,
+          sessionId: event.sessionId,
+          snapshot:
+            typeof event.snapshot === 'object' && 'value' in event.snapshot
+              ? { value: event.snapshot.value }
+              : event.snapshot,
+          event: event.event.type
+        };
+      })
+    ).toMatchInlineSnapshot(`
```
davidkpiano, line 160: cc. @gavination Here are what the events look like, roughly

packages/core/src/stateUtils.ts
```diff
@@ -1434,26 +1438,36 @@ export function resolveActionsAndContext<
       action: typeof action === 'string' ? { type: action } : (action as any)
     };
 
-    if (!('resolve' in resolved)) {
+    if (!('resolve' in resolvedAction)) {
       if (actorCtx?.self.status === ActorStatus.Running) {
-        resolved(args);
+        resolvedAction(args);
+        actorCtx?.system?._sendInspectionEvent({
```
rlaffers, line 1444: The ActorActionEvent is very useful for observing side effects performed by the actor. However, there is not enough information here. It is impossible to associate this action with the transition on which it was executed. Now is it possible to glean which actor performed this event. We might need the following:

```
{
  type: '@xstate.action`,
  sessionId: string, // <-- which actor executed this?
  event: AnyEventObject, // <-- as in ActorTransitionEvent.event
}
```

packages/core/src/system.ts
```diff
@@ -34,6 +46,36 @@ export function createSystem<T extends ActorSystemInfo>(): ActorSystem<T> {
...
+    },
+    inspect: (observer) => {
+      observers.add(observer);
+    },
+    _sendInspectionEvent: (event) => {
+      const resolvedInspectionEvent: ResolvedInspectionEvent = {
+        id: Math.random().toString(),
+        createdAt: new Date().toString(),
+        ...event,
+        systemId: system.root.sessionId
```
rlaffers, line 58: I find this a little confusing. It is called systemId, but is populated with the root sessionId. 

I understand that this is meant to identify the system within which this is happening, but the `systemId` prop is already used in the spawn and invoke options. A child machine can have both sessionId and systemId, which may have been assigned to it when it was spawned/invoked.
I guess, I'm trying to say that Actor.systemId has a different meaning than InspectionEvent.systemId. 

- Actor.systemId = a key in the system Map,
- InspectionEvent.systemId = sessionId of the system itself.

rlaffers (commented):
Great work, this will make implementing devtools much easier than in v4. :heart: 

Also, I have tested how it works - mostly very well.

One hitch is delayed actions - they are logged at the time their timer runs out, which makes sense. The inspector however still may wish to map them onto the transition which originally scheduled them. This is currently impossible, but adding an `event` prop onto the @xstate.action + a timestamp at which they were scheduled would allow for this mapping.

packages/core/src/system.ts
davidkpiano: Yeah I struggled with this too, maybe we should rename this to `rootSessionId`

rlaffers: Yes, I think rootSessionId is a good name.

packages/core/src/stateUtils.ts
rlaffers: Additionally, an indication whether this was an entry, exit, or a transition action would be nice. It would allow for a better visualization in the Inspector.

packages/core/src/actions/send.ts
```diff
@@ -123,10 +123,12 @@ function execute(
   const { to, event } = params;
 
   actorContext.defer(() => {
-    to.send(
+    actorContext?.system.sendTo(
```
davidkpiano, line 126: By using `system.sendTo`, we make it possible to track the source/target of events.

packages/core/src/interpreter.ts
```diff
@@ -439,6 +478,14 @@ export class Actor<
         `Only event objects may be sent to actors; use .send({ type: "${event}" }) instead`
       );
     }
+    if (!('__id' in event)) {
```
davidkpiano, line 481: The `__id` is a hacky workaround to ensure that the same event isn't inspected twice

packages/core/test/inspect.test.ts
```diff
@@ -0,0 +1,337 @@
...
+    const events: InspectionEvent[] = [];
+
+    const actor = interpret(machine, {
+      inspect: {
+        next(event) {
+          events.push(event);
+          // push events to websocket server
+          // server.clients.forEach((client) => {
+          //   client.send(JSON.stringify(event));
+          // });
```
davidkpiano, line 70:
```suggestion
```

packages/xstate-inspect/src/index.ts
```diff
@@ -1,7 +1,114 @@
...
+
 export {
   inspect,
   createWindowReceiver,
   createWebSocketReceiver,
   createDevTools
 } from './browser.ts';
 export * from './types.ts';
+
+const WebSocket = require('ws');
```
davidkpiano, line 18: Ignore this file, will probably revert these changes

packages/core/src/actors/observable.ts
```diff
@@ -45,9 +45,10 @@ export function fromObservable<T, TInput>(
   const errorEventType = '$$xstate.error';
   const completeEventType = '$$xstate.complete';
 
-  return {
+  // TODO: add event types
+  const logic: ObservableActorLogic<T, TInput> = {
```
Andarist, line 49: is there any need to move it to a variable? it's quite nice that we don't have to repeat the declared type here if we just rely on the contextual return type applied to the return statement

packages/core/src/actors/observable.ts
```diff
@@ -45,9 +45,10 @@ export function fromObservable<T, TInput>(
   const errorEventType = '$$xstate.error';
   const completeEventType = '$$xstate.complete';
 
-  return {
+  // TODO: add event types
```
Andarist, line 48: which event types do u think about here?

packages/core/src/types.ts
```diff
@@ -2154,11 +2156,31 @@ export interface ActorSystemInfo {
...
+   */
   _unregister: (actorRef: AnyActorRef) => void;
+  /**
+   * @private
+   */
   _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
   get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+  inspect: (observer: Observer<InspectionEvent>) => void;
+  _sendInspectionEvent: (event: InspectionEvent) => void;
+  sendTo: (
```
Andarist, line 2178: I like this method but I wonder about:
- could we use a different order of arguments, having `event` at the end would look visually better when we call this function
- could we make `target` and `source` required?

packages/core/src/types.ts
```diff
@@ -2154,11 +2156,31 @@ export interface ActorSystemInfo {
 }
 
 export interface ActorSystem<T extends ActorSystemInfo> {
+  /**
+   * @private
+   */
```
Andarist, line 2161: A neat thing that we could maybe utilize is `@internal` and TS `stripInternal: true` option

packages/core/src/actions/send.ts
```diff
@@ -123,10 +123,12 @@ function execute(
   const { to, event } = params;
 
   actorContext.defer(() => {
```
Andarist, line 125: I wonder if we could move this concern to the `system.sendTo`. It would be nice if we didn't have to worry about this here at all and if all sends would be processed in the very same way. That's outside of the scope of this PR though

packages/core/src/interpreter.ts
```diff
@@ -83,6 +87,8 @@ export class Actor<
   TEvent extends EventObject = EventFromLogic<TLogic>
 > implements ActorRef<TEvent, SnapshotFrom<TLogic>>
 {
+  public static defaults = defaultOptions as Partial<ActorOptions<any>> &
```
Andarist, line 90: Why do we expose this as a static property here? It feels like somebody could mutate this to change the default but I would consider that to be a bad thing overall. I wouldn't like a random~ call site to meaningfully change the behavior of all actors in the application. While it might make sense as something convenient in a self-contained system - it quickly breaks apart situations in which actors from different subsystems are put in a single application. A change "intended" for one of them should really not affect **all of them**.

I understand that this might be used to enable devtools globally but I would much prefer a dedicated way of doing this. Inspection is an extra thing that doesn't affect runtime behaviors of all contained subsystems so I think that setting this one thing globally should be fine.

davidkpiano: What would be the dedicated way of enabling devtools globally that you're thinking of?

packages/core/src/interpreter.ts
```diff
@@ -135,14 +141,18 @@ export class Actor<
...
+    } as ActorOptions<TLogic> & typeof Actor.defaults;
+
+    const { clock, logger, parent, id, systemId, inspect } = resolvedOptions;
 
-    const { clock, logger, parent, id, systemId } = resolvedOptions;
-    const self = this;
+    this.system = parent?.system ?? createSystem(this);
 
-    this.system = parent?.system ?? createSystem();
+    if (inspect) {
```
Andarist, line 152: I think there is no way atm to pass `inspect` when invoking or spawning children but perhaps better be safe than sorry here?
```suggestion
    if (inspect && !parent) {
```

packages/core/src/interpreter.ts
Andarist:
```ts
import { enableInspection } from 'xstate/devtools'
enableInspection(myInspector)
```

In the future we could even allow this:
```ts
import 'xstate/devtools/enable-browser-devtools'
```

Internally, we would just set a module-level variable and reach for it when instantiating a system.

packages/core/src/types.ts
davidkpiano: Sure, we can reorder it. It should be an internal-only method too, not meant for public usage.

packages/core/src/actors/observable.ts
davidkpiano: No need, can revert

packages/core/src/types.ts
```diff
@@ -2219,3 +2241,55 @@ export type PersistedMachineState<TState extends AnyState> = Pick<
     };
   };
 };
+
+export interface InspectedActorObject {
```
Andarist, line 2245: it would be nice if all of those inspector-related types would be moved closer to where they are actually used - which currently is the `system.ts`. I find it much easier to navigate the codebase when things stay colocated. Our `./types.ts` here is pretty huge already and I often get lost in it ;p

packages/core/src/types.ts
Andarist: > It should be an internal-only method too, not meant for public usage.

This has crossed my mind but it actually should be available for all custom actors so we can't just `_` it

packages/core/src/types.ts
```diff
@@ -2219,3 +2241,55 @@ export type PersistedMachineState<TState extends AnyState> = Pick<
...
+  sourceId: string | undefined; // Session ID
+  targetId: string; // Session ID, required
+}
+
+export interface ActorRegistrationEvent {
+  type: '@xstate.actor';
+  actorRef: AnyActorRef;
+  sessionId: string;
+  parentId?: string;
+  definition?: string; // JSON-stringified definition or URL
```
Andarist, line 2285:
```suggestion
```

packages/core/src/interpreter.ts
```diff
@@ -179,6 +189,13 @@ export class Actor<
     // Ensure that the send method is bound to this Actor instance
     // if destructured
     this.send = this.send.bind(this);
+    this.system._sendInspectionEvent({
+      type: '@xstate.actor',
+      actorRef: this,
+      sessionId: this.sessionId,
+      // definition: JSON.stringify(this.logic.config),
```
Andarist, line 196:
```suggestion
```

packages/core/src/interpreter.ts
```diff
@@ -193,7 +210,7 @@ export class Actor<
   // array of functions to defer
   private _deferred: Array<() => void> = [];
 
-  private update(state: InternalStateFrom<TLogic>): void {
+  private update(state: InternalStateFrom<TLogic>, event: TEvent): void {
```
Andarist, line 213: this would likely save you some casts:
```suggestion
  private update(state: InternalStateFrom<TLogic>, event: EventObject): void {
```

`TEvent` annotation isn't true here anyway since internal events don't belong to `TEvent`

packages/core/src/types.ts
davidkpiano: Will move 🚚

packages/core/src/interpreter.ts
```diff
@@ -282,13 +311,22 @@ export class Actor<
     }
     this.status = ActorStatus.Running;
 
+    const initEvent = createInitEvent(this.options.input);
+
+    this.system._sendInspectionEvent({
+      type: '@xstate.event',
+      event: initEvent,
+      sourceId: this._parent?.sessionId,
```
Andarist, line 319: I wonder if this is truly OK, is the init event received from the parent or is it raised internally? I can see arguments going both ways :p

packages/core/src/system.ts
```diff
@@ -1,13 +1,25 @@
...
+): ActorSystem<T> {
   const children = new Map<string, AnyActorRef>();
   const keyedActors = new Map<keyof T['actors'], AnyActorRef | undefined>();
   const reverseKeyedActors = new WeakMap<AnyActorRef, keyof T['actors']>();
+  const observers = new Set<Observer<ResolvedInspectionEvent>>();
 
   const system: ActorSystem<T> = {
-    _bookId: () => `x:${sessionIdCounter++}`,
+    root: rootActor,
+    _bookId: (name: string = 'x') => `${name}:${systemCounter++}`,
```
Andarist, line 22: This argument is never used in the codebase
```suggestion
    _bookId: () => `x:${systemCounter++}`,
```

packages/core/src/types.ts
```diff
@@ -2219,3 +2241,53 @@ export type PersistedMachineState<TState extends AnyState> = Pick<
...
+}
+
+export type InspectionEvent =
+  | ActorTransitionEvent
+  | ActorCommunicationEvent
+  | ActorRegistrationEvent;
+
+export type ResolvedInspectionEvent = InspectionEvent & BaseInspectionEvent;
+
+export type InspectorActorRef = ActorRef<InspectionEvent>;
```
Andarist, line 2293:
```suggestion
export type InspectorActorRef = ActorRef<InspectionEvent>;
```

Andarist, line 2293: such a thing doesn't exist anywhere so it's redundant, at least right now
```suggestion
```

packages/core/src/types.ts
```diff
@@ -2219,3 +2241,53 @@ export type PersistedMachineState<TState extends AnyState> = Pick<
     };
   };
 };
+
+export interface InspectedActorObject {
```
Andarist, line 2245: this whole type is not used anywhere, can it be removed?

packages/core/src/types.ts
```diff
@@ -2219,3 +2241,53 @@ export type PersistedMachineState<TState extends AnyState> = Pick<
...
+  systemId?: string; // Session ID
+  events: ActorCommunicationEvent[]; // events where targetId === sessionId
+  definition?: string; // JSON-stringified machine definition or URL
+  createdAt: number; // Timestamp
+  updatedAt: number; // Timestamp, derived from latest update createdAt
+  status: 0 | 1 | 2; // 0 = not started, 1 = started, 2 = stopped, derived from latest update status
+}
+
+export interface BaseInspectionEvent {
+  actorSystemId: string; // the session ID of the root
```
Andarist, line 2258: This one would sound a little bit more accurate. `actorSystemId` sounds a little bit like if we'd have special system ids but we don't - the system is attached to the root actor but it has no id of its own.
```suggestion
  rootId: string; // the session ID of the root
```

packages/core/src/types.ts
```diff
@@ -2219,3 +2241,53 @@ export type PersistedMachineState<TState extends AnyState> = Pick<
...
+  type: '@xstate.actor';
+  actorRef: AnyActorRef;
+  sessionId: string;
+  parentId?: string;
+}
+
+export type InspectionEvent =
+  | ActorTransitionEvent
+  | ActorCommunicationEvent
+  | ActorRegistrationEvent;
```
Andarist, line 2289: All of those have somewhat confusing names. I think it would be better if the name of the type would match the `.type` closer - but currently, we have `ActorTransitionEvent` with `.type: '@xstate.snapshot'` etc.

packages/core/src/interpreter.ts
```diff
@@ -282,13 +311,22 @@ export class Actor<
     }
     this.status = ActorStatus.Running;
 
+    const initEvent = createInitEvent(this.options.input);
+
+    this.system._sendInspectionEvent({
```
Andarist, line 316: I wish that this one would go through `system.sendTo` if that's treated as an explicit event here. It's not a blocker though since ultimately it doesn't matter that much and we can try to juggle things more in the future.

packages/core/src/interpreter.ts
davidkpiano: Conceptually, parent creates a child with input that is inherited from the parent

packages/core/src/types.ts
davidkpiano: Yes

packages/core/src/interpreter.ts
davidkpiano: Un-exposed

packages/core/src/types.ts
davidkpiano: Renamed to `relay(event, source, target)` as it is _relaying_ an event

packages/core/src/actions/send.ts
davidkpiano: Probably not easily at the moment; something we can revisit

packages/core/src/types.ts
```diff
@@ -2176,11 +2181,30 @@ export interface ActorSystemInfo {
...
+   * @internal
+   */
   _unregister: (actorRef: AnyActorRef) => void;
+  /**
+   * @internal
+   */
   _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
   get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+  inspect: (observer: Observer<InspectionEvent>) => void;
+  _sendInspectionEvent: (event: InspectionEvent) => void;
```
Andarist, line 2202:
```suggestion
  /**
   * @internal
   */
  _sendInspectionEvent: (event: InspectionEvent) => void;
```

Andarist (approved):

PR #4086, opened 2023-06-20
Minor tweaks to the root `tsconfig.json`

Andarist:


packages/core/package.json
```diff
@@ -120,6 +120,7 @@
   "devDependencies": {
     "@scion-scxml/test-framework": "^2.0.15",
     "@xstate-repo/jest-utils": "*",
+    "ajv": "^8.12.0",
```
Andarist, line 123: We were implicitly relying on `ajv@6.x`, we never specified what version of `ajv` we want to actually use.

tsconfig.json
```diff
@@ -1,16 +1,11 @@
 {
   "compilerOptions": {
-    "allowSyntheticDefaultImports": true,
     "target": "esnext",
     "isolatedModules": true,
     "moduleResolution": "node16",
     "allowImportingTsExtensions": true,
     "noEmit": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
```
Andarist, line 10: those are linting concerns and we should use a linter for those kind of things, I think it's best to remove them for now from here and consider adding ESlint to the project at some point

davidkpiano (approved):

PR #4088, opened 2023-06-21
Enable TS strict rules that don't affect many locations

Andarist:


packages/core/src/StateNode.ts
```diff
@@ -350,7 +350,7 @@ export class StateNode<
         guardPassed =
           !guard ||
           evaluateGuard<TContext, TEvent>(guard, resolvedContext, event, state);
-      } catch (err) {
+      } catch (err: any) {
```
Andarist, line 353: it's an easy fix to actually fix those error-related locations but I'd like to take a look at how other libraries handle those cases before committing to any particular solution

davidkpiano (approved):

PR #4092, opened 2023-06-21
Enable `noImplicitAny`

Andarist:


packages/core/test/input.test.ts
```diff
@@ -198,7 +198,7 @@ describe('input', () => {
       {
         invoke: {
           src: 'child',
-          input: ({ event }) => {
+          input: ({ event }: any) => {
```
davidkpiano, line 201: See #4054 - that would remove the need for this

Andarist: I know. That's why I didn't even bother typing this here. It will be an easy conflict resolution either way.

davidkpiano (approved):

PR #4098, opened 2023-06-25
Add missing actions to main export

davidkpiano:
The `log`, `pure`, `choose`, and `stop` actions were added to the main export:

```ts
import { log, pure, choose, stop } from 'xstate';
```


mellson (approved):
It looks fine to me. And since it's not breaking any old APIs I guess the change should be harmless.

Andarist (approved):

PR #4102, opened 2023-06-27
add deep persistence example

RainerAtSpirit:


davidkpiano:
Doing some cleanup 🧹 

Taking a different approach for persistence in v6

PR #4103, opened 2023-06-27
Remove `delimiter` option

Andarist:


packages/core/src/scxml.ts
```diff
@@ -71,7 +75,9 @@ function executableContent(elements: XMLElement[]) {
 function getTargets(targetAttr?: string | number): string[] | undefined {
   // return targetAttr ? [`#${targetAttr}`] : undefined;
   return targetAttr
-    ? `${targetAttr}`.split(/\s+/).map((target) => `#${target}`)
+    ? `${targetAttr}`
+        .split(/\s+/)
+        .map((target) => `#${sanitizeStateId(target)}`)
```
davidkpiano, line 80: Hah this is a much smarter solution than what I did before

davidkpiano (approved):

PR #4110, opened 2023-06-30
Remove `getContext`

Andarist:


packages/core/src/StateMachine.ts
```diff
@@ -362,10 +359,7 @@ export class StateMachine<
   }
 
   public get definition(): StateMachineDefinition<TContext, TEvent> {
-    return {
-      context: this.getContext(),
```
Andarist, line 366: This was *not* safe. Since `.definition` is a getter we can't pass any arguments to it and `input` might be required

davidkpiano (approved):

PR #4114, opened 2023-07-03
Make options params optional in useMachine hook type

ilkaydnc:
Currently, the useMachine hook's second parameter is optional and has a default fallback value of `{}`. However, this fallback value is not declared in the type definition file, causing a type error when it is used.

This pull request addresses the issue by updating the type definition file to include the fallback value {} for the second parameter of the useMachine hook. This change ensures that the type declarations align with the actual behavior of the hook, preventing type errors in usage.

To provide more context, the screenshot below illustrates the current type error:

image

This fix will enhance the developer experience and prevent potential issues related to incorrect typings when using the useMachine hook.

### DIFF
--- a/packages/xstate-react/index.v3.d.ts
+++ b/packages/xstate-react/index.v3.d.ts
@@ -16,62 +16,70 @@ type UseMachineReturn<
   TInterpreter = InterpreterFrom<TMachine>
 > = [StateFrom<TMachine>, Prop<TInterpreter, 'send'>, TInterpreter];
 
-type UseMachineRestParams<
-  TMachine extends AnyStateMachine
-> = AreAllImplementationsAssumedToBeProvided<
-  TMachine['__TResolvedTypesMeta']
-> extends false
-  ? [
-      InterpreterOptions &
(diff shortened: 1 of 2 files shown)


packages/xstate-react/index.v3.d.ts
```diff
@@ -16,62 +16,70 @@ type UseMachineReturn<
...
-          TMachine['__TContext'],
-          TMachine['__TEvent'],
-          TMachine['__TResolvedTypesMeta']
-        >)?,
-      (Observer<StateFrom<TMachine>> | ((value: StateFrom<TMachine>) => void))?
-    ];
+type UseInterpretRestParams<TMachine extends AnyStateMachine> =
+  AreAllImplementationsAssumedToBeProvided<
+    TMachine['__TResolvedTypesMeta']
+  > extends false
```
Andarist, line 55: This check is intentional and it's responsible for making the options required (conditionally). Please provide a runnable repro case of your problem so we could assess the whole situation.

Andarist:
I'm closing this PR for housekeeping purposes as I think that it's not something that we want to pull. If you feel that we should reconsider this decision - feel free to comment back so we could continue the discussion.

PR #4117, opened 2023-07-04
[v5] `self` for all actor logic

davidkpiano:
This PR allows access to `self` from all actor logic creators:

```ts
const promiseLogic = fromPromise(({ self }) => { ... });

const observableLogic = fromObservable(({ self }) => { ... });

const callbackLogic = fromCallback((sendBack, receive, { self }) => { ... });

const transitionLogic = fromTransition((state, event, { self }) => { ... }, ...);
```

## Motivating example

```ts
const callbackLogic = fromCallback((sendBack, receive, { system, self }) => {
  const otherActor = system.get('other');

  if (otherActor) {
    otherActor.send({
      type: 'hello',
      respondTo: self
    });
  }

  receive(event => {
    // hopefully get a response from `otherActor`
  });

  // ...
});
```


Andarist:
I think that we should always back up new features with some motivating examples. TypeScript has this in their feature request template and I dig it:
```ts
## 📃 Motivating Example

<!--
  If you were announcing this feature in a blog post, what's a short explanation that shows
  a developer why this feature improves the language?
-->
```

All I can get from this PR is **what** but there is no **why** here.

.changeset/orange-pillows-deny.md
```diff
@@ -0,0 +1,15 @@
+---
+'xstate': minor
+---
+
+Actor logic cerators now have access to `self`:
```
Andarist, line 5:
```suggestion
Actor logic creators now have access to `self`:
```

Andarist:
> but a likely bigger use-case is enabling the request-response pattern for any actor, which I demonstrated in fromCallback(...).

I can't find an example like that in the tests, nor in the PR description. If I understand correctly, to achieve this pattern we need smth like this:
```ts
createMachine({
  invoke: {
    src: fromCallback((sendBack, receive, { self }) => {
      receive((event) => {
        switch (event.type) {
          case "PONG": {
            // ...
          }
        }
      });

      sendBack({
        type: "PING",
        ref: self,
      });
    }),
  },
  on: {
    PING: {
      actions: sendTo(
        ({ event }) => event.ref,
        ({ event }) => ({ type: "PONG" })
      ),
    },
  },
});

```

davidkpiano:
@Andarist Okay to merge?

PR #4123, opened 2023-07-06
Drop support for `TransitionsConfigArray`

Andarist:
I think that:
1. `TransitionsConfigArray` complicated our types
2. it also complicated our implementation and slowed down the most basic scenario (we were searching for candidates with explicit descriptors in the array instead of just looking it up in a map)
3. it's **rarely** used feature (if used at all)
4. its existence it makes vizualization harder/less obvious. We don't visually order (nor denote order) anyhow. By removing the ability to order transitions arbitrarily in an array we make it more intuitive to interpret the visualization: explicit descriptions > longest-to-shortest partial descriptors > wildcard

This PR is not meant to refactor all inner bits. The main goal here was to remove this from the types and the public API, the internals could still use some further cleanups but that can be done at any time.

### DIFF
--- a/packages/core/test/scxml.test.ts
+++ b/packages/core/test/scxml.test.ts
@@ -120,7 +120,7 @@ const testGroups: Record<string, string[]> = {
   // script: ['test0', 'test1', 'test2'], // <script/> conversion not implemented
   // 'script-src': ['test0', 'test1', 'test2', 'test3'], // <script/> conversion not implemented
   'scxml-prefix-event-name-matching': [
-    'star0'
+    // 'star0' // this relies on the source order of transitions where * is first and it's supposed to get macthed over an explicit descriptor
     // prefix event matching not implemented yet
     // 'test0',
     // 'test1'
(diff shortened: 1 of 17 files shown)


packages/core/test/scxml.test.ts
```diff
@@ -120,7 +120,7 @@ const testGroups: Record<string, string[]> = {
   // script: ['test0', 'test1', 'test2'], // <script/> conversion not implemented
   // 'script-src': ['test0', 'test1', 'test2', 'test3'], // <script/> conversion not implemented
   'scxml-prefix-event-name-matching': [
-    'star0'
+    // 'star0' // this relies on the source order of transitions where * is first and it's supposed to get macthed over an explicit descriptor
```
davidkpiano, line 123: Do you think there is value in supporting this use-case?

davidkpiano (approved):

packages/core/test/scxml.test.ts
Andarist: I don't think so - it makes visualization less obvious. If somebody really needs smth like this they can always divide their state into 2 nested ones and put "higher priority" transitions in the inner one.

PR #4127, opened 2023-07-07
Refactor how builtin actions are structured and how all actions are stored and resolved

Andarist:


davidkpiano:
Looked mainly at the tests; LGTM but CI needs to pass

packages/core/src/stateUtils.ts
```diff
@@ -365,7 +354,7 @@ export function formatTransition<
   }
   const transition = {
     ...transitionConfig,
-    actions: toActionObjects(toArray(transitionConfig.actions)),
```
davidkpiano, line 368: Why is `toActionObjects` no longer called? A string `"action"` should be converted to `{ type: 'action' }`

davidkpiano: This is also what is causing the test failures, I believe

Andarist: > Why is toActionObjects no longer called? A string "action" should be converted to { type: 'action' }

Why? :p I mean, this isn't really strictly needed for quite anything (as showcased by this PR). I could add back this particular conversion but I wonder if it's worth it. From the types PoV, this makes the publicly accepted action different from the one that we operate on internally and right now both can just use `Action` and it's pretty easy to work with.

> This is also what is causing the test failures, I believe

One test was related to the lack of `.type` on builtin actions - just fixed it. There wasn't any test that would test that actions are called with an object form of the action even if it was configured as string though - just added one and pushed out a fix for it.

The second one is related to JSON schema serialization. I didn't exactly dig into what has changed and what it would take to fix it right now. I think it's worth considering just removing this test for now (and perhaps the JSON schema as well). We can always bring this back later but right now this relies on the fact that actions are always objects but with this PR they are not. I think it would be better to remove `toJSON` methods and implement converting logic in something akin to `machineToJson`. It would also be good to specify what's the goal of this JSON schema and what one could actually do with it today.

Andarist: For the time being I fixed the JSON serialization - changed around this can be moved to other discussions/PRs.

packages/core/src/StateNode.ts
```diff
@@ -208,6 +207,20 @@ export class StateNode<
    * The well-structured state node definition.
    */
   public get definition(): StateNodeDefinition<TContext, TEvent> {
+    const toSerializableActon = (action: Action<any, any, any>) => {
```
davidkpiano, line 210: Nit: this can be pulled out to module scope

packages/core/src/actions/assign.ts
```diff
@@ -28,71 +85,17 @@ export function assign<
...
-): DynamicAssignAction<TContext, TExpressionEvent, TEvent> {
-  return createDynamicAction<
-    TContext,
-    TExpressionEvent,
-    TEvent,
-    AssignActionObject<TContext>,
-    {
-      assignment: typeof assignment;
+) {
+  function assign(_: ActionArgs<TContext, TExpressionEvent>) {
```
davidkpiano, line 89: Nit: was a little confused, can we call this something different than `assign` to disambiguate?

davidkpiano (approved):
Some nits, otherwise LGTM

packages/core/src/actions/assign.ts
Andarist: this was somewhat deliberate here - a) naming is hard b) it will show be printed in the console as `ƒ assign() {}` which is kinda nice. I agree that this is a little bit confusing internally but I just don't know what is a better name for this ;p

davidkpiano: All good 👍

packages/xstate-scxml/src/index.ts
```diff
@@ -26,21 +25,20 @@ export function functionToExpr(fn: Function): string {
...
+      event: raiseAction.event.type
     }
   };
 }
 
-function actionToSCXML(action: BaseActionObject): XMLElement {
-  if (action.type === 'xstate.raise') {
-    return raiseActionToSCXML(action as RaiseActionObject<any, any>);
+function actionToSCXML(action: Action<any, any>): XMLElement {
+  // TODO: this won't work rn since we don't keep `builtinAction.type` around
```
Andarist, line 39:
```suggestion
```

Issue #4105, opened 2023-06-28
Bug: Support Svelte 4 in @xstate/svelte


danawoodman:
### Description

Installing @xstate/svelte with an app using Svelte v4 fails with 

![image](https://github.com/statelyai/xstate/assets/157695/26783f83-ee14-4cd0-9d63-67872df271c6)

### Expected result

It installs on Svelte 4

### Actual result

It doesn't 😅 

### Reproduction

tried to with codesandbox but it kept crashing

Andarist:
It might be worth rechecking if we are affected anyhow by a breaking change from this PR: https://github.com/sveltejs/svelte/pull/6750

davidkpiano:
> xstate is working fine for me in svelte 4. is there are work around for this? I can't deploy my app.

Is the error you're getting the same as the original issue?

jer-0:
> > xstate is working fine for me in svelte 4. is there are work around for this? I can't deploy my app.
> 
> Is the error you're getting the same as the original issue?

yes sir:
```
#0 10.77 npm ERR! While resolving: web-app@0.0.1
#0 10.77 npm ERR! Found: svelte@4.0.5
#0 10.77 npm ERR! node_modules/svelte
#0 10.77 npm ERR!   dev svelte@"^4.0.0" from the root project
#0 10.77 npm ERR!
#0 10.77 npm ERR! Could not resolve dependency:
#0 10.77 npm ERR! peer svelte@"^3.24.1" from @xstate/svelte@2.0.1
#0 10.77 npm ERR! node_modules/@xstate/svelte
#0 10.77 npm ERR!   @xstate/svelte@"^2.0.1" from the root project
```

danawoodman:
I believe [this line](https://github.com/statelyai/xstate/blob/main/packages/xstate-svelte/package.json#L49) should change to something like:

```
"svelte": "^3.x || ^4.x",
```

jer-0:
> ^3.24.1
I think this is better: 
```js
"svelte": "3.24.1 - 4"
```

ref: https://stackoverflow.com/a/60828558

danawoodman:
If the intent is to support 3 and 4, then doing `3 - 4` or ` `3 || 4` should be equivalent:

![image](https://github.com/statelyai/xstate/assets/157695/859c3652-1c90-4873-84d1-3f36e8e35e23)

![image](https://github.com/statelyai/xstate/assets/157695/cf49a9cf-5735-41df-a637-607b250da14c)

[...]

PR #4128, opened 2023-07-08
Allow Svelte 4 in the specified peer dependency range

jer-0:
closes: https://github.com/statelyai/xstate/issues/4105, make xstate-svelte compatible with svelte 4.

### DIFF
--- a/packages/xstate-svelte/package.json
+++ b/packages/xstate-svelte/package.json
@@ -46,7 +46,7 @@
   },
   "peerDependencies": {
     "@xstate/fsm": "^2.1.0",
-    "svelte": "^3.24.1",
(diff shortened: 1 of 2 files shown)


packages/xstate-svelte/package.json
```diff
@@ -46,7 +46,7 @@
   },
   "peerDependencies": {
     "@xstate/fsm": "^2.1.0",
-    "svelte": "^3.24.1",
+    "svelte": "3.24.1 - 4",
```
Andarist, line 49:
```suggestion
    "svelte": "^3.24.1 || ^4",
```

Andarist (approved):

PR #4145, opened 2023-07-13
[v5] Do not swallow errors

davidkpiano:


davidkpiano:
@Andarist Will need help with the couple of tests that are failing. They should expect an unhandled rejection, but I'm unsure of how to best test that in Jest.

packages/core/src/interpreter.ts
```diff
@@ -302,15 +310,10 @@ export class Interpreter<
         this._complete();
       }
     } catch (err) {
-      // TODO: properly handle errors
-      if (this.observers.size > 0) {
-        this.observers.forEach((observer) => {
-          observer.error?.(err);
-        });
-        this.stop();
```
Andarist, line 310: `.stop()` was removed from here. TODO: reevaluate its importance and role here. If anything this should likely call `._stop` to avoid the "A non-root actor cannot be stopped directly." error - unless this throw can only happen in a root actor (it's unlikely that this is actually the case here).

packages/core/src/reportUnhandledError.ts
```diff
@@ -0,0 +1,5 @@
+export function reportUnhandledError(err: unknown) {
+  setTimeout(() => {
+    throw err;
+  });
```
davidkpiano, line 4: Can we add a quick comment explaining why this needs to be in `setTimeout`?

packages/core/src/interpreter.ts
```diff
@@ -310,9 +325,13 @@ export class Interpreter<
...
     } catch (err) {
-      this.observers.forEach((observer) => {
-        observer.error?.(err);
-      });
+      for (const observer of this.observers) {
+        try {
+          observer.error?.(err);
+        } catch (err2) {
+          reportUnhandledError(err2);
+        }
```
davidkpiano: So the logic is that if _any_ observer is missing an error handler (even if other observers have error handlers), rethrow anyway?

Andarist: Yeah, smth like that. In RxJS would get an error for each observer that is missing an error handler but it would still be the same error, it feels somewhat redundant. In RxJS each observer is something that is notified about the error and calling the error listener is delegated to it, so there is no~ loop over existing observers etc so they can't even apply such logic that would count error listeners on those observers etc (that's not a detail that is surfaced to the code that forwards the error to multiple observers). We have that looping centralized~ so I don't see why we wouldn't leverage that.

Note though that the code that you have quoted here isn't exactly about that. This is an extra `try/catch` that ensures that we report errors that might be thrown by error listeners themselves. At the same time, we ensure that each error listener is notified about the original error.

packages/core/src/interpreter.ts
Andarist: note for me: this should stop the actor by using appropriate error procedures - `stop` schedules a stop signal and that shouldn't happen

packages/core/src/interpreter.ts
```diff
@@ -288,29 +317,29 @@ export class Interpreter<
...
+    let caughtError;
     try {
-      const nextState = this.logic.transition(
-        this._state,
-        event,
-        this._actorContext
-      );
+      nextState = this.logic.transition(this._state, event, this._actorContext);
+    } catch (err) {
+      // we wrap it in a box so we can rethrow it later even if falsy value gets caught here
```
davidkpiano, line 326: Clever! (in a good way)

packages/core/test/errors.test.ts
```diff
@@ -40,18 +42,102 @@ describe('error handling', () => {
...
+    expect(errorSpy).toHaveBeenCalledTimes(0);
+
+    installGlobalOnErrorHandler((ev) => {
+      ev.preventDefault();
+      expect(ev.error.message).toEqual(
+        'doesnt_notify_error_listener_when_error_is_thrown_in_subscribe'
+      );
+      setTimeout(() => {
+        done();
+      }, 2000);
```
davidkpiano, line 128: Should it be this long?

Andarist: actually, this is just some debugging leftover - it should just call `done()` right away

.changeset/chilled-bobcats-lay.md
```diff
@@ -0,0 +1,16 @@
...
+- Actors will no longer crash when an error is thrown in an observer (`actor.subscribe(observer)`).
+- Errors will be handled by observer's `.error()` handler:
+  ```ts
+  actor.subscribe({
+    error: (error) => {
+      // handle error
+    }
+  });
+  ```
+- If an observer does not have an error handler, the error will be thrown normally.
```
Andarist, line 16:
```suggestion
- If an observer does not have an error handler, the error will be thrown in a clear stack so bug tracking services can collect it.
```

PR #4157, opened 2023-07-26
[v5] Invoke callback as single object arg

Valkendorm:
Merge `sendBack` and `receive` with other properties on `InvokeCallback` type.

```ts
const callbackLogic = fromCallback(({ input, service, self, sendBack, receive }) => { ... });
```

### Previously

```ts
const callbackLogic = fromCallback((sendBack, receive, { input, self, service }) => { ... });
```

From discussion: https://discord.com/channels/795785288994652170/1133419826111578112/1133503134245859329


davidkpiano:
Thanks, this looks great! While waiting for @Andarist's feedback, can you write a changeset?

`yarn changeset` and follow the instructions (`minor` for `xstate`)

packages/core/src/types.ts
```diff
@@ -301,15 +301,19 @@ export type Receiver<TEvent extends EventObject> = (
 export type InvokeCallback<
```
Andarist, line 301: while we are at it, could we move this type to `packages/core/src/actors/callback.ts`? Same for `Receiver`

packages/core/src/actors/callback.ts
```diff
@@ -50,10 +50,12 @@ export function fromCallback<TEvent extends EventObject>(
...
         };
 
-        state.dispose = invokeCallback(sender, receiver, {
+        state.dispose = invokeCallback({
           input: state.input,
           system,
-          self: self as TODO
+          self: self as TODO,
+          sendBack: sender,
+          receive: receiver
```
Andarist, line 58: It would make sense to rename the local variables to use the same names as the ones that we expose to the user now

packages/core/src/types.ts
Valkendorm: Makes sense. I've kept them exported, but might not be no longer necessary. Let me know what you want.

Valkendorm:
> Could you sync this with `next`? There are merge conflicts here now (sorry about that!)

Should be good now. Thanks for the heads up.

PR #4168, opened 2023-07-31
Make sure useSelector.ts doesn’t create stale values

enyo:
The function provided to svelte's `readable` is only invoked when the first subscriber subscribes to this store.

But because stately’s `actor.subscribe` does not fire immediately after subscribing (it only fires when something changes in the state) the svelte store could have a stale value (with the initial `prevSelected`) until the state in the machine changes and triggers the update.

This PR fixes that, by making sure that the value is not stale when the first subscription happens.

### DIFF
--- a/packages/xstate-svelte/src/useSelector.ts
+++ b/packages/xstate-svelte/src/useSelector.ts
@@ -15,13 +15,18 @@ export const useSelector = <TActor extends ActorRef<any, any>, T>(
   let prevSelected = selector(actor.getSnapshot());
 
   const selected = readable(prevSelected, (set) => {
-    sub = actor.subscribe((state) => {
+    const onNext = (state: SnapshotFrom<TActor>) => {
       const nextSelected = selector(state);
       if (!compare(prevSelected, nextSelected)) {
         prevSelected = nextSelected;
         set(nextSelected);
       }
-    });
+    }
+
+    // Make sure the store gets updated when it's subscribed to.
+    onNext(actor.getSnapshot());
+
+    sub = actor.subscribe(onNext);
 
     return () => {
       sub.unsubscribe();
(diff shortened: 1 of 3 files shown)


enyo:
@Andarist done!

I think the comments in the code should be self descriptive, but here is a short summary since it’s pretty obscure Svelte behavior:

Using `$count` _anywhere_ in the Svelte template, immediately creates a subscription of the store under the hood. That’s why it’s not possible to test this behavior with:

```svelte
{#if $state.context.count >= 2}
  <div data-testid="selectorOutput”>{$count}</div>
{/if}
```

Instead I created another variable (`readCount`) that uses the `get` Svelte store function to read the value only when it’s supposed to be displayed (which means that the store is only subscribed to and read at a later time, and not from the start).

Without the fix in this PR, the value would still read `0` at the end, because the `count` store would still provide the initial value it received when it was created.

enyo:
You might be wondering why this fix is needed in the first place, since you nearly always use a Svelte store in the template with the `$count` shorthand.

The bug appeared in our case, because we created the store and put it in a svelte context:

```ts
const count = useSelector(/* etc... */);
setContext(‘count’, count);
```

And then in some child component, at a later time, we used that store:

```ts
const count = getContext(‘count’);
```

And this is where the value was wrong.

Andarist:
I wonder, are you using `@xstate/svelte@beta` by copy-pasting it into your project? That's fine - I was just double-checking things and this package isn't yet released, hence the question.

enyo:
> I wonder, are you using `@xstate/svelte@beta` by copy-pasting it into your project? That's fine - I was just double-checking things and this package isn't yet released, hence the question.

Yes that’s exactly what I’m doing.

enyo:
The same bug exists in v4, but since I’m using v5 and this is going to be the next version anyway, I thought I might just fix it in v5.

PR #4172, opened 2023-08-02
[v5] `onSnapshot` for all actors

davidkpiano:
This PR adds `onSnapshot: { ... }` support for all supported invoked actor logic.


davidkpiano:
@Andarist I need help because the test failures are only related to `console.warn` unexpected calls (locally, `yarn test` passes ✅ )

packages/core/src/StateMachine.ts
```diff
@@ -384,10 +384,10 @@ export class StateMachine<
     state: State<TContext, TEvent, TActor, TOutput, TResolvedTypesMeta>
   ) {
     return state.error
-      ? { status: 'error', data: state.error }
+      ? ({ status: 'error', error: state.error } as const)
       : state.done
-      ? { status: 'done', data: state.output }
-      : { status: 'active' };
+      ? ({ status: 'done', data: state.output } as const)
+      : ({ status: 'active' } as const);
```
Andarist, line 390: this is unrelated change

packages/core/src/actions/invoke.ts
```diff
@@ -123,6 +144,7 @@ export function invoke<
   invoke.systemId = systemId;
   invoke.src = src;
   invoke.input = input;
+  invoke.subscribe = !!onSnapshot;
```
Andarist, line 147: I like this one very much - but perhaps we could rename this somehow? `syncSnapshot`? `iLikeChatter`? `listenToSnapshots`?

packages/core/src/actions/invoke.ts
```diff
@@ -53,6 +55,23 @@ function resolve(
             })
           : configuredInput
     });
+
+    if (subscribe) {
+      actorRef.subscribe({
+        next: (snapshot) => {
+          if (actorContext.self.status === ActorStatus.Running) {
```
Andarist, line 62: is there any test case that would exercise the need for this if statement?

packages/core/src/actions/invoke.ts
```diff
@@ -53,6 +55,23 @@ function resolve(
...
+            actorContext.self.send({
+              type: `xstate.snapshot.${id}`,
+              id,
+              snapshot
+            });
+          }
+        },
+        error: () => {
+          /* TODO */
+        }
```
Andarist, line 72: that shouldn't be here as if we are going to add an error listener to the `actorRef` then it will be done outside of this if block

packages/core/src/spawn.ts
```diff
@@ -45,16 +45,52 @@ export function createSpawner(
             : input,
         systemId
       }) as any;
-      spawnedChildren[actor.id] = actor;
-      return actor;
+      spawnedChildren[actorRef.id] = actorRef;
+
+      // if (options.subscribe) {
```
Andarist, line 50: looks like a leftover - this if statement here should be uncommented

packages/core/src/spawn.ts
```diff
@@ -45,16 +45,52 @@ export function createSpawner(
...
+            actorContext.self.send({
+              type: `xstate.snapshot.${actorRef.id}`,
+              snapshot,
+              id: actorRef.id
+            });
+          }
+        },
+        error: () => {
+          /* TODO */
+        }
```
Andarist, line 63: same as https://github.com/statelyai/xstate/pull/4172/files#r1304354021

packages/core/src/spawn.ts
```diff
@@ -45,16 +45,52 @@ export function createSpawner(
...
       // TODO: this should also receive `src`
-      return createActor(src, {
+      const actorRef = createActor(src, {
         id: options.id,
         parent: actorContext.self,
         input: options.input,
         systemId
       });
+
+      // if (options.subscribe) {
```
Andarist, line 76: same as above

packages/core/src/spawn.ts
```diff
@@ -45,16 +45,52 @@ export function createSpawner(
...
+            actorContext.self.send({
+              type: `xstate.snapshot.${actorRef.id}`,
+              snapshot,
+              id: actorRef.id
+            });
+          }
+        },
+        error: () => {
+          /* TODO */
+        }
```
Andarist, line 89: same as above

packages/core/test/invoke.test.ts
```diff
@@ -2048,20 +2069,23 @@ describe('invoke', () => {
       actor.subscribe({
         complete: () => {
           done();
-        }
+        },
+        error: () => {}
```
Andarist, line 2073: this looks like an unrelated change - and one that could suppress legit errors in the future

packages/core/src/actions/invoke.ts
davidkpiano: `syncSnapshot` it is

packages/core/src/actions/invoke.ts
davidkpiano: Not sure how to resolve this; removing this causes a test failure

packages/core/src/interpreter.ts
```diff
@@ -225,8 +225,8 @@ export class Actor<
         break;
       case 'error':
         this._stopProcedure();
-        this._error(status.data);
-        this._parent?.send(error(this.id, status.data));
+        this._parent?.send(error(this.id, status.error));
+        this._error(status.error);
```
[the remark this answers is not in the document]
davidkpiano: I don't remember so I un-reordered it

.changeset/mean-pianos-explode.md
```diff
@@ -0,0 +1,45 @@
+---
+'xstate': minor
+---
+
+The `onSnapshot: { ... }` transition object is now supported for invoked machines, observables, promises, and transition functions:
+
+```ts
+const machine = Machine({
```
Andarist, line 8:
```suggestion
const machine = createMachine({
```

.changeset/mean-pianos-explode.md
```diff
@@ -0,0 +1,45 @@
...
+      }
+    },
+    {
+      src: fromPromise(() => { ... }),
+      onSnapshot: {
+        actions: (context, event) => {
+          event.snapshot; // promise value
+        }
+      }
+    },
```
Andarist, line 26:
```suggestion
```

packages/core/src/actions/invoke.ts
Andarist: I'm not 100% sure what I was thinking about here 😅 I had some concerns related to this right now but I think they are addressed, see this test case:
https://github.com/statelyai/xstate/blob/23402e34c178ba9a000e22c54765090b8fcfc52f/packages/core/test/errors.test.ts#L213

packages/core/test/actor.test.ts
```diff
@@ -487,7 +491,10 @@ describe('spawning observables', () => {
         idle: {
           entry: assign({
             observableRef: ({ spawn }) => {
-              const ref = spawn(observableLogic, { id: 'int' });
+              const ref = spawn(observableLogic, {
+                id: 'int',
+                syncSnapshot: true
```
Andarist, line 496: It's worth noting that we don't have any means right now to validate that this is actually safe on the type-level

Andarist (approved):

PR #3940, opened 2023-04-04
[v5] Strongly-typed state value

davidkpiano:
https://user-images.githubusercontent.com/1093738/229780669-d0af6444-d356-4846-9282-aaf9f4513329.mp4

CleanShot 2023-04-03 at 19 17 16@2x

PR #4173, opened 2023-08-03
[v5] Strongly-typed state value

davidkpiano:
Trying this again - supersedes #3940

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1044,7 +1044,7 @@ export type MachineConfig<
     : { context: InitialContext<LowInfer<TContext>, TInput> });
 
 export type MachineStates = {
-  [key: string]: any;
+  states?: Record<string, MachineStates>;
 };
 
 export interface ProvidedActor {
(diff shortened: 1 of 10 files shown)


packages/core/src/types.ts
```diff
@@ -1043,6 +1043,14 @@ export type MachineConfig<
     ? { context?: InitialContext<LowInfer<TContext>, TInput> }
     : { context: InitialContext<LowInfer<TContext>, TInput> });
 
+export type MachineStates<Literal = string> = {
+  states?: Record<string, MachineStates<Literal>>;
+  on?: unknown;
+  invoke?: SingleOrArray<{
+    src: Literal | NonReducibleUnknown;
+  }>;
```
Andarist, line 1051: this fixes the contextual type for `invoke.src` so it can be retained as a string literal

Andarist:
So far we can't infer what is needed for this feature and the current attempt at inferring this is in: https://github.com/statelyai/xstate/pull/4262 . I think there is no reason to have this PR open until we manage to work out the inference part. Let's close it for now, for housekeeping purposes. It's referenced by that other PR so we can always track it back and reopen when/if needed.

PR #4176, opened 2023-08-05
[v5] Rename `interpret` to `createActor`

davidkpiano:
The `interpret(...)` function has been deprecated and renamed to `createActor(...)`:

```diff
-import { interpret } from 'xstate';
+import { createActor } from 'xstate';

-const actor = interpret(machine);
+const actor = createActor(machine);
```

### DIFF
--- a/.changeset/ten-singers-rhyme.md
+++ b/.changeset/ten-singers-rhyme.md
@@ -1,0 +1,13 @@
+---
+'xstate': major
+---
+
+The `interpret(...)` function has been deprecated and renamed to `createActor(...)`:
+
+```diff
+-import { interpret } from 'xstate';
++import { createActor } from 'xstate';
+
+-const actor = interpret(machine);
++const actor = createActor(machine);
+```
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -6,7 +6,12 @@ export { log } from './actions/log.ts';
 export { pure } from './actions/pure.ts';
 export { raise } from './actions/raise.ts';
 export { stop } from './actions/stop.ts';
-import { interpret, Interpreter, ActorStatus } from './interpreter.ts';
+import {
+  createActor,
+  interpret,
+  Interpreter,
+  ActorStatus
+} from './interpreter.ts';
 import { createMachine } from './Machine.ts';
 import { mapState } from './mapState.ts';
 import { State } from './State.ts';
@@ -26,7 +31,8 @@ export {
   sendTo,
   sendParent,
   forwardTo,
-  interpret,
+  createActor,
+  interpret, // deprecated
   Interpreter,
   ActorStatus as InterpreterStatus,
   doneInvoke,
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -534,6 +534,34 @@ export class Interpreter<
  * @param machine The machine to interpret
  * @param options Interpreter options
  */
+export function createActor<TMachine extends AnyStateMachine>(
+  machine: AreAllImplementationsAssumedToBeProvided<
+    TMachine['__TResolvedTypesMeta']
+  > extends true
+    ? TMachine
+    : MissingImplementationsError<TMachine['__TResolvedTypesMeta']>,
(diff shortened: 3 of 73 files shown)


packages/core/src/interpreter.ts
```diff
@@ -534,6 +534,34 @@ export class Interpreter<
...
+}
+
+/**
+ * Creates a new Interpreter instance for the given machine with the provided options, if any.
+ *
+ * @deprecated Use `createActor` instead
+ * @param machine The machine to interpret
+ * @param options Interpreter options
+ */
 export function interpret<TMachine extends AnyStateMachine>(
```
Andarist, line 565: this can probably be replaced by just `export function interpret = createActor;` (with the `@deprecated` tag)

packages/core/src/interpreter.ts
```diff
@@ -534,6 +534,34 @@ export class Interpreter<
  * @param machine The machine to interpret
```
Andarist, line 534: Interpreter is still mentioned in the comments here. It would be cool to also rename this file and the `Interpreter` class itself

packages/core/src/index.ts
```diff
@@ -26,7 +31,8 @@ export {
   sendTo,
   sendParent,
   forwardTo,
-  interpret,
+  createActor,
+  interpret, // deprecated
   Interpreter,
   ActorStatus as InterpreterStatus,
```
Andarist, line 37: `InterpreterStatus` should be phased out (or it should just be an alias of `ActorStatus` with the `@deprecated` tag). Other packages should start depending on the `ActorStatus` itself.

packages/core/src/interpreter.ts
```diff
@@ -534,6 +534,34 @@ export class Interpreter<
...
  * @param options Interpreter options
  */
+export function createActor<TMachine extends AnyStateMachine>(
+  machine: AreAllImplementationsAssumedToBeProvided<
+    TMachine['__TResolvedTypesMeta']
+  > extends true
+    ? TMachine
+    : MissingImplementationsError<TMachine['__TResolvedTypesMeta']>,
+  options?: InterpreterOptions<TMachine>
```
Andarist, line 543: we probably should also rename all other types that mention the interpreter, for example: `InterpreterOptions` -> `ActorOptions`

packages/xstate-vue/test/UseActorCreateSimple.vue
```diff
@@ -13,10 +13,10 @@ import { ActorRef } from 'xstate';
 import { defineComponent, shallowRef } from 'vue';
 
 import { useActor } from '../src/index.ts';
-import { interpret } from 'xstate/src';
+import { createActor } from 'xstate/src';
```
Andarist, line 16:
```suggestion
import { createActor } from 'xstate';
```

.changeset/ten-singers-rhyme.md
```diff
@@ -0,0 +1,13 @@
+---
+'xstate': major
+---
+
+The `interpret(...)` function has been deprecated and renamed to `createActor(...)`:
+
+```diff
+-import { interpret } from 'xstate';
++import { createActor } from 'xstate';
```
Andarist, line 9: just a note: it might be confusing that we have `createActor` and `createMachine` but they don't actually do a similar thing whereas things that do similar things to `createMachine` don't share the same naming scheme (`createMachine`, `fromPromise`, etc)

packages/core/src/interpreter.ts
davidkpiano: Ooh smart

.changeset/ten-singers-rhyme.md
davidkpiano: Yeah perhaps, but the alternatives aren't really better:

- `createMachine` to `fromMachine` = we're creating a machine from a machine...?
- `fromMachineConfig` is a bit too wordy
- `createMachineFromPromise` wordy and confusing
- `createPromise` we're not creating a promise though

davidkpiano:
Merging this in (big PR, all comments resolved, straightforward change) - won't release until next week

PR #4184, opened 2023-08-09
[v5] Routable states

davidkpiano:
This PR enables developers to mark states as "routes" by providing a `route` config, which is just a normal transition config without the target:

```ts
const todoMachine = createMachine({
  type: 'parallel',
  states: {
    todo: {
      initial: 'new',
      states: {
        new: {},
        editing: {}
      }
    },
    filter: {
      initial: 'all',
      states: {
        all: {
          route: {}
        },
        active: {
          route: {}
        },
        completed: {
          route: {}
        }
      }
    }
  }
});

const todoActor = createActor(todoMachine).start();

todoActor.send({ type: 'xstate.route.filter.active' });
// => transitions directly to { todo: 'new', filter: 'active' }
```

The motivation for this is that there are many use-cases for wanting to transition directly to a state, such as in modeling page routes or steps in a multi-step form. Previously, you needed to manually hard-code transitions for these, which resulted in an unnatural and verbose state machine definition.

### DIFF
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -453,6 +453,18 @@ export function formatTransitions<
     }
     existing.push(delayedTransition);
   }
+  Object.values(stateNode.states).forEach((sn) => {
+    if (sn.config.route) {
+      const eventType = `xstate.route.${sn.path.join('.')}`;
+      const transition: TransitionConfig<TContext, TEvent> = {
+        target: `.${sn.key}`
+      };
+
+      transitions.set(eventType, [
+        formatTransition(stateNode, eventType, transition)
+      ]);
+    }
+  });
   return transitions as Map<string, TransitionDefinition<TContext, any>[]>;
 }
 
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -457,6 +457,7 @@ export function formatTransitions<
     if (sn.config.route) {
       const eventType = `xstate.route.${sn.path.join('.')}`;
       const transition: TransitionConfig<TContext, TEvent> = {
+        ...sn.config.route,
         target: `.${sn.key}`
       };
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -669,7 +669,19 @@ export interface StateNodeConfig<
    * A default target for a history state
    */
   target?: string;
-  route?: boolean;
+  route?: RouteTransitionConfig<TContext, TEvent>;
+}
+
+export interface RouteTransitionConfig<
+  TContext extends MachineContext,
+  TEvent extends EventObject
+> {
+  guard?: GuardConfig<TContext, TEvent>;
+  actions?: Actions<TContext, TEvent>;
+  reenter?: boolean;
+  meta?: Record<string, any>;
+  description?: string;
+  // no target because target is self
 }
 
 export interface StateNodeDefinition<
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -669,6 +669,7 @@ export interface StateNodeConfig<
    * A default target for a history state
    */
   target?: string;
+  route?: boolean;
 }
 
 export interface StateNodeDefinition<
--- a/packages/core/test/route.test.ts
+++ b/packages/core/test/route.test.ts
@@ -1,0 +1,31 @@
+import { createActor, createMachine } from '../src';
+
+describe('route', () => {
+  it('should transition directly to a route if route: true', () => {
+    const machine = createMachine({
+      id: 'routeTest',
+      initial: 'a',
+      states: {
+        a: {},
+        b: {
+          route: true
+        },
+        c: {}
+      }
+    });
+
+    const actor = createActor(machine).start();
+
+    actor.send({
+      type: 'xstate.route.b'
+    });
+
+    expect(actor.getSnapshot().value).toEqual('b');
+
+    actor.send({
+      type: 'xstate.route.c'
+    });
+
+    expect(actor.getSnapshot().value).toEqual('b');
+  });
+});
--- a/packages/core/test/route.test.ts
+++ b/packages/core/test/route.test.ts
@@ -1,14 +1,14 @@
 import { createActor, createMachine } from '../src';
 
 describe('route', () => {
-  it('should transition directly to a route if route: true', () => {
+  it('should transition directly to a route if route is an empty transition config', () => {
     const machine = createMachine({
       id: 'routeTest',
       initial: 'a',
       states: {
         a: {},
         b: {
-          route: true
+          route: {}
         },
         c: {}
       }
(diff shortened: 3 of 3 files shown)


christophe-g:
@davidkpiano 

I am quite interested in this for binding the route with machine state actors. 

Now that `emit`  has landed (thanks!), would it make sense to emit a `route` event any time a  state enters via `actor.send({ type: 'xstate.route.anyRouteState')}` ? 

This would facilitate keeping the route in sync with the active state. Happy to provide more context if needed.

christophe-g:
> More context would be appreciated.

@davidkpiano - this is not really more context, but a new lit controller binding xstate actor machine with a router. 

- More info: https://github.com/lit-apps/lit-app/tree/dev/packages/xstate-route
- Life example: https://xstate-route.web.app/home/fsm/one

> and it may be better to .subscribe(…) to the state changes and map those to URL states.

I took your advice for this controller ...

Any change to have this PR merged ?

christophe-g:
Dear @davidkpiano - just interested to know if this PR has any chance to be merged, or if there are any blocking issues ?

I have build a lit-controller around this (see above). [This PR allows to de-couple business logic (handled by xstate machine), routing / navigation (natively handled by the browser), and views (provided by the router / also bound to state machine)](https://github.com/lit-apps/lit-app/tree/dev/packages/xstate-route), which is great!

christophe-g:
> It does have a chance! Going to update it soon.

@davidkpiano - Fantastic, thank you !

handletastic:
What's the current state? 👍🏼 Would love to see this ! 🚀

christophe-g:
I see that this PR was marked as ready for review, but this might have gone unnoticed ; )

Just adding a comment here to try revive this powerful feature - and thanks anyway for this great piece of code !

Hebilicious:
For anyone that needs this, I've implemented a verbose workaround: 

1) Create a `routeTo` property in the context for the route
2) Define a guard per routable state that check for `routeTo`
3) Create a `RouteTo` event with an array of all routes+ guard

Then you can route to a specific state by assigning `routeTo` and sending the `RouteTo` event

christophe-g:
> Routable state would simplify things massively though

Routable states are a hidden gem of xstate. Using with a lot of joy for instance in relatively complex stepper UI, allowing to bind the view with the state through the route.

joernroeder:
I'd love to see this as part of xstates core. still plans to merge this in @davidkpiano?

minhazfm:
This would be a huge help in keeping browser routing in sync with state machines! Thankful for your hard work @davidkpiano! Hope to see this available in V5 soon!

davidkpiano:
> I'd love to see this as part of xstates core. still plans to merge this in @davidkpiano?

Yes, I'd still like to see this merged in soon

handletastic:
🙏🏼 What faith must I convert to for my plea to realise? 🥺 @davidkpiano

packages/core/src/types.ts
```diff
@@ -2542,6 +2580,32 @@ export type StateId<
         }>
       : never);
 
+export type RoutableStateId<
```
Andarist: I wonder if we shouldnt restrict those routes to be addressable only by state IDs. It would mostly just simplify those types here ;p

davidkpiano: Do you mean something like:

`actor.send({ type: 'xstate.route', path: 'someNamedPath' })` alias?

Andarist: Ah, ye - that would be a nice improvement too, to separate the path from the event's type. But this comment was about `xstate.route` only working with `#statesWithIds`.

packages/core/src/types.ts
```diff
@@ -1038,6 +1038,43 @@ export interface StateNodeConfig<
 
   /** A default target for a history state */
   target?: string | undefined; // `| undefined` makes `HistoryStateNodeConfig` compatible with this interface (it extends it) under `exactOptionalPropertyTypes`
+  route?: RouteTransitionConfig<
```
Andarist, line 1041: to clarify - this is a transition config that will be unique now because it's defined on the *target* state and not on the *source* state, right?

packages/core/src/stateUtils.ts
```diff
@@ -380,6 +380,28 @@ export function formatTransitions<
     }
     existing.push(delayedTransition);
   }
+  // Collect routes from all descendants (not just direct children)
+  // so routes are accessible from anywhere in the machine
```
Andarist: I get why this is implemented this way... but damn, that's quite terrible from the Big O's PoV. I don't think we should ship such an algorithmic deficiency

packages/core/src/types.ts
Andarist: If we'd go with a single `type: 'xstate.route'` approach it would be rather easy to address [this](https://github.com/statelyai/xstate/pull/4184/changes#r2783843078) by specialcasing routes. You could collect routes at the root and just handle them right away by matching the collected routes instead of going through `.on`/transitions lookup. In fact, you could probably only collect routes at the root's `.on` anyway, right? They are available from anywhere with the machine and the root transitions are just that. You don't have to repeat them an all states

packages/core/src/types.ts
```diff
@@ -1038,6 +1038,43 @@ export interface StateNodeConfig<
...
+    TExpressionEvent,
+    TEvent,
+    undefined,
+    TActor,
+    TAction,
+    TGuard,
+    TDelay,
+    TEmitted
+  >;
+  reenter?: boolean;
```
Andarist: given [this](https://github.com/statelyai/xstate/pull/4184#discussion_r2783828377) I'm not sure how to even interpret `reenter` on those route transitions, should this be configurable for them?

packages/core/test/route.test.ts
Andarist, line 1: some tests related to taking a route transition while being within the routable state already would be appreciated

packages/core/src/stateUtils.ts
pedrovhb: Sorry to butt in, but I was just looking around and this PR caught my eye because, if it were merged, it would allow for some really nasty code to disappear 😄 

If I understand correctly, this isn't really bad, or even improvable, is it? It's possible I'm missing some context, but it seems to me -

It looks dangerous at first glance because there's a recursive call inside `Object.values(states).forEach` - but really it's just a DFS traversal, visiting each node once. If the goal is to discover all routes defined anywhere in the hierarchy, I’m not sure there’s an asymptotically better approach than O(N), short of indexing routes somehow which would be a whole other thing.

packages/core/src/types.ts
davidkpiano: Yes

packages/core/src/types.ts
davidkpiano: Implemented this here; now states require an ID to be routable and the event is simplified/merged:

`actor.send({ type: 'xstate.route', to: 'someKnownId' })`

[9d02e91](https://github.com/statelyai/xstate/pull/4184/commits/9d02e91011f0b6a1f09306b571574112ffd78e41)

packages/core/src/types.ts
davidkpiano: Removing this

packages/core/test/route.test.ts
davidkpiano: Added [cf31cfd](https://github.com/statelyai/xstate/pull/4184/commits/cf31cfd393371fd54012fdeffdc0167b72a52b2c)

packages/core/src/stateUtils.ts
davidkpiano: Optimized the heck out of it: now it only runs on the root once - O(N) - and I'm sure it can be optimized further but it's fine for now: 4e09eedc7b96479a8e06dc825b7cf521653812c0

packages/core/src/types.ts
```diff
@@ -2499,6 +2535,7 @@ export type ToChildren<TActor extends ProvidedActor> =
 
 export type StateSchema = {
   id?: string;
+  route?: any;
```
Andarist: nit:
```suggestion
  route?: unknown;
```

packages/core/test/route.test.ts
```diff
@@ -282,7 +333,7 @@ describe('route', () => {
     // Should be able to route to deeply nested state from root
     expect(actor.getSnapshot().value).toEqual('home');
 
-    actor.send({ type: 'xstate.route.app.dashboard.overview' });
+    actor.send({ type: 'xstate.route', to: 'overview' });
```
Andarist: making this `to: "#overview"` would likely allow people to target descendants of "named" states - like `to: "#overview.nested.child"`. Given we could just use our regular target resolution for that (if my memory serves me well). Whereas I don't think it would be good to allow the same with `overview.nested.child`.

Making the `#` explicit would also just nicely match what u have to use as a target pattern in the machine's config - so it would nicely pair up.

davidkpiano: Discussed offline: only change here is to require the `#` - we aren't going to do descendant target resolution (yet? remains TBD)

davidkpiano: Done here [01d3fff](https://github.com/statelyai/xstate/pull/4184/commits/01d3fff8965573f0eb1df81320f418e586491589)

handletastic: I am far removed about the implications for any of the implementation details, or from the internal mechanics at play even plain ergonomics. I am aware that this has been already merged and I am not trying to be a PITA.

However, if I understand correctly,  and correct me firmly if I am wrong (just trying to learn it because I intend to use this functionality asap), the intention here is to enforce that that nested child routes to require a `#` prepended. 

How would this then handle if we were to have "injected" configurations that go multiple levels deep?

If I am totally off mark, if you could please provide a simplified example/explanation.

Thanks so much for your great work guys. ❤️

davidkpiano: @handletastic Great question, and you're not off at all.

Yes, the intent is to route by explicit state node ID, not by nested path-like syntax. So #dashboard.overview is intentionally not supported.

A route target is considered valid when a state has both:
```ts
id: 'someId'
route: { ... }
```
Depth does not matter. Even if the state is injected or many levels deep, routing still works as long as that nested state has a stable explicit id and route config. Then you can route to it from anywhere with:
```ts
{ type: 'xstate.route', to: '#someId' }
```
So the model is: explicit IDs for routable states, globally addressable by `#id`.

What specifically did you mean by "injected" configurations?

PR #4190, opened 2023-08-12
[v5] Resumable promise logic

davidkpiano:
This PR introduces a `step(...)` wrapper to persist intermediate state in promises.

The `step(name, promiseFn)` function executes the promise created by the `promiseFn` the first time, and persists its result to the actor's internal state.

The promise actor's state can then be persisted (`actor.getPersistedState()`) and restored normally (`createActor(logic, { state })`) and the logic will essentially "resume" at the latest incomplete `step(...)`. The way this works is by using persisted values for previous steps; if there is a persisted value, no need to run the promise again.

```ts
const flow = fromPromise(async ({ step }) => {
  const user = await step('user', () => fetchUser());

  const friends = await step('friends', () => fetchFriends(user.id));

  if (friends.length > 0) {
    const party = await step('plan party', () => createEvent(user, friends));
    return party;
  }

  return null;
});
```

**Motivation**

This would be _completely opt-in_ - you can create promise logic as normal. Calling async functions in a named `step(name, promiseFn)` gives you a couple benefits:

- **Simple visualization** - we can show a basic partial state machine for previously executed steps
- **Persistence** - we can persist the results of each step (make sure they're uniquely named!) so that when we restore a promise, it can fast-forward to where you left off
- **Observability** - when hooked up to Stately in the future, you can see the status/result of each step visualized on a sequence diagram

### DIFF
--- a/packages/core/src/actors/promise.ts
+++ b/packages/core/src/actors/promise.ts
@@ -50,7 +50,7 @@ export function fromPromise<T, TInput>(
     input: TInput;
     system: AnyActorSystem;
     self: PromiseActorRef<T>;
-    step: <T>(name: string, promise: Promise<T>) => Promise<T>;
+    step: <T>(name: string, promiseCreator: () => Promise<T>) => Promise<T>;
   }) => PromiseLike<T>
 ): PromiseActorLogic<T, TInput> {
   // TODO: add event types
@@ -101,7 +101,12 @@ export function fromPromise<T, TInput>(
         return;
       }
 
-      function step(name: string, promise: Promise<any>) {
+      function step(name: string, promiseCreator: () => Promise<any>) {
+        if (state.steps?.[name]) {
+          return state.steps[name][1];
+        }
+
+        const promise = promiseCreator();
         return promise.then(
           (result) => {
             self.send({
--- a/packages/core/src/actors/promise.ts
+++ b/packages/core/src/actors/promise.ts
@@ -10,6 +10,7 @@ export interface PromiseInternalState<T, TInput = unknown> {
   status: 'active' | 'error' | 'done' | 'canceled';
   data: T | undefined;
   input: TInput | undefined;
+  steps?: Record<string, [status: 0 | 1 | 2, result: any]>;
 }
 
 const resolveEventType = '$$xstate.resolve';
@@ -49,6 +50,7 @@ export function fromPromise<T, TInput>(
     input: TInput;
     system: AnyActorSystem;
     self: PromiseActorRef<T>;
+    step: <T>(name: string, promise: Promise<T>) => Promise<T>;
   }) => PromiseLike<T>
 ): PromiseActorLogic<T, TInput> {
   // TODO: add event types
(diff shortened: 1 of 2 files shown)


packages/core/src/actors/promise.ts
```diff
@@ -91,8 +101,32 @@ export function fromPromise<T, TInput>(
         return;
       }
 
+      function step(name: string, promiseCreator: () => Promise<any>) {
+        if (state.steps?.[name]) {
+          return state.steps[name][1];
+        }
+
```
pixtron, line 108: Could the step be visualized as "pending" when something like a `$$xstate.step.pending` event would be sent?

phpnode:
I really like this idea, but I think it's going to be very easy to forget to use `step()` somewhere, and that would have really unpleasant consequences when trying to resume a function. TypeScript can't help either, as there's no way to constrain what can be awaited. If this used generators instead it would be possible to ensure that `step()` is always used, both at runtime and when type checking.

ellemedit:
could we remove first parameter? how do you think about using reference check of promise as React's useThenable?

davidkpiano:
> I really like this idea, but I think it's going to be very easy to forget to use `step()` somewhere, and that would have really unpleasant consequences when trying to resume a function. TypeScript can't help either, as there's no way to constrain what can be awaited. If this used generators instead it would be possible to ensure that `step()` is always used, both at runtime and when type checking.

You have the same problem with generators, since it's possible to make a `promise.then(...)` call without wrapping it or execute some unwrapped side-effect.

-----

The important thing to remember here is that this is an **opt-in** feature. The default behavior is that the promise (or parts of the promise) will _re-run_. So it will still work, but this _may_ be undesirable. However, I'd argue that it's impractical to prevent every possible way that a developer may forget to do something or do things wrongly, outside of providing lint rules.

IMO if strict guardrails are desired, using a state machine would be the way to go.

davidkpiano:
> could we remove first parameter of step? how do you think about using reference check of promise as React's useThenable?

If we remove the first parameter, then 1) we don't have explicit names for steps, which makes it harder to visualize, and 2) similar to React hooks, we're now dependent on the order in which these steps get called, which limits the ability to call these within e.g. loops or if-statements.

Reference checking wouldn't work when restoring state (e.g. from localstorage or a DB).

pixtron:
> If we remove the first parameter, then 1) we don't have explicit names for steps

Maybe `name` could be inferred from named functions, and overridden by `name` attribute?

```ts
function step(promiseCreator: () => Promise<any>, nameOverride?: string) {
  let name = promiseCreator.name;
  if (nameOverride) name = nameOverride;
  
  if(!name) {
    if (isDevelopment) {
      console.warn('...')
    }
    return promiseCreator();
  }
        
  if (state.steps?.[name]) {
    return state.steps[name][1];
  }

  ...
}
```

Usage:
```ts
const fetchUser = () => fetch(url);
step(fetchUser); // => name = fetchUser
step(fetchUser, 'user'); // => name = user
step(() => fetch(url), 'user'); // => name = user
step(() => fetch(url)); // => no name, just returns the promise and warns in dev mode
```

ellemedit:
I guess depending 'name' property could make confusion for production as names are mingled formally?

pixtron:
> I guess depending 'name' property of constructor or instance could make confusion for production as names are mingled formally?

That might indeed be problematic for visualization.

davidkpiano:
Will be revisited in v6.

PR #4191, opened 2023-08-15
[v5] provide AbortSignal to fromPromise

pixtron:
This PR allows fromPromise actors to consume an AbortSignal which is aborted once the actor receives the `stop` event.
A consumer can then cancel (long) running promises. Example:

```js
const logic = fromPromise(({event, signal}) => {
  const response = await fetch(`https://example.com/users/${event.userId}`, {signal});
  return response.json();
})
```


Andarist:
This is something that we'd likely want to pull in but it's a purely additive change and it can wait to be done after the v5 release.

davidkpiano:
This has already been done a while ago.

PR #4195, opened 2023-08-17
Remove the script that was rewriting the computed property name in a TS declaration file

Andarist:
We had a note to remove this at some point:
https://github.com/statelyai/xstate/blob/a171ea15100f027eb883c20735767181443e3887/packages/core/rollup.config.js#L38

We already require at least TS 4.1. I think that it's also OK to bump that to 4.2 (in practice we might require a higher version for some things, I only mention the absolute minimum requirement that I'm aware of at the moment).


davidkpiano (approved):
I trust you

PR #4197, opened 2023-08-18
Introduce `TExpressionAction` to make `pure`/`choose` types stricter and tweak `action` arg for inline actions

Andarist:


packages/core/src/stateUtils.ts
```diff
@@ -1432,12 +1440,12 @@ export function resolveActionsAndContext<
       event,
       self: actorCtx?.self,
       system: actorCtx?.system,
-      // TODO: figure out story for `action` and inline actions
-      // what those ones should receive?
-      //
-      // entry: ({ action }) => {}
-      // exit: assign(({ action }) => {})
-      action: typeof action === 'string' ? { type: action } : (action as any)
+      // andarist
```
davidkpiano, line 1443: Andarist

davidkpiano (approved):

PR #4198, opened 2023-08-20
`toPromise(actorRef)`

davidkpiano:
This PR introduces `toPromise(actor)`, which creates a promise from an actor that resolves with `output` when the actor is done.

```ts
import { createMachine, createActor, toPromise } from 'xstate';

const actor = createActor(createMachine({
  // ...
  states: {
    // ...
    done: { type: 'final', output: 42 }
  }
});

actor.start();

const output = await toPromise(actor);
// => 42
```


Andarist:
This supersedes https://github.com/statelyai/xstate/pull/3977 , right? I certainly like this one better 😉

packages/core/src/interpreter.ts
```diff
@@ -320,6 +321,10 @@ export class Actor<
     return this;
   }
 
+  public getOutput() {
+    return this.logic.getOutput?.(this._state);
+  }
```
Andarist, line 326: IMHO it would be could to take another look at how we can standardize status thing, if it would be standard then perhaps we could read it without introducing yet another method to the logic interface. Output is already available within the return value of `getStatus` - it's just that there is no strong contract around the return type of this method so right now it's not that obvious that the logic implemented can do it.

https://github.com/statelyai/xstate/blob/0c2e56270c884abba5d5c200fbede11f071ad2a2/packages/core/src/interpreter.ts#L224

davidkpiano: ~~Done~~ Started here: https://github.com/statelyai/xstate/pull/4199

Necmttn:
Looking forward to this, 
Currently, i have actor machines that execute in serverless environments. 
```
const entryNode = di.editor.getNode(payload.workflowNodeId);
if (!entryNode) {
  throw new Error("Entry node not found");
}
let state = entryNode.actor.getSnapshot();
entryNode.actor.subscribe({
  next: (data) => {
    io.logger.info("STATE", data);
    state = data;
  },
  complete: () => {
    io.logger.log("COMPLETE", state.output);
  },
});
di.engine.execute(entryNode.id, undefined, payload.executionId);
await waitFor(entryNode.actor, (state) => state.matches("complete"), {
  timeout: 1000 * 60 * 5,
});
io.logger.info("STATE", state);
```
Currently, I'm using the `waitFor` method to wait for the actor to get the final state, but as API design wise this doesn't feel good to me. 
another alternative of it kind a like this;
```ts
let state = entryNode.actor.getSnapshot();
await new Promise((resolve, reject) => {
  entryNode.actor.subscribe({
    error(err) {
      reject(err);
    },
    next: (data) => {
      io.logger.info("STATE", data);
      state = data;
    },
    complete: () => {
      resolve(state);
      io.logger.log("COMPLETE", state.output);
    },
  });
});
``` 
Which is also doesn't feel like the best.

davidkpiano:
> This should be enough to "await" the actor with the v5 API:
> 
> ```ts
> await waitFor(actorRef, s => s.status === 'done')
> ```
> 
> We might still implement `toPromise` later on as a convenience but I don't feel like this one-liner is that bad even today.

For reference, this would be more than a one-liner unfortunately:
```ts
await waitFor(actorRef, s => s.status === 'done');
const output = actorRef.getSnapshot().output!;
```

Andarist:
Still a one-liner :P
```ts
const output = (await waitFor(actorRef, s => s.status === 'done')).output!;
```

packages/core/test/toPromise.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+describe('toPromise', () => {
+  it('should be awaitable', async () => {
+    const promiseActor = createActor(
+      fromPromise(() => Promise.resolve(42))
+    ).start();
+
+    const result = await toPromise(promiseActor);
+
+    ((_accept: number) => {})(result);
```
technoplato, line 17: @davidkpiano what is this for?

davidkpiano: Just asserting the type. We can do it using `satisfies`

packages/core/test/toPromise.test.ts
Andarist, line 1: could we also add an error-related test case here? just to verify that the correct data gets thrown - it should be the error itself, not the error snapshot (and the implementation is already working like this)

packages/core/src/toPromise.ts
```diff
@@ -0,0 +1,47 @@
...
+ * const output = await toPromise(actor);
+ *
+ * console.log(output);
+ * // logs { count: 42 }
+ * ```
+ */
+export function toPromise<T extends AnyActorRef>(
+  actor: T
+): Promise<OutputFrom<T>> {
+  // TODO: this is typed as `any`
```
Andarist, line 28: We can't do much about it as long as we use `T extends AnyActorRef` pattern. You unwrap here what you provided for the `TSnapshot` within `ActorRef` but that's just `any` (since you are using `AnyActorRef`)

packages/core/src/toPromise.ts
```diff
@@ -0,0 +1,47 @@
...
+  const currentSnapshot = actor.getSnapshot();
+
+  if (currentSnapshot.status === 'done') {
+    return Promise.resolve(currentSnapshot.output);
+  }
```
Andarist, line 33: This is redundant - we already resolve `complete` callbacks immediately if the actor is already done.

packages/core/src/toPromise.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  if (currentSnapshot.status === 'error') {
+    return Promise.reject(currentSnapshot.error);
+  }
```
Andarist, line 37: Interestingly, this one is not redundant right now. I think this is a problem and I even thought that we had some tests related to this (although maybe indirect ones). I'll take a look at this right now.

packages/core/src/toPromise.ts
Andarist: I believe that we'll be able to remove this once this lands: https://github.com/statelyai/xstate/pull/4570

packages/core/src/toPromise.ts
Andarist: I believe that we'll be able to remove this once this lands: https://github.com/statelyai/xstate/pull/4570

davidkpiano: Approved ✅

Andarist (approved):

PR #4199, opened 2023-08-21
[v5] Actor status object

davidkpiano:


davidkpiano:
@Andarist Need help with the types 🥺

packages/core/src/StateMachine.ts
```diff
@@ -377,10 +377,10 @@ export class StateMachine<
     state: State<TContext, TEvent, TActor, TOutput, TResolvedTypesMeta>
   ) {
     return state.error
-      ? { status: 'error', data: state.error }
+      ? { status: 'error' as const, data: state.error }
```
Andarist, line 380: As done in other PR, we could rename this here: `data` -> `error`

packages/core/src/actors/observable.ts
```diff
@@ -64,14 +66,15 @@ export function fromObservable<T, TInput>(
...
+            output: (event as any).data
           };
         case errorEventType:
           return {
             ...state,
             status: 'error',
+            data: (event as any).data,
             input: undefined,
-            data: (event as any).data, // TODO: if we keep this as `data` we should reflect this in the type
+            output: (event as any).data, // TODO: if we keep this as `data` we should reflect this in the type
```
Andarist, line 77: after the change we'll be able to drop this comment

packages/core/src/interpreter.ts
Andarist, line 1: It would be cool to remove `ActorStatus` from here? piggy-back on the `getStatus` or smth like that?

packages/core/src/actors/observable.ts
```diff
@@ -64,14 +67,15 @@ export function fromObservable<T, TInput>(
           });
           return {
             ...state,
-            data: (event as any).data
+            output: (event as any).data
```
Andarist, line 70: note: this is a mistake - `nextEventType` shouldn't assign to `output`. I'm working on this branch locally so I'll fix this.

PR #4202, opened 2023-08-21
[v5] createSystemObservable utility

pixtron:
This PR does two things: 

- Implements Symbol.iterator on `ActorSystem`
- Adds a new utility function `createSystemObservable(system: ActorSystem)` that lets observer all actors in a system

```js
const systemObservable = createSystemObservable(actor.system);
const sub = systemObservable.subscribe({
  next: ({actorRef, snapshot}) => {},
  error: ({actorRef, error}) => {},
  complete: () => {}
})
```

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -21,6 +21,7 @@ import { StateNode } from './StateNode.ts';
 export { SimulatedClock } from './SimulatedClock.ts';
 export { StateMachine } from './StateMachine.ts';
 export { getStateNodes } from './stateUtils.ts';
+export { createSystemObservable } from './systemObservable.ts';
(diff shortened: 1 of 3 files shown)


Andarist:
What's the most motivating use case for this?

pixtron:
> What's the most motivating use case for this?

There was a question "how to persists parent's state when children change" in discord today. Observers of the parent actor won't get notified upon transitions in children.

https://discord.com/channels/795785288994652170/1134223771792846848/1143185996947652668

smakinson:
Hi everyone. Now that xstate is released, is this something that is on the radar?

davidkpiano:
Reworking how systems work in v6

Issue #4003, opened 2023-05-08
Bug: [v5] TypeScript reports type error when using 'or', 'and' or 'not' on a guarded transition


Valkendorm:
### Description

While strict mode is enabled, with the following machine definition, TypeScript reports a "Types of property 'EVENT' are incompatible." error.

```
const machine = createMachine(
  {
    id: 'b',
    types: {} as {
      events: { type: 'EVENT' };
    },
    on: {
      // < type error
      EVENT: {
        target: '#b',
        guard: not('falsy'),
      },
    },
  },
  {
    guards: {
      falsy: () => false,
    },
  }
);
```

### Expected result

No type error.

### Actual result

```
Type '{ EVENT: { target: string; guard: BooleanGuardDefinition<MachineContext, EventObject>; }; }' is not assignable to type 'TransitionsConfig<MachineContext, { type: "EVENT"; }> | undefined'.
  Types of property 'EVENT' are incompatible.
    Type '{ target: string; guard: BooleanGuardDefinition<MachineContext, EventObject>; }' is not assignable to type 'TransitionConfigOrTarget<MachineContext, { type: "EVENT"; }, { type: "EVENT"; }>'.
      Types of property 'guard' are incompatible.
        Type 'BooleanGuardDefinition<MachineContext, EventObject>' is not assignable to type 'GuardConfig<MachineContext, { type: "EVENT"; }> | undefined'.
          Type 'BooleanGuardDefinition<MachineContext, EventObject>' is not assignable to type 'DefaultGuardObject<MachineContext, { type: "EVENT"; }>'.
            Types of property 'children' are incompatible.
              Type 'GuardDefinition<MachineContext, EventObject>[] | undefined' is not assignable to type 'GuardObject<MachineContext, { type: "EVENT"; }>[] | undefined'.
                Type 'GuardDefinition<MachineContext, EventObject>[]' is not assignable to type 'GuardObject<MachineContext, { type: "EVENT"; }>[]'.
                  Type 'GuardDefinition<MachineContext, EventObject>' is not assignable to type 'GuardObject<MachineContext, { type: "EVENT"; }>'.
                    Type 'GuardDefinition<MachineContext, EventObject>' is not assignable to type 'DefaultGuardObject<MachineContext, { type: "EVENT"; }>'.
                      Types of property 'predicate' are incompatible.
[...]

PR #4206, opened 2023-08-23
Refactor composite guards

Andarist:
fixes https://github.com/statelyai/xstate/issues/4003

### DIFF
--- a/packages/core/test/guards.test.ts
+++ b/packages/core/test/guards.test.ts
@@ -306,7 +306,7 @@ describe('guard conditions', () => {
     });
   });
 
-  it('should allow a matching transition', () => {
+  it.skip('should allow a matching transition', () => {
     const machine = createMachine({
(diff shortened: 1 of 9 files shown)


packages/core/test/guards.test.ts
```diff
@@ -792,7 +744,7 @@ describe('not() guard', () => {
       {
         guards: {
           greaterThan10: ({ guard }) => {
-            return guard.params.value > 10;
+            return (guard as any).params.value > 10;
```
davidkpiano, line 747: Is strong typing possible?

davidkpiano (approved):

packages/core/test/guards.test.ts
Andarist: We’ll learn tomorrow :p but ye - it should be possible to have this one typed

PR #4211, opened 2023-08-28
[v5] Add `.stateNode` for actions and guards

davidkpiano:
The `stateNode` where an action or guard originates can now be read in the action or guard's provided implementation, or inline:

```ts
  initial: 'idle',
  states: {
    idle: {
      meta: { message: 'I am idle', enabled: true },

      entry: ({ stateNode }) => {
        stateNode.message; // 'I am idle'
      },

      on: {
        someEvent: {
          guard: ({ stateNode }) => {
            return stateNode.enabled; // true
          }
        }
      }
    }
  }
});
```

### DIFF
--- a/.changeset/shiny-laws-juggle.md
+++ b/.changeset/shiny-laws-juggle.md
@@ -1,0 +1,28 @@
+---
+'xstate': minor
+---
+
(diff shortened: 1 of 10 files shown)


davidkpiano:
Reworking this for v6

PR #4217, opened 2023-08-29
Introduce a light representation of state nodes

Andarist:
This could be used a value given as `stateNode` to the user-provided executable content in https://github.com/statelyai/xstate/pull/4211

### DIFF
--- a/packages/core/src/StateNode.ts
+++ b/packages/core/src/StateNode.ts
@@ -67,6 +67,12 @@ interface StateNodeOptions<
   _machine: AnyStateMachine;
 }
 
+interface LightStateNode {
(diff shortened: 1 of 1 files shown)


packages/core/src/StateNode.ts
```diff
@@ -67,6 +66,12 @@ interface StateNodeOptions<
   _machine: AnyStateMachine;
 }
 
+interface LightStateNode {
```
davidkpiano, line 69: Can we also have `key` in there?

Andarist: yes, definitely :)

davidkpiano:
We'll give them the full thing in v6; no reason not to (and they can already get it anyway)

PR #4219, opened 2023-08-30
Refactor all test cases to use `.types` and not explicit generic type params

Andarist:


davidkpiano (approved):
Pre-approved, waiting for CI to be happy

PR #4220, opened 2023-08-30
[v5] Strongly-typed partial event descriptors

davidkpiano:
```ts
createMachine({
  types: {} as {
    events:
      | { type: 'mouse.click.up'; direction: 'up' }
      | { type: 'mouse.click.down'; direction: 'down' }
      | { type: 'mouse.move' }
      | { type: 'anotherEvent' };
  },
  on: {
    'mouse.click.*': {
      actions: ({ event }) => {
        event.type;
        // 'mouse.click.up' | 'mouse.click.down'
        event.direction;
        // 'up' | 'down'
      }
    },
    'mouse.*': {
      actions: ({ event }) => {
        event.type;
        // 'mouse.click.up' | 'mouse.click.down' | 'mouse.move'
      }
    }
  }
});
```

### DIFF
--- a/.changeset/dull-actors-suffer.md
+++ b/.changeset/dull-actors-suffer.md
@@ -1,0 +1,33 @@
+---
+'xstate': minor
+---
+
+Partial event descriptors are now type-safe:
+
+```ts
+createMachine({
+  types: {} as {
+    events:
+      | { type: 'mouse.click.up'; direction: 'up' }
+      | { type: 'mouse.click.down'; direction: 'down' }
+      | { type: 'mouse.move' }
+      | { type: 'anotherEvent' };
+  },
+  on: {
+    'mouse.click.*': {
+      actions: ({ event }) => {
+        event.type;
+        // 'mouse.click.up' | 'mouse.click.down'
+        event.direction;
+        // 'up' | 'down'
+      }
+    },
+    'mouse.*': {
+      actions: ({ event }) => {
+        event.type;
+        // 'mouse.click.up' | 'mouse.click.down' | 'mouse.move'
+      }
+    }
+  }
+});
+```
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -472,26 +472,39 @@ export type TransitionsConfig<
   TAction extends ParameterizedObject,
   TGuard extends ParameterizedObject,
   TDelay extends string
-> = {
-  // TODO: this doesn't support partial descriptors
-  [K in TEvent['type'] | '*']?: K extends '*'
-    ? TransitionConfigOrTarget<
-        TContext,
-        TEvent,
-        TEvent,
-        TAction,
-        TGuard,
-        TDelay
-      >
-    : TransitionConfigOrTarget<
-        TContext,
-        ExtractEvent<TEvent, K>,
-        TEvent,
-        TAction,
-        TGuard,
-        TDelay
-      >;
-};
+> =
+  | {
+      [K in
+        | TEvent['type']
+        | PartialEventType<TEvent['type']>
+        | '*']?: K extends '*'
+        ? TransitionConfigOrTarget<
+            TContext,
+            TEvent,
+            TEvent,
+            TAction,
+            TGuard,
+            TDelay
+          >
+        : TransitionConfigOrTarget<
+            TContext,
+            ExtractEvent<TEvent, K>,
+            TEvent,
+            TAction,
+            TGuard,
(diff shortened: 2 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -714,6 +727,20 @@ export interface StateNodeConfig<
...
+        [Z in TEvent['type']]: Z extends PartialEventTypeMatcher<K>
+          ? TEvent & { type: Z }
+          : never;
+      }>,
+      TEvent,
+      TAction,
+      TGuard,
+      TDelay
+    >;
+  };
```
Andarist, line 743:
```suggestion
```

packages/core/src/types.ts
```diff
@@ -472,26 +472,31 @@ export type TransitionsConfig<
...
         TGuard,
         TDelay
       >;
-};
+    };
+
+type PartialEventType<TEventType extends string> =
+  TEventType extends `${infer TEventPrefix}.${infer TEventSuffix}`
+    ? `${TEventPrefix}.*` | `${TEventPrefix}.${PartialEventType<TEventSuffix>}`
+    : never;
```
Andarist, line 493: A version of this that would leverage tail-recursion elimination could look smth like this:
```suggestion
type _PartialEventType<
  TEventType extends string,
  Consumed extends string = "",
  Acc extends string = never,
> = TEventType extends `${Consumed}${infer TEventPrefix}.${string}`
  ? _PartialEventType<
      TEventType,
      `${Consumed}${TEventPrefix}.`,
      Acc | `${Consumed}${TEventPrefix}.*`
    >
  : Acc;

type PartialEventType<TEventType extends string> = _PartialEventType<TEventType>
```

It's hard to imagine somebody creating an event type with more than 50 segments though... and this definitely looks more complex 😅 (even though matching with `Consumed` is more straightforward in a sense). So I'd say that... ye, it's not worth to apply this optimization here.

packages/core/test/types.test.ts
```diff
@@ -3193,4 +3193,45 @@ describe('tags', () => {
     // @ts-expect-error
     actor.getSnapshot().hasTag('other');
   });
+
+  it('supports partial event descriptor types', () => {
```
Andarist, line 3197: I'd prefer to have separate tests for each case. It helps long-term because each case receives a unique test title so it becomes obvious what exact requirement is not met when something breaks.

At the same time, currently - all of this testing happens in a single function call. I've seen in the past situations in which one error could accidentally impact another one, some error location could get moved, or perhaps only a subset of errors was actually reported. Either way, it's safer~ to just have such things isolated in granular tests.

packages/core/src/types.ts
```diff
@@ -472,26 +472,33 @@ export type TransitionsConfig<
...
-        TEvent,
-        TEvent,
-        TAction,
-        TGuard,
-        TDelay
-      >
-    : TransitionConfigOrTarget<
+> =
+  | {
+      [K in EventDescriptor<TEvent>]?: TransitionConfigOrTarget<
```
Andarist, line 477: we should still decide if we want to allow actor-related events here (or other auto-generated events like the ones related to timers, `done.state.*` etc)

packages/core/src/State.ts
```diff
@@ -257,7 +258,7 @@ export class State<
   /**
    * The next events that will cause a transition from the current state.
    */
-  public get nextEvents(): Array<string> {
+  public get nextEvents(): Array<EventDescriptor<TEvent>> {
```
Andarist, line 261: WDYT about renaming this getter?

.changeset/dull-actors-suffer.md
```diff
@@ -0,0 +1,33 @@
...
+Partial event descriptors are now type-safe:
+
+```ts
+createMachine({
+  types: {} as {
+    events:
+      | { type: 'mouse.click.up'; direction: 'up' }
+      | { type: 'mouse.click.down'; direction: 'down' }
+      | { type: 'mouse.move' }
+      | { type: 'anotherEvent' };
```
Andarist, line 14: This would match the rest of the event names here better :p
```suggestion
      | { type: 'keypress' };
```

Andarist (approved):

packages/core/src/types.ts
davidkpiano: Eventually yes

packages/core/src/State.ts
davidkpiano: I'm fine with any name

PR #4228, opened 2023-09-04
Allow action and guard params to be dynamic

Andarist:


davidkpiano (approved):
Fantastic! LGTM but CI needs to pass

PR #4231, opened 2023-09-04
Allow optional actor in `useSelector`

davidkpiano:
This PR allows an actor to be `undefined` here, with proper typing:

```ts
const someState = useSelector(maybeActor, (snapshot) => ...);

someState; // derived state or undefined
```


with-heart:
What particular use case(s) are you envisioning here?

davidkpiano:
> What particular use case(s) are you envisioning here?

```ts
const maybeChild = useSelector(parentActor, s => s.children.child);
// const maybeCount = useSelector(maybeChild ?? createEmptyActor(), s => s?.context.count);
const maybeCount = useSelector(maybeChild, s => s?.context.count);
```

Necmttn:
Is there a similar method for a non-react environment?  
```ts
const count = actor.watch((state) => state.context.count)
``` 
something like this, maybe ? 

cc: @davidkpiano

davidkpiano:
> Is there a similar method for a non-react environment?
> 
> ```ts
> const $count = actor.watch((state) => state.context.count)
> ```
> 
> something like this, maybe ?
> 
> cc: @davidkpiano

I'd like to introduce something like this:

```ts
// PSEUDOCODE
import { fromSelector } from 'xstate';

const totalActor = fromSelector(someActor, s => {
  return s.context.items.reduce((sum, item) => sum + item.price, 0)
});

totalActor.subscribe(s => s.context /* = the total */)
```

packages/xstate-react/src/useSelector.ts
```diff
@@ -9,17 +10,32 @@ function defaultCompare<T>(a: T, b: T) {
...
+  actor: TActor | undefined,
+  selector: (emitted: SnapshotFrom<TActor> | undefined) => T,
+  compare?: (a: T, b: T) => boolean
+): T;
+export function useSelector<TActor extends ActorRef<any, any> | undefined, T>(
+  actor: TActor | undefined,
+  selector: (emitted: SnapshotFrom<TActor> | undefined) => T,
   compare: (a: T, b: T) => boolean = defaultCompare
 ): T {
+  const resolvedActor = actor ?? createEmptyActor();
```
Andarist, line 25: The types mention that `selector` might receive `undefined` but since this is implemented using `createEmptyActor` it actually can receive `Snapshot<undefined>`. I think it's going to be easier for people to handle actual `undefined` so I'm going to push out the "fix" for it in a moment

packages/xstate-react/src/useSelector.ts
```diff
@@ -2,24 +2,41 @@ import { useCallback } from 'react';
...
 
+type SyncExternalStoreSubscribe = Parameters<
+  typeof useSyncExternalStoreWithSelector
+>[0];
+
 function defaultCompare<T>(a: T, b: T) {
   return a === b;
 }
 
 export function useSelector<TActor extends ActorRef<any, any>, T>(
```
Andarist, line 13: a non-overload version could look smth like this:
```diff
diff --git a/packages/xstate-react/src/createActorContext.ts b/packages/xstate-react/src/createActorContext.ts
index aa4668aa9..d217c2f6d 100644
--- a/packages/xstate-react/src/createActorContext.ts
+++ b/packages/xstate-react/src/createActorContext.ts
@@ -7,7 +7,8 @@ import {
   SnapshotFrom,
   ActorOptions,
   StateMachine,
-  AnyActorLogic
+  AnyActorLogic,
+  ActorRef
 } from 'xstate';
 
 export function createActorContext<TLogic extends AnyActorLogic>(
@@ -84,7 +85,7 @@ export function createActorContext<TLogic extends AnyActorLogic>(
     compare?: (a: T, b: T) => boolean
   ): T {
     const actor = useContext();
-    return useSelectorUnbound(actor, selector, compare);
+    return useSelectorUnbound(actor as ActorRef<any, any>, selector, compare);
   }
 
   return {
diff --git a/packages/xstate-react/src/useSelector.ts b/packages/xstate-react/src/useSelector.ts
index f2647d19f..96de665a2 100644
--- a/packages/xstate-react/src/useSelector.ts
+++ b/packages/xstate-react/src/useSelector.ts
@@ -10,19 +10,13 @@ function defaultCompare<T>(a: T, b: T) {
   return a === b;
 }
 
-export function useSelector<TActor extends ActorRef<any, any>, T>(
+export function useSelector<TActor extends ActorRef<any, any> | undefined, T>(
   actor: TActor,
-  selector: (emitted: SnapshotFrom<TActor>) => T,
-  compare?: (a: T, b: T) => boolean
-): T;
-export function useSelector<TActor extends ActorRef<any, any>, T>(
-  actor: TActor | undefined,
-  selector: (emitted: SnapshotFrom<TActor> | undefined) => T,
-  compare?: (a: T, b: T) => boolean
-): T;
-export function useSelector<TActor extends ActorRef<any, any>, T>(
-  actor: TActor | undefined,
-  selector: (emitted: SnapshotFrom<TActor> | undefined) => T,
+  selector: (
+    emitted: TActor extends ActorRef<any, any>
+      ? SnapshotFrom<TActor>
+      : undefined
+  ) => T,
   compare: (a: T, b: T) => boolean = defaultCompare
 ): T {
   const subscribe: SyncExternalStoreSubscribe = useCallback(
```

.changeset/beige-donuts-grab.md
Andarist, line 1: If we are changing this in `@xstate/react` then we should probably do a similar change in `@xstate/vue` and `@xstate/svelte`.

Andarist (approved):

.changeset/beige-donuts-grab.md
```diff
@@ -0,0 +1,14 @@
...
+'@xstate/react': minor
+---
+
+The `actor` passed to `useSelector(actor, selector)` is now allowed to be `undefined` for an actor that may not exist yet. For actors that may be `undefined`, the `snapshot` provided to the `selector` function can also be `undefined`:
+
+```ts
+const count = useSelector(maybeActor, (state) => {
+  // `state` may be undefined
+  return state?.context.count;
+});
```
Andarist, line 11:
```suggestion
const count = useSelector(maybeActor, (snapshot) => {
  // `snapshot` may be undefined
  return snapshot?.context.count;
});
```

packages/xstate-react/src/useSelector.ts
davidkpiano: I feel like the overload is more for helping devs; it's better to see a simpler type when they pass in an `actorRef` that won't be undefined.

.changeset/beige-donuts-grab.md
davidkpiano: I added it for `@xstate/vue` but I'm not familiar enough with Svelte to get it working there.

packages/xstate-react/src/useSelector.ts
Andarist: Yeah, that's why I didn't commit it. I just put it here for future reference. I don't think the more complex type is that big of an issue since TS displays resolved types~ with the inference result already substituted: [TS playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMBGAHgBUA+AClgHMAueSgbQF0AaeRgJhbsOAShYA3HFmABuAFChIsBCnTY8BYmSp1GgzvAA+8NKERZUIYD34sTIMxeCj4EqXNmESpWm1Ij3CtBwSGiYuPieGHxU8CAAHoSowETGqADWqDgA7qj6RnYOlnSy8LwwzKxcJWUMAqmm5paxCSBJKZTwAPz19o3A8IJVzq4ysh6afD5+wnJAA). One could still see those in the source code though. Or when hovering over a referenced (but not called!) function, the conditional type version arguably displays better at that point. It might be slightly longer but at least it displays the whole thing instead of the first signature and the info that there is more there (`function test1<T>(arg: T[], arg2: T[]): void (+1 overload)`)

I'm still somewhat struggling when it comes to picking overloads over conditional types in situations like this. Overloads often suffer from cryptic errors but those probably happen in more complex scenarios. When it comes to adding just optionality to some arguments etc then overloads are quite fine from that PoV. I don't like how overloads don't quite work with some TS features (like conditional types, including "builtin" ones like `ReturnType` or `Parameters` but tbh they just don't work well with generic functions so 🤷‍♂️ ).

packages/xstate-vue/src/useSelector.ts
```diff
@@ -26,15 +38,15 @@ export function useSelector<TActor extends ActorRef<any, any>, T>(
   watch(
     actorRefRef,
     (newActor, _, onCleanup) => {
-      selected.value = selector(newActor.getSnapshot());
-      const { unsubscribe } = newActor.subscribe({
+      selected.value = selector(newActor?.getSnapshot() ?? undefined);
```
Andarist, line 41:
```suggestion
      selected.value = selector(newActor?.getSnapshot());
```

packages/xstate-vue/src/useSelector.ts
```diff
@@ -12,10 +12,22 @@ const noop = () => {
...
+export function useSelector<TActor extends ActorRef<any, any>, T>(
+  actor: (TActor | undefined) | Ref<TActor | undefined>,
+  selector: (snapshot: SnapshotFrom<TActor> | undefined) => T,
+  compare: (a: T | undefined, b: T) => boolean = defaultCompare
+): Ref<T | undefined> {
   const actorRefRef = isRef(actor) ? actor : shallowRef(actor);
-  const selected = shallowRef(selector(actorRefRef.value.getSnapshot()));
+  const selected = shallowRef(
+    actorRefRef.value ? selector(actorRefRef.value.getSnapshot()) : undefined
+  );
```
Andarist, line 30: This is not correct because we want `selector` to be called with `undefined` if the user doesn't pass `actor` in. Otherwise, you won't get `T` back.

This kinda gets patched immediately by the `watch` call that follows this since it uses `immediate: true`. I'm not sure if `watch` callbacks would be called server-side though - so it's better to fix this here to not rely on this implicit "fixup"

Andarist: I just pushed out a fix for this

.changeset/beige-donuts-grab.md
```diff
@@ -0,0 +1,14 @@
+---
+'@xstate/react': minor
```
Andarist, line 2:
```suggestion
'@xstate/react': minor
'@xstate/vue': minor
```

.changeset/beige-donuts-grab.md
Andarist: > but I'm not familiar enough with Svelte to get it working there.

Yeah, me neither. I don't think we even support `actorRef` changes in its `useSelector`. Let's leave it out for now then

packages/xstate-react/src/useSelector.ts
davidkpiano: I removed the overloads

PR #4233, opened 2023-09-04
Remove getMachineShortestPaths and getMachineSimplePaths

davidkpiano:
Remove `getMachineShortestPaths` and `getMachineSimplePaths`

```diff
import {
- getMachineShortestPaths,
+ getShortestPaths,
- getMachineSimplePaths,
+ getSimplePaths
} from '@xstate/graph';

-const paths = getMachineShortestPaths(machine);
+const paths = getShortestPaths(machine);

-const paths = getMachineSimplePaths(machine);
+const paths = getSimplePaths(machine);
```

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,7 +7,6 @@
   "baseBranch": "main",
   "ignore": [
     "@xstate/analytics",
-    "@xstate/graph",
     "@xstate/immer",
     "@xstate/inspect",
     "@xstate/solid",
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,6 +7,7 @@
   "baseBranch": "main",
   "ignore": [
     "@xstate/analytics",
+    "@xstate/graph",
     "@xstate/immer",
     "@xstate/inspect",
     "@xstate/solid",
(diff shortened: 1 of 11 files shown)


.changeset/config.json
Andarist, line 1: shouldn't `@xstate/test` be removed from the ignored list here too? but either way - let's keep changes to this config for a separate, follow up PR. Releasing those isn't directly related to the removal on which this PR is focused on.

packages/xstate-graph/src/graph.ts
```diff
@@ -169,13 +171,30 @@ export interface AdjacencyMap<TState, TEvent> {
...
+  TEvent extends EventFromLogic<TLogic>
+>(
+  logic: AnyActorLogic,
   traversalOptions?: TraversalOptions<TState, TEvent>,
   defaultOptions?: TraversalOptions<TState, TEvent>
 ): TraversalConfig<TState, TEvent> {
+  const resolvedDefaultOptions =
+    defaultOptions ??
+    (isMachineLogic(logic)
+      ? (createDefaultMachineOptions(
```
Andarist, line 190: On the one hand, this will be easier to use. On the other hand, it feels like those abstractions don't line up nicely if we need such special cases under the hood.

Issue #4218, opened 2023-08-30
Bug: Unexpected re-entry with an internal transition in v5


rlaffers:
### Description

An internal transition specified with `reenter: false` still caused the entry action to be executed.

```javascript
const machine = createMachine({
	id: 'root',
	initial: 'Ready',
	states: {
		Ready: {
			entry: () => console.log('enter Ready'),
	    	on: {
	        	TOGGLE: {
					target: "#root.Ready",
					reenter: false, // 🪲 despite this, the Ready state will be re-entered
		        }
			}
		}
	}
})
```

### Expected result

When the machine receives a TOGGLE event, the "Ready" state should not be reentered. There should be a single log "enter Ready".

### Actual result

The Ready state is re-entered even when the transition is explicitly marked as internal with `reenter: false`.
"enter Ready" is logged twice (once on xstate.init, the second time on TOGGLE).

### Reproduction

https://codesandbox.io/s/bitter-star-63z86x

### Additional context

For comparison, XState v4 did not execute the entry action in this exact scenario. Compare with:
https://codesandbox.io/s/focused-monad-4f3v9d

Andarist:
If we treat `reenter` as just a renamed `internal` then the current behavior is correct (according to SCMXL):
> The behavior of transitions with 'type' of "internal" is identical, except in the case of a transition whose source state is a compound state and whose target(s) is a descendant of the source. In such a case, an internal transition will not exit and re-enter its source state, while an external one will, as shown in the example below.

Since the target here is not a descendant of the source (it is the source itself), the transition reenters the source.

It is quite confusing though - especially with the name that we have chosen for this transition property (`reenter`). So I think that it would be good to diverge from SCXML here.

rlaffers:
Hmm, your call guys. On hand, sticking to the standard should be quite important. On the other hand, this is indeed confusing. Also, changing this behavior makes migration to v5 more difficult.


It wouldn't be the only diversion by the way. The spec says:
>  The behavior of a transition with 'type' of "external" (the default) ...

Whereas, xstate v5 defaults to internal transitions (reenter:false).

rlaffers:
Well, the spec later says:
> Note that this is different from a <transition> whose 'target' is its source state. In the latter case, the state is exited and reentered, triggering execution of its <onentry> and <onexit> executable content.

It seems that the spec explicitly says the state should be re-entered in this case.

Andarist:
> It wouldn't be the only diversion by the way.

[...]

PR #4234, opened 2023-09-05
Always avoid source reentrancy with `reenter: false` (the default)

Andarist:
closes https://github.com/statelyai/xstate/issues/4218


packages/core/src/stateUtils.ts
```diff
@@ -1153,10 +1164,8 @@ function enterStates(
     actions.push(...stateNodeToEnter.entry);
 
     if (statesForDefaultEntry.has(stateNodeToEnter)) {
-      for (const stateNode of statesForDefaultEntry) {
-        const initialActions = stateNode.initial!.actions;
-        actions.push(...initialActions);
-      }
```
Andarist, line 1159: this was quite wrong, I was surprised that the test suite passed with it - I have a strong feeling that we might be missing some test case, I'll try to figure it out tomorrow

davidkpiano (approved):

PR #4238, opened 2023-09-06
[v5] Improve paths in `@xstate/graph`

davidkpiano:
This PR changes the way that paths are represented.

CleanShot 2023-09-06 at 00 42 32

Instead of:
```
state D
steps:
  state A + next event
  state B + next event
  state C + next event

```

The _previous_ event is used:

```
state D
steps:
  init event + state A
  prev event + state B
  prev event + state C
  prev event + state D
```

This has a couple of benefits:

- Structure of `steps` contains all of the path information
- More declarative: "this state was reached from this event" instead of "from this state, do this event next"
- Input data can be represented as part of init event (very important!)

The algorithms need to eventually be changed so that the previous event is represented instead of the next event, so we can eventually remove `alterPath(path)`, which is an internal workaround.

Motivating example: https://github.com/statelyai/xstate/pull/4239

### DIFF
--- a/packages/xstate-graph/src/alterPath.ts
+++ b/packages/xstate-graph/src/alterPath.ts
@@ -1,5 +1,6 @@
 import { StatePath } from './types';
 
+// TODO: rewrite parts of the algorithm leading to this to make this function obsolete
 export function alterPath<T extends StatePath<any, any>>(path: T): T {
   let steps: T['steps'] = [];
 
--- a/packages/xstate-graph/src/alterPath.ts
+++ b/packages/xstate-graph/src/alterPath.ts
@@ -1,0 +1,31 @@
+import { StatePath } from './types';
+
+export function alterPath<T extends StatePath<any, any>>(path: T): T {
+  let steps: T['steps'] = [];
+
+  if (!path.steps.length) {
+    steps = [
+      {
+        state: path.state,
+        event: { type: 'xstate.init' } as any
+      }
+    ];
+  } else {
+    for (let i = 0; i < path.steps.length; i++) {
+      const step = path.steps[i];
+
+      steps.push({
+        state: step.state,
+        event: i === 0 ? { type: 'xstate.init' } : path.steps[i - 1].event
+      });
+    }
+    steps.push({
+      state: path.state,
+      event: path.steps[path.steps.length - 1].event
+    });
+  }
+  return {
+    ...path,
+    steps
+  };
+}
--- a/packages/xstate-graph/src/shortestPaths.ts
+++ b/packages/xstate-graph/src/shortestPaths.ts
@@ -17,6 +17,7 @@ import {
 } from './types';
 import { resolveTraversalOptions, createDefaultMachineOptions } from './graph';
 import { getAdjacencyMap } from './adjacency';
+import { alterPath } from './alterPath';
 
 export function getShortestPaths<
   TLogic extends AnyActorLogic,
@@ -103,7 +104,6 @@ export function getShortestPaths<
 
   weightMap.forEach(
     ({ weight, state: fromState, event: fromEvent }, stateSerial) => {
-
       const state = stateMap.get(stateSerial)!;
       const steps = !fromState
         ? []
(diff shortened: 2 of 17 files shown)


packages/xstate-test/src/TestModel.ts
```diff
@@ -283,28 +283,21 @@ export class TestModel<TState, TEvent extends EventObject> {
         testPathResult.steps.push(testStepResult);
 
         try {
-          await this.testState(params, step.state, options);
+          await this.testTransition(params, step);
```
Andarist, line 286: would this somehow allow for testing `xstate.init`? do we imagine it being useful?

packages/xstate-test/src/TestModel.ts
```diff
@@ -283,28 +283,21 @@ export class TestModel<TState, TEvent extends EventObject> {
...
+          await this.testState(params, step.state, options);
         } catch (err: any) {
-          testStepResult.event.error = err;
+          testStepResult.state.error = err;
 
           throw err;
         }
       }
-
-      try {
```
Andarist, line 302: shouldn't similar changes be made to `testPathSync`? those are 2 sibling methods that have almost the same implementation

packages/xstate-graph/src/shortestPaths.ts
```diff
@@ -133,8 +132,40 @@ export function getShortestPaths<
...
+    return paths
+      .filter((path) => resolvedOptions.toState!(path.state))
+      .map(alterPath);
   }
 
-  return paths;
+  return paths.map(alterPath);
+}
+
+export function alterPath<T extends StatePath<any, any>>(path: T): T {
```
Andarist, line 143: I think that this should be moved to a different file since this is used by both `shortestPaths` and `simplestPaths`

packages/xstate-graph/src/shortestPaths.ts
```diff
@@ -103,28 +103,27 @@ export function getShortestPaths<
 
   weightMap.forEach(
     ({ weight, state: fromState, event: fromEvent }, stateSerial) => {
+      weightMap;
```
Andarist, line 106:
```suggestion
```

packages/xstate-graph/src/shortestPaths.ts
```diff
@@ -133,8 +132,40 @@ export function getShortestPaths<
...
   }
 
-  return paths;
+  return paths.map(alterPath);
+}
+
+export function alterPath<T extends StatePath<any, any>>(path: T): T {
+  let steps: T['steps'] = [];
+
+  if (!path.steps.length) {
```
Andarist, line 146: it feels off that we need to alter the steps/paths after the fact, couldn't all of this be just made upfront? in other words: why the internal shape of this is different than the public one that we return to the user?

packages/xstate-test/src/TestModel.ts
davidkpiano: Yes it would, which can be very useful

Andarist: could u show an example of that?

packages/xstate-test/src/TestModel.ts
davidkpiano: Going to remove that one

davidkpiano: Will remove later, this is done

packages/xstate-graph/src/alterPath.ts
```diff
@@ -0,0 +1,31 @@
+import { StatePath } from './types';
+
+export function alterPath<T extends StatePath<any, any>>(path: T): T {
```
Andarist, line 3:
```suggestion
// TODO: rewrite parts of the algorithm leading to this to make this function obsolete
export function alterPath<T extends StatePath<any, any>>(path: T): T {
```

packages/xstate-test/src/TestModel.ts
```diff
@@ -206,6 +206,7 @@ export class TestModel<TState, TEvent extends EventObject> {
     return adjList;
   }
 
+  // TODO: remove this
```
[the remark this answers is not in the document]
davidkpiano: Removed comment for now

packages/xstate-graph/src/shortestPaths.ts
davidkpiano: Comment added for this (thanks)

Andarist (approved):

PR #4240, opened 2023-09-06
[v5] Alias `useMachine` in `@xstate/react`

davidkpiano:
This PR makes it so that `useMachine` (very commonly used hook) doesn't appear as strikethrough in code (technically aliased, not deprecated)

### DIFF
--- a/packages/xstate-react/src/useMachine.ts
+++ b/packages/xstate-react/src/useMachine.ts
@@ -9,8 +9,7 @@ import {
 import { useActor } from './useActor.ts';
 
 /**
- *
- * @deprecated Use `useActor(...)` instead.
+ * @alias useActor
  */
 export function useMachine<TMachine extends AnyStateMachine>(
   machine: AreAllImplementationsAssumedToBeProvided<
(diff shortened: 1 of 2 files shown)


packages/xstate-react/src/useMachine.ts
```diff
@@ -9,8 +9,7 @@ import {
 import { useActor } from './useActor.ts';
 
 /**
- *
- * @deprecated Use `useActor(...)` instead.
+ * @alias useActor
```
Andarist, line 12: what kind of an effect on the overall in-IDE experience does it have?

davidkpiano: Right now, it looks like `useMachine` is invalid because it is crossed out (~~`useMachine`~~) which can cause a lot of confusion because it's by far the most popular hook in `@xstate/react`.

If we treat it as an alias, we prevent confusion.

davidkpiano:
Merging this in as `useMachine` actually serves an important purpose for typegen (not present on `useActor`)

Issue #809, opened 2019-11-20
[Feature] Enforce/describe State Meta Data with TS


semopz:
## Bug or feature request?
Feature
## Description:
Currently metadata is of any/unknown type but in the case of actors, I need to be able to lock down what kind of metadata each event can or should have. I'm not sure if this is wading too much into context (extended state) territory but if states can be configured to only accept certain events, they should also be able to emit only a certain type of metadata.

## (Feature) Potential implementation:
This is what the API might look like
```typescript
type Locales = 'en-US' | 'en-GB' | 'fr-CA';
enum EventIntendedFor {
  NoOneOrSilent,
  User,
  Console,
  EventLog,
  System,
}

type Translations = { [code in Locales]: string };
type EventInfo = {
  for: EventIntendedFor,
  message: Translations
};

interface MachineMeta {
  // Tells the app what the machine state is in the local language (if the app wants to display that info)
  strings: {
    label: Translations;
    desc: Translations;
  },
  info?: EventInfo;
}

interface Color extends MachineMeta {
  rgb: number;
  // Different beeps and chirps or instructions
  audio?: number[];
};


interface LightStateSchema {
  states: {
    green: {
      meta: Color
    };
    yellow: {
      meta: Color
    };
    red: {
      meta: Color
      states: {
        walk: {
          meta?: MachineMeta | any
        };
        wait: {
          meta?: MachineMeta | any
        };
        stop: {
          meta?: MachineMeta | any
        };
      };
    };
  };
}
```
It shouldn't be a breaking change as omitting the meta definition should just imply meta can be `any`.

Leaving the flexibility in makes the definition a bit verbose (maybe I don't care what the nested states have as metadata but if I decide to put it in, I'd still like some intellisense).  
  
Would be nice if you could define a meta inside the `states` definition so that all nested states inherit the meta definition. But would that also apply to the nested states within (might break builtin states like onDone)?

## Link to reproduction or proof-of-concept:

Sorry, not sure how to implement this

Allaoua9:
It would be nice to have this :+1:  !

nstadigs:
I have a use case where the UI elements and the states of my physical devices in my automated home is driven by the metadata of the states of an xstate machine. Being able to enforce the type of the meta property would really improve my workflow!

davidkpiano:
[...]

PR #4241, opened 2023-09-06
[v5] Strongly type state and transition `meta`

davidkpiano:
You can now strongly type `meta` on state nodes and transitions via `stateMeta` and `transitionMeta` respectively:

```ts

  types: {
    stateMeta: {} as { title: string },
    transitionMeta: {} as { button: 'primary' | 'secondary' }
  },
  initial: 'idle',
  states: {
    idle: {
      // Strongly typed from `stateMeta`
      meta: { title: 'Idle' },
      on: {
        CLICK: {
          target: 'loading',
          // Strongly typed from `transitionMeta`
          meta: { button: 'primary' }
        }
      }
    }
  }
});
```

Fixes https://github.com/statelyai/xstate/issues/809

---

Motivation: strongly-typing things like this: https://github.com/statelyai/xstate/pull/4239/files#diff-a43cf6af3000f32addb10c6176a04f8a5066bc04358b4805726165c65d8565d3R64

### DIFF
--- a/.changeset/ten-rats-worry.md
+++ b/.changeset/ten-rats-worry.md
@@ -1,0 +1,30 @@
+---
+'xstate': minor
+---
+
+You can now strongly type `meta` on state nodes and transitions via `stateMeta` and `transitionMeta` respectively:
+
+```ts
+import { createMachine } from 'xstate';
+
+const machine = createMachine({
+  types: {
+    stateMeta: {} as { title: string },
+    transitionMeta: {} as { button: 'primary' | 'secondary' }
+  },
+  initial: 'idle',
+  states: {
+    idle: {
+      // Strongly typed from `stateMeta`
+      meta: { title: 'Idle' },
+      on: {
+        CLICK: {
+          target: 'loading',
+          // Strongly typed from `transitionMeta`
+          meta: { button: 'primary' }
+        }
+      }
+    }
+  }
+});
+```
--- a/packages/core/src/Machine.ts
+++ b/packages/core/src/Machine.ts
@@ -28,7 +28,7 @@ export function createMachine<
   TInput = any,
   TOutput = NonReducibleUnknown,
   TStateMeta extends MetaObject = MetaObject,
-  TEventMeta extends MetaObject = MetaObject,
+  TTransitionMeta extends MetaObject = MetaObject,
   TTypesMeta extends TypegenConstraint = TypegenDisabled
 >(
   config: MachineConfig<
@@ -42,7 +42,7 @@ export function createMachine<
     TInput,
     TOutput,
     TStateMeta,
-    TEventMeta,
+    TTransitionMeta,
     TTypesMeta
   >,
   implementations?: InternalMachineImplementations<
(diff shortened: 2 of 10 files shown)


.changeset/ten-rats-worry.md
```diff
@@ -0,0 +1,30 @@
...
+
+You can now strongly type `meta` on state nodes and transitions via `stateMeta` and `transitionMeta` respectively:
+
+```ts
+import { createMachine } from 'xstate';
+
+const machine = createMachine({
+  types: {
+    stateMeta: {} as { title: string },
+    transitionMeta: {} as { button: 'primary' | 'secondary' }
```
Andarist, line 13: do we need this if it's not used by our runtime *anyhow*? One can only get access to this by introspecting the machine object - which is far from the most common use case (introducing a whole new generic for this purpose feels to me like a huge huge overcomplication).

On top of that, currently this type doesn't even flow to things like:
```ts
machine.getTransitionData(machine.getInitialState({} as any), { type: '' })[0].meta
```

.changeset/ten-rats-worry.md
```diff
@@ -0,0 +1,30 @@
...
+---
+
+You can now strongly type `meta` on state nodes and transitions via `stateMeta` and `transitionMeta` respectively:
+
+```ts
+import { createMachine } from 'xstate';
+
+const machine = createMachine({
+  types: {
+    stateMeta: {} as { title: string },
```
Andarist, line 12: I always imagined this to be a per-state information and not a global one. Here, you are going to restrict all state metas to have the same shape. This definitely simplifies some things but it also feels like it cripples the utility of this type.

I think that before introducing any changes to this it would be great to establish where this information pops up in our runtime and types (I wonder if we could accurately list all of those places without actually rechecking the current implementation 😉 ). Having a list like that would allow us to assess the idea of this being a single shape for all states.

davidkpiano: I think per-state is more of a lesser user-case than a common use-case. And even so, we can use (discriminated) unions for the `stateMeta` value.

Please see here for a motivating example: https://github.com/statelyai/xstate/pull/4239/files#diff-a43cf6af3000f32addb10c6176a04f8a5066bc04358b4805726165c65d8565d3R64

.changeset/ten-rats-worry.md
Andarist: I actually think that the `.configuration` should be removed from the user-facing `state` 😅 If there is a need for introspection then a dedicated API should be introduced for that (even if that would be a getter~ on the `state` or smth). The current `state`'s shape is too fat and it actually creates problems for `@xstate/solid`

.changeset/ten-rats-worry.md
Andarist: This is still a very advanced use case - at that point, I think it should be rather OK to require this metadata to be casted~/asserted at some point in the user's code. The current implementation doesn't even *require* those metas to be provided for each state/transition (and that would complicate types even more so I would prefer not doing that). So there is no strict guarantee that each state/transition will have this information and the user is left alone when it comes to remembering that they need to add this. If they need to add this then requiring smth like `meta: {} satisfies MyShape` doesn't sound like a big ask to me.

I understand the appeal of such generated UIs etc but I'm just not sure if this narrow use case should motivate extra generics here. The `statesMeta` is at least accessible on `state.meta` today (although in an awkward way) so that could be useful at typing this. The `transitionMeta` isn't as publicly available though.

davidkpiano: See #809 for more motivation

Andarist:
superseded by https://github.com/statelyai/xstate/pull/4863

PR #4245, opened 2023-09-06
Remove `StateNode['target']` getter

Andarist:


davidkpiano (approved):
Huh, I wonder why I added that in the first place

PR #4255, opened 2023-09-07
Move event utils to a separate file

Andarist:


packages/core/src/eventUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @param delayRef The delay in milliseconds
+ * @param id The state node ID where this event is handled
+ */
+export function after(delayRef: number | string, id?: string) {
+  const idSuffix = id ? `#${id}` : '';
+  return `${constantPrefixes.after}(${delayRef})${idSuffix}`;
+}
+
+export function doneStateEventType<T extends string = string>(id: T) {
```
Andarist, line 21: let's bikeshed how those helper functions should be named (how to call the utils responsible for creating the event type and how do we call utils responsible for creating the event itself). We also need to decide which ones should be publicly exported (if any)

packages/core/src/constantPrefixes.ts
```diff
@@ -1,7 +1,6 @@
...
-export const doneState = ConstantPrefix.DoneState;
-export const errorExecution = ConstantPrefix.ErrorExecution;
-export const errorPlatform = ConstantPrefix.ErrorPlatform;
-export const error = ConstantPrefix.ErrorCustom;
+export const after = 'xstate.after';
+export const doneInvoke = 'done.invoke';
+export const doneState = 'done.state';
+export const error = 'xstate.error';
+export const errorExecution = 'error.execution';
+export const errorPlatform = 'error.platform';
```
Andarist, line 6: I think I would take a step further and just remove this file altogether. Inlining those where needed should be fine - it's not like we are using this all over the place

packages/core/src/eventUtils.ts
davidkpiano: By default, let's not export any

davidkpiano (approved):

packages/core/src/constantPrefixes.ts
davidkpiano: Agree

packages/core/src/eventUtils.ts
Andarist: I pushed out extra changes, could you re-review?

davidkpiano (approved):

PR #4256, opened 2023-09-07
Add a brand to `Function` to help distinguish resolved values from resolvers in the types

Andarist:
I'm honestly not sure if I want to land this. It feels gross but helps to avoid legitimate errors 🤷‍♂️

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -197,6 +197,14 @@ type ConditionalRequired<T, Condition extends boolean> = Condition extends true
   ? Required<T>
   : T;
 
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -206,7 +214,7 @@ export type WithDynamicParams<
       {
         type: T['type'];
         params?:
-          | T['params']
+          | NotAFunction<T['params']>
```
Andarist, line 217: if we decide to land this PR then this helper should be added to many more places

davidkpiano:
Closing per @Andarist's request

PR #4264, opened 2023-09-11
Update API docs link

davidkpiano:
https://doc.deno.land/https://esm.sh/v131/xstate

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -306,9 +306,9 @@ module.exports = {
       }
     },
     nav: [
-      { text: 'API', link: 'https://paka.dev/npm/xstate/' },
+      { text: 'API', link: 'https://doc.deno.land/https://esm.sh/v131/xstate' },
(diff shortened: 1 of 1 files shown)


docs/.vuepress/config.js
```diff
@@ -306,9 +306,9 @@ module.exports = {
       }
     },
     nav: [
-      { text: 'API', link: 'https://paka.dev/npm/xstate/' },
+      { text: 'API', link: 'https://doc.deno.land/https://esm.sh/v131/xstate' },
```
Andarist, line 309: The paka thing feels lighter to me and less daunting. It also seems to favor JSDoc a little bit more (often you can find the JSDoc comments earlier on the displayed page) and displays some information from base types as well when interfaces extend other interfaces.

davidkpiano: The paka thing doesn't work though for XState beta. Wanted to update it here for consistency but I guess we can keep it.

PR #4265, opened 2023-09-11
[v5] Remove FSM-related code

davidkpiano:


andy-zhou:
Hey, happened to see this PR. Is @xstate/fsm getting sunset?

davidkpiano:
> Happened to see this PR. Is @xstate/fsm getting sunset?

Yes; the plan is that `xstate` itself should be used most of the time, and if bundle size is a concern, perhaps a `fromFSM(...)` actor logic creator can be provided inside `xstate` itself, which would be tree-shakeable.

What is your use-case for `@xstate/fsm`? Is it just the bundle size?

andy-zhou:
No usecase here - I was just taking a look at the differences between the two. Deprecate away.

abmagil:
@davidkpiano - since you were asking about use cases. I encountered the situation where I was trying to introduce state machines to an app with an already present state management system (redux). I got pushback because of the bundle size, as you suggest. Having an easy-to-bundle-shake solution would let me keep using xstate's API in the small footprint that I actually need it for.

hysan:
@davidkpiano hi, I found this issue while looking for docs on `@xstate/fsm`. My use case is the same as @abmagil. Is there a small bundle size solution in v5, or should I be using `@xstate/fsm`? <- if I go that route, is there anything I should be worried about when using it with xstate/react?

Andarist:
V5 reduced the bundlesize of XState quite a bit - but, of course, it's still not exactly close to the size of `@xstate/fsm`. If this is very important to you then you can still use `@xstate/fsm` but then please continue using `@xstate/react@^3` and import hooks from `@xstate/react/fsm`

PR #4269, opened 2023-09-12
[v5] Export actor logic types

davidkpiano:
Actor types are now exported from `xstate`:

```ts
import {
  type CallbackActorLogic,
  type ObservableActorLogic,
  type PromiseActorLogic,
  type TransitionActorLogic
} from 'xstate';
```

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,15 +1,5 @@
 export * from './actions.ts';
-export {
-  fromCallback,
-  fromEventObservable,
-  fromObservable,
-  fromPromise,
-  fromTransition,
-  type CallbackActorLogic,
-  type ObservableActorLogic,
-  type PromiseActorLogic,
-  type TransitionActorLogic
-} from './actors/index.ts';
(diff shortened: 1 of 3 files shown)


packages/core/src/index.ts
```diff
@@ -4,7 +4,11 @@ export {
   fromEventObservable,
   fromObservable,
   fromPromise,
-  fromTransition
+  fromTransition,
+  type CallbackActorLogic,
+  type ObservableActorLogic,
+  type PromiseActorLogic,
+  type TransitionActorLogic
 } from './actors/index.ts';
```
Andarist, line 12: We can collapse this whole export into `export * from './actors/index.ts'`. Since both of those are publicly available~ (`xstate` and `xstate/actors`) and we export those `from*` factories from both, I think it's best to just always reexport here whatever is exported by that nested entrypoint

davidkpiano: Sounds good

Andarist (approved):

PR #4270, opened 2023-09-12
Put all special events under `xstate.*` namespace

Andarist:


packages/core/src/eventUtils.ts
```diff
@@ -48,7 +48,7 @@ export function createDoneInvokeEvent(
   output?: unknown
 ): DoneInvokeEventObject {
   return {
-    type: `done.invoke.${invokeId}`,
+    type: `xstate.done.invoke.${invokeId}`,
```
Andarist, line 51: I'd also like to propose an extra change to this. Since the same events are also generated for spawned actors, I think it would be best to use `xstate.done.actor.INVOKE_ID` here.

packages/core/src/eventUtils.ts
```diff
@@ -57,7 +57,7 @@ export function createErrorPlatformEvent(
   id: string,
   data?: unknown
 ): ErrorPlatformEvent {
-  return { type: `error.platform.${id}`, data };
+  return { type: `xstate.error.platform.${id}`, data };
```
Andarist, line 60: To accompany the other proposed change, I propose to rename this to `xstate.error.actor.INVOKE_ID`.

packages/core/test/invoke.test.ts
```diff
@@ -2046,9 +2046,10 @@ describe('invoke', () => {
               onError: {
                 target: 'success',
                 guard: ({ context, event }) => {
-                  expect(event.data.message).toEqual('some error');
+                  expect((event.data as any).message).toEqual('some error');
```
davidkpiano, line 2049: What's wrong here?

PR #4281, opened 2023-09-15
Remove `deferEvents` option

Andarist:
I doubt this was a popular option.


davidkpiano (approved):

davidkpiano:
Let's add a changeset for this just in case

PR #4282, opened 2023-09-15
Avoid mutating the received event

Andarist:


packages/core/src/types.ts
```diff
@@ -1841,6 +1841,8 @@ export interface ActorRef<TEvent extends EventObject, TSnapshot = any>
    */
   id: string;
   sessionId: string;
+  /** @internal */
+  _send: (event: TEvent) => void;
```
Andarist, line 1845: Is it great to have this on the `ActorRef`? not really, I wish I could figure out some good alternative to this

But I think that, at the end of the day, `ActorRef` is not something that people should implement on their own. The primary way of creating actors is to create an actor logic and pass that to `createActor`. That should ensure that `_send` is actually always there and that the `system.sendTo` can always call that.

davidkpiano (approved):

PR #4290, opened 2023-09-17
Error on incompatible state value when resolving state

davidkpiano:


packages/core/test/rehydration.test.ts
```diff
@@ -122,6 +122,37 @@ describe('rehydration', () => {
 
       expect(actual).toEqual([]);
     });
+
+    it('should handle incompatible state value (shallow)', () => {
```
Andarist, line 126: those tests should be moved to another location - we don't even attempt to rehydrate here since this is an early error so those tests don't quite belong to `rehydration.test.ts`

I would also suggest to improve those test titles. "Handle" implies to me that we somehow manage to deal with such a situation but in fact we just throw - so the simplest adjustment would be to rephrase this to smth like:
```suggestion
    it('should error on incompatible state value (shallow)', () => {
```

.changeset/olive-sheep-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+An error will now be thrown if an incompatible state value is passed to `machine.resolveStateValue(value)`.
```
Andarist, line 5:
```suggestion
An error will now be thrown if an incompatible state value is passed to `machine.resolveState({ value })`.
```

packages/core/test/rehydration.test.ts
```diff
@@ -123,6 +123,37 @@ describe('rehydration', () => {
...
+          valid: {}
+        }
+      });
+
+      expect(() => {
+        machine.resolveState({ value: 'invalid' });
+      }).toThrowError(/invalid/);
+    });
+
+    it('should handle incompatible state value (deep)', () => {
```
Andarist, line 140:
```suggestion
    it('should error on incompatible state value (deep)', () => {
```

Andarist (approved):

PR #4294, opened 2023-09-20
Rollback unpacking type params as defaults from `TConfig`

Andarist:


Andarist:
Superseded by https://github.com/statelyai/xstate/pull/4498

PR #4301, opened 2023-09-22
[v5] Persist timers

davidkpiano:


packages/core/test/scxml.test.ts
```diff
@@ -151,7 +151,7 @@ const testGroups: Record<string, string[]> = {
     // 'test183.txml', // idlocation not implemented yet
     'test185.txml',
     'test186.txml',
-    'test187.txml',
+    // 'test187.txml', // we don't automatically cancel delayed send actions - those should immediately execute
```
Andarist, line 154: this test should actually pass - we currently have a bug when it comes to id-less timers, they are not canceled when their "owner" actor gets stopped. I'll fix this as part of this PR

packages/core/test/scheduler.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    const machine = createMachine({
+      id: 't',
+      after: {
+        1000: {}
+      }
+    });
+
+    const actor = createActor(machine, {
+      scheduler: {
+        setTimeout: (_system, scheduledEvent) => {
```
Andarist, line 14: if we are gonna keep those arguments i think we should rename `setTimeout`/`cancelTimeout` options on this. They are too close to the JS timers API and it's confusing that they accept different arguments here.

Any particular reason why you have restructured it this way? could we add a test case that would show why this is useful?

packages/core/src/types.ts
```diff
@@ -1700,14 +1700,15 @@ export interface StateConfig<
   tags?: Set<string>;
   machine?: StateMachine<TContext, TEvent, any, any, any, any, any, any, any>;
   _internalQueue?: Array<TEvent>;
+  timers?: ScheduledEvent[];
```
Andarist, line 1703: should this be an array? this is not quite an ordered collection (order isn't that relevant to us) and when removing an item (when canceling an event) it would be easier to remove a key from an object than to remove an array item

packages/core/src/interpreter.ts
```diff
@@ -511,9 +518,13 @@ export class Actor<TLogic extends AnyActorLogic>
...
+    const timerId = this.system.scheduler.setTimeout(this, {
+      delay,
+      event,
+      id: Math.random().toString(),
+      startedAt: Date.now(),
+      target: to ?? this
+    });
 
     // TODO: consider the rehydration story here
     if (id) {
```
Andarist, line 530: TODO for myself: we should always keep those timer ids around - that would help us to fix test187

PR #4304, opened 2023-09-22
Tweak types in a few actor-related tests

Andarist:


packages/core/test/actor.test.ts
```diff
@@ -299,7 +307,6 @@ describe('spawning promises', () => {
           idle: {
             entry: assign({
               promiseRef: ({ spawn }) =>
-                // TODO: this is typed as AnyActorRef instead of PromiseActorRef<string>
```
Andarist, line 302: we can't infer that without `types.actors` (which I added here)

packages/core/test/actor.test.ts
```diff
@@ -352,9 +361,8 @@ describe('spawning callbacks', () => {
         idle: {
           entry: assign({
             callbackRef: ({ spawn }) =>
-              // TODO: this is typed as AnyActorRef instead of CallbackActorRef<{ type: 'START' }>
```
Andarist, line 355: this is also not possible to infer from bare `fromCallback` - `{ type: 'START' }` didn't appear anywhere in its signature and contextual types for return positions in nested calls (`assign` within `createMachine`) are sometimes not working (that's why we sometimes have to use `as const` within `raise` etc)

packages/core/test/actor.test.ts
```diff
@@ -787,77 +795,6 @@ describe('communicating with spawned actors', () => {
 
     parentService.start();
   });
-
-  it.skip('should be able to name existing actors', (done) => {
```
Andarist, line 791: I don't think this test has merit. It doesn't make sense to `spawn` an existing actor ref

packages/core/test/actor.test.ts
```diff
@@ -1229,10 +1166,9 @@ describe('actors', () => {
 
   it('should be able to spawn callback actors in (lazy) initial context', (done) => {
     const machine = createMachine({
-      types: {} as { context: { ref: CallbackActorRef<{ type: 'TEST' }> } },
+      types: {} as { context: { ref: CallbackActorRef<EventObject> } },
```
Andarist, line 1169: this type is about the **receiver**, not about the `sendBack`, and `{ type: 'TEST' }` is used by `sendBack`, so the previous annotation was wrong

davidkpiano (approved):

PR #4306, opened 2023-09-22
Move machine's `output` to top level

Andarist:


packages/core/test/types.test.ts
```diff
@@ -282,19 +258,13 @@ describe('output', () => {
         };
       },
       context: { password: 'okoń' },
```
davidkpiano, line 260: What kind of password is this?

davidkpiano (approved):

davidkpiano:
Does this break existing behavior of `output` on top-level final state appearing in `state.output`?

packages/core/test/types.test.ts
Andarist: 😂 I forgot that i sneaked in this polish easter egg here in the past

Andarist:
> Does this break existing behavior of output on top-level final state appearing in state.output?

Yes - I don't see how to maintain the backward compatibility in a sensible way here. This doesn't feel right:
```ts
createMachine({
  types: {} as { output: number },
  initial: 'foo',
  states: {
    foo: {
      type: 'final',
      output: 'bazinga'   
   }
  }
})
```

Configuring some backwards compatible logic at type level (like if `types.output` is not configured) to match what we could figure out here for the runtime world would be quite clunky

PR #4307, opened 2023-09-23
Persist callback input

davidkpiano:
This PR persists the `input` of callback, which was coming back as `undefined` when restoring callback state.

### DIFF
--- a/packages/core/test/actorLogic.test.ts
+++ b/packages/core/test/actorLogic.test.ts
@@ -489,6 +489,44 @@ describe('callback logic (fromCallback)', () => {
 
     createActor(machine).start();
   });
+
+  it('should persist the input of a callback', () => {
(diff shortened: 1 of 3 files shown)


packages/core/test/actorLogic.test.ts
```diff
@@ -489,6 +489,44 @@ describe('callback logic (fromCallback)', () => {
 
     createActor(machine).start();
   });
+
+  it('should persist the input of a callback', () => {
```
Andarist, line 493: I think it would be great to tweak this test case because it's not apparent that it tests the right thing. It passes if we remove `state` from the second `createActor` test (because then it just starts a new actor completely and the input is static).

I'll push a small change to this test case in a sec

Andarist (approved):

PR #4313, opened 2023-09-25
Add warning for missing top-level output

davidkpiano:


packages/core/src/StateMachine.ts
```diff
@@ -193,6 +194,19 @@ export class StateMachine<
 
     this.states = this.root.states; // TODO: remove!
     this.events = this.root.events;
+
+    if (isDevelopment) {
+      if (
+        Object.values(this.states).some(
+          (state) => state.type === 'final' && !!state.output
+        ) &&
+        !this.root.output
```
Andarist, line 203: I'd flip those conditions - first check the `!this.root.output` thing, then do the `Object.values` work.

It's a nit though since it doesn't matter here

packages/core/test/invoke.test.ts
```diff
@@ -53,7 +53,8 @@ const fetchMachine = createMachine({
     failure: {
       entry: sendParent({ type: 'REJECT' })
     }
-  }
+  },
+  output: ({ event }) => event.output
```
Andarist, line 57: I'm baffled by this - the tests in https://github.com/statelyai/xstate/pull/4306 pass even without it so it indicates that something is wrong.

packages/core/test/json.test.ts
```diff
@@ -88,7 +88,8 @@ describe('json', () => {
             }
           }
         }
-      }
+      },
+      output: {}
```
Andarist, line 92: This could look slightly less odd
```suggestion
      output: { result: 42 }
```

packages/core/test/invoke.test.ts
Andarist: I have no idea what that test would supposed to test 😅 I restructured it slightly and removed `output` from here completely - since the test wasn't testing it anyway.

Issue #4311, opened 2023-09-25
Bug: During the ongoing execution of the exit action, the state machine has already transitioned to the next state.


howard-57blocks:
### Description

```javascript
import {createMachine, interpret} from "@xstate/fsm";
let service;
// Edit your machine(s) here
const machine = createMachine({
    id: "machine",
    initial: "inactive",
    states: {
        inactive: {
            exit: 'inactiveExitAction',
            on: { TOGGLE: "active" }
        },
        active: {
            on: { TOGGLE: "inactive" }
        }
    }
}, {
    actions: {
        inactiveExitAction: (context,) => {
            console.log('inactiveExitAction', service.state.value)
        }
    }
});

service = interpret(machine).start()
service.send({type: 'TOGGLE'})
``` 

the output is `inactiveExitAction active`
why is not `inactiveExitAction inactive` ?

the doc says

An entry/exit action means “execute this action on any transition that enters/exits this state”. Use entry/exit actions when the action is only dependent on the state node that it’s in, and not on previous/next state nodes or events.

I am not sure if this is a bug

I am using the @xstate/fsm@2.1.0

### Expected result

the output should be `inactiveExitAction inactive`

### Actual result

 inactiveExitAction active

### Reproduction

https://codesandbox.io/s/cool-oskar-jscsc7?file=/src/index.js

davidkpiano:
This behavior is fixed in XState v5 beta (`xstate@beta`): https://codesandbox.io/s/damp-wave-4frjj9?file=/src/index.js

PR #4322, opened 2023-09-27
fix: exit actions not running before transition (#1)

mlevy-liftoff:
This PR:

- Fixes an issue in `@xstate/fsm` where exit actions run after transition and not before

### DIFF
--- a/packages/xstate-fsm/test/fsm.test.ts
+++ b/packages/xstate-fsm/test/fsm.test.ts
@@ -113,8 +113,15 @@ describe('@xstate/fsm', () => {
   it('should transition correctly', () => {
     const nextState = lightFSM.transition('green', 'TIMER');
     expect(nextState.value).toEqual('yellow');
+    expect(nextState.exitActions.map((action) => action.type)).toEqual([
+      'exitGreen'
+    ]);
+    expect(nextState.exitContext).toEqual({
+      count: 2,
+      foo: 'static++',
+      go: true
+    });
     expect(nextState.actions.map((action) => action.type)).toEqual([
-      'exitGreen',
       'g-y 1',
       'g-y 2'
     ]);
@@ -355,6 +362,30 @@ describe('interpreter', () => {
     });
 
     interpret(machine).start();
+    expect(executed).toBe(true);
+  });
+
+  it('should execute exit action', () => {
+    let executed = false;
+    let service;
+    const machine = createMachine({
+      initial: 'foo',
+      states: {
+        foo: {
+          exit: () => {
+            executed = service.state.value === 'foo';
+          },
+          on: {
+            BAR: 'bar'
+          }
+        },
+        bar: {}
+      }
+    });
+
+    service = interpret(machine);
+    service.start();
+    service.send('BAR');
     expect(executed).toBe(true);
   });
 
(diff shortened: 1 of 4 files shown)


packages/xstate-fsm/test/fsm.test.ts
```diff
@@ -358,6 +365,30 @@ describe('interpreter', () => {
...
+  it('should execute exit action', () => {
+    let executed = false;
+    let service;
+    const machine = createMachine({
+      initial: 'foo',
+      states: {
+        foo: {
+          exit: () => {
+            executed = service.state.value === 'foo';
+          },
```
Andarist, line 377: It's not exactly something we are inclined to support. You shouldn't be checking `service.state` in your actions like this. A machine is like a class whereas the service is like an instance of that class - checking the instance's state through an external reference doesn't make a lot of sense.

mlevy-liftoff: To add a little more context. [This issue](https://github.com/statelyai/xstate/issues/4311) was raised as we noticed this behaviour in the full package in addition to the `@xstate/fsm` package. We use the `@xstate/fsm` package in our projects.

The response on the issue is that it has been fixed in v5 but we couldn't see a fix for the fsm package. Hence the PR proposing a fix for it.

Andarist: I understand but how those things are executed internally is kinda of an implementation detail really. Conceptually the exit actions only have to be executed before transition actions - that's an observable effect that we care about.

mlevy-liftoff: Isn't the response to [the issue](https://github.com/statelyai/xstate/issues/4311) raised an acknowledgement that exit actions should run **before** state changes?

Andarist: The issue that you mentioned wasn't exactly "fixed" in v5. The fact that `.getSnapshot()` within an exit action still refers to the previous state is also just an implementation detail that semi-leaks to such a call. After all, we need to update the internal data structures that this reaches into *at some point* and naturally it might be observable one way or another. However, relying on those details is not an idiomatic way of using XState.

To achieve semantics parity between `xstate@beta` and `@xstate/fsm` here you'd have to execute actions within the `machine.transition` call - instead of executing them **after** that call. That would break other scenarios though, like this one:
```ts
let executed = false;
let service;
  initial: "foo",
  states: {
    foo: {
      on: {
        BAR: "bar",
      },
    },
    bar: {
      entry: () => {
        executed = service.state.value === "bar";
      },
    },
  },
});

service = interpret(machine);
service.start();
service.send("BAR");
expect(executed).toBe(true);
```

This assertion passes today but it would fail with the change that I mentioned (but it would match the equivalent `xstate@beta` code).

For those reasons, I don't think it's something that we'd like to "fix". it's just not something that you should rely on in the first place.

mlevy-liftoff: Ok, I understand. Thanks for clarifying @Andarist

Andarist:
Resolved through a discussion [here](https://github.com/statelyai/xstate/pull/4322#discussion_r1338370420)

PR #4324, opened 2023-09-27
Use TEvent['type'] for .nextEvents

mellson:
This PR changes the type of `state.nextEvents` to use `TEvent['type']`. This is to make the type of the next events compatible with the `send` methods input which use `EventFromLogic`.

### DIFF
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -10,21 +10,20 @@ import {
 } from './stateUtils.ts';
 import { TypegenDisabled, TypegenEnabled } from './typegenTypes.ts';
 import type {
-  ProvidedActor,
   ActorRefFrom,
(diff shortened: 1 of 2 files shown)


davidkpiano (approved):

packages/core/src/State.ts
```diff
@@ -257,7 +256,7 @@ export class State<
   /**
    * The next events that will cause a transition from the current state.
    */
-  public get nextEvents(): Array<EventDescriptor<TEvent>> {
+  public get nextEvents(): Array<TEvent['type']> {
```
Andarist, line 259: This change was not correct and I quickly force-pushed the previous version of `next` once this was merged in. Please share a test case demoing your problems so I can help with your exact problem.

PR #4329, opened 2023-10-01
[v5] Add `spawn()` action

davidkpiano:


packages/core/src/actions/spawn.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ *
+ * @param expr The expression function to evaluate which will be logged.
+ *  Takes in 2 arguments:
+ *  - `ctx` - the current state context
+ *  - `event` - the event that caused this action to be executed.
+ * @param label The label to give to the logged expression.
+ */
+export function spawn<
```
Andarist, line 67: q: isn't this - in a way - just an exposed `invoke` action? :p

packages/core/test/spawn.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    );
+    const actor = createActor(
+      createMachine({
+        types: {
+          actors: {} as {
+            src: 'fetchNum';
+            logic: typeof fetchNum;
+          }
+        },
+        entry: spawn('fetchNum', { id: 'child', input: 21 })
```
Andarist, line 37: I wonder if we could make it easier to spawn N actors at the same time. I was thinking about the API that would allow for a resolver to return many actors:
```ts
spawn(spawn => ({
  id1: spawn('foo', { input: 42 }),
  id2: spawn('bar', { input: 'answer' })
}))
```

Of course, I don't like using `spawn` name for both here. A nice thing about this one is that we could more easily **derive** spawned IDs in the future from types. Although, perhaps we could also do that from the `spawn` action since it returns, well, an action object - and not the actual actor. So we still have a place to keep the inferred ID at the type level. 

So basically... I'm probably just thinking about something that would allow us to spawn multiple actors at the same time without using `pure`

davidkpiano: Discussed offline, we'll stick with spawning one at a time (there's `pure` for dynamic spawns)

packages/core/src/actions/spawn.ts
Andarist: I managed to deduplicate those

Andarist (approved):

PR #4332, opened 2023-10-01
setup function may lead to unexpected behavior

sachin-nanayakkara:
In Vue , returning null from the setup function may lead to unexpected behavior, so it's best to follow the recommendation to return undefined. This change should resolve the warning message mentioned.

### DIFF
--- a/packages/xstate-vue/test/typegenTypes.test.tsx
+++ b/packages/xstate-vue/test/typegenTypes.test.tsx
@@ -48,7 +48,7 @@ describe('useMachine', () => {
       setup() {
         // @ts-expect-error
         useMachine(machine);
-        return null;
+        return undefined;
       }
     });
   });


davidkpiano:
Thank you! Can you add a quick changeset?

`yarn changeset` (patch)

packages/xstate-vue/test/typegenTypes.test.tsx
```diff
@@ -48,7 +48,7 @@ describe('useMachine', () => {
       setup() {
         // @ts-expect-error
         useMachine(machine);
-        return null;
+        return undefined;
```
Andarist, line 51:
```suggestion
        return;
```

Andarist:
> Can you add a quick changeset?

This one doesn't need a changeset - only a test file is touched here.

PR #4335, opened 2023-10-02
[v5] Composable logic

davidkpiano:
This PR fixes state machine logic so it can be composable.


packages/core/src/StateMachine.ts
```diff
@@ -182,7 +182,13 @@ export class StateMachine<
     };
     this.version = this.config.version;
     this.types = this.config.types ?? ({} as any as this['types']);
+
+    // I'm sure there's a better way of doing this...
```
davidkpiano, line 186:
```suggestion
```

PR #4343, opened 2023-10-08
[v5] `stop()` for multiple actors

davidkpiano:
You can now stop multiple actors in the `stop(...)` action creator:

```ts
  types: {
    context: {} as {
      actors: Array<AnyActorRef>;
    }
  },
  context: ({ spawn }) => {
    return {
      actors: [
        spawn(fromPromise(() => Promise.resolve('foo'))),
        spawn(fromPromise(() => Promise.resolve('bar'))),
        spawn(fromPromise(() => Promise.resolve('baz')))
      ]
    };
  },
  on: {
    stopAll: {
      // Return a single actorRef or an array of actorRefs
      actions: stop(({ context }) => context.actors)
    }
  }
});
```

Discussion:
- https://discord.com/channels/795785288994652170/1160570997209051256


Andarist:
I find it slightly weird that this would introduce a lack of symmetry between [the proposed `spawn` action](https://github.com/statelyai/xstate/pull/4329) - that action currently *doesn't* support spawning of multiple actors at the same time. This one here is also a little bit opinionated about the "collection type" of actors - it only supports arrays. It feels to me that usually actor collections are not ordered and for that reason, I'd choose an indexed object over an array to keep them around and get easier/faster access to any particular actor. 

Overall, I like the idea - but I have the mentioned concerns. I also think that it's the post-v5 item just because ideally we'd bring support for this to the Studio at the same time and I'd prefer not to work on such improvements before we publish a stable version of v5.

davidkpiano:
> I find it slightly weird that this would introduce a lack of symmetry between [the proposed `spawn` action](https://github.com/statelyai/xstate/pull/4329) - that action currently _doesn't_ support spawning of multiple actors at the same time. This one here is also a little bit opinionated about the "collection type" of actors - it only supports arrays. It feels to me that usually actor collections are not ordered and for that reason, I'd choose an indexed object over an array to keep them around and get easier/faster access to any particular actor.
> 
> Overall, I like the idea - but I have the mentioned concerns. I also think that it's the post-v5 item just because ideally we'd bring support for this to the Studio at the same time and I'd prefer not to work on such improvements before we publish a stable version of v5.

Agree, this is post-v5 release 👍

davidkpiano:
🧹 Closing as this can be done using `enqueueActions(…)`:

```ts
actions: enqueueActions(x => {
  x.context.actors.forEach(actor => x.enqueue.stop(actor));
})
```

PR #4344, opened 2023-10-09
Export inspection event types

davidkpiano:
Inspection events are now exported:

```ts
import type {
  InspectedActorEvent,
  InspectedEventEvent,
  InspectedSnapshotEvent,
  InspectionEvent
} from 'xstate';
```

### DIFF
--- a/packages/core/test/inspect.test.ts
+++ b/packages/core/test/inspect.test.ts
@@ -4,9 +4,9 @@ import {
   fromPromise,
   sendParent,
   sendTo,
-  waitFor
+  waitFor,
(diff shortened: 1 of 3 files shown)


packages/core/test/inspect.test.ts
```diff
@@ -4,9 +4,9 @@ import {
   fromPromise,
   sendParent,
   sendTo,
-  waitFor
+  waitFor,
+  InspectionEvent
```
mellson, line 8: nit: do we want these sorted?

mellson (approved):

PR #4351, opened 2023-10-12
Make updated context available to invokes

Andarist:


packages/core/src/actors/callback.ts
```diff
@@ -63,10 +63,10 @@ export function fromCallback<TEvent extends EventObject, TInput = unknown>(
   const logic: CallbackActorLogic<TEvent, TInput> = {
     config: invokeCallback,
     start: (_state, { self, system }) => {
-      system._relay(self, self, { type: XSTATE_INIT });
```
Andarist, line 66: this was *super* confusing because `XSTATE_INIT` is a special event and yet this "custom" logic was sending the same type to itself manually

packages/core/src/interpreter.ts
```diff
@@ -500,19 +498,19 @@ export class Actor<TLogic extends AnyActorLogic>
...
     id: string | undefined;
     delay: number;
     to?: AnyActorRef;
   }): void {
+    const { event, id, delay } = params;
     const timerId = this.clock.setTimeout(() => {
-      this.system._relay(this, to ?? this, event as EventFromLogic<TLogic>);
+      this.system._relay(
+        this,
+        params.to ?? this,
```
Andarist, line 511: TODO: this likely deserves a test case

packages/core/src/stateUtils.ts
```diff
@@ -1191,9 +1213,9 @@ function enterStates(
           stateNodeToEnter.output
             ? resolveOutput(
                 stateNodeToEnter.output,
-                currentState.context,
+                nextState.context, // TODO: this was likely buggy, write a test case that proved it
```
Andarist, line 1216: TODO: I need to look into this before landing this

packages/core/test/predictableExec.test.ts
```diff
@@ -424,11 +415,14 @@ describe('predictableExec', () => {
           entry: sendTo('myChild', { type: 'KNOCK_KNOCK' }),
           invoke: {
             id: 'myChild',
-            src: fromCallback(({ receive }) => {
```
Andarist, line 427: Currently, a callback actor couldn't be used here for this test case because the event sent from an entry action to an invoked actor is being received by it *before* a callback actor has a chance to setup listeners. 

It's a timing issue that we should fix but that's outside of the scope of this PR.

packages/core/test/system.test.ts
```diff
@@ -273,9 +273,14 @@ describe('system', () => {
         src: createMachine({}),
         systemId: 'test'
       },
-      entry: assign(({ system }) => {
-        expect(system!.get('test')).toBeDefined();
-      })
```
Andarist, line 278: well, this is an ability that we'd "lose" if we gonna land this PR. I think it's fine though because this:
```ts
invoke: { src: logic, id: 'myId' },
entry: sendTo('myId', event)
```
is way more useful and probable than:
```ts
invoke: { src: logic, systemId: 'mySystemId' },
entry: sendTo(({ system }) => system.get('mySystemId'), event)
```

Note that this whole problem is **only** relevant when an entry action tries to grab the actor that is just about to be invoked by the very same state from the system. As showcased by the test adjustment - the actor becomes available in the system right after entry+invoke on a single state are resolved

davidkpiano: Yeah, that makes sense

packages/core/src/interpreter.ts
davidkpiano, line 511:
```suggestion
        // TODO: add test case
        params.to ?? this,
```

davidkpiano (commented):
I'm a bit confused about the `retries` part - what's happening there?

packages/core/src/interpreter.ts
Andarist: done https://github.com/statelyai/xstate/pull/4351/commits/8973e43ba6dde285b925fc5c198c9d1727765c45

Andarist:
> I'm a bit confused about the retries part - what's happening there?

Entry actions are executed first (before invokes) so `sendTo('foo', ...)` might try to target a yet-not-available invoked actor that is meant to be created by the same "action block". When resolving entry blocks, I set aside all actions that have `.retryResolve` on them (so basically just send actions) and I attempt to call those retries at the end of the block. At that time, the invoked actors are already available so I have a chance to mutate the already created action params. This wouldn't work for just any action type though - it specifically works for sends because their execution is wrapped with `defer`.

davidkpiano (approved):

PR #4352, opened 2023-10-12
Modify useMachine config instance warning

kirill578:
This change will address the 2nd edge case covered in the documentation I updated. currently there is no way to modify a reference of an action without triggering the useMachine warning.

### DIFF
--- a/docs/recipes/react.md
+++ b/docs/recipes/react.md
@@ -296,11 +296,11 @@ const machine = createMachine(config);
 
 const Component = () => {
   const doStuff = useSomeHook();
-  const [state, send] = useMachine(machine, React.useMemo(() => {
+  const [state, send] = useMachine(machine, React.useMemo(() => ({
     actions: {
       someHookAction: (ctx) => doStuff(ctx);
     }
-  }, [doStuff]));
+  }), [doStuff]));
   return <></>
 }
 ```
--- a/docs/recipes/react.md
+++ b/docs/recipes/react.md
@@ -280,3 +280,64 @@ class Toggle extends React.Component {
   }
 }
 ```
+
+## Reference hooks safely
+
+When working with React hook we might want to pass a reference to a hook callback that is changing between renders. When done incorrectly you might notice this warning in the console log `Machine given to 'useMachine' has changed between renders.`. This can be addressed in the following ways:
+
+```jsx
+const config = {
+  id: 'machine',
+  /* ... */
+  entry: 'someHookAction', // actions have to be referenced by name
+} as const;
+
+const machine = createMachine(config);
+
+const Component = () => {
+  const doStuff = useSomeHook();
+  const [state, send] = useMachine(machine, React.useMemo(() => {
+    actions: {
+      someHookAction: (ctx) => doStuff(ctx);
+    }
+  }, [doStuff]));
+  return <></>
+}
+```
+
+The above approach won't work if you invoke nested machine. Although a bit more verbose the example below addresses the issue where the config is preserved while the reference to `doStuff` can chance between renders
+
+```jsx
+const configA = {
+  id: 'machineA',
+  /* ... */
+  entry: 'someHookAction', // actions have to be referenced by name
+} as const;
+
+const createMachineA = ({ doStuff }) => createMachine(configA, {
+
+});
+
+const configB = {
+  id: 'machineB',
+  /* ... */
+  invoke: {
+      src: 'machineA'
+      id: 'machineA'
+  }
+} as const;
+
+const createMachineB = ({ doStuff }) => createMachine(configB, {
+  services: {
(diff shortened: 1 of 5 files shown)


docs/recipes/react.md
```diff
@@ -280,3 +280,64 @@ class Toggle extends React.Component {
...
+
+const machine = createMachine(config);
+
+const Component = () => {
+  const doStuff = useSomeHook();
+  const [state, send] = useMachine(machine, React.useMemo(() => {
+    actions: {
+      someHookAction: (ctx) => doStuff(ctx);
+    }
+  }, [doStuff]));
```
Andarist, line 303:
```suggestion
  const [state, send] = useMachine(machine, React.useMemo(() => ({
    actions: {
      someHookAction: (ctx) => doStuff(ctx);
    }
  }), [doStuff]));
```

docs/recipes/react.md
```diff
@@ -280,3 +280,64 @@ class Toggle extends React.Component {
...
+The above approach won't work if you invoke nested machine. Although a bit more verbose the example below addresses the issue where the config is preserved while the reference to `doStuff` can chance between renders
+
+```jsx
+const configA = {
+  id: 'machineA',
+  /* ... */
+  entry: 'someHookAction', // actions have to be referenced by name
+} as const;
+
+const createMachineA = ({ doStuff }) => createMachine(configA, {
```
Andarist, line 317: We don't want to put this into documentation because it makes it harder for our tooling to perform static analysis when visualizing machines, generating typegen information etc.

docs/recipes/react.md
```diff
@@ -280,3 +280,64 @@ class Toggle extends React.Component {
...
+  const [state, send] = useMachine(machine, React.useMemo(() => {
+    actions: {
+      someHookAction: (ctx) => doStuff(ctx);
+    }
+  }, [doStuff]));
+  return <></>
+}
+```
+
+The above approach won't work if you invoke nested machine. Although a bit more verbose the example below addresses the issue where the config is preserved while the reference to `doStuff` can chance between renders
```
Andarist, line 308: I don't quite understand this. We update all implementation types here:
https://github.com/statelyai/xstate/blob/f807e7a98d3854eb789260bd769bf1c3bc7880bb/packages/xstate-react/src/useInterpret.ts#L74-L83

So if you only pass a new second argument to the `useMachine` hook you should benefit from this mentioned `useEffect` "sync".

kirill578: The issue is not with the root machine that is passed to `useMachine` but the nested actions inside a service that is used by the root machine.

we have a pattern in our application where we inject the dependencies like so:
```js
const buildMachineA = (args) => createMachine({
  ...
}, {
   services: {
      machineB: () => buildMachineB(args)
   }
}

const buildMachineB = (args) => createMachine({
   ...
}, {
   actions: {
      doStuff: () => args.makeApiCall()
   }
}
```

the goal is to initialize the machine like so 
```js
useMachine(buildMachineA(args))
```
but it triggers the warning unless we make the change to the warning logic.

if you are referring to something like that:
```js
useMachine(buildMachineA(args), { services: {
     machineB: () => buildMachineB(args)
}})
```
it will still trigger the warning since the whole machine reference is replaced. In reality our pattern has nesting that can be infinity deep so I don't we could use the 2nd argument approach. 


if my example is still unclear take a look at the unit tests that I update it highlights the difference between what is a valid reference chance and what is no. 

for more details see this discord conversation:  
https://discord.com/channels/795785288994652170/801127053092716596/1159892665408684062

kirill578: let me know if I am missing something :)

docs/recipes/react.md
kirill578: is the tooling using the machine parser? if that is the case from my experience it is not thrown off by a constructor pattern. what would be the alternative. I just wanted to make sure if someone is getting the error in their application a quick google search will lead them to here so they can better understand what the issue is and how it should be addressed

docs/recipes/react.md
Andarist: I'm still somewhat confused about this. Could you wrap this in a codesandbox that would showcase your current problem?

kirill578: Just realized that I totally missed the root cause of the issue. Rewriting it into a basic demo uncovered the problem as I could reproduce it at first.
here is an example of our setup that reproduces the warning
https://codesandbox.io/p/sandbox/weathered-resonance-krwyst?file=%2Fsrc%2FTodo.tsx%3A10%2C22
mirror: https://gist.github.com/kirill578/fb4ff09e412519935f47e201e027fc9d

the issue is caused by this piece of code where we initialize the context of the inner machine
image

swapping it for withContext eliminates the warning
image
 
 
 I think we can close the PR. Let me know if you have any concerns about the approach.

PR #4353, opened 2023-10-15
[v5] `setup()` API

davidkpiano:
```ts
const getChatCompletion = fromPromise(async () => { ... });
const processResult = fromPromise(async () => { ... });
const sendToDiscord = fromPromise(async () => { ... });

const machine = setup({
  actors: {
    getChatCompletion,
    processResult,
    sendToDiscord
  }
}).createMachine({
  // ...
  states: {
    thinking: {
      invoke: { src: 'getChatCompletion', onDone: 'processing' }
    },
    processing: {
      invoke: { src: 'processResult', onDone: 'sending' }
    },
    sending: {
      invoke: { src: 'sendToDiscord', onDone: 'done' }
    },
    done: { type: 'final' }
  }
});
```


packages/core/test/assign.test.ts
```diff
@@ -361,3 +367,56 @@ describe('assign meta', () => {
...
+      stringSrc
+    },
+    actions: {
+      doSomething({ params }: { params: { foo: string } }) {
+        return params.foo;
+      }
+    }
+  });
+
+  p.createMachine({
```
davidkpiano, line 385: This fails without `provide()`

davidkpiano:
```ts
types: {
  children: {} as {
    [key: `todo-${number}`]: 'todo';
    ['my-notifier']: 'notifierSrc';
  }
}
```

Or:

```ts
types: {
  childrenIds: {} as {
    todo: `todo-${number}`,
    notifierSrc: 'my-notifier',
  }
}
```

nstadigs:
Big fan of this change! Would it be weird to also support setting the event type in `setup()`? A framework could then use `setup` to export a fully preconfigured `createMachine` function where all the possible events and side effects are provided beforehand without using type hacks, no? See this discussion for context: https://github.com/statelyai/xstate/discussions/4315

Andarist:
> Big fan of this change! Would it be weird to also support setting the event type in setup()? 

That's the current plan.

> A framework could then use setup to export a fully preconfigured createMachine function where all the possible events and side effects are provided beforehand without using type hacks, no? See this discussion for context: https://github.com/statelyai/xstate/discussions/4315

I'm not sure if that would solve this particular issue though. The event type could be "preconfigured" here but other than that it would be set in stone. IIRC, you wanted to extend it with more event types.

with-heart:
Don't forget to add `setup` to the `core` exports 😆

with-heart:
Hey @davidkpiano, could you expand the example in the PR description to show envisioned usage for the other keys of the `setup` object param?

I really like the idea of defining the implementations up front and deriving types from them as much as possible but I'm struggling with thinking through how this could work with built-in actions like `assign`.

davidkpiano:
> Hey @davidkpiano, could you expand the example in the PR description to show envisioned usage for the other keys of the `setup` object param?
> 
> I really like the idea of defining the implementations up front and deriving types from them as much as possible but I'm struggling with thinking through how this could work with built-in actions like `assign`.

Hey sure!

So, you can always inline `assign(...)`. But if you want to "serialize" them:

```ts
const machine = setup({
  actions: {
    setToZero: assign({ 
      count: 0
    })
  }
}).createMachine({
  entry: 'setToZero'
});
```

Not sure how context typings will work, or if/how parameters will work with this yet (cc. @Andarist)

nstadigs:
> > Big fan of this change! Would it be weird to also support setting the event type in setup()?
> 
> That's the current plan.
> 
> > A framework could then use setup to export a fully preconfigured createMachine function where all the possible events and side effects are provided beforehand without using type hacks, no? See this discussion for context: #4315
> 
> I'm not sure if that would solve this particular issue though. The event type could be "preconfigured" here but other than that it would be set in stone. IIRC, you wanted to extend it with more event types.

In my case the file providing the preconfigured 'createMachine' (using setup) would be generated, so extra event types can be specified before generating the code.

Andarist (approved):

Necmttn:
THANK YOU for this! This cleared up quite a bit type inference I was doing.

joernroeder:
I'm just stumbling upon this from within the changelog. can someone give more context what this replaces? is it just the call signature of `createMachine`? what's the difference to `myMachine.provide({ actors: { myActor } }}`

thanks in advance

joernroeder:
Thanks @davidkpiano for the swift response and clarification, love it!

PR #4371, opened 2023-10-19
Avoid selecting always transitions when the preceeding transition doesn't change the state

Andarist:


packages/core/src/stateUtils.ts
```diff
@@ -1621,7 +1626,7 @@ export function macrostep(
       false,
       internalQueue
     );
-
+    shouldSelectEventlessTransitions = nextState !== previousState;
```
Andarist, line 1629: It's worth noting that this also checks non-always transitions here. This check also doesn't cover the first `selectTransitions`+`microstep` within this `macrostep`.

To be more consistent we should either:
- check the same for that first `microstep`
- or only check this after executing an `always` transition

Which one sounds better to you?

davidkpiano: I think "or only check this after executing an always transition" sounds better

Andarist: I agree and I've pushed out an according change

davidkpiano (approved):

PR #4372, opened 2023-10-19
Remove `State['_internalQueue']`

Andarist:


davidkpiano (approved):
Approved when CI passes

PR #4374, opened 2023-10-19
Add StateValue @remarks

audionerd:
this fixes an error when generating API documentation which prevents an documentation index page from being created

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -284,6 +284,8 @@ export interface StateValueMap {
 /**
  * The string or object representing the state value relative to the parent state node.
  *
+ * @remarks
+ *
  * - For a child atomic state node, this is a string, e.g., `"pending"`.
  * - For complex state nodes, this is an object, e.g., `{ success: "someChildState" }`.
  */


packages/core/src/types.ts
```diff
@@ -284,6 +284,8 @@ export interface StateValueMap {
 /**
  * The string or object representing the state value relative to the parent state node.
  *
+ * @remarks
```
Andarist, line 287: can this whole JSDoc be rewritten somehow? `@remarks` seems to be somewhat non-standard so, for example, the IDE doesn't really understand it and displays `@remarks` verbatim

audionerd: For context, this is related to a PR I’m working on to generate XState API docs for the Stately site using API Extractor.

API Extractor expects TSDoc, and I realize now that @remarks happens to be something only TSDoc has:

https://api-extractor.com/pages/tsdoc/tag_remarks

Andarist: Yeah, so I think that perhaps there is something here that specifically breaks the API generation, like this markdown list (but maybe that's not it). So perhaps by rewriting this, by avoiding the problematic pattern - we wouldn't have to introduce non-standard `@remarks`

audionerd: Yes, it should be possible to re-write this comment to be JSDoc compatible, and still ensure the Markdown generated by API Extractor can be rendered without error by Docusaurus. I’ll take a look.

Also this will make more sense once I have the API docs PR up for review.

Andarist:
I'm closing this one as superseded by https://github.com/statelyai/xstate/pull/4391

Issue #4167, opened 2023-07-29
Bug: Exit action triggered twice when invoked machine escalates from final state


andrejpavlovic:
### Description

If a child machine has a top-level exit action, this action will be triggered twice when the child machine ends in a final state that triggers an escalate action.

Basically a child machine like this one will trigger exit action twice when reaching final `cancelled` state.

```typescript
const childMachine = createMachine({
  predictableActionArguments: true,
  id: "child",
  initial: "start",
  entry: () => {
    console.log('child-entry')
  },
  exit: () => { // <-- triggered twice when reaching final `cancelled` state
    console.log('child-exit')
  },
  states: {
    start: {
      on: {
        COMPLETE: "completed",
        CANCEL: "cancelled"
      }
    },
    completed: {
      type: "final"
    },
    cancelled: { // <-- final cancelled state
      type: "final",
      entry: actions.escalate({ type: "cancelled" })
    }
  }
});

send('CANCEL')
```

### Expected result

```
child-entry
child-exit
```

### Actual result

```
child-entry
child-exit
child-exit
```

### Reproduction

https://codesandbox.io/s/ecstatic-euler-pqtsfj?file=/src/index.js

### Additional context

- Tested with version `4.38.2`
- Used to work fine with version `4.32.1`, regression starts with `4.33.0` and higher
- Regression most likely caused by #3126
- Looks like latest v5 (5.0.0-beta.20) also has this issue. Repro https://codesandbox.io/s/silly-maxwell-49h4v5?file=/src/index.js

PR #4376, opened 2023-10-20
Fixed an issue with exit actions being called twice when machine reached its final state

Andarist:
fixes https://github.com/statelyai/xstate/issues/4167

### DIFF
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -1578,7 +1578,7 @@ export function macrostep(
 
   // Handle stop event
   if (event.type === XSTATE_STOP) {
-    nextState = stopStep(event, nextState, actorCtx);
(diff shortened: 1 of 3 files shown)


packages/core/src/stateUtils.ts
```diff
@@ -1638,8 +1638,7 @@ export function macrostep(
   }
 
   if (nextState.status !== 'active') {
-    // Perform the stop step to ensure that child actors are stopped
-    stopStep(nextEvent, nextState, actorCtx);
+    stopChildren(nextState, nextEvent, actorCtx);
```
Andarist, line 1641: this likely tries to stop invoked actors for the second time and I have a feeling that this might be a problem as well, I'm gonna investigate this separately

davidkpiano (approved):

PR #4377, opened 2023-10-20
Stop calling exit actions on stop

Andarist:
It's important for rehydration. A stopped aka persisted machine can't call its exit actions if we don't call entry actions later on when rehydrating that state. And we can't call those entry actions because we have no sensible `event` to call them with. In addition to that - the whole persisting/rehydrating cycle should be mostly transparent to the user and if we call those actions it's no longer transparent.

Originally, we only started to call those actions in this situation because it was what SCXML mandated. It didn't come from a user request.

### DIFF
--- a/.changeset/wild-apes-play.md
+++ b/.changeset/wild-apes-play.md
@@ -1,0 +1,5 @@
+---
+'xstate': major
+---
+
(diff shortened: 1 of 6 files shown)


davidkpiano (approved):
Can we have a changeset for this?

PR #4390, opened 2023-10-24
Remove `TPersisted` generic

Andarist:


davidkpiano (commented):
I'm worried about the `any`'s - how will this affect dev code?

Andarist:
> I'm worried about the any's - how will this affect dev code?

It kinda shouldn't really. We assume that the persisted state comes from an untrusted source (like a network or a local storage). At that point, the true type is `unknown` anyway - despite what our types or user's casts might say. We don't have any means of actually validating those things so we were always trusting the inputs here and that was always unsafe but it gave an illusion of safety. 

It might only affect scenarios like this:
```ts
const machine1 = createMachine({ /* ... */ });
const machine2 = createMachine({ /* ... */ });

const actorRef1 = createActor(machine1).start()

createActor(machine2, { state: actorRef1.getPersistsedState() })
```

In this scenario, we lose some type-safety because everything stays "local" so the type-safety was actually here before this change. I'd say though that such in-memory persistence+rehydration isn't the primary use of those APIs and for simplicity's sake I'd prefer to treat those with the same type-safety guarantees as `JSON.stringify` and `JSON.parse` come with in TS.

davidkpiano (approved):

PR #4391, opened 2023-10-24
Fixes to support Generating API Documentation

audionerd:
Companion PR for https://github.com/statelyai/docs/pull/237, which now uses `xstate` as a submodule, and requires these changes in order to build API docs.

This PR fixes issues that prevented the XState API docs index page from rendering properly, and fixes two warnings that occurred when API Documenter ran.

Includes and closes https://github.com/statelyai/xstate/pull/4374

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -290,6 +290,7 @@ export interface StateValueMap {
  * @remarks
  *
  * - For a child atomic state node, this is a string, e.g., `"pending"`.
+ *
  * - For complex state nodes, this is an object, e.g., `{ success: "someChildState" }`.
  */
 export type StateValue = string | StateValueMap;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -23,6 +23,9 @@ export type HomomorphicOmit<T, K extends keyof any> = {
 };
 
 /**
+ *
+ * @remarks
+ *
  * `T | unknown` reduces to `unknown` and that can be problematic when it comes to contextual typing.
  * It especially is a problem when the union has a function member, like here:
  *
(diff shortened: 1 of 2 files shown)


davidkpiano (approved):

packages/core/src/types.ts
```diff
@@ -284,6 +287,8 @@ export interface StateValueMap {
 /**
  * The string or object representing the state value relative to the parent state node.
  *
+ * @remarks
```
Andarist, line 290: I'm still quite unsure why we need this. RxJS has *extensive* and long JSDocs and they use api-extractor too but there is no `@remarks` in sight in their codebase. I wonder what's the difference? 🤔

audionerd: On the index page a description is provided for each item.

With `@remarks`, we can choose to show only the first summary line in the descriptions:
index-with-remarks

Without `@remarks`, the description is longer (and most of the index descriptions are a single line so this one stands out):
index-without-remarks

So it’s a subjective change, but I thought it was better to have a shorter description. If you disagree I’m happy to change it back.

I had also noticed the list items weren’t formatted properly on the index, so I wanted to omit them, but I just found a fix for that I can push to this PR, so it's not a deciding factor.

audionerd: API Extractor uses TSDoc, and `@remarks` is specific to TSDoc. See: https://tsdoc.org/pages/tags/remarks

packages/core/src/types.ts
```diff
@@ -835,7 +841,7 @@ export interface StateNodeConfig<
     TDelay
   >;
   /**
-   * @private
+   * @internal
```
Andarist, line 844: We are using `stripInternal` and I'm not sure if we want to "hide" this parent property like this. I'd remove this particular comment completely.

Andarist (approved):

packages/core/src/types.ts
audionerd: That's a good point. I’ll see what I can find out about TSDocs in VS Code. Might be some ways to improve the experience.

PR #4393, opened 2023-10-24
Switch from JSDoc to TSDoc

audionerd:
I linted with `eslint-plugin-tsdoc` to identify cases where the JSDoc syntax used is not compatible with TSDoc. Found a couple cases and fixed them.

There is one remaining issue with `@param`, but I'm ignoring that for now. In JSDoc [`@param` can be name-only](https://jsdoc.app/tags-param.html#examples). But in TSDoc, `@param` is expected to always have a hyphen after the name (even if description is missing). We could avoid that linter warning by simply appending a hyphen after every `@param` name. But because it’s just a linter warning, and doesn’t make a difference to API Extractor, I didn’t bother.

### DIFF
--- a/packages/core/src/waitFor.ts
+++ b/packages/core/src/waitFor.ts
@@ -6,7 +6,7 @@ interface WaitForOptions {
    * How long to wait before rejecting, if no emitted
    * state satisfies the predicate.
    *
-   * @default 10_000 (10 seconds)
+   * @defaultValue 10_000 (10 seconds)
    */
   timeout: number;
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/waitFor.ts
```diff
@@ -6,7 +6,7 @@ interface WaitForOptions {
    * How long to wait before rejecting, if no emitted
    * state satisfies the predicate.
    *
-   * @default 10_000 (10 seconds)
+   * @defaultValue 10_000 (10 seconds)
```
Andarist, line 9: I'm quite surprised that this one matters - they are supposed to be synonyms in JSDoc 🤷

Andarist: It's worth noting that they display slightly differently in VS Code but the difference is kinda negligible and I'm not even sure which one is "better"/more expected in this context

PR #4399, opened 2023-10-25
Add warning for sending events to non-started actors

davidkpiano:


Andarist:
Sending to non-started actors is a valid thing to do when integrating with UI frameworks. The user might send the event to the actor from a child component’s `useEffect` and those are execited before the parent has a chance to start the actor

Issue #4364, opened 2023-10-17
Bug: Actor with system ID '...' already exists. (v5)


rlaffers:
### Description

Invoked actors are not cleared from the system upon re-entry.

```javascript
const machine = createMachine({
  initial: 'Listening',
  states: {
    Listening: {
      invoke: {
        systemId: 'listener', // <-- This will clash with the subsequently invoked listener
        src: fromCallback(() => {
          console.log('✅ invoked the listener')
          return () => {
            console.log('⛔ clearing the listener')
          }
        }),
      },
    },
  },
  on: {
    RESTART: {
      // reenters Listening, tries to invoke the listener, but:
      // 🪲 Actor with system ID 'listener' already exists.
      target: '.Listening',
    },
  },
})
```

### Expected result

After receiving the RESTART event, the first listener is cleared first (its unsubscribe function should run), its systemId should be removed from actor.system, and then a fresh listener should be invoked.

### Actual result

Error upon receiving the RESTART event:

> Error: Actor with system ID 'listener' already exists.

### Reproduction

https://codesandbox.io/s/optimistic-bush-7lh38j?file=/src/index.js

### Additional context

This is an issue with xstate v5.
Occurs with: xstate@5.0.0-beta.34

Andarist:
The issue is caused by the fact that we defer stopping the previous child:
https://github.com/statelyai/xstate/blob/eea74c5943e9b2157934d260b793e06169099b41/packages/core/src/actions/stop.ts#L51-L71

but we register a new one in the system asap, when it's created. An actor instantiation is not deferred (and can't be deferred).

So the fix for this is to unregister the old actor from the system eagerly.

PR #4405, opened 2023-10-26
Unregister `systemId`s eagerly

Andarist:
fixes https://github.com/statelyai/xstate/issues/4364

### DIFF
--- a/packages/core/test/system.test.ts
+++ b/packages/core/test/system.test.ts
@@ -424,4 +424,49 @@ describe('system', () => {
 
     expect(actor.system.get('test')).toBeDefined();
   });
+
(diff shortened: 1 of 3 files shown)


packages/core/test/system.test.ts
```diff
@@ -424,4 +424,49 @@ describe('system', () => {
 
     expect(actor.system.get('test')).toBeDefined();
   });
+
+  it('da', () => {
```
davidkpiano, line 428: Da

Andarist: it's **da** test

davidkpiano (approved):

PR #4407, opened 2023-10-27
[v5] ActorContext -> ActorScope

davidkpiano:
Changed the actor context type from `ActorContext` to `ActorScope` to mitigate confusion.


Andarist (approved):
I'll miss the idea of `ActorStuff` but `ActorScope` is a great name for this

PR #4414, opened 2023-10-29
[v5] Consistent `actorRef` property for inspection events

davidkpiano:
All inspector events (snapshot, event, actor) now have a common `actorRef` property. This makes it easier to discern which inspection event is for which actor:

```ts
const actor = createActor(someMachine, {
  inspect: (event) => {
    // Was previously a type error
    if (event.actorRef === actor) {
      // This event is for the root actor
    }

    if (event.type === '@xstate.event') {
      // previously event.targetRef
      event.actorRef;
    }
  }
});
```

In the `'xstate.event'` event, the `actorRef` property is now the target actor (recipient of the event). Previously, this was the `event.targetRef` property (which is now removed).

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -313,7 +313,7 @@ export class Actor<TLogic extends AnyActorLogic>
     this.system._sendInspectionEvent({
       type: '@xstate.event',
       sourceRef: this._parent,
-      targetRef: this,
+      actorRef: this,
       event: initEvent
     });
 
--- a/packages/core/src/system.ts
+++ b/packages/core/src/system.ts
@@ -59,7 +59,7 @@ export function createSystem<T extends ActorSystemInfo>(
       system._sendInspectionEvent({
         type: '@xstate.event',
         sourceRef: source,
-        targetRef: target,
+        actorRef: target,
         event
       });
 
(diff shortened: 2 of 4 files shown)


packages/core/src/system.ts
```diff
@@ -71,11 +71,17 @@ export function createSystem<T extends ActorSystemInfo>(
...
+   * - For snapshot events, this is the `actorRef` of the snapshot.
+   * - For event events, this is the target `actorRef` (recipient of event).
+   * - For actor events, this is the `actorRef` of the registered actor.
+   */
+  actorRef: AnyActorRef;
 }
 
 export interface InspectedSnapshotEvent extends BaseInspectionEventProperties {
   type: '@xstate.snapshot';
-  actorRef: AnyActorRef; // Only available locally
```
Andarist, line 78: note that by "duplicating" this property here you could have a JSDoc unique to this particular event type here

packages/core/src/interpreter.ts
```diff
@@ -313,7 +313,7 @@ export class Actor<TLogic extends AnyActorLogic>
     this.system._sendInspectionEvent({
       type: '@xstate.event',
       sourceRef: this._parent,
-      targetRef: this,
+      actorRef: this,
```
Andarist, line 316: I see the argument of having a "common" `actorRef` but when looking at this particular event in isolation... I think that the `sourceRef` and `targetRef` were more readable. I see `actorRef` here and I'm like - ye, sure, but which one?

Andarist (approved):

packages/core/src/interpreter.ts
davidkpiano: Can sorta be a process-of-elimination though: if it's not the source, it must be the target

PR #4416, opened 2023-10-30
API Docs: createActor

audionerd:
Adds documentation for createActor with two overloaded signatures.

Moved the most common signature (logic) to the top, so VS Code IntelliSense will suggest it first.

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -559,7 +559,33 @@ export class Actor<TLogic extends AnyActorLogic>
 /**
  * Creates a new actor instance for the given actor logic with the provided options, if any.
  *
- * @param logic - The actor logic to create an actor from
+ * @remarks
+ * When you create an actor from actor logic via `createActor(logic)`, you implicitly create an actor system where the created actor is the root actor.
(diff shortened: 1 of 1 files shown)


packages/core/src/interpreter.ts
```diff
@@ -570,10 +580,6 @@ export function createActor<TMachine extends AnyStateMachine>(
     : MissingImplementationsError<TMachine['__TResolvedTypesMeta']>,
   options?: ActorOptions<TMachine>
 ): Actor<TMachine>;
-export function createActor<TLogic extends AnyActorLogic>(
-  logic: TLogic,
-  options?: ActorOptions<TLogic>
-): Actor<TLogic>;
```
davidkpiano, line 576: We might still need this, since machines are handled differently.

audionerd: Wasn’t removed, but was moved up. 

Moved because I noticed IntelliSense will suggest the first function override it finds, and I thought we'd want the actor logic one to be listed first in the list.

audionerd:
I could also throw in the notes and example from the main docs, if that would be helpful: https://stately.ai/docs/actors#creating-actors

davidkpiano (approved):

PR #4419, opened 2023-10-31
API Docs: ActorOptions

audionerd:
- [x] `inspect`
- [x] `clock`
- [x] `logger`
- [x] `state`, and `Actor#getPersistedState` (cross-referenced)
- [x] `parent` - Not sure how this is used or how to describe it. Is it used within actor systems?
- [x] `sync` - Is this used?
- [x] rewrite `inspect` example to reflect recent changes (https://github.com/statelyai/xstate/pull/4414)

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1704,6 +1704,45 @@ export interface ActorOptions<TLogic extends AnyActorLogic> {
    */
   src?: string | AnyActorLogic;
 
+  /**
+   * A callback function which can be used to inspect actor system updates.
+   *
+   * @remarks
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -1708,6 +1741,49 @@ export interface ActorOptions<TLogic extends AnyActorLogic> {
    */
   src?: string;
 
+  /**
+   * A callback function which can be used to inspect actor system updates.
```
davidkpiano, line 1745: It can be a callback function or observer (`{ next?, error?, complete? }`)

davidkpiano (approved):

PR #4429, opened 2023-11-05
[v5] `createAction()` (replaces `pure()` and `choose()`)

davidkpiano:
```ts
actions: createAction(({ context, event, exec }) => {
  exec.assign({
    count: context.count + 1
  });

  // Conditional actions (replaces choose(...))
  if (event.someOption) {
    exec.sendTo('someActor', { type: 'blah', thing: context.thing });
    
    // other actions
    exec.action('namedAction');
    // with params
    exec.action({ type: 'greet', params: { message: 'hello' }});
  } else {
    // inline
    exec.action(() => console.log('hello'));
  }

  // no return
})
```

### DIFF
--- a/packages/core/test/actions.test.ts
+++ b/packages/core/test/actions.test.ts
@@ -1,9 +1,7 @@
 import { ActorRef, EventObject } from '../src/index.ts';
 import {
   cancel,
-  choose,
   log,
-  pure,
   raise,
   sendParent,
   sendTo,
@@ -18,6 +16,7 @@ import {
 } from '../src/index.ts';
 import { CallbackActorRef, fromCallback } from '../src/actors/callback.ts';
 import { trackEntries } from './utils.ts';
+import { createAction } from '../src/actions/pure.ts';
 
 const originalConsoleLog = console.log;
 
@@ -2385,11 +2384,17 @@ describe('purely defined actions', () => {
         context: {
           items: [{ id: 1 }, { id: 2 }, { id: 3 }]
         },
-        entry: pure(({ context }) => {
-          return {
+        // entry: pure(({ context }) => {
+        //   return {
+        //     type: 'doSomething',
+        //     params: { length: context.items.length }
+        //   };
+        // })
+        entry: createAction(({ exec, context }) => {
+          exec.action({
             type: 'doSomething',
             params: { length: context.items.length }
-          };
+          });
         })
       },
       {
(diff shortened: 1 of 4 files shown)


packages/core/test/actions.test.ts
```diff
@@ -2950,9 +3027,14 @@ describe('choose', () => {
...
-          conditionallyRevealAnswer: choose([
-            { guard: 'worstGuard', actions: 'revealAnswer' }
-          ])
+          // conditionallyRevealAnswer: choose([
+          //   { guard: 'worstGuard', actions: 'revealAnswer' }
+          // ])
+          conditionallyRevealAnswer: createAction(({ exec, guard }) => {
+            if (guard({ type: 'worstGuard' })) {
+              exec.action({ type: 'revealAnswer' });
+            }
```
with-heart, line 3036: I love this. A little more verbose but feels more natural than `choose`

tivac:
I love this, is there any conversation or references anywhere about where it came from?

davidkpiano:
> I love this, is there any conversation or references anywhere about where it came from?

Some conversation here: https://discord.com/channels/795785288994652170/1169717611416789083

It builds up from previous discussions about a `capture()` API, but most recently, just the frustration in typing `choose()` and `pure()` 😅

asfktz:
How will it behave when with async logic?

I imagine new users will attempt something like this:

```js
actions: createAction(async ({ context, event, exec }) => {
  exec.assign({ isFetching: true });
  
  const res = await fetchTodos(1000);
  
  exec.assign({ todos: res.data, isFetching: false });  
})
```

Andarist:
Async logic would not be permitted here

asfktz:
Sounds Perfect 👌

Also, I'm curious, what is the reason behind using `exec.assign` / `exec.action` / etc.. 
over just `exec(assign({}))`?


Are they just shortcut for better DX?

While I do like it, I wonder about the case of using a custom assign, like [@xstate/immer](https://xstate.js.org/docs/packages/xstate-immer/).
If the rest of the machine is using [@xstate/immer](https://xstate.js.org/docs/packages/xstate-immer/), it my be confusing to have access to the default assign on `exec.assign`.

Without the shortcuts, it's one less thing to think about.


```js

import { assign } from '@xstate/immer';

//...

actions: [
  // a custom assign imported from "@xstate/immer"
  assign(({ context }) => {
    context.value = 'foo';
  }),
  createAction(({ context, event, exec }) => {
    // The assign we imported
    exec(assign({
      count: context.count + 1
    }));
  })
]
```

What do you think?

davidkpiano:
> Also, I'm curious, what is the reason behind using `exec.assign` / `exec.action` / etc..
> over just `exec(assign({}))`?

Makes typing stronger; no need to infer types, which isn't perfect. We can strongly type `exec.assign` etc. upfront.

packages/core/src/actions/enqueueActions.ts
```diff
@@ -0,0 +1,224 @@
...
+  TActor extends ProvidedActor = ProvidedActor,
+  TAction extends ParameterizedObject = ParameterizedObject,
+  // `TGuard` is only required by `choose`, if we remove `choose` then we can remove this whole type parameter
+  TGuard extends ParameterizedObject = ParameterizedObject,
+  TDelay extends string = string
+>(
+  collect: ({
+    context,
+    event,
+    guard,
```
Andarist, line 182: WDYT about calling this `check`?

Andarist:
> Without the shortcuts, it's one less thing to think about.

Note that this should still "just work" - so you can very much do this.

Andarist (approved):

PR #4433, opened 2023-11-07
Remove the unsupported machine implementation types from `useActorRef`'s signature

Andarist:
`useActorRef(machine, { actions: { doStuff: () => {} } })` is not a supported flavour in v5 (it got replaced with `useActorRef(machine.provide({ actions: { doStuff: () => {} } }))`) but it confusingly was still available in the types


packages/xstate-react/test/typegenTypes.test.tsx
```diff
@@ -226,217 +226,6 @@ describe('useMachine', () => {
 });
 
 describe('useActorRef', () => {
-  it('should allow to be used with a machine without any missing implementations', () => {
```
Andarist, line 229: Those tests were no longer relevant to `useActorRef`. All of the implementation requirements are handled by `.provide`. Unfortunately, we allow for machines without all required implementation to be used with `useActorRef` *but* this PR doesn't change this. 

If anything we could follow the `useMachine`'s signature and do smth like this:
https://github.com/statelyai/xstate/blob/dce3b54fb4c7998f9111c67d15283b74790b0cba/packages/xstate-react/src/useMachine.ts#L15-L19

but, of course, it's a little bit unfortunate to embed machine-specific checks in `useActorRef`. The hope is that the `setup` API should just make it impossible to create machines without all of the required implementations and that would "solve" this problem.

davidkpiano (approved):

PR #4435, opened 2023-11-07
[v5] Remove `waitFor` timeout

davidkpiano:
The default `timeout` for `waitFor(...)` is now `Infinity` instead of 10 seconds.


packages/core/src/waitFor.ts
```diff
@@ -12,7 +12,7 @@ interface WaitForOptions {
 }
 
 const defaultWaitForOptions: WaitForOptions = {
-  timeout: 10_000 // 10 seconds
+  timeout: Infinity // much more than 10 seconds
```
Andarist, line 15: 😂

PR #4436, opened 2023-11-07
Improved Fast Refresh support

Andarist:


packages/core/src/interpreter.ts
```diff
@@ -158,7 +158,7 @@ export class Actor<TLogic extends AnyActorLogic>
     this.clock = clock;
     this._parent = parent;
     this.options = resolvedOptions;
-    this.src = resolvedOptions.src;
+    this.src = resolvedOptions.src || logic;
```
davidkpiano, line 161:
```suggestion
    this.src = resolvedOptions.src ?? logic;
```

davidkpiano (approved):

with-heart:
Big if true! 🙏👏

Andarist:
Note that this definitely has some caveats - but some of the basic scenarios should work fine and we do a better job at rehydrating state right now.

PR #4438, opened 2023-11-08
Removed `State#toStrings` method

Andarist:
This method seems to be so niche that I don't think it has to be on `State`


davidkpiano:
This still has usage: https://sourcegraph.com/search?q=context:global+state.toStrings+AND+xstate+lang:ts+-repo:%5Egithub%5C.com/statelyai/xstate%24+&patternType=standard&sm=1&groupBy=repo

We should at least move it to a helper function

Andarist:
We discussed this privately and decided that it should be OK to get rid off this for now and reevaluate later if the need arises.

PR #4440, opened 2023-11-08
Deduplicate `State` creation methods

Andarist:


davidkpiano (approved):
Just needs a changeset

davidkpiano (changes requested):
We actually need `createState` for an important reason: developer testing of pure logic. `new State(...)` is not really satisfactory since you manually have to specify the `configuration` and other things.

davidkpiano (approved):

PR #4441, opened 2023-11-08
Remove the internal `getInitialConfiguration`

Andarist:


packages/core/src/stateUtils.ts
```diff
@@ -1795,28 +1795,3 @@ export function stateValuesEqual(
     aKeys.every((key) => stateValuesEqual(a[key], b[key]))
   );
 }
-
-export function getInitialConfiguration(
-  rootNode: AnyStateNode
-): AnyStateNode[] {
-  const configuration: AnyStateNode[] = [];
-  const initialTransition = rootNode.initial;
```
Andarist, line 1803: I noticed that this whole function was returning an empty array for parallel roots (since they don't have `.initial`) and it turns out that the initial configuration is resolved anyway even without this whole function... so: 🗑️

davidkpiano (approved):

PR #4443, opened 2023-11-09
Declassify `State`

Andarist:


packages/xstate-graph/test/graph.test.ts
```diff
@@ -30,14 +29,14 @@ function getPathSnapshot(path: StatePath<Snapshot<unknown>, any>): {
 } {
   return {
     state:
-      path.state instanceof State
+      'machine' in path.state && 'value' in path.state
```
davidkpiano, line 32: We should have an `isState` helper

davidkpiano (approved):
Just needs a changeset

packages/xstate-graph/test/graph.test.ts
Andarist: Agreed. `isMachineSnapshot`?

PR #4444, opened 2023-11-09
Remove `mapState` util

Andarist:


davidkpiano (approved):
Let's add a quick changeset, just in case anyone was using this

PR #4445, opened 2023-11-09
Add `ActorLogic` interface docs

audionerd:
While working on docs for "Custom actor logic" I felt it would be helpful to have the `ActorLogic` interface better documented.

Related to https://github.com/statelyai/docs/pull/252

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -2078,7 +2078,7 @@ export interface ActorLogic<
   TInput = unknown,
   TSystem extends ActorSystem<any> = ActorSystem<any>
 > {
-  /** Optional configuration for the actor logic. */
+  /** The initial setup/configuration used to create the actor logic. */
   config?: unknown;
   /**
    * Transition function that processes the current state and an incoming message
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -2064,29 +2064,67 @@ export type Snapshot<TOutput> =
       error: undefined;
     };
 
+/**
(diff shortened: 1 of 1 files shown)


packages/core/src/types.ts
```diff
@@ -2064,29 +2064,67 @@ export type Snapshot<TOutput> =
...
+ * @template TSystem - The type of the actor system.
+ */
 export interface ActorLogic<
   TSnapshot extends Snapshot<unknown>,
   TEvent extends EventObject,
   TInput = unknown,
   TSystem extends ActorSystem<any> = ActorSystem<any>
 > {
+  /** Optional configuration for the actor logic. */
   config?: unknown;
```
audionerd, line 2082: Could use a better comment here re: why/how `config` is used.

davidkpiano (approved):

packages/core/src/types.ts
```diff
@@ -2064,29 +2064,67 @@ export type Snapshot<TOutput> =
...
+ * @template TInput - The type of the input.
+ * @template TSystem - The type of the actor system.
+ */
 export interface ActorLogic<
   TSnapshot extends Snapshot<unknown>,
   TEvent extends EventObject,
   TInput = unknown,
   TSystem extends ActorSystem<any> = ActorSystem<any>
 > {
+  /** Optional configuration for the actor logic. */
```
davidkpiano, line 2081:
```suggestion
  /** The initial setup/configuration used to create the actor logic. */
```

PR #4447, opened 2023-11-10
Reuse methods across machine snapshot instances

Andarist:


packages/core/src/State.ts
```diff
@@ -110,21 +110,39 @@ interface MachineSnapshotBase<
 
   /**
    * Whether the current state value is a subset of the given parent state value.
-   * @param parentStateValue
+   * @param testValue
```
Andarist, line 113: `parentStateValue` was quite confusing to me so I decided to change it

packages/core/src/State.ts
```diff
@@ -110,21 +110,39 @@ interface MachineSnapshotBase<
...
    */
   matches: <
     TSV extends TResolvedTypesMeta extends TypegenEnabled
       ? Prop<Prop<TResolvedTypesMeta, 'resolved'>, 'matchesStates'>
       : StateValue
   >(
-    parentStateValue: TSV
+    this: MachineSnapshotBase<
```
Andarist, line 120: I tried to just use `this: this` but unfortunately TS doesn't like it: [TS playground](https://www.typescriptlang.org/play?ts=5.4.0-dev.20231110#code/JYOwLgpgTgZghgYwgAgEJwM4oN4ChnIAmA9gMpgCuMMAXMgBRgAWwGdzrAlMgLwB8yAG7FghANy4AvrlyhIsRCgCCCMMEEoIAD0ghCGNJhz5kGMHEptkAIkRqN1idNnho8JMgAixEJp0Q9A3QsZDwCMwsKK2sSX0cpGTAATwAHFHILFB5kFXsUAB8vHwgJXEIIBAAbOCgUBB8zZBhiYjoMyFLm4gA6EnIqGHpOCSA)

It's also not mandatory at all to have `this` here in the annotation but it nicely guards people against mistakes like this: [TS playground](https://www.typescriptlang.org/play?ts=5.4.0-dev.20231110#code/JYOwLgpgTgZghgYwgAgMpjpA6sMALAFT2AGdkBvAKGWQFtME8IT1MIAKfUgLjQ21yFiJADTIAbnAA2AVwi8SYKKADmASl4AjAPbapEOCEoBfSpQAmEBFLhQUCbSEXJFbAIy9WA-EVKUHTmB0DEwkbsgAvC78EG4AdPRgjMxeEJSJyWHsAOTZamagkLCIKKk4+NoyYL5kVDQZoanskrLy0cog6lq6+oYmZpbWtvaOzq6QAEyeMeV4ldXC-qNBDcwTkdFsEwkhKTHpuyQTOXlAA)

packages/core/src/State.ts
```diff
@@ -288,57 +384,27 @@ export function createMachineSnapshot<
...
+    can: machineSnapshotCan,
+    toJSON: machineSnapshotToJSON
+  };
 
-      return (
-        !!transitionData?.length &&
-        // Check that at least one transition is not forbidden
-        transitionData.some((t) => t.target !== undefined || t.actions.length)
-      );
+  Object.defineProperties(snapshot, {
```
Andarist, line 394: I'd really love to get rid of those getters. We don't even benefit from them being getters since we "tend to" spread those objects anyway, like here:
https://github.com/statelyai/xstate/blob/7c1497318799c6abe638e3dfa4639f86b880c18f/packages/core/src/State.ts#L415-L416

That evaluates them anyway, almost for each state. Currently we end up computing "a lot" of temporary states within a single transition and each `cloneMachineSnapshot` pays the price of invoking them. 

The problem isn't only unique to those getters. Some other things are computed by each `createMachineSnapshot` as well.

packages/xstate-solid/src/deriveServiceState.ts
Andarist, line 1: 😍

packages/core/src/State.ts
davidkpiano: What's the alternative? I think a `getNextTransitions()` is more accurate than a `nextEvents`, and we can have a `getMeta()` function.

davidkpiano (approved):

packages/core/src/State.ts
Andarist: > What's the alternative?

I'd like to just replace those with methods or simple properties - this way we would end up with more predictable behavior around those and they just would work like everything else here. 

> I think a getNextTransitions() is more accurate than a nextEvents, and we can have a getMeta() function.

Ye, I think that making them methods like those would be good. That's a topic for a follow up PR though. I'm also not sure if `getNextTransitions` would be the best replacement name for `nextEvents` since we get a list of strings back and that doesn't scream "transitions" to me. Either way, we can bikeshed later on :p

PR #4448, opened 2023-11-10
Replace `isState`/`isStateConfig` with `isMachineSnapshot`

Andarist:


davidkpiano (commented):
Are we sure we don't want to at least alias `AnyState` -> `AnyMachineSnapshot`? It's a pretty popular type to use.

Andarist:
> Are we sure we don't want to at least alias AnyState -> AnyMachineSnapshot? It's a pretty popular type to use.

I brought it back as a deprecated alias. I'm not sure if it's that popular though - we use it a lot internally but the sourcegraph search doesn't return a lot of results. However, I'm not sure how to properly make it an exact match. With an exact match maybe it could show better results since right now I get a lot of false positives matching `AnyStateMachine`

davidkpiano (approved):

PR #4449, opened 2023-11-10
Do not select always transitions when no regular transition gets selected

Andarist:
This essentially reverts https://github.com/statelyai/xstate/pull/2343 . I open this PR to spark the conversation about this behavior. It wasn't a change requested by any user - we just did it to comply with SCXML. However, this behavior "promotes" impurity and I really wonder if we want to keep this behavior.

### DIFF
--- a/packages/core/test/transient.test.ts
+++ b/packages/core/test/transient.test.ts
@@ -474,6 +474,29 @@ describe('transient states (eventless transitions)', () => {
     expect(actorRef.getSnapshot().value).toBe('a');
   });
 
+  it('should be taken if there is a wildcard transition', () => {
+    const machine = createMachine({
+      initial: 'a',
+      states: {
+        a: {
+          always: {
+            target: 'b',
+            guard: ({ event }) => event.type === 'WHATEVER'
+          },
+          on: {
+            '*': {}
+          }
+        },
+        b: {}
+      }
+    });
+    const actorRef = createActor(machine).start();
+
+    actorRef.send({ type: 'WHATEVER' });
+
+    expect(actorRef.getSnapshot().value).toBe('b');
+  });
+
   it('should select subsequent always transitions after selecting a regular transition', () => {
     let shouldMatch = false;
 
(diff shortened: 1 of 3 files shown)


davidkpiano:
What about the failing tests?

Andarist:
Those are directly related to this change. So if we decide to go with this change then I'll just adjust them before landing this.

davidkpiano (approved):

packages/core/test/transient.test.ts
```diff
@@ -614,32 +614,27 @@ describe('transient states (eventless transitions)', () => {
     expect(() => actorRef.start()).not.toThrow();
   });
 
-  it('should be taken even in absence of other transitions', () => {
-    let shouldMatch = false;
-
+  it('should not be taken even in absence of other transitions', () => {
```
davidkpiano, line 617: Confused about the wording, isn't it taken?

davidkpiano (approved):

packages/core/test/transient.test.ts
Andarist: The wording is correct when it comes to the goal of this - but the assertions in this test were still wrong (the CI was failing because of that). I just pushed fixes for those tests and removed some old duplicates

davidkpiano (commented):
Thinking about this

packages/core/test/transient.test.ts
```diff
@@ -255,63 +255,6 @@ describe('transient states (eventless transitions)', () => {
     expect(actorRef.getSnapshot().value).toEqual({ A: 'A4', B: 'B4' });
   });
 
-  it('should execute all eventless transitions in the same microstep (with `always`)', () => {
```
davidkpiano, line 258: What was wrong with these deleted tests?

Andarist: Those were copies of other test nearby. In the past we had 2 copies of those since one was using `''` and the other one was using `always` but now both were using the latter

Andarist:
@davidkpiano reopened this one like we discussed offline

davidkpiano (approved):

PR #4452, opened 2023-11-10
API Docs: fromPromise

audionerd:


packages/core/src/actors/promise.ts
```diff
@@ -38,14 +38,64 @@ export type PromiseActorRef<TOutput> = ActorRefFrom<
...
+ *   - `input` - Data that was provided to the promise actor
+ *   - `self` - The parent actor of the promise actor
+ *   - `system` - The actor system to which the promise actor belongs
+ * @see {@link https://stately.ai/docs/input | Input docs} for more information about how input is passed
+ *
+ * @example
+ * ```ts
+ * const promiseLogic = fromPromise(() => {
+ *   return fetch('https://example.com/...').then((data) => data.json());
+ * });
```
davidkpiano, line 61: Nit: should we make this an async/await example, as that may be more idiomatic?

```ts
 * const promiseLogic = fromPromise(async () => {
 *   const result = await fetch('https://example.com/...')
 *     .then((data) => data.json());
 * 
 *   return result;
 * });

davidkpiano (approved):

packages/core/src/actors/promise.ts
audionerd: Good call. I'll update.

davidkpiano (approved):

PR #4463, opened 2023-11-15
Require required inputs for invoked and spawned actors

Andarist:


packages/core/src/types.ts
```diff
@@ -622,21 +630,7 @@ type DistributeActors<
...
-            id: TSpecificActor['id'];
-          }
-        : {
-            /**
-             * The unique identifier for the invoked machine. If not specified, this
-             * will be the machine's own `id`, or the URL (from `src`).
-             */
-            id?: string;
-          })
+      } & { [K in RequiredActorOptions<TSpecificActor>]: unknown }
```
davidkpiano, line 633: Q: why `unknown`?

davidkpiano (approved):

packages/core/src/types.ts
Andarist: https://twitter.com/AndaristRake/status/1621075741470519296

PR #4464, opened 2023-11-15
Fixed registered `systemId`s availability in initial actions

Andarist:


packages/core/src/interpreter.ts
```diff
@@ -186,12 +186,17 @@ export class Actor<TLogic extends AnyActorLogic>
...
 
-    if (systemId && (this._state as any).status === 'active') {
+    if (systemId) {
       this._systemId = systemId;
       this.system._set(systemId, this);
     }
+
+    this._initState(options?.state);
+
+    if (systemId && (this._state as any).status !== 'active') {
```
davidkpiano, line 197: Nit: do we still need `any`?

packages/core/test/system.test.ts
```diff
@@ -510,4 +510,34 @@ describe('system', () => {
...
     ]);
   });
+
+  it('should be able to send an event to an ancestor with a registered `systemId` from an initial entry action', () => {
+    const spy = jest.fn();
+
+    const child = createMachine({
+      entry: sendTo(
+        ({ system }) => {
+          debugger;
```
davidkpiano, line 520:
```suggestion
```

packages/core/src/interpreter.ts
Andarist: Yes. The whole `SnapshotFrom<TLogic>` pattern (and that's how `this._state` is typed here) just makes it impossible to use *anything* on it. This type is a deferred conditional here since `TLogic` is generic and TS can't guarantee that it won't resolve to `never`... and we can't access anything on `never`. So it errors on any property access here since that's a situation that can happen from its PoV. I think that we shouldn't use this pattern in our signatures/classes.

packages/core/test/system.test.ts
Andarist: oopsies

davidkpiano (approved):

PR #4472, opened 2023-11-17
[v5] `state.configuration` -> `state._nodes`

davidkpiano:
The `state.configuration` property has been renamed to `state._nodes`.

```diff
- state.configuration
+ state._nodes
```


packages/core/src/State.ts
```diff
@@ -90,7 +90,7 @@ interface MachineSnapshotBase<
   /**
    * The enabled state nodes representative of the state value.
    */
-  configuration: Array<StateNode<TContext, TEvent>>;
+  nodes: Array<StateNode<TContext, TEvent>>;
```
Andarist, line 93: while we are at it I'd propose to rename this to `_nodes`

davidkpiano: Sure - go for it

packages/core/test/typegenTypes.test.ts
```diff
@@ -7,7 +7,7 @@ import {
 } from '../src/index.ts';
 import { fromPromise } from '../src/actors/index.ts';
 import { fromCallback } from '../src/actors/index.ts';
-import { createMachine } from '../src/Machine.ts';
+import { createMachine } from '../src/createMachine.ts';
```
Andarist, line 10: bold move

Andarist (approved):

packages/core/test/typegenTypes.test.ts
davidkpiano: Needed to be done

PR #4474, opened 2023-11-20
Rework setup inference and specifically guard inferences to aid higher order guards

Andarist:


packages/core/test/assign.test.ts
```diff
@@ -367,56 +367,3 @@ describe('assign meta', () => {
     service.send({ type: 'EVENT' });
   });
 });
-
-it('should type events from onDone properly', () => {
```
davidkpiano, line 371: Is this test no longer valid?

davidkpiano (approved):

packages/core/test/assign.test.ts
Andarist: it's valid - i just moved this to `setup.types.test.ts`

PR #4475, opened 2023-11-20
API Docs: fromObservable

audionerd:


packages/core/src/actors/observable.ts
```diff
@@ -26,6 +26,45 @@ export type ObservableActorRef<TContext> = ActorRefFrom<
...
+ * - Emit snapshots of the observable’s emitted value
+ *
+ * The observable’s emitted value is used as its observable actor’s `context`.
+ *
+ * Sending events to observable actors will have no effect.
+ *
+ * @param observableCreator A function that creates an observable. It receives one argument, an object with the following properties:
+ * - `input` - Data that was provided to the observable actor
+ * - `system` - The actor system to which the observable actor belongs
+ * - `self` - The parent actor
```
davidkpiano, line 41:
```suggestion
 * - `self` - The parent actor
 * - `system` - The actor system to which the observable actor belongs
```
Switching in order of importance

PR #4478, opened 2023-11-21
Remove `getNextEvents` method from the `MachineSnapshot`

Andarist:


davidkpiano (approved):
This can go in, will do a follow up for helper functions

Necmttn:
This turned out to be a breaking change as we had some logic built around this; What is the new helper function gonna be ? cc: @davidkpiano  
If it's gonna take a while, what is the way around getting similar results compatible with the upcoming helper function?

davidkpiano:
> This turned out to be a breaking change as we had some logic built around this; What is the new helper function gonna be ? cc: @davidkpiano If it's gonna take a while, what is the way around getting similar results compatible with the upcoming helper function?

Here is a workaround helper function that you can use:

```ts
function getNextEvents(snapshot: AnyMachineSnapshot) {
  return [...new Set([...snapshot._nodes.flatMap((sn) => sn.ownEvents)])];
}

// state.nextEvents;
getNextEvents(state);
```

derek-primumco:
Thanks for the workaround! We were also using `getNextEvents` for console logging.

davidkpiano:
> Thanks for the workaround! We were also using `getNextEvents` for console logging.

No problem! For others, this is also documented here: https://stately.ai/docs/migration#statenextevents-has-been-removed

Issue #4486, opened 2023-11-22
Bug: Nested machines / actors breaks restoring from persisted state


ShravanSunder:
### Description

If there are grand children state machines where the grandchild is using `fromPromise`. Then loading a state that we was persisted fails.  

## Reproduction
**Here is reproduction:**  https://codesandbox.io/p/sandbox/peaceful-glitter-2qr6qw

The reproduction has 3 nested machines and the last machine uses fromPromise.   It seems like the shape of the snapshot provided to `restoreState` changes after the 3rd machine.  

## Some code snippets
Here is the related xstate code
```ts
const actorData = snapshotChildren[actorId];
      const childState = actorData.state;
      const src = actorData.src;
      const logic = typeof src === 'string' ? guards_dist_xstateGuards.resolveReferencedActor(this, src) : src;
      if (!logic) {
        return;
      }
      const actorState = logic.restoreState?.(childState, _actorScope);
```

my code
```
	const preRunActor = fromPromise(async (actorInvoke): Promise<void> => {
		const jobActorRef = getJobActorRef(actorInvoke.self);
		match(params).with({ kind: P.union('hasSideEffects', 'withStepTools') }, () => {
			jobActorRef.send({ type: 'job.persistWorkflow' } satisfies FromJobInternalEvents);
		});
		await defer();
		jobActorRef.send({ type: 'job.toRun' } satisfies FromJobInternalEvents);
	});
```

This promise is a subchild of a machine.  It seems when the state is being restored, the logic is looking for a child state in the fromPromise actor.  the child state exists in runtime as `_state` in the child but not as `state`

### Expected result

I wanted clarification on why the restore would fail.  I would expect fromPromise woudln't have state and the restoration would continue

### Actual result

TypeError: Cannot read properties of undefined (reading 'children')

      at StateMachine2.restoreState

### Reproduction

attached code

ShravanSunder:
It seems that in my fromPromise runtime, actorData._state exists but actorData.state does not.  😖

Andarist:
Thanks for the report and the repro case. I think I already recognized the problem so you can expect the fix for this tomorrow.

ShravanSunder:
@Andarist thanks!  🥳

ShravanSunder:
@Andarist would it be possible to release this today?   Would help unblock me  :)

Andarist:
Done

PR #4491, opened 2023-11-23
Fixed an issue with actors deep in the tree failing to rehydrate

Andarist:
fixes https://github.com/statelyai/xstate/issues/4486

### DIFF
--- a/packages/core/src/StateMachine.ts
+++ b/packages/core/src/StateMachine.ts
@@ -560,13 +560,11 @@ export class StateMachine<
         return;
       }
 
-      const actorState = logic.restoreState?.(childState, _actorScope);
-
       const actorRef = createActor(logic, {
         id: actorId,
         parent: _actorScope?.self,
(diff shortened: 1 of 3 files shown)


packages/core/src/StateMachine.ts
```diff
@@ -560,13 +560,11 @@ export class StateMachine<
...
       }
 
-      const actorState = logic.restoreState?.(childState, _actorScope);
-
       const actorRef = createActor(logic, {
         id: actorId,
         parent: _actorScope?.self,
         syncSnapshot: actorData.syncSnapshot,
-        state: actorState,
+        state: childState,
```
Andarist, line 567: `restoreState` is already called within `createActor` ([here](https://github.com/statelyai/xstate/blob/1b6e26482344a7de2bf86e6f3a75340528500a52/packages/core/src/interpreter.ts#L211)) so we were calling it twice~ and that was creating problems

davidkpiano (approved):

PR #4492, opened 2023-11-24
Further errors-related adjustments

Andarist:


packages/core/src/interpreter.ts
```diff
@@ -224,19 +235,43 @@ export class Actor<TLogic extends AnyActorLogic>
...
+        for (const observer of this.observers) {
+          try {
+            observer.next?.(snapshot);
+          } catch (err) {
+            reportUnhandledError(err);
+          }
+        }
+        break;
       case 'done':
+        // it's disputable if `next` observers should be notified about done snapsots
```
Andarist, line 266: While somewhat disputable - I'm coming to the realization that maybe it's just better in our case. We don't necessarily have to maintain 1 to 1 semantics with RxJS and we can treat done snapshots as regular snapshots, just with a small addition of executing `complete` right after updating our internal state with one.

packages/core/src/interpreter.ts
```diff
@@ -515,6 +548,22 @@ export class Actor<TLogic extends AnyActorLogic>
...
+        this._parent,
+        createErrorActorEvent(this.id, err)
+      );
+    }
+  }
+  // TODO: atm children don't belong entirely to the actor so
+  // in a way - it's not even super aware of them
+  // so we can't stop them from here but we really should!
+  // right now, they are being stopped within the machine's transition
+  // but that could throw and leave us with "orphaned" active actors
```
Andarist, line 566: This is a fairly big concern. The core of the problem is outlined in this comment. Children's lifetime is somewhat related to the parent actor... but that parent actor alone doesn't really manage them. On top of that, `children` are exclusive to machines today.

To somewhat remedy that I could "duplicate" some of the try/catches within the `StateMachine` and update the snapshots and things there. We can't remove those try/catches from the `Actor` though. Other logic types might not be as nice and they might just throw. 

Alternatively, we could make an assumption that actor logics can't really throw and that the error handling is their responsibility... that's kinda a big assumption though (even if we'd document this)

packages/core/src/stateUtils.ts
```diff
@@ -1643,7 +1644,25 @@ export function macrostep(
...
-    const transitions = selectTransitions(nextEvent, nextState);
+    const currentEvent = nextEvent;
+    const isErr = isErrorActorEvent(currentEvent);
+
+    const transitions = selectTransitions(currentEvent, nextState);
+
+    if (isErr && !transitions.length) {
+      // TODO: we should likely only allow transitions selected by very explicit descriptors
+      // `*` shouldn't be matched, likely `xstate.error.*` shouldnt be either
+      // what about `xstate.error.actor.*`? what about `xstate.error.actor.todo.*`?
```
Andarist, line 1655: related to: https://github.com/statelyai/xstate/pull/2935#discussion_r781992122

packages/core/src/interpreter.ts
```diff
@@ -206,11 +206,22 @@ export class Actor<TLogic extends AnyActorLogic>
...
+    try {
+      this._state = persistedState
+        ? this.logic.restoreState
+          ? this.logic.restoreState(persistedState, this._actorScope)
+          : persistedState
+        : this.logic.getInitialState(this._actorScope, this.options?.input);
+    } catch (err) {
+      // TODO: if we get here then it means that we assign a value to this._state that is not of the correct type
+      // we can't get the true `TSnapshot & { status: 'error'; }`, it's impossible
+      // should the error snapshot's type to always be a minimalistic one to reflect the reality?
```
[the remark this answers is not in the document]
davidkpiano: I think some sync errors are desirable too.

packages/core/test/actions.test.ts
```diff
@@ -3185,7 +3185,8 @@ describe('sendTo', () => {
     service.send({ type: 'EVENT', value: 'foo' });
   });
 
-  it('should throw if given a string', () => {
+  // TODO: figure out how to best test this
+  it.skip('should throw if given a string', () => {
```
davidkpiano, line 3189: Why skipped? This should synchronously error

packages/core/test/actions.test.ts
```diff
@@ -3410,7 +3411,8 @@ describe('raise', () => {
     }, 10);
   });
 
-  it('should throw if given a string', () => {
+  // TODO: figure out how to best test this
+  it.skip('should throw if given a string', () => {
```
davidkpiano, line 3415: Same here - this should sync error

davidkpiano (commented):
This LGTM - just need to adjust (un-skip + refactor) the tests

.changeset/olive-months-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': major
+---
+
+All errors caught while executing the actor should now consistently error its snapshot and should be reported to the outermost `error` listener if they stay unhandled.
```
davidkpiano, line 5:
```suggestion
All errors caught while executing the actor should now consistently include the error in its `snapshot.error` and should be reported to the closest `error` listener.
```

PR #4497, opened 2023-11-26
Scheduler

davidkpiano:


packages/core/src/actions/raise.ts
```diff
@@ -85,7 +85,8 @@ function executeRaise(
   }
 ) {
   if (typeof params.delay === 'number') {
```
Andarist, line 87: I'd consider wrapping this in `defer` (ain't sure yet though)

packages/core/src/interpreter.ts
```diff
@@ -150,7 +142,12 @@ export class Actor<TLogic extends AnyActorLogic>
     const { clock, logger, parent, syncSnapshot, id, systemId, inspect } =
       resolvedOptions;
 
-    this.system = parent?.system ?? createSystem(this);
+    this.system =
+      parent?.system ??
+      createSystem(this, {
+        clock,
+        snapshot: (options as any)?.snapshot?.['xstate.system']
```
Andarist, line 149: for the root we probably should "split" the `options.snapshot` so `restoreSnapshot` of the root logic doesn't receive this `xstate.system` key there. We need to contain all of this attaching/detaching of the `xstate.system` magical thing in the Actor

packages/core/src/scheduler.ts
```diff
@@ -0,0 +1,113 @@
...
+      to?: AnyActorRef;
+    }
+  ): void;
+  cancel(id: string): void;
+  cancelAll(actorRef: AnyActorRef): void;
+  getPersistedSnapshot(): Record<string, ScheduledEvent>;
+  start(): void;
+}
+
+const defaultClock: Clock = {
```
Andarist, line 35: this is unused currently if I read it right

packages/core/src/scheduler.ts
```diff
@@ -0,0 +1,113 @@
...
+      source: AnyActorRef,
+      data: {
+        id: string | undefined;
+        event: EventObject;
+        delay: number;
+        to: AnyActorRef;
+      }
+    ) => {
+      // TODO: `id` has to be separated from `data.id` completely
+      // `data.id` is supposed to be unique within an actor, `id` is supposed to be unique globally
```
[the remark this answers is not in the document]
Andarist: we need tests for this :P

packages/core/src/scheduler.ts
```diff
@@ -0,0 +1,113 @@
...
+        id,
+        source,
+        target: data.to || source,
+        startedAt: Date.now()
+      };
+      scheduler.events[id] = scheduledEvent;
+
+      const timeout = clock.setTimeout(() => {
+        const target = data.to || source;
+        // TODO: explain this hack, it should also happen sooner, not within this timeout
```
Andarist, line 76: this likely is fixable (but I didn't check it) by deferring `scheudle`+`cancel` in our machine actions

packages/core/src/scheduler.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+      }
+    },
+    getPersistedSnapshot: () => {
+      return { ...scheduler.events };
+    },
+    start: () => {
+      for (const id in scheduler.events) {
+        const scheduledEvent = scheduler.events[id];
+        scheduler.schedule(scheduledEvent.source, scheduledEvent);
```
Andarist, line 108: TODO: we should diff the `now - createdAt +-/ delayThing`

Andarist: an option for the behavior? but which one is the default?

packages/core/src/interpreter.ts
```diff
@@ -695,7 +660,15 @@ export class Actor<TLogic extends AnyActorLogic>
...
   public getPersistedSnapshot(options?: unknown): Snapshot<unknown> {
-    return this.logic.getPersistedSnapshot(this._snapshot, options);
+    const persisted = this.logic.getPersistedSnapshot(this._snapshot, options);
+    if (!this._parent) {
+      const rootPersistedSnapshot = {
+        ...persisted,
+        'xstate.system': this.system.getPersistedSnapshot()
+      };
+      return rootPersistedSnapshot;
+    }
```
Andarist, line 670: I'd temporarily disable this part of the code to figure out system-wide rehydration of actor references.

packages/core/src/interpreter.ts
davidkpiano: This part is removed

packages/core/src/interpreter.ts
davidkpiano: This part is removed

packages/core/src/system.ts
```diff
@@ -1,22 +1,109 @@
...
+   */
+  _relay: (
+    source: AnyActorRef | undefined,
+    target: AnyActorRef,
+    event: AnyEventObject
+  ) => void;
+  scheduler: Scheduler;
+  getSnapshot: () => {
+    _scheduledEvents: Record<string, ScheduledEvent>;
+  };
```
Andarist, line 75: we'll likely need `getPersistedSnapshot` here too - but that can be added when we manage to implement serialization/deserialization of actor refs

packages/core/src/system.ts
```diff
@@ -1,22 +1,109 @@
...
+
+export type AnyActorSystem = ActorSystem<any>;
+
 let idCounter = 0;
 export function createSystem<T extends ActorSystemInfo>(
-  rootActor: AnyActorRef
+  rootActor: AnyActorRef,
+  options: {
+    clock: Clock;
+    snapshot?: unknown;
```
Andarist, line 92: this stays unused but we will need this in the future so I'm OK with keeping it here

packages/core/src/system.ts
```diff
@@ -1,22 +1,109 @@
...
+  /**
+   * @internal
+   */
+  _relay: (
+    source: AnyActorRef | undefined,
+    target: AnyActorRef,
+    event: AnyEventObject
+  ) => void;
+  scheduler: Scheduler;
+  getSnapshot: () => {
```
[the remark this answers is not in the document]
davidkpiano: We'll figure it out later; it's for our internal use only for now

packages/core/src/scheduler.ts
Andarist: done

packages/core/src/scheduler.ts
Andarist: we'll be doing the actual timers persistence later

packages/core/src/system.ts
```diff
@@ -1,22 +1,182 @@
...
+  }
+
+  return systemPath;
+}
+
+function createScheduledEventId(
+  actorRef: AnyActorRef,
+  id: string
+): ScheduledEventId {
+  return `${getSystemPath(actorRef)}.${id}` as ScheduledEventId;
```
Andarist, line 58: Couldn't this be simplified like this?
```suggestion
  return `${actorRef.sessionId}.${id}` as ScheduledEventId;
```

Andarist (approved):

PR #4498, opened 2023-11-26
Strongly-typed `state.value` and `state.matches`

davidkpiano:
This only works with machines from `setup(...).createMachine(...)`


packages/core/src/State.ts
```diff
@@ -39,7 +41,8 @@ interface MachineSnapshotBase<
   TChildren extends Record<string, AnyActorRef | undefined>,
   TTag extends string,
   TOutput,
-  TResolvedTypesMeta = TypegenDisabled
+  TResolvedTypesMeta = TypegenDisabled,
+  TConfig extends StateConfig2 = any
```
Andarist, line 45: this shouldn't receive the whole `TConfig` but rather the `TStateValue` - in other words, we should precompute this and pass the resolved value here

packages/core/src/types.ts
```diff
@@ -2368,3 +2354,40 @@ export type ToChildren<TActor extends ProvidedActor> =
...
+    ?
+        | keyof TConfig['states']
+        | {
+            [K in keyof TConfig['states']]?: StateValueFromConfig<
+              TConfig['states'][K]
+            >;
+          }
+    : never;
+
+export type StateValueFrom2<T extends StateConfig2> = IsAny<T> extends true
```
Andarist, line 2375: it would be *very* nice to rewrite this as a tail-recursive conditional type - but we can optimize this later on

packages/core/src/types.ts
```diff
@@ -2368,3 +2354,40 @@ export type ToChildren<TActor extends ProvidedActor> =
...
               : 'exclude']
       >;
+
+export type StateConfig2 = {
+  states?: {
+    [K: string]: StateConfig2;
+  };
+};
+
+export type StateValueFromConfig<TConfig extends StateConfig2> =
```
Andarist, line 2364: this stays unused so it can probably be removed

packages/core/src/StateMachine.ts
```diff
@@ -83,7 +84,8 @@ export class StateMachine<
     TGuard,
     TDelay,
     TTag
-  >
+  >,
+  TConfig extends StateConfig2 = any
```
Andarist, line 88: We can't infer it here at the moment so it shouldn't even be added to the type parameters list. If we stop passing `TConfig` down (like suggested in another comment) then we could just pass down directly `StateConfig2` (or similar) as `TStateValue`.

packages/core/src/State.ts
```diff
@@ -235,39 +247,44 @@ export type MachineSnapshot<
...
```
Andarist, line 251: Please keep `TResolvedTypesMeta` as the last type parameter in the list

packages/core/src/types.ts
```diff
@@ -2173,25 +2173,11 @@ export type SnapshotFrom<T> = ReturnTypeOrValue<T> extends infer R
     ? TSnapshot
     : R extends Actor<infer TLogic>
       ? SnapshotFrom<TLogic>
-      : R extends StateMachine<
```
Andarist, line 2176: There was a reason why this had two separate branches for `StateMachine` and `ActorLogic`, although I don't recall what that was right now. I'd probably prefer to leave this change for another PR

davidkpiano (commented):
LGTM, merge when ready

PR #4499, opened 2023-11-26
Fixed `ToProvidedActor` helper type

davidkpiano:


packages/core/test/types.test.ts
```diff
@@ -4413,4 +4413,27 @@ describe('self', () => {
...
+  it('should work with this case that was not working before', () => {
+    function init() {
+      return { foo: 'bar' };
+    }
+
+    setup({}).createMachine({
+      types: {
+        events: {} as { type: 'RESTART' },
+        context: {} as ReturnType<typeof init>
+      },
```
Andarist, line 4426: When using `setup` you are supposed to pass `.types` within `setup` and not within `createMachine`

Andarist: That said, it doesn't work when I refactor it this way either. I'll investigate this soon.

PR #4509, opened 2023-11-28
Refactor callback logic to not send self-event

davidkpiano:


packages/core/src/actors/callback.ts
```diff
@@ -134,36 +134,25 @@ export function fromCallback<TEvent extends EventObject, TInput = unknown>(
...
-          sendBack,
-          receive
-        });
-
-        return state;
-      }
-
+        },
+        receive: (listener) => {
+          (state._receivers ?? (state._receivers = new Set())).add(listener);
```
davidkpiano, line 151: Bonus: receivers `Set()` is now created lazily

packages/core/src/actors/observable.ts
```diff
@@ -209,6 +209,9 @@ export function fromObservable<TContext, TInput>(
  * ```
  */
 
+const errorEventType = '$$xstate.error';
+const completeEventType = '$$xstate.complete';
```
Andarist, line 213: We can also deduplicate it with the same constants that are defined within `fromObservable`. I would also drop `$$` prefix from all of them

packages/core/src/actors/callback.ts
```diff
@@ -134,36 +134,25 @@ export function fromCallback<TEvent extends EventObject, TInput = unknown>(
 ): CallbackActorLogic<TEvent, TInput> {
   const logic: CallbackActorLogic<TEvent, TInput> = {
     config: invokeCallback,
-    start: (_state, { self, system }) => {
-      system._relay(self, self, { type: 'xstate.create' });
```
Andarist, line 138: The previous version was somewhat more correct. At the moment, `start` should not be able to modify the snapshot. `transition` should also do it in an immutable way and even the previous version didn't do it "correctly" - but at least the mutation was located in a more appropriate place (it was easier to refactor it towards the immutable version than it is with the proposed change)

packages/core/src/actors/callback.ts
```diff
@@ -134,36 +134,25 @@ export function fromCallback<TEvent extends EventObject, TInput = unknown>(
...
-      if (event.type === 'xstate.create') {
-        const sendBack = (eventForParent: AnyEventObject) => {
-          if (state.status === 'stopped') {
+    start: (state, { self, system }) => {
+      state._dispose = invokeCallback({
+        input: state.input,
+        system,
+        self,
+        sendBack: (event) => {
+          if (self.getSnapshot().status === 'stopped') {
```
Andarist, line 143: This line looks like an actual bug fix - it looks like we were using stale closure data here before. A test case for this should be added

Andarist (approved):

PR #4510, opened 2023-11-28
Merge `createActor` overloads into one

Andarist:
fixes https://github.com/statelyai/xstate/pull/4416/files#r1407630794

Before the final release, we might consider removing all those `AreAllImplementationsAssumedToBeProvided` code paths. For the time being though this is kinda a bug so let's fix it and then bikeshed that removal.


davidkpiano (approved):

davidkpiano:
> Before the final release, we might consider removing all those `AreAllImplementationsAssumedToBeProvided` code paths

Let's do this

Issue #4496, opened 2023-11-25
Bug: [V5] Recommended way to set higher-level logic breake actor


christophe-g:
### Description

The recommended way to set higher-level logic creates a `TypeError: this.logic.getPersistedState is not a function` downstream. 

```js
function withLogging(actorLogic) {
  const enhancedLogic = {
    ...actorLogic,
    transition: (state, event, actorCtx) => {
      console.log('State:', state);
      return actorLogic.transition(state, event, actorCtx);
    },
  };
  return enhancedLogic;
}

const counterActor = createActor(withLogging(counterMachine));
```

### Expected result

Actor system work with recommended way to add higher level logic.

### Actual result

Type Error when calling `actor.getPersistedState`

### Reproduction

https://lit.dev/playground/#sample=examples%2Fhello-world&gist=2c690b9eabd345221116ad5722d78a5f

### Additional context

Workaround: 

```js
function withLogging(actorLogic) {
  const transition = actorLogic.transition
  const enhancedLogic = {
    transition: (state, event, actorCtx) => {
      console.log('State:', state);
      return transition(state, event, actorCtx);
    },
  };
  return enhancedLogic;
}

```

Andarist:
Hmm, this is because this particular logic (`StateMachine`) is implemented using a class. When you spread it... its prototype methods are not spread onto the new object.

christophe-g:
> Hmm, this is because this particular logic (StateMachine) is implemented using a class

Yes, I agree it is not a big deal - it is just that following the current leads to this issue.

PR #4513, opened 2023-11-29
Fix higher-level logic for machines

davidkpiano:


packages/core/test/actorLogic.test.ts
```diff
@@ -971,4 +972,41 @@ describe('composable actor logic', () => {
...
+          on: { more: 'done' }
+        },
+        done: {}
+      }
+    });
+
+    const actor = createActor(withLogging(machine)).start();
+
+    actor.send({ type: 'next' });
+    actor.send({ type: 'more' });
```
christophe-g, line 1006: In #4496, the actor breaks on calling `getPersistedSnaphot`. This is what should be tested in addition.

Andarist: As a quick/temporary fix, we could just `.bind` this method along with the other ones that we already bind:
https://github.com/statelyai/xstate/blob/eef14a2d558a00ad75558fb44432601608690890/packages/core/src/StateMachine.ts#L145-L148

davidkpiano: Just did both of these thing; tests pass

packages/core/test/actorLogic.test.ts
```diff
@@ -982,4 +983,45 @@ describe('composable actor logic', () => {
       }
     });
   });
+
+  it('higher-level logic should work', () => {
```
Andarist, line 987: this fix is specific to machines so I'd include "machine" in  the test title
```suggestion
  it('higher-level logic wrapping a machine should be able to persist a snapshot', () => {
```

packages/core/test/actorLogic.test.ts
```diff
@@ -982,4 +983,45 @@ describe('composable actor logic', () => {
...
+    actor.send({ type: 'next' });
+    actor.send({ type: 'more' });
+
+    expect(logged).toEqual(['next', 'more']);
+
+    expect(actor.getSnapshot().value).toBe('done');
+
+    expect(() => {
+      actor.getPersistedSnapshot();
+    }).not.toThrow();
```
Andarist, line 1025: it might be worth testing the output of `getPersistedSnapshot` here. If it would return `undefined` or smth else (for whatever reason) then this test would still pass

packages/core/test/actorLogic.test.ts
christophe-g: Thanks a lot for all your hard (and high quality) work !

Andarist (approved):

PR #4521, opened 2023-11-29
Edit after events

davidkpiano:
The event type of internal `after` events changed from `xstate.after(1000)#some.state.id` to `xstate.after.1000.#some.state.id` for consistency.


packages/core/src/eventUtils.ts
```diff
@@ -10,7 +10,7 @@ import { DoneActorEvent, DoneStateEvent, ErrorActorEvent } from './types.ts';
  */
 export function createAfterEvent(delayRef: number | string, id?: string) {
   const idSuffix = id ? `#${id}` : '';
-  return { type: `xstate.after(${delayRef})${idSuffix}` } as const;
+  return { type: `xstate.after.${delayRef}.${idSuffix}` } as const;
```
Andarist, line 13: `id` is always provided from what I can see, let's make that argument required and drop this conditional expression (just inline the `id` directly in the `type`)

.changeset/weak-swans-care.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+The event type of internal `after` events changed from `xstate.after(1000)#some.state.id` to `xstate.after.1000.#some.state.id` for consistency.
```
Andarist, line 5: One cool thing about the previous format was that "segments" were delimited in a somewhat better way. Theoretically one could define this:
```ts
setup({
  delays: {
    'wow.this.is.a.delay': () => 100
  }
})
```

With smth like this, we'll have a lot of segments delimited just by dots and it might be harder to tell which exact segments are coming from the `delayRef`. It's not exactly impossible using the proposed format but the previous format was more intuitive in this regard. OTOH, one could also define a stupid `')FU'` delay and mess up the matcher that would try to look between the opening and closing parenthesis. So perhaps it's not a big deal after all? Hm...

Please also note that even with the proposed adjustment we still can have some "special characters" in those generated event types:
https://github.com/statelyai/xstate/blob/63d923857592437dc174518ba02e061082f629cf/packages/core/src/stateUtils.ts#L295-297

It feels like this PR should adjust both or leave both alone.

davidkpiano: The small edge-case benefit of `xstate.after.*` is the partial event descriptor, for very advanced use-cases. The previous syntax was just me not knowing what I was doing.

I'll adjust our other magic IDs.

packages/core/src/utils.ts
```diff
@@ -381,13 +381,13 @@ export function toObserver<T>(
...
 export function createInvokeId(stateNodeId: string, index: number): string {
-  return `${stateNodeId}[${index}]`;
+  return `${index}.#${stateNodeId}`;
 }
 
 export function resolveReferencedActor(machine: AnyStateMachine, src: string) {
-  if (src.startsWith('xstate#')) {
-    const [, indexStr] = src.match(/\[(\d+)\]$/)!;
-    const node = machine.getStateNodeById(src.slice(7, -(indexStr.length + 2)));
+  if (src.startsWith('xstate.source')) {
```
Andarist, line 388: I'd expect this to be enough already:
```suggestion
  if (src.startsWith('xstate.')) {
```

Andarist: kinda addressed this in a different way in https://github.com/statelyai/xstate/pull/4524

packages/core/src/eventUtils.ts
```diff
@@ -8,9 +8,8 @@ import { DoneActorEvent, DoneStateEvent, ErrorActorEvent } from './types.ts';
  * @param delayRef The delay in milliseconds
  * @param id The state node ID where this event is handled
  */
-export function createAfterEvent(delayRef: number | string, id?: string) {
-  const idSuffix = id ? `#${id}` : '';
-  return { type: `xstate.after(${delayRef})${idSuffix}` } as const;
+export function createAfterEvent(delayRef: number | string, id: string) {
+  return { type: `xstate.after.${delayRef}.#${id}` } as const;
```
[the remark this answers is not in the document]
davidkpiano: Sure; also in general, we just have to make sure that any `.` in places like `delayRef` are escaped and account for those; that's an edge-case though.

Andarist (approved):

.changeset/weak-swans-care.md
Andarist, line 5:
```suggestion
The event type of internal `after` events changed from `xstate.after(1000)#some.state.id` to `xstate.after.1000.some.state.id` for consistency.
```

PR #4524, opened 2023-11-30
Adjust after event types and invoke IDs further

Andarist:


packages/core/src/StateNode.ts
```diff
@@ -305,7 +305,7 @@ export class StateNode<
         const resolvedSrc =
           typeof src === 'string'
             ? src
-            : `xstate.source.${createInvokeId(this.id, i)}`;
+            : `xstate.invoke.${createInvokeId(this.id, i)}`;
```
Andarist, line 308: We might want to do a similar thing to `spawnChild` action in the future so I think it makes sense to use `invoke` here and not a generic `source`. This might become important for `resolveReferencedActor`

packages/core/src/utils.ts
```diff
@@ -381,21 +381,22 @@ export function toObserver<T>(
 }
 
 export function createInvokeId(stateNodeId: string, index: number): string {
-  return `${index}.#${stateNodeId}`;
+  return `${index}.${stateNodeId}`;
```
Andarist, line 384: feel free to revert this part of the change. It just felt that `#` is also a special character that we don't necessarily have to include here.

davidkpiano (approved):

PR #4547, opened 2023-12-03
Add `assertEvent`

davidkpiano:
Add `assertEvent(...)` to help provide strong typings for events that can't be easily inferred, such as events in `entry` and `exit` actions, or in `invoke.input`.

The `assertEvent(event, 'someType')` function will _throw_ if the event is not the expected type. This ensures that the `event` is guaranteed to have that type, and assumes that the event object has the expected payload (naturally enforced by TypeScript, as long as you don't do anything silly).

```ts
// ...
entry: ({ event }) => {
  assertEvent(event, 'greet');
  // event is { type: 'greet'; message: string }

  assertEvent(event, ['greet', 'notify']);
  // event is { type: 'greet'; message: string }
  // or { type: 'notify'; message: string; level: 'info' | 'error' }
},
exit: ({ event }) => {
  assertEvent(event, 'doNothing');
  // event is { type: 'doNothing' }
}
```


.changeset/cuddly-taxis-walk.md
```diff
@@ -0,0 +1,23 @@
+---
+'xstate': patch
+---
+
+Add `assertEvent(...)` to help provide strong typings for events that can't be easily inferred, such as events in `entry` and `exit` actions, or in `invoke.input`.
+
+The `assertEvent(event, 'someType')` function will _throw_ if the event is not the expected type. This ensures that the `event` is guaranteed to have that type, and assumes that the event object has the expected payload (naturally enforced by TypeScript, as long as you don't do anything silly).
```
Andarist, line 7:
```suggestion
The `assertEvent(event, 'someType')` function will _throw_ if the event is not the expected type. This ensures that the `event` is guaranteed to have that type, and assumes that the event object has the expected payload (naturally enforced by TypeScript).
```

packages/core/src/assert.ts
```diff
@@ -0,0 +1,42 @@
...
+    // or { type: 'notify'; message: string; level: 'info' | 'error' }
+  },
+  exit: ({ event }) => {
+    assertEvent(event, 'doNothing');
+    // event is { type: 'doNothing' }
+  }
+  ```
+ */
+export function assertEvent<
+  TEvent extends AnyEventObject,
```
Andarist, line 26:
```suggestion
  TEvent extends EventObject,
```

packages/core/src/assert.ts
```diff
@@ -0,0 +1,42 @@
...
+  TSpecificType extends TEvent['type']
```
Andarist, line 27: Maybe just `TType` or `TAssertedType`?

packages/core/src/assert.ts
```diff
@@ -0,0 +1,42 @@
...
+>(
+  eventObject: TEvent,
```
Andarist, line 29: usually we call this just an `event`

packages/core/src/assert.ts
```diff
@@ -0,0 +1,42 @@
...
+    // event is { type: 'greet'; message: string }
+
+    assertEvent(event, ['greet', 'notify']);
+    // event is { type: 'greet'; message: string }
+    // or { type: 'notify'; message: string; level: 'info' | 'error' }
+  },
+  exit: ({ event }) => {
+    assertEvent(event, 'doNothing');
+    // event is { type: 'doNothing' }
+  }
```
Andarist, line 22: I'd swap those examples to have the shorter one (easier one) first.

.changeset/cuddly-taxis-walk.md
```diff
@@ -0,0 +1,23 @@
+---
+'xstate': patch
```
Andarist, line 2:
```suggestion
'xstate': minor
```

packages/core/src/assert.ts
```diff
@@ -0,0 +1,42 @@
+import { AnyEventObject } from '.';
```
Andarist, line 1:
```suggestion
import { AnyEventObject } from './types.ts';
```

packages/core/src/assert.ts
```diff
@@ -0,0 +1,42 @@
...
+    // event is { type: 'greet'; message: string }
+    // or { type: 'notify'; message: string; level: 'info' | 'error' }
+  },
+  exit: ({ event }) => {
+    assertEvent(event, 'doNothing');
+    // event is { type: 'doNothing' }
+  }
+  ```
+ */
+export function assertEvent<
```
Andarist, line 25: we need to re-export this from the `src/index.ts`

packages/core/test/assertionHelpers.test.ts
```diff
@@ -0,0 +1,101 @@
...
+          `[Error: Expected event {"type":"count","value":42} to have one of types "greet", "notify"]`
+        );
+        done();
+      }
+    });
+
+    actor.start();
+
+    actor.send({ type: 'count', value: 42 });
+  });
```
Andarist, line 100: What do we want to do about situations like this:
```ts
it('TODO: test title', () => {
  setup({
    types: {} as {
      events: {
        type: 'FETCH';
      };
    },
    actors: {
      fetchRandom: fromPromise(() => Promise.resolve(Math.random())),
      replicate: fromPromise(({ input }: { input: number }) =>
        Promise.resolve([input, input])
      )
    }
  }).createMachine({
    initial: 'idle',
    states: {
      idle: {
        on: { FETCH: 'step1' }
      },
      step1: {
        invoke: {
          src: 'fetchRandom',
          onDone: 'step2'
        }
      },
      step2: {
        invoke: {
          src: 'replicate',
          input: ({ event }) => {
            assertEvent(event, 'done.invoke.??????');
            return event.output;
          }
        }
      }
    }
  });
});
```

davidkpiano: We can use the same types from the partial event descriptors so that devs can use:
```ts
assertEvent(event, 'done.invoke.*')
```
But let's save that for another PR. If the ID is known, `'done.invoke.theId'` will suffice (and it encourages them to use IDs).

Andarist: Even with that, the example above won't work. We still need the "all events" type - smth like [this](https://github.com/statelyai/xstate/commit/4f80aea97a124b04b5d0517649085384b6ef6e13) (it breaks a lot of stuff right now though).

Andarist (approved):

Issue #3925, opened 2023-03-27
Bug: @xstate/solid: Incorrect `after` event is firing


exclipy:
### Description

I am observing that an "after" event being fired, for a state that is not the current state. The "after" event that *should* be fired for the current state is lost.

Go here: https://codesandbox.io/s/silly-margulis-fbmzd8?file=/src/index.js
Open the console, which displays the states and transitions as they occur.

### Expected result

The final state should be `NotHeld.Idle` because there is an transition to go from `NotHeld.TripleClicked` to `NotHeld.Idle` after 400ms.

### Actual result

It gets stuck in `NotHeld.TripleClicked` state.

Observe an event `xstate.after(100)#stepper.Held.HeldIncrement` but this is ignored because HeldIncrement is not the current state.

### Reproduction

https://codesandbox.io/s/silly-margulis-fbmzd8?file=/src/index.js

### Additional context

The bug does not appear if I use `interpret` from `xstate` instead of `useMachine` from `@xstate/solid`

Andarist:
Unfortunately, `state` holds some data structures that are created internally by XState (such as `state.actions`) and when the Solid store is updated... it accidentally proxify/reconcile them and later on mutates them which has a very negative effect on our internals.

A dirty hack to work around this could look something like that:
```diff
diff --git a/packages/xstate-solid/src/useMachine.ts b/packages/xstate-solid/src/useMachine.ts
index 2afde4e9d..83cdbe485 100644
--- a/packages/xstate-solid/src/useMachine.ts
+++ b/packages/xstate-solid/src/useMachine.ts
@@ -40,7 +40,16 @@ export function useMachine<TMachine extends AnyStateMachine>(
   onMount(() => {
     const { unsubscribe } = service.subscribe((nextState) => {
       setState(
-        deriveServiceState(nextState, unwrap(state)) as StateFrom<TMachine>
+        deriveServiceState(
+          { ...nextState, actions: undefined },
+          unwrap(state)
+        ) as StateFrom<TMachine>
+      );
+      setState(
+        deriveServiceState(
+          { ...nextState, actions: nextState.actions },
+          unwrap(state)
+        ) as StateFrom<TMachine>
       );
     });
```

[...]

PR #4567, opened 2023-12-07
[@xstate/solid] Fix full array replacement bug

### LINKED ISSUE #4369: Bug: @xstate/solid not updating arrays properly (createImmutable.ts)
### Description

First off thank you for an amazing open sourced library. I know the title is a duplicate of bug #3853 . That was done intentionally as it seems the fix for the issue did work however it only works if the arrays are different sizes or the length is divisible by two. 

### Expected result

Given these two arrays as an example
```javascript
const next = [4,3,2,1,0]
const prev = [1,2,3,4,5]
```
The output should be `[4,3,2,1,0]`

### Actual result

However the current output is 
```javascript
[4,3,3,1,0]
```

 The issue lies in the for loop here 

https://github.com/statelyai/xstate/blob/98f9f9165df328bba68622b69ec46dada0e3a4f7/packages/xstate-solid/src/createImmutable.ts#L56-L59C8

```typescript
      // Diff array
      for (let start = 0, end = largestSize - 1; start < end; start++, end--) {
        diff(next[start], prev[start], [...path, start] as Path);
        diff(next[end], prev[end], [...path, end] as Path);
      }
```

In this loop both `start` and `end` are incremented / decremented on each iteration.  Because of the use of th `<` operator the start and end will miss the center value in an odd length array.  

## Example
```typescript
const next = [4,3,2,1,0]
const prev = [1,2,3,4,5]
for (let start = 0, end = next.length - 1; start < end; start++, end--) {
  console.log(`start: ${start}, next[start]: ${next[start]}, prev[start]: ${prev[start]}`)
  console.log(`end: ${end}, next[end]: ${next[end]}, prev[end]: ${prev[end]}\n`)
  if (next[start] !== prev[start]) prev[start] = next[start]
  if (next[end] !== prev[end]) prev[end] = next[end]
}
console.log(`updated: [${prev}]`)
```

The above code will output the following

```bash
start: 0, next[start]: 4, prev[start]: 1
end: 4, next[end]: 0, prev[end]: 5

start: 1, next[start]: 3, prev[start]: 2
end: 3, next[end]: 1, prev[end]: 4

updated: [4,3,3,1,0]
```

As you can see because `start` and `end` don't converge the 3rd element in the `next` array doesn't get assigned to the updated state. 

One solution would be to change `start < end` to `start < end + 1` then add a conditional equality check between the two diff operations for `start === end` and break the loop.

The other is obviously using a traditional unidirectional for loop. So I thought I would compare the approaches and see which one preformed better.  

```javascript
import { Bench } from 'tinybench';
const bench = new Bench();

function diff1(prev, next) {
  const state = [...prev];
  for (let start = 0, end = next.length - 1; start < end; start++, end--) {
    if (next[start] !== prev[start]) state[start] = next[start]
    if (next[end] !== prev[end]) state[end] = next[end]
  }
  return state;
}

function diff2(prev, next) {
  const state = [...prev];
  for (let start = 0, end = next.length; start < end; start++) {
    if (next[start] !== prev[start]) state[start] = next[start]
  }
  return state;
}

function diff3(prev, next) {
  const state = [...prev];
  for (let start = 0, end = next.length - 1; start < end + 1; start++, end--) {
    if (next[start] !== prev[start]) state[start] = next[start]
    if (start == end) break;
    if (next[end] !== prev[end]) state[end] = next[end]
  }
  return state;
}

function randomPrevAndNext() {
  return [Array.from({length: 5}, () => Math.floor(Math.random() * 100)), Array.from({length: 5}, () => Math.floor(Math.random() * 100))]
}


bench.add("diff fixed multidirection", diff3.bind(null, ...randomPrevAndNext()))
bench.add("diff multidirection broken", diff1.bind(null, ...randomPrevAndNext()))
bench.add("diff unidirection", diff2.bind(null, ...randomPrevAndNext()))
await bench.run();
console.table(bench.table());
```

I used a randomly generated array of length 5 to eliminate any caching or JIT optimizations that would have skewed the results. 

```bash
┌─────────┬──────────────────────────────┬──────────────┬───────────────────┬──────────┬─────────┐
│ (index) │          Task Name           │   ops/sec    │ Average Time (ns) │  Margin  │ Samples │
├─────────┼──────────────────────────────┼──────────────┼───────────────────┼──────────┼─────────┤
│    0    │ 'diff fixed multidirection'  │ '13,478,123' │ 74.19430570991092 │ '±1.52%' │ 6739062 │
│    1    │ 'diff multidirection broken' │ '13,588,260' │ 73.59293492150188 │ '±0.84%' │ 6794131 │
│    2    │     'diff unidirection'      │ '13,700,434' │ 72.99038840504755 │ '±0.94%' │ 6850218 │
└─────────┴──────────────────────────────┴──────────────┴───────────────────┴──────────┴─────────┘
```

I think 0 & 2 are basically the same in performance as running this multiple times will flip flop the winner randomly. Personally I would go with the unidirectional comparison as it's less code. 

### Reproduction

Not necessary

### Additional context

_No response_

GoldingAustin:
- Additional iteration on store array diffs to ensure middle value is checked
- Update package.json to match script and export pattern of other packages
Fixes #4369 & #3853

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,7 +9,6 @@
     "@xstate/immer",
     "@xstate/graph",
     "@xstate/inspect",
-    "@xstate/solid",
     "@xstate/test"
   ],
   "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
--- a/.changeset/yellow-wasps-switch.md
+++ b/.changeset/yellow-wasps-switch.md
@@ -1,0 +1,5 @@
+---
+'@xstate/solid': patch
+---
+
+Additional iteration on store array diffs to ensure middle value is checked
--- a/packages/xstate-solid/package.json
+++ b/packages/xstate-solid/package.json
@@ -30,7 +30,6 @@
     },
     "./package.json": "./package.json"
   },
-  "types": "lib/index.d.ts",
   "files": [
     "dist"
   ],
@@ -38,12 +37,7 @@
     "type": "git",
     "url": "git+ssh://git@github.com/statelyai/xstate.git"
   },
-  "scripts": {
-    "clean": "rm -rf dist lib es tsconfig.tsbuildinfo",
-    "build": "tsc && tsc --outDir es --module es2015 && rollup -c",
-    "test": "jest",
-    "prepare": "npm run build"
-  },
+  "scripts": {},
   "bugs": {
     "url": "https://github.com/statelyai/xstate/issues"
   },
--- a/packages/xstate-solid/src/createImmutable.ts
+++ b/packages/xstate-solid/src/createImmutable.ts
@@ -53,8 +53,9 @@ const updateStore = <Path extends unknown[]>(
       const largestSize = Math.max(next.length, prev.length);
 
       // Diff array
-      for (let start = 0, end = largestSize - 1; start < end; start++, end--) {
+      for (let start = 0, end = largestSize - 1; start <= end; start++, end--) {
         diff(next[start], prev[start], [...path, start] as Path);
+        if (start === end) break;
         diff(next[end], prev[end], [...path, end] as Path);
       }
 
--- a/packages/xstate-solid/test/useMachine.test.tsx
+++ b/packages/xstate-solid/test/useMachine.test.tsx
@@ -485,6 +485,52 @@ describe('useMachine hook', () => {
     expect(countEl.textContent).toEqual('1');
   });
 
+  it('useMachine array with odd number of items should be replaceable', () => {
+    const machine = createMachine({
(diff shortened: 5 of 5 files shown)


davidkpiano (approved):

.changeset/config.json
```diff
@@ -9,7 +9,6 @@
     "@xstate/immer",
     "@xstate/graph",
     "@xstate/inspect",
-    "@xstate/solid",
```
Andarist, line 12: I think that before unignoring this we should get a better understanding of: https://github.com/statelyai/xstate/issues/3925

PR #4582, opened 2023-12-11
Test: a stopped rehydrated actor should be able to subscribe

christophe-g:
Related to #4583

### DIFF
--- a/packages/core/test/rehydration.test.ts
+++ b/packages/core/test/rehydration.test.ts
@@ -179,6 +179,44 @@ describe('rehydration', () => {
     expect(rehydratedActor.getSnapshot().value).toBe('c');
   });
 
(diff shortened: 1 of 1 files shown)


packages/core/test/rehydration.test.ts
```diff
@@ -179,6 +179,44 @@ describe('rehydration', () => {
...
+          }
+        },
+        b: {},
+        c: {}
+      }
+    });
+
+    const actor = createActor(machine).start();
+    actor.stop();
+    const persistedState = actor.getPersistedSnapshot();
```
Andarist, line 203: You should persist the snapshot before stopping the `actor`.
```suggestion
    const persistedState = actor.getPersistedSnapshot();
    actor.stop();
```

PR #4584, opened 2023-12-12
Invariants POC

davidkpiano:
CleanShot 2023-12-11 at 19 54 06@2x

### DIFF
--- a/packages/core/src/setup.ts
+++ b/packages/core/src/setup.ts
@@ -1,3 +1,5 @@
+import { createActor } from '.';
+import { MachineSnapshot } from './State';
 import { StateMachine } from './StateMachine';
 import { createMachine } from './createMachine';
 import { GuardPredicate } from './guards';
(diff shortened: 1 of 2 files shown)


jordanebelanger:
Very interested in this, until this drops it's hard to have real type safety with regard to the context which is crucial for my application.

davidkpiano:
> Very interested in this, until this drops it's hard to have real type safety with regard to the context which is crucial for my application.

Good to know; we're still continuing to think about the best way to tackle this.

technoplato:
> > Very interested in this, until this drops it's hard to have real type safety with regard to the context which is crucial for my application.
> 
> Good to know; we're still continuing to think about the best way to tackle this.

Any movement on this @davidkpiano? This is definitely a killer limitation for trying to get greater buy in for anything remotely complex. Totally understand that is complicated though.

PR #4590, opened 2023-12-15
[poc] Use a `functionBrand` to make limited callback actors possible

Andarist:
reusing the brand trick from https://github.com/statelyai/xstate/pull/4256 to allow implementing type-safe~ callbacks actors that can only be used within machines that accept events that they might `sendBack` to them

related to https://discord.com/channels/795785288994652170/1177607380905775165/1184957932878696458

### DIFF
--- a/packages/core/test/setup.types.test.ts
+++ b/packages/core/test/setup.types.test.ts
@@ -1,9 +1,11 @@
 import {
+  ActorRef,
   ActorRefFrom,
   and,
   assign,
   createActor,
   createMachine,
+  fromCallback,
   fromPromise,
   fromTransition,
   not,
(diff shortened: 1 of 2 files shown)


packages/core/test/setup.types.test.ts
```diff
@@ -1599,3 +1601,54 @@ describe('setup()', () => {
...
+      actions: 'myAction'
+    }
+  }
+});
+
+createMachine({
+  types: {
+    events: {} as { type: 'DIFFERENT_EVENT' }
+  },
+  invoke: registerEvent('MY_EVENT')
```
Andarist, line 1653: the error here is expected, that's the whole point of this PR :P

PR #4591, opened 2023-12-15
Svelte template

davidkpiano:


packages/xstate-vue/src/useActor.ts
```diff
@@ -13,7 +13,7 @@ import { useSelector } from './useSelector.ts';
 
 export function useActor<TLogic extends AnyActorLogic>(
   actorLogic: TLogic,
-  options: ActorOptions<TLogic>
+  options?: ActorOptions<TLogic>
```
Andarist, line 16: we don't implement this in any of the integration packages yet - but this reminds me that we should make `input` required here when it's required

Andarist (approved):

PR #4596, opened 2023-12-17
[core] `getNextSnapshot(...)`

davidkpiano:
Introduce `getNextSnapshot(...)`, which determines the next snapshot for the given `actorLogic` based on the given `snapshot` and `event`.

If the `snapshot` is `undefined`, the initial snapshot of the `actorLogic` is used.

```ts
import { getNextSnapshot } from 'xstate';
import { trafficLightMachine } from './trafficLightMachine.ts';

const nextSnapshot = getNextSnapshot(
  trafficLightMachine, // actor logic
  undefined, // snapshot (or initial state if undefined)
  { type: 'TIMER' }
); // event object

console.log(nextSnapshot.value);
// => 'yellow'

const nextSnapshot2 = getNextSnapshot(
  trafficLightMachine, // actor logic
  nextSnapshot, // snapshot
  { type: 'TIMER' }
); // event object

console.log(nextSnapshot2.value);
// =>'red'
```

Supersedes #4056


packages/core/src/getNextSnapshot.ts
```diff
@@ -0,0 +1,55 @@
...
+    nextSnapshot, // snapshot
+    { type: 'TIMER' }); // event object
+
+  console.log(nextSnapshot2.value);
+  // =>'red'
+  ```
+ */
+export function getNextSnapshot<T extends AnyActorLogic>(
+  actorLogic: T,
+  snapshot: SnapshotFrom<T> | undefined,
```
andyjakubowski, line 40: Would it make sense to position the `snapshot` parameter last, so that you can omit it at the call site instead of passing `undefined`? That is: `getNextSnapshot(machine, { type: 'TIMER' })` instead of `getNextSnapshot(machine, undefined, { type: 'TIMER' })`.

packages/core/src/getNextSnapshot.ts
```diff
@@ -0,0 +1,55 @@
+import {
+  AnyActorLogic,
+  AnyActorScope,
+  EventFromLogic,
+  SnapshotFrom,
+  createEmptyActor
+} from '.';
```
Andarist, line 7: we should avoid circular deps - let's import those from places in which they originate and not from the `index.ts`

packages/core/src/getNextSnapshot.ts
```diff
@@ -0,0 +1,56 @@
...
+    self,
+    defer: () => {},
+    id: '',
+    logger: () => {},
+    sessionId: '',
+    stopChild: () => {},
+    system: self.system
+  };
+  const resolvedSnapshot =
+    snapshot ?? actorLogic.getInitialSnapshot(inertActorScope, null as any);
```
Andarist, line 54: IMHO this `null as any` is a big red flag here. It shows how easy it is to end up with problems here as `null` won't satisfy all `TInput`s.

packages/core/src/getNextSnapshot.ts
```diff
@@ -0,0 +1,56 @@
...
+  return actorLogic.transition(resolvedSnapshot, event, inertActorScope);
```
Andarist, line 55: One extra problem that I see here is that each time this `getNextSnapshot` is called one ends up with a new `inertActorScope`. While it probably won't create big problems in the vast majority of cases, it feels incoherent that a basic assumption that `self` is the same across the whole "session" doesn't hold up when using this.

packages/core/src/getNextSnapshot.ts
Andarist: I don't think this would be beneficial. Almost all our methods accept `snapshot` before the `event`. Making an exception here would be odd. It also "forces" you to make some decision about this `snapshot` - optional parameters are so easy to forget about that bugs often are caused by people just forgetting to pass them in.

andyjakubowski: I wasn’t aware of your ordering convention. And if it’s preferable to have people make a conscious decision about `snapshot`, then it all makes sense as is 👍

packages/core/src/getNextSnapshot.ts
davidkpiano: Added `getInitialSnapshot(logic, input)` to address this

packages/core/src/getNextSnapshot.ts
davidkpiano: The preferred solution (IMO) would be to attach `self` to the snapshot.

packages/core/src/getNextSnapshot.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  console.log(nextSnapshot2.value);
+  // =>'red'
+  ```
+ */
+export function getNextSnapshot<T extends AnyActorLogic>(
+  actorLogic: T,
+  snapshot: SnapshotFrom<T>,
+  event: EventFromLogic<T>
+): any {
```
Andarist, line 71:
```suggestion
): SnapshotFrom<T> {
```

packages/core/src/getNextSnapshot.ts
```diff
@@ -0,0 +1,74 @@
...
+): AnyActorScope {
+  const self = createEmptyActor() as ActorRefFrom<T>;
+  const inertActorScope: ActorScope<SnapshotFrom<T>, EventFromLogic<T>, any> = {
+    self,
+    defer: () => {},
+    id: '',
+    logger: () => {},
+    sessionId: '',
+    stopChild: () => {},
+    system: (self as any).system // TODO: fix types?
```
Andarist, line 24: This is caused by `T extends AnyActorLogic` pattern. Or rather with the `ActorRefFrom<T>` pattern. It's fundamentally at odds with how TS works and for that reason, I'm, at the moment, not a fan of this pattern.

The problem is that TS can't guarantee that `ActorRefFrom<T>` "returns" an `ActorRef` type. It's a conditional type and it got fed up generic type argument - as such it stays deferred and can't be "resolved". And well, you can't access properties on a conditional type :v

packages/core/src/getNextSnapshot.ts
```diff
@@ -0,0 +1,74 @@
...
+  EventFromLogic,
+  InputFrom,
+  SnapshotFrom
+} from './types.ts';
+
+/** @internal */
+export function createInertActorScope<T extends AnyActorLogic>(
+  _actorLogic: T
+): AnyActorScope {
+  const self = createEmptyActor() as ActorRefFrom<T>;
```
Andarist, line 16: Sample case why this is unsafe:
```ts
createMachine({
  entry: assign(({ self }) => {
     self.getSnapshot().value // oops! it throws
     return {}
  })
})
```

`createEmptyActor` returns `Snapshot<undefined>` so it's not compatible with the snapshot type of the actual `self` that an actor (and its `transition`) expects

packages/core/src/getNextSnapshot.ts
Andarist: Which again is at odds with snapshot management since you'd force every implementer of the actor logic to attach this to its snapshot ;p The pure/impure "split" in our API is quite challenging 😢 Looking at it from a different angle - the problem that we might notice is that "pure" bits have access to the `actorScope`. If the pure bits were decoupled from it that would simplify some things (but it's also not a change that we can take now anyway...)

packages/core/src/getNextSnapshot.ts
davidkpiano: What would be the solution?

packages/core/src/getNextSnapshot.ts
davidkpiano: We can change this with `const self = createActor(actorLogic)`

packages/core/src/getNextSnapshot.ts
Andarist: Stop using the `ActorRefFrom<T>` pattern ;p While separate generics are somewhat annoying - they fit the type system better. If we don't want to do this here then we have to be fine with this `as any` and we should remove this TODO comment - since it's not something that can be fixed while using the `ActorRefFrom<T>` pattern

packages/core/src/getNextSnapshot.ts
Andarist: This would definitely be better 👍

packages/core/src/getNextSnapshot.ts
davidkpiano: Removed

packages/core/src/getNextSnapshot.ts
```diff
@@ -0,0 +1,74 @@
...
+  // =>'red'
+  ```
+ */
+export function getNextSnapshot<T extends AnyActorLogic>(
+  actorLogic: T,
+  snapshot: SnapshotFrom<T>,
+  event: EventFromLogic<T>
+): SnapshotFrom<T> {
+  const inertActorScope = createInertActorScope(actorLogic);
+  return actorLogic.transition(snapshot, event, inertActorScope);
```
Andarist, line 72: if you do what you have mentioned [here](https://github.com/statelyai/xstate/pull/4596/files#r1432858780) then you kinda should do smth like this here:
```suggestion
  const inertActorScope = createInertActorScope(actorLogic);
  (inertActorScope.self as any)._snapshot = snapshot;
  return actorLogic.transition(snapshot, event, inertActorScope);
```

Andarist: otherwise the `self.getSnapshot()` will always return the initial snapshot in actions and that doesn't match how this would behave in a running machine

we also can't use `createActor(logic, { snapshot })` because this argument is meant to be the persisted snapshot... but we only have a snapshot here

Andarist (approved):

PR #4597, opened 2023-12-17
[core] Add `self` and `system` to `enqueueActions(...)`

davidkpiano:
Updates the argument object of `enqueueActions(...)` to include a `self` property:

```ts
// ...
entry: enqueueActions(({ self }) => {
  // ...
});
```

### DIFF
--- a/.changeset/strong-jobs-unite.md
+++ b/.changeset/strong-jobs-unite.md
@@ -1,0 +1,12 @@
+---
+'xstate': patch
+---
+
+Update the argument object of `enqueueActions(...)` to include a `self` property:
+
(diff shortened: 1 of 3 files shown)


.changeset/strong-jobs-unite.md
```diff
@@ -0,0 +1,12 @@
+---
+'xstate': patch
+---
+
+Update the argument object of `enqueueActions(...)` to include a `self` property:
```
Andarist, line 5: I pushed out a change to include `system` here too

Andarist (approved):

PR #4598, opened 2023-12-17
Add JSDocs

davidkpiano:


packages/core/src/State.ts
```diff
@@ -85,8 +112,8 @@ interface MachineSnapshotBase<
   children: TChildren;
 
   /**
-   * Whether the current state value is a subset of the given parent state value.
-   * @param testValue
+   * Whether the current state value is a subset of the given partial state value.
+   * @param partialStateValue
```
Andarist, line 116: partial sounds like literally a partial of the state value but for a state value like `{ a: 'b' }` we can also test against `'a'`

packages/core/src/State.ts
```diff
@@ -68,8 +71,32 @@ interface MachineSnapshotBase<
...
+  // single-level state node
+  state.value; // => 'yellow'
+
+  // nested state nodes
+  state.value; // => { red: 'wait' }
+  ```
+   */
   value: TStateValue;
+  /**
+   * The current status of this state.
```
Andarist, line 98:
```suggestion
   * The current status of this snapshot.
```

packages/core/src/State.ts
```diff
@@ -68,8 +71,32 @@ interface MachineSnapshotBase<
...
+   * This represents the active state nodes in the state machine.
+   * - For atomic state nodes, it is a string. 
+   * - For compound parent state nodes, it is an object where:
+   *   - The key is the parent state node's key
+   *   - The value is the current state value of the active child state node(s)
+   * 
+   * @example
+  ```ts
+  // single-level state node
+  state.value; // => 'yellow'
```
Andarist, line 90:
```suggestion
  snapshot.value; // => 'yellow'
```

packages/core/src/State.ts
```diff
@@ -68,8 +71,32 @@ interface MachineSnapshotBase<
...
+
+  // nested state nodes
+  state.value; // => { red: 'wait' }
```
Andarist, line 93:
```suggestion
  snapshot.value; // => { red: 'wait' }
```

packages/core/src/State.ts
davidkpiano: Yeah, was just trying to figure out something less ambiguous than `testValue`, which doesn't suggest a state value

ducin:
One thing I'd like to add here is - it would be super cool to provide simple example of usage for most important functions, such as `createActor` or whatever. `rxjs` is an example package where internal impl. is full of explanations and examples. I know it might be quite some effort to get there, but at least simple usage examples would be very useful.

davidkpiano:
> One thing I'd like to add here is - it would be super cool to provide simple example of usage for most important functions, such as `createActor` or whatever. `rxjs` is an example package where internal impl. is full of explanations and examples. I know it might be quite some effort to get there, but at least simple usage examples would be very useful.

The `createActor` function, as well as many other functions, do have examples in them 👍 

Just some gaps that need to be filled

packages/core/src/guards.ts
```diff
@@ -138,6 +138,32 @@ function checkNot(
...
+        guard: not('someNamedGuard'),
+        actions: () => {
+          // will be executed if guard in `not(...)`
+          // evaluates to `false`
+        }
+      }
+    }
+  });
+  ```
+ * @returns A guard action object
```
Andarist, line 165: the return value is not exactly a plain object (it's a function)

packages/core/src/createMachine.ts
```diff
@@ -91,6 +128,10 @@ export function createMachine<
     TOutput,
     TTypesMeta
   >,
+  /**
+   * @internal
+   * @deprecated Use `setup({ ... })` or `machine.provide({ ... })` instead
+   */
```
Andarist, line 134: You can't add JSDocs directly to parameters. They would have to be hoisted to the JSDoc for the function declaration (and annotated through `@params` and such). Adding `@internal` could be considered a breaking change. But anyway, I'm not really sure if either of those is supported for params - since you have to hoist those for them to actually have an effect... they start to be applied to the function declaration itself and not to individual parameters.

packages/core/src/utils.ts
```diff
@@ -258,6 +259,9 @@ export function resolveReferencedActor(machine: AnyStateMachine, src: string) {
   ).src;
 }
 
+/**
+ * @internal
+ */
```
Andarist, line 264: does it have any effect? We are actually exporting this as `__unsafe_getAllOwnEventDescriptors` - but I'm sure if this is applied transitively and even if it is then we should just... remove that export instead of marking this as internal here.

packages/core/src/utils.ts
```diff
@@ -75,6 +75,7 @@ export function toStateValue(
   return pathToStateValue(statePath);
 }
 
+/** @internal */
```
Andarist, line 78: Similarly here - this is **explicitly** exported today. It's not like it got exported accidentally through a wildcard or something. I think we should keep it or remove the export itself (from `src/index.ts`).

packages/core/src/createActor.ts
```diff
@@ -674,6 +676,9 @@ export class Actor<TLogic extends AnyActorLogic>
     return this.logic.getPersistedSnapshot(this._snapshot, options);
   }
 
+  /**
+   * @internal
+   */
```
Andarist, line 681: If you mark this as internal it will break people relying on this. This one exists for a reason here.
```suggestion
```

packages/core/src/StateMachine.ts
```diff
@@ -525,6 +529,9 @@ export class StateMachine<
     return restoredSnapshot;
   }
 
-  /** @deprecated an internal property that was acting as a "phantom" type, it's not used by anything right now but it's kept around for compatibility reasons */
+  /**
+   * @internal
```
Andarist, line 533: > but it's kept around for compatibility reasons

this is the important bit... ;p by marking it as `@internal` you essentially remove this from the generated declaration files and for now it was kept here to avoid breaking people when they install a new version of XState v5 without installing a new version of `@xstate/react@^4`

Andarist:
Let's merge this sooner than later. We don't need a long-lived umbrella branch for this - those JSDoc improvements can land incrementally.

packages/core/src/guards.ts
davidkpiano: For the developer's intents, it is an object (saying it's a function would be confusing, and is an implementation detail)

Andarist: "guard action" is definitely confusing, we can just make it "guard". Similarly, we don't even need to specify what it is - for the developer's intent it's just a guard (a "guard thing" but calling it a "guard thing" would be silly :D), they shouldn't inspect it anyhow so the extra "object" bit isn't that helpful here anyway

packages/core/src/typegenTypes.ts
```diff
@@ -10,12 +10,26 @@ import {
   IndexByProp
 } from './types.ts';
 
+/**
+ * @internal
```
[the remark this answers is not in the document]
davidkpiano: I don't think it disappears from the generated declaration files: https://www.typescriptlang.org/play#example/jsdoc-deprecated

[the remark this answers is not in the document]
Andarist: Sorry, I meant "marking those as internal"

packages/core/src/utils.ts
davidkpiano: This only removes it from the generated documentation.

PR #4600, opened 2023-12-17
[core] Remove some typegen-specific types

davidkpiano:
Motivation:

- The preferred way of avoiding missing implementations is to use `setup(...).createMachine(...)`
- Specialized type-checking should be via a language server (e.g. in a future extension) or something like eslint
- Desire to simplify the types in XState and integrations


Andarist:
> The preferred way of avoiding missing implementations is to use setup(...).createMachine(...)

It's just that this one isn't the case yet. `setup` isn't fully capable of avoiding missing implementations.

packages/core/src/typegenTypes.ts
```diff
@@ -104,40 +101,6 @@ export interface ResolvedTypegenMeta extends TypegenMeta {
 
 export type TypegenConstraint = TypegenEnabled | TypegenDisabled;
 
-// if combined union of all missing implementation types is never then everything has been provided
-export type AreAllImplementationsAssumedToBeProvided<
```
Andarist, line 108: older versions of integration packages still refer to this so it might be best to just `@deprecate` it

davidkpiano: Older integration packages don't work with XState v5 anyway

Andarist: > unless you mean minor/patch versions from the current major?

This

davidkpiano:
@Andarist Is this good to go?

Andarist:
I'm working on some further code cleanups related to this.

packages/xstate-solid/test/useMachine.test.tsx
```diff
@@ -1270,11 +1278,13 @@ describe('useMachine hook', () => {
     });
     const [isAwesome, setIsAwesome] = createSignal(false);
     const App = () => {
-      const [state, send] = useMachine(machine, {
-        guards: {
-          isAwesome: () => isAwesome()
-        }
-      });
+      const [state, send] = useMachine(
```
Andarist, line 1281: in a separate PR we should remove some of the tests here - we have 2 describe blocks here, one is targeting the strict mode... but Solid doesn't have strict mode. Those tests were just copy-pasted from `@xstate/react` and didn't go through a purge of redundancy :P

davidkpiano: Sounds good

.changeset/mean-pillows-move.md
```diff
@@ -0,0 +1,8 @@
+---
+'@xstate/svelte': patch
+'@xstate/react': patch
+'@xstate/vue': patch
+'xstate': patch
+---
+
+Typegen types have been mostly removed internally.
```
Andarist, line 8:
```suggestion
Typegen-based types for detecting missing implementations have been removed internally.
```

PR #4609, opened 2023-12-20
Call `complete` listeners exclusively when the actor reaches its done status

Andarist:
related to https://github.com/statelyai/xstate/discussions/4608

this removes the ability to listen for "on stop" **but** `complete` was already **not** called in different scenarios for stopped snapshots:
- like in `subscribe` [here](https://github.com/statelyai/xstate/blob/12e062bd594a1d05c8bc3eab7c15300384c04367/packages/core/src/interpreter.ts#L368)
- or like when starting an actor with a rehydrated stopped snapshot [here](https://github.com/statelyai/xstate/blob/12e062bd594a1d05c8bc3eab7c15300384c04367/packages/core/src/interpreter.ts#L440)

So this is currently inconsistent, one way or another, and we should unify it. When it comes to the "on stop" listener - personally, I just don't see a lot of use cases for it. A different modeling approach is always possible and I'd prefer to steer people towards explicit modeling and away from adding implicit ad-hoc logic in listeners like this.

### DIFF
--- a/packages/core/src/interpreter.ts
+++ b/packages/core/src/interpreter.ts
@@ -513,7 +513,6 @@ export class Actor<TLogic extends AnyActorLogic>
     this.update(nextState, event);
     if (event.type === XSTATE_STOP) {
       this._stopProcedure();
-      this._complete();
     }
   }
 
--- a/packages/core/test/interpreter.test.ts
+++ b/packages/core/test/interpreter.test.ts
@@ -1414,20 +1414,18 @@ describe('interpreter', () => {
       expect(completeCb).toHaveBeenCalledTimes(1);
     });
 
-    it('should call complete() once the interpreter is stopped', () => {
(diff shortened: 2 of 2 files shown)


davidkpiano (commented):
I checked, and this PR removed the ability to listen for an actor stopping. I added code to enable this (tests pass locally, not sure why they're failing in CI), can you please review?

Andarist:
> For "on stop", which is still an important "event" to listen for

Why? 😉 Whenever I think about it, I always conclude that I wouldn't ever want to listen to it. `subscribe` is currently mainly used to notify people about a meaningful change in the underlying data and I think that, more often than not, people wouldn't consider a stopped snapshot something that they need to account for. The stopped snapshot is important for us internally but users should rarely interact with it in their code.

I think it's important to back up the claim of its importance with some real-life use cases oriented on app development and their needs.

PR #4610, opened 2023-12-21
format with prettier. update getPersistedSnapshot() function call

gavination:
Just keeping up with the new updates for V5

### DIFF
--- a/examples/express-workflow/index.ts
+++ b/examples/express-workflow/index.ts
@@ -1,5 +1,5 @@
-import { createActor } from 'xstate';
-import bodyParser from 'body-parser';
+import { createActor } from "xstate";
+import bodyParser from "body-parser";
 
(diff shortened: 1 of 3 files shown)


examples/express-workflow/index.ts
```diff
@@ -35,21 +35,21 @@ app.post('/workflows', (req, res) => {
...
   const snapshot = persistedStates[workflowId];
 
   if (!snapshot) {
-    return res.status(404).send('Actor not found');
+    return res.status(404).send("Actor not found");
   }
 
   const event = req.body;
-  const actor = createActor(machine, { state: snapshot }).start();
+  const actor = createActor(machine, snapshot).start();
```
davidkpiano, line 47:
```suggestion
  const actor = createActor(machine, { snapshot }).start();
```

Issue #4612, opened 2023-12-21
Bug: 'self' missing in context


josephfh:
### XState version

XState version 5

### Description

_self_ is available to the assign function, which makes it possible to pass the parent machine down to a spawned child machine:

```
assign({
    spawnie: ({ self, spawn }) => spawn(someMachine, {
        input: { parent: self },
    })
})
```

However _self_ is not available when specifying the initial context of a machine:

```
context: ({ self, spawn }) => spawn(someMachine, {
    input: { parent: self },
})
```

### Expected result

It should be possible to deconstuct self in the context assignation

### Actual result

It doesn't work. When trying to use context.parent.send(...) from the spawned child:

`Uncaught TypeError: Cannot read properties of undefined (reading 'send')`

### Reproduction

https://stackblitz.com/edit/github-keqmzr-y5weyb?file=src%2FfeedbackMachine.ts

PR #4616, opened 2023-12-22
Pass `self` to `context` factories

Andarist:
fixes https://github.com/statelyai/xstate/issues/4612


packages/core/src/interpreter.ts
```diff
@@ -697,6 +697,11 @@ export class Actor<TLogic extends AnyActorLogic>
    * @see {@link Actor.getPersistedSnapshot} to persist the internal state of an actor (which is more than just a snapshot).
    */
   public getSnapshot(): SnapshotFrom<TLogic> {
+    if (isDevelopment && !this._snapshot) {
+      throw new Error(
+        `Snapshot can't be read while the actor initializes itself`
```
davidkpiano, line 702:
```suggestion
        `Snapshot cannot be read before actor is started.`
```

davidkpiano (approved):

packages/core/src/interpreter.ts
Andarist: The suggested change is not correct. This is perfectly valid:
```ts
const actorRef = createActor(machine)
actorRef.getSnapshot()
actorRef.start()
```

This isn't:
```ts
  context: ({ self }) => {
    self.getSnapshot() // we can't do this
    return {}
  }
})
createActor(machine)
```

davidkpiano: Oh okay, makes sense

PR #4621, opened 2023-12-22
An attempt to fix compatibility with `exactOptionalPropertyTypes`

Andarist:
Weirdly `onSnapshot` behaves differently from `onDone` and I don't understand yet why because they are typed in a very similar way.

I might have to explore alternative solutions


davidkpiano:
Closing as stale 🧹

PR #4638, opened 2023-12-27
[work in progress] feature/xstate/angular

ducin:
This is a very early implementation that already does its job. It's not ready for a merge yet, however, I'd like to start review with @Andarist asap so that it's easier later on. There might be some setup stuff (or some conventions?) I'm missing, most probably.

# usage

```ts
export class AppComponent {
  #injector = inject(Injector);
  toggleActor = injectActor(toggleMachine, undefined, {
    injector: this.#injector
  });

  constructor() {
    this.toggleActor.ref.start();
  }

  handleToggleClick() {
    this.toggleActor.send({ type: 'toggle' });
  }
}
```

Nowadays, this is the standard approach in angular to get dependencies straight into a component (or another service):

```ts
  #injector = inject(Injector);
  toggleActor = injectActor(toggleMachine, undefined, {
    injector: this.#injector
  });
```

injecting the Injector looks odd, but that's actually expected, since there's a whole hierarchy of injectors, so you might get lots of different injectors when needed. And passing the injector via parameter is also a standard approach in libraries (either your call is within, so called, _injection context_ and our bridge would get the injector itself, or one could pass it explicitly via param). All standard.

# requirements

As @Andarist mentioned, there has to be a way to allow components having multiple state machines.

# cleanups

the most important part is actually this:
https://github.com/ducin/xstate/blob/feature/xstate/angular/packages/xstate-angular/src/injectActor.ts#L50-L53
all it does is it grabs destroyRef (an angular kinda-equivalent to useEffect return-callback) from a certain injector. Also, pretty standard.

# further work

I want to finish the following before considering the PR ready:
- cleanups (e.g. comments), obviously
- some more examples (working on it)
- analyzing possibility to provide globally available machines (could be achieved by explicitly creating a service and making it available in certain areas of application, so it's an **optional** feature)


ducin:
@Andarist what should we do with this?
Angular v17 requires node v18+. Going back to v16 makes little sense for the adapter since there's no stable signal support, and without the signal support the adaption would be significantly smaller (it's the new standard).

```
error @angular/common@17.0.7: The engine "node" is incompatible with this module. Expected version "^18.13.0 || >=20.9.0". Got "16.18.1"
error Found incompatible module.
```

Andarist:
> @Andarist what should we do with this?

This might have already been addressed in the meantime. Could you sync with `main`? Its CI is already using node 20

examples/toggle-angular/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@angular/common": "^17.0.0",
+    "@angular/compiler": "^17.0.0",
+    "@angular/core": "^17.0.0",
+    "@angular/forms": "^17.0.0",
+    "@angular/platform-browser": "^17.0.0",
+    "@angular/platform-browser-dynamic": "^17.0.0",
+    "@angular/router": "^17.0.0",
+    "rxjs": "~7.8.0",
+    "tslib": "^2.3.0",
+    "xstate": "^5.3.0",
```
Andarist, line 23: Shouldn't this dependency list mention `@xstate/angular` too? It has not been published yet though and `examples` are outside of the core monorepo... so this might be slightly problematic right now. Maybe it would be worth to split this (before the merge) into 2 different PRs? One could introduce `@xstate/angular` and the other one could introduce an example using it.

examples/toggle-angular/src/app/app.component.ts
```diff
@@ -0,0 +1,48 @@
...
+        <output>{{ toggleActor.snapshot().value }}</output>
+        <button type="button" (click)="handleToggleClick()">Toggle</button>
+      </div>
+    </div>
+  `,
+  styleUrl: './app.component.css'
+})
+export class AppComponent {
+  @ViewChild('output') outputEl!: ElementRef<HTMLDivElement>;
+  // protected stateLabel: string = '';
```
Andarist, line 30: those comments here should be cleaned up before merge

examples/toggle-angular/tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    "experimentalDecorators": true,
+    "moduleResolution": "node",
+    "importHelpers": true,
+    "target": "ES2022",
+    "module": "ES2022",
+    "useDefineForClassFields": false,
+    "lib": ["ES2022", "dom"],
+    "paths": {
+      "@xstate/angular": ["../../packages/xstate-angular"]
+    }
```
Andarist, line 25: If we split this PR into two (like mentioned in the other comment) then we could remove this, right? Then `@xstate/angular` would just get installed from the registry and things would "just work"

examples/toggle-angular/yarn.lock
Andarist, line 1: It seems that this PR introduces 2 different lockfiles in this example - we should stick to one 😉 I think most of our examples are meant to be using pnpm - but I see that some of them might have some leftover yarn/npm lockfiles too.

examples/toggle/yarn.lock
Andarist, line 1: this likely shouldnt be touched as part of this PR 😉

packages/xstate-angular/CHANGELOG.md
Andarist, line 1: I'd probably remove this file altogether. It will just get auto-created in the future when we release new versions of this package.

packages/xstate-angular/README.md
Andarist, line 1: It would be great to have at least a basic usage example in this README

packages/xstate-angular/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "state",
+    "machine",
+    "statechart",
+    "scxml",
+    "state",
+    "graph",
+    "angular",
+    "inject"
+  ],
+  "author": "Tomasz Ducin <tomasz.ducin@gmail.com>",
```
Andarist, line 15: All other packages mention David as the author so it feels weird to have a different author here. But I don't know, an "author" field in the MIT-licensed OSS software is a weird concept to me 🤷‍♂️ Maybe we should just remove this field altogether from (almost?) all packages?

packages/xstate-angular/package.json
```diff
@@ -0,0 +1,81 @@
...
+  "homepage": "https://github.com/statelyai/xstate/tree/main/packages/xstate-angular#readme",
+  "license": "MIT",
+  "main": "dist/xstate-angular.cjs.js",
+  "module": "dist/xstate-angular.esm.js",
+  "type": "module",
```
Andarist, line 20: is this a requirement? It certainly makes the current `exports` configuration incorrect since now `.js` files will be treated as modules but they are actually scripts

Andarist: If possible - I would prefer removing this field from here.

packages/xstate-angular/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "@angular/core": "^17.0.0",
+    "@angular/platform-browser": "^17.0.0",
+    "@angular/platform-browser-dynamic": "^17.0.0",
+    "xstate": "^5.3.0"
+  },
+  "peerDependenciesMeta": {
+    "xstate": {
+      "optional": true
+    }
+  },
```
Andarist, line 70: If we still have this in other packages we should remove this in another PR
```suggestion
```

packages/xstate-angular/package.json
```diff
@@ -0,0 +1,81 @@
...
+  "bugs": {
+    "url": "https://github.com/statelyai/xstate/issues"
+  },
+  "peerDependencies": {
+    "@angular/common": "^17.0.0",
+    "@angular/compiler": "^17.0.0",
+    "@angular/compiler-cli": "^17.0.0",
+    "@angular/core": "^17.0.0",
+    "@angular/platform-browser": "^17.0.0",
+    "@angular/platform-browser-dynamic": "^17.0.0",
```
Andarist, line 63: are all of those required as peer deps? :o it's kinda crazy 😅

packages/xstate-angular/package.json
```diff
@@ -0,0 +1,81 @@
...
+    }
+  },
+  "dependencies": {},
+  "devDependencies": {
+    "@angular/common": "^17.0.0",
+    "@angular/compiler": "^17.0.0",
+    "@angular/compiler-cli": "^17.0.0",
+    "@angular/core": "^17.0.0",
+    "@angular/platform-browser": "^17.0.0",
+    "@angular/platform-browser-dynamic": "^17.0.0",
```
Andarist, line 78: do we use anything beyond `@angular/core` here? I understand that a lot of this might be required in the final app - but are those anyhow related to this standalone~ library? 🤔

packages/xstate-angular/src/false.ts
Andarist, line 1: If we don't use `#is-development` for anything within this package then let's remove this (and the associated configuration from the `package.json`)

packages/xstate-angular/src/provideActor.ts
Andarist, line 1: What would this potential export do? Is this related to the global actors idea that you have mentioned in the description of this PR?

packages/xstate-angular/src/injectActor.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    const snapshot = signal(actorInstance.getSnapshot());
+    const result = {
+      snapshot,
+      send: actorInstance.send,
+      ref: actorInstance
+    } as any;
+
+    const subscription = actorInstance.subscribe((currentSnapshot) => {
+      // result.snapshot = snapshot;
```
Andarist, line 46:
```suggestion
```

packages/xstate-angular/src/injectActor.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export function injectActor<TLogic extends AnyActorLogic>(
+  logic: TLogic,
+  options?: ActorOptions<TLogic>,
+  injectOptions?: injectActorOptions
+): InjectedActor<TLogic> {
+  const injector = injectOptions?.injector ?? inject(Injector);
+  const destroyRef = injector.get(DestroyRef);
+  // I'm afraid I stepped into this: https://github.com/angular/angular/issues/34478
```
Andarist, line 34: While this comment might be relevant to a problem that you have faced it feels out of place **here**. Did you end up using any of the workarounds mentioned [here](https://github.com/angular/angular/issues/34478#issuecomment-567579719)?

examples/toggle-angular/yarn.lock
Andarist: I removed the mentioned leftovers here: https://github.com/statelyai/xstate/pull/4641

packages/xstate-angular/src/injectActor.ts
```diff
@@ -0,0 +1,57 @@
...
+  return runInInjectionContext(injector, () => {
+    const actorInstance = createActor(logic as any, options).start();
```
Andarist, line 36: If you sync with `main` you should be able to remove this cast:
```suggestion
    const actorInstance = createActor(logic, options).start();
```

packages/xstate-angular/src/injectActor.ts
```diff
@@ -0,0 +1,57 @@
...
+interface InjectedActor<TLogic extends AnyActorLogic> {
+  snapshot: Signal<SnapshotFrom<TLogic>>;
+  send: ActorRefFrom<TLogic>['send'];
+  ref: ActorRefFrom<TLogic>;
+}
+
+export function injectActor<TLogic extends AnyActorLogic>(
+  logic: TLogic,
+  options?: ActorOptions<TLogic>,
+  injectOptions?: injectActorOptions
```
Andarist, line 30: I'd probably merge those two:
```suggestion
  { injector = inject(Injector), ...options }: ActorOptions<TLogic> & { injector?: Injector } = {},
```

packages/xstate-angular/src/injectActor.ts
```diff
@@ -0,0 +1,57 @@
...
+): InjectedActor<TLogic> {
+  const injector = injectOptions?.injector ?? inject(Injector);
```
Andarist, line 32: is it a common practice to use the default injector like this?

Andarist: I definitely like that this is an option and that people won't have to always pass `injector` in explicitly

packages/xstate-angular/src/injectActor.ts
Andarist, line 36: > wouldn't it make sense for the adapter to automatically start() the machine? Wouldn't it be the most common scenario to immediately start the machine (without the need to do that explicitly)?

Yes. That would be best, in my opinion.

I also wonder, what's the story around SSR in Angular? Is it enough to put `.start()` in `OnInit` to have this compatible with SSR?

packages/xstate-angular/src/injectActor.ts
```diff
@@ -0,0 +1,57 @@
...
+    } as any;
+
+    const subscription = actorInstance.subscribe((currentSnapshot) => {
+      // result.snapshot = snapshot;
+      snapshot.set(currentSnapshot);
+    });
+
+    destroyRef.onDestroy(() => {
+      actorInstance.stop();
+      subscription.unsubscribe();
```
Andarist, line 52: I usually find it much less error-prone to dispose of local things first before stopping the external~ resource. It usually leads to removing some edge cases around synchronous notification from within the stopping call or ones related to the fact that the other call might throw and interrupt the whole cleanup. I don't think we have to wrap this in any try/catches or nothing - nothing here should be able to throw (if it does then I'd consider that to be a likely bug in XState). It's just a rule of thumb that I developed over the years and try to follow as much as possible
```suggestion
      subscription.unsubscribe();
      actorInstance.stop();
```

davidkpiano:
Closing as stale; would be better to target v6 when the alpha is released

PR #4639, opened 2023-12-27
Bump version of @xtate/immer package to align with updated dependencies

lozandier:
Fixes inconsistent non-bumping of package version like other xstate packages that had the same recent updates of peer dependencies four days ago (inconsistent how packages such as `@xstate/react` and `@xstate/svelte` were handled when their peer dependencies were updated).

### DIFF
--- a/packages/xstate-immer/package.json
+++ b/packages/xstate-immer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@xstate/immer",
-  "version": "0.3.3",
+  "version": "0.3.4",
   "description": "XState with Immer",
   "keywords": [
     "immer",


Andarist:
The packages that you have mentioned were bumped because something has changed in their source code - not because the peer dep range has been changed. I don't see a reason here why `@xstate/immer` would have to be bumped.

lozandier:
Only reason I’m bumping is because importing it as is scans/reads xstate 4
as a dependency instead of 5 in several tools/scripts that imports NPM
packages in enterprise settings—that isn’t the case with xstate

Kevin Lozandier
***@***.*** ***@***.***>

davidkpiano:
Closing this as it is now recommended to use `immer` directly with XState (if desired) instead of with `@xstate/immer`: https://stately.ai/docs/immer

PR #4643, opened 2023-12-28
Add `InputFrom<TLogic>` to Spawner input option

DrSammyD:


packages/core/src/spawn.ts
```diff
@@ -58,7 +58,7 @@ export type Spawner<TActor extends ProvidedActor> = IsLiteralString<
       options?: {
         id?: string;
         systemId?: string;
-        input?: unknown;
+        input?: InputFrom<TLogic>;
```
DrSammyD, line 61:
```suggestion
        input?: TLogic extends string ? unknown : InputFrom<TLogic>;;
```

PR #4656, opened 2024-01-07
Update xstate-react.umd.min.js url to valid link

RohanArepally:
Resource `https://unpkg.com/@xstate/react/dist/xstate-react.umd.min.js` no longer exists so added a valid link to the `README.md`

### DIFF
--- a/packages/xstate-react/README.md
+++ b/packages/xstate-react/README.md
@@ -16,7 +16,7 @@ npm i xstate @xstate/react
 **Via CDN**
 
 ```html
-<script src="https://unpkg.com/@xstate/react/dist/xstate-react.umd.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/@xstate/react/dist/xstate-react.umd.min.js"></script>
 ```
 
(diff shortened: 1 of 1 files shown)


davidkpiano (approved):

packages/xstate-react/README.md
```diff
@@ -16,7 +16,7 @@ npm i xstate @xstate/react
 **Via CDN**
 
 ```html
-<script src="https://unpkg.com/@xstate/react/dist/xstate-react.umd.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/@xstate/react/dist/xstate-react.umd.min.js"></script>
```
Andarist, line 19: This is not a correct link. UMD build doesn't exist for `@xstate/react@4` and this likely points to some v3 version of that. The correct thing to do would be to remove this from the docs altogether or point people to skypack's CDN or smth here.

RohanArepally: Ah my mistake! I'll create a PR removing the line from the docs.

PR #4658, opened 2024-01-07
Fix the react friends machine example for v5

HartS:


davidkpiano:
Thanks for the PR! I made relevant changes and added you as a co-contributor on the commit.

PR #4661, opened 2024-01-09
Remove the parameter initializer from `spawn`'s signature

Andarist:


packages/core/src/spawn.ts
```diff
@@ -48,7 +48,7 @@ export type Spawner<TActor extends ProvidedActor> = IsLiteralString<
 > extends true
   ? <TSrc extends TActor['src']>(
       logic: TSrc,
-      ...[options = {} as any]: SpawnOptions<TActor, TSrc>
+      ...[options]: SpawnOptions<TActor, TSrc>
```
Andarist, line 51: This was a silly mistake. The initializers only make sense at runtime and this is a pure type-level signature. A misapplied initializer like this becomes an error in the upcoming TS 5.4 - hence the fix. We don't need any changeset here though since the emitted declaration files today are already correct and error-less - the error only happens when typechecking our source code

davidkpiano (approved):

Issue #4664, opened 2024-01-10
Bug: After 5.5, type of state.context from @xstate/react useMachine being inferred incorrectly with @xstate/react < 4.0.2


pckilgore:
### XState version

XState version 5

### Description

Updated from 5.4 -> 5.5

### Expected result

Everything works, everything compiles (patch change).

### Actual result

New type errors (state.context values being inferred to any instead of the correct type).

### Reproduction

[...]

PR #4667, opened 2024-01-11
Fixed compatibility with older versions of integration packages

Andarist:
fixes https://github.com/statelyai/xstate/issues/4664

### DIFF
--- a/packages/core/src/typegenTypes.ts
+++ b/packages/core/src/typegenTypes.ts
@@ -123,8 +123,16 @@ export type TypegenConstraint = TypegenEnabled | TypegenDisabled;
  */
 export type AreAllImplementationsAssumedToBeProvided<
   _TResolvedTypesMeta,
-  _TMissingImplementations
+  _TMissingImplementations = never
 > = true;
+
+/**
+ * @deprecated Always resolves to `never`
+ */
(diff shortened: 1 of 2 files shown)


packages/core/src/typegenTypes.ts
```diff
@@ -123,9 +123,17 @@ export type TypegenConstraint = TypegenEnabled | TypegenDisabled;
  */
 export type AreAllImplementationsAssumedToBeProvided<
   _TResolvedTypesMeta,
-  _TMissingImplementations
+  _TMissingImplementations = never
```
Andarist, line 126: There is a q if we even need to do this - but I think we should. We left `AreAllImplementationsAssumedToBeProvided` for compat reasons after all in https://github.com/statelyai/xstate/pull/4600 ... we just didn't account for its default type param (integration packages were only supplying one type arg) and we didn't do the same for `MissingImplementationsError`. 

The goal **was** to preserve compatibility and avoid issues and that fixes it. Since it's easy to do this, I think it's worth considering.

davidkpiano: I think this is fine

davidkpiano (approved):

PR #4674, opened 2024-01-13
Fix mutating actor in strict mode

davidkpiano:
There are potential problems with mutating the actor in the `stopRootWithRehydration` method, which is a hack to work around React strict mode. At least one problem has already surfaced: when inspecting machines (e.g. using `@statelyai/inspect`), it reports the _current actor_ as stopped, instead of the previous actor.

This is probably because the actor is being mutated, yet we're still reading the actor in e.g. an `@xstate.snapshot` event where `status: stopped`; however, by the time we read it, it's too late, the actor has been mutated to be reused in the second render, so `actorRef.sessionId` is the new actor.

I'd prefer _not_ to work around this in the inspector and instead avoid actor mutation. The only failing tests are with render counts, which aren't that big of a deal.

### DIFF
--- a/packages/xstate-react/src/useActor.ts
+++ b/packages/xstate-react/src/useActor.ts
@@ -1,9 +1,18 @@
 import isDevelopment from '#is-development';
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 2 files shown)


Andarist:
We decided to handle this in the inspector and investigate the implementation of `actor.restart`.

PR #4677, opened 2024-01-13
add support for subscribing to system registration events

cevr:
This adds support for subscribing to registration events within an actor's system.

### DIFF
--- a/packages/core/src/system.ts
+++ b/packages/core/src/system.ts
@@ -4,6 +4,7 @@ import {
   AnyActorRef,
   Observer,
   Snapshot,
+  Subscription,
   HomomorphicOmit,
   EventObject
 } from './types.ts';
@@ -61,6 +62,20 @@ export interface ActorSystem<T extends ActorSystemInfo> {
    */
   _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
   get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+  subscribe: {
+    (
+      listener: (
+        event: RegistrationEvent<
+          T['actors'][keyof T['actors']],
+          keyof T['actors'] & string
+        >
+      ) => void
+    ): Subscription;
+    <K extends keyof T['actors']>(
+      key: K,
+      listener: (event: RegistrationEvent<T['actors'][K], K & string>) => void
+    ): Subscription;
+  };
   inspect: (observer: Observer<InspectionEvent>) => void;
   /**
    * @internal
@@ -91,6 +106,8 @@ export interface ActorSystem<T extends ActorSystemInfo> {
 
 export type AnyActorSystem = ActorSystem<any>;
 
+const rootSubscriptionKey = '@xstate.system.root' as const;
+
 let idCounter = 0;
 export function createSystem<T extends ActorSystemInfo>(
   rootActor: AnyActorRef,
@@ -103,6 +120,10 @@ export function createSystem<T extends ActorSystemInfo>(
   const keyedActors = new Map<keyof T['actors'], AnyActorRef | undefined>();
   const reverseKeyedActors = new WeakMap<AnyActorRef, keyof T['actors']>();
   const observers = new Set<Observer<InspectionEvent>>();
+  const systemSubscriptions = new Map<
+    keyof T['actors'] | typeof rootSubscriptionKey,
+    Set<(event: RegistrationEvent<T['actors'][keyof T['actors']]>) => void>
+  >();
   const timerMap: { [id: ScheduledEventId]: number } = {};
   const clock = options.clock;
 
@@ -164,6 +185,19 @@ export function createSystem<T extends ActorSystemInfo>(
     _bookId: () => `x:${idCounter++}`,
     _register: (sessionId, actorRef) => {
       children.set(sessionId, actorRef);
+      const systemId = reverseKeyedActors.get(actorRef);
+      if (systemId !== undefined) {
+        const event = {
+          type: `@xstate.system.actor.register.${systemId as string}`,
+          actorRef: actorRef as T['actors'][keyof T['actors']]
+        } as const;
+        systemSubscriptions.get(systemId)?.forEach((listener) => {
+          listener(event);
+        });
+        systemSubscriptions.get(rootSubscriptionKey)?.forEach((listener) => {
+          listener(event);
+        });
+      }
       return sessionId;
     },
     _unregister: (actorRef) => {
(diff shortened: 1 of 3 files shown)


packages/core/src/system.ts
```diff
@@ -61,6 +62,20 @@ export interface ActorSystem<T extends ActorSystemInfo> {
    */
   _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
   get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+  subscribe: {
```
cevr, line 65: Not sure if overloading is preferred here, can break this out if not.

packages/core/src/system.ts
```diff
@@ -91,6 +106,8 @@ export interface ActorSystem<T extends ActorSystemInfo> {
 
 export type AnyActorSystem = ActorSystem<any>;
 
+const rootSubscriptionKey = '@xstate.system.root' as const;
```
cevr, line 109: added a root subscription key that seemed unique enough while following some conventions I found.

This would be used in the case `actor.system.subscribe(event => ...)` was used with no `systemId`.

packages/core/src/system.ts
```diff
@@ -262,3 +341,26 @@ export type InspectionEvent =
   | InspectedSnapshotEvent
   | InspectedEventEvent
   | InspectedActorEvent;
+
+export interface ActorRegisteredEvent<
+  TActorRef extends AnyActorRef,
+  TSystemId extends string = string
```
cevr, line 347: This made sense to me if there were any plans to allow for extracting `systemId` types from a machine.

Let me if it doesn't!

packages/core/src/system.ts
```diff
@@ -173,11 +207,56 @@ export function createSystem<T extends ActorSystemInfo>(
...
+        | K
+        | ((event: RegistrationEvent<T['actors'][keyof T['actors']]>) => void),
+      maybeListener?:
+        | ((event: RegistrationEvent<T['actors'][keyof T['actors']]>) => void)
+        | undefined
+    ) => {
+      let listener =
+        typeof maybeSystemIdOrListener === 'function'
+          ? maybeSystemIdOrListener
+          : maybeListener!;
```
cevr, line 236: not sure if this is okay, typescript would prevent this but there's no runtime check

packages/core/src/system.ts
```diff
@@ -173,11 +207,56 @@ export function createSystem<T extends ActorSystemInfo>(
...
+      let systemId =
+        typeof maybeSystemIdOrListener === 'function'
+          ? rootSubscriptionKey
+          : maybeSystemIdOrListener;
+
+      let subscriptions = systemSubscriptions.get(systemId) || new Set();
```
cevr, line 242: I think we could also design this differently by using a single set, and overriding the listeners if a systemId key is present. This would also mean passing the systemId directly in the event so its easier to parse

something like:

```tsx
let listener = 
  typeof maybeSystemIdOrListener === 'function' ? 
  maybeSystemIdOrListener 
  : 
    (event) => { 
      if (event.systemId === maybeSystemIdOrListener) maybeListener(event)
     }
```

what do you think is preferable here?

packages/core/src/system.ts
```diff
@@ -173,11 +228,54 @@ export function createSystem<T extends ActorSystemInfo>(
...
+      errorListener?: (error: any) => void,
+      completeListener?: () => void
+    ) => {
+      const observer = toObserver(
+        nextListenerOrObserver,
+        errorListener,
+        completeListener
+      );
+
+      if (rootActor._processingStatus !== ProcessingStatus.Stopped) {
```
cevr, line 257: Copied the implementation in `createActor`. I think it would make sense to keep this since system whose root actor is stopped won't be receiving updates, right?

packages/core/src/system.ts
```diff
@@ -173,11 +228,54 @@ export function createSystem<T extends ActorSystemInfo>(
...
+        const snapshot = rootActor.getSnapshot();
+        switch (snapshot.status) {
+          case 'done':
+            try {
+              observer.complete?.();
+            } catch (err) {
+              reportUnhandledError(err);
+            }
+            break;
+          // can this error?
```
cevr, line 269: What would an error look like here?

cevr:
Disregard the resolved comments, did a refactoring to make it use the same logic as `actor.subscribe` for consistency

Andarist:
Please include some motivation behind this feature in the PR's description.

cevr:
@Andarist Oops! Back to you

Andarist:
The motivation mentions that today it's tedious to access such actors in React but it doesn't show how this PR solves that. Could you include before/after comparison?

cevr:
@Andarist You're right. Updated it with some examples!

Andarist:
I think this particular design doesn't fit XState well. When you subscribe to an actor you are subscribed to its snapshot values but here, when subscribing to the system, you are subscribed to events. A proposal that would bring this closer to how things work with actors would fit XState better.

cevr:
@Andarist Fair enough. What are you thinking would be closer? There is `inspect`, which seems to also be a sort of subscription to events. Would it be better to appropriate `inspect` and make it also take `RegistrationEvents`? Currently it doesn't return a `Subscription` but that could be done rather easily.

What do you think?

Andarist:
The proposed implementation of `useSystemActorRef` shows how you are trying to create an implicit relationship between those events and the actor's existence in the system. I think a better design would be to decouple this from events completely and focus on what you are interested in and that's "the state" of the system. 

Currently `system` isn't an Actor at all - but what would it take to make it one (or at least make it compatible with the `ActorRef` interface etc)? It would be best if you could do this:
```ts
const notifier = useSelector(actorRef.system, system => system.get('notifier'))
```

On the other hand, the system is mutable today and that doesn't play with `useSelector` so maybe this is a dead end right now.

Could you write down how you want to use this in your app? What's the rough sketch of the architecture that you have there and why do you find yourself in a position of reaching for an actor that might not exist at a time etc.

cevr:
@Andarist I think this problem manifests itself only if you orchestrate the entirety of your application state in XState. 

The best example I can give is one I face currently:

App has three states:

Cover | Menu | Checkout

Cover has a bunch of substates
Idle | Wizard | Auth | Complete

Menu has a couple of substates as well
Browsing | Item

And each states can potentially have its own states as well. These states are all machines that are instantiated by its parent machine using `invoke`. To get access to these machines requires the sort of boilerplate I mentioned in the motivation. This could be resolved by decoupling the machines from their parent states, and have them orchestrated together entirely by React. It can also be resolved by having a flat registry of all actors within an actor's scope. I think option #2 is better, since it easier to reason about when looking at machine definitions how each state is related and it's already implemented via an actor's `system`.

The only missing piece to `system` is being able to observe changes to the system's registry. Which is what lead me to this PR.

cevr:
I think option 3 would also be to have each child state just be part of the root machine rather than a separate machine that is invoked. This comes with added complexity in types, which tends to be a deal breaker as the app scales.

cevr:
@Andarist What would you say the root issue is? Am I just holding xstate wrong, and should rather let React do the instantiations?

Andarist:
I think I stand by what I described in: https://github.com/statelyai/xstate/pull/4677#issuecomment-1893545070 . To improve the situation here we'd have to actor-ify the system further. Its current snapshot should be cached, each change in the registered actors should update it (in an immutable manner) and notify the subscribers about the change. This way you should be able to get this working:
```diff
-const notifier = useSelector(actorRef.system, system => system.get('notifier'))
+const notifier = useSelector(actorRef.system, s => s.actors.notifier)
```

cevr:
@Andarist Sounds good to me. What would a snapshot look like in this case?

image

Would we move `children, keyedActors, reversedKeyedActors` into one object?

cevr:
Also by making this an actor, do you mean to hook the system into the rootActor's `update` method? 

Currently looking at `useSelector` implementations, and it seems it subscribes to the rootActor only. So the only way for these changes to be propagated would be through the rootActor unless we change the implementation of `useSelector`

Andarist:
> Would we move children, keyedActors, reversedKeyedActors into one object?

Not necessarily. I'd like to keep the snapshot as lean as possible so it should only be extended with `actors`

> Also by making this an actor, do you mean to hook the system into the rootActor's update method?

No. I'm sorry for not being explicit enough. I meant that its implementation should get closer to an actor - not necessarily that it should literally use `Actor` class/`createActor` factory, or hook into `rootActor` anyhow. It only has to be extended with `subscribe` and its snapshot has to be extended.

cevr:
@Andarist back to you!

There are better tests that can be written, if you could give me some test cases you'd like to see I can get that done.

One question I have is: what would an `error` and `complete` call look like for an actor's system?

cevr:
Anything I can do to get this past the finish line?

cevr:
unfortunately, not yet. the system is still untyped. you'd have to annotate it manually in the selector or in the function

K3TH3R:
@cevr thank you for taking this on, being able to access deeply nested actors from anywhere in the system/UI was one of the biggest struggles I've had with building our app in XState (v4 mind you).

cevr:
Anything I can do to improve this pr? @Andarist @davidkpiano

cevr:
should have waited until the 2 year anniversary, but hopefully we can revive this lol

PR #4685, opened 2024-01-18
Allow escaped periods in state IDs

davidkpiano:
State IDs that have periods in them are now supported if those periods are escaped.

The motivation is that external tools, such as [Stately Studio](https://stately.ai/studio), may allow users to enter any text into the state ID field. This change allows those tools to escape periods in state IDs, so that they don't conflict with the internal path-based state IDs.

E.g. if a state ID of `"Loading..."` is entered into the state ID field, instead of crashing either the external tool and/or the XState state machine, it should be converted by the tool to `"Loading\\.\\.\\."`, and those periods will be ignored by XState.

---

Quick demo of the problem in Stately Studio:

https://github.com/statelyai/xstate/assets/1093738/3da093ac-fdb0-4534-a12b-1534b86b6112

The solution would be to escape the periods, but for that to be fully compatible with XState, we need to handle use-cases like that on the XState side too.

### DIFF
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -581,9 +581,7 @@ export function getStateNode(
       `Unable to retrieve child state '${stateKey}' from '${stateNode.id}'; no child states exist.`
     );
   }
-  // replace escaped period ('\.') with just a period
-  const resolvedStateKey = stateKey.replace(/\\\./g, '.');
-  const result = stateNode.states[resolvedStateKey];
+  const result = stateNode.states[stateKey];
   if (!result) {
     throw new Error(
       `Child state '${resolvedStateKey}' does not exist on '${stateNode.id}'`
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -581,7 +581,7 @@ export function getStateNode(
   const result = stateNode.states[stateKey];
   if (!result) {
     throw new Error(
-      `Child state '${result}' does not exist on '${stateNode.id}'`
+      `Child state '${stateKey}' does not exist on '${stateNode.id}'`
     );
   }
   return result;
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -236,8 +236,8 @@ export function getCandidates<TEvent extends EventObject>(
         // split the event descriptor from unescaped periods
         // e.g. 'foo.bar' is ['foo','bar']
         // e.g. 'foo\.bar' is ['foo\.bar']
-        const partialEventTokens = eventDescriptor.split(/(?<!\\)\./g);
-        const eventTokens = receivedEventType.split(/(?<!\\)\./g);
+        const partialEventTokens = eventDescriptor.split('.');
+        const eventTokens = receivedEventType.split('.');
 
         for (
           let tokenIndex = 0;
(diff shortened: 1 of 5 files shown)


packages/core/test/id.test.ts
```diff
@@ -108,4 +109,37 @@ describe('State node IDs', () => {
       }
     });
   });
+
+  it('should work with IDs that have escaped periods', () => {
```
Andarist, line 113: I don't think this test title is accurate - this test should be renamed (and maybe even moved?). This focuses on a state key that contains such escaped dots. 

A test case of this exact test title is needed though but it's a different test case:
```ts
      initial: 'start',
      states: {
        start: {
          on: {
            escaped: '#xyz\\.bar',
            unescaped: '#xyz.bar2'
          }
        },
        foo: {
          id: 'xyz',
          initial: 'bar1',
          states: {
            bar1: {},
            bar2: {}
          }
        },
        baz: {
          id: 'xyz\\.bar'
        },
      }
    });
```

packages/core/src/utils.ts
```diff
@@ -56,7 +56,9 @@ export function toStatePath(stateId: string | string[]): string[] {
     return stateId;
   }
 
-  return stateId.split(STATE_DELIMITER);
+  // Split on periods, but not if they are escaped
+  // i.e. match a period only if it is not preceded by a backslash.
+  return stateId.split(/(?<!\\)\./g);
```
Andarist, line 61: note: a negative lookbehind requires ES2018

packages/core/src/stateUtils.ts
```diff
@@ -216,25 +216,28 @@ export function getCandidates<TEvent extends EventObject>(
...
           );
         }
 
-        const partialEventTokens = descriptor.split('.');
-        const eventTokens = receivedEventType.split('.');
+        // split the event descriptor from unescaped periods
+        // e.g. 'foo.bar' is ['foo','bar']
+        // e.g. 'foo\.bar' is ['foo\.bar']
+        const partialEventTokens = eventDescriptor.split(/(?<!\\)\./g);
+        const eventTokens = receivedEventType.split(/(?<!\\)\./g);
```
Andarist, line 240: This requires extra test cases. I'm not completely sure if that's needed here at all. A test case would help me to clarify this. It feels like, in here, we might always treat dots in a special way.

packages/core/src/utils.ts
Andarist: Correct me if I'm wrong but this doesn't look correct.

An escaping character escapes the *next* character so the next character loses its special meaning. So while `\.` escapes a dot, this `\\.` should escape the backslash character and the dot should stay unescaped.

Since we are going between languages~ (JS -> XState), in the source code that would actually be `\\\\.` (just how `\.` in the source code is `\\.`)

packages/core/src/stateUtils.ts
davidkpiano: Oops, yeah these should be removed

packages/core/src/stateUtils.ts
```diff
@@ -578,10 +581,12 @@ export function getStateNode(
       `Unable to retrieve child state '${stateKey}' from '${stateNode.id}'; no child states exist.`
     );
   }
-  const result = stateNode.states[stateKey];
+  // replace escaped period ('\.') with just a period
+  const resolvedStateKey = stateKey.replace(/\\\./g, '.');
+  const result = stateNode.states[resolvedStateKey];
```
Andarist, line 586: There is no test case that would fail if I revert this change:
```suggestion
  const result = stateNode.states[stateKey];
```

PR #4693, opened 2024-01-20
Add inspection event for microsteps

davidkpiano:
This PR adds the `@xstate.microstep` inspection event, which allows inspectors to inspect microsteps within a transition. This is useful for observing:

- "Transient events" (raised/eventless)
- Enabled transitions
- Guards that evaluated to `true`
- Actions that were executed


packages/core/test/inspect.test.ts
```diff
@@ -6,36 +6,45 @@ import {
...
-      event: inspectionEvent.event
-    };
-  }
-  if (inspectionEvent.type === '@xstate.actor') {
-    return {
-      type: inspectionEvent.type,
-      actorId: inspectionEvent.actorRef.sessionId
-    };
-  }
+function simplifyEvents(inspectionEvents: InspectionEvent[]) {
```
Andarist, line 17: I'd recommend adding a `filter` parameter here or something and using this utility across the whole test file. This way every test case will be structured in the same way - and that makes it easier to copy-paste things around or just make bulk changes to this file.

packages/core/src/stateUtils.ts
```diff
@@ -1616,11 +1631,12 @@ export function macrostep(
         status: 'stopped'
       }
     );
-    states.push(nextSnapshot);
+    // states.push(nextSnapshot);
```
Andarist, line 1634:
```suggestion
```

packages/core/src/stateUtils.ts
```diff
@@ -1642,10 +1658,11 @@ export function macrostep(
         status: 'error',
         error: currentEvent.error
       });
-      states.push(nextSnapshot);
+      addMicrostate(nextSnapshot, currentEvent, []);
+      // states.push(nextSnapshot);
```
Andarist, line 1662:
```suggestion
```

packages/core/src/stateUtils.ts
```diff
@@ -1616,11 +1631,12 @@ export function macrostep(
         status: 'stopped'
       }
     );
-    states.push(nextSnapshot);
+    // states.push(nextSnapshot);
+    addMicrostate(nextSnapshot, event, []);
```
Andarist, line 1635: I wonder if this wouldn't be better served by some "stop" event - this call site is unique to state machines but what about stopped promises etc?

packages/core/src/stateUtils.ts
```diff
@@ -1688,7 +1705,7 @@ export function macrostep(
       internalQueue
     );
     shouldSelectEventlessTransitions = nextSnapshot !== previousState;
-    states.push(nextSnapshot);
+    addMicrostate(nextSnapshot, nextEvent, enabledTransitions);
```
Andarist, line 1708: it would be great to somehow explicitly differentiate between raised microstates and the "always" microstates - that's a useful information for the end user

packages/core/test/inspect.test.ts
Andarist, line 1: could you add 2 test cases that will list **all** of the events received by the `inspect`?
```ts
// test1
  initial: 'a',
  states: {
    a: { on: { EV: 'b' } },
    b: {}
  }
})
const actorRef = createActor(machine).start()
actorRef.send({ type: 'EV' })
```

```ts
// test2
  initial: 'a',
  states: {
    a: { on: { EV: 'b' } },
    b: { always: 'c' },
    c: {}
  }
})
const actorRef = createActor(machine).start()
actorRef.send({ type: 'EV' })
```

packages/core/src/stateUtils.ts
davidkpiano: What might this look like?

packages/core/src/stateUtils.ts
davidkpiano: We could have `event: null` in the inspection event (matches SCXML's concept of "null events")

packages/core/test/inspect.test.ts
davidkpiano: Added (please rename the tests)

packages/core/src/stateUtils.ts
Andarist: `@xstate.actor` is already produced for all actors in `class Actor`. `'xstate.stop'` always originates in that same `class Actor` (in its `_stop`). So maybe this could just be moved closer to `Actor`? That would give us some symmetry around this

packages/core/src/stateUtils.ts
Andarist: Yeah, I was thinking about a similar thing. It would be great to know what microsteps were produced by what transitions - but if we wont be able to figure out how to "link" them on the inspector side of things then we'll just add more metadata to those events.

packages/core/src/stateUtils.ts
davidkpiano: I'm not sure how to accomplish this, but feel free to give it a try - I'm open to it

packages/core/test/inspect.test.ts
Andarist: Ok, so this matches my understanding. For a received event `EV` we get 2 inspected events (3 if we count the snapshot event): `['@xstate.event', `@xstate.microstep`]`. So this isn't about *extra* steps within the transition - but rather about *all* steps in it. 

I just wonder how this will get displayed by the inspector because this adds some noise here. Receiving an event isn't the same as processing it though (!) so there is also that. So probably nothing is wrong with how this is handled right now but I'm also really curious how this data will be presented visually.

davidkpiano: > I just wonder how this will get displayed by the inspector because this adds some noise here.

Hidden by default. I'll add some logic to correlate microstep events with snapshot events (easy because they should always happen before snapshot events), and in the events panel, you can expand individual snapshot events to see the microsteps that led to it.

packages/core/src/stateUtils.ts
```diff
@@ -1606,7 +1606,22 @@ export function macrostep(
...
+    microstate: AnyMachineSnapshot,
+    event: AnyEventObject,
+    transitions: AnyTransitionDefinition[]
+  ) {
+    actorScope.system._sendInspectionEvent?.({
+      type: '@xstate.microstep',
+      actorRef: actorScope.self,
+      event,
+      snapshot: microstate,
+      transitions
```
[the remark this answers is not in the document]
davidkpiano: Good catch; will revise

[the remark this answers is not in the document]
davidkpiano: Actually, to avoid extra processing, I'm going to send the transitions as-is. It is up to the inspector to get what it needs. Unless there's a better reason why we should do processing up-front?

packages/core/src/stateUtils.ts
davidkpiano: I realized that we already differentiate - we can see what kind of `eventType` caused the transition on `inspectionEvent.transitions[0].eventType`. This is actually more useful than `event: null` or specifying whether an event is from an always-transition or not because the transition object is more specific about that (can be an always event, or a wildcard, or a partial wildcard...)

packages/core/src/system.ts
```diff
@@ -190,17 +201,14 @@ export function createSystem<T extends ActorSystemInfo>(
...
-    _sendInspectionEvent: (event) => {
-      const resolvedInspectionEvent: InspectionEvent = {
-        ...event,
-        rootId: rootActor.sessionId
-      };
-      observers.forEach((observer) => observer.next?.(resolvedInspectionEvent));
+      inspectionObservers.add(observer);
+
+      // Only send inspection events if someone is listening
+      system._sendInspectionEvent = sendInspectionEvent as any;
```
Andarist, line 207: It would make sense to make `system._sendInspectionEvent` non-optional and return early there based on `!inspectionObservers.size`

davidkpiano:
@Andarist Is this good to go?

packages/core/src/stateUtils.ts
```diff
@@ -1518,12 +1518,30 @@ function resolveActionsAndContextWorker(
...
+          params: actionParams
+        }
+      });
+      resolvedAction(actionArgs, actionParams);
+    }
+
     if (!('resolve' in resolvedAction)) {
       if (actorScope.self._processingStatus === ProcessingStatus.Running) {
-        resolvedAction(actionArgs, actionParams);
+        // Execute action
```
Andarist, line 1540:
```suggestion
```

packages/core/src/stateUtils.ts
```diff
@@ -1461,7 +1461,7 @@ interface BuiltinAction {
   execute: (actorScope: AnyActorScope, params: unknown) => void;
 }
 
-function resolveActionsAndContextWorker(
+function resolveAndExecuteActionsWithContext(
```
Andarist, line 1464: nit: I don't understand the new name - what does the `withContext` bit refer to? 🤔

packages/core/src/stateUtils.ts
```diff
@@ -1518,12 +1518,30 @@ function resolveActionsAndContextWorker(
...
+      actorScope.system._sendInspectionEvent({
+        type: '@xstate.action',
+        actorRef: actorScope.self,
+        action: {
+          type:
+            typeof action === 'string'
+              ? action
+              : typeof action === 'object'
+                ? action.type
+                : action.toString(),
```
Andarist, line 1531: How helpful is stringifying the function? I'd be more content with a simpler version like this one: 
```suggestion
                : action.name || '(anonymous)',
```

packages/core/src/stateUtils.ts
```diff
@@ -1606,7 +1624,22 @@ export function macrostep(
...
+    microstate: AnyMachineSnapshot,
+    event: AnyEventObject,
+    transitions: AnyTransitionDefinition[]
+  ) {
+    actorScope.system._sendInspectionEvent({
+      type: '@xstate.microstep',
+      actorRef: actorScope.self,
+      event,
+      snapshot: microstate,
+      transitions
```
Andarist, line 1639: To err on the safe side of things I'd rename this to `_transitions`:
```suggestion
      _transitions: transitions
```

What information do you plan to use from those on the inspector's side?

Andarist (approved):

packages/core/src/stateUtils.ts
davidkpiano: Execute the actions with the latest context - just made the name more clear

Andarist: I still don't get it 😅 I'm not sure what is the context that is provided here - or if this refers to the `snapshot.context` or what. The `*Worker` convention from some other codebase cause I couldn't figure out a specific name for it. I kinda wanted to name 2 functions the same way - the `*Worker` was just a helper function.

I'm open to renaming this since the old name was definitely not perfect but the `*WithContext` confuses me here (especially given that we have `snapshot.context` and I don't see how that fits here).

Andarist: I think I misread the new name slightly at the beginning... but now I find the new name even more inaccurate :D this function *does* resolve both actions and context. I think I see how it could be named what you have named it but at that moment the "outer" call should also be somehow renamed. I don't see what makes this one to be called `resolveAndExecuteActionsWithContext` while the other one is called `resolveActionsAndContext`.

PR #4695, opened 2024-01-22
Merge `createActorContext` options with Provider options

davidkpiano:
Options in `createActorContext` are now properly merged with provider options. Previously, provider options replaced the actor options.

```tsx
const { inspect } = createBrowserInspector();

const SomeContext = createActorContext(someMachine, { inspect });

// ...
// Options are now merged:
// { inspect: inspect, input: 10 }
<SomeContext.Provider options={{ input: 10 }}>
  {/* ... */}
</SomeContext.Provider>;
```

### DIFF
--- a/packages/xstate-react/src/createActorContext.ts
+++ b/packages/xstate-react/src/createActorContext.ts
@@ -5,7 +5,7 @@ import { useSelector as useSelectorUnbound } from './useSelector';
 
 export function createActorContext<TLogic extends AnyActorLogic>(
   actorLogic: TLogic,
-  interpreterOptions?: ActorOptions<TLogic>
+  actorOptions?: ActorOptions<TLogic>
 ): {
   useSelector: <T>(
     selector: (snapshot: SnapshotFrom<TLogic>) => T,
@@ -30,7 +30,7 @@ export function createActorContext<TLogic extends AnyActorLogic>(
     children,
     logic: providedLogic = actorLogic,
     machine,
-    options: providedOptions = interpreterOptions
+    options: providedOptions = actorOptions
   }: {
     children: React.ReactNode;
     logic: TLogic;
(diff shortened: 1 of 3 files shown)


packages/xstate-react/src/createActorContext.ts
```diff
@@ -30,7 +30,7 @@ export function createActorContext<TLogic extends AnyActorLogic>(
     children,
     logic: providedLogic = actorLogic,
     machine,
-    options: providedOptions = interpreterOptions
+    options: providedOptions = actorOptions
```
Andarist, line 33: When we don't remove this default we often end up merging object A with object A (this becomes `{ ...actorOptions, ...actorOptions }`) and that bothers me slightly ;p
```suggestion
    options: providedOptions
```

We'll still be creating somewhat redundant copies of those options here since we spread them unconditionally but that should be fine an I'm OK with it.

To avoid those redundant copies we could have a `merge` util like:
```ts
function merge<T>(a: T | undefined, b: T | undefined): T | undefined {
  if (a === undefined) {
    return b;
  }
  if (b === undefined) {
    return a;
  }
  return {
    ...a,
    ...b,
  };

```

packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -470,4 +470,40 @@ describe('createActorContext', () => {
...
+
+    const Component = () => {
+      const count = SomeContext.useSelector((state) => state.context.count);
+
+      return <div data-testid="value">{count}</div>;
+    };
+
+    const App = () => {
+      return (
+        <SomeContext.Provider options={{ input: 10 }}>
```
Andarist, line 499: This test doesn't check that those `options` here were used - if the test is meant to check the merging behavior then we should have assertions validating that both of those are preserved somehow.

davidkpiano: Improved the test to check that `count: 10`

davidkpiano: It's already testing the top-level option by reading the `events` array

Andarist (approved):

Issue #4662, opened 2024-01-09
Bug: XState source code can't be typechecked with TS 5.4


Andarist:
### XState version

XState version 5

### Description

Our codebase doesn't typecheck with the upcoming TS 5.4 right now

### Expected result

`yarn typecheck` should work

### Actual result

```
packages/core/src/setup.ts:190:5 - error TS2344: Type 'ToStateValue<TConfig>' does not satisfy the constraint 'StateValue'.
  Type 'ToStateValue<MachineConfig<TContext, TEvent, ToProvidedActor<TChildrenMap, TActors, DefaultToUnknownActorLogic<TActors>>, ... 6 more ..., ResolveTypegenMeta<...>>>' is not assignable to type 'StateValue'.
    Type '{} | { [x: string]: {} | { [x: string]: {} | ... | Required<...> | undefined; } | Required<{ [x: string]: {} | ... | Required<...> | undefined; }> | undefined; }' is not assignable to type 'StateValue'.
      Type '{ [x: string]: {} | { [x: string]: {} | ... | Required<...> | undefined; } | Required<{ [x: string]: {} | ... | Required<...> | undefined; }> | undefined; }' is not assignable to type 'StateValue'.

190     ToStateValue<TConfig>,
        ~~~~~~~~~~~~~~~~~~~~~


Found 1 error in packages/core/src/setup.ts:190
```

### Reproduction

yarn typecheck

### Additional context

I bisected this problem to 5.4.0-dev.20231121 and thus to this changeset: https://github.com/microsoft/TypeScript/compare/c10edfbc8dd840c7e5d25b0e2ede2697ef4ac4b6...c266e47794fbd187425989448670238576523d4e

rjbrache:
I'm seeing a similar issue with XState v5.4.1 and the current TS 5.3.3

node_modules/xstate/dist/declarations/src/State.d.ts:5:49 - error TS2344: Type 'NonNullable<TStateValue[K]>' does not satisfy the constraint 'StateValue'.
  Type 'NonNullable<TStateValue[K]>' is not assignable to type 'StateValueMap'.
    Index signature for type 'string' is missing in type '{}'.

5     [K in keyof TStateValue]?: ToTestStateValue<NonNullable<TStateValue[K]>>;

Andarist:
@rjbrache do you have `strictNullChecks` enabled?

rjbrache:
strictNullChecks is not enabled.

Andarist:
[...]

PR #4706, opened 2024-01-26
Fixed compatibility with the upcoming TS 5.4

Andarist:
fixes https://github.com/statelyai/xstate/issues/4662

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -295,7 +295,7 @@ export type Actions<
 export type StateKey = string | AnyMachineSnapshot;
 
 export interface StateValueMap {
-  [key: string]: StateValue;
+  [key: string]: StateValue | undefined;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -2352,21 +2352,19 @@ export type ToChildren<TActor extends ProvidedActor> =
       // or maybe even `TActor["logic"]` since it's possible to configure `{ src: string; logic: SomeConcreteLogic }`
       // TODO: consider adding `| undefined` here
       Record<string, AnyActorRef>
-    : Compute<
```
Andarist, line 2355: Removing `Compute` might degrade type displays *slightly* but I wasn't able to fix this in any different way. I think this might be a bug in TS that this doesn't typecheck - I reported it here: https://github.com/microsoft/TypeScript/issues/57188

davidkpiano (approved):

PR #4717, opened 2024-01-31
[chore] refactored `guards` internals to standardize built-in guard definitions and clarify logic flow

Enteleform:
- [fixed typo in error message](https://github.com/statelyai/xstate/pull/4717/commits/4262d1126df873d59c978f8f7ae475e2495cb527)
- [added documentation for `stateIn`](https://github.com/statelyai/xstate/pull/4717/commits/f3f3f2fdf39f86c8b62cc1b9348ca55c0e677029)
- [refactored `evaluateGuard` to clarify logic flow](https://github.com/statelyai/xstate/pull/4717/commits/857463b3316c60012e82d6d519816e343325a377)
- [consolidated repeated definitions into `LogicGuards` type](https://github.com/statelyai/xstate/pull/4717/commits/86300daa985171737fabc3d7c47696c9f6a8d505)
- [implemented abstractions to standardize built-in guard definitions](https://github.com/statelyai/xstate/pull/4717/commits/eabe830604b4e7e3cc616a8532a082fff34e6cb9)
- [added changeset](https://github.com/statelyai/xstate/pull/4717/commits/d7ca549b92f84c3f3c5b69fcd7f70200ec2e2e60)

### DIFF
--- a/.changeset/friendly-goats-jam.md
+++ b/.changeset/friendly-goats-jam.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
+refactored `guards` internals to standardize built-in guard definitions and clarify logic flow
--- a/packages/core/src/guards.ts
+++ b/packages/core/src/guards.ts
@@ -163,10 +163,10 @@ function checkNot(
 }
 
 /**
- * Higher-order guard that evaluates to `true` if the `guard` passed to it evaluates to `false`.
- *
- * @category Guards
- * @example
+* Higher-order guard that evaluates to `true` if the `guard` passed to it evaluates to `false`.
+*
+* @category Guards
+* @example
   ```ts
   import { setup, not } from 'xstate';
 
@@ -186,8 +186,8 @@ function checkNot(
     }
   });
   ```
- * @returns A guard
- */
+* @returns A guard
+*/
 export function not<
   TContext extends MachineContext,
   TExpressionEvent extends EventObject,
@@ -224,11 +224,11 @@ function checkAnd(
 }
 
 /**
- * Higher-order guard that evaluates to `true` if all `guards` passed to it
- * evaluate to `true`.
- *
- * @category Guards
- * @example
+* Higher-order guard that evaluates to `true` if all `guards` passed to it
+* evaluate to `true`.
+*
+* @category Guards
+* @example
   ```ts
   import { setup, and } from 'xstate';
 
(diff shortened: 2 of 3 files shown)


packages/core/src/guards.ts
```diff
@@ -163,10 +163,10 @@ function checkNot(
 }
 
 /**
```
Enteleform, line 194: Seems like the Prettier output changed some whitespace throughout the document for some reason.  Go to the next comment for the actual updates 👇

packages/core/src/guards.ts
```diff
@@ -364,58 +364,40 @@ export function evaluateGuard<
   TContext extends MachineContext,
   TExpressionEvent extends EventObject
 >(
```
Enteleform, line 343: Updates start here 🎯

.changeset/friendly-goats-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': minor
+---
+
+refactored `guards` internals to standardize built-in guard definitions and clarify logic flow
```
with-heart, line 5: When adding a changeset, you want to document changes to the public api since those changes affect users and these entries are added to the changelog. 

Basically imagine yourself as a user reading the changelog entry for this change. You'd want to know:

- what changed? 
- why did it change? 
- how do I use the change? 

If there are no changes to the public api (I'm not sure), you shouldn't add a changeset.

Andarist:
I'm not the biggest fan of those changes to how guards are structured here. I don't see big issues with the current structure and I don't think this is a problem worth solving. it could also be seen as a breaking change (in quite an edge case scenarios but still). Also, the fact that each guard had its own "not callable builtin guard" with a unique `.name` that repeated the guard's name was very much intentional.

.changeset/friendly-goats-jam.md
Enteleform: Ah, gotcha.  It was generated during step 7 of [CONTRIBUTING #making-changes](https://github.com/statelyai/xstate/blob/main/CONTRIBUTING.md#making-changes).  The instructions don't give any criteria for when to include a changeset or not, they just say to generate one.

There are no changes to the public API in this PR.  Do you recommend removing the changeset?

Andarist:
I don't think we are going to proceed with this particular PR. I'm especially opposed to [this change](https://github.com/statelyai/xstate/pull/4717/commits/eabe830604b4e7e3cc616a8532a082fff34e6cb9). Maybe some other changes from this PR could be pulled in but I don't feel a strong need for this either way. I'm sorry for closing this as I know that you put some effort into this - we really appreciate that!

PR #4724, opened 2024-02-04
Add ability to spawn actor from any logic

davidkpiano:
This PR adds the ability for any actor logic to spawn an actor, which is a core tenet of the actor model.


packages/core/src/types.ts
```diff
@@ -2106,6 +2106,7 @@ export interface ActorScope<
   defer: (fn: () => void) => void;
   system: TSystem;
   stopChild: (child: AnyActorRef) => void;
+  spawn: (logic: AnyActorLogic) => AnyActorRef;
```
Andarist, line 2109:
```suggestion
  spawnChild: (logic: AnyActorLogic) => AnyActorRef;
```

Andarist, line 2109: This has to become generic so people could actually have some type safety here

packages/core/src/createActor.ts
```diff
@@ -178,6 +178,13 @@ export class Actor<TLogic extends AnyActorLogic>
           );
         }
         (child as any)._stop();
+      },
+      spawn: (logic) => {
+        const actor = createActor(logic, {
+          parent: this
+        });
+
+        return actor;
```
Andarist, line 187: when it is meant to be started?

davidkpiano: I knew I was forgetting something!

packages/core/src/createActor.ts
```diff
@@ -178,6 +179,21 @@ export class Actor<TLogic extends AnyActorLogic>
...
+        actorOptions?: ActorOptions<T>
+      ) => {
+        const actor = createActor(logic, {
+          parent: this,
+          ...actorOptions
+        });
+
+        if (this._processingStatus === ProcessingStatus.Running) {
+          actor.start();
+        }
```
Andarist, line 194: what about the actors spawned when `this._processingStatus === ProcessingStatus.NotStarted`?

packages/core/src/createActor.ts
```diff
@@ -178,6 +179,21 @@ export class Actor<TLogic extends AnyActorLogic>
           );
         }
         (child as any)._stop();
+      },
+      spawnChild: <T extends AnyActorLogic>(
+        logic: T,
+        actorOptions?: ActorOptions<T>
```
Andarist, line 185: this should reuse the type tricks from here: https://github.com/statelyai/xstate/blob/8a7ea9eb5faca407890cfa597eca2915660ab6bd/packages/core/src/createActor.ts#L763-L770

packages/core/src/actors/callback.ts
```diff
@@ -73,6 +75,7 @@ export type InvokeCallback<
    * the listener is then called whenever events are received by the callback actor
    */
   receive: Receiver<TEvent>;
+  spawn: ActorScope<any, any>['spawnChild'];
```
Andarist, line 78: shouldn't all of those actor logics receive `stopChild` as well?

Andarist (changes requested):
There is a substantial problem with this PR that is somewhat mentioned in this comment here:
https://github.com/statelyai/xstate/blob/91549ce8a6a5105b3d354ca442ea5c058e58e206/packages/core/src/createActor.ts#L588-L592

When the parent actor gets stopped its children should get stopped too. There is no mechanism implemented here that would achieve this for this new functionality.

davidkpiano:
> 831aeec

[831aeec](https://github.com/statelyai/xstate/pull/4724/commits/831aeecc9a0a17f297dccfac4c2099be4d328d07)

davidkpiano:
> There is a substantial problem with this PR that is somewhat mentioned in this comment here:
> 
> https://github.com/statelyai/xstate/blob/91549ce8a6a5105b3d354ca442ea5c058e58e206/packages/core/src/createActor.ts#L588-L592
> 
> When the parent actor gets stopped its children should get stopped too. There is no mechanism implemented here that would achieve this for this new functionality.

[831aeec](https://github.com/statelyai/xstate/pull/4724/commits/831aeecc9a0a17f297dccfac4c2099be4d328d07)

davidkpiano:
Bringing this to v6.

PR #4728, opened 2024-02-05
Fixed types compatibility with TS 5.0

Andarist:


packages/core/src/types.ts
```diff
@@ -2348,19 +2348,21 @@ export type ToChildren<TActor extends ProvidedActor> =
...
-          include: {
-            [id: string]: TActor extends any
-              ? ActorRefFrom<TActor['logic']> | undefined
-              : never;
-          };
-          exclude: {};
-        }[undefined extends TActor['id'] // if not all actors have literal string IDs then we need to create an index signature containing all possible actor types
-          ? 'include'
-          : string extends TActor['id']
+    : Compute<
```
Andarist, line 2351: This just brings back `Compute` here so it's essentially a revert of this change https://github.com/statelyai/xstate/pull/4706/files#r1467593357

packages/core/src/setup.ts
```diff
@@ -61,7 +61,7 @@ type ToProvidedActor<
     id: IsNever<TChildrenMap> extends true
       ? string | undefined
       : K extends keyof Invert<TChildrenMap>
-        ? Invert<TChildrenMap>[K]
+        ? Invert<TChildrenMap>[K] & string
```
Andarist, line 64: this is the true and correct fix for what I previously fixed by removing `Compute`. You can read my breakdown of the issue [here](https://github.com/microsoft/TypeScript/issues/57188#issuecomment-1926882621)

PR #4731, opened 2024-02-07
Expose getInitialSnapshot

davidkpiano:
You can now import `getInitialSnapshot(…)` from `xstate` directly:

```ts
import { getInitialSnapshot } from 'xstate';
```


.changeset/clean-brooms-yell.md
```diff
@@ -0,0 +1,9 @@
+---
+'xstate': patch
+---
+
+You can now import `getInitialSnapshot(…)` from `xstate` directly:
```
Andarist, line 5: it would be great if the changeset would mention what this is even useful for

Issue #4734, opened 2024-02-08
Bug: Actor Unions have typing issues related when calling methods


devanfarrell:
### XState version

XState version 5

### Description

Unions of actor types have issues when it comes to calling methods like `can` or `matches`.

### Expected result

A union type of actors that has the same states and events shouldn't complain if you try to use `can` or `matches`.

### Actual result

Typescript can't figure out that they are essentially the same outside of the context object and requires a `ts-expect-error` to use the methods due to the `this` property not being the same.

### Reproduction

[...]

PR #4739, opened 2024-02-10
#4734 bugfix: removed `this` from snapshot methods

devanfarrell:
Resolves https://github.com/statelyai/xstate/issues/4734. I tested locally using the same scenario as used in the issue and the types behaved as expected. I couldn't really figure out a way to programmatically prove work and test the types directly. Please advise if there is something you'd like to see there.

### DIFF
--- a/.changeset/fifty-beds-love.md
+++ b/.changeset/fifty-beds-love.md
@@ -2,4 +2,4 @@
 'xstate': patch
 ---
 
-Removed `this` from actor snapshot methods to resolve [#4734](https://github.com/statelyai/xstate/issues/4734)
+Removed `this` from machine snapshot methods to fix issues with accessing those methods from union of actors and their snapshots.
--- a/.changeset/fifty-beds-love.md
+++ b/.changeset/fifty-beds-love.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Removed `this` from actor snapshot methods to resolve [#4734](https://github.com/statelyai/xstate/issues/4734)
--- a/packages/core/test/types.test.ts
+++ b/packages/core/test/types.test.ts
@@ -17,7 +17,8 @@ import {
   not,
   sendTo,
   spawnChild,
-  stateIn
+  stateIn,
+  setup
 } from '../src/index';
 
 function noop(_x: unknown) {
@@ -4339,24 +4340,6 @@ describe('self', () => {
       })
     });
   });
-
-  it('should allow actor union snap snapshot methods to be called', () => {
-    const typeOne = createMachine({
-      types: {} as { events: { type: 'one' } }
-    });
-    type TypeOneRef = ActorRefFrom<typeof typeOne>;
-
-    const typeTwo = createMachine({
-      types: {} as { events: { type: 'one' } | { type: 'two' } }
-    });
-    type TypeTwoRef = ActorRefFrom<typeof typeTwo>;
-
-    const canMethod = (ref: TypeOneRef | TypeTwoRef) => {
-      ref.getSnapshot().can({ type: 'one' });
-    };
-
-    canMethod(createActor(typeOne));
-  });
 });
 
 describe('createActor', () => {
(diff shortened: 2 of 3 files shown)


davidkpiano (approved):

Andarist:
>  I couldn't really figure out a way to programmatically prove work and test the types directly. Please advise if there is something you'd like to see there.

You can add a test case in [this file](https://github.com/statelyai/xstate/blob/91549ce8a6a5105b3d354ca442ea5c058e58e206/packages/core/test/types.test.ts).

devanfarrell:
Added the regression test, confirmed that it fails on main.

packages/core/test/types.test.ts
```diff
@@ -4339,6 +4339,24 @@ describe('self', () => {
       })
     });
   });
+
+  it('should allow actor union snap snapshot methods to be called', () => {
```
Andarist, line 4343: This was added to the `describe` block that is about `self`. Could you create a new `describe` block and move this test into it?

devanfarrell: Improved the tests as well. Something I learned that may or may not be intentional. If you don't call setup, the `snapshot.matches` method parameters don't infer the different possible machine states.

Andarist: > If you don't call setup, the snapshot.matches method parameters don't infer the different possible machine states.

Supporting this in the bare `createMachine` would make us require TS 5.3 (which is the current latest version) so we decided not to do that there. The problem is that it would have negative effects on users of other TS versions.

packages/core/test/types.test.ts
```diff
@@ -4357,3 +4358,49 @@ describe('createActor', () => {
...
+      }
+    });
+    type TypeTwoRef = ActorRefFrom<typeof typeTwo>;
+
+    const ref = createActor(typeTwo) as TypeOneRef | TypeTwoRef;
+    const snapshot = ref.getSnapshot();
+
+    snapshot.can({ type: 'one' });
+    // @ts-expect-error
+    snapshot.can({ type: 'three' });
```
Andarist, line 4396: shouldn't we test `'one'` and `'two'` in all of those assertions? `'three'` is incompatible with either - it's still a valuable test since it might test that the argument isn't `any`. However, it would also be good to test that only a subset of the "inputs" is allowed here (`'one'` should be OK, and `'two'` should not be)

packages/core/test/types.test.ts
```diff
@@ -4357,3 +4358,58 @@ describe('createActor', () => {
...
+    // @ts-expect-error should error both snapshots don't match
+    snapshot.hasTag('two');
+    // @ts-expect-error should error if no snapshot matches
+    snapshot.hasTag('three');
+
+    snapshot.matches('one');
+    // @ts-expect-error should error both snapshots don't match
+    snapshot.matches('two');
+    // @ts-expect-error should error if no snapshot matches
+    snapshot.matches('three');
```
Andarist, line 4410:
```suggestion
    // @ts-expect-error
    snapshot.can({ type: 'two' });
    // @ts-expect-error
    snapshot.can({ type: 'three' });

    snapshot.hasTag('one');
    // @ts-expect-error
    snapshot.hasTag('two');
    // @ts-expect-error
    snapshot.hasTag('three');

    snapshot.matches('one');
    // @ts-expect-error
    snapshot.matches('two');
    // @ts-expect-error
    snapshot.matches('three');
```

.changeset/fifty-beds-love.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Removed `this` from actor snapshot methods to resolve [#4734](https://github.com/statelyai/xstate/issues/4734)
```
Andarist, line 5:
```suggestion
Removed `this` from machine snapshot methods to fix issues with accessing those methods from union of actors and their snapshots.
```

Andarist:
Thanks ❤️

PR #4744, opened 2024-02-16
Fix log example on readme super quick start section

LuccaBessa:
The count on the line 156 was 1 instead of 2

### DIFF
--- a/.changeset/friendly-spiders-cheer.md
+++ b/.changeset/friendly-spiders-cheer.md
@@ -1,5 +1,0 @@
----
-'xstate': major
----
-
-Fix log comment on README.md to the correct count number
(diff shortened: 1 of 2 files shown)


.changeset/friendly-spiders-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': major
+---
+
+Fix log comment on README.md to the correct count number
```
Andarist, line 5:
```suggestion
```

README.md
```diff
@@ -153,7 +153,7 @@ toggleActor.send({ type: 'TOGGLE' });
 // => logs 'active', { count: 1 }
 
 toggleActor.send({ type: 'TOGGLE' });
-// => logs 'inactive', { count: 1 }
+// => logs 'inactive', { count: 2 }
```
Andarist, line 156: This is not correct. This machine doesn't increment the counter when transitioning from the active state to the inactive state so the counter's value stays the same

PR #4746, opened 2024-02-18
[core] Event emitter

davidkpiano:
The new `emit(…)` action creator emits events that can be received by listeners. Actors are now event emitters.

```ts
import { emit } from 'xstate';

  // ...
  on: {
    something: {
      actions: emit({
        type: 'emitted',
        some: 'data'
      })
    }
  }
  // ...
});

actor.on('emitted', (event) => {
  console.log(event);
});

actor.send({ type: 'something' });
// logs:
// {
//   type: 'emitted',
//   some: 'data'
// }
```


packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,152 @@
...
+  const { event, delay, id } = params;
+  // if (typeof delay === 'number') {
+  actorScope.defer(() => {
+    actorScope.emit(event);
+  });
+  return;
+  // }
+}
+
+export interface RaiseAction<
```
Andarist, line 96:
```suggestion
export interface EmitAction<
```

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,152 @@
...
+      | EventObject
+      | SendExpr<
+          MachineContext,
+          EventObject,
+          ParameterizedObject['params'] | undefined,
+          EventObject,
+          EventObject
+        >;
+    id: string | undefined;
+    delay:
```
Andarist, line 37: If u plan to support `delay` here then you will have to account for that in the whacky persistence that we have today. Since this is additive I'd start without it to simplify things. And at that point you won't need `id` either

packages/core/src/types.ts
```diff
@@ -1952,6 +1952,10 @@ export interface ActorRef<
   /** @internal */
   _processingStatus: ProcessingStatus;
   src: string | AnyActorLogic;
+  on: (
+    emittedEventType: string,
+    handler: (emittedEvent: AnyEventObject) => void
```
Andarist, line 1957: Nice try with that `AnyEventObject` 🤣 We'll have to introduce `TNotification` in this PR an pass it around and around 😢

packages/core/src/createActor.ts
```diff
@@ -178,6 +181,12 @@ export class Actor<TLogic extends AnyActorLogic>
           );
         }
         (child as any)._stop();
+      },
+      emit: (emittedEvent) => {
+        for (const handler of this.eventListeners.get(emittedEvent.type) ??
```
Andarist, line 186: This should use `.slice()` to make the copy of the **current** listeners. We don't want to notify the listeners that might get added while notifying this listeners list etc

Andarist: Perhaps not `.slice` since that's a Set from what I can see but you know what I mean :p

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,152 @@
...
+      typeof configDelay === 'function'
+        ? configDelay(args, actionParams)
+        : configDelay;
+  } else {
+    resolvedDelay =
+      typeof delay === 'function' ? delay(args, actionParams) : delay;
+  }
+  if (typeof resolvedDelay !== 'number') {
+    internalQueue.push(resolvedEvent);
+  }
```
Andarist, line 75: This is wrong because it would send that as a raised event to the current machine. It would be good to add a test case for this (to test that it doesn't happen)
```suggestion
```

packages/core/src/createActor.ts
```diff
@@ -397,6 +406,22 @@ export class Actor<TLogic extends AnyActorLogic>
     };
   }
 
+  public on(
+    emittedEventType: string,
+    handler: EmittedEventHandler
+  ): Subscription {
+    const set = this.eventListeners.get(emittedEventType) ?? new Set();
+    set.add(handler);
```
Andarist, line 414: You have to wrap the input handler because otherwise you end up with weird inconsistencies:
```ts
actorRef.on('foo', handler)
const sub = actorRef.on('foo', handler)

sub.unsubscribe()

emit({ type: 'foo' })

// radio silence 😢
```

davidkpiano: Are you sure? `sub.unsubscribe()` is basically the same as a `.removeEventListener(handler)`. 

The radio silence seems like the correct behavior.

packages/core/src/State.ts
```diff
@@ -68,7 +68,8 @@ interface MachineSnapshotBase<
     TStateValue,
     TTag,
     unknown,
-    TOutput
+    TOutput,
+    any // TEmitted
```
Andarist, line 72:
```suggestion
    EventObject // TEmitted
```

packages/core/src/StateMachine.ts
```diff
@@ -65,6 +65,7 @@ export class StateMachine<
   TTag extends string,
   TInput,
   TOutput,
+  TEmitted extends EventObject,
```
Andarist, line 68: we might need to default those since `@xstate/react` and other packages might depend on the number of arguments - we already had small issues like this but we can take another look at the fixing PRs to understand if the nature of both is the same

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,106 @@
...
+      | SendExpr<
+          MachineContext,
+          EventObject,
+          ParameterizedObject['params'] | undefined,
+          EventObject,
+          EventObject
+        >;
+  }
+) {
+  if (typeof eventOrExpr === 'string') {
```
Andarist, line 31:
```suggestion
  if (isDevelopment && typeof eventOrExpr === 'string') {
```

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @param eventType The event to emit.
+ */
+export function emit<
+  TContext extends MachineContext,
+  TExpressionEvent extends EventObject,
+  TEvent extends EventObject = TExpressionEvent,
+  TParams extends ParameterizedObject['params'] | undefined =
+    | ParameterizedObject['params']
+    | undefined,
+  TEmitted extends EventObject = EventObject
```
Andarist, line 84: This is a new action type so I'd start without any defaults here
```suggestion
  TEvent extends EventObject,
  TParams extends ParameterizedObject['params'] | undefined,
  TEmitted extends EventObject
```

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,106 @@
...
+    args: ActionArgs<TContext, TExpressionEvent, TEvent>,
+    params: TParams
+  ) {
+    if (isDevelopment) {
+      throw new Error(`This isn't supposed to be called`);
+    }
+  }
+
+  emit.type = 'xstate.emit';
+  emit.event = eventOrExpr;
```
Andarist, line 100: While it is an event - we already have events that mean a different thing. What do you think about calling it differently? It gets confusing when u start saying "send an event" and "emit an event". What do you think about calling this a notification?

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,106 @@
...
+      : eventOrExpr;
+  return [snapshot, { event: resolvedEvent }];
+}
+
+function executeEmit(
+  actorScope: AnyActorScope,
+  params: {
+    event: EventObject;
+    id: string | undefined;
+    delay: number | undefined;
```
Andarist, line 48:
```suggestion
```

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+) {
+  const { event, delay, id } = params;
```
Andarist, line 51:
```suggestion
  const { event } = params;
```

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,106 @@
...
+  // if (typeof delay === 'number') {
```
Andarist, line 52:
```suggestion
```

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,106 @@
...
+  actorScope.defer(() => {
+    actorScope.emit(event);
+  });
+  return;
+  // }
```
Andarist, line 57:
```suggestion
```

packages/core/src/actions/enqueueActions.ts
Andarist, line 1: we should introduce `enqueue.emit` here

packages/core/test/types.test.ts
```diff
@@ -306,6 +306,32 @@ describe('output', () => {
   });
 });
 
+describe('emitted', () => {
+  it('emitted tpe should be represented in actor.on(…)', () => {
```
Andarist, line 310:
```suggestion
  it('emitted type should be represented in actor.on(…)', () => {
```

packages/core/src/createMachine.ts
```diff
@@ -115,6 +115,7 @@ export function createMachine<
   TTag extends string,
   TInput,
   TOutput extends NonReducibleUnknown,
+  TEmitted extends AnyEventObject,
```
Andarist, line 118:
```suggestion
  TEmitted extends EventObject,
```

packages/core/src/createActor.ts
```diff
@@ -745,11 +778,13 @@ type RequiredOptions<TLogic extends AnyActorLogic> =
 export function createActor<TLogic extends AnyActorLogic>(
   logic: TLogic,
   ...[options]: ConditionalRequired<
+    // Require this...
```
Andarist, line 781: perhaps this would be better suited as JSDoc for `ConditionalRequired`?

packages/core/src/createActor.ts
Andarist: The API is quite different here from the `removeEventListener` API though. With `addEventListener`/`removeEventListener` you don't get a subscription object back - so your `listener` acts as a cache key in the internal set. For that reason, if you add the same listener twice it will be called once and if you remove it once then it will remove all of the subscribers (well, the only one that was added). The actual storage algorithm is slightly more complicated because it also accounts for options - u might add the same listener with different options and IIRC that would create separate internal entries. That's beside the point though :P

With the subscription API, the situation is different because you get back a subscription object - the intuition is that this object is bound to a specific *instance* of the subscription.

This is how it always worked with RxJS and most of the other stream libraries ([stackblitz](https://stackblitz.com/edit/rxjs-uvymmo?devtoolsheight=60&file=index.ts)):
```ts
import { Subject } from 'rxjs';

const subject = new Subject();

function listener(v: unknown) {
  console.log('got', v);
}

const first = subject.subscribe({
  next: listener,
});
subject.subscribe({
  next: listener,
});

first.unsubscribe();

subject.next('hello!'); // it still gets logged
```

This is also how our own `.subscribe` API works, how Redux works (see [here](https://github.com/reduxjs/redux/blob/dfcfab2d9a399c3fe19e12a47607ced44acda6b5/src/createStore.ts#L223-L224)) and how TanStack Query v4 worked (see [here](https://github.com/TanStack/query/blob/bdc746abdbf40df809648b1501240391a06e833d/packages/query-core/src/subscribable.ts#L11-L13)). Interestingly, TanStack Query v5 has changed this behavior (see [here](https://github.com/TanStack/query/blob/55782aebbbc69020858f8ed611ab41d24eee215a/packages/query-core/src/subscribable.ts#L11-L12)).

I've also tried to take a look at the current proposed `Observable` API in Chrome Canary (spec [here](https://github.com/WICG/observable)). However... that just doesn't support `.unsubscribe` at all and each subscription just sets up a new subscription each time:
```ts
var stream$ = new Observable((s) => {
  console.log("setup");
  setTimeout(() => {
    s.next("hey!"), s.complete();
  }, 1);
});

function listener(v) {
  console.log("got", v);
}

stream$.subscribe({ next: listener });
stream$.subscribe({ next: listener });
// logs both `setup` and `got` twice
```

So this isn't exactly comparable because with this `Observable` spec we create two separate sources whereas our API is about subscribing to the same source. "Sharing" could be implemented here though - whatever you do in your setup callback is up to you, so  - in there - you could reach into a shared resource and setup ur own sharing logic etc. That sharing logic wouldn't even be able to "deduplicate" those subscriptions because the callback doesn't get any access to the user-provided callbacks. It just gets access to its own unique subscription object and only that can be used as "cache key" etc.

If we implement teardown on top of the example above we can see how it only cancels a single subscriber here:
```ts
const controller = new AbortController();

let id = 0;

const stream$ = new Observable((s) => {
  const myId = ++id;
  const timerId = setTimeout(() => {
    s.next(`hey from ${myId}!`);
  }, 1);

  s.addTeardown(() => {
    console.log("tearing down", myId);
    clearTimeout(timerIt);
  });
});

function listener(v) {
  console.log("got", v);
}

stream$.subscribe({ next: listener }, { signal: controller.signal });
stream$.subscribe({ next: listener });

controller.abort(); // this would only cancel the first one!
```

packages/core/src/types.ts
```diff
@@ -1106,7 +1161,8 @@ type MachineImplementationsActions<
     Cast<
       Prop<TIndexedDelays, keyof TIndexedDelays>,
       ParameterizedObject
-    >['type']
+    >['type'],
+    any // TODO
```
Andarist, line 1165: I can help with that

packages/core/src/createActor.ts
davidkpiano: Yeah, true

packages/core/src/actions/emit.ts
davidkpiano: Yeah, but conceptually they're still the same thing. An emitted event can be thought of as sending an event to implicit "listener actors", whose logic is simply to call a function when the event is received. Pseudocode:

```ts
function on(actor, eventType, fn) {
  const listenerActor = createListenerActor(fn);

  // the emit(...) action will send the event to all listener actors
  actor.send({
    type: 'xstate.listen',
    eventType,
    ref: listenerActor
  });

  return { ... } // subscription
}
```

In fact, we should probably model it this way in the future so we get visualization/tracing benefits etc. (maybe for v6)

packages/core/src/actions/enqueueActions.ts
davidkpiano: Done

packages/core/src/createActor.ts
davidkpiano: Makes sense; I'll wrap it then 👍 🍬

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function executeEmit(
+  actorScope: AnyActorScope,
+  params: {
+    event: EventObject;
+  }
+) {
+  const { event } = params;
+  actorScope.defer(() => {
```
Andarist, line 50: can we think of any sensible test case that would require `defer` here? I think that it's probably the right thing to use `defer` here - it just would be cool to have a test case that would prove its necessity

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,100 @@
...
+    actorScope.emit(event);
+  });
+  return;
```
Andarist, line 53:
```suggestion
```

packages/core/src/types.ts
```diff
@@ -1106,7 +1161,8 @@ type MachineImplementationsActions<
     Cast<
       Prop<TIndexedDelays, keyof TIndexedDelays>,
       ParameterizedObject
-    >['type']
+    >['type'],
+    AnyEventObject // TODO
```
Andarist, line 1165: TODO for myself: look into this

packages/core/src/actions/emit.ts
davidkpiano: There is no need; just removed it. It was a copypasta

Andarist: What snapshot should be read here? should it be the one from *before* or *after* the transition? I think making it the one *after* would be more consistent with `send`. We need `defer` for this and a test could be somewhat easily crafted for this situation
```ts
actorRef.on('foo', () => {
  actorRef.getSnapshot()
})
```

davidkpiano: Can you write a test for that?

davidkpiano: Added a test for this; it works without `defer`

davidkpiano: JK it doesn't

packages/core/src/actions/emit.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export interface EmitAction<
+  TContext extends MachineContext,
+  TExpressionEvent extends EventObject,
+  TParams extends ParameterizedObject['params'] | undefined,
+  TEvent extends EventObject,
+  TEmitted extends EventObject
+> {
+  (args: ActionArgs<TContext, TExpressionEvent, TEvent>, params: TParams): void;
+  _out_TEvent?: TEvent;
```
Andarist, line 63: I think that this one shouldnt be needed here:
```suggestion
```

Issue #4743, opened 2024-02-16
Bug: Compilation Error when using setup and enqueueActions


FabioGimmillaro:
### XState version

XState version 5

### Description

Hi,

I'm currently having an issue using the enqueueActions Action in combination with the new setup function.

I have reproduced this issue with a small machine:
```typescript
export const brokenMachine = setup(
  {
    types: {
      context: {} as any,
      events: {} as {} as
        | ({ type: "someEvent" }),
      input: {} as any
    },
    actions: {},
    actors: {},
    guards: {},
    delays: {},
  }).createMachine(
  {
    id: "Game",
    initial: "default",
    on: {
      someEvent: { // <== Error is here
        actions: enqueueActions(() => {})
      },
    },
    states: {},
  },
);
```
I am getting following error:
```
 Types of property 'actions' are incompatible.
    Type 'EnqueueActionsAction<any, { type: "someEvent"; }, { type: "someEvent"; }, { src: string; logic: UnknownActorLogic; id: string | undefined; }, { type: string; params: NonReducibleUnknown; }, { .
..; }, string>' is not assignable to type 'Actions<any, { type: "someEvent"; }, { type: "someEvent"; }, undefined, { src: string; logic: UnknownActorLogic; id: string | undefined; }, { type: string; para
ms: NonReducibleUnknown; }, { ...; }, never> | undefined'.
      Type 'EnqueueActionsAction<any, { type: "someEvent"; }, { type: "someEvent"; }, { src: string; logic: UnknownActorLogic; id: string | undefined; }, { type: string; params: NonReducibleUnknown; }, {
 ...; }, string>' is not assignable to type 'ActionFunction<any, { type: "someEvent"; }, { type: "someEvent"; }, undefined, { src: string; logic: UnknownActorLogic; id: string | undefined; }, { type: str
ing; params: NonReducibleUnknown; }, { ...; }, never>'.
        Types of property '_out_TDelay' are incompatible.
          Type 'string | undefined' is not assignable to type 'undefined'.
            Type 'string' is not assignable to type 'undefined'.

253       someEvent: {
          ~~~~~~~~~

ms: NonReducibleUnknown; }, { ...; }, never> | undefined'.
      Type 'EnqueueActionsAction<any, { type: "someEvent"; }, { type: "someEvent"; }, { src: string; logic: UnknownActorLogic; id: string | undefined; }, { type: string; params: NonReducibleUnknown; }, {
[...]

Issue #4545, opened 2023-12-03
Bug: Enqueued Actions throw type error inside setup()


KrisCoulson:
### Description

When using the new setup api to provide actions when you use the enqueueActions action typescript throws

### Expected result

This should not throw an error. It does not throw an error when it is used as an object passed to createMachine

![image](https://github.com/statelyai/xstate/assets/6943688/a2ea57b2-cd13-4964-b4c8-c63477fb2c9d)

### Actual result

![image](https://github.com/statelyai/xstate/assets/6943688/b045779f-39cd-4d10-8db0-1060121defd3)

### Reproduction

https://codesandbox.io/p/sandbox/silly-danny-vf63fc

### Additional context

Also just incase demo doesn't show it when you goto the repro it was being weird when I would refresh and it would show the right type but when I commented it out it would break again  🤔

https://github.com/statelyai/xstate/assets/6943688/aee6c437-21e0-4076-bf50-d6286c903d90

Andarist:
Unfortunately, we are hitting a limitation in TypeScript right now. I'd be happy to review any PR that fixes this but I already tried **extensively** to fix this issue with no luck. I will be working towards fixing this within the TS compiler but that will take some time to do, get reviewed, and released. 

For the time being, we need to classify this as a limitation that we are aware of and add a note about this in the docs or something. `enqueueActions` can't be used within `setup` - but it can be used within `createMachine` chained off `setup`

Jgordon-pencilwrench:
To expand on the (temporary?) workaround, use `enqueueActions` as an inline action within your event's action config, like:

```
 CANCEL_EDITING: {
            target: "editingComplete",
            actions: [
              enqueueActions(({ context, event, enqueue, check }) => {
                if (
                  check({
                    type: "formIsEmpty",
                    params: {
                      formData: context.formData,
                      additionalInformation: context.additionalInformation,
                    },
                  })
                ) {
                  enqueue({
                    type: "sendDeleteThisComplaintEventToParent",
                    params: { machineUuid: context.machineUuid },
                  });
                } else {
                  enqueue({
                    type: "editingCompleted",
                    params: {
                      machineUuid: context.machineUuid,
                      summary: event.payload.summary,
                      selectedChoiceUuid: context.selectedChoiceUuid ?? "",
                    },
                  });
                }
              }),
            ],
          }, ```

Andarist:
[...]

PR #4750, opened 2024-02-19
Forbid unknown implementations from being used

Andarist:
fixes https://github.com/statelyai/xstate/issues/4743
fixes https://github.com/statelyai/xstate/issues/4545

### DIFF
--- a/packages/core/src/actions/assign.ts
+++ b/packages/core/src/actions/assign.ts
@@ -13,7 +13,8 @@ import type {
   MachineContext,
   ParameterizedObject,
   PropertyAssigner,
-  ProvidedActor
+  ProvidedActor,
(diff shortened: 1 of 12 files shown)


packages/core/src/actions/assign.ts
```diff
@@ -146,7 +147,16 @@ export function assign<
         TEvent,
         TActor
       >
-): AssignAction<TContext, TExpressionEvent, TParams, TEvent, TActor> {
+): ActionFunction<
```
Andarist, line 150: We could revert this if we'd figure out how to include `| undefined` in our constraints for contextual type parameters (`_out_TActor` etc)

davidkpiano (approved):

PR #4752, opened 2024-02-20
XState Store

davidkpiano:
Initial release of `@xstate/store`

```ts

const store = createStore(
  // initial context
  { count: 0, greeting: 'hello' },
  // transitions
  {
    inc: {
      count: (context) => context.count + 1
    },
    updateBoth: {
      count: () => 42,
      greeting: 'hi'
    }
  }
);

store.send({
  type: 'inc'
});

console.log(store.getSnapshot());
// Logs:
// {
//   status: 'active',
//   context: {
//     count: 1,
//     greeting: 'hello'
//   }
// }
```


packages/xstate-store/src/react.ts
Andarist, line 1: u need to define an entrypoint for this one, like here:
https://github.com/statelyai/xstate/blob/f86a59ced85febed3ee67ea2bbe28fd34c465c6d/packages/core/package.json#L130-L138

packages/xstate-store/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+
+      return {
+        unsubscribe() {
+          return observers?.delete(observer);
+        }
+      };
+    },
+    [symbolObservable](): InteropSubscribable<StoreSnapshot<TContext>> {
+      return this;
+    }
```
Andarist, line 168: if you decouple this pkg from `xstate` then to make this work you should also include smth like this at the bottom of this file or somewhere:
https://github.com/statelyai/xstate/blob/f86a59ced85febed3ee67ea2bbe28fd34c465c6d/packages/core/src/index.ts#L42-L46

packages/xstate-store/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+        partialUpdate[key] =
+          typeof propAssignment === 'function'
+            ? (
+                propAssignment as StorePartialAssigner<
+                  TContext,
+                  StoreEvent,
+                  typeof key
+                >
+              )(currentContext, event)
+            : propAssignment;
```
Andarist, line 135: this branch doesn't use `updater` at all and it should

davidkpiano: Just like in `@xstate/immer`, it doesn't make sense for the `produce(…)` function (the main use-case for `updater`) to operate on individual properties.

Andarist: why? 🤔 it feels odd that I would be allowed to "mutate" things in the regular assigned but I wouldn't be able to do that in per-property assigner

Andarist: Ok, i guess i see how it would be quirky but the point above still stands - it's not intuitive that a "whole" assigner works differently from the per-property assigners when u configure the whole thing with `produce`.

davidkpiano: What do you recommend to do? I would rather have the constraint that per-property assigners are not allowed if a producer is provided.

davidkpiano: Also, for per-property, you can/should just use "idiomatic immer"; same # of lines:

```ts
// No producer
addTodo: {
  todos: (context, event) => context.todos.concat(event.todo)
}

// Producer
addTodos: (context, event) => {
  context.todos.push(event.todo);
}
```

packages/xstate-store/package.json
```diff
@@ -0,0 +1,66 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/statelyai/xstate.git"
+  },
+  "scripts": {},
+  "bugs": {
+    "url": "https://github.com/statelyai/xstate/issues"
+  },
+  "dependencies": {
+    "@testing-library/react": "^14.2.1",
```
Andarist, line 39: this one should be moved to devDeps

packages/xstate-store/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "use-sync-external-store": "^1.2.0"
+  },
+  "devDependencies": {
+    "immer": "^10.0.2",
+    "react": "^18.0.0",
+    "xstate": "^5.8.2"
+  },
+  "peerDependencies": {
+    "react": "^18.2.0",
+    "xstate": "^5.8.2"
```
Andarist, line 49: I guess i find it a little bit strange that this gateway lib *does* depend on `xstate`. I know that it's just an optional peer dependency but for many TS users this is really like a regular peer dependency here. Their projects won't compile without `xstate`.

packages/xstate-store/src/index.ts
Andarist: > I would rather have the constraint that per-property assigners are not allowed if a producer is provided.

this seems to be a reasonable way out of this for now

Andarist: Ah, it's already limited like this (sort of). Nice. I pushed out a small change to make it impossible to create producer-based stores with those property assigners using `createStore` API.

packages/xstate-store/src/index.ts
```diff
@@ -0,0 +1,226 @@
...
+  store.getSnapshot(); // { context: { count: 0 } }
+  store.send({ type: 'inc', by: 5 });
+  store.getSnapshot(); // { context: { count: 5 } }
+  ```
+ */
+export function createStoreWithProducer<
+  TProducer extends <TContext>(
+    context: TContext,
+    recipe: (context: TContext) => void
+  ) => TContext,
```
Andarist, line 204: does this have to be generic? 🤔 it feels like u could replace this with a simpler:
```ts
  producer: (context: TContex, recipe: (context: TContext) => void) => TContext,
```

packages/xstate-store/src/index.ts
```diff
@@ -0,0 +1,226 @@
...
+          { type: K } & TEventPayloadMap[K]
+        >;
+  },
+  updater?: (
+    context: NoInfer<TContext>,
+    recipe: (context: NoInfer<TContext>) => NoInfer<TContext>
+  ) => NoInfer<TContext>
+): Store<TContext, ExtractEventsFromPayloadMap<TEventPayloadMap>> {
+  type StoreEvent = ExtractEventsFromPayloadMap<TEventPayloadMap>;
+  const setter = updater ?? defaultSetter;
```
Andarist, line 61: nit: if im not mistaken u don't need this. Whenever you call `setter` you are actually calling `defaultSetter`

packages/xstate-store/src/types.ts
```diff
@@ -0,0 +1,51 @@
...
+  K extends keyof TContext
+> = (ctx: TContext, ev: TEvent) => Partial<TContext>[K];
+export type StorePropertyAssigner<TContext, TEvent extends EventObject> = {
+  [K in keyof TContext]?:
+    | TContext[K]
+    | StorePartialAssigner<TContext, TEvent, K>;
+};
+
+export interface StoreSnapshot<TContext> {
+  status: 'active'; // TODO: it can only be active
```
Andarist, line 32: if it can only be active what's left TODO here? :P

packages/xstate-store/src/index.ts
davidkpiano: Ah yes you're right

packages/xstate-store/package.json
Andarist: that's not what i've tried to convey 😅 i think it would make sense here to just copy-paste those types that you need here and remove a direct dependency (or peer dependency) on XSTate

packages/xstate-store/src/store.ts
```diff
@@ -0,0 +1,273 @@
...
+  TContext extends StoreContext,
+  TEventPayloadMap extends EventPayloadMap
+>(
+  producer: (
+    context: TContext,
+    recipe: (context: TContext) => void
+  ) => TContext,
+  initialContext: TContext,
+  transitions: {
+    [K in keyof TEventPayloadMap & string]: (
```
Andarist, line 189: I have a strong feeling that this wont work with TS versions earlier than 5.4. It's a new package though so I guess it's fine.

Andarist: And you are using `NoInfer` so it already requires 5.4 anyway

Andarist (approved):

davidkpiano:
Will add JSDocs in follow-up PR

PR #4759, opened 2024-02-23
[@xstate/vue] Fix `send(…)` type in `useActor(…)`

davidkpiano:


packages/xstate-vue/src/useActor.ts
```diff
@@ -15,7 +15,7 @@ export function useActor<TLogic extends AnyActorLogic>(
   options?: ActorOptions<TLogic>
 ): {
   snapshot: Ref<SnapshotFrom<TLogic>>;
-  send: (event: Actor<TLogic>['send']) => void;
```
Andarist, line 18: yikes, quite bizarre that this has not be caught by our tests anyhow

packages/xstate-vue/test/UseActorWithTransitionLogic.vue
Andarist, line 1: I'm worried that this file might be unused. Does any script error without ur fix based on this file alone?

davidkpiano: Added a test

Andarist (approved):

PR #4775, opened 2024-02-28
[@xstate/lit] Add Lit Controller

oscarmarina:
This pull request adds compatibility for linking XState with Lit.

## Motivation

XState's state machines offer a structured and predictable approach to handle complex logic, while Lit facilitates reactive UI updates in response to state changes.

## Changes

It follows the established structure of the xstate repository, including:

### 📂 packages/xstate-lit/src/

Adds xstate-lit to packages with code, tests, and documentation.

▨  **UseMachine.ts**

Implements the @xstate/lit controller, referencing other packages for guidance as much as possible.
@xstate/svelte, @xstate/vue and [Lifecycle: reactive controller adapters for other frameworks](https://github.com/lit/lit/issues/1682#issue-836224813)

- Provides get `actor`,  get `snapshot`, and `send(ev: EventFrom<TMachine>)` method for interacting with the XState actor.
- Exposes an `unsubscribe` method 
- Option to pass a reactive property for use in the actor's subscribe method.
- Documentation of the API in the README file.

```js
this.toggleController = new UseMachine(this, {
  machine: toggleMachine,
  options: { inspect },
  subscriptionProperty: '_xstate',
});

// ...

private get _turn() {
    return this.toggleController.snapshot.matches('inactive');
  }

render() {
  return html`
    <button @click=${() => this.toggleController.send({ type: 'TOGGLE' })}>
      ${this._turn ? 'Turn on' : 'Turn off'}
    </button>
  `;
}

```

▨  **useActorRef.ts**

Creates and returns the XState actor without Lit-specific dependencies (handled in UseMachine.js).

▨  **index.ts:**

Exports only UseMachine.ts.

> It is a structure more in line with the working approach used by Lit's controllers.

### 📂 packages/xstate-lit/test
Leverages @open-wc/testing-helpers for unit testing components, drawing inspiration from existing tests in Svelte and Vue integrations.

▨  **useActor.test.ts**

```js
it('should be able to spawn an actor from actor logic', async () => {
    const el: UseActorWithTransitionLogic = await fixture(
      html`<use-actor-with-transition-logic></use-actor-with-transition-logic>`
    );
    const buttonEl = getByTestId(el, 'count');
    await waitFor(() => expect(buttonEl.textContent?.trim()).toEqual('0'));
    await fireEvent.click(buttonEl);
    await el.updateComplete;
    await waitFor(() => expect(buttonEl.textContent?.trim()).toEqual('1'));
  });
```

▨  **useActorRef.test.ts**
```js
it('observer should be called with next state', async () => {
    const el: UseActorRef = await fixture(
      html`<use-actor-ref></use-actor-ref>`
    );
    const buttonEl = getByTestId(el, 'button');
    await waitFor(() => expect(buttonEl.textContent?.trim()).toBe('Turn on'));
    await fireEvent.click(buttonEl);
    await el.updateComplete;
    await waitFor(() => expect(buttonEl.textContent?.trim()).toBe('Turn off'));
  });
```

### 📂 templates/lit-ts/
Adds examples and documentation.

#### Usage:
```bash
npm i && npm start
```

**Provides two demos:**
- `<lit-ts>` & feedbackMachine: Equal to existing templates in other packages.
- `<lit-ts-counter>` & counterMachine: Illustrates using a reactive property and the [inspect API](https://stately.ai/docs/inspection#snapshot-inspection-events) to listen for events that caused transitions and reset reactive property.

| <lit-ts>                           | <lit-ts-counter>                           |
|---------------------------------|---------------------------------|
|  <lit-ts>    | <lit-ts-counter> |

<hr>

>  ⚠ **Update before "publish"**

```js
import { UseMachine } from '../../../packages/xstate-lit/src/index.js';
// import { UseMachine } from '@xstate/lit';
```

- templates/lit-ts/src/LitTs.ts (line 4 and 5)
- templates/lit-ts/src/LitTsCounter.ts (line 5 and 6)

<hr>

## Other Modified Files:

**▨  jest.config.js**

Add transformIgnorePatterns to accommodate Lit and Open-WC.

```js
transformIgnorePatterns: [
  'node_modules/(?!(@open-wc|lit-html|lit-element|lit|@lit)/)'
],
```

### 📂 scripts/jest-utils/

**▨  setup.js**

Filters out Lit's "Lit is in dev mode..." console logs during tests.

<hr>

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,6 +5,9 @@ const { constants } = require('jest-config');
  */
 module.exports = {
   setupFilesAfterEnv: ['@xstate-repo/jest-utils/setup'],
+  transformIgnorePatterns: [
+    'node_modules/(?!(@open-wc|lit-html|lit-element|lit|@lit)/)'
+  ],
   transform: {
     [constants.DEFAULT_JS_PATTERN]: 'babel-jest',
     '^.+\\.vue$': '@vue/vue3-jest',
--- a/packages/xstate-lit/README.md
+++ b/packages/xstate-lit/README.md
@@ -1,0 +1,147 @@
+# @xstate/lit
+
+The [@xstate/lit](https://github.com/lit/lit) package contains a [Reactive Controller](https://lit.dev/docs/composition/controllers/) for using XState with Lit.
+
+- [Read the full documentation in the XState docs](https://stately.ai/docs/xstate-lit/).
+- [Read our contribution guidelines](https://github.com/statelyai/xstate/blob/main/CONTRIBUTING.md).
+
+## Quick Start
+
+1. Install `xstate` and `@xstate/lit`:
+
+```bash
+npm i xstate @xstate/lit
+```
+
+**Via CDN**
+
+```html
+<script src="https://unpkg.com/@xstate/lit/dist/xstate-lit.esm.js"></script>
+```
+
+2. Import the `UseMachine` Lit controller:
+
+**`new UseMachine(this, {machine, options?, subscriptionProperty?)`**
+
+```js
+import { html, LitElement } from 'lit';
+import { customElement, state } from 'lit/decorators.js';
+import { createBrowserInspector } from '@statelyai/inspect';
+import { createMachine } from 'xstate';
+import { UseMachine } from '@xstate/lit';
+
+const { inspect } = createBrowserInspector({
+  // Comment out the line below to start the inspector
+  autoStart: false,
+});
+
+const toggleMachine = createMachine({
+  id: 'toggle',
+  initial: 'inactive',
+  states: {
+    inactive: {
+      on: { TOGGLE: 'active' },
+    },
+    active: {
+      on: { TOGGLE: 'inactive' },
+    },
+  },
+});
+
+@customElement('toggle-component')
+export class ToggleComponent extends LitElement {
+  toggleController: UseMachine<typeof toggleMachine>;
+
+  @state()
+  protected _xstate: { [k: string]: unknown } = {};
+
+  constructor() {
+    super();
+    this.toggleController = new UseMachine(this, {
+      machine: toggleMachine,
+      options: { inspect },
+      subscriptionProperty: '_xstate',
+    });
+  }
+
+  override updated(props: Map<string, unknown>) {
+    super.updated && super.updated(props);
+    if (props.has('_xstate')) {
+      const { value } = this._xstate;
+      const toggleEvent = new CustomEvent('togglechange', {
+        detail: value,
+      });
+      this.dispatchEvent(toggleEvent);
+    }
+  }
+
+  private get _turn() {
+    return this.toggleController.snapshot.matches('inactive');
+  }
+
+  render() {
+    return html`
+      <button @click=${() => this.toggleController.send({ type: 'TOGGLE' })}>
+        ${this._turn ? 'Turn on' : 'Turn off'}
+      </button>
+    `;
+  }
+}
+```
+
+## API
+
+### `new UseMachine(host, {machine, options?, subscriptionProperty?)`
+
+A class that creates an actor from the given machine and starts a service that runs for the lifetime of the component.
+
+#### Constructor Options:
+
+`host`: ReactiveControllerHost: The Lit component host.
+`machine`: AnyStateMachine: The XState machine to manage.
+`options?`: ActorOptions<TMachine>: Optional options for the actor.
+`subscriptionProperty?`: string: Optional property on the host to update with the state snapshot.
+
+#### Return Methods:
+
+`actor` - Returns the actor (state machine) instance.
+`snapshot` - Returns the current state snapshot.
+`send` - Sends an event to the state machine.
+`unsubscribe` - Unsubscribes from state updates.
+
+## Matching States
+
+When using [hierarchical](https://xstate.js.org/docs/guides/hierarchical.html) and [parallel](https://xstate.js.org/docs/guides/parallel.html) machines, the state values will be objects, not strings. In this case, it is best to use [`state.matches(...)`](https://xstate.js.org/docs/guides/states.html#state-methods-and-properties).
+
+```js
+${this.myXsateController.snapshot.matches('idle')}
+//
+${this.myXsateController.snapshot.matches({ loading: 'user' })}
+//
+${this.myXsateController.snapshot.matches({ loading: 'friends' })}
+```
+
+## Persisted and Rehydrated State
+
+You can persist and rehydrate state with `useMachine(...)` via `options.snapshot`:
+
+```js
+// Get the persisted state config object from somewhere, e.g. localStorage
+// highlight-start
+const persistedState = JSON.parse(
+  localStorage.getItem('some-persisted-state-key'),
+);
+// highlight-end
+
+connectedCallback() {
+  super.connectedCallback && super.connectedCallback();
+  this.fetchController = this.fetchController ?? new UseMachine(this, {
+    machine: someMachine,
+    options: {
+      snapshot: this.persistedState
+    }
+  });
+}
+
+// state will initially be that persisted state, not the machine’s initialState
+```
--- a/packages/xstate-lit/package.json
+++ b/packages/xstate-lit/package.json
@@ -1,0 +1,61 @@
+{
+  "name": "@xstate/lit",
+  "version": "1.0.0",
+  "description": "XState tools for Lit",
+  "keywords": [
+    "state",
+    "machine",
+    "statechart",
+    "scxml",
+    "state",
+    "graph",
+    "store",
+    "lit",
+    "reactive controller",
+    "web components"
+  ],
+  "author": "David Khourshid <davidkpiano@gmail.com>",
+  "homepage": "https://github.com/statelyai/xstate/tree/main/packages/xstate-lit#readme",
+  "license": "MIT",
+  "main": "dist/xstate-lit.cjs.js",
+  "module": "dist/xstate-lit.esm.js",
+  "exports": {
+    ".": {
+      "types": {
+        "import": "./dist/xstate-lit.cjs.mjs",
+        "default": "./dist/xstate-lit.cjs.js"
+      },
+      "module": "./dist/xstate-lit.esm.js",
+      "import": "./dist/xstate-lit.cjs.mjs",
+      "default": "./dist/xstate-lit.cjs.js"
+    },
+    "./package.json": "./package.json"
+  },
+  "sideEffects": false,
+  "files": [
+    "dist"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/statelyai/xstate.git"
+  },
+  "bugs": {
+    "url": "https://github.com/statelyai/xstate/issues"
+  },
+  "peerDependencies": {
+    "lit": "^3.1.2",
+    "xstate": "^5.8.2"
+  },
+  "peerDependenciesMeta": {
+    "xstate": {
+      "optional": true
+    }
+  },
+  "devDependencies": {
+    "@open-wc/testing-helpers": "^3.0.0",
+    "@testing-library/dom": "^9.3.4",
+    "@types/jest": "^29.5.10",
+    "lit": "^3.1.2",
+    "xstate": "5.8.2"
+  }
+}
--- a/packages/xstate-lit/src/UseMachine.ts
+++ b/packages/xstate-lit/src/UseMachine.ts
@@ -1,0 +1,95 @@
+import { ReactiveController, ReactiveControllerHost } from 'lit';
+import {
+  Actor,
+  AnyStateMachine,
+  ActorOptions,
+  Subscription,
+  EventFrom,
+  SnapshotFrom
+} from 'xstate';
+import { useActorRef } from './useActorRef.ts';
+
+export class UseMachine<TMachine extends AnyStateMachine>
+  implements ReactiveController
+{
+  private host: ReactiveControllerHost;
+  private machine: TMachine;
+  private options?: ActorOptions<TMachine>;
+  private subscriptionProperty: string;
+  private shouldUpdateSubscriptionProperty: boolean;
+  private actorRef = {} as Actor<TMachine>;
+  private subs: Subscription = { unsubscribe: () => {} };
+  private currentSnapshot: SnapshotFrom<TMachine>;
+
+  constructor(
+    host: ReactiveControllerHost,
+    {
+      machine,
+      options,
+      subscriptionProperty = ''
+    }: {
+      machine: TMachine;
+      options?: ActorOptions<TMachine>;
+      subscriptionProperty?: string;
+    }
+  ) {
+    this.machine = machine;
+    this.options = options;
+    this.subscriptionProperty = subscriptionProperty;
+    this.shouldUpdateSubscriptionProperty = subscriptionProperty in host;
+    this.currentSnapshot = this.snapshot;
+    this.onNext = this.onNext.bind(this);
+
+    (this.host = host).addController(this);
+  }
+
+  get actor() {
+    return this.actorRef;
+  }
+
+  get snapshot() {
+    return this.actorRef?.getSnapshot?.();
+  }
+
+  send(ev: EventFrom<TMachine>) {
+    this.actorRef?.send(ev);
+  }
+
+  unsubscribe() {
+    this.subs.unsubscribe();
+  }
+
+  protected updateSubscriptionProperty(snapshot: SnapshotFrom<TMachine>) {
+    if (this.shouldUpdateSubscriptionProperty) {
+      (this.host as unknown as Record<string, unknown>)[
+        this.subscriptionProperty
+      ] = snapshot;
+    }
+  }
+
+  protected onNext(snapshot: SnapshotFrom<TMachine>) {
+    if (this.currentSnapshot !== snapshot) {
+      this.currentSnapshot = snapshot;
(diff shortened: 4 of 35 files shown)


packages/xstate-lit/README.md
```diff
@@ -0,0 +1,147 @@
...
+        ${this._turn ? 'Turn on' : 'Turn off'}
+      </button>
+    `;
+  }
+}
+```
+
+## API
+
+### `new UseMachine(host, {machine, options?, subscriptionProperty?)`
```
davidkpiano, line 94: Typo? `{machine`

davidkpiano:
```suggestion
### `new UseMachine(host, machine, options?, subscriptionProperty?)`
```

packages/xstate-lit/test/tsconfig.json
Andarist, line 1: do we need this file? at the moment we have a single top-level tsconfig that handles everything - it's perhaps slightly less accurate but way more convenient

jest.config.js
```diff
@@ -5,6 +5,9 @@ const { constants } = require('jest-config');
  */
 module.exports = {
   setupFilesAfterEnv: ['@xstate-repo/jest-utils/setup'],
+  transformIgnorePatterns: [
+    'node_modules/(?!(@open-wc|lit-html|lit-element|lit|@lit)/)'
+  ],
```
Andarist, line 10: why do we need this? does it turn off ESM->CJS transform?

packages/xstate-lit/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "url": "https://github.com/statelyai/xstate/issues"
+  },
+  "peerDependencies": {
+    "lit": "^3.1.2",
+    "xstate": "^5.8.2"
+  },
+  "peerDependenciesMeta": {
+    "xstate": {
+      "optional": true
+    }
```
Andarist, line 52: Having this optional doesn't make a lot of sense to me right now. If we do this in other packages that is probably just a leftover from v4 when we were integrating with both `xstate` and `@xstate/fsm`

packages/xstate-lit/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+  @state()
+  protected _xstate: { [k: string]: unknown } = {};
+
+  constructor() {
+    super();
+    this.toggleController = new UseMachine(this, {
+      machine: toggleMachine,
+      options: { inspect },
+      subscriptionProperty: '_xstate',
```
Andarist, line 63: couldnt we just have a subscription callback for people to hook into and do their thing? I'm not used to Lit so maybe that's how things are done in that ecosystem but this option feels clumsy to me and makes types very loosely coupled (to the point that this doesn't really deliver good type-safety)

packages/xstate-lit/README.md
oscarmarina: Hi, my mistake I forgot to close the curly braces in the definition of UseMachine and I misspelled myXStateController.

packages/xstate-lit/README.md
```diff
@@ -0,0 +1,147 @@
...
+## Matching States
+
+When using [hierarchical](https://xstate.js.org/docs/guides/hierarchical.html) and [parallel](https://xstate.js.org/docs/guides/parallel.html) machines, the state values will be objects, not strings. In this case, it is best to use [`state.matches(...)`](https://xstate.js.org/docs/guides/states.html#state-methods-and-properties).
+
+```js
+${this.myXsateController.snapshot.matches('idle')}
+//
+${this.myXsateController.snapshot.matches({ loading: 'user' })}
+//
+${this.myXsateController.snapshot.matches({ loading: 'friends' })}
```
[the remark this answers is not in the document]
oscarmarina: Yes, wrong text

packages/xstate-lit/test/tsconfig.json
oscarmarina: Hi, fixed, I copied it from the 'xstate-svelte' package

packages/xstate-lit/package.json
oscarmarina: Hi, fixed, I copied it from the 'xstate-vue' package

jest.config.js
oscarmarina: It is necessary to use Jest with Lit.
You can see more information here.

https://github.com/jestjs/jest/issues/11783#issuecomment-1107861603

===
jest-and-lit

packages/xstate-lit/package.json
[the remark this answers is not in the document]
oscarmarina: Hi, I  added 'type:module' and that fixed the error in "yarn typecheck".

https://github.com/preconstruct/preconstruct/pull/586/checks#discussion_r1452789690

oscarmarina:
Hi, @davidkpiano, @Andarist
I have uploaded to stackblitz "a version of the demo located in the template folder" if you want to see how it works.

- https://stackblitz.com/github/oscarmarina/XstateController?file=src%2FLitTs.ts

The folder xstat-lit corresponds to what is in the folder packages/xstate-lit

packages/xstate-lit/README.md
oscarmarina: Hi @Andarist,
I have replaced 'subscriptionProperty' as you suggest with a 'callback', this makes it much more aligned and also with [lit --> observers](https://github.com/lit/lit/tree/main/packages/labs/observers)

packages/xstate-lit/package.json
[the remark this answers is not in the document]
Andarist: it still should use that experimental option - otherwise, I'd expect us to run into problems with `preconstruct dev/validate/build`

packages/xstate-lit/src/useActorRef.ts
```diff
@@ -0,0 +1,9 @@
+import { Actor, ActorOptions, AnyActorLogic, createActor } from 'xstate';
+
+export const useActorRef = <TMachine extends AnyActorLogic>(
```
Andarist, line 3: this one should be exported from the index and it should also accept `TLogic` (not just `TMachine`)

packages/xstate-lit/src/useActorRef.ts
```diff
@@ -0,0 +1,9 @@
...
+  logic: TMachine,
+  options?: ActorOptions<TMachine>
+): Actor<TMachine> => {
+  const actorRef = createActor(logic, options);
+  return actorRef;
+};
```
Andarist, line 9: it's just a wrapper around `createActor` - do you even need it here? couldn't u just call `createActor` directly in ur `UseMachine`?

packages/xstate-lit/src/UseMachine.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.actorRef?.start();
+  }
+
+  private stopService() {
+    this.actorRef?.stop();
+  }
+
+  hostConnected() {
+    this.startService();
+  }
```
Andarist, line 80: interesting, out of curiosity - does Lit has to iterate over all properties, recognize what might be a reactive controller and what is not and then handle all of the recognized controllers in a special way (to call their lifecycle-related methods etc)?

packages/xstate-lit/test/UseActor.ts
```diff
@@ -0,0 +1,80 @@
...
+const fMachine = fetchMachine.provide({
+  actors: {
+    fetchData: fromPromise(onFetch)
+  }
+});
+
+export class UseActor extends LitElement {
+  fetchController: UseMachine<typeof fetchMachine> = {} as UseMachine<
+    typeof fetchMachine
+  >;
```
Andarist, line 21: wouldn't it be possible to call `new UseMachine(...)` here? declaring a property like this would be much easier for consumers

packages/xstate-lit/src/UseMachine.ts
Andarist: ahh, scratch that - I've missed `addController` call

packages/xstate-lit/test/UseActor.ts
oscarmarina: Yes, updated.

packages/xstate-lit/package.json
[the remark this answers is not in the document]
oscarmarina: Done!

oscarmarina:
Hi @Andarist 
I'm thinking that if the "PR" is correct, I would need to decline this PR and create two new ones, one for `@xstate/lit` and another for the `templates/lit-ts`.

**This way, I can add the dependency to the `templates/lit-ts` package file and also update the "import(s)".**

Here are the steps I would take:

- Decline the current PR.
- Create a new PR for "@xstate/lit".
- Create a new PR for the "templates/lit-ts".
  - In the "templates/lit-ts" package file, add the dependency on "@xstate/lit".
  - In the "templates/lit-ts" update the "import(s)" to use "@xstate/lit".
- Submit the two new PRs for review.

Does this approach sound reasonable?

Andarist:
I'd keep this PR open but remove templates from it. Then you can have a branch with templates on it, and point us to it so we can see how it's used in practice (although tests in this PR here might/should be enough too). Once we land this PR that introduces `@xstate/lit` then you'll be able to open a new one with the templates.

Also, please note that I'm aware of this PR and I plan to review it thoroughly (so far I didn't really do a proper review - just a driveby one). It might take some time before I properly get to it because at the moment I'm focusing on something else. We really appreciate your contribution!

oscarmarina:
Perfect, I'll leave it as it is for now to not drive you crazy, and if it eventually makes sense and gets approved, I'll make the necessary changes.

Thank you for taking the time to review the PR.

christophe-g:
For those interested, this is an alternate lit-controller: https://github.com/lit-apps/lit-app/tree/dev/packages/actor

It leverages https://github.com/lit-apps/lit-app/tree/dev/packages/state and wraps xstate actors so that lit element re-renders when the snapshot changes. 

**example usage:**
```js
const actor = new Actor(workflow)
 
export default class fsmTest extends LitElement {
  // bind actor state to fsmTest element, so it will re-render when actor snapshot changes
  bindActor = new StateController(this, actor)
  
  override render() {

   const send = () => actor.send({ 
    type: 'NewPatientEvent', name: 'John', condition: 'healthy' 
   })
   
   return html`
    <div>
     <div>status: ${actor.status}</div>
     <div>any context value: ${actor.context.anyValue}</div>
     <div>value: ${JSON.stringify(actor.value)}</div>
    </div>
    <md-filled-button @click=${send}>NewPatientEvent</md-filled-button>
   `;
  }
}
```

jest.config.js
justinfagnani: Jest still doesn't support modules??

packages/xstate-lit/src/UseMachine.ts
```diff
@@ -0,0 +1,85 @@
...
+  Actor,
+  AnyStateMachine,
+  ActorOptions,
+  createActor,
+  EventFrom,
+  SnapshotFrom,
+  Subscription
+} from 'xstate';
+
+export class UseMachine<TMachine extends AnyStateMachine>
```
justinfagnani, line 12: Since this is a controller and not a React hook, I'd recommend against the "Use" name, and call this something like `MachineController`

packages/xstate-lit/src/UseMachine.ts
```diff
@@ -0,0 +1,85 @@
...
+  implements ReactiveController
+{
+  private host: ReactiveControllerHost;
```
justinfagnani, line 15: I'd recommend prefixing private fields with `_` so that they're more obviously private even to plain JavaScript users.

Even better if you can, I'd use standard private fields. They have very good [browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties#javascript.classes.private_class_fields) these days, and even better tool support.

packages/xstate-lit/src/UseMachine.ts
```diff
@@ -0,0 +1,85 @@
...
+      machine: TMachine;
+      options?: ActorOptions<TMachine>;
+      callback?: (snapshot: SnapshotFrom<TMachine>) => void;
+    }
+  ) {
+    this.machine = machine;
+    this.options = options;
+    this.callback = callback;
+    this.currentSnapshot = this.snapshot;
+    this.onNext = this.onNext.bind(this);
```
justinfagnani, line 39: Defining `onNext` as an instance field with an arrow function value will prevent this reassignment, which changes the class shape.

packages/xstate-lit/test/UseActor.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  override createRenderRoot() {
+    return this;
+  }
+
+  override render() {
+    return html`
+      <slot></slot>
+      <div>
+        ${this.fetchController.snapshot.matches('idle')
```
justinfagnani, line 38: I don't know the XState API, but could the control offer a `@lit/task`-like render helper that's a bit more declarative?

Something like:
```ts
this.fetchController.render({
  idle: () => html`...`,
  loading: () => html`...`,
  success: () => html`...`,
});
```

Could such an API be type safe wrt the the state names?

packages/xstate-lit/test/UseActor.ts
```diff
@@ -0,0 +1,71 @@
...
+          ? html`
+              <button
+                @click=${() => this.fetchController.send({ type: 'FETCH' })}
```
justinfagnani, line 41: This may be a super small DX improvement, but the controller could offer a helper that created an event handler for the user, instead of requiring an arrow function. It would also eliminate the closure creation on every render for a small perf gain.

Something like
```ts
html`
    <button
      @click=${this.fetchController.sendHandler({ type: 'FETCH' })}
```

`sendHandler()` is probably a terrible name, but I couldn't think of something better right now.

justinfagnani (commented):
This is great! Hope a drive-by review is welcome!

packages/xstate-lit/src/UseMachine.ts
oscarmarina: As with my previous comment, I followed the style used in xstate here. I reviewed some controllers in Lit, such as:

- [context-consumer](https://github.com/lit/lit/blob/main/packages/context/src/lib/controllers/context-consumer.ts#L40)
- [mutation-controller](https://github.com/lit/lit/blob/main/packages/labs/observers/src/mutation-controller.ts#L63)

And decided to maintain that style.

packages/xstate-lit/src/UseMachine.ts
oscarmarina: Changed 👌

packages/xstate-lit/test/UseActor.ts
oscarmarina: I’m not quite sure what you're suggesting here. Could you please clarify?

Here is a repository where I created some demos before making the PR. If possible, you can add your suggestions there:

https://github.com/oscarmarina/XstateController/blob/main/xstate-lit/src/UseMachine.ts

packages/xstate-lit/test/UseActor.ts
oscarmarina: Hmm, I'm not quite sure if those two concepts can be combined. We would need @davidkpiano  and  @Andarist 's input on this.

packages/xstate-lit/src/UseMachine.ts
justinfagnani: Oops... those don't actually follow our code style, so I'll fix them. Thanks for pointing them out!

One reason why we prefix all private fields is so that they can be renamed in a minifier like Terser with a simple rule.

packages/xstate-lit/test/UseActor.ts
justinfagnani: Which two concepts?

packages/xstate-lit/test/UseActor.ts
justinfagnani: So right now you need to use an arrow function in the click handler:

```ts
html`
  <button
    @click=${() => this.fetchController.send({ type: 'FETCH' })}
  >
    Fetch
  </button>
`
```

This is fine, but it does allocate a new function object for this binding every render, and it's a slight bit of boilerplate.

Instead you can have the controller allocate a function object once and reuse it every render.

Using it would look like:

```ts
html`
  <button
    @click=${this.fetchController.sendHandler({ type: 'FETCH' })}
  >
    Fetch
  </button>
`
```

and implementing it would look like:

```ts
class MachineController {
  #sendHandler;
  
  /** Returns a callback that calls this.send() with the given arguments **/
  sendHandler(...args) {
    return this.#sendHandler ??= () => this.send(...args);
  }
}
```

packages/xstate-lit/src/UseMachine.ts
justinfagnani: As someone more planted in the Lit ecosystem, I think it'd be nice to match Lit naming too - which just follows the somewhat cross-language standard OO noun-phrase vs verb-phrase naming. The class should be the noun-phrase name of the category of thing that the instances are, while functions should be verb-phrases.

`UseMachine` is a verb-phrase, and I think it reads a bit weird to have statements like:

```ts
this.fetchController = new UseMachine(...);
```

Because this is saying that `fetchController` _is_ a "UseMachine". But what is a "UseMachine"? It's a controller - an object with a verb-phrased name. But it sounds like a function.

This is why most reactive controllers are named {X}Controller.

It's possible to vend a function that makes the class for you, so maybe that's better:

```ts
this.fetchController = useMachine(...);
```

But even then, in the Lit ecosystem I'd still want to shy away from the "use" name because people tend to think that it's a hook and has React-like "rule of hooks" semantics, when it doesn't.

oscarmarina:
I’m closing this PR as it has become outdated. I have updated everything and split the work into two new pull requests for easier review and integration. Thank you!

PR #4782, opened 2024-03-04
[core] Mark properties as @hidden instead of @internal

davidkpiano:


Andarist:
Please include some reasoning behind the change. At least some of those changed locations were marked as internal very intentionally (to remove them from emitted types)

packages/core/src/StateMachine.ts
```diff
@@ -94,10 +94,7 @@ export class StateMachine<
 
   public implementations: MachineImplementationsSimplified<TContext, TEvent>;
 
-  /** @internal */
-  public __xstatenode: true = true;
```
Andarist, line 98: Technically it is a breaking change - one that we don't have to make. We shouldn't freely remove things unless there is a very strong reason to do that.

packages/core/src/types.ts
```diff
@@ -1777,7 +1777,7 @@ export interface ActorOptions<TLogic extends AnyActorLogic> {
   logger?: (...args: any[]) => void;
   parent?: AnyActorRef;
   /**
-   * @internal
+   * @hidden
```
Andarist, line 1780: this one is one of the things that were hidden on purpose because this isn't supposed to be a valid call outside of the codebase:
```ts
createActor(m, { syncSnapshot: true })
```

It doesn't make sense on its own, we have privileged access to stuff like this because only we are responsible for actually creating child actors etc.

All of those "hidden" things now will pollute~ autocompletes and stuff so overall I'd call that a pretty big downside of the proposed change. 

The whole change seems to be kinda motivated by the discussion on Discord in which:
- Tim couldn't find `AnyActorSystem`
- Tim tried to call an internal `._relay`

This PR doesn't do anything about `AnyActorSystem` - it stays unexported. Tim also has stopped using `._relay` (which likely he shouldn't be using in the first place since it's meant to be a private method here).

So overall this change feels a little bit misfired, so to speak.

PR #4789, opened 2024-03-07
TypeScript 5.4

davidkpiano:
This PR updates the TypeScript version in the XState monorepo to 5.4, which includes replacing `NoInfer` with the native `NoInfer` (doesn't work yet)

### DIFF
--- a/.changeset/tall-actors-call.md
+++ b/.changeset/tall-actors-call.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Upgrade to TypeScript 5.4
--- a/.changeset/tall-actors-call.md
+++ b/.changeset/tall-actors-call.md
@@ -1,5 +1,0 @@
----
-'xstate': patch
----
-
-Upgrade to TypeScript 5.4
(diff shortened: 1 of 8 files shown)


Andarist:
We cant easily replace our `NoInfer` with the native one because we’d break consumers using other TS versions.

.changeset/tall-actors-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Upgrade to TypeScript 5.4
```
Andarist, line 5: This changeset should be removed - our consumers are not that concerned with what TS version our monorepo is using.

yarn.lock
```diff
@@ -9198,11 +9198,16 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=
 
-typescript@^5.0.3, typescript@^5.2.2:
+typescript@^5.0.3:
   version "5.2.2"
   resolved "https://registry.yarnpkg.com/typescript/-/typescript-5.2.2.tgz#5ebb5e5a5b75f085f22bc3f8460fba308310fa78"
   integrity sha512-mI4WrpHsbCIcwT9cF4FZvr80QUeKvsUsUvKDoR+X/7XHQH98xYD8YHZg7ANtz2GtZt/CBq2QJ0thkGJMHfqc1w==
 
+typescript@^5.4.2:
```
Andarist, line 9206: for some reason we now have two TS versions installed - I hope it wont create problems, it would be great to deduplicate this

PR #4792, opened 2024-03-10
[core] Add `children` to `Snapshot` interface for all logic types

davidkpiano:
This will be required for adding the ability to `spawn(…)` in any logic


boneskull:
As written, this should not conflict with #4929

davidkpiano:
Will rethink this for v6

PR #4794, opened 2024-03-12
[core] Add dev-time warning for events sent before actor is started

davidkpiano:


packages/core/src/createActor.ts
```diff
@@ -678,10 +678,17 @@ export class Actor<TLogic extends AnyActorLogic>
...
+      if (typeof event === 'string') {
+        throw new Error(
+          `Only event objects may be sent to actors; use .send({ type: "${event}" }) instead`
+        );
+      }
+      if (this._processingStatus === ProcessingStatus.NotStarted) {
+        console.warn(
+          `Event "${event.type}" sent to actor "${this.id}" before it started. Remember to call \`actor.start()\` before sending events.`
+        );
+      }
```
Andarist, line 691: It's not a good warning when it's applied like this. Sending events like this is totally legal - I'm pretty sure that we do this in our tests and that the CI might quickly fail because of this. 

But it's not only about our test suite. Sending from a child effect to the parent would start warning (it's not desired) since it happens before the parent's effect runs (so before it has its chance to start the target actor).

PR #4800, opened 2024-03-15
Types Improvement: prevent deep _parent access (a TODO fix)

korniychuk:
```ts
 self._parent._parent
```

Screenshot 2024-03-15 at 14 52 00@2x

This is a suggestion, I didn't check all the side effects.


Andarist:
Even shallow access is discouraged. This is a private property as denoted by the leading underscore. I prefer to avoid the Omit type when possible since it’s not the most performant type util.

I think the added benefit here doesn't justify this code change.

korniychuk:
Thanks for the explanation!

PR #4806, opened 2024-03-18
Allow inline actor logic

davidkpiano:


Andarist:
The problem is that by allowing inline actor logics you allow them to override the ones that are configured and that leads to problems:
```ts
const m = setup({
  types: { children: { foo: 'myChild' } },
  actors: {
    myChild: fromPromise(async () => "")
  }
}).createMachine({
  entry: spawnChild(fromPromise(async () => 100), { id: 'foo' })
})

const actorRef = createActor(m)
const val = actorRef.getSnapshot().children.foo.getSnapshot().output // actual: number | undefined, expected: string | undefined
```

davidkpiano:
> The problem is that by allowing inline actor logics you allow them to override the ones that are configured and that leads to problems:
> 
> ```ts
> const m = setup({
>   types: { children: { foo: 'myChild' } },
>   actors: {
>     myChild: fromPromise(async () => "")
>   }
> }).createMachine({
>   entry: spawnChild(fromPromise(async () => 100), { id: 'foo' })
> })
> 
> const actorRef = createActor(m)
> const val = actorRef.getSnapshot().children.foo.getSnapshot().output // actual: number | undefined, expected: string | undefined
> ```

This really makes me think that the way that child actors are defined is a bit awkward currently. Would love to revisit this for v6

Andarist (approved):

PR #4809, opened 2024-03-19
[core] Ensure "stop" transitions are called in machine actors

davidkpiano:
This PR ensures that "stop" transitions are called in machines:

```ts
  on: {
    'xstate.stop': {
      actions: () => {
        // do cleanup, etc...
        console.log('Machine actor was just stopped');
      }
    }
  }
});

actor.stop();
// Logs "Machine actor was just stopped"
```

### DIFF
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -1647,22 +1647,6 @@ export function macrostep(
     microstates.push(microstate);
   }
 
-  // Handle stop event
-  if (event.type === XSTATE_STOP) {
-    nextSnapshot = cloneMachineSnapshot(
-      stopChildren(nextSnapshot, event, actorScope),
-      {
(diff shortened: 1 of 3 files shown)


packages/core/src/stateUtils.ts
```diff
@@ -1647,22 +1647,6 @@ export function macrostep(
...
-  if (event.type === XSTATE_STOP) {
-    nextSnapshot = cloneMachineSnapshot(
-      stopChildren(nextSnapshot, event, actorScope),
-      {
-        status: 'stopped'
-      }
-    );
-    addMicrostate(nextSnapshot, event, []);
-
-    return {
```
davidkpiano, line 1660: The main problem was that this was returning too early, without giving the machine a chance to process the `"xstate.stop"` event

PR #4816, opened 2024-03-25
First Draft for @xstate/angular

niklas-wortmann:
This is a first draft for an angular implementation for XState.

As you will notice, it does need some tests, and also for convenience, we could add an angular add schematic, but other than that it is working in the tests I ran. 

It did take heavy inspiration from the xstate/vue implementation but also ngrx signalstore. 

Let me know what you think!


niklas-wortmann:
Just saw that the pipeline is failing, and I also did some copy paste mistakes in the package Json, I will fix those tomorrow morning

packages/xstate-angular/package.json
```diff
@@ -0,0 +1,70 @@
...
+  "peerDependenciesMeta": {
+    "xstate": {
+      "optional": true
+    }
+  },
+  "devDependencies": {
+    "@angular/core": "^17.0.0",
+    "@testing-library/angular": "15.2.0",
+    "@testing-library/jest-dom": "^6.1.4",
+    "ng-packagr": "^17.0.0",
```
Andarist, line 67: How strongly required this is? it might be slightly challenging to incorporate this into our current setup. It's certainly doable but if we wouldn't have to use this... that would be kinda simpler for us

Andarist:
I wonder if you had a chance to take a look at https://github.com/statelyai/xstate/pull/4638 . How would you compare both approaches?

packages/xstate-angular/package.json
niklas-wortmann: I think it is required but I will check in the morning

niklas-wortmann:
@Andarist overall the implementation is pretty similar. I am slightly biased here, but I think my implementation is a little more ergonomic and flexible, also I prefer not to expose the user to deal with the injection context. My implementation abstracts this better. Also, this approach provides more flexibility in the way the service is provided, as a service can extend useActor to add additional logic. But as I said, I might be biased here and would appreciate other opinions, maybe @ducin has an opinion on this

packages/xstate-angular/README.md
```diff
@@ -0,0 +1,53 @@
...
+      toggleMachine.snapshot().value === 'inactive'
+        ? 'Click to activate'
+        : 'Active! Click to deactivate'
+    }}
+  </button>
+  `,
+  styleUrl: './toggle.component.css'
+})
+export class ToggleComponent {
+  public toggleMachine = inject({providedIn: "root", }ToggleMachine);
```
Andarist, line 51:
```suggestion
  public toggleMachine = inject({ providedIn: "root" }, ToggleMachine);
```

packages/xstate-angular/src/useActor.ts
```diff
@@ -0,0 +1,73 @@
...
+  signal,
+  Type,
+  WritableSignal
+} from '@angular/core';
+import { useSelector } from './useSelector.ts';
+import { useActorRef } from './useActorRef.ts';
+
+export interface UseActorConfig {
+  providedIn: 'root';
+}
```
Andarist, line 21: Could you ELI5 this thing?

Andarist: Could this become part of the `options` parameter (it could be typed as `ActorOptions<TLogic> & { providedIn?: string }`). I'm not a particular fan of making this an optional prepended parameter - it makes consuming those internally more complicated.

niklas-wortmann: possible yes, just using it wouldn't be as nice, because you can easily distinguish right now based on the usage where it is provided, e.g. 
`useActor({providedIn: 'root'}, toggleMachine)` => globally provided
`useActor(toggleMachine)` => locally provided

If I understand you correctly yours would look kinda like this,
`useActor({...toggleMachine, providedIn: 'root'})` => globally provided
`useActor(toggleMachine)` => locally provided

Andarist: I was thinking about:
```ts
useActor(toggleMachine, { providedIn: 'root' }) => globally provided
useActor(toggleMachine) => locally provided
useActor(toggleMachine, { input: 42, providedIn: 'root' }) => globally provided, with some other options
```

niklas-wortmann: I like that, I will make changes accordingly! Thanks for the input

packages/xstate-angular/package.json
niklas-wortmann: yeah it is needed. Not sure what the best way to move forward is on this one, as using babel/preconstruct doesn't make a whole lot of sense here (from what I can say)

Andarist: I can figure it out on my own before we land this - I don't want to put on you the burden of figuring intricacies of our build pipeline. I wonder, what makes it *required*? Angular surely has to be able to consume regular npm packages - what's different about a package like this one?

niklas-wortmann: https://angular.io/guide/angular-package-format They are doing pre-optimizations, it would probably work fine. 
For now I will just exclude it from preconstruct.

We can also schedule a call and figure out the build together, also feel somewhat bad to put that burden on you :D

niklas-wortmann:
What is the best way forward with this? I'd be more than happy to add tests and schematics if this is of interest (probably in a separate pr though)

Andarist:
Tests should be added here - as part of this PR. I'm not sure what do you mean by schematics. 

> Also is there anything I can do about the ci/codesandbox issue?

I wouldn't be worried about CodeSandbox. It's not a blocker - but we should bump the node version used by that tool. This can be done [here](https://github.com/statelyai/xstate/blob/f86a59ced85febed3ee67ea2bbe28fd34c465c6d/.codesandbox/ci.json#L9) but:
- I'm not sure if they read config from the PR or from the base branch
- I'm not sure if they already support node 20

It might be good to schedule that call to talk through the build intricacies if you are still up for it. I would mainly like to just get myself up to speed with the angular ways of doing things ;p

niklas-wortmann:
@Andarist Alrighty I added a couple of tests, let me know what you think.
I did test the implementation in an SSR scenario and it worked just fine with the current implementation 🎉 

I also checked in with the other Angular GDEs about the ng-packagr. It seems like ng-packagr mostly ensures efficient builds with the angular cli and the guarantee that it works with the bundler. But as Angular Package Format is just a specification it is not a set requirement to use it. So I guess you can try to move forward with preconstruct and I can help test builds if you want.

niklas-wortmann:
I also noticed that the build is now failing, it doesn't really seem to be related to any of my changes unless I am missing something?

Andarist:
> I also noticed that the build is now failing, it doesn't really seem to be related to any of my changes unless I am missing something?

It seems that you have forcefully updated the lockfile or something. I've fixed it with `git checkout origin/main yarn.lock && yarn`

> I did test the implementation in an SSR scenario and it worked just fine with the current implementation 🎉

I suspect it still might be a problem. Could you check what happens with a machine like this?
```ts
createMachine({
  entry: () => { throw new Error('oops') }
})
```

This error should only be thrown on the client side. 

>  So I guess you can try to move forward with preconstruct and I can help test builds if you want.

I'll try to do that 👍

niklas-wortmann:
You were right the error was thrown on the server, but that is now fixed!

niklas-wortmann:
The failing test actually made me aware of a behavior that I did not necessarily expect. 
With the change I did to handle the SSR case, now a machine needs to be injected into a component to be started, this means if you have a global machine but it isn't used in a component yet state transitions just get lost. This doesn't sound like the correct behavior to me.
@Andarist can you elaborate on the original issue you had in the SSR case?

Andarist:
> this means if you have a global machine but it isn't used in a component yet state transitions just get lost.

Could you elaborate on what happens in this scenario? 

> @Andarist can you elaborate on the original issue you had in the SSR case?

In every other framework actors are not meant to be started at the server. Each SSR render (yes, ATM machine) is meant to be scoped to the current request. If an actor gets started on the server then it might execute some side-effects and SSR is meant to be pure-ish when it comes to what happens in the components to avoid leaks between requests and stuff like that. Also, we need a "symmetry" - if an actor gets started during SSR... what about the cleanup?

niklas-wortmann:
> Could you elaborate on what happens in this scenario?

So the machine will be started if it is injected in a component that is getting rendered. If this is not the case because it is a global state that is not rendered somewhere yet, every state transition will get lost. 

Here's an example that makes it more tangible

```
const counterMachine = createMachine({
  initial: 'active',
  context: {
    count: 0
  },
  states: { active: {
      on: {
        INC: {
          // Inferred automatically in most cases
          actions: assign({
            count: (context: any) => {
              // context: { count: number }
              return context.context.count + 1;
            }
          })
        },
        DEC: {
          // Inferred automatically in most cases
          actions: assign({
            count: (context: any) => {
              console.log(context)
              // context: { count: number }
              return context.context.count - 1;
            }
          })
        }
      }
    }}
});

const CounterMachine = useMachine(counterMachine, {providedIn: "root"})

@Component({
  selector: 'app-test',
  standalone: true,
  imports: [
    JsonPipe
  ],
  providers: [CounterMachine],
  template: `
    <button (click)="counterMachine.send({ type: 'DEC' })"

    >Decrement</button>
    {{counterMachine.snapshot().context.count }}
    <button (click)="counterMachine.send({ type: 'INC' })">Increment</button>

  `,
  styleUrl: './test.component.css'
})
export class TestComponent {
  protected counterMachine = inject(CounterMachine)
}


@Injectable({providedIn: 'root'})
export class Service {

  constructor() {
    const machine = inject(CounterMachine);
    machine.send({ type: 'INC' }); // THIS IS NOT HAVING ANY EFFECT AS THE MACHINE IS NOT STARTED YET
  }
}
```

I'd also assume that you easily get hydration mismatch errors if the machine is not started on the server but on the client (not super familiar with SSR, so I might be missing something here)

Andarist:
So I think what you are saying is that transitions won't get executed on the server side (because the machine is not started yet) and that can lead to a hydration mismatch. I mean, I'm just repeating what you said - I just want to make sure that you are talking about a *single* problem (I see those 2 as basically the same problem).

You are right but I think this is expected. If you want to avoid a hydration mismatch then you need to "stabilize" your initial value that gets SSRed - and not rely on transitions being executed there to achieve that.

niklas-wortmann:
Sorry I could have expressed this better, but I think those are 2 separate issues (at least for me)
Regarding the mismatch issues, I agree with you, this is probably easy to prevent by stabilizing your initial state.

The other issue is still relevant and unrelated to SSR, but rather to the fact that it is dependent on the component lifecycle.
The current implementation allows for using machines in a global context. This global context might not be tied to a component. As long as a globally used machine is not used within a component context state transitions will be lost, which I'd consider a bug. This bug could be prevented if the machine is started once it is injected, but from your [example](https://github.com/statelyai/xstate/pull/4816#issuecomment-2036603924) that would lead to the error being thrown on the server side. I guess I don't understand the issue around the error being thrown on the server side properly, to make some sort of decision on this issue.

niklas-wortmann:
@Andarist how do you want to proceed? I can delete/fix the test or adjust the implementation?

packages/xstate-angular/src/useActor.ts
```diff
@@ -0,0 +1,51 @@
...
+      const listener = (nextSnapshot: Snapshot<unknown>) => {
+        this._snapshot?.set(nextSnapshot as any);
+      };
+
+      this.actorRef = useActorRef(actorLogic, options, listener);
+      this._snapshot = signal(useSelector(this.actorRef as any, (s) => s)());
+      this.send = this.actorRef.send;
+    }
+  }
+  return ActorStore;
```
eneajaho, line 44: Instead of creating an injectable like this, I think we can just use an injection token with a factory which should simplify this a lot.

packages/xstate-angular/src/useActorRef.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (observerOrListener) {
+    sub = actorRef.subscribe(toObserver(observerOrListener));
+  }
+
+  afterNextRender(
+    () => {
+      actorRef.start();
+    },
+    { phase: AfterRenderPhase.Read }
+  );
```
eneajaho, line 37: Can't we just start it directly? What's the idea of running it after next render?

matheo:
@niklas-wortmann thanks for this!
I'm just getting to know `xstate` and I needed a ngrx-signals'like adaptation <3

matheo:
@Andarist playing with this I wonder one thing:
If I create a `machine` and an `actor` separately, they can communicate each other via `systemId`? or they need to be in the same parent/child hierarchy? I ask this because in the Angular context we create the instances "on demand", and we are used to have independent services which would be actors in this case.

So I'm trying to create a Machine that consumes different actors via `systemId` to access the Angular providers/instances.

It works defining the actors in the machine and using `spawnChild`, but it doesn't work just by creating the actor in another section of the code with `createActor(... { ..., systemId })` .
Can you give me some hints if this is possible please? TIA

SerkanSipahi:
any updates?

davidkpiano:
Let's revisit this for v6

PR #4822, opened 2024-04-03
[core] Propagate `clock` and `logger` from system

davidkpiano:
The `clock` and `logger` specified in the `options` object of `createActor(logic, options)` will now propagate to all actors created within the same actor system.

```ts
import { setup, log, createActor } from 'xstate';

const childMachine = setup({
  // ...
}).createMachine({
  // ...
  // Uses custom logger from root actor
  entry: log('something')
});

const parentMachine = setup({
  // ...
}).createMachine({
  // ...
  invoke: {
    src: childMachine
  }
});

const actor = createActor(parentMachine, {
  logger: (...args) => {
    // custom logger for args
  }
});

actor.start();
```


packages/core/src/actions/spawnChild.ts
```diff
@@ -210,6 +212,7 @@ export function spawnChild<
   spawnChild.src = src;
   spawnChild.input = input;
   spawnChild.syncSnapshot = syncSnapshot;
+  spawnChild.logger = logger;
```
Andarist, line 215: have you considered propagating this through the system? it feels like it could be cleaner

packages/core/test/logger.test.ts
Andarist, line 1: i'd move this to some other file - I'm not sure if this single test case is worth having a separate file ;p

packages/core/src/createActor.ts
```diff
@@ -166,8 +167,8 @@ export class Actor<TLogic extends AnyActorLogic>
 
     this.sessionId = this.system._bookId();
     this.id = id ?? this.sessionId;
-    this.logger = logger;
-    this.clock = clock;
+    this.logger = options?.logger ?? this.system._logger;
+    this.clock = options?.clock ?? this.system._clock;
```
Andarist, line 171: this change should get a dedicated test case

packages/core/test/logger.test.ts
davidkpiano: I plan on adding more tests

PR #4832, opened 2024-04-07
`fromPromise` passes signal to creator

cevr:
Currently there is no notion of cancelling promises that are passed into a `fromPromise` logic creator if the state transitions before the promise is resolved.

This is a problem if the promise has side effects that run after resolving/rejecting - eg retries, cache invalidations, etc.

Now the signal can be used to cancel any side effects


packages/core/src/actors/promise.ts
```diff
@@ -132,29 +139,38 @@ export function fromPromise<TOutput, TInput = NonReducibleUnknown>(
...
-          system._relay(self, self, {
-            type: XSTATE_PROMISE_REJECT,
-            data: errorData
-          });
-        }
-      );
+        )
+        .finally(() => {
+          done = true;
+        });
```
Andarist, line 173: To avoid potential race conditions between scheduled microtasks, I would strongly prefer for this to be set asap - within each then/catch callback. Notice how this microtask is executed *before* the finally handler:
```ts
Promise.resolve(42)
  .then(() => {
    queueMicrotask(() => console.log("micro"));
  })
  .finally(() => console.log("end"));
```

Andarist: And actually, maybe you could just clear `controller` instead of introducing a new `done` variable.

packages/core/src/actors/promise.ts
```diff
@@ -88,8 +88,11 @@ export function fromPromise<TOutput, TInput = NonReducibleUnknown>(
      * The parent actor of the promise actor
      */
     self: PromiseActorRef<TOutput>;
+    signal: AbortSignal;
   }) => PromiseLike<TOutput>
 ): PromiseActorLogic<TOutput, TInput> {
+  const controller = new AbortController();
+  let done = false;
```
Andarist, line 95: This introduces shared state between all of the actors created based on this promise logic. Each actor should get its own controller/signal - it would also be great to add a test case for this.

You can use this is the starting point for this test case. The second region should be completed when we cancel the first one
```ts
const p = fromPromise(() => { /* ... */ })
createMachine({
  type: 'parallel',
  states: {
    p1: {
      initial: 'running',
      states: {
        running: {
          invoke: {
            src: p
          },
          on: {
            CANCEL_1: 'canceled'
          }
        },
        canceled: {}
      }
    },
    p2: {
      initial: 'running',
      states: {
        running: {
          invoke: {
            src: p,
            onDone: 'done'
          },
        },
        done: {}
      }
    }
  }
})
```

packages/core/test/actorLogic.test.ts
```diff
@@ -232,6 +232,46 @@ describe('promise logic (fromPromise)', () => {
...
     createActor(promiseLogic).start();
   });
+
+  it('should abort when stopping', (done) => {
+    let signal: AbortSignal;
+    const promiseLogic = fromPromise((ctx) => {
+      signal = ctx.signal;
+      return new Promise((res) => {
+        setTimeout(() => res(42), 1000);
+      });
```
Andarist, line 242: could we rewrite those test cases in a way that would show more real-life usage? `signal` shouldn't have to be leaked to the outer scope to get a meaningful test case for this

.changeset/sweet-sheep-arrive.md
```diff
@@ -0,0 +1,13 @@
+---
+'xstate': patch
```
Andarist, line 2:
```suggestion
'xstate': minor
```

packages/core/src/actors/promise.ts
```diff
@@ -88,8 +88,11 @@ export function fromPromise<TOutput, TInput = NonReducibleUnknown>(
      * The parent actor of the promise actor
      */
     self: PromiseActorRef<TOutput>;
+    signal: AbortSignal;
   }) => PromiseLike<TOutput>
 ): PromiseActorLogic<TOutput, TInput> {
+  const controller = new AbortController();
```
Andarist, line 94: `AbortController` is only supported since node 15. I think it's a reasonable threshold. XState v5 was released when node 18 was already the LTS

.changeset/sweet-sheep-arrive.md
cevr: [`446ce64` (#4832)](https://github.com/statelyai/xstate/pull/4832/commits/446ce64ce763d10357fb7ca51322ef83879f08d7)

packages/core/test/actorLogic.test.ts
cevr: [`c06b029` (#4832)](https://github.com/statelyai/xstate/pull/4832/commits/c06b029019d1aac6d9d66166a5a940c4f49661f8)

packages/core/src/actors/promise.ts
cevr: [`c06b029` (#4832)](https://github.com/statelyai/xstate/pull/4832/commits/c06b029019d1aac6d9d66166a5a940c4f49661f8)

packages/core/src/actors/promise.ts
cevr: [`c06b029` (#4832)](https://github.com/statelyai/xstate/pull/4832/commits/c06b029019d1aac6d9d66166a5a940c4f49661f8)

cevr:
@Andarist Thanks for the quick review! I added some more test cases in line with your comments, and opted to using a WeakMap to capture references to unique actors, since `id` is not guaranteed to be unique within a machine. 

What are your thoughts on that?

packages/core/src/actors/promise.ts
```diff
@@ -71,6 +72,9 @@ export type PromiseActorRef<TOutput> = ActorRefFrom<
  * // }
  * ```
  */
+
+let controllerMap = new WeakMap<AnyActorRef, AbortController>();
```
Andarist, line 76:
```suggestion
const controllerMap = new WeakMap<AnyActorRef, AbortController>();
```

packages/core/src/actors/promise.ts
```diff
@@ -114,12 +119,17 @@ export function fromPromise<TOutput, TInput = NonReducibleUnknown>(
             error: (event as any).data,
             input: undefined
           };
-        case XSTATE_STOP:
+        case XSTATE_STOP: {
+          const controller = controllerMap.get(scope.self);
+          if (controller) {
+            controller.abort();
+          }
```
Andarist, line 126:
```suggestion
          controllerMap.get(scope.self)?.abort();
          controllerMap.delete(scope.self);
```

packages/core/test/actorLogic.test.ts
```diff
@@ -1032,3 +1273,25 @@ describe('composable actor logic', () => {
...
   });
 });
+
+interface Deferred<T> {
+  resolve(value: T): void;
+  reject(reason: any): void;
+  promise: Promise<T>;
+}
+
+function defer<T>(): Deferred<T> {
```
Andarist, line 1283: You should already have this available here at type-level (but we still have to shim the runtime)
```suggestion
function withResolvers<T>(): PromiseWithResolvers<T> {
```

packages/core/test/actorLogic.test.ts
```diff
@@ -232,6 +232,247 @@ describe('promise logic (fromPromise)', () => {
...
+      });
+    });
+
+    const actor = createActor(promiseLogic).start();
+
+    actor.stop();
+
+    deferred.resolve(42);
+    await deferred.promise;
+    expect(fn).toHaveBeenCalledWith(true);
```
Andarist, line 252: I was thinking about something closer to this:
```suggestion
    const aborted = jest.fn();
    const promiseLogic = fromPromise(({ signal }) => {
      return new Promise(() => {
        signal.on("abort", aborted);
      })
    });

    const actor = createActor(promiseLogic).start();

    actor.stop();

    expect(aborted).toHaveBeenCalledWith(true);
```

Andarist:
Using `WeakMap` is fine 👍

cevr:
@Andarist Updated the tests. Though I had to add `ESNext.Promise` to the tsconfig `lib`. Is that alright?

cevr:
also my changes in [`29ae633` (#4832)](https://github.com/statelyai/xstate/pull/4832/commits/29ae6330faf88e884f3092c66285ce8d4856e6f1) seem to have broken one seemingly unrelated test 🤔 not sure how yet

Andarist:
> also my changes in [29ae633 (#4832)](https://github.com/statelyai/xstate/pull/4832/commits/29ae6330faf88e884f3092c66285ce8d4856e6f1) seem to have broken one seemingly unrelated test 🤔 not sure how yet

it's a common~ issue with async tests that share a thread ;p an error that was "created" by one of the tests might get thrown (and caught by the global handler) while another test is already executing

packages/core/test/actorLogic.test.ts
```diff
@@ -1032,3 +1264,25 @@ describe('composable actor logic', () => {
     );
   });
 });
+
+interface Deferred<T> {
+  resolve(value: T): void;
+  reject(reason: any): void;
+  promise: Promise<T>;
+}
```
Andarist, line 1272: couldn't this be replaced everywhere with `PromiseWithResolvers<T>`?

cevr: oops, forgot to remove that. thanks for catching this

.changeset/sweet-sheep-arrive.md
```diff
@@ -0,0 +1,13 @@
+---
+'xstate': minor
+---
+
+`fromPromise` now passes a signal into it's creator function.
```
davidkpiano, line 5:
```suggestion
`fromPromise` now passes a signal into its creator function.
```

PR #4839, opened 2024-04-09
[@xstate/store] Add JSDocs

davidkpiano:


Andarist (approved):

packages/xstate-store/examples/counter-react/.eslintrc.cjs
Andarist, line 1: we keep all the examples in a separate directory so this is an outlier here (in `packages`)

davidkpiano: Okay, I'll move it

PR #4844, opened 2024-04-10
[@xstate/store] Compatibility with `@xstate/react`'s `useSelector(…)` hook

davidkpiano:


packages/xstate-react/src/useSelector.ts
```diff
@@ -10,10 +10,15 @@ function defaultCompare<T>(a: T, b: T) {
   return a === b;
 }
 
-export function useSelector<TActor extends AnyActorRef | undefined, T>(
+export function useSelector<
+  TActor extends Pick<AnyActorRef, 'subscribe' | 'getSnapshot'> | undefined,
```
Andarist, line 14: why is the previous version not compatible?

davidkpiano: Because `Store` is not fully compatible with `AnyActorRef`

Andarist: Hm, ok - lots of things are missing from this. The solution is OK 👍 I would do the same in all of the other integration packages though. `@xstate/react` shouldn't be the only one that becomes compatible with this

.changeset/tall-bobcats-trade.md
```diff
@@ -0,0 +1,34 @@
+---
+'@xstate/react': patch
+'@xstate/store': patch
```
Andarist, line 3: To be fair - this change doesn't really affect `@xstate/store` so it feels off to me that it's going to be mentioned in its CHANGELOG cause of this changeset. The change itself is solely in `@xstate/react`

packages/xstate-react/src/useSelector.ts
davidkpiano: Agreed

.changeset/tall-bobcats-trade.md
Andarist:
```suggestion
```

Andarist (approved):

PR #4850, opened 2024-04-13
Fix depart date bug: 7GUIs Flight Booker React

stevebarakat:
* Fixed depart date bug
* Remodeled state machine
* Refactored code
* Improved UI
* Updated screenshot
* Added _Open in StackBlitz_ button


davidkpiano:
I was able to book a flight in the past

CleanShot 2024-04-15 at 13 48 48@2x

davidkpiano (approved):

PR #4863, opened 2024-04-22
[core] Add `meta` types

### LINKED ISSUE #809: [Feature] Enforce/describe State Meta Data with TS

davidkpiano:
Meta objects for state nodes and transitions can now be specified in `setup({ types: … })`:

```ts
const machine = setup({
  types: {
    meta: {} as {
      layout: string;
    }
  }
}).createMachine({
  initial: 'home',
  states: {
    home: {
      meta: {
        layout: 'full' // strongly typed
      }
    }
  }
});

actor.getSnapshot().getMeta().home!;
// typed as { layout: string }
```

Fixes #809

### DIFF
--- a/.changeset/metal-elephants-impress.md
+++ b/.changeset/metal-elephants-impress.md
@@ -1,0 +1,30 @@
+---
+'xstate': minor
+---
+
+Meta objects for state nodes and transitions can now be specified in `setup({ types: … })`:
+
+```ts
+const machine = setup({
+  types: {
+    meta: {} as {
+      layout: string;
+    }
+  }
+}).createMachine({
+  initial: 'home',
+  states: {
+    home: {
+      meta: {
+        layout: 'full'
+      }
+    }
+  }
+});
+
+const actor = createActor(machine).start();
+
+actor.getSnapshot().getMeta().home;
+// => { layout: 'full' }
+// if in "home" state
+```
--- a/packages/core/src/State.ts
+++ b/packages/core/src/State.ts
@@ -16,7 +16,8 @@ import type {
   AnyActorRef,
   Snapshot,
   ParameterizedObject,
-  IsNever
+  IsNever,
+  MetaObject
 } from './types.ts';
 import { matchesState } from './utils.ts';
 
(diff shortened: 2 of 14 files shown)


jonestristand:
This would be very useful to me!

davidkpiano:
cc. @Andarist Quick review? 🥺

Andarist (approved):

jonestristand:
Amazing, thank you!

PR #4876, opened 2024-04-29
[core] Add warning for imperative built-in action calls

davidkpiano:
This PR warns for improper imperative calls of built-in actions, which is a very common problem.

https://stately.ai/docs/actions#built-in-actions

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,6 +4,7 @@ const { constants } = require('jest-config');
  * @type {import('@jest/types').Config.InitialOptions}
  */
 module.exports = {
+  prettierPath: null,
   setupFilesAfterEnv: ['@xstate-repo/jest-utils/setup'],
   transform: {
     [constants.DEFAULT_JS_PATTERN]: 'babel-jest',
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -1554,7 +1554,6 @@ function resolveAndExecuteActionsWithContext(
           : undefined;
 
     function executeAction() {
-      executingCustomAction = resolvedAction;
       actorScope.system._sendInspectionEvent({
         type: '@xstate.action',
         actorRef: actorScope.self,
@@ -1569,6 +1568,7 @@ function resolveAndExecuteActionsWithContext(
         }
       });
       try {
+        executingCustomAction = resolvedAction;
         resolvedAction(actionArgs, actionParams);
       } finally {
         executingCustomAction = false;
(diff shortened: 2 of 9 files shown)


jest.config.js
```diff
@@ -4,6 +4,7 @@ const { constants } = require('jest-config');
  * @type {import('@jest/types').Config.InitialOptions}
  */
 module.exports = {
+  prettierPath: null,
```
davidkpiano, line 7: Mitigates this error:
```
Jest: Inline Snapshots are not supported when using Prettier 3.0.0 or above.
```

fekete965 (approved):
LGTM 👍

davidkpiano:
Ping @Andarist

jest.config.js
Andarist: Supposedly it is fixed in Jest@30 (alpha) but 29 is still the latest version 😢

packages/core/src/stateUtils.ts
```diff
@@ -1563,15 +1567,25 @@ function resolveAndExecuteActionsWithContext(
...
         }
       });
-      resolvedAction(actionArgs, actionParams);
+      try {
+        resolvedAction(actionArgs, actionParams);
+      } catch (err) {
+        executingCustomAction = false;
+        // rethrow err
+        throw err;
+      }
```
Andarist, line 1576: this would have the same effect
```suggestion
      } finally {
        executingCustomAction = false;
      }
```

packages/core/src/stateUtils.ts
```diff
@@ -1563,15 +1567,25 @@ function resolveAndExecuteActionsWithContext(
...
     }
 
     if (!('resolve' in resolvedAction)) {
       if (actorScope.self._processingStatus === ProcessingStatus.Running) {
+        executingCustomAction = resolvedAction;
         executeAction();
+        executingCustomAction = false;
```
Andarist, line 1583: why do we reset this both here and in `executeAction`? couldn't everything related to this change be handled by `executeAction`?

davidkpiano: [e1ec0f2](https://github.com/statelyai/xstate/pull/4876/commits/e1ec0f267cf8a65c333e1ae4370ca08a7c33c779)

packages/core/src/stateUtils.ts
```diff
@@ -1550,6 +1554,7 @@ function resolveAndExecuteActionsWithContext(
           : undefined;
 
     function executeAction() {
+      executingCustomAction = resolvedAction;
```
Andarist, line 1557: i'd move this into the try block, it will be more clear that it goes together with the reset in the finally block

Andarist (approved):

PR #4896, opened 2024-05-14
Combine `@xstate/graph` and `@xstate/test`

davidkpiano:


davidkpiano:
@Andarist When you get the chance, can you set this up so that it will do a proper 2.0 release?

davidkpiano:
@Andarist Ping 🏓

.changeset/eighty-papayas-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/graph': patch
+---
+
+The `@xstate/graph` package now includes everything from `@xstate/test`.
```
Andarist, line 5: What are the current roles of both packages? I guess that with the "combine" in the title of this PR I expected this one would consolidate those 2 in a single package but we still have 2

packages/xstate-graph/src/TestModel.ts
Andarist, line 1: I diffed this and the other one in xstate-test. They are **very very** similar but yet they have some very small differences. I'm confused 😅

.changeset/config.json
```diff
@@ -5,12 +5,7 @@
   "linked": [],
   "access": "public",
   "baseBranch": "main",
-  "ignore": [
```
Andarist, line 8: q: do u plan to release `@xstate/graph` as a stable release **soon**? we probably should add `@xstate/graph` back here and do some cleanup in https://github.com/statelyai/xstate/tree/7c6e2eaa4e5021738efec345b51f59a8e2de662e/.changeset-xstate-test-beta and only then execute the final release

packages/xstate-graph/test/shortestPaths.test.ts
```diff
@@ -131,7 +131,7 @@ describe('getShortestPaths', () => {
           todo: 'two'
         } as const
       ],
-      filter: (state) => state.context.todos.length < 3
+      stopWhen: (state) => state.context.todos.length >= 3
```
Andarist, line 134: this feels like a breaking change that wasn't mentioned in changesets

davidkpiano: I will add a changeset in a follow-up (clean-up) PR

.changeset/config.json
davidkpiano: Yes, I plan to release it soon

packages/xstate-graph/test/shortestPaths.test.ts
davidkpiano: #4911

.changeset/config.json
davidkpiano: #4911

packages/xstate-graph/src/TestModel.ts
davidkpiano: Changesets in #4911

.changeset/eighty-papayas-exist.md
davidkpiano: Deleted `@xstate/test` in #4911

PR #4898, opened 2024-05-16
Add Trivia Game App to examples 

javiert01:
This is a trivia game app made using XState v5. The purpose is to give users an example of how to use the different functionality, methods and properties of actors and state machines.


examples/trivia-game-example/public/vite.svg
davidkpiano, line 1: This can probably be removed

examples/trivia-game-example/src/machines/triviaMachine.ts
davidkpiano, line 1: Tiny suggestion - may be better to put this directly in `src/` for easier discoverability (flattening file structure)

examples/trivia-game-example/src/machines/triviaMachine.ts
```diff
@@ -0,0 +1,245 @@
...
+            randomCharacters: Array<RMCharacter>
+            isClueOpened: boolean
+            points: number
+            question: number
+            lifes: number
+        }
+    },
+    guards: {
+        isAnswerCorrect: ({context, event}) => {
+            if(event.type !== "user.selectAnswer") return false
```
davidkpiano, line 33: Nit: better to use `params` as a future refactor item, but for now, since we're only expecting the guard to be called in one place:

```suggestion
            assertEvent(event, 'user.selectAnswer');
```

davidkpiano (commented):
This is very awesome! Just some small nits.

examples/trivia-game-example/src/machines/triviaMachine.ts
javiert01: Sounds good!

javiert01:
Thanks for the review @davidkpiano ! I have updated the PR

davidkpiano (approved):

PR #4903, opened 2024-05-19
Example: 7GUIs-circleDrawer-react

stevebarakat:


examples/7GUIs-circle-drawer-react/.eslintrc.cjs
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
davidkpiano, line 1: Let's remove this file in favor of a shared `.eslintrc` in `examples/`

PR #4905, opened 2024-05-19
[core] Wildcard emitted event listener

davidkpiano:
You can now use a wildcard to listen for _any_ emitted event from an actor:

```ts
actor.on('*', (emitted) => {
  console.log(emitted); // Any emitted event
});
```

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -2148,8 +2148,7 @@ export interface ActorLike<TCurrent, TEvent extends EventObject>
 
 export interface ActorRef<
   TSnapshot extends Snapshot<unknown>,
-  TEvent extends EventObject,
-  TEmitted extends EventObject = EventObject
+  TEvent extends EventObject
 > extends Subscribable<TSnapshot>,
     InteropObservable<TSnapshot> {
   /**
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -2171,13 +2170,9 @@ export interface ActorRef<
   /** @internal */
   _processingStatus: ProcessingStatus;
   src: string | AnyActorLogic;
-  on: <TType extends TEmitted['type']>(
-    type: TType,
-    handler: (emitted: TEmitted & { type: TType }) => void
-  ) => Subscription;
```
Andarist, line 2177: this could easily be seen as a breaking change

davidkpiano: How? It still shows up on the Actor type, which is what is returned from createActor. It shouldn't have been on ActorRef, since that's supposed to be a minimal interface to reference the actor. 

I can add it back, though.

Andarist: It was on an `ActorRef` as that's what u get from `.children` etc. Without it you can't easily subscribe to events emitted within the same "actor tree"

davidkpiano: Added back

davidkpiano:
@Andarist ping 🏓

PR #4911, opened 2024-05-27
Release `@xstate/graph`

davidkpiano:
Getting `@xstate/graph` ready for release


.changeset/eleven-bees-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/graph': major
+---
+
+The `adjacencyMapToArray(…)` helper function has been introduced, which converts an adjacency map to an array of `{ state, event, nextState }` objects.
```
with-heart, line 5: Might be a good place to add a simple use case example. I don't really know what this means or how I might use it.

davidkpiano: Added an example

.changeset/eight-pianos-divide.md
```diff
@@ -0,0 +1,24 @@
+---
+'@xstate/graph': major
```
Andarist, line 2: i'm not sure how those major bumps will behave here since this package is already in the prerelease mode (sort of). We can review what Changesets will do to all of this after this PR lands - then we'll be able to just tweak the content or the changeset files as needed

.changeset/eight-pianos-divide.md
```diff
@@ -0,0 +1,24 @@
...
+---
```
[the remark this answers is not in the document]
Andarist: that's just to fix the generated links because those changes were not introduced by this PR here but by the previous one

.changeset/eight-pianos-divide.md
```diff
@@ -0,0 +1,28 @@
...
+
+pr: #4896
+commit: 7c6e2ea
```
Andarist, line 6: just remember to add this to other changeset files here - i think at least most of them should have this but u'll be able to tell best

PR #4929, opened 2024-06-08
feat(core): expose UnknownActorRef

boneskull:
`AnyActorRef` is problematic for some use cases, because the result of the `AnyActorRef['getSnapshot']` function is of type `any`.

However, changing `AnyActorRef` to use a narrow type in `ActorRef`'s first type argument, `TSnapshot`, breaks conditional types which perform inference based on `TSnapshot`.

This change introduces `UnknownActorRef`, which is like `AnyActorRef`, but is not intended to be inferred from in future conditional types. A consumer can use `UnknownActorRef` like so:

```ts
const actor: UnknownActorRef = getSomeActor();

// inferred as `Snapshot<unknown>`
const snapshot = actor.getSnapshot();

// do things w/ snapshot
```

`AnyActorRef` would return `any` from `actor.getSnapshot()`, which make it unsuitable for this use-case in a strictly-typed environment.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -2186,6 +2186,8 @@ export interface ActorRef<
 
 export type AnyActorRef = ActorRef<any, any, any>;
 
+export type UnknownActorRef = ActorRef<Snapshot<unknown>, EventObject>;
+
 export type ActorLogicFrom<T> = ReturnTypeOrValue<T> extends infer R
   ? R extends StateMachine<
       any,
--- a/.changeset/strange-fireants-tap.md
+++ b/.changeset/strange-fireants-tap.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Expose type `UnknownActorRef` for use when calling `getSnapshot()` on an unknown `ActorRef`.
--- a/packages/core/test/types.test.ts
+++ b/packages/core/test/types.test.ts
@@ -2,7 +2,12 @@ import { from } from 'rxjs';
 import { log } from '../src/actions/log';
 import { raise } from '../src/actions/raise';
 import { stopChild } from '../src/actions/stopChild';
-import { PromiseActorLogic, fromCallback, fromPromise } from '../src/actors';
+import {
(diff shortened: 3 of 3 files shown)


boneskull:
regardless; if this (or the other suggestion) makes sense, I'll go in and make the change and fix the type tests.

davidkpiano:
Thank you for this! It seems like TS is complaining in some of the test files (you can run `yarn typecheck` to check) 🕵️ 

cc. @Andarist

boneskull:
@davidkpiano I don't quite understand why the 3rd `any` in `AnyActorRef` was removed in dbeafeb25eed63ee1d2b027f10bc63d9937ab073 (#4905).  That change seems to break some stuff anyhow.  I'll create a simple reproduction.

boneskull:
See https://github.com/statelyai/xstate/issues/4931 for the "other problem" I mentioned above.

davidkpiano:
There may be some overlap with #4932

boneskull:
@davidkpiano Yeah.  I'll wait until #4932 is merged then retry on top of it.

davidkpiano:
@boneskull It's merged; please merge `main` and try again?

boneskull:
looking

boneskull:
@davidkpiano OK, I rebased, but changed my strategy. Please read updated issue description

packages/core/src/types.ts
```diff
@@ -2186,6 +2186,8 @@ export interface ActorRef<
 
 export type AnyActorRef = ActorRef<any, any, any>;
 
+export type SomeActorRef = ActorRef<Snapshot<unknown>, EventObject>;
```
Andarist, line 2189: I'd call this `UnknownActorRef`. It might fit better some other existing types of this kind

boneskull: works for me

boneskull:
@davidkpiano I've added a changeset for this.  I'm calling it "patch" since it doesn't affect any functionality. Happy to change it

Andarist (approved):

PR #4932, opened 2024-06-10
[core] Fix type issues with emitted events

### LINKED ISSUE #4931: Bug: type disagreement between emitted events and toPromise
### XState version

XState version 5

### Description

After #4905, there's some problem using `toPromise` on an actor created from a machine where `types.emitted` is specified.  

Example machine:

```ts
const machine = setup({
  types: {
    emitted: {} as {type: 'FOO'}
  }
}).createMachine({
  initial: 'init',
  states: {
    init: {
      entry: [
        emit({type: 'FOO'})
      ],
      always: 'done'
    },
    done: {type: 'final'}
  }
});
```

Usage:

```ts
await toPromise(actor) // <-- TS error
```

### Expected result

Would not expect a TS error here.

### Actual result

(See [playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgUxMGAaOBnZMCuYWMEAClBGrlgMZTICGMyAgjSVHAL5wBmFIOAHIAHthhNkQgNwAoWTQgA7cXBAMaAC2BLkcALw48hABQJZcODACeYZNgBciC5ZRoYzACZOEPBtkQbOychADEAeXChLhcYrgBKADo6RmYAWQ1tXTMXHXRgBgAbELyYIQwXcUlHZ1c4Up8XOstkJRgoaycAbSbmutR0MyDkEIiohN66gF0KvssigHcGaxqhT2UpXq5Z13XdH2GQ3h0i6NjZBLkFZVUNDgM4FMk2DhN1LR1kJKrYE3j5EjkSjAXAmO7Qf6yIA)

```
Argument of type 'Actor<StateMachine<MachineContext, AnyEventObject, {}, never, never, never, never, "init" | "done", string, NonReducibleUnknown, NonReducibleUnknown, { ...; }, MetaObject, ResolveTypegenMeta<...>>>' is not assignable to parameter of type 'AnyActorRef'.
  Types of property 'on' are incompatible.
    Type '<TType extends "FOO" | "*">(type: TType, handler: (emitted: { type: "FOO"; } & (TType extends "*" ? {} : { type: TType; })) => void) => Subscription' is not assignable to type '<TType extends string>(type: TType, handler: (emitted: EventObject & (TType extends "*" ? {} : { type: TType; })) => void) => Subscription'.
      Types of parameters 'type' and 'type' are incompatible.
        Type 'TType' is not assignable to type '"FOO" | "*"'.
          Type 'string' is not assignable to type '"FOO" | "*"'.
```

### Reproduction

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgUxMGAaOBnZMCuYWMEAClBGrlgMZTICGMyAgjSVHAL5wBmFIOAHIAHthhNkQgNwAoWTQgA7cXBAMaAC2BLkcALw48hABQJZcODACeYZNgBciC5ZRoYzACZOEPBtkQbOychADEAeXChLhcYrgBKADo6RmYAWQ1tXTMXHXRgBgAbELyYIQwXcUlHZ1c4Up8XOstkJRgoaycAbSbmutR0MyDkEIiohN66gF0KvssigHcGaxqhT2UpXq5Z13XdH2GQ3h0i6NjZBLkFZVUNDgM4FMk2DhN1LR1kJKrYE3j5EjkSjAXAmO7Qf6yIA

### Additional context

_No response_

davidkpiano:
Fixes #4931

### DIFF
--- a/packages/core/src/actors/index.ts
+++ b/packages/core/src/actors/index.ts
@@ -27,7 +27,8 @@ const emptyLogic = fromTransition((_) => undefined, undefined);
 
 export function createEmptyActor(): ActorRef<
   Snapshot<undefined>,
-  AnyEventObject
+  AnyEventObject,
+  any
 > {
   return createActor(emptyLogic);
 }
--- a/packages/core/src/actors/index.ts
+++ b/packages/core/src/actors/index.ts
@@ -28,7 +28,7 @@ const emptyLogic = fromTransition((_) => undefined, undefined);
 export function createEmptyActor(): ActorRef<
   Snapshot<undefined>,
   AnyEventObject,
-  any
+  AnyEventObject
 > {
   return createActor(emptyLogic);
 }
--- a/packages/core/src/createActor.ts
+++ b/packages/core/src/createActor.ts
@@ -71,7 +71,8 @@ const defaultOptions = {
  * An Actor is a running process that can receive events, send events and change its behavior based on the events it receives, which can cause effects outside of the actor. When you run a state machine, it becomes an actor.
  */
 export class Actor<TLogic extends AnyActorLogic>
-  implements ActorRef<SnapshotFrom<TLogic>, EventFromLogic<TLogic>>
+  implements
+    ActorRef<SnapshotFrom<TLogic>, EventFromLogic<TLogic>, EmittedFrom<TLogic>>
 {
   /**
    * The current internal state of the actor.
@@ -106,7 +107,11 @@ export class Actor<TLogic extends AnyActorLogic>
   public _parent?: AnyActorRef;
   /** @internal */
   public _syncSnapshot?: boolean;
-  public ref: ActorRef<SnapshotFrom<TLogic>, EventFromLogic<TLogic>>;
+  public ref: ActorRef<
+    SnapshotFrom<TLogic>,
+    EventFromLogic<TLogic>,
+    EmittedFrom<TLogic>
+  >;
   // TODO: add typings for system
   private _actorScope: ActorScope<
     SnapshotFrom<TLogic>,
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -121,7 +121,8 @@ export interface UnifiedArg<
       unknown,
       TODO // TMeta
     >,
-    TEvent
+    TEvent,
+    any // TODO
   >;
   system: AnyActorSystem;
 }
@@ -169,7 +170,7 @@ export type OutputFrom<T> = T extends ActorLogic<
   infer _TEmitted
 >
   ? (TSnapshot & { status: 'done' })['output']
-  : T extends ActorRef<infer TSnapshot, infer _TEvent>
+  : T extends ActorRef<infer TSnapshot, infer _TEvent, infer _TEmitted>
     ? (TSnapshot & { status: 'done' })['output']
     : never;
 
@@ -1521,7 +1522,8 @@ export type ContextFactory<
       unknown,
       TODO // TMeta
     >,
-    TEvent
+    TEvent,
+    any // TODO
   >;
 }) => TContext;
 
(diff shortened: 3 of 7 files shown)


packages/core/src/createActor.ts
```diff
@@ -106,7 +107,11 @@ export class Actor<TLogic extends AnyActorLogic>
   public _parent?: AnyActorRef;
   /** @internal */
   public _syncSnapshot?: boolean;
-  public ref: ActorRef<SnapshotFrom<TLogic>, EventFromLogic<TLogic>>;
+  public ref: ActorRef<
+    SnapshotFrom<TLogic>,
+    EventFromLogic<TLogic>,
+    EmittedFrom<TLogic>
+  >;
```
boneskull, line 114: It looks like this field (`ref`) is unused in any practical sense.  Backwards-compat thing?

packages/core/src/types.ts
```diff
@@ -169,7 +170,7 @@ export type OutputFrom<T> = T extends ActorLogic<
   infer _TEmitted
 >
   ? (TSnapshot & { status: 'done' })['output']
-  : T extends ActorRef<infer TSnapshot, infer _TEvent>
+  : T extends ActorRef<infer TSnapshot, infer _TEvent, infer _TEmitted>
```
boneskull, line 173: it looks like these unused type args are just a stylistic convention?

packages/xstate-solid/src/fromActorRef.ts
```diff
@@ -4,7 +4,9 @@ import { createImmutable } from './createImmutable.ts';
 
 const noop = () => {};
 
-export function fromActorRef<TActor extends ActorRef<any, any> | undefined>(
+export function fromActorRef<
+  TActor extends ActorRef<any, any, any> | undefined
```
boneskull, line 8:
```suggestion
  TActor extends AnyActorRef | undefined
```

packages/core/src/types.ts
davidkpiano: Not convention; without this, the alternative is `<infer Something, any, any>` and those `any`s can actually break inference because they may not match with whatever you're checking with `T extends ...`. So by using `<infer Something, infer _Unused, ...>`, you're saying "I don't care what that type is (more broad than `any`); I'm not going to use it anyway"

packages/core/src/createActor.ts
davidkpiano: Yeah, will remove in a separate PR

davidkpiano:
cc. @Andarist

packages/core/src/actors/index.ts
```diff
@@ -27,7 +27,8 @@ const emptyLogic = fromTransition((_) => undefined, undefined);
 
 export function createEmptyActor(): ActorRef<
   Snapshot<undefined>,
-  AnyEventObject
+  AnyEventObject,
+  any
```
Andarist, line 31: Would it still work with this?
```suggestion
  AnyEventObject
```

packages/core/src/createActor.ts
Andarist: it's a public property, removing it now would be a breaking change

packages/core/test/emit.test.ts
```diff
@@ -3,7 +3,8 @@ import {
   createActor,
   createMachine,
   enqueueActions,
-  setup
+  setup,
+  toPromise
```
Andarist, line 7:
```suggestion
  setup
```

packages/core/test/types.test.ts
```diff
@@ -4552,4 +4553,17 @@ describe('snapshot methods', () => {
     snapshot.getMeta();
     snapshot.toJSON();
   });
+
+  // https://github.com/statelyai/xstate/issues/4931
+  it('fromPromise should not have issues with actors with emitted types', () => {
```
Andarist, line 4558: describe blocks are evil 🤪 this should be moved outside of the current one because it has nothing to do with it

Andarist (approved):

PR #4935, opened 2024-06-12
[core] Add emit support for all actor logic creators

davidkpiano:
This PR adds support for emitting events from all actor logic types:

**Promise actors**

```ts
const logic = fromPromise(async ({ emit }) => {
  // ...
  emit({
    type: 'emitted',
    msg: 'hello'
  });
  // ...
});
```

**Transition actors**

```ts
const logic = fromTransition((state, event, { emit }) => {
  // ...
  emit({
    type: 'emitted',
    msg: 'hello'
  });
  // ...
  return state;
}, {});
```

**Observable actors**

```ts
const logic = fromObservable(({ emit }) => {
  // ...

  emit({
    type: 'emitted',
    msg: 'hello'
  });

  // ...
});
```

**Callback actors**

```ts
const logic = fromCallback(({ emit }) => {
  // ...
  emit({
    type: 'emitted',
    msg: 'hello'
  });
  // ...
});
```

### DIFF
--- a/.changeset/smooth-crabs-dress.md
+++ b/.changeset/smooth-crabs-dress.md
@@ -2,7 +2,7 @@
 'xstate': minor
 ---
 
-All actor logic creators now support emitting events:
+All actor logic creators now support [emitting events](https://stately.ai/docs/event-emitter):
 
 **Promise actors**
 
--- a/.changeset/smooth-crabs-dress.md
+++ b/.changeset/smooth-crabs-dress.md
@@ -1,0 +1,60 @@
+---
+'xstate': minor
+---
+
+All actor logic creators now support emitting events:
+
+**Promise actors**
+
+```ts
+const logic = fromPromise(async ({ emit }) => {
+  // ...
+  emit({
+    type: 'emitted',
+    msg: 'hello'
+  });
+  // ...
+});
+```
+
+**Transition actors**
+
+```ts
+const logic = fromTransition((state, event, { emit }) => {
+  // ...
+  emit({
+    type: 'emitted',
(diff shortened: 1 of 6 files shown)


packages/core/src/actors/callback.ts
```diff
@@ -171,7 +176,8 @@ export function fromCallback<
         receive: (listener) => {
           callbackState.receivers ??= new Set();
           callbackState.receivers.add(listener);
-        }
+        },
+        emit
```
Andarist, line 180: could you add a test that would successfully `emit` from a *rehydrated* callback actor?

davidkpiano: I will do this in a separate PR

Andarist: 1. create a machine with a callback actor
2. persist the root root actor
3. rehydrate a new instance of the root actor using the persisted snapshot
4. now verify that the *rehydrated* actor can correctly use `emit` after step 3

davidkpiano: 53c3c4dd58d29cf003959ffbd5ae279a01cd2ece

.changeset/smooth-crabs-dress.md
```diff
@@ -0,0 +1,60 @@
+---
+'xstate': minor
+---
+
+All actor logic creators now support emitting events:
```
with-heart, line 5:
```suggestion
All actor logic creators now support [emitting events](https://stately.ai/docs/event-emitter):
```

Andarist (approved):

PR #4936, opened 2024-06-12
[core] Inspect subscription

davidkpiano:
Inspecting an actor system via `actor.system.inspect(ev => …)` now accepts a function or observer, and returns a subscription:

```ts
const actor = createActor(someMachine);

const sub = actor.system.inspect((inspectionEvent) => {
  console.log(inspectionEvent);
});

// Inspection events will be logged
actor.start();
actor.send({ type: 'anEvent' });

// ...

sub.unsubscribe();

// Will no longer log inspection events
actor.send({ type: 'someEvent' });
```

### DIFF
--- a/.changeset/cold-planes-clap.md
+++ b/.changeset/cold-planes-clap.md
@@ -1,0 +1,24 @@
+---
+'xstate': minor
+---
+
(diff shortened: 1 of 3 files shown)


boneskull:
The subscription is welcome, but this also makes it easier to attach an inspector on-the-fly, because you don't need to use `toObserver` to do so.

Andarist (approved):

PR #4942, opened 2024-06-15
feat(core): add actor ID to DoneActorEvent, ErrorActorEvent

boneskull:
This changes `DoneActorEvent` (`xstate.done.actor.*`) and `ErrorActorEvent` (`xstate.error.actor.*`) to have an `actorId` prop which is equal to the ID of the Actor associated with the event.

Previously, this information was only available by extracting it from the `type` property of the event object using string operations.

* * *

Updated a test for `DoneActorEvent` to prove that `actorId` gets passed, but may want the same test for an `ErrorActorEvent`.

Ref: https://discord.com/channels/795785288994652170/1250894522616123494

### DIFF
--- a/packages/core/test/invoke.test.ts
+++ b/packages/core/test/invoke.test.ts
@@ -20,7 +20,8 @@ import {
   createActor,
   sendParent,
   Snapshot,
-  ActorRef
+  ActorRef,
+  DoneActorEvent
 } from '../src/index.ts';
 import { sleep } from '@xstate-repo/jest-utils';
 
@@ -3081,7 +3082,7 @@ describe('invoke', () => {
   });
 
   it('xstate.done.actor events should have unique names when invokee is a machine with an id property', (done) => {
-    const actual: string[] = [];
+    const actual: DoneActorEvent<void>[] = [];
 
     const childMachine = createMachine({
       id: 'child',
@@ -3121,7 +3122,7 @@ describe('invoke', () => {
       on: {
         '*': {
           actions: ({ event }) => {
-            actual.push(event.type);
+            actual.push(event as any);
           }
         }
       }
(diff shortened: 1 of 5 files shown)


davidkpiano (commented):
This looks great. I'm wondering if we should use `actorId` instead of `id`, since intuitively, `id` is self-referential (e.g. the `id` of the event) rather than something that references something else.

boneskull:
New prop is now `actorId`

packages/core/test/invoke.test.ts
```diff
@@ -3121,7 +3122,7 @@ describe('invoke', () => {
       on: {
         '*': {
           actions: ({ event }) => {
-            actual.push(event.type);
+            actual.push(event as any);
```
Andarist, line 3125: we shouldn't have to cast here, what is the reporter error without it?

boneskull: This test machine does not declare its event types, so any event it handles will be of type `AnyEventObject`--which is not `DoneActorEvent<void>` (which `actual` is an array of).  We could cast to `DoneActorEvent<void>`, but that seemed a little wordy. 🙂

boneskull: Previously, `actual` was `string[]`, which did not need a type assertion.

packages/core/test/invoke.test.ts
```diff
@@ -3081,7 +3082,7 @@ describe('invoke', () => {
   });
 
   it('xstate.done.actor events should have unique names when invokee is a machine with an id property', (done) => {
-    const actual: string[] = [];
+    const actual: DoneActorEvent<void>[] = [];
```
davidkpiano, line 3085:
```suggestion
    const actual: AnyEventObject[] = [];
```

packages/core/test/invoke.test.ts
```diff
@@ -20,7 +20,8 @@ import {
   createActor,
   sendParent,
   Snapshot,
-  ActorRef
+  ActorRef,
+  DoneActorEvent
```
davidkpiano, line 24:
```suggestion
  AnyEventObject
```

packages/core/test/invoke.test.ts
davidkpiano, line 3125:
```suggestion
            actual.push(event);
```

boneskull:
@davidkpiano Suggestions committed

davidkpiano (approved):
Just need to add a changeset (minor change: `yarn changeset`) 🚀 

cc. @Andarist

boneskull:
@davidkpiano Is that something I should be doing?

davidkpiano:
> @davidkpiano Is that something I should be doing?

Yes please

boneskull:
@davidkpiano Done. Hope I did that right. It looks like a minor bump to me

PR #4944, opened 2024-06-16
[core] Add basic support for state `invariant`

davidkpiano:


packages/core/test/invariant.test.ts
Andarist, line 1: we need such tests:
- show the behavior of those invariants for transitions within states that already have `invariant`, the current tests only show what's the behavior when *entering* such states
- show that `invariant` doesn't create a problems when we exit a state that has some invariant requirement (like, the invariant requires `context.user` but an exit action within the exited state resets the `user` to `null`)
- show the interactions with `entry`, `always` and `parallel`

davidkpiano: Added most of these

packages/core/test/invariant.test.ts
```diff
@@ -0,0 +1,221 @@
...
+    });
+    actor.start();
+
+    actor.send({ type: 'logout' });
+
+    expect(spy).not.toHaveBeenCalled();
+    expect(actor.getSnapshot().value).toEqual('idle');
+  });
+
+  it('interacts correctly with parallel states', () => {
```
Andarist, line 147: this test title tells me nothing ;p could we use something more specific? what kind of actual granular-ish behavior is being tested here?

davidkpiano: [879ae52](https://github.com/statelyai/xstate/pull/4944/commits/879ae529e2f354af0b8e773d6cf4c96d78de1b50)

PR #4949, opened 2024-06-20
[core] Remove typegen types

davidkpiano:
This PR removes typegen types to clean up the types. Typegen is _not_ used in XState v5.


with-heart:
RIP but also good riddance

packages/xstate-react/test/typegenTypes.test.tsx
Andarist, line 1: this whole file is related to typegen so it should be removed with typegen's removal

packages/core/src/types.ts
```diff
@@ -1259,239 +1252,53 @@ export interface MachineImplementationsSimplified<
...
-      Prop<TIndexedDelays, keyof TIndexedDelays>,
-      ParameterizedObject
-    >['type'],
-    Cast<TEmitted, EventObject>
+type MachineImplementationsActions<TTypes extends StateMachineTypes> = {
+  [K in TTypes['actions']['type']]?: ActionFunction<
+    TTypes['context'],
+    TTypes['events'],
+    TTypes['events'],
+    (TTypes['actions'] & { type: K })['params'],
```
Andarist, line 1260: we had problems with this pattern elsewhere, see: https://github.com/statelyai/xstate/pull/4568/files#diff-af7184a4b0cdcc69ce20eda13e1aceda04efde70d1b26482886cf31e245ad049R40-R46

PR #4954, opened 2024-06-26
[core] Transition function

davidkpiano:
This PR adds the `transition(…)` function for returning a `[snapshot, actions]` tuple given the machine, state, and event.

Its main use-case is for server-side workflows and a pure way of handling state transitions for actor logic.


with-heart:
Just a reminder to remove commented lines before merging 💜

packages/core/src/StateNode.ts
```diff
@@ -220,8 +206,28 @@ export class StateNode<
...
+      action: UnknownAction,
+      kind: 'entry' | 'exit',
+      i: number
+    ): UnknownActionObject => {
+      if (typeof action === 'string') {
+        return { type: action };
+      }
+      if (typeof action === 'function' && !('resolve' in action)) {
+        const type = `${this.id}|${kind}:${i}`;
+        this.machine.implementations.actions[type] = action as any;
```
Andarist, line 219: 1. I don't particularly like that we are writing to this object because it might be shared among different machines in some way and we might potentially leak things between instances
2. i'd be wary of the code [here](https://github.com/statelyai/xstate/blob/6341cfbb00b3767dd075ffe49466dd2e58d4f948/packages/xstate-react/src/useActorRef.ts#L36-L40), I wonder if this gets "fixed" somehow but I think there might run into stale closure hazard by resolving this here and caching it

packages/core/src/StateNode.ts
```diff
@@ -220,8 +206,28 @@ export class StateNode<
     this.history =
       this.config.history === true ? 'shallow' : this.config.history || false;
 
-    this.entry = toArray(this.config.entry).slice();
-    this.exit = toArray(this.config.exit).slice();
+    const convertAction = (
```
Andarist, line 209: we also have a very similar thing in stateUtils now

packages/core/src/stateUtils.ts
```diff
@@ -333,9 +346,25 @@ export function formatTransition<
...
+
+  const convertAction = (
+    action: UnknownAction,
+    i: number
+  ): UnknownActionObject => {
+    if (typeof action === 'string') {
+      return { type: action };
+    }
+    if (typeof action === 'function' && !('resolve' in action)) {
+      const type = `${stateNode.id}|${descriptor}:${index}:${i}`;
```
Andarist, line 358: is index necessary in this `type`? shouldn't `i` already be enough? for the record: I'm not sure if it's needed or not but im curious ;p

packages/core/src/actions/assign.ts
```diff
@@ -177,5 +177,9 @@ export function assign<
 
   assign.resolve = resolveAssign;
 
+  assign.toJSON = () => ({
+    ...assign
+  });
```
Andarist, line 182: why do we need all of those toJSONs?

packages/core/src/actions/spawnChild.ts
```diff
@@ -1,7 +1,7 @@
 import isDevelopment from '#is-development';
 import { cloneMachineSnapshot } from '../State.ts';
 import { ProcessingStatus, createActor } from '../createActor.ts';
-import { executingCustomAction } from '../stateUtils.ts';
+// import { executingCustomAction } from '../createActor.ts';
```
Andarist, line 4:
```suggestion
```

packages/core/src/stateUtils.ts
```diff
@@ -1491,15 +1530,27 @@ interface BuiltinAction {
...
-export let executingCustomAction:
-  | ActionFunction<any, any, any, any, any, any, any, any, any>
-  | false = false;
+export interface ExecutableAction {
+  /**
+   * The action type
+   */
+  type: string;
+  // TODO: don't know if we need this
+  _info: ActionArgs<MachineContext, EventObject, EventObject>;
```
Andarist, line 1539: then let's remove it :P

packages/core/src/stateUtils.ts
```diff
@@ -1554,35 +1605,43 @@ function resolveAndExecuteActionsWithContext(
...
+      //   actorScope.actionExecutor({
+      //     type: typeof action === 'object' ? action.type : 'TODO',
+      //     _info: actionArgs,
+      //     params: actionParams,
+      //     execute: () => resolvedAction(actionArgs, actionParams)
+      //   });
+      //   // resolvedAction(actionArgs, actionParams);
+      // } finally {
+      //   executingCustomAction = false;
+      // }
```
Andarist, line 1633: ? :p
```suggestion
```

packages/core/src/stateUtils.ts
```diff
@@ -1554,35 +1605,43 @@ function resolveAndExecuteActionsWithContext(
...
-        });
-      }
+      executeAction();
+      // if (actorScope.self._processingStatus === ProcessingStatus.Running) {
+      //   executeAction();
+      // } else {
+      //   actorScope.defer(() => {
+      //     executeAction();
+      //   });
+      // }
```
Andarist, line 1644:
```suggestion
```

packages/core/src/types.ts
```diff
@@ -254,6 +255,23 @@ export type Action<
       TEmitted
     >;
 
+export type ActionObject<
+  TContext extends MachineContext,
+  TExpressionEvent extends EventObject,
+  TAction extends ParameterizedObject
+> = {
+  type: string;
+} & ( // this way we could iterate over `TAction` (and `TGuard` in the `Guard` type) once and not twice // TODO: consider merging `NoRequiredParams` and `WithDynamicParams` into one
```
Andarist, line 264: i'd try to recheck why this is needed for myself

packages/core/src/actions/enqueueActions.ts
Andarist, line 1: is enqueue callback resolved when `transition` gets called?

packages/core/test/transition.test.ts
```diff
@@ -0,0 +1,80 @@
+import { assign, setup, transition } from '../src';
+import { initialTransition } from '../src/transition';
+
+describe('transition function', () => {
```
Andarist, line 4: we need tests with `enqueueActions`

packages/core/src/stateUtils.ts
```diff
@@ -1491,15 +1530,27 @@ interface BuiltinAction {
...
+}
+
+export type ActionExecutor = (actionToExecute: ExecutableAction) => void;
 
 function resolveAndExecuteActionsWithContext(
   currentSnapshot: AnyMachineSnapshot,
   event: AnyEventObject,
   actorScope: AnyActorScope,
-  actions: UnknownAction[],
+  actions: UnknownActionObject[],
```
Andarist, line 1553: this type is incorrect, enqueued actions can reach this as functions :P

packages/core/src/actions/assign.ts
davidkpiano: So that these are serializable to JSON when calling `machine.definition` or `machine.toJSON()`

packages/core/src/StateNode.ts
davidkpiano: Deduped

packages/core/src/stateUtils.ts
davidkpiano: It is now correct; only action objects reach this now

packages/core/test/transition.test.ts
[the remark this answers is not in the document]
davidkpiano:
```ts
import { executeAction } from 'xstate';

// ...

executeAction(action, someActor); // or createEmptyActor()
//                    ^^^^^^^^^
```

davidkpiano:
@Andarist Final review?

packages/core/src/actions/assign.ts
Andarist: But whatever you get back from this... is not JSON, it's far from it. I truly don't understand what purpose does it serve.

packages/core/src/stateUtils.ts
```diff
@@ -50,6 +52,8 @@ import {
 } from './utils.ts';
 import { ProcessingStatus } from './createActor.ts';
 
+export const defaultActionExecutor: ActionExecutor = executeAction;
```
Andarist, line 55: this stays unused
```suggestion
```

packages/core/src/stateUtils.ts
```diff
@@ -1485,15 +1521,26 @@ interface BuiltinAction {
...
 
-export let executingCustomAction:
-  | ActionFunction<any, any, any, any, any, any, any, any, any>
-  | false = false;
+export interface ExecutableAction {
+  /** The action type */
+  type: string;
+  info: ActionArgs<MachineContext, EventObject, EventObject>;
+  params: NonReducibleUnknown;
+  function:
```
Andarist, line 1529: nit: `function` used as a property name looks weird ;p `exec`/`execute` would fit my brain better

packages/core/src/stateUtils.ts
Andarist: it's still not correct, you can easily test it by inserting a code like this here:
```ts
if (actions.some(a => typeof a === 'function') {
  throw new Error('oops')
}
```

A ton of tests will fail. The function body here checks `typeof action === 'function'` and performs specific logic related to that but TS doesn't see that it's even an acceptable solution:
```ts
// within the loop
    if (typeof action === 'function') {
      action // never
    }
```

packages/core/src/stateUtils.ts
```diff
@@ -1548,35 +1591,23 @@ function resolveAndExecuteActionsWithContext(
           : undefined;
 
     function executeAction() {
```
Andarist, line 1593: what about actions such as `sendTo`? how those are supposed to be handled with `transition`? they are not passed through this `executeAction` function

packages/core/test/getNextSnapshot.test.ts
```diff
@@ -18,10 +19,10 @@ describe('getNextSnapshot', () => {
       { count: 0 }
     );
 
-    const init = getInitialSnapshot(logic, undefined);
-    const s1 = getNextSnapshot(logic, init, { type: 'next' });
```
Andarist, line 22: this is a test file that's called `getNextSnapshot` but you are swapping *some* `getNexSnapshot` calls to `transition`. You have a different file with `transition` tests so I think this file should be left alone

packages/core/test/transition.test.ts
Andarist, line 1: I'd really like to see some tests here that would exercise the whole persists+rehydrate cycle - unless u are not aiming at those use cases at all with this functionality

packages/core/src/actions/assign.ts
davidkpiano: This is necessary for making the tests in `json.test.ts` pass

packages/core/test/transition.test.ts
davidkpiano: This is tested here:

```ts
 it('actor can be specified', () => {
      entry: (x) => {
        x.self.send({ type: 'next' });
      },
      initial: 'a',
      states: {
        a: {
          on: { next: 'b' }
        },
        b: {}
      }
    });

    const [state, actions] = initialTransition(machine);

    const actor = createActor(machine, {
      snapshot: state
    }).start();

    expect(actor.getSnapshot().matches('a')).toBeTruthy();

    actions.forEach((action) => {
      executeAction(action, actor);
    });

    expect(actor.getSnapshot().matches('b')).toBeTruthy();
  });
```

packages/core/src/stateUtils.ts
davidkpiano: Oh, good call. There are some built-in actions such as `assign(…)` and `raise(…)` that immediately affect the next `state`, but `sendTo(…)` may affect it later if it's a delayed send.

Do you think we should include built-in actions anyway in `actions`? What should they do if the `state` already represents the action taken by them?

packages/core/src/stateUtils.ts
davidkpiano: Renamed to `exec`

packages/core/test/transition.test.ts
Andarist: What's the benefit of this here if `actions` are immediately executed in the same scope? Why would anyone want to do it?

davidkpiano: I'll make a separate test that shows actions being executed in a different scope

Jamscott:
Is there any update on the status of this, as I really need this with the deprecation of starting interpreter at a state on server side machines

davidkpiano:
> Is there any update on the status of this, as I really need this with the deprecation of starting interpreter at a state on server side machines

We're planning on merging it soon (cc. @Andarist )

packages/core/test/transition.test.ts
davidkpiano: Actually, no extra test is needed - it is not in the same scope. The snapshot and actions don't come from any scope at all.

Andarist: by scope i meant here a different execution context, literally a scenario like:
1. call `transition`
2. send the result over the wire or serialize it somehow or something
3. execute things "on the other side"

I don't understand the purpose of `executeAction` (beyond some niche use cases) in scenarios when everything happens in the same execution context.

davidkpiano: > I don't understand the purpose of `executeAction` (beyond some niche use cases) in scenarios when everything happens in the same execution context.

1. This replaces `action.execute()`, like we talked about previously
2. In server-side workflow scenarios, it's possible that an action needs to be sent to a task queue to be executed later, where it will be executed e.g. in a completely separate serverless function.

In other words, for serverless workflows, or workflows that defer tasks to a task queue, effects probably will not run in the same execution context.

Do you want me to write a test demonstrating the 2nd case?

packages/core/src/stateUtils.ts
davidkpiano: The reason this is `UnknownActionObject` is that it expects all the actions to be action objects, whether built-in (which are functions too) or not.

A more accurate check is this:

```ts
  if (actions.some((a) => !('type' in a))) {
    throw new Error('oops');
  }
```

And all tests still pass with that

packages/core/src/stateUtils.ts
davidkpiano: This is done - `sendTo`, `raise`, `cancel` built-in actions are now handled through `executeAction` as well

packages/core/test/transition.test.ts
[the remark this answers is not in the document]
davidkpiano: This is for more advanced use-cases; the `actor` provided to `executeAction` is a "handle" to e.g. the worker or lambda instance; sort of a "proxy" so that `actor.send(…)` will do the domain-specific action

packages/core/test/invoke.test.ts
```diff
@@ -3357,10 +3358,13 @@ describe('invoke', () => {
...
+    actorRef.system.inspect((e) => {
+      e;
+    });
     actorRef.send({ type: 'NEXT' });
-    await sleep(3);
-    expect(actorRef.getSnapshot().status).toBe('done');
-  });
+
+    await waitFor(actorRef, (s) => s.status === 'done');
+  }, 100000);
```
Andarist, line 3367: I'm pretty sure you have some debugging leftovers here, changes to this file should be reverted

packages/core/src/actions/cancel.ts
```diff
@@ -35,7 +35,7 @@ function resolveCancel(
 
 function executeCancel(actorScope: AnyActorScope, resolvedSendId: string) {
   actorScope.defer(() => {
-    actorScope.system.scheduler.cancel(actorScope.self, resolvedSendId);
+    // actorScope.system.scheduler.cancel(actorScope.self, resolvedSendId);
```
with-heart, line 38: Should `executeCancel` do something?

packages/core/src/actions/raise.ts
```diff
@@ -87,10 +87,10 @@ function executeRaise(
...
   const { event, delay, id } = params;
   if (typeof delay === 'number') {
-    actorScope.defer(() => {
-      const self = actorScope.self;
-      actorScope.system.scheduler.schedule(self, self, event, delay, id);
-    });
+    // actorScope.defer(() => {
+    //   const self = actorScope.self;
+    //   actorScope.system.scheduler.schedule(self, self, event, delay, id);
+    // });
```
with-heart, line 93: Same thing here with `executeRaise`

packages/core/src/StateNode.ts
```diff
@@ -296,21 +307,25 @@ export class StateNode<
...
         const { src, systemId } = invokeConfig;
         const resolvedId = invokeConfig.id ?? createInvokeId(this.id, i);
-        const resolvedSrc =
+        const sourceName =
           typeof src === 'string'
             ? src
             : `xstate.invoke.${createInvokeId(this.id, i)}`;
+        if (typeof src !== 'string') {
+          this.machine.implementations.actors[sourceName] = src;
+        }
```
Andarist, line 316: This leaks in cases like this:
```ts
const sharedActors = {}

setup({ actors: sharedActors }).createMachine({
    invoke: {
        src: fromPromise(async () => '')
    }
})

setup({ actors: sharedActors }).createMachine({
  invoke: {
    src: fromPromise(async () => 100),
  },
});
```

We should add a test case for this

packages/core/src/actions/cancel.ts
```diff
@@ -33,9 +33,7 @@ function resolveCancel(
 }
 
 function executeCancel(actorScope: AnyActorScope, resolvedSendId: string) {
-  actorScope.defer(() => {
-    actorScope.system.scheduler.cancel(actorScope.self, resolvedSendId);
-  });
+  actorScope.system.scheduler.cancel(actorScope.self, resolvedSendId);
```
Andarist, line 36: do we have some new test that shows this is a desired change?

packages/core/src/actions/enqueueActions.ts
```diff
@@ -135,7 +136,16 @@ function resolveEnqueueActions(
   const enqueue: Parameters<typeof collect>[0]['enqueue'] = function enqueue(
     action
   ) {
-    actions.push(action);
+    actions.push(
+      convertAction(
+        action as any,
+        snapshot.machine.root,
+        'enqueue' + Math.random(), // TODO: this should come from state node ID which isn't provided
```
Andarist, line 143: how hard would it be to fix it here?

packages/core/src/actions/raise.ts
```diff
@@ -75,25 +76,27 @@ function resolveRaise(
...
     internalQueue.push(resolvedEvent);
   }
-  return [snapshot, { event: resolvedEvent, id, delay: resolvedDelay }];
+  return [
+    snapshot,
+    {
+      event: resolvedEvent,
+      id,
+      delay: resolvedDelay,
+      startedAt: resolvedDelay === undefined ? undefined : Date.now()
```
Andarist, line 85: conceptually this is rather incorrect, if anything `startedAt` is the execution's responsibility, not the resolving

packages/core/src/actions/raise.ts
```diff
@@ -162,9 +165,24 @@ export function raise<
   raise.event = eventOrExpr;
   raise.id = options?.id;
   raise.delay = options?.delay;
+  raise.startedAt = options?.delay === undefined ? undefined : Date.now();
```
Andarist, line 168:
```suggestion
```

packages/core/test/transition.test.ts
```diff
@@ -0,0 +1,479 @@
...
+    }).start();
+
+    actions.forEach((action) => {
+      machine.executeAction(action, actor);
+    });
+
+    await waitFor(actor, (s) => s.matches('b'));
+  });
+
+  it('Delayed transitions should be returned', async () => {
```
Andarist, line 184:
```suggestion
  it('raise actions related to delayed transitions should be returned', async () => {
```

packages/core/src/types.ts
```diff
@@ -2604,3 +2623,60 @@ export type ToStateValue<T extends StateSchema> = T extends {
...
+    infer _TTag,
+    infer _TInput,
+    infer _TOutput,
+    infer _TEmitted,
+    infer _TMeta,
+    infer _TConfig
+  >
+    ?
+        | SpecialExecutableAction
+        | (string extends TAction['type'] ? never : ToExecutableAction<TAction>)
```
Andarist, line 2676:
```suggestion
        | SpecialExecutableAction
        | ToExecutableAction<
            string extends TAction['type']
              ? { type: string & {}; params: NonReducibleUnknown }
              : TAction
          >
```

Andarist: we should add type tests for both setup and bare createMachine

packages/core/src/actions/cancel.ts
davidkpiano: There is no reason to defer it - it's a clean-up

packages/core/src/stateUtils.ts
```diff
@@ -1586,6 +1604,15 @@ function resolveAndExecuteActionsWithContext(
...
+        {
+          type: builtinAction.type,
+          info: actionArgs,
+          params,
+          exec: () => {} // noop
+        },
+        actorScope.self
+      );
       if (actorScope.self._processingStatus === ProcessingStatus.Running) {
         builtinAction.execute(actorScope, params);
```
Andarist, line 1617: what happens here is beyond convoluted:

1. a `noop` exec gets passed into `actorScope.actionExecutor`
2. then that gets mutated internally there for some actions (raise and sendTo)
3. but then we can also learn that `actorScope.actionExecutor` isn't entirely responsible for the actions' execution because we continue to call `builtinAction.execute` **here**

It's very confusing and hard to reason about as the logic is spread across those 2 places weirdly

packages/core/src/system.ts
```diff
@@ -125,7 +131,18 @@ export function createSystem<T extends ActorSystemInfo>(
...
+        const resolvedTarget =
+          typeof target === 'string'
+            ? (source.getSnapshot() as AnyMachineSnapshot).children[target]
+            : target;
+
+        if (!resolvedTarget) {
+          throw new Error(
+            `Actor with id ${typeof target === 'string' ? target : target.sessionId} not found in the system.`
+          );
+        }
```
Andarist, line 143: IMHO this doesn't fit here. It's forcefully trying to make something pass without considering what area of the code was designed to deal with this kind of resolving. `Scheduler` works with actor refs so even assuming here that the source is a machine actor is already a big stretch. The target should always come pre-resolved to the Scheduler as it was before.

packages/core/test/actions.test.ts
```diff
@@ -3178,6 +3178,49 @@ describe('sendTo', () => {
...
+
+    expect(spy).toMatchMockCallsInlineSnapshot(`
+[
+  [
+    {
+      "counter": 1,
+    },
+  ],
+]
+`);
```
Andarist, line 3222: I just pushed out this test case that shows, what I believe to be, a regression. Note that I pushed out 2 test cases here: https://github.com/statelyai/xstate/pull/4954/commits/ece423d68c449eeb0ae5dee8392e4e10a989ce63

They both are identical, the only difference is that one uses `raise(...)` whereas the other one uses `sendTo(({ self }) => self, ...)`. While the raise one doesn't make a lot of sense, it clearly shows that something is broken here because those 2 tests are exactly the same if we consider `raise` to be a sugar for the other one.

packages/core/test/actions.test.ts
```diff
@@ -3753,51 +3753,6 @@ describe('cancel', () => {
     await sleep(10);
     expect(spy.mock.calls.length).toBe(0);
   });
-
-  it('should not be able to cancel a just scheduled non-delayed event to a just invoked child', async () => {
```
Andarist: I don't understand why this one was removed. What's wrong with this test? This specifically tests that an invalid usage of `cancel` does nothing. It's not like this is testing that `cancel` can cancel a non-delayed `send` - it's testing the opposite. A "negative" test case like this still serves a purpose.

davidkpiano: The `cancel(…)` action is only for delayed events: 

> The <cancel> element is used to cancel a delayed <send> event. […]
>
> https://www.w3.org/TR/scxml/#cancel

davidkpiano: Sorry, we can add this one back in. I misinterpreted it.

packages/core/src/stateUtils.ts
Andarist: This is a different test and it tests a different thing. I'm saying that this type claims that functions can't be received by this function but they totally can be - and the throw-based snippet I proposed proves that they can be received here at runtime. That, in turn, means that this type annotation is not correct - or that parts of the implementation were not adjusted to make this annotation correct

packages/core/src/index.ts
```diff
@@ -8,6 +8,7 @@ export {
   type Interpreter,
   type RequiredActorOptionsKeys as RequiredActorOptionsKeys
 } from './createActor.ts';
+// TODO: decide from where those should be exported
```
Andarist, line 11: what this comment is about? i somewhat doubt it's about `createMachine` 😉

packages/core/src/actions/enqueueActions.ts
Andarist: 👋

packages/core/src/actions/spawnChild.ts
```diff
@@ -103,6 +107,8 @@ function executeSpawn(
   }
 
   actorScope.defer(() => {
+    actorRef._parent = actorScope.self;
+    actorRef.system = actorScope.system;
```
Andarist, line 111: tbh, this looks like a massive hack

Andarist: There is also no test in the codebase that would prove this is needed. I can see how it might be needed for something but:
- we should have a test for it
- this should be resolved at some other time and not in this `execute` handler for the `spawnChild` action

packages/core/src/actions/enqueueActions.ts
davidkpiano: If you do the refactor where we don't convert actions eagerly, should we try to fix this still?

Andarist: ye, this becomes obsolete when we don't convert actions eagerly

packages/core/src/stateUtils.ts
```diff
@@ -1823,3 +1817,51 @@ export function resolveStateValue(
...
+    }
+
+    action.exec?.(
+      {
+        ...action.info,
+        self: actor,
+        system: actor.system
+      },
+      action.params
+    );
```
Andarist, line 1863: @davidkpiano this is missing action lookup in `machine.implementations.actions`. You can't rely on `.exec` being available if the goal is to have serializable actions. In fact, I'd probably just forbid combining inline actions with `machine.executeAction`. It should be an invariant that to use this API all actions must be declared explicitly in `machine.implementations.actions`. It's impossible to resolve an enqueued inline action, they are not discoverable in the static config at all. So it would be best to forbid all inline actions to make the rules clear around this.

davidkpiano: We fixed this, right?

packages/core/src/stateUtils.ts
davidkpiano: This has been reverted

packages/core/src/actions/send.ts
```diff
@@ -124,6 +124,7 @@ function resolveSendTo(
     snapshot,
     {
       to: targetActorRef,
+      actorId: typeof resolvedTarget === 'string' ? resolvedTarget : undefined,
```
Andarist, line 127: we could name this `targetId` to slightly disambiguate it (with `actorId` it's not immediately obvious if it contains the source's id or target's id)

packages/core/test/transition.test.ts
```diff
@@ -20,6 +21,7 @@ import { createDoneActorEvent } from '../src/eventUtils';
 import { initialTransition } from '../src/transition';
 import assert from 'node:assert';
 import { resolveReferencedActor } from '../src/utils';
+import { parseArgs } from 'node:util';
```
Andarist, line 24: this should be removed

packages/core/src/StateMachine.ts
```diff
@@ -629,4 +637,57 @@ export class StateMachine<
...
+        case 'xstate.raise':
+          if (typeof (action as any).params.delay !== 'number') {
+            return;
+          }
+          executeRaise(actorScope, action.params as any);
+          return;
+        case 'xstate.sendTo':
+          executeSendTo(actorScope, action.params as any);
+          return;
+      }
```
Andarist, line 677: I believe this switch/case misses some of the other actions, like `emit`, `log`. It's also still not completely obvious to me how the user should deal with `spawnChild` and `stopChild` - it would be great if we'd have some tests showing how to interact with those.

Andarist (approved):

PR #4961, opened 2024-07-04
Update TypeScript to 5.5.3

davidkpiano:


package.json
```diff
@@ -78,9 +78,9 @@
     "prettier": "^3.1.0",
     "spawn-command": "0.0.2-1",
     "synckit": "^0.8.5",
-    "tslib": "^2.3.1",
-    "tslint": "^5.11.0",
-    "typescript": "^5.4.2",
+    "tslib": "^2.6.3",
+    "tslint": "^6.1.3",
+    "typescript": "^5.5.3",
```
Andarist, line 83: I'd hold off for a little bit longer. I know about some declaration emit bugs that crept into the 5.5 release. They probably won't affect us but we are not in a hurry here and it's safer to wait.

PR #4962, opened 2024-07-07
run prettier on all files

with-heart:
This PR runs `prettier` on all files in the project (`yarn prettier --write --ignore-unknown .`) so that PRs will no longer include accidental unrelated formatting changes.

It also creates `.prettierignore` and adds `pnpm-lock.yaml` to it as this file is used in some examples projects and shouldn't be formatted.


docs/zh/recipes/svelte.md
```diff
@@ -68,14 +70,15 @@ The toggleService has a `.subscribe` function that is similar to Svelte stores,
...
 
-    const toggleService = interpret(toggleMachine).start();
+  const toggleService = interpret(toggleMachine).start();
 </script>
 
-<button on:click={() => toggleService.send({ type: 'TOGGLE' })}>
-    {$toggleService.matches('inactive') ? 'Off' : 'On'}
+<button on:click="{()" ="">
+  toggleService.send({ type: 'TOGGLE' })}> {$toggleService.matches('inactive') ?
+  'Off' : 'On'}
```
davidkpiano, line 81: Whoa, what happened here?

with-heart: Looks like an issue with the `html` tag on those code blocks. There's [a plugin for `svelte`](https://github.com/sveltejs/prettier-plugin-svelte) but I think it'd be easier to just add `docs` to `.prettierignore` since it's not really updated anymore. What do you think?

Andarist (approved):

docs/zh/recipes/svelte.md
davidkpiano: Yes, can you please do that?

templates/vue-ts/src/Feedback.vue
```diff
@@ -36,29 +32,37 @@ const { snapshot, send } = useMachine(feedbackMachine, {
...
       <h2>What can we do better?</h2>
-      <textarea name="feedback" rows={4} placeholder="So many things..." @input="(ev) => {
-        send({
-          type: 'feedback.update',
-          value: (ev.target as HTMLTextAreaElement).value
-        })
-      }" />
+      <textarea
+        name="feedback"
+        rows="{4}"
```
with-heart, line 51: I don't understand why it's adding quotes around `{4}`. Also don't know Vue so I'm not sure if `rows={4}` is supposed to be valid syntax or not.

Andarist: let's revert that change if possible then and land this PR without it

davidkpiano, line 51:
```suggestion
        rows="4"
```

davidkpiano (approved):

templates/vue-ts/src/Feedback.vue
with-heart: I was getting a type error locally on that about it needing to be a number but I realized it's because it was loading react types for vue files? No idea what was going on there

davidkpiano: Might have been my fault

Andarist: > I was getting a type error locally on that about it needing to be a number but I realized it's because it was loading react types for vue files? No idea what was going on there

Vue relies on the global JSX namespace so it tends to conflict with React which still has that global namespace too (no more in the upcoming React 19!). We have a patch meant to fight it [here](https://github.com/statelyai/xstate/blob/71183da18ae6b0b6b2e8f0c52ea9976232e54f41/patches/%40vue%2Bruntime-dom%2B3.0.11.patch) but I'm not sure what's the current status of this thing. It could easily stop working at some point

PR #4969, opened 2024-07-11
use prettier-plugin-jsdoc

with-heart:
This PR installs and configures [`prettier-plugin-jsdoc`](https://github.com/hosseinmd/prettier-plugin-jsdoc):

> Prettier plugin for format comment blocks and convert to standard Match with Visual studio and other IDE which support jsdoc and comments as markdown.

> Configured with best practices of jsDoc style guides.

The biggest benefit imo is that it automatically formats jsdoc code blocks using `prettier`. But it's also nice in that it formats jsdoc comments in general so it's one less thing to have to think about when developing.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -5,5 +5,7 @@
   "semi": true,
   "singleQuote": true,
   "trailingComma": "none",
-  "bracketSpacing": true
+  "bracketSpacing": true,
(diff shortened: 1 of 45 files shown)


Andarist (approved):

davidkpiano:
Love this! I've been pulling my hair out trying to get the indentation right for JSDoc examples 💇‍♂️

PR #4974, opened 2024-07-11
Clean up files, exports, dependencies

with-heart:
This PR installs and configures [`knip`](https://knip.dev/) which I then used to clean up a number of things in the repo:

- exports
  - deleted unused types/vars that were only marked as used due to `export` keyword but aren't imported/re-exported
  - removed `export` from vars/types that aren't imported/re-exported
- removed unused files
- dependencies
  - removed unused dependencies
  - removed dependencies from packages that are specified in root
  - added installed but unspecified dependencies ([phantom dependencies](https://rushjs.io/pages/advanced/phantom_deps/#phantom-dependencies)) in the appropriate places

I think it'd be cool to eventually integrate `knip` into our pipeline to be able to automate catching these things but for now doing going through them manually is a good start.

P.S.: Since stuff is changing all over the place, it might be easier to [read the changes commit-by-commit](https://github.com/statelyai/xstate/pull/4974/commits)

P.P.S.: Press <kbd>n</kbd> to go to the next commit and press <kbd>p</kbd> to go to the previous commit


packages/core/src/actors/observable.ts
```diff
@@ -35,7 +35,7 @@ export type ObservableActorLogic<
   TEmitted
 >;
 
-export type ObservableActorRef<TContext> = ActorRefFrom<
+type ObservableActorRef<TContext> = ActorRefFrom<
```
davidkpiano, line 38: Hmm, may want to make this one exported in `index.ts`

packages/core/src/actors/transition.ts
```diff
@@ -25,7 +25,7 @@ export type TransitionActorLogic<
   TEmitted
 >;
 
-export type TransitionActorRef<
+type TransitionActorRef<
```
davidkpiano, line 28: Same here; we may want this exported in `index.ts`?

packages/core/src/actors/observable.ts
with-heart: image

Caaaaaan dooooooo

I had actually gone the opposite direction and started adding all of them to `index.ts` but realized it'd be less effort to just remove them and add back the ones that are worth it.

with-heart: Resolved in #4976

packages/core/src/actors/transition.ts
with-heart: Resolved in #4976

package.json
```diff
@@ -30,7 +30,6 @@
     "typecheck": "tsc",
     "test": "jest",
     "test:core": "jest packages/core",
-    "dev": "node ./scripts/dev.js",
```
davidkpiano, line 33: cc. @Andarist

with-heart: There wasn't actually a `scripts/dev.js` file so I was thinking the file was removed at some point but the script was forgotten

davidkpiano: Ah okay!

Andarist: ye, I used this to develop v4

davidkpiano (approved):
cc. @Andarist for final approval

.knip.jsonc
```diff
@@ -0,0 +1,30 @@
...
+  "ignoreDependencies": [
+    "@tsconfig/svelte",
+    "@xstate-repo/jest-utils",
+    "@xstate-repo/jest-utils/setup",
+    "babel-preset-solid",
+    "husky",
+    "lint-staged",
+    "svelte-jester",
+    "synckit",
+    "tslib"
```
Andarist, line 20: why this one has been ignored? do we explicitly depend on this anywhere? that might be a mistake

.knip.jsonc
Andarist, line 1: can this be added as a CI check?

scripts/typecheck-examples.js
Andarist, line 1: i think those 3 scripts were created by @davidkpiano for manual execution - I never used them but it feels like those shouldn't be removed (unless David is OK with removing them)

davidkpiano: Yes, I use these scripts

davidkpiano: @with-heart Can you add these back? 🙏

.knip.jsonc
with-heart: Ye that's what I'm planning to do. It's still reporting a few warnings that I need to figure out how to take care of, gonna do that in a follow-up PR

.knip.jsonc
```diff
@@ -6,7 +6,8 @@
     // used for `#is-development` conditional import
     "packages/**/{true,false}.ts",
     // file acts as a type test
-    "packages/xstate-svelte/test/interpreterAsReadable.svelte"
+    "packages/xstate-svelte/test/interpreterAsReadable.svelte",
+    "scripts/{typecheck-examples,update-example-*}.js"
```
davidkpiano, line 10: Maybe this or similar just in case we have other scripts?
```suggestion
    "scripts/{*}.js"
```

with-heart: I was thinking about that but if it had been in the config to start it wouldn't have detected the unused `dev.js` script so might be better to be more specific? Not sure

.knip.jsonc
with-heart: It's also in `resolutions`:

https://github.com/statelyai/xstate/blob/9841fab65f65d76349386b9489413f8fd134ee84/package.json#L89

with-heart:
The `workspaces` config I used for `examples` and `templates` was incorrect (`examples` instead of `examples/*`). Actually enabling them creates a bunch of extra work like needing to install all of their dependencies so I just removed it and will revisit later maybe

Andarist:
> @tsconfig/svelte is confusing because of packages/xstate-svelte/test/package.json - maybe just add to ignoreDependencies

I assume this might be some issue with pkg.json lookup. Such "package.json scopes" to add `type` are... weird but I've seen them used them in the wild a couple of times already. Maybe it would be worth ignoring them in Knip? This is a documented pattern in the node.js docs.

with-heart:
@webpro Thanks so much for taking a look! Glad you were able to fix two bugs while helping us clean up the config. And also glad to learn about pkg.pr.new, that's quite a useful tool.

I was thinking about posting on twitter to ask if I knew any `knip` experts that wanted to take a look but opening an issue got us _the_ expert. The power of open source ❤️

with-heart:
I believe this is good to go now 😁

package.json
```diff
@@ -46,47 +45,40 @@
...
-    "gh-pages": "^2.0.1",
     "husky": "^3.1.0",
     "jest": "^29.7.0",
     "jest-config": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
-    "jest-regex-util": "^29.6.3",
+    "jest-resolve": "^29.7.0",
+    "jest-snapshot": "^29.7.0",
     "jest-watch-typeahead": "^2.2.2",
+    "knip": "https://pkg.pr.new/knip@bd16c98",
```
with-heart, line 73: I subscribed to `knip` release notifications so I can switch this to an actual version once it's published

Andarist: I'd prefer to block this PR on that release. I had unpleasant experiences in the past with such alternative registered - mainly CodeSandbox CI just decided to forget a package that was published using that service.

webpro:
> > @tsconfig/svelte is confusing because of packages/xstate-svelte/test/package.json - maybe just add to ignoreDependencies
> 
> I assume this might be some issue with pkg.json lookup. Such "package.json scopes" to add `type` are... weird but I've seen them used them in the wild a couple of times already. Maybe it would be worth ignoring them in Knip? This is a documented pattern in the node.js docs.

Sorry, not sure exactly what you are referring to here. Knip follows the root `package.json#workspaces` to find workspaces (extra workspaces can be configured, though). If we ignore `packages/xstate-svelte/test/package.json` in this regard, then the easiest path is to configure `test/tsconfig.json` as the path of the TypeScript config for this workspace:

```jsonc
{
  "workspaces": {
    "packages/xstate-svelte": {
      "typescript": "test/tsconfig.json",
      // shorthand for "typescript": { "config": ["test/tsconfig.json"] }
    }
  }
}
```

package.json
webpro: Absolutely.

with-heart: > I'd prefer to block this PR on that release. I had unpleasant experiences in the past with such alternative registered - mainly CodeSandbox CI just decided to forget a package that was published using that service.

Damn you, Codesandbox! ✊☁️

with-heart:
Since `knip@5.27.0` was released this morning, I've replaced `pkg.pr.new` with `^5.27.0`. Running it results in no errors so we should be good to go! 🤞

webpro:
Thanks @with-heart! ➕❤️ 

You could also use the config from [my latest comment](https://github.com/statelyai/xstate/pull/4974#issuecomment-2241596411) to get rid of of the ignored `@tsconfig/svelte` issue, but it's minor.

Andarist:
I guess I just don't understand why this has to be declared as an extra workspace in the first place - couldn't it be picked up automatically? It's fine as a solution right now - I just wonder how Knip's magic sauce is made and if there is any room for improvements on that front

webpro:
> I guess I just don't understand why this has to be declared as an extra workspace in the first place - couldn't it be picked up automatically?

There's no magic here, Knip just uses `package.json#workspaces` - which does not include `packages/xstate-svelte/test`.

Andarist:
Right, because it's not a workspace. But including it in Knip's workspaces it feels like we dilute the meaning of a workspace. Multiple `tsconfig.json`s (and `package.json`s for that matter) are fair game within a single workspace. So I just wonder if they couldn't be discovered automatically? It would require less work for the user.

webpro:
> Right, because it's not a workspace. But including it in Knip's workspaces it feels like we dilute the meaning of a workspace.

We're not including it as a workspace, it's only Knip's TypeScript plugin that pulls in `test/tsconfig.json` instead of `./tsconfig.json` so it can extract dependencies from that file. There's nothing more to it, it's not like we're defining `./test` as a TS project or a (npm/Yarn) workspace.

> Multiple `tsconfig.json`s (and `package.json`s for that matter) are fair game within a single workspace. So I just wonder if they couldn't be discovered automatically? It would require less work for the user.

Unfortunately we're simply not there yet.

davidkpiano:
Is this one good to go on (pending merge conflicts)? @Andarist @with-heart

with-heart:
Yeap we're good to go!

Andarist (approved):

PR #4976, opened 2024-07-12
export actor logic-specific ActorRef types

with-heart:
This PR exports actor logic-specific `ActorRef` types. Noticed in #4974 that these weren't actually exported.

In addition to exporting them, I added a `jsdoc` comment block for each.

Also fixed a small issue where the `jsdoc` block for `fromPromise` was attached to a variable defined before `fromPromise` rather than `fromPromise` itself.


davidkpiano (approved):

davidkpiano:
This is awesome!

PR #4979, opened 2024-07-12
[core] Strongly-typed meta keys

davidkpiano:
This PR introduces:

- Internal types for computing state IDs from a state schema (the machine configuration)
- Strongly-typing `snapshot.getMeta()` to use those state IDs as keys


packages/core/src/types.ts
```diff
@@ -2456,9 +2458,52 @@ export type ToChildren<TActor extends ProvidedActor> =
...
+          walk: {
+            id: 'walk';
+          };
+          wait: {};
+        };
+      };
+    };
+  }>
+>;
+
```
with-heart, line 2506:
```suggestion
```

with-heart (changes requested):
Nice lil addition!

packages/core/src/setup.ts
```diff
@@ -216,7 +216,8 @@ export function setup<
     TInput,
     TOutput,
     TEmitted,
-    TMeta
+    TMeta,
+    TConfig
```
davidkpiano, line 220: @Andarist I included `TConfig` because providing just `ToStateValue<TConfig>` loses too much information. I think I prefer passing the `TConfig` (as a `StateSchema`) instead of reducing the types in a lossy way. WDYT?

packages/core/test/setup.types.test.ts
```diff
@@ -2184,9 +2184,48 @@ describe('setup()', () => {
 
     const actor = createActor(machine);
 
-    actor.getSnapshot().getMeta().a satisfies { layout: string } | undefined;
+    actor.getSnapshot().getMeta()['(machine)'] satisfies
```
Andarist, line 2187:
```suggestion
    actor.getSnapshot().getMeta()['(machine).a'] satisfies
```

packages/core/test/setup.types.test.ts
```diff
@@ -2184,9 +2184,48 @@ describe('setup()', () => {
...
+      | { layout: string }
+      | undefined;
 
     // @ts-expect-error
     actor.getSnapshot().getMeta().a?.whatever;
   });
+
+  it('should strongly type the state IDs in snapshot.getMeta()', () => {
```
Andarist, line 2195: this PR introduces very similar tests in 2 places - are there any different? It seems like keeping this in a single place would be better

packages/core/src/State.ts
```diff
@@ -53,7 +55,8 @@ interface MachineSnapshotBase<
   TTag extends string,
   TOutput,
   TMeta,
-  _TUnusedButLeftForCompatReasons = never
+  _TUnusedButLeftForCompatReasons = never,
```
Andarist, line 58: it's really hard to judge those things but it's probably fine to remove this one now

packages/core/src/setup.ts
Andarist: Conceptually it would be better to pass minimal required information down here instead of the whole `TConfig`. That said, with this change, it becomes even less likely for one to create 2 different state machines that could be used interchangeably... so 🤷‍♂️

packages/core/test/setup.types.test.ts
```diff
@@ -2184,9 +2184,48 @@ describe('setup()', () => {
 
     const actor = createActor(machine);
 
-    actor.getSnapshot().getMeta().a satisfies { layout: string } | undefined;
+    actor.getSnapshot().getMeta()['(machine).a'] satisfies
```
Andarist, line 2187: it looks like this revealed some issue with the `StateId` type

davidkpiano: See [520a829](https://github.com/statelyai/xstate/pull/4979/commits/520a829fb6c8180cc9b82a8015216886dd00c2d2)

packages/core/src/State.ts
davidkpiano: [5a3404d](https://github.com/statelyai/xstate/pull/4979/commits/5a3404dc0e7250e0a9b861287235121fe45a5f95)

packages/core/test/setup.types.test.ts
```diff
@@ -2184,9 +2184,48 @@ describe('setup()', () => {
...
+      | { layout: string }
+      | undefined;
 
     // @ts-expect-error
     actor.getSnapshot().getMeta().a?.whatever;
   });
+
+  it('should strongly type the state IDs in snapshot.getMeta()', () => {
+    const machine = setup({}).createMachine({
+      id: 'root',
```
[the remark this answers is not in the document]
davidkpiano: [54738b5](https://github.com/statelyai/xstate/pull/4979/commits/54738b585180cf3fee7128b0674d1f49ce769199)

packages/core/test/setup.types.test.ts
Andarist: And that's why we should not write huge type tests like "should support typing meta properties". Every "assertion" contained in it should be a separate test to avoid issues like this

davidkpiano: I'll separate this into multiple tests

davidkpiano: Separated: [e337836](https://github.com/statelyai/xstate/pull/4979/commits/e3378365c6d04fe2ea7b40520320b741da7c4e8a)

packages/core/test/setup.types.test.ts
davidkpiano: Consolidated here: [e79ad4e](https://github.com/statelyai/xstate/pull/4979/commits/e79ad4e223053310e9d63e9192198cf19df7b979)

Andarist (approved):

PR #4983, opened 2024-07-15
remove unmaintained v4 docs

with-heart:
This PR removes the unmaintained v4 docs and the associated `docs` workflow.

As the `concepts.md` page notes, these docs are no longer maintained. So let's make that a `true` statement 😁


davidkpiano (changes requested):
We'd like to still keep the v4 documentation up for developers still using XState v4 and ideally not break any links 🤔

with-heart:
Do you happen to have usage data from Plausible for https://xstate.js.org/docs/? That's what the `docs/` folder deploys to right?

I'm wondering if maybe it'd make more sense to just redirect that to https://stately.ai/docs/xstate-v4/studio since that's the more recently-unmaintained unmaintained v4 docs.

Reasons for opening this PR:

- `docs` folder that isn't related to the version of the code it shares a codebase with is potentially confusing for users/contributors
- multiple v4 docs deployments is confusing for everyone (by which I mean me, at a minimum 😁)

davidkpiano:
Let's close this since we should still keep v4 docs for now.

PR #4984, opened 2024-07-15
add .git-blame-ignore-revs

with-heart:
This PR adds `.git-blame-ignore-revs` with commit SHAs for recent-ish `prettier` formatting commits (#4295, #4962, and #4969).

This is a `git`+GitHub feature I just learned about today that allows hiding certain commits from blame results. It's useful for hiding commits that are only related to formatting the code rather than introducing anything to the code.

GitHub displays a banner in blame view for files that are hidden and `git blame` can be configured locally to hide these commits using `git config blame.ignoreRevsFile .git-blame-ignore-revs`.


davidkpiano (approved):

Andarist (approved):

.git-blame-ignore-revs
Andarist, line 1: I tried to leverage this once in TS repo since they have a couple of commits ignored like this but they still were appearing in git blame results etc. So... I'm not sure how to make this work for myself but certainly it doesn't hurt to have it here ;p

with-heart: Yeah I'm gonna be honest, I haven't actually tried it but it sounded neat if it actually works! 🙃😁

PR #4987, opened 2024-07-17
switch package manager to pnpm

with-heart:
This PR changes our package manager from `yarn` to [`pnpm`](https://pnpm.io/).

- replaced `yarn` with `pnpm` in `CONTRIBUTING.md`
- `mv scripts/ensure-{yarn,pnpm}.js` and updated test with `pnpm/9`
- removed `patch-package` in place of `pnpm patch`
  - also removed patch for `@vue/runtime-dom@3.0.11` as we're on a much newer version that doesn't have the declared globals we had been patching
- changed `devDependencies.xstate` versions to `workspace:^`
  - upgraded `@manypkg/cli` to latest in order to support `workspace:` protocol
- GitHub Actions
  - added "Install pnpm" step to `ci-setup`
  - replaced `yarn` with `pnpm` in actions/workflows (except `docs` workflow)


Andarist:
> changed devDependencies.xstate versions to workspace:^

don't do this if not needed - this might interfere with how Changesets are set up right now, you'd have to rework publishing scripts to enable this

with-heart:
> don't do this if not needed - this might interfere with how Changesets are set up right now, you'd have to rework publishing scripts to enable this

Oof. Okay I'll see what I can do

with-heart:
@Andarist do you mean `scripts/bump-peer-dep-ranges.js` or something else? I'm not sure where `devDependencies` using `workspace:` would affect things

Andarist:
`changeset publish` might have issues with workspace ranges. Although perhaps they don’t affect pnpm repos - that would have to be tested. It definitely doesnt work with Yarn Berry

with-heart:
I'll set up a repo and test it later just to see what happens

with-heart:
I resolved the `typecheck` errors for all of the packages except `@xstate/vue`. I don't know Vue so gonna need some help with that bit

Andarist:
I fixed the Vue issue. I think something weird is happening with Vue types - their `shallowRef` function introduces some weirdness with deferred conditional types. I prepared a slimmed-down xstate-less repro of the problem: [TS playground](https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240718#code/JYWwDg9gTgLgBAbzgdwIYwMYAsA0cBKApgGZ4DKWqANlRMkaXAM6U10NwC+cxUEIcAEQA3AK6FBAbjgB6GXBgBPMISYAuOAGYAdABZtmgEwAoY0pVwAghhjQ4AXkTG4cAOaEYZAHaowLCDAAFACUGgjOLnAYEKJeMBpeoiAARoRQkhGcGVmmACaEGFSoUIRREF5M8Kg20AwaDAA81rZQAHwZ+YXFpdEVVTVQFNS09CQaQ2yjxE0D7aZomFiB1S0MeMsDwQ6tTi4r0Nrunj5+WAEh2cEZxgvYGy0TI2tw99Bb9jvhewOHHt6+-iCV2MnGBxk6RRKZT6cBAwAAHoRcnUFMpCBBiHB9lAOAAfVEqDFYgaPdgka63JZwxHIkjrbHvT4RbG-Y4As5Ay7XYixGzAco8LwNAAqcEI8JghC8uSYVlmyygrg0wq2X2hlWJqxIDkQ3FQssawrmLl6GuxpKmOoQetlFsNcwilNeOLpLwZ212kRZR3+p3OwJcoOuLid5tYT1dzsZnu+LVZvsBF0yYJN5TNJPDZOIRCYoio8EcLGGWYVrhTKHQdzDxamObzMHpmw9apcckiAD0APwRWMHH0nRMByJtrvJ4Pq+DUpEcRzWrGy8zozHYvEEpea6B2kh1-Pjp1T2mMKPNnuyeRwUeRDdQeMDjlJq8j7uB4GcYxAA)

packages/xstate-svelte/test/package.json
```diff
@@ -1,3 +1,6 @@
 {
-  "type": "module"
+  "type": "module",
+  "devDependencies": {
+    "@xstate/svelte": "workspace:^"
+  }
```
with-heart, line 5: This seems fine? Not really sure

with-heart:
@Andarist I found references to the `workflow:` protocol in `changesets/changesets` in both `CHANGELOG.md` files and the codebase. 

I also created a test project locally and verified that `changeset version` has no issues correctly handling versioning when using the protocol, including bumping the version of packages that depend on workspace packages when that dependency's version is bumped.

I think we should be good to go assuming everything else checks out

davidkpiano (approved):

davidkpiano (changes requested):
I got this error when trying this out locally:
```
(base) davidkpiano@davids-mbp-3 xstate % pnpm i
Scope: all 12 workspace projects
 WARN  Ignoring broken lockfile at /Users/davidkpiano/Code/xstate: Lockfile /Users/davidkpiano/Code/xstate/pnpm-lock.yaml not compatible with current pnpm
 WARN  deprecated @babel/plugin-proposal-class-properties@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-class-properties instead.
packages/xstate-solid                    |  WARN  deprecated solid-testing-library@0.3.0
 ERR_PNPM_PATCH_NOT_APPLIED  The following patches were not applied: solid-js@1.8.18

Either remove them from "patchedDependencies" or update them to match packages in your dependencies.
Progress: resolved 1181, reused 1181, downloaded 0, added 0
(base) davidkpiano@davids-mbp-3 xstate % pnpm test

> xstate-monorepo@ test /Users/davidkpiano/Code/xstate
> jest

sh: jest: command not found
 ELIFECYCLE  Test failed. See above for more details.
(base) davidkpiano@davids-mbp-3 xstate % 
```

with-heart:
@davidkpiano what pnpm version are you on? The lockfile format changed in v8 I think so you might be on an older version

davidkpiano:
> @davidkpiano what pnpm version are you on? The lockfile format changed in v8 I think so you might be on an older version

I might have been on an older version.

Can you fix the conflicts? 🙏

packages/xstate-solid/package.json
```diff
@@ -53,6 +53,6 @@
   "devDependencies": {
     "solid-js": "^1.7.6",
     "solid-testing-library": "^0.3.0",
-    "xstate": "5.17.1"
+    "xstate": "workspace:^"
```
with-heart, line 56: I've updated all `xstate` versions to use the `workspace:` protocol.

This should reduce the number of merge conflicts we need to resolve due to bumped versions.

.github/workflows/release.yml
```diff
@@ -30,8 +31,8 @@ jobs:
...
         uses: changesets/action@v1
         with:
-          publish: yarn run release
-          version: yarn run version
+          publish: pnpm -r publish --access=public
+          version: pnpm run version
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
with-heart, line 38: In order to make the `workspace:` protocol work with publishing, I replaced `changeset publish` with `pnpm -r publish`. `pnpm publish` uses `NODE_AUTH_TOKEN` so I also renamed the `env` var.

`pnpm publish` replaces `workspace:` references with the specific version of the package in the workspace using the semver range specified by the workspace.

For example:

```json
{
	"dependencies": {
		"foo": "workspace:*",
		"bar": "workspace:~",
		"qar": "workspace:^",
		"zoo": "workspace:^1.5.0"
	}
}
```

is transformed into:

```json
{
	"dependencies": {
		"foo": "1.5.0",
		"bar": "~1.5.0",
		"qar": "^1.5.0",
		"zoo": "^1.5.0"
	}
}
```

`pnpm publish` also publishes packages using a dependency graph which guarantees that packages are published in the correct order.

To validate that `pnpm publish` works correctly with `changesets`, I created a repo with dummy packages: https://github.com/with-heart/changesets-pnpm-workspace-protocol-test

[`@pnpm/pnpm`](https://github.com/pnpm/pnpm) itself also uses `changesets` with `pnpm publish`.

package.json
Andarist, line 1: I'd appreciate if you could add a `packageManager` field here. To get it right u can generate it using `corepack`

with-heart: That's funny. I was going to add a `packageManager` field in a separate PR a few weeks ago, but then after doing some more research into `corepack` I changed my mind and didn't think it was a good idea, but now I can't remember why I changed my mind 😆

Will do!

.github/workflows/release.yml
with-heart: Good call! Played around with it for a bit to verify that `changeset tag` works. Lemme update our script

package.json
with-heart: Interesting side effect here: `pnpm/action-setup` checks to see if both `with.version` and `packageManager` are specified. If they are it fails the action with a message noting that you should use one or the other but not both to avoid version mismatch errors.

.github/workflows/release.yml
with-heart: You can see it working on my test repo: https://github.com/with-heart/changesets-pnpm-workspace-protocol-test

with-heart:
Is there anything else I need to do to move forward with this?

packages/xstate-svelte/test/package.json
Andarist: Interesting, this is weird and it doesn't make sense (to me :P). This file doesn't define dependencies at all. On principle, I'd like to have this removed from here... but also it's probably not worth putting the effort into fixing this 🤷‍♂️

pnpm-workspace.yaml
```diff
@@ -0,0 +1,4 @@
+packages:
+  - 'packages/*'
+  - 'packages/xstate-svelte/test'
```
Andarist, line 3: I guess this might be an accompanying change to https://github.com/statelyai/xstate/pull/4987/files#r1683562667 but I similarly don't think this should be a workspace. 

Could you confirm that this new setup doesn't work after removing this line and the change from that other `package.json`?

package.json
```diff
@@ -23,16 +23,16 @@
...
+    "postinstall": "manypkg check && preconstruct dev",
     "build": "preconstruct build",
     "fix": "manypkg fix",
     "typecheck": "tsc",
     "test": "jest",
     "test:core": "jest packages/core",
     "changeset": "changeset",
-    "release": "changeset publish",
-    "version": "changeset version && node ./scripts/bump-peer-dep-ranges.js"
+    "version": "changeset version && node ./scripts/bump-peer-dep-ranges.js",
```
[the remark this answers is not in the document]
with-heart: Good catch!

pnpm-workspace.yaml
with-heart: I changed the import in `UseActor.svelte` and everything is fine now:

https://github.com/statelyai/xstate/pull/4987/files#diff-a89e578c4c6140c646c35fd9f9b0f217b614fd78861ae130c3326300058c756aL4

with-heart: I removed `xstate-svelte/test/package.json` and set `verbatimModuleSyntax: false` in `xstate-svelte/test/tsconfig.json`. That allows `svelte-check` to pass. Not sure if that's fine or not.

package.json
```diff
@@ -90,5 +88,14 @@
...
+  },
+  "pnpm": {
+    "patchedDependencies": {
+      "@vue/test-utils@2.4.6": "patches/@vue__test-utils@2.4.6.patch",
+      "jest-snapshot@29.7.0": "patches/jest-snapshot@29.7.0.patch",
+      "solid-js@1.8.18": "patches/solid-js@1.8.18.patch",
+      "solid-testing-library@0.3.0": "patches/solid-testing-library@0.3.0.patch"
+    }
+  },
+  "packageManager": "pnpm@9.7.1+sha512.faf344af2d6ca65c4c5c8c2224ea77a81a5e8859cbc4e06b1511ddce2f0151512431dd19e6aff31f2c6a8f5f2aced9bd2273e1fed7dd4de1868984059d2c4247"
```
davidkpiano, line 100: What's this?

Andarist: it's a field describing the exact version of the package manager that should be used, tools like `corepack` read this and use that exact version to run the commands etc

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  - 'packages/*'
+  - 'scripts/jest-utils'
```
Andarist, line 3:
```suggestion
  - 'scripts/*'
```

Issue #5001, opened 2024-07-28
Bug: clearTimeout being called with argument 'undefined'


arlyon:
### XState version

XState version 5

### Description

I am trying to integrate temporal and xstate 5 which I have working bar a few minor roadblocks. Temporal overrides clearTimeout which calls into their runtime however it crashes because xstate has a habit of calling clearTimeout with `undefined`. Temporal then tries to find a timer with id 0 (presumably coercing undefined across the FFI boundary) and erroring out because it doesn't exist. A workaround is to override this:

```ts
const clearTimeoutInner = global.clearTimeout;

/**
 * xstate passes undefined here, which temporal doesn't like
 * instead, we intercept and log it
 *
 * @param id
 */
global.clearTimeout = (id) => {
	if (id !== undefined) {
		clearTimeoutInner(id);
	} else {
		console.log("INVALID TIMEOUT");
	}
};
```

However, this seems like a bug as I don't think clearTimeout should ever be called with no arguments. From testing, xstate does still correctly cancel all the timers as it should. I do not know the cause of the extra invalid calls. 

### Expected result

Do not call clearTimeout with undefined.

### Actual result

When timers elapse, clearTimeout is called with `undefined`.

### Reproduction

https://codesandbox.io/p/live/c4801092-33a5-4d8e-9e9f-366078f138ab

Open up the sandbox, see the console print undefined.

PR #5002, opened 2024-07-29
[core] Do not clear undefined timer IDs

davidkpiano:
Fix an issue where `clearTimeout(undefined)` was sometimes being called, which can cause errors for some clock implementations. See https://github.com/statelyai/xstate/issues/5001 for details.

### DIFF
--- a/packages/core/test/after.test.ts
+++ b/packages/core/test/after.test.ts
@@ -43,6 +43,30 @@ describe('delayed transitions', () => {
     expect(actorRef.getSnapshot().value).toBe('yellow');
   });
 
+  it('should never try to clear an undefined timeout ID', () => {
+    // https://github.com/statelyai/xstate/issues/5001
+    jest.useFakeTimers();
+
+    const actorRef = createActor(lightMachine, {
+      clock: {
+        setTimeout,
+        clearTimeout(id) {
+          if (id === undefined) {
+            // in some workflow implementations, clearing an undefined ID will error
+            throw new Error('undefined ID');
+          }
+        }
+      }
+    }).start();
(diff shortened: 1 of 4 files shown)


packages/core/test/after.test.ts
```diff
@@ -43,6 +43,30 @@ describe('delayed transitions', () => {
     expect(actorRef.getSnapshot().value).toBe('yellow');
   });
 
+  it('should never try to clear an undefined timeout ID', () => {
```
Andarist, line 46: why did this one fail before the change? how did that `undefined` creep in there? 🤔

davidkpiano: There's 2 places where the ID can be deleted:

```ts
 const scheduler: Scheduler = {
    schedule: (
      // ...
    ) => {
      // ...

      const timeout = clock.setTimeout(() => {
        delete timerMap[scheduledEventId]; // here
        delete system._snapshot._scheduledEvents[scheduledEventId];

        system._relay(source, target, event);
      }, delay);

      timerMap[scheduledEventId] = timeout;
    },
    cancel: (source, id: string) => {
      // ...

      delete timerMap[scheduledEventId]; // and here
      delete system._snapshot._scheduledEvents[scheduledEventId];

      if (timeout !== undefined) {
        clock.clearTimeout(timeout);
      }
    },
```
If the timeout happens before `cancel(…)` happens, the ID is already deleted, hence `undefined`

Andarist: ah, ok - cause the cancellation itself is not canceled in a sense when the timeout completes. This makes sense 👍

Andarist (approved):

Issue #5008, opened 2024-07-30
Bug: 5.17.0 Property `"(machine)"` is missing in type...


wKovacs64:
### XState version

XState version 5

### Description

Hey folks. Getting a type error in xstate v5.17.0 when passing a few of my machines to `useMachine` from `@xstate/react`, but not all of them - can't quite figure out what's different enough between them to nail down the culprit. Is it obvious to you based on this error?

![image](https://github.com/user-attachments/assets/ccc0cd59-f67e-4178-9083-58cb59ee7d12)

### Expected result

No type errors.

### Actual result

```
Argument of type 'StateMachine<CheatCodeContext, KeydownEvent, {}, never, Values<{ record: { type: "record"; params: NonReducibleUnknown; }; resetTypedKeys: { ...; }; }>, ... 8 more ..., { ...; }>' is not assignable to parameter of type 'AnyStateMachine'.
  The types returned by 'resolveState(...)' are incompatible between these types.
    Type 'MachineSnapshot<CheatCodeContext, KeydownEvent, {}, { disabled?: "idle" | "recording" | undefined; enabled?: "idle" | "recording" | undefined; }, string, NonReducibleUnknown, MetaObject, { ...; }>' is not assignable to type 'MachineSnapshot<any, any, any, any, any, any, any, StateSchema>'.
      Type 'ActiveMachineSnapshot<CheatCodeContext, KeydownEvent, {}, { disabled?: "idle" | "recording" | undefined; enabled?: "idle" | "recording" | undefined; }, string, NonReducibleUnknown, MetaObject, { ...; }>' is not assignable to type 'MachineSnapshot<any, any, any, any, any, any, any, StateSchema>'.
        Type 'ActiveMachineSnapshot<CheatCodeContext, KeydownEvent, {}, { disabled?: "idle" | "recording" | undefined; enabled?: "idle" | "recording" | undefined; }, string, NonReducibleUnknown, MetaObject, { ...; }>' is not assignable to type 'ActiveMachineSnapshot<any, any, any, any, any, any, any, StateSchema>'.
          The types returned by 'getMeta()' are incompatible between these types.
[...]

PR #5009, opened 2024-07-31
[core] Tighten `StateMachine` types and fix issue #5008

davidkpiano:
The internal types for `StateMachine<...>` have been improved so that all type params are required, to prevent errors when using the types. This fixes weird issues like #5008.

### DIFF
--- a/packages/core/src/StateMachine.ts
+++ b/packages/core/src/StateMachine.ts
@@ -66,9 +66,9 @@ export class StateMachine<
   TTag extends string,
   TInput,
   TOutput,
-  TEmitted extends EventObject = EventObject, // TODO: remove default
-  TMeta extends MetaObject = MetaObject,
-  TConfig extends StateSchema = StateSchema
+  TEmitted extends EventObject,
+  TMeta extends MetaObject,
+  TConfig extends StateSchema
 > implements
     ActorLogic<
       MachineSnapshot<
--- a/packages/core/src/StateMachine.ts
+++ b/packages/core/src/StateMachine.ts
@@ -199,7 +199,8 @@ export class StateMachine<
     TInput,
     TOutput,
     TEmitted,
-    TMeta
+    TMeta,
+    TConfig
   > {
     const { actions, guards, actors, delays } = this.implementations;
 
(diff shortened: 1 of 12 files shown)


packages/xstate-react/test/types.test.tsx
```diff
@@ -119,3 +120,24 @@ describe('useMachine', () => {
...
 });
+
+it('useMachine types work for machines with a specified id and state with an after property #5008', () => {
+  // https://github.com/statelyai/xstate/issues/5008
+  const cheatCodeMachine = setup({}).createMachine({
+    id: 'cheatCodeMachine',
+    initial: 'disabled',
+    states: {
+      disabled: {
+        after: {}
```
davidkpiano, line 131: This is extremely weird: if I remove `after: {}`, then (without any changes to XState) this type checks without errors 😵

packages/xstate-graph/test/index.test.ts
```diff
@@ -413,33 +413,33 @@ describe('state tests', () => {
   });
 
   it('should test with input', () => {
-    const model = createTestModel(
```
davidkpiano, line 416: This was a little weird too... for some reason, TS complained if `setup(…)` was inside `createTestModel(setup(…).createMachine(…))` but not if I moved it outside.

packages/core/src/types.ts
```diff
@@ -2443,29 +2460,35 @@ export type StateSchema = {
   states?: Record<string, StateSchema>;
 };
 
+type Test = SnapshotFrom<AnyStateMachine>;
+
+['getMeta'];
+
```
davidkpiano, line 2466:
```suggestion
```

packages/core/src/StateMachine.ts
```diff
@@ -66,9 +66,9 @@ export class StateMachine<
   TTag extends string,
   TInput,
   TOutput,
-  TEmitted extends EventObject = EventObject, // TODO: remove default
-  TMeta extends MetaObject = MetaObject,
-  TConfig extends StateSchema = StateSchema
+  TEmitted extends EventObject,
+  TMeta extends MetaObject,
+  TConfig extends StateSchema
```
davidkpiano, line 71: This is the main change. By having these as optional, we were (well, at least I was) miscounting the generic type params, and some were erroneously using their default types.

PR #5011, opened 2024-07-31
[core] Fix `Actor<T>` not assignable to `ActorRef<T>` problem

davidkpiano:


packages/core/test/types.test.ts
```diff
@@ -4580,3 +4585,25 @@ it('UnknownActorRef should return a Snapshot-typed value from getSnapshot()', ()
...
+  const actor = createActor(logic);
+  actor satisfies ActorRefFrom<typeof logic>;
+
+  class ActorThing<T extends AnyActorLogic> {
+    // actorRef: ActorRefFrom<T>;
+    actorRef: ActorRefFrom<T>;
+    constructor(actorLogic: T) {
+      const actor = createActor(actorLogic);
+
+      // But it doesn't work here...
```
davidkpiano, line 4602: cc. @Andarist Here is the weird part

davidkpiano:
@Andarist Is this good to go?

packages/core/src/types.ts
```diff
@@ -2006,50 +2006,19 @@ export type ActorLogicFrom<T> = ReturnTypeOrValue<T> extends infer R
       : never
   : never;
 
+// TODO: in v6, this should only accept AnyActorLogic, like ActorRefFromLogic
 export type ActorRefFrom<T> = ReturnTypeOrValue<T> extends infer R
```
Andarist, line 2010: technically this change was a breaking change, support for multiple input types has been removed from this helper

davidkpiano: Should I adjust it to "minor" then? https://github.com/statelyai/xstate?tab=readme-ov-file#typescript-changes

It's very unlikely that anyone was using `ActorRefFrom<typeof someFunction>`, and if they are, it's easy enough to do `ActorRefFrom<ReturnType<typeof someFunction>>` instead.

Andarist: The `ReturnType` thing is still handled here. What was removed was support for promises and a special priority case for state machines (that could have been redundant but it's hard to tell). 

> Should I adjust it to "minor" then? https://github.com/statelyai/xstate?tab=readme-ov-file#typescript-changes

IMHO this part of the change should be reverted. It's the kind of a breaking change that doesn't really buy us anything so I think it's not worth doing it within the current major line

davidkpiano: https://github.com/statelyai/xstate/pull/5029

PR #5022, opened 2024-08-07
switch from jest to vitest

with-heart:
This PR switches our test setup from `jest` to [`vitest`](https://github.com/vitest-dev/vitest).

## How it's configured

`vitest.workspace.json` defines our workspaces as `["packages/*"]`. `vitest` looks for a `vitest.config` file in each workspace and uses it to run that workspace's tests. This allows each workspace to define the plugins (like `vite-plugin-solid` for `@xstate/solid`) and test configuration used to run that workspace's tests.

This allows us to run tests for specific workspaces using `vitest --project <package name>` (`vitest --project @xstate/react`).

## Major changes

- removed all `jest` dependencies (11❗), configs, scripts, etc.
- replaced `jsdom` with [`happy-dom`](https://github.com/capricorn86/happy-dom)
- replaced deprecated `done` callbacks with `new Promise` callbacks + `resolve`
- added `vite` plugins to some packages for out-of-the-box support in tests
- removed `@xstate-repo/jest-utils` package
  - replaced `toMatchMockCallsInlineSnapshot` with `toMatchInlineSnapshot` (`vitest` doesn't support wrapping snapshot functions the way we could with `jest-snapshot`)
  - replaced `spyOnConsole` with inline `console` spies for the tests that require it
  - replaced `sleep` with `setTimeout` from `node:timers/promises`
- removed redundant patches


Andarist (changes requested):
I'd like to block this on updating the VS Code [`launch.json` script](https://github.com/statelyai/xstate/blob/437738dbe1c325ffe5807f39aa929c07e91bd005/.vscode/launch.json). It's essential to my workflow. 

Extra questions:
1. Does Vitest have an interactive watch mode in which I could filter tests using filenames or/and test names?
2. If yes, does it have something akin to [`jest-watch-typehead`](https://github.com/jest-community/jest-watch-typeahead)?

davidkpiano:
> I'd like to block this on updating the VS Code [`launch.json` script](https://github.com/statelyai/xstate/blob/437738dbe1c325ffe5807f39aa929c07e91bd005/.vscode/launch.json). It's essential to my workflow.
> 
> Extra questions:
> 
> 1. Does Vitest have an interactive watch mode in which I could filter tests using filenames or/and test names?
> 2. If yes, does it have something akin to [`jest-watch-typehead`](https://github.com/jest-community/jest-watch-typeahead)?

Yes; here is an example in `@statelyai/agent`:

CleanShot 2024-08-07 at 07 39 00

with-heart:
Does anyone know anything about `svelte-check`? I have no idea how to get it to pick up the `types: ["vitest/globals"]` in `tsconfig.json`. Couldn't really find anything on its repo either.

with-heart:
> I'd like to block this on updating the VS Code [launch.json script](https://github.com/statelyai/xstate/blob/437738dbe1c325ffe5807f39aa929c07e91bd005/.vscode/launch.json). It's essential to my workflow.

I've got you buddy 🤗

with-heart:
> I'd like to block this on updating the VS Code [launch.json script](https://github.com/statelyai/xstate/blob/437738dbe1c325ffe5807f39aa929c07e91bd005/.vscode/launch.json). It's essential to my workflow.

@Andarist Wanna _test_ it locally and make sure it's working the way you expect?

Andarist:
I noticed it's fairly easy to lose those typeahead filters in watch mode (for example pressing Enter clears them). It's annoying but not a big deal.

When debugging with VS Code, I can't disable source mapped stepping - it still shows me a location in the source file (albeit a different/broken one). This is way more annoying, probably something I can live with but well. I just wish that console evaluation could use original references, is this too much to ask for? 😭 😭 😭 (to clarify, this is just something that current debuggers suck at cause they are not utilizing sourcemapped info to its full extent)

with-heart:
> I noticed it's fairly easy to lose those typeahead filters in watch mode (for example pressing Enter clears them). It's annoying but not a big deal.

Yeah for some reason they chose to use `r` for re-running filtered tests. Quite an annoying change if you're a long-time `jest` user 🥲

packages/core/test/after.test.ts
```diff
@@ -85,77 +85,83 @@ describe('delayed transitions', () => {
...
-      initial: 'nested',
-      states: {
-        nested: {
-          initial: 'wait',
-          states: {
-            wait: {
-              after: {
-                10: '#end'
+  it('should be able to transition with delay from nested initial state', () =>
+    new Promise<void>((resolve) => {
```
Andarist, line 89: Those kind of changes could probably be replaced with `async` functions and `waitFor`. Some might not be but then I'd go with `Promise.withResolvers` (which might not be available in our testing envs, I'm not sure). This would allow u to create a deferred thing at the top of the async function, await it at the bottom before ur `expect` calls and `.resolve`/`.reject` it from callbacks. This would avoid adding extra indentation to the tests that got refactored by introducing `new Promise`

with-heart: imo the `new Promise` callback style works well for assertion-free tests (where the test is basically just that it doesn't time out (because `resolve` was called)).

I think this example shows the pitfalls of `waitFor`/`Promise.withResolvers` with this style of test:

```ts
test('new Promise', () =>
  new Promise((resolve) => {
      initial: 'waiting',
      states: {
        waiting: {
          after: {
            10: 'done'
          }
        },
        done: {
          entry: resolve
        }
      }
    });
    createActor(machine).start();

    // no assertions or other logic necessary. the test only passes if the
    // promise resolves; otherwise, it times out
  }));

test('waitFor', async () => {
    initial: 'waiting',
    states: {
      waiting: {
        after: {
          10: 'done'
        }
      },
      done: {}
    }
  });
  actor.start();

  // we need to wait for + assert the transition. if we forget to `await` or
  // forget `waitFor` entirely, the test passes even though we haven't actually
  // tested anything
  await waitFor(actor, (snapshot) => snapshot.matches('done'));
});

test('Promise.withResolvers', async () => {
  const { promise, resolve } = Promise.withResolvers();

    initial: 'waiting',
    states: {
      waiting: {
        after: {
          10: 'done'
        }
      },
      done: {
        entry: resolve
      }
    }
  });
  actor.start();

  // if we don't `await promise` or `return promise`, the test passes even
  // though `resolve` isn't called. would be an easy mistake to make and the
  // cost is that the test isn't actually testing anything even though it's
  // passing
  return promise;
});
```

Definitely agree with `waitFor`/`Promise.withResolvers` for tests where we're actually making assertions and such though.

Andarist: > Though actually now that I'm thinking about it, maybe we shouldn't have assertion-free tests to begin with 🙃

This. I think it's always best to assert something explicitly. That always gives some extra insight to what is actually tested

boneskull:
Just wanted to plug [Wallaby](https://wallabyjs.com) for running tests in "watch" mode (or even as you type).  I've been a user for years and I haven't found anything better.

Andarist:
@with-heart could u take a look at Knip's errors here?

with-heart:
@Andarist yeap I'll take a look today!

davidkpiano:
Superseded by #5293

Andarist:
The fork's permissions has changed and we no longer can push to it - that's the reason this work has been pulled into a fresh PR.

PR #5027, opened 2024-08-10
[@xstate/store] Add `store.inspect(…)`

davidkpiano:
You can now inspect XState stores using the `.inspect(inspector)` method:

```ts
import { someStore } from './someStore';

someStore.inspect((inspEv) => {
  console.log(inspEv);
  // logs "@xstate.event" events and "@xstate.snapshot" events
  // whenever an event is sent to the store
});
// The "@xstate.actor" event is immediately logged
```

### DIFF
--- a/packages/xstate-store/src/store.ts
+++ b/packages/xstate-store/src/store.ts
@@ -1,3 +1,4 @@
+import type { InspectionEvent } from 'xstate';
 import {
   Recipe,
   EventPayloadMap,
@@ -41,6 +42,11 @@ function setter<TContext extends StoreContext>(
   return recipe(context);
 }
 
+const inspectionObservers = new WeakMap<
+  Store<any, any>,
+  Set<Observer<InspectionEvent>>
+>();
+
 function createStoreCore<
   TContext extends StoreContext,
   TEventPayloadMap extends EventPayloadMap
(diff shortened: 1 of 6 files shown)


packages/xstate-store/src/store.ts
```diff
@@ -76,11 +82,31 @@ function createStoreCore<
...
+        actorRef: store,
+        rootId: store.sessionId
+      });
+    });
+
     observers?.forEach((o) => o.next?.(currentSnapshot));
   }
 
   const store: Store<TContext, StoreEvent> = {
+    sessionId: 'test',
```
Andarist, line 99: :cough: :cough:

packages/xstate-store/src/store.ts
```diff
@@ -1,3 +1,4 @@
+import type { InspectionEvent } from 'xstate';
```
Andarist, line 1: `@xstate/store` doesn't depend on `xstate`, you can't reach for its types here

packages/xstate-store/src/types.ts
```diff
@@ -1,3 +1,5 @@
+import { InspectionEvent } from 'xstate';
```
Andarist, line 1: same as https://github.com/statelyai/xstate/pull/5027/files#r1716421770

PR #5037, opened 2024-08-14
[@xstate/store] Fix store inspection issues

davidkpiano:
#5027

### DIFF
--- a/packages/xstate-store/src/store.ts
+++ b/packages/xstate-store/src/store.ts
@@ -1,4 +1,3 @@
-import type { InspectionEvent } from 'xstate';
 import {
   Recipe,
   EventPayloadMap,
@@ -11,7 +10,8 @@ import {
   StorePropertyAssigner,
   Observer,
   StoreContext,
-  InteropSubscribable
+  InteropSubscribable,
+  InspectionEvent
 } from './types';
 
 const symbolObservable: typeof Symbol.observable = (() =>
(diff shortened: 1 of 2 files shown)


davidkpiano:
@Andarist I'm not sure why `store` does not extend `ActorRefLike` - TypeScript isn't too helpful here

Andarist:
By passing `actorRef` to `observer.next` (and more) you make it possible for the receiver of that reference to call `actorRef.send`. On the receiver's end, it's possible to call that `.send` method with `EventObject` - but the actual `actorRef` that you have passed there doesn't allow just any `EventObject`. It only accepts `ExtractEventsFromPayloadMap<TEventPayloadMap>` and conceptually it's just a subtype of `EventObject`. That's why it errors.

packages/xstate-store/src/store.ts
```diff
@@ -341,3 +341,8 @@ export function createStoreTransition<
     return { ...snapshot, context: currentContext };
   };
 }
+
+// create a unique 6-char id
+export function uniqueId() {
+  return Math.random().toString(36).substring(2, 8);
```
Andarist, line 347: this would have the same effect but it would be shorter :D
```suggestion
  return Math.random().toString(36).slice(6);
```

packages/xstate-store/src/types.ts
```diff
@@ -197,3 +195,72 @@ export type EventObject = {
...
+}
+
+// export type ActorRefLike = Pick<
+//   AnyActorRef,
+//   'sessionId' | 'send' | 'getSnapshot'
+// >;
+
+export type ActorRefLike = {
+  sessionId: string;
+  send: (...args: never) => void;
```
Andarist, line 264: this is how I fixed the issue here, this makes `send` method not callable so TS can assume now that we can pass a store expecting some concrete event types to an external consumer as that consumer won't be able to actually call `.send`

PR #5041, opened 2024-08-16
add eslint

with-heart:
This PR installs and configures [`eslint`](https://eslint.org/) with [`typescript-eslint`](https://typescript-eslint.io/).

It uses the new [flat config format](https://eslint.org/docs/latest/use/configure/migration-guide#start-using-flat-config-files) (which is such a nice improvement imo), along with the `recommended` rules from `@eslint/js` and `typescript-eslint`.

Since #5022 is a thing, I added all `test.*` files to the global `ignores`. Once that lands we can remove this and maybe add [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest).


with-heart:
I wanted to try this with [type-aware linting](https://typescript-eslint.io/getting-started/typed-linting) but it reported hundreds of errors that didn't seem like they'd be easy to fix. They're likely related to having to set `@typescript-eslint/no-explicit-any` to `warn` (581 warnings for that rule alone 😬).

Andarist:
how many we'd get after disabling `@typescript-eslint/no-explicit-any`?

with-heart:
@Andarist 0. Should we disable it? I wasn't sure if this was something that'd be worth trying to deal with or not

with-heart:
I'm not sure what the deal is with the failed check. The error it reports isn't actually relevant.

To make build work, I have to add a trailing comma to type params in `packages/xstate-solid`:

```diff
diff --git a/packages/xstate-solid/src/createImmutable.ts b/packages/xstate-solid/src/createImmutable.ts
index 5624473d5b..e3237c7414 100644
--- a/packages/xstate-solid/src/createImmutable.ts
+++ b/packages/xstate-solid/src/createImmutable.ts
@@ -19,7 +19,7 @@ const updateStore = <Path extends unknown[]>(
   store: Store<any>
 ) => {
   const valueRefs = new WeakMap<any, unknown>();
-  const diff = <CompareValue>(
+  const diff = <CompareValue,>(
     next: CompareValue,
     prev: CompareValue,
     path: Path
diff --git a/packages/xstate-solid/src/deepClone.ts b/packages/xstate-solid/src/deepClone.ts
index e829300845..58bc92e28d 100644
--- a/packages/xstate-solid/src/deepClone.ts
+++ b/packages/xstate-solid/src/deepClone.ts
@@ -17,7 +17,7 @@ export function isWrappable(obj: any): obj is object {
  * @param valueRefs A WeakMap that stores a reference from the original
  *   object/array to the cloned object/array
  */
-const clone = <T>(value: T, valueRefs: WeakMap<any, any>): T => {
+const clone = <T,>(value: T, valueRefs: WeakMap<any, any>): T => {
   if (!isWrappable(value)) {
     return value;
   }
@@ -47,4 +47,4 @@ const clone = <T>(value: T, valueRefs: WeakMap<any, any>): T => {
   return clonedValue;
 };
 
-export const deepClone = <T>(value: T): T => clone(value, new WeakMap());
+export const deepClone = <T,>(value: T): T => clone(value, new WeakMap());
```

I don't understand why. And that's not gonna work anyway because `prettier` removes the trailing commas.

boneskull:
> I wanted to try this with [type-aware linting](https://typescript-eslint.io/getting-started/typed-linting) but it reported hundreds of errors that didn't seem like they'd be easy to fix. They're likely related to having to set `@typescript-eslint/no-explicit-any` to `warn` (581 warnings for that rule alone 😬).

IMO this is worth doing.  You will probably need to disable stuff if using the `recommendedTypeChecked` preview; it's pretty strict.

Here's some of the rules I'd suggest tweaking:

```js
// and sometimes you gotta use any
'@typescript-eslint/no-explicit-any': 'off',

// allow ! 
'@typescript-eslint/no-non-null-assertion': 'off',

// allow prefixing unused vars with _
'@typescript-eslint/no-unused-vars': [
  'error',
  {
    argsIgnorePattern: '^_',
    varsIgnorePattern: '^_',
  },
],

// better for overloads
'@typescript-eslint/unified-signatures': [
  'error',
  {
    ignoreDifferentlyNamedParameters: true,
  },
],
```

Something like the above might get you out of the woods.

Also I'd recommend enabling this (if it's not already enabled by default), which surfaces unused `// eslint-disable` directives:

```js
{
  linterOptions: {
    reportUnusedDisableDirectives: 'error',
  },
}
```

I'm not sure what you were using for the parser config, but this is what has worked well for me with tseslint@v8. 

```js
languageOptions: {
  parserOptions: {
    projectService: {
      // type-aware linting for non-project config files. modify as necessary
      allowDefaultProject: ['*.js', '.*.js'],
    },
    tsconfigRootDir: import.meta.dirname, // (or __dirname, depending)
  },
},
```

boneskull:
I don't think `T extends unknown` is doing anything in that function, and can just be `T`.

davidkpiano:
@with-heart Can you resolve merge conflicts?

with-heart:
> @with-heart Can you resolve merge conflicts?

Done! I think we should be good to go assuming the changes look fine 🙏

davidkpiano (approved):
cc. @Andarist

packages/xstate-store/src/types.ts
```diff
@@ -109,7 +109,7 @@ export interface Store<
 
 export type AnyStore = Store<any, any, any>;
 
-export type Compute<A extends any> = { [K in keyof A]: A[K] } & unknown;
+export type Compute<A> = { [K in keyof A]: A[K] };
```
Andarist, line 112: `& unknown` is required sometimes
```suggestion
export type Compute<A> = { [K in keyof A]: A[K] } & unknown;
```

packages/xstate-solid/src/deepClone.ts
```diff
@@ -17,6 +17,7 @@ export function isWrappable(obj: any): obj is object {
  * @param valueRefs A WeakMap that stores a reference from the original
  *   object/array to the cloned object/array
  */
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint
 const clone = <T extends unknown>(
```
Andarist, line 21: would this work instead?
```suggestion
const clone = <T,>(
```

packages/xstate-inspect/src/server.ts
```diff
@@ -73,6 +73,7 @@ export function inspect(options: ServerInspectorOptions): Inspector {
         return;
       }
 
+      // eslint-disable-next-line @typescript-eslint/no-base-to-string
       const jsonMessage = JSON.parse(data.toString());
```
Andarist, line 77: Would it be OK with this?
```suggestion
      const jsonMessage = JSON.parse(String(data));
```

packages/core/src/types.ts
```diff
@@ -485,7 +485,7 @@ export type StateTypes =
   | 'parallel'
   | 'final'
   | 'history'
-  | string; // TODO: remove once TS fixes this type-widening issue
+  | ({} & string); // TODO: remove once TS fixes this type-widening issue
```
Andarist, line 488:
```suggestion
  | ({} & string);
```

packages/core/src/types.ts
```diff
@@ -50,7 +50,7 @@ type ReturnTypeOrValue<T> = T extends AnyFunction ? ReturnType<T> : T;
 export type IsNever<T> = [T] extends [never] ? true : false;
 export type IsNotNever<T> = [T] extends [never] ? false : true;
 
-export type Compute<A extends any> = { [K in keyof A]: A[K] } & unknown;
+export type Compute<A> = { [K in keyof A]: A[K] };
```
Andarist, line 53:
```suggestion
export type Compute<A> = { [K in keyof A]: A[K] } & {};
```

packages/core/src/types.ts
```diff
@@ -41,7 +41,7 @@ export type GetParameterizedParams<T extends ParameterizedObject | undefined> =
  * This type can be used to avoid this problem. This union represents the same
  * value space as `unknown`.
  */
-export type NonReducibleUnknown = {} | null | undefined;
+export type NonReducibleUnknown = NonNullable<unknown> | null | undefined;
```
Andarist, line 44: I officially hate this... :joy: I kinda understand why `{}` has such a bad rep... but if I'll have to start using `NonNullable<unknown>` everywhere to workaround a lint rule... 😬 

Personally, I'd vote for disabling that rule and allowing `{}`

packages/core/src/guards.ts
```diff
@@ -115,8 +115,8 @@ export function stateIn<
   any // TODO: recheck if we could replace this with something better here
 > {
   function stateIn(
-    args: GuardArgs<TContext, TExpressionEvent>,
-    params: TParams
+    _args: GuardArgs<TContext, TExpressionEvent>,
+    _params: TParams
```
Andarist, line 119: those might be seen by users when debugging in devtools, I think it would be best to avoid `_` for those

packages/core/src/actions/stopChild.ts
```diff
@@ -102,8 +101,8 @@ export function stopChild<
   actorRef: ResolvableActorRef<TContext, TExpressionEvent, TParams, TEvent>
 ): StopAction<TContext, TExpressionEvent, TParams, TEvent> {
   function stop(
-    args: ActionArgs<TContext, TExpressionEvent, TEvent>,
-    params: TParams
+    _args: ActionArgs<TContext, TExpressionEvent, TEvent>,
```
Andarist, line 104: same in all of those actions

PR #5042, opened 2024-08-16
feat(core): add signal option to waitFor

boneskull:
This PR adds a `signal?: AbortSignal` option to `WaitForOptions`, so that a call to `waitFor()` can be externally aborted.

### DIFF
--- a/packages/core/src/waitFor.ts
+++ b/packages/core/src/waitFor.ts
@@ -9,6 +9,9 @@ interface WaitForOptions {
    * @defaultValue Infinity
    */
   timeout: number;
+
+  /** A signal which stops waiting when aborted. */
+  signal?: AbortSignal;
 }
 
 const defaultWaitForOptions: WaitForOptions = {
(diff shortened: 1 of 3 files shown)


packages/core/src/waitFor.ts
```diff
@@ -64,6 +72,9 @@ export function waitFor<TActorRef extends AnyActorRef>(
       clearTimeout(handle!);
       done = true;
       sub?.unsubscribe();
+      if (abortListener) {
+        signal!.removeEventListener('abort', abortListener);
```
Andarist, line 76: timeout should also lead to removing this

davidkpiano (approved):

packages/core/src/waitFor.ts
```diff
@@ -57,13 +65,19 @@ export function waitFor<TActorRef extends AnyActorRef>(
         ? undefined
         : setTimeout(() => {
             sub!.unsubscribe();
+            if (abortListener) {
+              signal!.removeEventListener('abort', abortListener);
+            }
```
Andarist, line 70: replacing those 4 lines with `dispose()` should be fine here from what I can tell

Andarist (approved):

Issue #4928, opened 2024-06-07
Bug: error thrown from initial context assignment thrown too much


boneskull:
### XState version

XState version 5

### Description

Given:

1. Root machine actor _A_
2. No `error` subscriber is registered with _A_
3. _A_'s initial `context` assignment function throws an error _E_

_E_ cannot be caught by any mechanism other than a global error listener (e.g., `window.onerror` or `process.on('uncaughtException')`.

Even if a machine actor is wrapped in `toPromise()`, _E_ is still thrown this way _in addition to_ being rejected out of the `Promise`.

This behavior _does not occur_ if the machine actor has a subscriber with an `error` listener.

I _really_ don't want to create a bunch of subscriptions where `toPromise()` would work for my purposes. 😄 

### Expected result

The error thrown by the `context` initializer should not be thrown in addition to being rejected.

### Actual result

The "extra" error thrown is only catchable via a global error handler.

### Reproduction

https://stackblitz.com/edit/github-trfney?file=src%2Fmain.ts

### Additional context

This may be expected behavior.  I understand that an actor that is started without an `error` listener _and_ isn't wrapped in a `Promise` _has no other choice_ but to throw an error this way; there's nowhere for it to go.

In Node.js, an EE will emit an `error` event, which can be listened to.  If it is listened to, then the EE won't throw an uncaught exception.  However, the `once` wrapper (which is only vaguely analogous to `toPromise()`) will _also_ avoid the uncaught exception:

```ts
import {once, EventEmitter} from 'node:events';

const emitter = new EventEmitter();

// without this, the error would be uncaught
once(emitter, 'error').then(err => {
  // handle error
});

emitter.emit('error', new Error('handled'));
```

That's why I think this is a bug. Like above, `toPromise` should _act like_ a subscriber with an `error` listener.

It may also not be unique to the `context` assignment function, either...

boneskull:
I updated the title to better describe what I think is happening; my hunch is that it can be fixed in the `toPromise` implementation.

boneskull:
It may be worth mentioning _why_ the `context` assignment function would ever want to throw.  Because:

```ts
const machine = setup({
  types: {
    input: {} as {signal: AbortSignal}
  }
}).createMachine({
  context: ({input: {signal}}) => {
    if (signal.aborted) {
      throw new AbortError(signal.reason);
    }
  }
});
```

Andarist:
[...]

PR #5045, opened 2024-08-17
fix(core): errored initial snapshot throws sync w/o observers

boneskull:
This modifies the error-handling logic for when an error is thrown during the creation of an Actor's initial snapshot (during `start()`).  _If_ the actor has _no_ observers, the error will now be thrown synchronously out of `start()` instead of to the global error handler.

Example use case:

```js
  context: () => {
    throw new Error('egad!');
  }
});

try {
  await toPromise(actor.start());
} catch (err) {
  err.message === 'egad!' // true
}
```

Note that this _does not impact child actors_, but it might want to!

Fixes: #4928

### DIFF
--- a/packages/core/src/createActor.ts
+++ b/packages/core/src/createActor.ts
@@ -491,7 +491,26 @@ export class Actor<TLogic extends AnyActorLogic>
         // TODO: rethink cleanup of observers, mailbox, etc
         return this;
       case 'error':
+        // in this case, creation of the initial snapshot caused an error.
+
+        // **if the actor has no observer when start() is called**, this error would otherwise be
+        // thrown to the global error handler. to prevent this--and allow start() to be wrapped in try/catch--we
+        // create a temporary observer that receives the error (via _reportError()) and rethrows it from this call stack.
+        let err: unknown;
+        let errorTrapSub: Subscription | undefined;
+        if (!this.observers.size) {
+          errorTrapSub = this.subscribe({
+            error: (error) => {
+              // we cannot throw here because it would be caught elsewhere and rethrown as unhandled
+              err = error;
+            }
+          });
+        }
         this._error((this._snapshot as any).error);
+        errorTrapSub?.unsubscribe();
+        if (err) {
+          throw err;
+        }
         return this;
     }
 
--- a/packages/core/test/errors.test.ts
+++ b/packages/core/test/errors.test.ts
@@ -5,7 +5,8 @@ import {
   createMachine,
   fromCallback,
   fromPromise,
-  fromTransition
+  fromTransition,
+  toPromise
 } from '../src';
 
 const cleanups: (() => void)[] = [];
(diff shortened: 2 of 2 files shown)


boneskull:
For reference, the current workaround is this:

```js
const actor = createBrokenActor();
  context: () => {
    throw new Error('egad!');
  }
});

const p = toPromise(actor);
actor.start();
try {
  await p;
} catch (err) {
  err.message === 'egad!' // true
}
```

boneskull:
I don't know if the new behavior is desirable to the maintainers, but _I_ find it more ergonomic. Thought I would restart that conversation here!

davidkpiano (commented):
I agree with this PR in principle; waiting to see what @Andarist thinks

boneskull:
My main concern is that it is somewhat inconsistent, since if the example machine was spawned as a child actor, it _would_ reach the global error handler.

boneskull:
This looks like a breaking change to me unless you feel this is _actually_ a "bug fix" and not just a "revision of intent" 😜

Andarist:
> ...that being said, if using spawnChild(), there's really nowhere else for the error to go.

Could you show the exact situation you are talking about here?

boneskull:
> > ...that being said, if using spawnChild(), there's really nowhere else for the error to go.
> 
> Could you show the exact situation you are talking about here?

@Andarist 

This, I think:

https://github.com/boneskull/xstate/blob/0dabaa7be7e11b7567def5e1b747b4077cd8703f/packages/core/test/errors.test.ts#L336C4-L367

packages/core/src/createActor.ts
```diff
@@ -491,7 +491,26 @@ export class Actor<TLogic extends AnyActorLogic>
         // TODO: rethink cleanup of observers, mailbox, etc
         return this;
       case 'error':
+        // in this case, creation of the initial snapshot caused an error.
+
+        // **if the actor has no observer when start() is called**, this error would otherwise be
+        // thrown to the global error handler. to prevent this--and allow start() to be wrapped in try/catch--we
+        // create a temporary observer that receives the error (via _reportError()) and rethrows it from this call stack.
```
davidkpiano, line 498: I'm wondering if we can do this a different way without creating a temp sub

boneskull: @davidkpiano The folded section [in this comment](https://github.com/statelyai/xstate/issues/4928#issuecomment-2294458587) has an alternate proposal, which I explored but didn't get working fully.

packages/core/test/errors.test.ts
```diff
@@ -892,4 +893,41 @@ describe('error handling', () => {
...
+
+    const actor = createActor(machine);
+
+    try {
+      await toPromise(actor.start());
+      fail();
+    } catch (err) {
+      expect((err as Error).message).toEqual('oh no');
+    }
+  });
```
Andarist, line 932: This doesn't work like you think it does. This error is still very much thrown synchronously. `toPromise` isn't ever called during this test case.

packages/core/test/errors.test.ts
```diff
@@ -892,4 +893,41 @@ describe('error handling', () => {
...
+      }
+    });
+
+    const actor = createActor(machine);
+
+    installGlobalOnErrorHandler(() => {
+      done.fail();
+    });
+
+    expect(() => actor.start()).toThrowErrorMatchingInlineSnapshot(`"oh no"`);
```
Andarist, line 910: actors are always async conceptually, a synchronous throw is something that might sound convenient at first but you essentially create 2 error channels by introducing this. Now, if you want to handle errors, u need to both install an error listener and wrap this in try/catch. The guiding principle that we had when designing this was that you should always be notified about errors in the very same way (through error listeners).

davidkpiano: @Andarist In theory I agree, but in practice, if we know that an actor is going to fail right at the start, rather than later during event processing, it's a _much_ better DX (IMO) to throw right away.

packages/core/test/errors.test.ts
boneskull: I'm not sure what you mean.  If `createActor` threw synchronously, wouldn't this test fail?

packages/core/test/errors.test.ts
boneskull: This is my feeling too.  User experience > theoretical purity

davidkpiano: @boneskull I may have changed my stance on this: https://www.youtube.com/watch?v=JCXZhe6KsxQ&t=294s

boneskull: I don't follow

boneskull: @davidkpiano Should I close this?

davidkpiano: Yes, let's close it for now, given that it's a breaking change.

PR #5056, opened 2024-08-27
Add initial support for xstate/store/solid

steveadams:
I've added `useSelector` support to xstate/store for SolidJS which seems to be working well in my tests. Let me know what I can improve and I'd be happy to do it.

### DIFF
--- a/packages/xstate-store/src/solid.ts
+++ b/packages/xstate-store/src/solid.ts
@@ -7,27 +7,10 @@ function defaultCompare<T>(a: T | undefined, b: T) {
   return a === b;
 }
 
-function useSelectorWithCompare<
-  TStore extends
-    | Store<any, any>
-    | Pick<AnyActorRef, 'subscribe' | 'getSnapshot'>,
-  T
->(
-  selector: (
-    snapshot: TStore extends Store<any, any>
-      ? SnapshotFromStore<TStore>
-      : TStore extends { getSnapshot(): infer TSnapshot }
-        ? TSnapshot
-        : never
-  ) => T,
-  compare: (a: T | undefined, b: T) => boolean = defaultCompare
-): (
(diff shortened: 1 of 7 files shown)


steveadams:
Apologies for the force pushes; I can move this to a clean branch with a single commit instead if they're a problem.

packages/xstate-store/src/solid.ts
```diff
@@ -0,0 +1,79 @@
...
+ * ```
+ *
+ * @param store The store, created from `createStore(…)`
+ * @param selector A function which takes in the snapshot and returns a selected
+ *   value from it
+ * @param compare An optional function which compares the selected value to the
+ *   previously selected value
+ * @returns A read-only signal of the selected value
+ */
+export function useSelector<TStore extends Store<any, any>, T>(
```
steveadams, line 57: I initially went with the name `createSelector`, and I might actually prefer it. I'm not sure what the consensus is on naming functions like this in SolidJS. The library itself has `use*` as well as `create*` naming conventions for functions like this.

davidkpiano: That's fine, whatever is most idiomatic!

steveadams: I'm just noticing the other solid-related code uses the `useActor` or `useMachine` convention so I guess this makes sense.

packages/xstate-store/src/solid.ts
```diff
@@ -0,0 +1,79 @@
+/* @jsxImportSource solid-js */
+import { AnyActorRef } from 'xstate';
```
steveadams, line 2: Whoops, will remove (left over from experiments)

steveadams:
You're welcome! I really enjoyed this process and I'm stoked to contribute to XState.

PR #5064, opened 2024-09-04
[@xstate/store] Event emitter

davidkpiano:
You can now emit events from a store:

```ts

const store = createStore({
  context: {
    count: 0
  },
  on: {
    increment: (context, event, { emit }) => {
      emit({ type: 'incremented' });
      return { count: context.count + 1 };
    }
  }
});

store.on('incremented', () => {
  console.log('incremented!');
});
```

You can make emitted events type-safe via `createStore({ types: { … } })`


packages/xstate-store/src/setup.ts
```diff
@@ -0,0 +1,51 @@
...
+            NoInfer<TContext>,
+            { type: K } & TEventPayloadMap[K],
+            TEmitted
+          >;
+    }
+  ) => Store<TContext, ExtractEventsFromPayloadMap<TEventPayloadMap>, TEmitted>;
+};
+
+export function setup<
+  const T extends { schemas: { emitted: Record<string, {}> } }
```
Andarist, line 42:
```suggestion
  const T extends { schemas: { emitted: Record<string, { _output: unknown }> } }
```

Andarist: I think this should help with mistakes like this:
```ts
    const store = setup({
      schemas: {
        emitted: {
          increased: { not: 'zod' },
        }
      }
    })
```

packages/xstate-store/src/setup.ts
```diff
@@ -0,0 +1,51 @@
...
+  Compute,
+  EventPayloadMap,
+  ExtractEventsFromPayloadMap,
+  Store,
+  StoreAssigner,
+  StoreContext,
+  StorePropertyAssigner
+} from './types';
+
+type MaybeZod<T> = T extends { _output: infer U } ? U : T;
```
Andarist, line 13: likely this should be OK with the other change:
```suggestion
type MaybeZod<T> = T extends { _output: infer U } ? U : never;
```

or once u fix constraints:
```suggestion
type MaybeZod<T extends { _output: unknown }> = T["_output"];
```

davidkpiano:
@tonivj5 Do you have any thoughts on this to share? (responding to emoji reaction)

packages/xstate-store/src/types.ts
```diff
@@ -265,3 +301,9 @@ export type ActorRefLike = {
   send: (...args: never) => void;
   getSnapshot: () => any;
 };
+
+export type Prop<T, K> = K extends keyof T ? T[K] : never;
+
+export interface StoreSchemas {
+  emitted?: Record<string, { _output: unknown }>;
+}
```
Andarist, line 309: TBH, I'm not sure why this is introduced here. It should be a separate followup PR. I also don't necessarily understand why this one only adds support for `schema.emitted`. What about `schema.events`?

packages/xstate-store/src/fromStore.ts
```diff
@@ -1,11 +1,47 @@
...
+//       type: "inc";
+//   }) => StoreSnapshot<StoreContext>;
+//   start: () => void;
+//   getInitialSnapshot: (_: any, input: number) => {
+//       ...;
+//   };
+//   getPersistedSnapshot: (s: StoreSnapshot<...>) => StoreSnapshot<...>;
+//   restoreSnapshot: (s: Snapshot<unknown>) => StoreSnapshot<any>;
+// }
+
```
Andarist, line 44:
```suggestion
```

packages/xstate-store/src/fromStore.ts
```diff
@@ -1,11 +1,47 @@
...
+import {
+  createStore,
+  createStoreTransition,
+  EmittedFromSchemas,
+  TransitionsFromEventPayloadMap
+} from './store';
 import {
   EventPayloadMap,
   StoreContext,
+  // StoreSnapshot,
```
Andarist, line 10:
```suggestion
```

packages/xstate-store/src/fromStore.ts
```diff
@@ -16,17 +52,50 @@ import {
...
+//   on: TransitionsFromEventPayloadMap<
+//     TEventPayloadMap,
+//     TContext,
+//     EmittedFromSchemas<TSchemas['emitted']>
+//   >;
+// } & { schemas?: TSchemas }): StoreLogic<
+//   TContext,
+//   ExtractEventsFromPayloadMap<TEventPayloadMap>,
+//   TInput
+// >;
```
Andarist, line 75:
```suggestion
```

packages/xstate-store/src/fromStore.ts
```diff
@@ -16,17 +52,50 @@ import {
...
+  //: StoreLogic<TContext, ExtractEventsFromPayloadMap<TEventPayloadMap>, TInput>;
+  // export function fromStore(
+  //   initialContextOrObject: any,
+  //   transitions?: any
+  // ): StoreLogic<any, any, any>
+  // let [initialContext, actualTransitions] =
+  //   transitions === undefined
+  //     ? [initialContextOrObject.context, initialContextOrObject.on]
+  //     : [initialContextOrObject, transitions];
+
```
Andarist, line 97:
```suggestion
```

packages/xstate-store/src/fromStore.ts
```diff
@@ -42,6 +111,6 @@ export function fromStore<
       } satisfies StoreSnapshot<TContext>;
     },
     getPersistedSnapshot: (s: StoreSnapshot<TContext>) => s,
-    restoreSnapshot: (s: Snapshot<unknown>) => s as StoreSnapshot<TContext>
+    restoreSnapshot: (s: Snapshot<unknown>) => s as StoreSnapshot<any>
```
Andarist, line 114: why do we want to implicitly leak `any` here?

packages/xstate-store/src/store.ts
```diff
@@ -43,41 +46,59 @@ function setter<TContext extends StoreContext>(
...
+): Store<TContext, ExtractEventsFromPayloadMap<TEventPayloadMap>, TEmitted> {
   type StoreEvent = ExtractEventsFromPayloadMap<TEventPayloadMap>;
   let observers: Set<Observer<StoreSnapshot<TContext>>> | undefined;
+  let listeners: Map<TEmitted['type'], Set<any>> | undefined;
   const initialSnapshot = {
     context: initialContext,
     status: 'active',
     output: undefined,
     error: undefined
   } satisfies StoreSnapshot<TContext> as StoreSnapshot<TContext>;
```
Andarist, line 85:
```suggestion
  const initialSnapshot: StoreSnapshot<TContext> = {
    context: initialContext,
    status: 'active',
    output: undefined,
    error: undefined
  };
```

packages/xstate-store/src/store.ts
```diff
@@ -95,7 +116,23 @@ function createStoreCore<
...
+      }
+      const set = listeners.get(event)!;
+      set.add(fn);
+
+      return {
+        unsubscribe() {
+          set.delete(fn);
+        }
+      };
+    },
```
Andarist, line 135: please follow the binding scheme used by `createActor` here: https://github.com/statelyai/xstate/blob/5afff2cd0662347b6c031a3362392eb33ad15da0/packages/core/src/createActor.ts#L423-L442

this has some semantic differences and that binding scheme improves compat with the observable spec proposal

packages/xstate-store/src/store.ts
```diff
@@ -162,6 +199,144 @@ function createStoreCore<
...
+ *
+ * store.subscribe((snapshot) => {
+ *   console.log(snapshot);
+ * });
+ *
+ * store.send({ type: 'inc', by: 5 });
+ * // Logs { context: { count: 5 }, status: 'active', ... }
+ * ```
+ */
+export function createStore<
```
Andarist, line 253: why do we need **four** overloads here? this is soemwhat surprising

packages/xstate-store/src/store.ts
```diff
@@ -346,3 +590,16 @@ export function createStoreTransition<
...
+
+type MaybeZod<T> = T extends { _output: infer U } ? U : never;
+
+export type EmittedFromSchemas<TSchemas extends StoreSchemas> =
+  TSchemas extends {
+    emitted: infer TEmitted;
+  }
+    ? {
+        [K in keyof TEmitted]: Compute<
+          { type: K } & MaybeZod<TEmitted[K]> & EventObject
```
Andarist, line 602: `EventObject` doesn't bring any value here:
```suggestion
          { type: K } & MaybeZod<TEmitted[K]>
```

packages/xstate-store/test/fromStore.test.ts
```diff
@@ -21,4 +21,27 @@ describe('fromStore', () => {
...
+  //   const actor = createActor(storeLogic, {
+  //     input: 42
+  //   });
+
+  //   actor.start();
+
+  //   actor.send({ type: 'inc', by: 8 });
+
+  //   expect(actor.getSnapshot().context.count).toEqual(50);
+  // });
```
Andarist, line 46:
```suggestion
```

packages/xstate-store/src/store.ts
davidkpiano: I couldn't figure out "either `types` or `schema`, or none of them" and have it infer correctly. Maybe you can?

packages/xstate-store/src/types.ts
davidkpiano: I'll move schema types to a separate PR

packages/xstate-store/src/store.ts
davidkpiano: Typechecking fails without this

.changeset/silver-maps-grab.md
```diff
@@ -0,0 +1,56 @@
...
+  }
+});
+
+store.on('incremented', (event) => {
+  console.log('Incremented by', event.by);
+  // => "Incremented by 1"
+});
+
+store.send({ type: 'increment' });
+```
```
Andarist, line 56: this has been moved out of this PR
```suggestion
```

.changeset/silver-maps-grab.md
```diff
@@ -0,0 +1,56 @@
...
+  context: {
+    count: 0
+  },
+  on: {
+    increment: (context, event, { emit }) => {
+      emit({ type: 'incremented' });
+      return { count: context.count + 1 };
+    }
+  }
+});
```
Andarist, line 20: this is a new shape of this `createStore` API, right? it would be great to mention that

packages/xstate-store/src/fromStore.ts
Andarist, line 1: I'm not sure but shouldn't the new types and API shapes be applied to `fromStore` too?

davidkpiano:
@Andarist Types are failing

tonivj5:
hey @davidkpiano! My "corcerns" (or doubts) about these changes were about make zod "first citizen", I love the idea itself, but I personally prefer to avoid "hard" coupling and let it open to integrate other solutions using adapters or some other kind of technique (and I think @colinhacks is drafting an awesome solution to achive it with https://github.com/standard-schema/standard-schema).

Zod is a brilliant piece of software, I've used it before although I love and use other libraries too as https://github.com/sinclairzx81/typebox and I think this field is a moving target that it's going to beneficate from making it elegible.

Anyway, I'm just passing through here so take my reactions and comments with a grain of salt 😅. I love from time to time review your work and check what's cooking here 😸

davidkpiano:
@tonivj5 You're in luck - we moved the Zod stuff out and that will probably be a separate PR later, if we decide to do it.

Andarist (approved):

PR #5071, opened 2024-09-12
[core] Synchronizers

davidkpiano:


packages/core/src/createActor.ts
```diff
@@ -207,7 +211,23 @@ export class Actor<TLogic extends AnyActorLogic>
...
+
+    this._initState(initialSnapshot);
+
+    if (this._synchronizer) {
+      this._synchronizerSubscription = this._synchronizer.subscribe(
+        (rawSnapshot) => {
+          const restoredSnapshot =
+            this.logic.restoreSnapshot?.(rawSnapshot, this._actorScope) ??
+            rawSnapshot;
+          this.update(restoredSnapshot, { type: '@xstate.sync' });
```
Andarist, line 227: We don't use `@` for other builtin events like init, done.invoke and more
```suggestion
          this.update(restoredSnapshot, { type: 'xstate.sync' });
```

packages/core/src/createActor.ts
```diff
@@ -567,6 +588,7 @@ export class Actor<TLogic extends AnyActorLogic>
       return this;
     }
     this.mailbox.clear();
+    this._synchronizerSubscription?.unsubscribe();
```
Andarist, line 591: what snapshot should the synchronizer be aware of when somebody does smth like this:
```ts
const actorRef = createActor(m, { sync: someSynchronizer })

// ... do stuff

actorRef.stop()
actorRef.getSnapshot() // .status === 'stopped'
```

If I'm not mistaken the known snapshot after this operation has a stopped status but the synchronizer won't be aware of it because we have unsubscribed earlier.

packages/core/src/createActor.ts
Andarist, line 227: I think it should get enqueued and not handled immediately because with this model we allow for `this.update` to be potentially reentered (and that's a bad thing™). And if we are in the middle of processing 5 events mailbox and we receive a `rawSnapshot` here then likely we should append to the mailbox instead of processing it right away

packages/core/src/StateMachine.ts
```diff
@@ -471,13 +471,15 @@ export class StateMachine<
       TConfig
     >
   ): void {
-    Object.values(snapshot.children as Record<string, AnyActorRef>).forEach(
-      (child: any) => {
-        if (child.getSnapshot().status === 'active') {
-          child.start();
+    if (snapshot.children) {
```
Andarist, line 474: this could likely be reverted - no tests fails without this `if` 😉

PR #5072, opened 2024-09-13
Fix `docs` workflow

Andarist:
follow up to https://github.com/statelyai/xstate/pull/4987

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Build docs
         working-directory: docs
         run: |
-          NODE_OPTIONS='--openssl-legacy-provider' yarn docs:build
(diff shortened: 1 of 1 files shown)


davidkpiano (approved):

.github/workflows/docs.yml
```diff
@@ -15,15 +15,11 @@ jobs:
...
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 20.x
+      - uses: ./.github/actions/ci-setup
 
       - name: Build docs
         working-directory: docs
         run: |
-          yarn install
           NODE_OPTIONS='--openssl-legacy-provider' yarn docs:build
```
davidkpiano, line 23: Should this be this?

```suggestion
          NODE_OPTIONS='--openssl-legacy-provider' pnpm docs:build
```

PR #5078, opened 2024-09-16
add `registry-url` to `actions/setup-node`

with-heart:
This PR should hopefully fix the `ENEEDAUTH` error we're receiving in actions. 

According to https://docs.github.com/en/actions/use-cases-and-examples/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry:

> Please note that you need to set the `registry-url` to https://registry.npmjs.org/ in `setup-node` to properly configure your credentials.

### DIFF
--- a/.github/actions/ci-setup/action.yml
+++ b/.github/actions/ci-setup/action.yml
@@ -10,6 +10,7 @@ runs:
       with:
         node-version: 20.x
         cache: 'pnpm'
+        registry-url: 'https://registry.npmjs.org'
 
     - name: Install Dependencies
       run: pnpm install


.github/actions/ci-setup/action.yml
```diff
@@ -10,6 +10,7 @@ runs:
       with:
         node-version: 20.x
         cache: 'pnpm'
+        registry-url: 'https://registry.npmjs.org'
```
Andarist, line 13: bizarre that we might need to configure this given this really is the default

Andarist (approved):

.github/actions/ci-setup/action.yml
with-heart: Yeah I don't understand why that'd be the case 😭

PR #5080, opened 2024-09-18
[@xstate/store] Fix `useSelector(…)` infinite loop 

davidkpiano:
The `useSelector(…)` hook will no longer result in an infinite loop when an object-like value is returned from the selector function.

cc. @TkDodo - I had to bring back in `useSyncExternalStoreWithSelector` (copy-pasted, no dependencies), but if you have any better ideas for solving this, I'm open to a simpler solution.

### DIFF
--- a/packages/xstate-store/src/useSyncExternalStoreWithSelector.ts
+++ b/packages/xstate-store/src/useSyncExternalStoreWithSelector.ts
@@ -3,8 +3,6 @@ import * as React from 'react';
 const is = (a: any, b: any) => a === b;
 import { useSyncExternalStore } from 'react';
 
-// Intentionally not using named imports because Rollup uses dynamic dispatch
-// for CommonJS interop.
 const { useRef, useEffect, useMemo, useDebugValue } = React;
 
 // Same as useSyncExternalStore, but supports selector and isEqual arguments.
@@ -15,7 +13,6 @@ export function useSyncExternalStoreWithSelector<Snapshot, Selection>(
   selector: (snapshot: Snapshot) => Selection,
   isEqual?: (a: Selection, b: Selection) => boolean
 ): Selection {
-  // Use this to track the rendered snapshot.
   const instRef = useRef<
     | {
         hasValue: true;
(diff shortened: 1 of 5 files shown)


packages/xstate-store/src/useSyncExternalStoreWithSelector.ts
```diff
@@ -0,0 +1,101 @@
+// Copied from https://github.com/facebook/react/blob/0a76aecbfa3d493c6b97998d0444f6505f5e5365/packages/use-sync-external-store/src/useSyncExternalStoreWithSelector.js
```
Andarist, line 1: this hurts my soul, I completely miss the point of copy-pasting code just to claim the package has no redundant dependencies. Now imagine 10 libraries doing the same. As a result, the final application will have *more* code and not less.

packages/xstate-store/test/react.test.tsx
```diff
@@ -266,3 +266,35 @@ it('useActorRef (@xstate/react) should work', () => {
 
   expect(countDiv.textContent).toEqual('1');
 });
+
+it('useSelector should not infinite loop', () => {
```
Andarist, line 270: this is a known caveat listed in [the React docs](https://react.dev/reference/react/useSyncExternalStore#caveats):
> The store snapshot returned by getSnapshot must be immutable. If the underlying store has mutable data, return a new immutable snapshot if the data has changed. Otherwise, return a cached last snapshot.

Is it worth "patching" the problem on our side if this is just how things are "supposed" to work?

TkDodo: Wouldn't zustand v5 have the same problem? The solution here is heavily inspired by how they do it: 

https://github.com/pmndrs/zustand/blob/b93e11681d7b70c60dc99ae5296216a066f466b7/src/react/shallow.ts#L4

I think the solution here would be to have the user provide a different `compare` method like `shallow`. Otherwise, yes, returning new instances will have this impact.

packages/xstate-store/src/useSyncExternalStoreWithSelector.ts
davidkpiano: If there's a way we can fix the existing code (cc @tkdodo) then let's do that. Otherwise, I don't mind reintroducing the dependency.

TkDodo: I'm not sure there is something to fix. If you create a new object or array in your selector, the comparison method must account for that. Per default, we use `===` to compare, so it won't work.

It would need:

```
const count = useSelector(
  store,
  (s) => s.context.count.slice(),
  shallowEqual
);
```

davidkpiano: I wonder why it worked with `useSyncExternalStoreWithSelector` then?

packages/xstate-store/test/react.test.tsx
davidkpiano: To clarify, the problem here isn't that it rerenders every time; that's expected and a good use-case for providing a different `compare` method.

The problem is that it will cause an infinite loop:

> Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

Andarist: When this happens with the current code React even prints a warning about it:
> The result of getSnapshot should be cached to avoid an infinite loop

packages/xstate-store/src/useSyncExternalStoreWithSelector.ts
Andarist: this is why https://github.com/statelyai/xstate/pull/5084 :)

PR #5084, opened 2024-09-18
Avoid calling `selector` when the snapshot doesn't change

Andarist:
alternative to https://github.com/statelyai/xstate/pull/5080/files

### DIFF
--- a/packages/xstate-store/src/react.ts
+++ b/packages/xstate-store/src/react.ts
@@ -9,13 +9,18 @@ function useSelectorWithCompare<TStore extends AnyStore, T>(
   selector: (snapshot: SnapshotFromStore<TStore>) => T,
   compare: (a: T | undefined, b: T) => boolean
 ): (snapshot: SnapshotFromStore<TStore>) => T {
-  const previous = useRef<T>();
+  const previous = useRef<readonly [SnapshotFromStore<TStore>, T]>();
 
   return (state) => {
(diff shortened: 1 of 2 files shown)


davidkpiano:
I think there may be a hidden problem with this logic:

```ts
  const items = useSelector(store, s => {
    return s.context.items.map(item => ({
      ...item,
      isSelected: props.selected === item.id
    }));
  });
}
```

Not the best example but the point is that if we're selecting something that mixes in data outside the snapshot (whether that's best practice or not), we may end up with stale data.

TkDodo (approved):

Andarist:
That's a fair concern - this is how the React team has implemented' useSyncExternalStoreWithSelector', so I just used their logic here.

TkDodo:
> Not the best example but the point is that if we're selecting something that mixes in data outside the snapshot (whether that's best practice or not), we may end up with stale data.

Ah, I see. closing over values should still work - we could compare the selectorFn for equality but that wouldn't work with inline selectors :/

Andarist:
For context, this is where the original `useSyncExternalStoreWithSelector` is returning early:
https://github.com/facebook/react/blob/8dfbd16fce9077ab4e5fe85a7b86fa7c97a5ae04/packages/use-sync-external-store/src/useSyncExternalStoreWithSelector.js#L83-L86

davidkpiano:
Closing as we went with a different approach

PR #5095, opened 2024-09-25
fix: triggerRef in the update selected function

Hebilicious:


Andarist:
Please add a test case verifying the fixed behavior.

Andarist:
I'd prefer new dedicated PRs instead of repurposing this one

Hebilicious:
for reference https://github.com/statelyai/xstate/pull/5382

PR #5100, opened 2024-10-10
[@xstate/solid] Fix context mutation with new values nested in arrays

### LINKED ISSUE #5099: Bug: useActor from @xstate/solid mutates context
### XState version

XState version 5

### Description

Hi guys. Have been tinkering with XState and Solid recently. Found a surprising issue. Seems like diffing and setting updated store values in `createImmutable` implementation mutates the context within machine.

In my example, on `next` event first item within `current` (which is an object) gets removed and is put into the other array (`done`). `next` performed 3 times totally (until `current` is empty). Then `prev` event performed 3 times. I expect to see initial items within `current` array, but all of them are `{value: 'Stack #2'}`

### Expected result

Objects within arrays in context shouldn't mutate

### Actual result

Objects within arrays in XState context gets mutated when used with `useActor`

### Reproduction

https://codesandbox.io/p/devbox/runtime-thunder-qy36ws?embed=1&file=%2Fsrc%2FApp.tsx

### Additional context

_No response_

GoldingAustin:
Fix new array index value not being cloned before insertion into store - Fix #5099

* [`packages/xstate-solid/src/createImmutable.ts`](diffhunk://#diff-189122ca05a9c2d9e0559f05565de61a9b74b5c74be087a09f71dd6970a4b564R56-L71): Refactored the `updateStore` function to deep clone new elements in the array, ensuring immutability.

### Additions to test cases:

* [`packages/xstate-solid/test/useActor.test.tsx`](diffhunk://#diff-bd88a12ef25cb8c59659c67547866b04e5da5745f8e83c4accb6e98c0e73d16dR689-R814): Added a new test case to verify that moving objects between arrays does not mutate the context.

### DIFF
--- a/packages/xstate-solid/src/createImmutable.ts
+++ b/packages/xstate-solid/src/createImmutable.ts
@@ -53,6 +53,13 @@ const updateStore = <Path extends unknown[]>(
       const smallestSize = Math.min(prev.length, next.length);
       const largestSize = Math.max(next.length, prev.length);
 
+      // Update new or now undefined indexes
+      if (newIndices !== 0) {
+        for (let newEnd = smallestSize; newEnd <= largestSize - 1; newEnd++) {
+          set(...path, newEnd, deepClone(next[newEnd]));
+        }
+      }
+
       // Diff array
       for (let start = 0, end = largestSize - 1; start <= end; start++, end--) {
         diff(next[start], prev[start], [...path, start] as Path);
@@ -60,14 +67,9 @@ const updateStore = <Path extends unknown[]>(
         diff(next[end], prev[end], [...path, end] as Path);
       }
 
-      // Update new or now undefined indexes
-      if (newIndices !== 0) {
-        for (let newEnd = smallestSize; newEnd <= largestSize - 1; newEnd++) {
-          set(...path, newEnd, next[newEnd]);
-        }
-        if (prev.length > next.length) {
-          set(...path, 'length', next.length);
-        }
+      // Update length if it has changed
+      if (prev.length !== next.length) {
+        set(...path, 'length', next.length);
       }
     } else {
       // Update new values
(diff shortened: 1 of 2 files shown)


packages/xstate-solid/src/createImmutable.ts
```diff
@@ -53,21 +53,23 @@ const updateStore = <Path extends unknown[]>(
       const smallestSize = Math.min(prev.length, next.length);
       const largestSize = Math.max(next.length, prev.length);
 
+      // Update new or now undefined indexes
+      if (newIndices !== 0) {
+        for (let newEnd = smallestSize; newEnd <= largestSize - 1; newEnd++) {
+          set(...path, newEnd, deepClone(next[newEnd]));
```
Andarist, line 59: i thought we might only have to deep clone everything once upfront - do we deep clone objects anywhere else while performing other types of *updates* right now?

Andarist: Perhaps my question is - wow is this handled when a new property is being added to an object?

GoldingAustin: From my initial testing, only inserting into a new array index seems to require a deep clone. Moving nested objects around and into new properties works as expected. I will dig deeper into this to see if something is happening with new array indexes that are not happening with object updates. It might only be the root reference that is mutated, which would mean we could just shallow clone, but I want to confirm why the current behavior is happening before considering that.

Andarist: Thanks Austin. I appreciate it ❤️

GoldingAustin: After additional reading/testing, I found that we only needed to place an empty array/object in the new array index; the diff loop properly updates all changes after that. Objects automatically populate with the path provided to set, but arrays require the nested object/array to be set before using the path

Andarist: > Objects automatically populate with the path provided to set, but arrays require the nested object/array to be set before properly updating the path

Why? What is the difference between arrays and objects in this regard? 🤔

PR #5131, opened 2024-11-28
Add Svelte 5 to the allowed peer dependency range

lendle:
Svelte 5 is "almost" entirely backwards compatible with svelte 4. Breaking changes can be found [here](https://svelte.dev/docs/svelte/v5-migration-guide#Breaking-changes-in-runes-mode). 

`@xstate/svelte` only uses `onDestroy`, `readable`, and `Readable` from svelte, which are backwards compatible.

### DIFF
--- a/.changeset/long-ravens-yell.md
+++ b/.changeset/long-ravens-yell.md
@@ -1,0 +1,5 @@
+---
+'@xstate/svelte': patch
+---
+
+Bump svelte version to include svelte 5
--- a/.changeset/long-ravens-yell.md
+++ b/.changeset/long-ravens-yell.md
@@ -2,4 +2,4 @@
 '@xstate/svelte': patch
 ---
 
-Bump svelte version to include svelte 5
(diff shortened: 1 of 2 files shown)


lendle:
There are some failing svelte related tests, but I think this is coincidental, they're also failing in main. I created the branch from the last working commit on main (65088b6a595a0c69f312fec5f91916e1dbc81b46) and it works locally. So I think the ci build will succeed once main is fixed.

lendle:
Build seems to be working now if anyone is available to review. cc @Andarist

.changeset/long-ravens-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/svelte': patch
+---
+
+Bump svelte version to include svelte 5
```
Andarist, line 5:
```suggestion
Add Svelte 5 to the allowed peer dependency range
```

lendle:
🙇 Thanks!

PR #5132, opened 2024-12-03
[core] Add snapshot update feature to promise logic

davidkpiano:


packages/core/test/actorLogic.test.ts
```diff
@@ -464,6 +465,44 @@ describe('promise logic (fromPromise)', () => {
...
+    });
+    actor.start();
+
+    const res = await toPromise(actor);
+
+    expect(res).toEqual('done data');
+
+    expect(stuff).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({}),
```
Andarist, line 491: it feels like a better "assertion" could be put here, this one looks suspicious - maybe we could use `{ status: 'active' }` or smth like that?

Andarist: then it would make sense to include the status in all of those assertions~

packages/core/test/actorLogic.test.ts
```diff
@@ -464,6 +465,44 @@ describe('promise logic (fromPromise)', () => {
...
+        expect.objectContaining({ value: 'starting', context: undefined }),
+        expect.objectContaining({
+          value: 'loading',
+          context: { progress: 0.6 }
+        }),
+        expect.objectContaining({ value: 'finished', context: undefined }),
```
Andarist, line 497: I feel like this test case is too big. "can emit updates" is too broad and doesn't give me much insight into what is actually expected because it reads as "it works".

For instance, I noticed that an update with just one of those fields would wipe out the other one. I was wondering how deliberate that was and if we have a test for it. It turns out we do - but it's just so buried in this item here. It's hard to tell if this was intentional or accidental. I think it was intentional and that's fine - but my point stands: a test with so many details in it isn't particularly informative/explicit about its intentions

packages/core/src/actors/promise.ts
```diff
@@ -5,24 +5,42 @@ import {
...
+  StateValue
 } from '../types.ts';
 
-export type PromiseSnapshot<TOutput, TInput> = Snapshot<TOutput> & {
-  input: TInput | undefined;
-};
+export interface PromiseState {
+  value?: StateValue;
+  context?: MachineContext;
+}
```
Andarist, line 17: it would be nice to add some type-level tests for usage with discriminated unions, including tests that would exercise `.getSnapshot()`

davidkpiano:
Going a different direction in v6: `enq.emit(…)` should work in promise logic too

PR #5139, opened 2024-12-09
Required input for `spawn` when defined inside referenced actor

SandroMaglione:
Similar to #5055, this PR makes `input` required when using `spawn` and the `input` type  in the referenced actor is defined.

This prevents issues when forgetting to pass an input when using `spawn` (causing runtime errors).

```ts
  types: { input: {} as { value: number } }
});

  types: {} as { context: { ref: ActorRefFrom<typeof childMachine> } },
  context: ({ spawn }) => ({
    ref: spawn(
      childMachine,
      // Input is now required!
      { input: { value: 42 } }
    )
  })
});
```

### DIFF
--- a/packages/core/src/spawn.ts
+++ b/packages/core/src/spawn.ts
@@ -13,7 +13,8 @@ import {
   IsNotNever,
   ProvidedActor,
   RequiredActorOptions,
-  TODO
+  TODO,
+  type RequiredLogicInput
 } from './types.ts';
 import { resolveReferencedActor } from './utils.ts';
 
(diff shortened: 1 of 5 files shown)


packages/core/src/spawn.ts
```diff
@@ -123,5 +138,5 @@ export function createSpawner(
       actorRef.start();
     });
     return actorRef;
-  };
+  }) as Spawner<any>;
```
SandroMaglione, line 141: The return type errors without this bypass. I think this is expected given that typescript has not enough information to correctly type-check all possible configurations.

All tests are passing, so I think it shouldn’t be an issue.

davidkpiano (approved):

packages/core/test/spawn.types.test.ts
```diff
@@ -0,0 +1,49 @@
+import { ActorRefFrom, assign, createMachine } from '../src';
+
+describe('spawn inside machine', () => {
+  it('input is required when defined in actor', () => {
```
Andarist, line 4: we don't really check that it's required here, u just pass an input as an argument but the same test would pass without the patch introduced by this PR

SandroMaglione: Should this be inverted with a `@ts-expect-error` instead? Or is there another better way to test types?

PR #5143, opened 2024-12-12
Support generic sent event types (sendBack events) to callback-based actors

steveadams:
As mentioned in Discord, I don't think this is supported presently. With no typing, it's possible to place arbitrary events into the `sendBack` method. With this change (currently intended to be a suggestion; some things are broken), the types do work as I want them to in my use-cases so this seems to be on the right track.

One existing solution to this problem is to type the events like this:

```ts
export type SegmenterEvent =
  | {
      type: 'SEGMENT_CREATED';
      eventID: string;
      eventData: EventData;
    }
  | {
      type: 'SEGMENTER_FINISHED';
      segmenterID: string;
    }
  | {
      type: 'SEGMENTER_FAILED';
      segmenterID: string;
      error: unknown;
    };

export const segmenter = fromCallback(
  ({ input, sendBack }: { input: SegmenterInput; sendBack: (e: SegmenterEvent) => void }) => {
    const segmentEventData = async () => {
```

With this proposed change, the goal is to be able to type them like this:

```ts
export const segmenter = fromCallback<AnyEventObject, SegmenterInput, AnyEventObject, SegmenterEvent>(
  ({ input, sendBack }) => {
    const segmentEventData = async () => {
```

Both versions are quite verbose, but for callbacks with received, emitted, and sent events, it could be quite a bit cleaner.

Remaining:

- [ ] Determine if anyone else thinks this is a good idea
- [ ] Include changeset
- [ ] Figure out why typecheck is failing. It appears to be failing on main as well, but not sure if this is a caching/local issue of some sort. I'll look closer
- [ ] Review if the changes are made correctly, or only match my use case coincidentally while it might break others

### DIFF
--- a/packages/core/src/actors/callback.ts
+++ b/packages/core/src/actors/callback.ts
@@ -186,7 +186,7 @@ export function fromCallback<
   TEvent extends EventObject,
   TInput = NonReducibleUnknown,
   TEmitted extends EventObject = EventObject,
-  TSentEvent extends EventObject = AnyEventObject
+  TSentEvent extends EventObject = EventObject
 >(
   callback: CallbackLogicFunction<TEvent, TSentEvent, TInput, TEmitted>
 ): CallbackActorLogic<TEvent, TInput, TEmitted, TSentEvent> {
--- a/packages/core/src/actors/callback.ts
+++ b/packages/core/src/actors/callback.ts
@@ -27,13 +27,15 @@ export type CallbackSnapshot<TInput> = Snapshot<undefined> & {
 export type CallbackActorLogic<
   TEvent extends EventObject,
   TInput = NonReducibleUnknown,
-  TEmitted extends EventObject = EventObject
+  TEmitted extends EventObject = EventObject,
+  TSentEvent extends EventObject = AnyEventObject
 > = ActorLogic<
   CallbackSnapshot<TInput>,
   TEvent,
   TInput,
   AnyActorSystem,
-  TEmitted
+  TEmitted,
+  TSentEvent
 >;
 
 /**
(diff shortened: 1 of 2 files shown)


Andarist:
It requires a little bit of manual plumbing but the preferred approach to get what you want is this ([TS playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgZwKYwK5gDRwGZQQgDCAhgDbkBGpAxgNa4CCtM0ASqnrgMoB2pMMgAWEeAF98hEHABEAD2QxSMVLIDccAPRa4MAJ5hUyAFxwArADoAjAE5LABgBQTg0bgAFUlFR8YAUQA3X3gAXkQ9Q1QzWQAVAHkAfQ8mdn8AOVjZOEkAHwi3aLkmdPjYgAl-dkSE5NSMrJz1FzRMMAAKBCc4SKNTRElSZERunrhUYL9+rx8-IJDmnvFsUbo2KH6usfwICDMCIjJKGgZ20e3O8+24YD4wDBgzJDBvEJyV65ynq+3b+8eIi9ZgCWOtOHgADz8QQiMQQjB8eh8CAAdz4AD5cDMQvM-Oimj8cgBKOChfFbT5wIEhSxoPgAE0ulLGhRitRSaUysg+lPERMWlOpflpvkZFOZOjgAAEYMgALSoeRGVgKqCEKCE7asuQIpGovjczXEgWfHyYKB8ODtElkgYmsbLK5EnmOvmWWg+FSoACydGEt1QTJufECEHoRXFKCgtBieF2ht+dweZk6KFQ5DwxNJ+NTQoBaAzxJdKz5zSAA)):
```ts
import { setup, fromCallback, ActorRef, Snapshot } from "xstate"; // types: 5.19.0

type ParentEvent = { type: "TO_PARENT" } | { type: "ANOTHER_TO_PARENT" };

setup({
  types: {} as {
    events: ParentEvent;
  },
  actors: {
    foo: fromCallback(
      ({
        input: { parent },
      }: {
        input: { parent: ActorRef<Snapshot<unknown>, ParentEvent> };
      }) => {
        parent.send({
          type: "TO_PARENT",
        });
        parent.send({
          // @ts-expect-error
          type: "unknown",
        });
        return () => {};
      },
    ),
  },
}).createMachine({
  invoke: {
    src: "foo",
    input: ({ self }) => ({ parent: self }),
  },
});
```

Introducing a new generic introduces coupling between actor definitions in a way that is meaningless for non-callback actors. Other actor types can't send events to their parents without using a pattern like the above after all.

steveadams:
Ah, I see why this wasn't done already then. Thank you for explaining it to me. I'll stick with the manual plumbing!

PR #5175, opened 2025-01-22
[@xstate/store] v3

davidkpiano:


packages/xstate-store/src/store.ts
```diff
@@ -289,13 +289,18 @@ export const createStore: {
...
-  >(
-    ...args: CreateStoreParameterTypes<TContext, TEventPayloadMap, TTypes>
-  ): CreateStoreReturnType<TContext, TEventPayloadMap, TTypes>;
+  <TContext extends StoreContext, TEvent extends EventObject>(definition: {
+    context: TContext;
+    on: {
+      [K in TEvent['type']]?: StoreAssigner<
+        NoInfer<TContext>,
+        { type: K } & TEvent,
+        TEvent
```
Andarist, line 298: Isn't this one incorrect? You can't emit `TEvent`, that should be `TEmitted` or just `EventObject`

.changeset/big-schools-hang.md
```diff
@@ -0,0 +1,37 @@
...
+  // Context
+  {
+    count: number;
+  },
+  // Sent events
+  {
+    inc: {
+      by: number;
+    };
+  },
```
Andarist, line 18: I think the need for this explicit slot is a significant DX downgrade, but I guess you have already made up your mind on this. It feels like this now uses different set of tradeoffs than `setup` from `xstate` which is somewhat weird, even if both libraries are not strictly related.

packages/xstate-store/src/store.ts
```diff
@@ -253,85 +285,40 @@ export type TransitionsFromEventPayloadMap<
...
-      initialContextOrObject.context,
-      initialContextOrObject.on
-    );
-  }
-  return createStoreCore(initialContextOrObject, transitions);
+): CreateStoreReturnType<TContext, TEventPayloadMap, TEmitted> {
+  return createStoreCore(context, on);
 }
 
+export const createStore: {
```
Andarist, line 302: If you are going with "Type parameters should now be explicitly provided to `createStore` and `createStoreWithProducer`" then those overloads are not needed at all here and this could be just simplified/inlined

.changeset/big-schools-hang.md
Andarist, line 18: This payload map doesn't feel like an ergonomic explicit type argument - you won't need it anywhere else than at the `createStore` call. A union plays better with other utils you might want to build around this API as a user. Like, even a `logEvent` becomes much more cumbersome to type because you need to somehow extract the event type from your store instead of just using the same union type that you could use here

packages/xstate-store/test/store.test.ts
```diff
@@ -350,3 +351,101 @@ it('emitted events occur after the snapshot is updated', () => {
...
+        count: ctx.count - 1
+      })
+    }
+  });
+
+  store.send({ type: 'inc' });
+
+  expect(store.getSnapshot().context.count).toEqual(1);
+
+  await new Promise((resolve) => setTimeout(resolve, 10));
```
Andarist, line 384: you can import a shared util for this instead:
```ts
import { sleep } from '@xstate-repo/jest-utils';
```

.changeset/big-schools-hang.md
davidkpiano: You don't need this if you're not using emitted events; it's still inferred. But if you are, this:

```ts
createStore({
  types: {
    emitted: {} as ...
  },
  // …
});
```

is also not really a good DX.

davidkpiano: The payload map is mostly to still allow for inference when the type parameters are not provided.

davidkpiano: I'll change this since this is now inferred

expelledboy:
Oh no... 😭 https://github.com/statelyai/xstate/pull/5183

davidkpiano:
> Oh no... 😭 #5183
> 
> Figuring out how to pass the types around was difficult. But I think I made some improvements, even beyond the new getter API.
> 
> But when I saw how many changes where in this PR... I am going to get a beer.

Hey @expelledboy – computed properties are a great idea! I think you'll find the types even simpler to work with in this PR. Let's make it work. 🛠️

packages/xstate-store/src/fromStore.ts
```diff
@@ -97,17 +93,11 @@ export function fromStore<
...
-                { type: K } & TEventPayloadMap[K],
-                Cast<TTypes['emitted'], EventObject>
-              >;
+          [K in keyof TEventPayloadMap & string]: StoreAssigner<
+            NoInfer<TContext>,
+            { type: K } & TEventPayloadMap[K],
+            Cast<TTypes['emitted'], EventObject>
+          >;
         };
       } & { types?: TTypes }),
```
Andarist, line 102: should `types` property still be supported here?

Andarist: and similarly, should `fromStore` still support `initialContextOrObj` parameter?

davidkpiano: Fixed

.changeset/big-schools-hang.md
davidkpiano: We no longer use explicit type arguments

packages/xstate-store/test/fromStore.test.ts
```diff
@@ -2,34 +2,15 @@ import { createActor } from 'xstate';
...
-    });
-
-    actor.start();
-
-    actor.send({ type: 'inc', by: 8 });
-
-    expect(actor.getSnapshot().context.count).toEqual(50);
-  });
-
   it('creates an actor from store logic with input (object API)', () => {
```
Andarist, line 5: this is now the only API ;p
```suggestion
  it('creates an actor from store logic with input', () => {
```

Andarist (approved):

PR #5202, opened 2025-02-14
Introduce nix and just for reproducible CI locally

expelledboy:
The goal is to allow dev to quickly become productive, and be able to execute exactly what is run during CI before they submit a PR.

- **Nix Integration:**
  - Added `.envrc` for direnv integration with `use flake` to load the Nix environment.
  - Introduced `flake.nix` and `flake.lock` to define and lock dependencies.
  - Updated `.gitignore` to include `.direnv`, which is generated by Direnv and Nix workflow.

- **Just Task Runner:**
  - Added `justfile` to define reproducible commands for tasks such as:
    - `init`: Installs node dependencies if `node_modules` does not exist.
    - `lint`: Runs linting on the codebase.
    - `build`: Compiles the project.
    - `test`: Typechecks and tests the project.
    - `ci`: Comprehensive task that runs `init`, `build`, `lint`, and `test` stages.

- **Updates to Pre-commit Hook:**
  - Modified the `pre-commit` hook in `package.json` to run `jest -o` along with `lint-staged`. This ensures that only changed files are tested during a commit, optimizing the commit workflow.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ api/
 
 # pnpm
 .pnpm-lock.yaml
+
+# nix
+.direnv
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,59 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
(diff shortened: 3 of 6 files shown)


Andarist:
While those tools look useful and all… I think this is a maintenance burden for the core contributors (as far as I know none of us use those tools). This repo is using pretty standard JS tooling and reproducibility has not been a problem so far

PR #5203, opened 2025-02-14
[@xstate/store] Typestates failing test

davidkpiano:


Andarist:
It's not quite possible to make it work today. I reported the first underlying issue here https://github.com/microsoft/TypeScript/issues/61197 and proposed a fix for it here https://github.com/microsoft/TypeScript/pull/61185 

That's everything we can do about it (for now), as far as I know.

PR #5205, opened 2025-02-17
[@xstate/store] `useStore()` and other improvements

davidkpiano:
There is now a `useStore()` hook that allows you to create a local component store from a config object.

```tsx
import { useStore, useSelector } from '@xstate/store/react';

function Counter() {
  const store = useStore({
    context: {
      name: 'David',
      count: 0
    },
    on: {
      inc: (ctx, { by }: { by: number }) => ({
        ...ctx,
        count: ctx.count + by
      })
    }
  });
  const count = useSelector(store, (state) => state.count);

  return (
    <div>
      <div>Count: {count}</div>
      <button onClick={() => store.trigger.inc({ by: 1 })}>
        Increment by 1
      </button>
      <button onClick={() => store.trigger.inc({ by: 5 })}>
        Increment by 5
      </button>
    </div>
  );
}
```

---

Added `createStoreConfig` to create a store config from an object. This is an identity function that returns the config unchanged, but is useful for type inference.

```tsx
const storeConfig = createStoreConfig({
  context: { count: 0 },
  on: { inc: (ctx) => ({ ...ctx, count: ctx.count + 1 }) }
});

// Reusable store config:

const store = createStore(storeConfig);

// ...
function Comp1() {
  const store = useStore(storeConfig);

  // ...
}

function Comp2() {
  const store = useStore(storeConfig);

  // ...
}
```

---

The `createStoreWithProducer(config)` function now accepts an `emits` object (fix).

### DIFF
--- a/.changeset/modern-scissors-bow.md
+++ b/.changeset/modern-scissors-bow.md
@@ -1,0 +1,29 @@
+---
+'@xstate/store': minor
+---
+
+Added `createStoreConfig` to create a store config from an object. This is an identity function that returns the config unchanged, but is useful for type inference.
+
+```tsx
+const storeConfig = createStoreConfig({
+  context: { count: 0 },
+  on: { inc: (ctx) => ({ ...ctx, count: ctx.count + 1 }) }
+});
+
+// Reusable store config:
+
+const store = createStore(storeConfig);
+
+// ...
+function Comp1() {
+  const store = useStore(storeConfig);
+
+  // ...
+}
+
+function Comp2() {
+  const store = useStore(storeConfig);
+
+  // ...
+}
+```
--- a/packages/xstate-store/src/store.ts
+++ b/packages/xstate-store/src/store.ts
@@ -9,6 +9,7 @@ import {
   Store,
   StoreAssigner,
   StoreContext,
+  StoreConfig,
   StoreEffect,
   StoreInspectionEvent,
   StoreProducerAssigner,
@@ -68,6 +69,7 @@ function createStoreCore<
       TEmitted
     >;
   },
+  emits?: Record<string, (payload: any) => void>, // TODO: improve this type
   producer?: (
     context: NoInfer<TContext>,
     recipe: (context: NoInfer<TContext>) => void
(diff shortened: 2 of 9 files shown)


.changeset/modern-scissors-bow.md
```diff
@@ -0,0 +1,29 @@
...
+  const store = useStore(storeConfig);
+
+  // ...
+}
+
+function Comp2() {
+  const store = useStore(storeConfig);
+
+  // ...
+}
```
Andarist, line 28: Shouldn't one just use a reusable hook in this case?

packages/xstate-store/src/store.ts
```diff
@@ -364,13 +382,18 @@ export function createStoreWithProducer<
...
+      ) => void;
+    };
   }
 ): Store<
   TContext,
   ExtractEvents<TEventPayloadMap>,
   ExtractEvents<TEmittedPayloadMap>
 > {
-  return createStoreCore(config.context, config.on, producer);
+  return createStoreCore(config.context, config.on, config.emits, producer);
```
Andarist, line 396: perhaps `createStoreCore` should just accept `config`

packages/xstate-store/src/types.ts
```diff
@@ -113,6 +113,24 @@ export interface Store<
...
+  TEventPayloadMap extends EventPayloadMap,
+  TEmitted extends EventPayloadMap
+> = {
+  context: TContext;
+  emits?: {
+    [K in keyof TEmitted & string]: (payload: TEmitted[K]) => void;
+  };
+  on: {
+    [K in keyof TEventPayloadMap & string]: StoreAssigner<
+      NoInfer<TContext>,
```
Andarist, line 127: I'm guilty as charged but on principle `NoInfer` **shouldn't** be used within type definitions. It's a signature-level "annotation". But ye, well - I see how this type has both an inferrable and non-inferrable usage of `TContext` and having it here makes this reusable... so perhaps it's fine. I know of situations in which this kind of a thing creates problems though. David from `arktype` can confirm 😉. Our usage might be different from `arktype` though so we might be not as susceptible to those problems

Andarist (approved):

packages/xstate-store/src/store.ts
davidkpiano: Yeah, that makes sense

packages/xstate-store/src/types.ts
davidkpiano: I'll try without it

PR #5215, opened 2025-02-26
[@xstate/store] Add `store.transition(…)` API

davidkpiano:
Added `store.transition(state, event)` method which is a pure function that returns the next state and effects for a given state and event as a tuple, without actually updating the store. This is useful for computing state changes before committing them, or controlling the execution of effects.

Example:

```ts
const [nextState, effects] = store.transition(store.getSnapshot(), {
  type: 'increment',
  by: 1
});
```


packages/xstate-store/src/store.ts
```diff
@@ -131,6 +131,7 @@ function createStoreCore<
         }
       };
     },
+    transition: (state, event) => transition(state, event),
```
Andarist, line 134: nit:
```suggestion
    transition,
```

PR #5217, opened 2025-03-04
infered typehint for state machine properties 'on' and 'initial'

Syth-1:
Added a new flag 'scriptType' to state machine which provides different levels of typehint for states 'initial' and 'on' properties.

`strictType` can be:
- `"strict"` : add type-hints and errors if types do not match, this will break some features eg spreading vars into the machine,
- `"loose"` : adds type-hint but does not give errors when keys dont exist, as far as i know this keeps all compatibility, 
- `undefined` : no type-hint, the default behaviour if you do not provide a `strictType` param

I wasn't too sure how states inside states should work, so i made it scoped to the nested state you are currently on.

### DIFF
--- a/.changeset/many-tips-sit.md
+++ b/.changeset/many-tips-sit.md
@@ -1,0 +1,5 @@
+---
+'xstate': minor
+---
+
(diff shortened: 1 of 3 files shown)


davidkpiano:
Closing as stale; it's something we'll revisit in v6 though

PR #5220, opened 2025-03-07
[xstate/react] [xstate/store] Update React dependencies

davidkpiano:
Doing this cleanup since React 19 is out

### DIFF
--- a/packages/xstate-react/package.json
+++ b/packages/xstate-react/package.json
@@ -54,7 +54,7 @@
     "url": "https://github.com/statelyai/xstate/issues"
   },
   "peerDependencies": {
-    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0-0",
+    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0",
     "xstate": "workspace:^"
   },
   "peerDependenciesMeta": {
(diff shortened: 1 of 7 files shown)


packages/xstate-react/test/createActorContext.test.tsx
```diff
@@ -8,25 +8,6 @@ import {
 import { fireEvent, screen, render, waitFor } from '@testing-library/react';
 import { useSelector, createActorContext, shallowEqual } from '../src';
 
-const originalConsoleError = console.error;
-
-afterEach(() => {
-  console.error = originalConsoleError;
-});
-
-function checkConsoleErrorOutputForMissingProvider() {
```
Andarist, line 17: this was really just testing React's own behavior and that has changed, we don't need this

packages/xstate-react/test/useActorRef.test.tsx
```diff
@@ -611,10 +611,10 @@ describeEachReactMode('useActorRef (%s)', ({ suiteKey, render }) => {
     fireEvent.click(screen.getByText('Reload machine'));
 
     // while those numbers might be a little bit surprising at first glance they are actually correct
```
Andarist, line 613: the composition of functions changed between React versions. Previously it was smth like:
```ts
let children = renderWithHooks(...)
if (isSrictMode) {
  children = renderWithHooks(...)
}
```

where `renderWithHooks` had an extra loop based on `didScheduleRenderPhaseUpdateDuringThisPass`.

But now `renderWithHooks` itself is like
```ts
let children = callComponentInDEV(...);
if (didScheduleRenderPhaseUpdateDuringThisPass){
  children = renderWithHooksAgain(...)
}
if (isStrictMode) {
  children = renderWithHooksAgain(...)
}
```

Where `renderWithHooksAgain` has the infinite render loop protection

packages/xstate-react/package.json
```diff
@@ -54,7 +54,7 @@
     "url": "https://github.com/statelyai/xstate/issues"
   },
   "peerDependencies": {
-    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0-0",
```
Andarist, line 57: this wasn't really necessary - the users who might still be at some React 19 beta versions will now get peer dep mismatch warnings

Andarist (approved):

PR #5221, opened 2025-03-08
[xstate/store] Atoms

davidkpiano:
Added `createAtom()` for creating reactive atoms that can be combined with other atoms and stores:

- Create simple atoms with initial values:

  ```ts
  import { createAtom } from '@xstate/store';

  const countAtom = createAtom(0);
  countAtom.get(); // 0
  countAtom.set(1); // or use setter function: (prev) => prev + 1
  ```

- Subscribe to atom changes:

  ```ts
  countAtom.subscribe((value) => console.log(value));
  ```

- Combine multiple atoms:

  ```ts
  const nameAtom = createAtom('hello');
  const countAtom = createAtom(3);
  const combinedAtom = createAtom((read) =>
    read(nameAtom).repeat(read(countAtom))
  );
  combinedAtom.get(); // "hellohellohello"
  ```

- Seamlessly combine atoms with stores:

  ```ts
  const countAtom = createAtom(0);
  const nameStore = createStore({
    context: { name: 'David' }
    // ... store config
  });

  const combinedAtom = createAtom(
    (read) => read(nameStore).context.name + ` ${read(countAtom)}`
  );
  combinedAtom.get(); // "David 0"
  ```

Atoms automatically update when their dependencies change, making it easy to create derived state from both atoms and stores.


.changeset/tricky-comics-know.md
```diff
@@ -0,0 +1,49 @@
...
+  countAtom.subscribe((value) => console.log(value));
+  ```
+
+- Combine multiple atoms:
+
+  ```ts
+  const nameAtom = createAtom('hello');
+  const countAtom = createAtom(3);
+  const combinedAtom = createAtom((read) =>
+    read(nameAtom).repeat(read(countAtom))
```
Andarist, line 29: That `repeat` threw me off at first 😅 i thought it was some magic new method but that’s just the unwrapped value’s method :D

packages/xstate-store/src/types.ts
```diff
@@ -340,3 +346,18 @@ export type Selector<TContext, TSelected> = (context: TContext) => TSelected;
...
   get: () => TSelected;
 }
+
+export interface Readable<T> extends Subscribable<T> {
+  get(): T;
+}
+
+export interface Atom<T> extends Subscribable<T>, Readable<T> {
+  /** Gets the current value of the atom. */
+  get(): T;
```
Andarist, line 356: this method here is redundant because this interface already extends `Readable`
```suggestion
```

packages/xstate-store/src/types.ts
```diff
@@ -340,3 +346,18 @@ export type Selector<TContext, TSelected> = (context: TContext) => TSelected;
...
+  /** Sets the value of the atom using a function. */
+  set(fn: (prevVal: T) => T): void;
+  /** Sets the value of the atom. */
+  set(value: T): void;
+}
+
+export interface ReadOnlyAtom<T> extends Readable<T> {}
```
Andarist, line 363: Why do we need both of those interfaces? shouldn't we just use a single `ReadOnlyAtom`?

Andarist, line 363: nit: in TS we already have a builtin `Readonly` and `ReadonlyArray` types. I'd match spelling here and rename this to `ReadonlyAtom`

packages/xstate-store/src/types.ts
```diff
@@ -11,6 +11,9 @@ export type EnqueueObject<TEmittedEvent extends EventObject> = {
     [E in TEmittedEvent as E['type']]: (payload: Omit<E, 'type'>) => void;
   };
   effect: (fn: () => void) => void;
+  spawn: <T extends StoreConfig<any, any, any>>(
+    config: T
+  ) => Store<any, any, any>;
```
Andarist, line 16: I'd remove this method from this PR. From what I understand, this is unrelated to the atoms functionality and there are no new tests for this here either.

packages/xstate-store/src/types.ts
```diff
@@ -340,3 +346,18 @@ export type Selector<TContext, TSelected> = (context: TContext) => TSelected;
...
+  get(): T;
+}
+
+export interface Atom<T> extends Subscribable<T>, Readable<T> {
+  /** Gets the current value of the atom. */
+  get(): T;
+  /** Sets the value of the atom using a function. */
+  set(fn: (prevVal: T) => T): void;
+  /** Sets the value of the atom. */
+  set(value: T): void;
```
Andarist, line 360: For improved type-safety let's use non-bivariant functions instead of bivariant methods
```suggestion
  /** Sets the value of the atom using a function. */
  set: (fn: (prevVal: T) => T) => void;
  /** Sets the value of the atom. */
  set: (value: T) => void;
```

packages/xstate-store/src/types.ts
```diff
@@ -340,3 +346,18 @@ export type Selector<TContext, TSelected> = (context: TContext) => TSelected;
 export interface Selection<TSelected> extends Subscribable<TSelected> {
   get: () => TSelected;
 }
+
+export interface Readable<T> extends Subscribable<T> {
+  get(): T;
```
Andarist, line 351:
```suggestion
  get: () => T;
```

packages/xstate-store/src/types.ts
```diff
@@ -340,3 +346,18 @@ export type Selector<TContext, TSelected> = (context: TContext) => TSelected;
 export interface Selection<TSelected> extends Subscribable<TSelected> {
   get: () => TSelected;
 }
```
Andarist, line 348: This is basically redundant now but to avoid breaking the external types you can keep it as an alias:
```suggestion
export type Selection<TSelected> = Readable<TSelected>;
```

packages/xstate-store/src/types.ts
davidkpiano: I can't do this since overloads are only possible with `fn(...): ...` AFAIK

Andarist: You could declare it like this:
```ts
export interface Atom<T> extends Subscribable<T>, Readable<T> {
  set: {
    /** Sets the value of the atom using a function. */
    (fn: (prevVal: T) => T): void;
    /** Sets the value of the atom. */
    (value: T): void;
  };
}
```

buuuut that actually breaks things and the current types sort-of work. From types PoV, this `set` signature is somewhat weird given it both consumes and produces `T` in the first overload. So, all in all, I think we can't do any better here.

packages/xstate-store/src/types.ts
Andarist: makes sense

packages/xstate-store/src/atom.ts
```diff
@@ -0,0 +1,68 @@
...
+        return val;
+      }
+      const sub = atom.subscribe(recompute);
+      subs.set(atom, sub);
+      return val;
+    };
+
+    function recompute() {
+      current.value = getValue(read);
+      observers?.forEach((o) => o.next?.(current.value));
```
Andarist, line 30: This should help with some edge cases:
```suggestion
      const newValue = getValue(read);
      current.value = newValue;
      observers?.forEach((o) => o.next?.(newValue));
```

packages/xstate-store/src/atom.ts
```diff
@@ -0,0 +1,68 @@
...
+      if (subs.get(atom)) {
+        return val;
+      }
+      const sub = atom.subscribe(recompute);
+      subs.set(atom, sub);
+      return val;
+    };
+
+    function recompute() {
+      current.value = getValue(read);
```
Andarist, line 29: note: this might lead to oversubscribing to sources this atom no longer depends on cause there is no logic to unsubscribe sources that were not read during a recomputation

packages/xstate-store/src/atom.ts
```diff
@@ -0,0 +1,69 @@
...
+      const obs = toObserver(observerOrFn);
+      observers ??= new Set();
+      observers.add(obs);
+      return {
+        unsubscribe() {
+          observers?.delete(obs);
+          if (typeof valueOrFn === 'function') {
+            subs.forEach((sub) => sub.unsubscribe());
+            subs.clear();
+          }
```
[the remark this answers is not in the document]
davidkpiano: I honestly don't remember why I added that 😅

Andarist (approved):

PR #5223, opened 2025-03-10
[xstate/store] [xstate/react] Changeset for React 19 changes

davidkpiano:


.changeset/sharp-moose-matter.md
```diff
@@ -0,0 +1,6 @@
+---
+'@xstate/react': patch
+'@xstate/store': patch
+---
+
+Added React 19 as a peer dependency.
```
Andarist, line 6: It was already there 😉

davidkpiano: I thought I forgot to add it in the last PR 🤦

Andarist: What I meant is that it was already a valid peer dependency even before https://github.com/statelyai/xstate/pull/5220 . If anything this changeset *could* say that pre-release React 19 versions were removed from the peer dep range

PR #5230, opened 2025-03-15
[xstate/store] Fix types for emitting with no payload

davidkpiano:
The types for emitting events with no payload have been fixed so that the following code works:

```ts
const store = createStore({
  emits: {
    incremented: () => {}
  },
  on: {
    inc: (ctx, ev, enq) => {
      // No payload is expected
      enq.emit.incremented();
    }
  }
});
```

Previously, this would have been an error because the `incremented` event was expected to have a payload.

### DIFF
--- a/packages/xstate-store/src/types.ts
+++ b/packages/xstate-store/src/types.ts
@@ -6,15 +6,13 @@ export type ExtractEvents<T extends EventPayloadMap> = Values<{
 
 export type Recipe<T, TReturn> = (state: T) => TReturn;
 
-type HasNoPayload<T extends EventObject> = { type: T['type'] } extends T
-  ? keyof T extends 'type'
-    ? true
-    : false
-  : false;
+type AllKeys<T> = T extends any ? keyof T : never;
 
 type EmitterFunction<TEmittedEvent extends EventObject> = (
-  ...args: HasNoPayload<TEmittedEvent> extends true
-    ? []
+  ...args: { type: TEmittedEvent['type'] } extends TEmittedEvent
+    ? AllKeys<TEmittedEvent> extends 'type'
+      ? []
+      : [DistributiveOmit<TEmittedEvent, 'type'>?]
     : [DistributiveOmit<TEmittedEvent, 'type'>]
 ) => void;
 
--- a/packages/xstate-store/src/types.ts
+++ b/packages/xstate-store/src/types.ts
@@ -11,7 +11,7 @@ export type EnqueueObject<TEmittedEvent extends EventObject> = {
     [E in TEmittedEvent as E['type']]: IsEmptyObject<
       Omit<E, 'type'>
     > extends true
-      ? () => void
+      ? (_?: {}) => void
       : (payload: Omit<E, 'type'>) => void;
   };
   effect: (fn: () => void) => void;
--- a/packages/xstate-store/src/types.ts
+++ b/packages/xstate-store/src/types.ts
@@ -8,7 +8,11 @@ export type Recipe<T, TReturn> = (state: T) => TReturn;
 
 export type EnqueueObject<TEmittedEvent extends EventObject> = {
   emit: {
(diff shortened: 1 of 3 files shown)


packages/xstate-store/test/store.test.ts
```diff
@@ -339,6 +339,43 @@ it('emitted events occur after the snapshot is updated', () => {
...
 });
 
+it('events can be emitted with no payload', () => {
+  const spy = jest.fn();
+
+  const store = createStore({
+    emits: {
+      incremented: () => {},
+      decremented: () => {},
+      expectsPayload: (_: { payload: string }) => {}
```
Andarist, line 349: what about `_: { payload?: string }`? what should be accepted arguments for that?

davidkpiano: Ah let me add a test for that

davidkpiano: Just added a test:

```ts
  createStore({
    emits: {
      optionalPayload: (_: { payload?: string }) => {}
    },
    context: {},
    on: {
      inc: (_ctx, _ev, enq) => {
        enq.emit
          // @ts-expect-error
          .optionalPayload();

        enq.emit.optionalPayload({ payload: 'hello' });

        enq.emit.optionalPayload({});
      }
    }
  });
  ```

File: packages/xstate-store/test/store.test.ts [hunk 2]
```diff
@@ -339,6 +339,63 @@ it('emitted events occur after the snapshot is updated', () => {
...
+  createStore({
+    emits: {
+      optionalPayload: (_: { payload?: string }) => {}
+    },
+    context: {},
+    on: {
+      inc: (_ctx, _ev, enq) => {
+        enq.emit
+          // @ts-expect-error
+          .optionalPayload();
```
Andarist, line 389: so that was kind of my question - should this be an error if all fields are optional?

davidkpiano: I don't think so; what would be the correct type in this case?

Andarist: but is requiring an empty object to be passed in natural? fwiw you can achieve conditional parameter optionality like this: [TS playground](https://www.typescriptlang.org/play/?ts=5.9.0-dev.20250317#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwFsoBrEAHgBV4QAPDEVYAZ3gG94MBPABxAC54TDDCyoA5vAC+APgAUASgGyAUPHgA6TbDFMB7LrwEUA2gCIDIUwF0p1Og2bwKqtfAD88UpxyInd+ows5jyWLq7u8MZWYa4CxtxQnBA4UMBuAgDyBFgYlAA08MG8ptLR4XEJSSnAmdm5FAVFlqXK8vAAvNLwAG44WMAA3MrKYHhCgg4AjB2EJOT6IQKmAGIZGaZScvJDykxTCkN7jJOyrJLbu-sLhoWr61IXI2MYE4wATDNEpGTX-LdrpgG8AARrABKhkARgSAYJsDsMjsA3vDEcizhdUacQWD4NNzjtRqhxoiAMyfOY-DiLf7rIGgmDpeAQqEwuGPUkohwk074y6MbnsekCPEXIA)

packages/xstate-store/src/types.ts
```diff
@@ -8,7 +8,11 @@ export type Recipe<T, TReturn> = (state: T) => TReturn;
 
 export type EnqueueObject<TEmittedEvent extends EventObject> = {
   emit: {
-    [E in TEmittedEvent as E['type']]: (payload: Omit<E, 'type'>) => void;
+    [E in TEmittedEvent as E['type']]: IsEmptyObject<
+      Omit<E, 'type'>
+    > extends true
+      ? (_?: {}) => void
```
Andarist, line 14: this is wrong because you now permit strings, numbers and whatnot here - remember that `{}` means in TS basically `NonNullable<unknown>`, it **doesn't** mean "an empty object"

Andarist (approved):

Issue #5243, opened 2025-03-27
Bug: @xstate/svelte cjs export breaks tests


Shyrro:
### XState version

XState version 5

### Description

Hey there, 

I'm using `@xstate/svelte` and Vitest for my components. 

Vitest tries to use the `cjs` file apparently, instead of the esm one. In that same CJS export, there is a `require('svelte/store')` and many other requires from svelte.

Svelte however, only exports ESM, which breaks since the `require` looks for CJS modules. 

SvelteKit seems to target the right file, Vitest on the other hand targets the CJS one. 

Here is a repo with a reproduction: https://github.com/Shyrro/xstate-svelte-cjs-issue

Here is a screen of the error when running `pnpm test:unit`

<img width="725" alt="Image" src="https://github.com/user-attachments/assets/c2929e5f-8d9d-447a-822a-fc46b9b3b775" />

**ISSUE:** 

The `import` value is wrong in the `exports` field of the `package.json`. 
The `import`currently links to `./dist/xstate-svelte.cjs.mjs` which imports from `./xstate-svelte.cjs.js` which is a **CJS** file. 

**SOLUTION**

I was able to fix the issue by changing the `import` value to: `./dist/xstate-svelte.esm.js` in my `node_modules` locally, which is the real ESM. 

To still comply with CJS, there should be a `require` entry in the exports file that links to `./dist/xstate-svelte.cjs.js`. 

**ADDITIONAL NOTE**

Since Svelte only uses ESM, maybe there is no need for the CJS export here?

### Expected result

The `import` should link to the ESM file rather than the CJS one.

### Actual result

The `import` links to an ESM file that imports from a `CJS` file.

### Reproduction

https://github.com/Shyrro/xstate-svelte-cjs-issue

### Additional context

I can open the PR to fix this, just let me know if the solution is okay with you or if there is something i am missing. Thanks.

PR #5244, opened 2025-03-27
Make `@xstate/svelte` ESM-only

Andarist:
related to https://github.com/statelyai/xstate/issues/5243


Andarist:
This is somewhat blocked on migrating the repo to Vitest. We'd have to fight Jest to make it work here and I don't think it's worth it.

davidkpiano:
@Andarist Should we redo this one?

PR #5247, opened 2025-03-31
[@xstate/store] Fix distributive event payload types

davidkpiano:
This PR fixes type inference for discriminated union event types in the `trigger` and `emit` object. Previously, using `Omit` with union types would incorrectly combine event types, breaking type inference for discriminated unions. This has been fixed by introducing a `DistributiveOmit` type that correctly preserves the relationship between discriminated properties.

### DIFF
--- a/packages/xstate-store/src/types.ts
+++ b/packages/xstate-store/src/types.ts
@@ -8,7 +8,9 @@ export type Recipe<T, TReturn> = (state: T) => TReturn;
 
 export type EnqueueObject<TEmittedEvent extends EventObject> = {
   emit: {
-    [E in TEmittedEvent as E['type']]: (payload: Omit<E, 'type'>) => void;
+    [E in TEmittedEvent as E['type']]: (
+      payload: DistributiveOmit<E, 'type'>
+    ) => void;
   };
   effect: (fn: () => void) => void;
 };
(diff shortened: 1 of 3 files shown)


packages/xstate-store/src/types.ts
```diff
@@ -109,10 +111,10 @@ export interface Store<
    */
   trigger: {
     [E in TEvent as E['type'] & string]: IsEmptyObject<
-      Omit<E, 'type'>
+      DistributiveOmit<E, 'type'>
     > extends true
-      ? () => Omit<E, 'type'>
-      : (eventPayload: Omit<E, 'type'>) => void;
+      ? () => DistributiveOmit<E, 'type'>
```
Andarist, line 116: it's unrelated to this PR but this looks like a mistake, it should be:
```suggestion
      ? () => void
```

packages/xstate-store/src/types.ts
```diff
@@ -8,7 +8,9 @@ export type Recipe<T, TReturn> = (state: T) => TReturn;
 
 export type EnqueueObject<TEmittedEvent extends EventObject> = {
   emit: {
-    [E in TEmittedEvent as E['type']]: (payload: Omit<E, 'type'>) => void;
+    [E in TEmittedEvent as E['type']]: (
+      payload: DistributiveOmit<E, 'type'>
+    ) => void;
```
Andarist, line 13: this is definitely correct 👍

Andarist (approved):

PR #5250, opened 2025-04-03
[@xstate/store] Improve atom architecture

davidkpiano:
- Improved atom architecture with better dependency management (the diamond problem is solved!)
- Optimized recomputation logic to prevent unnecessary updates
- Added support for custom equality functions through `compare` option in `createAtom`, allowing fine-grained control over when atoms update:

  ```ts
  const coordinateAtom = createAtom(
    { x: 0, y: 0 },
    {
      // only update when x and y change
      compare: (prev, next) => prev.x === next.x && prev.y === next.y
    }
  );
  ```


packages/xstate-store/src/types.ts
```diff
@@ -365,14 +366,21 @@ export interface Readable<T> extends Subscribable<T> {
   get: () => T;
 }
 
-export interface Atom<T> extends Subscribable<T>, Readable<T> {
+interface BaseAtom<T> extends Subscribable<T>, Readable<T> {
+  dependencies: Set<BaseAtom<any>>;
+  dependents: Set<BaseAtom<any>>;
+  recompute: () => void;
+  status: AtomStatus;
```
davidkpiano, line 373: Need to find a better way to not expose these (help @Andarist)

packages/xstate-store/src/store.ts
```diff
@@ -210,22 +216,15 @@ function createStoreCore<
...
-              observer.next?.(nextSelected);
-            }
-          });
-        },
-        get: () => selector(this.getSnapshot().context)
-      };
-    }
+      return createAtom((read) => selector(read(store).context), {
+        compare: equalityFn
+      });
```
davidkpiano, line 221: Selected values can just be atoms; it works beautifully ✨

bowheart:
Hey, this is beautiful. Just wanted to give you a more official measuring stick for reactive competence than just the diamond problem.

### Measuring Stick

Many regard this repo as the gold standard for testing reactivity: https://github.com/milomg/js-reactivity-benchmark

There's a fork of it that Alien Signals uses here: https://github.com/transitive-bullshit/js-reactivity-benchmark

I just cloned XState and plugged this branch (at the last passing commit) into the original js-reactivity-benchmark. I see several "Assertion failed" logs in all the benchmarks, including `diamond`. For reference, here's my implementation:

```ts
import { ReactiveFramework } from '../util/reactiveFramework'
import { createAtom } from '../../xstate/packages/xstate-store/src/atom'

export const xStateStoreFramework: ReactiveFramework = {
  name: 'XState Store',
  signal: initialValue => {
    const s = createAtom(initialValue)
    return {
      write: v => s.set(v),
      read: () => s.get(),
    }
  },
  computed: fn => {
    const c = createAtom(fn)
    return {
      read: () => c.get(),
    }
  },
  effect: fn => createAtom(fn),
  withBatch: fn => fn(),
  withBuild: fn => fn(),
}
```

### Do You Need This?

No. This is technically for signals libs. While atomic libs typically have all the functionality of signals, they don't have to be valid signals implementations. In fact, Recoil and Nanostores are incapable of running these benchmarks. Jotai's latest version also can't run them, though earlier versions could get through all but the `triangle` benchmark.

But since your API is compatible, I would at least develop the initial implementation with this as an extra test suite. Making sure Zedux passes all these benchmarks has given us a lot of confidence that it's actually working and can handle all situations well. Unless I implemented it wrong, the fact that the `diamond` benchmark fails some assertions tells me this PR is probably missing something.

davidkpiano:
Thanks @bowheart! I'll take a look at these benchmarks and see how many of these we can get to pass.

packages/xstate-store/src/store.ts
```diff
@@ -149,21 +151,16 @@ function createStoreCore<
...
-      observers.add(observer);
-
-      return {
-        unsubscribe() {
-          return observers?.delete(observer);
-        }
-      };
+      return atom.subscribe(
+        // @ts-ignore help
+        observerOrFn
```
davidkpiano, line 162: @Andarist Can you help me figure out why this doesn't work?

Andarist: This boils down to how overloads work. Calling an overloaded function with an argument that has a union type of handled cases won't work (unless there is an overload that handles such an argument), see: https://www.typescriptlang.org/play/?ts=5.8.3#code/GYVwdgxgLglg9mABFApgZygCgIYCcDmAXIhrjGPgJTGnn4DcAUKJLAsulnkYmCALYAjFLmq8Bw3ExbR4SVBhwEaUMhUQAfcUJGVEAb0aJEuFFBC4k3JgF9GzcLPYB3XNgAO7kUp611Wvh1RAyMTMwt5Th9KW0YgA

In combination with how contextual parameter becomes a union of all possibilities when typed by an overloaded function prevents this from working: [TS playground](https://www.typescriptlang.org/play/?ts=5.8.3#code/GYVwdgxgLglg9mABDMYCmAnAFAQwwcwC5EBnKDFfASmIDc4YATAbgChRJYFlVNcDiYEAFsARphqJ6TNh2jwkKdNgD6eIqXKVEAH0RCxExAG8Avq1YQEZRHFEArYsdaJEwtFAAWcRsSgBPAAc0OGAeZTZTRABeExc3D29fRH5qGIA+ONdXAHoc7NcAPQB+eNclPnUqZkQ82zhAknjTABpWUzYgA)

You could just cast this but luckily we can also use `.bind`: [TS playground](https://www.typescriptlang.org/play/?ts=5.8.3#code/GYVwdgxgLglg9mABDMYCmAnAFAQwwcwC5EBnKDFfASmIDc4YATAbgChRJYFlVNcDiYEAFsARphqJ6TNh2jwkKdNgD6eIqXKVEAH0RCxExAG8Avq1aM0EADZ40iCAjKIcUOMOLHWiRMLRQABZwjMRQAJ4ADmhwwDzKbKZsrE5gLnCiAFZePn4BwaGIEdGx8ZiJiAC8Jrn+QSHEbh4AdHUFzaIojLjuwlQANKxJrEA)

I already pushed out this solution.

packages/xstate-store/src/store.ts
```diff
@@ -62,7 +63,7 @@ function createStoreCore<
   ) => NoInfer<TContext>
 ): Store<TContext, ExtractEvents<TEventPayloadMap>, TEmitted> {
   type StoreEvent = ExtractEvents<TEventPayloadMap>;
-  let observers: Set<Observer<StoreSnapshot<TContext>>> | undefined;
+  // let observers: Set<Observer<StoreSnapshot<TContext>>> | undefined;
```
Andarist, line 66:
```suggestion
```

packages/xstate-store/src/types.ts
```diff
@@ -385,7 +394,9 @@ export type AnyAtom = Atom<any>;
  * atom.set(43);
  * ```
  */
-export interface ReadonlyAtom<T> extends Readable<T> {}
+export interface ReadonlyAtom<T> extends BaseAtom<T>, Dependency, Subscriber {
+  update(): boolean;
```
Andarist, line 398: should `update` be on the atom? I thought it would be exclusive to computeds

packages/xstate-store/src/types.ts
```diff
@@ -365,14 +368,20 @@ export interface Readable<T> extends Subscribable<T> {
   get: () => T;
 }
 
-export interface Atom<T> extends Subscribable<T>, Readable<T> {
+interface BaseAtom<T> extends Subscribable<T>, Readable<T> {}
+
+export interface Atom<T> extends BaseAtom<T> {
```
Andarist, line 373: q: why `ReadonlyAtom` is a `Dependency` but an `Atom` isn't one? and similar question about `Subsciber`

packages/xstate-store/src/atom.ts
```diff
@@ -1,85 +1,221 @@
...
-        return val;
+    return computed(valueOrFn as () => T, options);
+  }
+
+  // Create plain object atom
+  const atom = {
+    currentValue: valueOrFn,
+    options,
+    // Dependency fields
+    subs: undefined as Link | undefined,
```
Andarist, line 51: I feel like this abuses casts and makes it harder to determine how those concepts compose here. The `atom`'s declaration should get a type annotation here (and the same thing applies to other objects that use this pattern in this file

packages/xstate-store/src/types.ts
davidkpiano: It is exclusive to computeds (`ReadonlyAtom` is the computed atom)

packages/xstate-store/src/types.ts
davidkpiano: Oversight; fixed

packages/xstate-store/src/atom.ts
davidkpiano: Cleaned this up; thanks for pointing it out

packages/xstate-store/src/atom.ts
```diff
@@ -1,85 +1,212 @@
...
+  return atom;
+}
+
+function computed<T>(
+  getter: (read: <U>(atom: Readable<U>) => U) => T,
+  options?: AtomOptions<T>
+): ReadonlyAtom<T> {
+  const computedAtom: ReadonlyAtom<T> = {
+    // Starts as undefined, but will be set by the computed getter (lazy)
+    currentValue: undefined as T,
```
Andarist, line 100: given this is a public property - shouldn't `computed` call `update` before returning `computedAtom` to the caller?

packages/xstate-store/src/types.ts
```diff
@@ -72,7 +74,8 @@ export interface Store<
   TEmitted extends EventObject
 > extends Subscribable<StoreSnapshot<TContext>>,
     InteropObservable<StoreSnapshot<TContext>>,
-    Readable<StoreSnapshot<TContext>> {
+    Readable<StoreSnapshot<TContext>>,
+    BaseAtom<StoreSnapshot<TContext>> {
```
Andarist, line 78: BaseAtom is already a Redable so this looks redundant to me:
```suggestion
    BaseAtom<StoreSnapshot<TContext>> {
```

packages/xstate-store/src/atom.ts
davidkpiano: It's not really supposed to be "public", not sure how to go about this though

packages/xstate-store/src/alien.ts
```diff
@@ -0,0 +1,696 @@
+// Adapted from Alien Signals
+// https://github.com/stackblitz/alien-signals/
+
+export interface Dependency {
+  /** @internal */
+  _subs: Link | undefined;
+  /** @internal */
+  _subsTail: Link | undefined;
+}
```
Andarist, line 9: I don't see those types being directly exposed outside of this package but be careful when marking everything as `@internal` on a type. That will become an empty interface in the emitted declaration and you don't want that.

packages/xstate-store/src/types.ts
```diff
@@ -365,14 +367,40 @@ export interface Readable<T> extends Subscribable<T> {
   get: () => T;
 }
 
-export interface Atom<T> extends Subscribable<T>, Readable<T> {
+export interface BaseAtom<T> extends Subscribable<T>, Readable<T> {}
+
+export interface InternalBaseAtom<T> extends Subscribable<T>, Readable<T> {
```
Andarist, line 372: Those `Internal*` types are leaking as part of the public exports of the package right now because we just blindly reexport everything contained in this file. Ideally this would get fixed

Andarist (approved):

PR #5264, opened 2025-04-29
[core] Add `actor.onStop(…)`

davidkpiano:
Added `onStop` handler to actors which allows registering callbacks that will be executed when an actor is stopped. This provides a way to perform cleanup or trigger side effects when an actor reaches its final state or is explicitly stopped.

```ts
const actor = createActor(someMachine);

actor.onStop(() => {
  console.log('Actor stopped');
});
```

### DIFF
--- a/.changeset/unlucky-news-drum.md
+++ b/.changeset/unlucky-news-drum.md
@@ -1,0 +1,13 @@
+---
+'xstate': minor
+---
+
+Added `onStop` handler to actors which allows registering callbacks that will be executed when an actor is stopped. This provides a way to perform cleanup or trigger side effects when an actor reaches its final state or is explicitly stopped.
+
+```ts
+const actor = createActor(someMachine);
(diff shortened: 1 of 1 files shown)


tnortman-jabra:
Any way for there to be some way to handle this in the state machine configuration itself?

For example:

```
setup({
  actions: {
    criticalAction: 'iWantToRunOnTeardownNoMatterWhat',
  },
}).createMachine({
  stop: 'iWantToRunOnTeardownNoMatterWhat',
});
```

Andarist:
@tnortman-jabra it already exists (sort of):
```ts
import { setup, fromCallback } from "xstate";

setup({
  actors: {
    criticalCleanup: fromCallback(() => () => {
      // do stuff
    }),
  },
}).createMachine({
  invoke: {
    src: "criticalCleanup",
  },
});
```

PR #5266, opened 2025-05-01
[@xstate/store] Async atoms

davidkpiano:
Added async atoms:

```ts
const atom = createAsyncAtom(async () => {
  const response = await fetch(`/api/something`);
  return response.json();
});

atom.subscribe((state) => {
  // Status can be 'pending', 'fulfilled', or 'rejected'
  if (state.status === 'fulfilled') {
    console.log(state.data);
  }
});
```

### DIFF
--- a/packages/xstate-store/src/atom.ts
+++ b/packages/xstate-store/src/atom.ts
@@ -6,6 +6,7 @@ import {
 } from './alien';
 import { toObserver } from './toObserver';
 import {
+  AsyncAtomOptions,
   Atom,
   AtomOptions,
   BaseAtom,
@@ -35,6 +36,30 @@ const {
 });
 
 let activeSub: Subscriber | undefined = undefined;
+
+type AsyncAtomState<Data, Error = unknown> =
+  | { status: 'pending' }
+  | { status: 'fulfilled'; data: Data }
+  | { status: 'rejected'; error: Error };
+
+export function createAsyncAtom<T>(
+  getValue: () => Promise<T>,
+  options?: AsyncAtomOptions<AsyncAtomState<T>>
+): ReadonlyAtom<AsyncAtomState<T>> {
+  const atom = createAtom<AsyncAtomState<T>>(() => {
+    getValue()
+      .then((value) => {
+        atom._update(() => ({ status: 'fulfilled', data: value }));
+      })
+      .catch((error) => {
+        atom._update(() => ({ status: 'rejected', error }));
+      });
+
+    return { status: 'pending' };
+  }, options) as InternalReadonlyAtom<AsyncAtomState<T>>;
+
+  return atom;
+}
 
 export function createAtom<T>(
   getValue: (read: <U>(atom: Readable<U>) => U) => T,
@@ -83,16 +108,32 @@ export function createAtom<T>(
           e.stop();
         }
       };
+    },
+    _update(getValue?: () => T): boolean {
+      const prevSub = activeSub;
+      const compare = options?.compare ?? Object.is;
+      activeSub = atom as InternalReadonlyAtom<T>;
+      startTracking(atom as InternalReadonlyAtom<T>);
+      try {
+        const oldValue = atom._snapshot;
+        const read = (atom: Readable<any>) => atom.get();
+        const newValue = getValue ? getValue() : getter(read);
+        if (oldValue === undefined || !compare(oldValue, newValue)) {
+          atom._snapshot = newValue;
+          return true;
+        }
+        return false;
+      } finally {
+        activeSub = prevSub;
+        endTracking(atom as InternalReadonlyAtom<T>);
+      }
     }
   };
 
   if (isComputed) {
     Object.assign<
       BaseAtom<T>,
-      Pick<
-        InternalReadonlyAtom<T>,
-        '_deps' | '_depsTail' | '_flags' | 'get' | '_update'
-      >
+      Pick<InternalReadonlyAtom<T>, '_deps' | '_depsTail' | '_flags' | 'get'>
     >(atom, {
       _deps: undefined,
       _depsTail: undefined,
@@ -107,37 +148,20 @@ export function createAtom<T>(
           link(atom, activeSub);
         }
         return atom._snapshot;
-      },
-      _update(): boolean {
-        const prevSub = activeSub;
-        const compare = options?.compare ?? Object.is;
-        activeSub = atom as InternalReadonlyAtom<T>;
-        startTracking(atom as InternalReadonlyAtom<T>);
-        try {
-          const oldValue = atom._snapshot;
-          const read = (atom: Readable<any>) => atom.get();
-          const newValue = getter(read);
-          if (oldValue === undefined || !compare(oldValue, newValue)) {
-            atom._snapshot = newValue;
-            return true;
-          }
-          return false;
-        } finally {
-          activeSub = prevSub;
-          endTracking(atom as InternalReadonlyAtom<T>);
-        }
       }
     });
   } else {
     Object.assign<BaseAtom<T>, Pick<Atom<T>, 'set'>>(atom, {
       set(valueOrFn: T | ((prev: T) => T)): void {
-        const compare = options?.compare ?? Object.is;
-        const value =
+        const prevValue = atom._snapshot;
+        const fn =
           typeof valueOrFn === 'function'
-            ? (valueOrFn as (prev: T) => T)(atom._snapshot)
-            : valueOrFn;
-        if (compare(atom._snapshot, value)) return;
-        atom._snapshot = value;
+            ? () => valueOrFn(atom._snapshot)
+            : () => valueOrFn;
+        atom._update(fn);
+        const nextValue = atom._snapshot;
+        const compare = options?.compare ?? Object.is;
+        if (compare(prevValue, nextValue)) return;
         const { _subs: subs } = atom;
         if (subs !== undefined) {
           propagate(subs);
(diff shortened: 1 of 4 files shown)


packages/xstate-store/src/atom.ts
```diff
@@ -39,6 +40,30 @@ const [
...
+  options?: AsyncAtomOptions<AsyncAtomState<T>>
+): ReadonlyAtom<AsyncAtomState<T>> {
+  const atom = createAtom<AsyncAtomState<T>>(() => {
+    getValue()
+      .then((value) => {
+        atom._update(() => ({ status: 'fulfilled', data: value }));
+      })
+      .catch((error) => {
+        atom._update(() => ({ status: 'rejected', error }));
+      });
```
Andarist, line 59: nit: I always felt this is more correct - because it "wraps" the most minimal thing with ur catch handler. We only want to catch errors coming from `getValue()` and not the ones coming from the `then`'s callback
```suggestion
      .then((value) => {
        atom._update(() => ({ status: 'fulfilled', data: value }));
      }, (error) => {
        atom._update(() => ({ status: 'rejected', error }));
      })
```

packages/xstate-store/src/atom.ts
```diff
@@ -39,6 +40,30 @@ const [
 
 let activeSub: Subscriber | undefined = undefined;
 
+type AsyncAtomState<Data, Error = unknown> =
+  | { status: 'pending' }
+  | { status: 'fulfilled'; data: Data }
+  | { status: 'rejected'; error: Error };
```
Andarist, line 46: we use `status: 'done' | 'error'` in XState. It would be nice to have this unified. It's worth noting though that in XState `status: "pending"` would be `status: "active"`

packages/xstate-store/test/atom.test.ts
```diff
@@ -585,3 +586,30 @@ it('Atom-specific properties should not be exposed', () => {
...
   store._depsTail;
 });
+
+describe('async atoms', () => {
+  it('async atoms should work (fulfilled)', async () => {
+    const atom = createAsyncAtom(async () => 'hello');
+
+    expect(atom.get()).toEqual({ status: 'pending' });
+
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
Andarist, line 596: nit: we have an util for this:
```ts
import { sleep } from '@xstate-repo/jest-utils';
```

packages/xstate-store/src/types.ts
```diff
@@ -382,14 +384,18 @@ export interface AtomOptions<T> {
   compare?: (prev: T, next: T) => boolean;
 }
 
+export interface AsyncAtomOptions<T> extends AtomOptions<T> {
+  signal?: AbortSignal;
```
Andarist, line 388: This was only added to the type here - but it's not handled by the implementation and there are no tests for this

packages/xstate-store/src/atom.ts
```diff
@@ -39,6 +40,30 @@ const [
...
+
+export function createAsyncAtom<T>(
```
Andarist, line 48: it would be good to have a test with a single async atom and multiple reads from it/subscribers, it should check that `getValue` is only called once (I think that's the expected result, right? 🤔 )

packages/xstate-store/src/types.ts
```diff
@@ -382,14 +384,18 @@ export interface AtomOptions<T> {
...
+
 export type AnyAtom = BaseAtom<any>;
 
 export interface InternalReadonlyAtom<T>
   extends InternalBaseAtom<T>,
     Dependency,
     Subscriber {
   /** @internal */
-  _update(): boolean;
+  _update(getValue?: () => T): boolean;
```
Andarist, line 398: This is already available on `InternalBaseAtom` and this interface extends from that. So you could remove it from here

packages/xstate-store/src/atom.ts
```diff
@@ -110,37 +151,20 @@ export function createAtom<T>(
...
-            ? (valueOrFn as (prev: T) => T)(atom._snapshot)
-            : valueOrFn;
-        if (compare(atom._snapshot, value)) return;
-        atom._snapshot = value;
+            ? () => valueOrFn(atom._snapshot)
+            : () => valueOrFn;
+        atom._update(fn);
+        const nextValue = atom._snapshot;
+        const compare = options?.compare ?? Object.is;
+        if (compare(prevValue, nextValue)) return;
```
Andarist, line 167: a good chunk of the `set`'s implementation is already handled by `._update`. Couldn't we rely on it more? 🤔

packages/xstate-store/src/atom.ts
davidkpiano: Yeah, I was thinking about this - I'm fine with unifying `fulfilling` -> `done` but I think it makes sense to have `status: 'active'` and `status: 'pending'` be semantically different

davidkpiano: [3bef5d8](https://github.com/statelyai/xstate/pull/5266/commits/3bef5d83c59c0d6b5b491ffc8292908c6cfc9c41)

packages/xstate-store/src/types.ts
davidkpiano: Yes, `getValue` is a way of "setting" a value of a derived atom, while tracking new dependencies. I'll unify

packages/xstate-store/src/atom.ts
davidkpiano: [357e414](https://github.com/statelyai/xstate/pull/5266/commits/357e414d250de3a3739062f00cb9b0c85fe21b40)

packages/xstate-store/src/atom.ts
```diff
@@ -39,6 +40,30 @@ const [
...
+type AsyncAtomState<Data, Error = unknown> =
+  | { status: 'pending' }
+  | { status: 'fulfilled'; data: Data }
+  | { status: 'rejected'; error: Error };
+
+export function createAsyncAtom<T>(
+  getValue: () => Promise<T>,
+  options?: AsyncAtomOptions<AsyncAtomState<T>>
+): ReadonlyAtom<AsyncAtomState<T>> {
+  const atom = createAtom<AsyncAtomState<T>>(() => {
```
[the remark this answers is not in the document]
davidkpiano: Yeah, I have a follow-up PR planned to refactor this actually

packages/xstate-store/src/atom.ts
```diff
@@ -39,6 +39,31 @@ const [
 
 let activeSub: Subscriber | undefined = undefined;
 
+type AsyncAtomState<Data, Error = unknown> =
+  | { status: 'pending' }
+  | { status: 'done'; data: Data }
+  | { status: 'error'; error: Error };
```
Andarist, line 45: This now doesn't match what is documented by the changeset (and by test titles)

packages/xstate-store/src/atom.ts
```diff
@@ -86,16 +111,32 @@ export function createAtom<T>(
...
+    },
+    _update(getValue?: () => T): boolean {
+      const prevSub = activeSub;
+      const compare = options?.compare ?? Object.is;
+      activeSub = atom as InternalReadonlyAtom<T>;
+      startTracking(atom as InternalReadonlyAtom<T>);
+      try {
+        const oldValue = atom._snapshot;
+        const read = (atom: Readable<any>) => atom.get();
+        const newValue = getValue ? getValue() : getter(read);
```
Andarist, line 123: this is a little bit tricky - does it mean a single implementation us reused by 2 quite different code paths? `getValue` is a dynamic updater~ wherear `getter` is somewhat static (configured when creating the atom).

This is fine but I think it deserves a code comment

packages/xstate-store/src/atom.ts
```diff
@@ -110,41 +151,23 @@ export function createAtom<T>(
...
-        const { _subs: subs } = atom;
-        if (subs !== undefined) {
-          propagate(subs);
-          processEffectNotifications();
+        if (
+          atom._update(
+            typeof valueOrFn === 'function'
+              ? () => (valueOrFn as (prev: T) => T)(atom._snapshot)
+              : () => valueOrFn
+          )
```
Andarist, line 164: it would be better to push the function resolving to the `atom._update` itself, this else branch here introduced a redundant function wrapper for the value, one that is going to be called synchronously *anyway*

packages/xstate-store/src/atom.ts
```diff
@@ -39,6 +39,31 @@ const [
...
+  | { status: 'error'; error: Error };
+
+export function createAsyncAtom<T>(
+  getValue: () => Promise<T>,
+  options?: AtomOptions<AsyncAtomState<T>>
+): ReadonlyAtom<AsyncAtomState<T>> {
+  const atom = createAtom<AsyncAtomState<T>>(() => {
+    getValue().then(
+      (value) => {
+        atom._update(() => ({ status: 'done', data: value }));
```
Andarist, line 54: nit:
```suggestion 
      (data) => {
        atom._update(() => ({ status: 'done', data }));
```

packages/xstate-store/src/atom.ts
davidkpiano: [c54de6b](https://github.com/statelyai/xstate/pull/5266/commits/c54de6bde210c577c414b9d08f3ce78fd47b22bc)

packages/xstate-store/src/atom.ts
davidkpiano: Fixed the changeset

packages/xstate-store/src/atom.ts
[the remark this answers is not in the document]
davidkpiano: Added

Andarist (approved):

Issue #5178, opened 2025-01-23
Bug: restoring state from serialized getPersistedSnapshot() breaks history behaviour


seanlennaerts:
### XState version

XState version 5

### Description

I think I've discovered unexpected behaviour with restoring from a persisted snapshot that has been serialized and the machine's historyValue not being restored correctly.

Take the following sample checkout machine:

![Image](https://github.com/user-attachments/assets/3914bd0b-0c2a-44fb-a913-1c5340777830)

[Link to stately](https://stately.ai/registry/editor/1c50a33b-67ca-4627-9dfb-92c44bdc289b?machineId=d3b775ba-58f9-4aa3-aee4-4aa9a7e76a0d&mode=design)

Say we go through the following path:

https://github.com/user-attachments/assets/63a68ed7-e080-41f3-975a-7db970a77646

So far this looks good. The History node successfully remembers the previous "OrderSummary" state by returning to "OrderSummary.Ready" the second time around.

The problem appears to only happen if we persist the state of the machine, serialize it, and restore it. The serializing bit is important because without serializing the snapshot it actually works as expected.

Take this minimal recreation: https://github.com/seanlennaerts/xstate-history-bug

If you run it with `npm run test` you'll see that the snapshots for original actor and the restored actor do not match:

![Image](https://github.com/user-attachments/assets/e3ba7630-c72b-474c-938d-dacca018ea8e)

The first log is the snapshot from the original actor and the historyValue is an instance of StateNode. The second log is a snapshot of a second actor that restored the serialized state of the first actor and the historyValue is a pojo - I'm thinking this may have something to do with the bug.

The code that is important:

```js
const actor = createActor(machine).start()
// send events that create a historyValue

// simulate persist
const snapshot = actor.getPersistedSnapshot()
const serialized = JSON.stringify(snapshot)

// restore persisted snapshot
const parsed = JSON.parse(serialized)
const secondActor = createActor(machine, { snapshot: parsed }).start()
secondActor.send({ type: 'update-shipping-address' })
```

After the "update-shipping-address" event and the promiseActor's "done" event emitted it should return to "OrderSummary.Ready" as we previously simulated. But instead it actually goes back to the initial state "ShippingAddress". The assertion at the end of the file fails.

Interestingly enough if we skip the serialization and restored from `snapshot` directly then the assertion at the end of the file succeeds.

### Expected result

After restoring from serialized persisted snapshot the machine should be able to go back to a history node.

### Actual result

[...]

PR #5269, opened 2025-05-06
[core]: bugfix restoring historyValue | feature reviveHistoryValue

chladog:
This is a fix for issue here: https://github.com/statelyai/xstate/issues/5178
featuring new helper functions to restore _StateNode class instances deeply.

### DIFF
--- a/packages/core/src/StateMachine.ts
+++ b/packages/core/src/StateMachine.ts
@@ -584,13 +584,71 @@ export class StateMachine<
       children[actorId] = actorRef;
     });
 
+    function resolveStateNodeFromIdOrPath(
+      idOrPath: string,
+      root: StateNode<TContext, TEvent>
+    ) {
+      try {
+        return root.machine.getStateNodeById(idOrPath);
+      } catch {
+        try {
+          return getStateNodeByPath(root, idOrPath.split('.'));
+        } catch {
+          if (isDevelopment) {
+            console.warn(
+              `Could not resolve StateNode for id/path: ${idOrPath}`
+            );
+          }
+          return idOrPath;
+        }
+      }
+    }
+
+    function reviveHistoryValue(
+      historyValue: HistoryValue<TContext, TEvent>,
+      root: StateNode<TContext, TEvent>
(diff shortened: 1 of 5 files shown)


packages/core/test/history.test.ts
```diff
@@ -1368,3 +1369,142 @@ describe('multistage history states', () => {
     });
   });
 });
+
+// --- Tests for reviveHistoryValue ---
+describe('revive history states', () => {
+  const {
+    restoreSnapshot,
+    getInitialSnapshot
+  } = require('../src/StateMachine');
```
Andarist, line 1378: 1. we shouldnt use `require` in tests
2. we should focus on testing behaviors thought public facing APIs alone

packages/core/src/StateMachine.ts
```diff
@@ -584,13 +584,95 @@ export class StateMachine<
...
+          }
+          return idOrPath;
+        }
+      }
+    }
+
+    function reviveHistoryValue(
+      historyValue: unknown,
+      root: StateNode<any, any>,
+      seen = new WeakSet()
```
Andarist, line 610: what do we need a recursion for? a `historyValue` value is defined as `Record<string, Array<StateNode<TContext, TEvent>>>` This only requires a single iteration - for each key we should "resolve" all (previously) serialized array items. Those should point to specific state nodes and that's it.

Note that the `historyValue` should get its designated serialization path in `getPersistedSnapshot` within `packages/core/src/State.ts`

packages/core/test/history.test.ts
chladog: 1. was autoimport mistake
2. I have refactored tests to public only API

packages/core/src/StateMachine.ts
chladog: I have refactored the implementation, removed the recursion, handle invalid state (get rid of it during reconstruction) and featured serialization of historyValue in State, as if I understood correctly, we need only StateNode IDs for correct reconstruction - StateNode itself is not holding any content that needs to be persisted.

davidkpiano (approved):

PR #5279, opened 2025-05-13
Fixed JSDoc `@link` references

Andarist:
This makes those references clickable in the IDE and they now correctly auto-link in the tooltips.


packages/core/src/createActor.ts
```diff
@@ -11,6 +11,16 @@ import { reportUnhandledError } from './reportUnhandledError.ts';
 import { symbolObservable } from './symbolObservable.ts';
 import { AnyActorSystem, Clock, createSystem } from './system.ts';
 
+// those are needed to make JSDoc `@link` work properly
+import type {
+  fromObservable,
```
Andarist, line 16: We can't freely `{@link fromObservable}` in the JSDoc without that being available in the scope.

davidkpiano (approved):

PR #5287, opened 2025-05-18
[core] Move graph/test utils to core

davidkpiano:
The graph and model-based testing utilities from `@xstate/graph` (and `@xstate/test` previously) were moved to the core `xstate` package.

```ts
import { getShortestPaths } from 'xstate/graph';

  // ...
});

const paths = getShortestPaths(machine, {
  fromState: 'a',
  toState: 'b'
});
```


davidkpiano:
> What's the plan for `@xstate/graph` package? Do you want to deprecate it completely? Should we remove its source code in this PR?

Not sure if removing the source code would make it more difficult for devs using `@xstate/graph` to reference code 🤔

Andarist:
The source code will still be here... in a way. I feel like having those files duplicates in the repo will be super annoying when searching through the repo, doing changes etc as it will be easy to accidentally change the wrong source file.

PR #5299, opened 2025-06-15
Actor select

Uniqen:
The select function in @xstate/store is very useful. This PR adds a similar select function to the actor. This is to make the transition from @xstate/store to an xstate machine easier. I've used the same naming of types as in @xstate/store to make it consistent.

I hope you find the PR useful! Thanks for an awesome library!

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1911,6 +1911,12 @@ export interface Subscription {
   unsubscribe(): void;
 }
 
+export type Selection<TSelected> = Readable<TSelected>;
+
+export interface Readable<T> extends Subscribable<T> {
+  get: () => T;
+}
+
 export interface InteropObservable<T> {
   [Symbol.observable]: () => InteropSubscribable<T>;
 }
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1996,6 +1996,10 @@ export interface ActorRef<
       emitted: TEmitted & (TType extends '*' ? unknown : { type: TType })
     ) => void
   ) => Subscription;
+  select<TSelected, TSnapshot>(
+    selector: (snapshot: TSnapshot) => TSelected,
+    equalityFn?: (a: TSelected, b: TSelected) => boolean
+  ): Readable<TSelected>;
 }
 
 export type AnyActorRef = ActorRef<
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -1911,6 +1911,12 @@ export interface Subscription {
   unsubscribe(): void;
 }
 
+export type Selection<TSelected> = Readable<TSelected>;
```
Andarist: this is unused, do we need this alias? it probably could be dropped

packages/core/src/types.ts
```diff
@@ -1990,6 +1996,10 @@ export interface ActorRef<
       emitted: TEmitted & (TType extends '*' ? unknown : { type: TType })
     ) => void
   ) => Subscription;
+  select<TSelected, TSnapshot>(
```
Andarist: `TSnapshot` is already available in the scope here (it's an existing type parameter of `ActorRef`), so u dont need it here - only `TSelected` is needed

packages/core/src/types.ts
Uniqen: Yes, good catch! Removing it.

packages/core/src/types.ts
Uniqen: I get compilation error when I remove TSnapshot and then run `pnpm typecheck`, see [GitHub Actions](https://github.com/statelyai/xstate/actions/runs/15810540587/job/44561254051?pr=5299#step:4:46). Not sure on how to resolve it. Reverting to use TSnapshot for now.

I looked into adding a generic for TSnapshot on Actor, see below. But that ended up in a lot of changes not really in the scope of this PR. @Andarist any suggestions?

```typescript
export class Actor<TLogic extends AnyActorLogic, TSnapshot extends SnapshotFrom<TLogic>>
  implements
    ActorRef<SnapshotFrom<TLogic>, EventFromLogic<TLogic>, EmittedFrom<TLogic>>
{
  ....
}
```

Uniqen: @Andarist Any idea how to resolve this without a major refactor? Is it required in order to merge the PR?

Uniqen:
@Andarist @davidkpiano 
Hi! I just wanted to check in on this PR to see if it’s relevant to you or if there’s anything I can do to help move it forward? I’m happy to make changes if needed.

PR #5302, opened 2025-06-19
[@xstate/store] Add `useAtom(atom)` + fix vitest config

davidkpiano:
This PR adds the `useAtom` hook for reading the value of an atom or selecting a value from the atom.

It is meant for the common use-case where the entire value of the atom (usually granular data) is needed.

```tsx
const atom = createAtom(0);

const Component = () => {
  const count = useAtom(atom);

  return (
    <>
      <div onClick={() => atom.set((c) => c + 1)}>{count}</div>
      <button onClick={() => atom.set(0)}>Reset</button>
    </>
  );
};
```

With selectors:

```tsx
const atom = createAtom({ count: 0 });

const Component = () => {
  const count = useAtom(atom, (s) => s.count);

  return <div>{count}</div>;
};
```

### DIFF
--- a/packages/xstate-store/src/react.ts
+++ b/packages/xstate-store/src/react.ts
@@ -8,7 +8,7 @@ import {
   ExtractEvents,
   Readable,
   AnyAtom,
-  SnapshotFromAtom
+  Atom
 } from './types';
 import { createStore } from './store';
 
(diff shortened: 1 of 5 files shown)


packages/xstate-store/src/react.ts
```diff
@@ -94,3 +100,48 @@ export const useStore: {
...
+ * @param compare An optional function which compares the selected value to the
+ *   previous value
+ */
+export function useAtom<TAtom extends AnyAtom>(
+  atom: TAtom
+): SnapshotFromAtom<TAtom>;
+export function useAtom<TAtom extends AnyAtom, T>(
+  atom: TAtom,
+  selector: (snapshot: SnapshotFromAtom<TAtom>) => T,
+  compare?: (a: T | undefined, b: T) => boolean
```
Andarist: To make it easier to work with this we should aim for this signature
```suggestion
  compare?: (a: T, b: T) => boolean
```

The first selected item doesn't ever have to be compared against the previous value - we know the previous doesn't exist

packages/xstate-store/src/react.ts
```diff
@@ -94,3 +100,48 @@ export const useStore: {
...
+ *
+ * @param atom The atom, created from `createAtom(…)`
+ * @param selector An optional function which takes in the `snapshot` and
+ *   returns a selected value
+ * @param compare An optional function which compares the selected value to the
+ *   previous value
+ */
+export function useAtom<TAtom extends AnyAtom>(
+  atom: TAtom
+): SnapshotFromAtom<TAtom>;
```
Andarist: for such easy utils, I feel this is just an anti-pattern - we are using it in other places because the alternative is just annoying but when u can avoid surfacing conditional types in the public signature, you should avoid it. Conditional types stay deferred at the generic level and makes wrapping functions using them much harder.

In here, we should strive for a simple:
```ts
export function useAtom<T>(
  atom: Atom<T>
): T
export function useAtom<T, S>(
  atom: Atom<T>,
  selector: (snapshot: T) => S,
  compare?: (a: S, b: S) => boolean
): S
```

But I guess we are using the proposed pattern already in some places so it might be more consistent to just keep it 🤷

Andarist (approved):

packages/xstate-store/src/react.ts
davidkpiano: Oh good call; I was curious about which one was better.

davidkpiano: Made this change

PR #5305, opened 2025-06-22
[@xstate/store] Undo redo

davidkpiano:
Added undo/redo functionality to XState Store via the `undoRedo` higher-order store logic:

- Adds `undo` and `redo` events to stores
- Supports grouping related events into transactions using `transactionId`
- Maintains event history for precise state reconstruction

- Automatically clears redo stack when new events occur

```ts
import { undoRedo } from '@xstate/store/undo';

const store = createStore(
  undoRedo({
    context: { count: 0 },
    on: {
      inc: (ctx) => ({ count: ctx.count + 1 }),
      dec: (ctx) => ({ count: ctx.count - 1 })
    }
  })
);

store.trigger.inc();
// count: 1
store.trigger.inc();
// count: 2
store.trigger.undo();
// count: 1
store.trigger.undo();
// count: 0
store.trigger.redo();
// count: 1
store.trigger.redo();
// count: 2
```

### DIFF
--- a/packages/xstate-store/src/undo.ts
+++ b/packages/xstate-store/src/undo.ts
@@ -16,6 +16,62 @@ type UndoEvent<TEvent extends EventObject> = {
   transactionId?: string;
 };
 
+/**
+ * Creates store logic with undo/redo functionality.
+ *
+ * It maintains an event history and allows reverting to previous states by
+ * replaying events from the beginning up to a certain point.
+ *
+ * @example
+ *
+ * ```ts
+ * // Basic usage - each event is its own transaction
+ * const store = createStore(
+ *   undoRedo({
+ *     context: { count: 0 },
+ *     on: {
+ *       inc: (ctx) => ({ count: ctx.count + 1 }),
+ *       dec: (ctx) => ({ count: ctx.count - 1 })
+ *     }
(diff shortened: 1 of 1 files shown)


packages/xstate-store/src/undo.ts
```diff
@@ -0,0 +1,167 @@
...
+          // Remove all events with the same transaction ID
+          while (
+            events.length > 0 &&
+            events[events.length - 1].transactionId === lastTransactionId
+          ) {
+            const event = events.pop()!;
+            eventsToUndo.unshift(event);
+            undoStack.push(event);
+          }
+        }
```
Andarist: nit: this can probably be squeezed into something like:
```ts
          while (true) {
            const event = events.pop()!;
            eventsToUndo.unshift(event);
            undoStack.push(event);
            if (lastTransactionId === undefined || !events.length || events[events.length - 1].transactionId !== lastTransactionId) {
              break;
            }
          }
```

Andarist (approved):

packages/xstate-store/src/undo.ts
davidkpiano: Nice, made this change :+1:

Issue #4754, opened 2024-02-21
Bug: `@xstate/vue` can't be used reliably to create a shared machine across components


IonianPlayboy:
### XState version

XState version 5

### Description

I need to use the same instance of a machine across different components, because the machine serves as the source of truth for the state of a core business logic in my app. 
The most idiomatic way of doing this in Vue (to my knowledge) would be to create a custom hook wrapping the base `useMachine` invocation with [`createSharedComposable`](https://vueuse.org/shared/createSharedComposable/), and then to use this custom hook everywhere I need to have this shared instance.

This is sadly not working currently in a couple of scenarios, because the machine is stopped prematurely and won't restart afterwards. It's really an issue for me, since it means that currently I can't use `@xstate/vue` as soon as my needs start to grow too big for its scope.

### Expected result

I have created a reproduction to highlight the issue, with the expected behavior showing on the right panel (named `FixedPageViews`) : 

- Open the repro, landing on the "Home page".  
<img width="278" alt="image" src="https://github.com/statelyai/xstate/assets/24679993/3b5800e4-1cba-4093-b515-2a27dfdd0bae">

Since our machine only tracks page views for the child routes of the `PagesToDiscover` page, currently `FixedPageViews` should only display its name, with no counter of page views below.
- Click on one of the link at the bottom of the page that is not "home". 
<img width="278" alt="image" src="https://github.com/statelyai/xstate/assets/24679993/6a066e66-9bb6-46dc-8357-b5ed5b617669">

`FixedPageViews` should now be displaying that the page you just visited has been viewed once.
- Click on other links that are not "home" any number of times.
<img width="278" alt="image" src="https://github.com/statelyai/xstate/assets/24679993/9241990f-c50a-4766-be1c-ca7ecc305670">

`FixedPageViews`should be displaying the correct count of views for each page you clicked.

- Click on the "home" link to go back to the home page
<img width="278" alt="image" src="https://github.com/statelyai/xstate/assets/24679993/ddddf0de-bcf9-4b16-8f91-0e1e74b0ed22">

`FixedPageViews`should correctly have reset to its initial state, since we unmounted the `PagesToDiscover` component before going back to `Home`.

Since it is back to its initial state, you should be able to restart this flow from the start and observe the same results.

### Actual result

The current behavior is showed in the repro on the left panel (named `PageViews`) : 

- Open the repro, landing on the "Home page".  
[...]

PR #4288, opened 2023-09-16
Update `@xstate/vue` hooks

davidkpiano:

Issue #3786, opened 2023-01-25
Bug: State from useActor (Vue) not fully reactive


ottopaulsen:
### Description

[This repo](https://github.com/ottopaulsen/failing-xstate-vue) shows a problem I am having with xstate and Vue.

There are two state machines, one parent and one child. The child is started from the parent using spawn. The parent is started using useMachine.

The parent starts in loading, then goes to runningChild, where the child machine is spawned on entry. The child machine starts in ready, and on a button click it goes to doing, where it is invoking a promise. When the promise is resolved, the onDone takes the machine to the final complete state, where a CHILD_COMPLETE event is sent to the parent. The parent then goes to the childComplete state and from there further to the complete state.

On the screen, I am displaying the states from the two machines. The parent state I am getting from the useMachine call, and the child state I am getting from the useActor call.

**The problem** is that when the child has completed, and the parent is in the complete state, the state displayed on the screen for the child is still doing. It does not show the complete state from the child.

However, if I read the childs state through the children property of the parent machine, I can see that the state is complete as expected.

Reading the state through the children property is awkward and fragile, so I believe the correct way is reading the state from useActor, but this may be where I am doing wrong. How else could I do this?

This is a problem in my real project because I am displaying text on the screen based on the state of the machines, and one text is hanging because the child machines state is not read correctly.

I have tried reading the state via the children, but this is also giving me similar problems, but I have not yet been able to isolate the other problems as good as this one.

So I need help. Is this a bug in @xstate/vue, or am I doing something wrong?

### Expected result

I expect the state retrieved from useActor to show the `complete` state.

### Actual result

The state retrieved from useActor shows the `doing` state

### Reproduction

https://github.com/ottopaulsen/failing-xstate-vue

### Additional context

See code in GitHub repo.

Andarist:
> So I need help. Is this a bug in @xstate/vue, or am I doing something wrong?

It might be a problem with `@xstate/vue`. This package was written by the community members and I lack the expertise to address this issue. We use `shallowRef` in the implementation - which makes me think that it's part of the problem because we don't exactly wrap the whole state recursively in proxies. We'd appreciate help with fixing this.

auvred:
[...]

PR #5311, opened 2025-06-26
[@xstate/vue] Run actor during setup, not mounted.

negezor:
I started migrating the chat logic in a Vue application using XState. After some time, I encountered an issue where the actor does not work during the execution of the `setup` hook. This makes it impossible to set any initial state based on external data. Here's a simplified example for context:

```ts
const { send, snapshot } = useActor({
  // ...
  context: {
    after: null,
    before: null,
  },
  // ...more logic
  states: {
    idle: {
      on: {
        JUMP_TO_CURSOR: {
          target: 'loadingAroundTop',
          actions: assign({
            after: ({ event }) => event.cursor,
            before: null,
          }),
        },
      }
    }
  }
});

const { data: conversation } = await useFetchConversation({
  variables: () => ({
    id: route.params.id,
  }),
});

if (conversation.value.lastSeenMessageCursor) {
  send({
    type: 'JUMP_TO_CURSOR',
    cursor: conversation.value.lastSeenMessageCursor,
  });
} else {
  send({
    type: 'INIT_WITHOUT_CURSOR',
  });
}

const { data: messagesData } = await useFetchConversationMessages({
  variables: () => ({
    // Variables depend on machine context
    after: snapshot.value.context.after,
    before: snapshot.value.context.before,
  }),
});
```

**Expected behavior:**

```ts
console.log(snapshot.value.value); // 'loadingAroundTop'
console.log(snapshot.value.context.after); // <cursor>
console.log(snapshot.value.context.before); // null
```

**Actual behavior:**

```ts
console.log(snapshot.value.value); // 'idle'
console.log(snapshot.value.context.after); // null
console.log(snapshot.value.context.before); // null
```

After looking into the hook implementation, I realized the problem is that the actor is only started inside the `onMounted()` hook. This creates an unnecessary limitation because:

* It makes the actor unusable during SSR;
* It prevents reusing logic immediately in the `setup()` function;
* And it forces awkward workarounds just to interact with machine state early.

I don't see any strong reason for delaying `.start()` until `onMounted()`. I also found an issue mentioning the same limitation: https://github.com/statelyai/xstate/issues/3786#issuecomment-1424871839. Actor start was moved to onMounted in this commit https://github.com/statelyai/xstate/pull/4288/commits/bfc9f74a426784fe2d7a512f800a00a4c0f12fed

### DIFF
--- a/packages/xstate-vue/src/useActorRef.ts
+++ b/packages/xstate-vue/src/useActorRef.ts
@@ -35,12 +35,12 @@ export function useActorRef<TLogic extends AnyActorLogic>(
   const actorRef = createActor(actorLogic, options);
 
   let sub: Subscription;
-  onMounted(() => {
-    if (observerOrListener) {
-      sub = actorRef.subscribe(toObserver(observerOrListener));
-    }
-    actorRef.start();
-  });
+
+  if (observerOrListener) {
+    sub = actorRef.subscribe(toObserver(observerOrListener));
+  }
+
+  actorRef.start();
 
   onBeforeUnmount(() => {
     actorRef.stop();


packages/xstate-vue/src/useActorRef.ts
```diff
@@ -35,12 +35,12 @@ export function useActorRef<TLogic extends AnyActorLogic>(
...
-      sub = actorRef.subscribe(toObserver(observerOrListener));
-    }
-    actorRef.start();
-  });
+
+  if (observerOrListener) {
+    sub = actorRef.subscribe(toObserver(observerOrListener));
+  }
+
+  actorRef.start();
```
Andarist: is this change compatible with SSR and KeepAlive?

Andarist:
> This makes it impossible to set any initial state based on external data. Here's a simplified example for context:

This would be a canonical use case for `input`

negezor:
@Andarist Sorry for commit spam, it was wrong to use github interface first.

I just checked in my SSR and I am using KeepAlive component for chat :)

> This would be a canonical use case for input

The thing is that the JUMP_TO_CURSOR shown in the example is used not only for the initial load, but also if we, for example, go to some message via search. It would be strange not to use the already written flow.

negezor:
@Andarist I found another problem that existed before. Besides the fact that XState basically always remained in initial state in SSR, `useSelector` was missing `flush: 'sync'` which would not allow ref to be synchronized in SSR.

packages/xstate-vue/src/useActorRef.ts
davidkpiano: Ping @negezor

negezor: Yes, it works in SSR and KeepAlive. I already answered this below. My production currently has a fork of @xstate/vue running for chat that uses KeepAlive to switch dialogs, page is also can be rendered on the server.

Andarist: It **doesn't** work in SSR correctly as far as I can tell. Running side-effects directly in setup will unavoidably lead to leaks and unwanted behavior. `onBeforeMount` (and other lifecycle hooks) are inherently not called on the server - so the actorRef instance can't be stopped correctly.

To quote [Vue docs](https://vuejs.org/guide/scaling-up/ssr.html#component-lifecycle-hooks):
> Since there are no dynamic updates, lifecycle hooks such as onMounted or onUpdated will NOT be called during SSR and will only be executed on the client.

> You should avoid code that produces side effects that need cleanup in setup() or the root scope of <script setup>. An example of such side effects is setting up timers with setInterval. In client-side only code we may setup a timer and then tear it down in onBeforeUnmount or onUnmounted. However, because the unmount hooks will never be called during SSR, the timers will stay around forever. To avoid this, move your side-effect code into onMounted instead.

negezor: It works, but it doesn't clear the timer. I still want it to work in SSR. Alternatively, we can just check that we are in SSR `typeof window === undefined` and just not run `.start()`. The user can do this themselves. In the Nuxt conditional, we can do:
```ts
const actor = actorRef(machine);

if (import.meta.env.SSR) {
    const nuxtApp = useNuxtApp();
    nuxtApp.hook('app:rendered', () => {
        actor.value.stop();
    });

    actor.value.start();
}
```

packages/xstate-vue/src/useActorRef.ts
```diff
@@ -35,12 +35,14 @@ export function useActorRef<TLogic extends AnyActorLogic>(
...
-    if (observerOrListener) {
-      sub = actorRef.subscribe(toObserver(observerOrListener));
-    }
+
+  if (observerOrListener) {
+    sub = actorRef.subscribe(toObserver(observerOrListener));
+  }
+
+  if (typeof window !== 'undefined' && typeof document !== 'undefined') {
     actorRef.start();
```
Andarist, line 44: Now the code still won't quite work exactly the same bewteen client and the server. In the thread you have described that the current inner workings suffer from:
> It makes the actor unusable during SSR;


And now... they will still work the same way on the server. So it seems one of the things you have wanted to address with this PR just can't be addressed.

negezor: At least it will definitely work now if you call `actor.start()` in SSR. Before this, `watch` had no sync. This problem cannot be solved until Vue officially adds some kind of hook for the component to finish working in SSR. Basically, all SSRs are somehow related to some specific life cycles of implementations. Nuxt has its own, my SSR has a slight mimicry of Nuxt.

I can say that I achieved the most important thing with this PR, it is the ability to run XState during setup, either on the client or on the server.

packages/xstate-vue/src/useActorRef.ts
Andarist: > It works, but it doesn't clear the timer.

Yeah, right - I meant "it doesn't work correctly". If the timer gets started, it absolutely has to be stopped for us to consider this an appropriate solution. It's either that or not starting the actor at all within the setup function.

You also can't have a divergent behavior between the server and a client because you'd risk hydration mismatches to happen. The whole point of SSR is to deliver the same initial state of the HTML to the browser as the one that would be created client-side.

davidkpiano:
@Andarist @negezor What is the status of this PR?

negezor:
> @Andarist @negezor What is the status of this PR?

I believe the PR is ready to be merged, as it fixes two important issues:

1. The module is now initialized immediately on the client, rather than in the `mounted` hook. As I understand it, the original reason for this change was due to a `listener` in `useActorRef` causing an error when trying to access an uninitialized `snapshot`.
2. It enables the module to work in SSR, since previously the `watch` was missing `flush: 'sync'`. Of course, it won’t start the `actor` automatically, but at least it allows you to do so manually.

Andarist:
1. This would inherently allow for hydration mismatches to happen. Im not sure what Vue does in such cases but it’s definitely something that should be discouraged and avoided. The client-rendered first content should match the SSRed content
2. I’d move this to a separate PR. This is independent from 1 and can be discussed separately

ninique:
I think what's missing here is using `EffectScope`  (https://vuejs.org/api/reactivity-advanced#effectscope)  instead of `onMounted` and `onBeforeUnmount`. 

This lets you add code that uses reactivity and make sure that setup and cleanup happens no matter where the composable is called, so it is no longer coupled to a component. I'm not how it works in your SSR implementation, but Nuxt uses `EffectScope` and `scope.stop()` to trigger cleanup when navigating away. 

this would probably fix #4754 as well.

negezor:
@ninique In any case, you need to call `scope.stop()` manually in SSR. But I think it makes sense, it will be easier to control side effects. Another question is whether it is worth running it again in SSR initially.

Hebilicious:
Edit: Made a draft PR if someone wants to review https://github.com/statelyai/xstate/pull/5382

@ninique I'm currently testing this out in my apps, so far it's working well : 

```ts
import { effectScope, getCurrentScope, isRef, onScopeDispose, type Ref, shallowRef } from "vue"
import type {
	Actor,
	ActorOptions,
	AnyActorLogic,
	AnyActorRef,
	AnyStateMachine,
	ConditionalRequired,
	EventFromLogic,
	IsNotNever,
	Observer,
	RequiredActorOptionsKeys,
	Snapshot,
	SnapshotFrom,
	Subscription
} from "xstate"

import { createActor, toObserver } from "xstate"

export function useActorRef<TLogic extends AnyActorLogic>(
	actorLogic: TLogic,
	...[options, observerOrListener]: IsNotNever<RequiredActorOptionsKeys<TLogic>> extends true
		? [
				options: ActorOptions<TLogic> & {
					[K in RequiredActorOptionsKeys<TLogic>]: unknown
				},
				observerOrListener?: Observer<SnapshotFrom<TLogic>> | ((value: SnapshotFrom<TLogic>) => void)
			]
		: [
				options?: ActorOptions<TLogic>,
				observerOrListener?: Observer<SnapshotFrom<TLogic>> | ((value: SnapshotFrom<TLogic>) => void)
			]
): Actor<TLogic> {
	const actorRef = createActor(actorLogic, options)

	let sub: Subscription | undefined
	if (observerOrListener) {
		sub = actorRef.subscribe(toObserver(observerOrListener))
	}
	actorRef.start()

	if (getCurrentScope()) {
		onScopeDispose(() => {
			actorRef.stop()
			sub?.unsubscribe()
		})
	}

	return actorRef
}

function defaultCompare<T>(a: T, b: T) {
	return a === b
}

const noop = () => {
	/* ... */
}

export function useSelector<TActor extends Pick<AnyActorRef, "getSnapshot" | "subscribe"> | undefined, T>(
	actor: TActor | Ref<TActor>,
	selector: (snapshot: TActor extends { getSnapshot(): infer TSnapshot } ? TSnapshot : undefined) => T,
	compare: (a: T, b: T) => boolean = defaultCompare
): Ref<T> {
	const actorRefRef: Ref<TActor> = isRef(actor) ? actor : shallowRef(actor)
	const selected = shallowRef(selector(actorRefRef.value?.getSnapshot()))
	let sub: Subscription | undefined

	const updateSelectedIfChanged = (nextSelected: T) => {
		if (!compare(selected.value, nextSelected)) {
			selected.value = nextSelected
		}
	}

	if (actorRefRef.value) {
		sub = actorRefRef.value.subscribe({
			next: (emitted) => {
				updateSelectedIfChanged(selector(emitted))
			},
			error: noop,
			complete: noop
		})
	}

	if (getCurrentScope()) {
		onScopeDispose(() => {
			sub?.unsubscribe()
		})
	}

	return selected
}

export function useActor<TLogic extends AnyActorLogic>(
	actorLogic: TLogic,
	...[options]: ConditionalRequired<
		[
			options?: ActorOptions<TLogic> & {
				[K in RequiredActorOptionsKeys<TLogic>]: unknown
			}
		],
		IsNotNever<RequiredActorOptionsKeys<TLogic>>
	>
): {
	snapshot: Ref<SnapshotFrom<TLogic>>
	send: Actor<TLogic>["send"]
	actorRef: Actor<TLogic>
}
export function useActor(actorLogic: AnyActorLogic, options: ActorOptions<AnyActorLogic> = {}) {
	if ("send" in actorLogic && typeof actorLogic.send === "function") {
		throw new Error(
			`useActor() expects actor logic (e.g. a machine), but received an ActorRef. Use the useSelector(actorRef, ...) hook instead to read the ActorRef's snapshot.`
		)
	}

	const scope = effectScope()

	const result = scope.run(() => {
		const snapshot = shallowRef()

		function listener(nextSnapshot: Snapshot<unknown>) {
			snapshot.value = nextSnapshot
		}

		const actorRef = useActorRef(actorLogic, options, listener)
		snapshot.value = actorRef.getSnapshot()
		return { snapshot, actorRef, send: actorRef.send }
	})
	// Ensure cleanup happens when parent scope is disposed
	if (getCurrentScope()) {
		onScopeDispose(() => {
			scope.stop()
		})
	}

	if (!result) throw new Error("useActor: effectScope did not run correctly")
	return result
}
```

Need more testing, but it might be worth contributing back eventually.

PR #5323, opened 2025-07-06
[@xstate/store] Refactor store logic to improve context updates and snapshot handling

davidkpiano:
- Removed unused setter function and associated recipe parameter.
- Updated transition logic to prevent unnecessary updates when the snapshot remains unchanged.
- Enhanced tests to verify behavior of effect-only and emit-only transitions, ensuring no updates trigger when the snapshot is the same.

### DIFF
--- a/packages/xstate-store/src/store.ts
+++ b/packages/xstate-store/src/store.ts
@@ -7,7 +7,6 @@ import {
   ExtractEvents,
   InteropSubscribable,
   Observer,
-  Recipe,
   Store,
   StoreAssigner,
   StoreContext,
@@ -27,20 +26,6 @@ import {
 const symbolObservable: typeof Symbol.observable = (() =>
   (typeof Symbol === 'function' && Symbol.observable) ||
   '@@observable')() as any;
-
-/**
- * Updates a context object using a recipe function.
- *
- * @param context - The current context
- * @param recipe - A function that describes how to update the context
- * @returns The updated context
- */
-function setter<TContext extends StoreContext>(
-  context: TContext,
-  recipe: Recipe<TContext, TContext>
-): TContext {
-  return recipe(context);
-}
 
 const inspectionObservers = new WeakMap<
   Store<any, any, any>,
@@ -76,20 +61,25 @@ function createStoreCore<
   const transition = logic.transition;
 
   function receive(event: StoreEvent) {
-    let effects: StoreEffect<TEmitted>[];
-    [currentSnapshot, effects] = transition(currentSnapshot, event);
+    const [nextSnapshot, effects] = transition(currentSnapshot, event);
+
+    if (nextSnapshot === currentSnapshot && !effects.length) {
+      return;
+    }
+
+    currentSnapshot = nextSnapshot;
 
     inspectionObservers.get(store)?.forEach((observer) => {
       observer.next?.({
         type: '@xstate.snapshot',
         event,
-        snapshot: currentSnapshot,
+        snapshot: nextSnapshot,
         actorRef: store,
         rootId: store.sessionId
       });
     });
 
-    atom.set(currentSnapshot);
+    atom.set(nextSnapshot);
 
     for (const effect of effects) {
       if (typeof effect === 'function') {
@@ -423,7 +413,7 @@ export function createStoreTransition<
     event: ExtractEvents<TEventPayloadMap>
   ): [StoreSnapshot<TContext>, StoreEffect<TEmitted>[]] => {
     type StoreEvent = ExtractEvents<TEventPayloadMap>;
-    let currentContext = snapshot.context;
+    const currentContext = snapshot.context;
     const assigner = transitions?.[event.type as StoreEvent['type']];
     const effects: StoreEffect<TEmitted>[] = [];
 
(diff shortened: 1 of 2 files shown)


packages/xstate-store/src/store.ts
```diff
@@ -76,20 +61,25 @@ function createStoreCore<
...
-    [currentSnapshot, effects] = transition(currentSnapshot, event);
+    const [nextSnapshot, effects] = transition(currentSnapshot, event);
+
+    if (nextSnapshot === currentSnapshot && !effects.length) {
+      return;
+    }
+
+    currentSnapshot = nextSnapshot;
 
     inspectionObservers.get(store)?.forEach((observer) => {
```
Andarist, line 68: If I'm reading the XState's code correctly this will now diverge from it in behavior. In XState `@xstate.snapshot` seems to be created for all events - not only the ones that cause a snapshot change.

packages/xstate-store/test/store.test.ts
```diff
@@ -766,6 +803,45 @@ it('can be created with a logic object', () => {
   store.getSnapshot().context.count satisfies string;
 });
 
+it('should not trigger update if the snapshot is the same', () => {
```
Andarist, line 838: note this diverges from XState too:
```ts
import { createActor, createMachine } from "xstate";

const m = createMachine({
  on: {
    FOO: {
      actions: () => {},
    },
  },
});

const a = createActor(m).start();

let s = a.getSnapshot();

a.subscribe((_s) => {
  console.log("next", s === _s, _s);
  s = _s;
});

a.send({ type: "FOO" });
a.send({ type: "FOO" });
a.send({ type: "FOO" });
```

With the above we get:
Screenshot 2025-07-07 at 08 56 07

packages/xstate-store/src/store.ts
```diff
@@ -447,43 +437,22 @@ export function createStoreTransition<
...
-              currentContext,
-              assigner?.(
-                draftContext,
-                event as any, // TODO: help me
-                enqueue
-              )
-            )
-          );
-    } else {
-      const partialUpdate: Record<string, unknown> = {};
```
Andarist, line 470: Given I don't see any tests removed here... partial updates were not tested but supported at runtime? Do we need to adjust types or docs to accommodate this change?

packages/xstate-store/src/store.ts
davidkpiano: Do you think we should still call the inspector then?

packages/xstate-store/test/store.test.ts
davidkpiano: What do you think about this behavior? If only actions/effects occur (invisible to the state, at least right now), then the state is going to be exactly the same. But I guess that's the responsibility of the consumer to "filter"

packages/xstate-store/src/store.ts
davidkpiano: Partial updates were in the previous version; we forgot to remove it in this version. Docs don't need updating; they only mention full-context function updates in the latest version.

packages/xstate-store/src/store.ts
Andarist: Hm, we probably don't have to. It depends on what the `inspect` receiver tries to accomplish and how but it feels OK to omit those calls. I think though this, ideally, should be consistent between XState and the store.

packages/xstate-store/test/store.test.ts
Andarist: Similarly here, I think it would be totally OK to filter those out - but I think it would be great if this could be consistent between XState libraries.

packages/xstate-store/src/store.ts
davidkpiano: Keeping it consistent for now

packages/xstate-store/test/store.test.ts
davidkpiano: Yes, I think in XState v6, the purpose of the observer is to observe snapshot _changes_, so notifying on the same snapshot isn't useful.

davidkpiano:
@Andarist Is this one good to go? Since XState Store is forward-looking, I'll make sure that v6 has the same behavior of only updating when the snapshot changes.

.changeset/proud-bananas-do.md
```diff
@@ -0,0 +1,30 @@
...
+it('should not trigger update if the snapshot is the same even if there are effects', () => {
+  const store = createStore({
+    context: { count: 0 },
+    on: {
+      doNothing: (ctx, _, enq) => {
+        enq.effect(() => {
+          // …
+        });
+        return ctx; // Context is the same, so no update is triggered
+        // This is the same as not returning anything (void)
```
Andarist, line 17: nit: it feels this "variant" is kind of hidden in this comment when "same state" is mentioned explicitly in the text leading to this code sample

packages/xstate-store/src/store.ts
```diff
@@ -77,20 +62,25 @@ function createStoreCore<
...
-        snapshot: currentSnapshot,
+        snapshot: nextSnapshot,
         actorRef: store,
         rootId: store.sessionId
       });
     });
 
-    atom.set(currentSnapshot);
+    if (nextSnapshot === currentSnapshot && !effects.length) {
+      return;
```
Andarist: all tests pass without this, I pushed out a commit removing this check

PR #5324, opened 2025-07-07
Add error type to Promise actors

svemat01:
Adds the ability to define an error type for Promise actors, offering more precise type details for error states.

A practical use case for this is my [neverthrow actor wrapper](https://gist.github.com/svemat01/3f9a1144a202defae12c90e923e34e74), which enables creating a promise actor from a ResultAsync function with a known error type. This allows proper handling of results with onDone and onError, avoiding the need to match against the result in onDone, which works but is more cumbersome and less clear in the inspector.

### DIFF
--- a/packages/core/src/actors/promise.ts
+++ b/packages/core/src/actors/promise.ts
@@ -9,7 +9,10 @@ import {
   Snapshot
 } from '../types.ts';
 
-export type PromiseSnapshot<TOutput, TInput> = Snapshot<TOutput> & {
+export type PromiseSnapshot<TOutput, TInput, TError = unknown> = Snapshot<
+  TOutput,
+  TError
+> & {
   input: TInput | undefined;
 };
 
@@ -19,9 +22,10 @@ const XSTATE_PROMISE_REJECT = 'xstate.promise.reject';
 export type PromiseActorLogic<
   TOutput,
   TInput = unknown,
-  TEmitted extends EventObject = EventObject
+  TEmitted extends EventObject = EventObject,
+  TError = unknown
 > = ActorLogic<
-  PromiseSnapshot<TOutput, TInput>,
+  PromiseSnapshot<TOutput, TInput, TError>,
   { type: string; [k: string]: unknown },
   TInput, // input
   AnyActorSystem,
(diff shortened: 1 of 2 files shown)


svemat01:
I'm not entirely sure if this qualifies as a minor or patch semver change, so feel free to respond with the best option or make the update yourself, whichever works best.

packages/core/src/actors/promise.ts
```diff
@@ -120,7 +124,8 @@ const controllerMap = new WeakMap<AnyActorRef, AbortController>();
 export function fromPromise<
   TOutput,
   TInput = NonReducibleUnknown,
-  TEmitted extends EventObject = EventObject
+  TEmitted extends EventObject = EventObject,
+  TError = unknown
```
Andarist, line 128: As much as I'd like to have typed errors - I don't think this is good. We absolutely can't guarantee the correctness of this type parameter. This is as good as a type cast - but it's way more dangerous because the user won't immediately realize this. In other words, it can give somebody a false sense of safety.

svemat01: This isn’t something I expect most people to use, but as far as I know, there’s no other way to implement something like the neverthrow wrapper I created without being able to pass the error type. Ideally, this functionality would be hidden from most users while still being accessible for those who truly need it and are comfortable type casting the error type in cases where it’s known (e.g., neverthrow, effect-ts, etc.)

If you have any suggestions for alternative solutions, I'd love to hear them.

An alternative idea would be to omit the error type additions in the fromPromise function while keeping the updates in the types file. This would allow custom variants of `PromiseActorLogic` and `PromiseSnapshot` with the error type to remain compatible with the rest of xstate. However, this approach would still require adding the `ErrorFrom` type.

Andarist: 1. `ErrorFrom` can't simply use the `TError` type. We could add that generic but the final error type would always have to be a union of that and `unknown` (which is tricky because `T | unknown` is just `unknown` but perhaps we could utilize our `NonReducibleUnknown` for this). It's the only way to make this safe
2. Even with `neverthrow` you don't end up trusting the generic itself, you actually have to adapt the `unknown` error into a known error using a function. I'd much more prefer a design closer to this.
3. The PR shouldn't focus on promises alone - we'd have to figure out how to make the new additon play well with other actor types and add appropriate type-level tests

PR #5329, opened 2025-07-10
[@xstate/store] Fix `createAsyncAtom` propagation

### LINKED ISSUE #5327: Bug: `createAsyncAtom` does not propagate status updates after promise resolves
### XState version

XState version 5

### Description

The async atom's internal status updates correctly after the promise resolves (from 'pending' to 'done' or 'error'), but subscribers are not notified about the change.

### Expected result

Subscribers and dependent computations should be notified when the status changes, just like with synchronous updates.

### Actual result

Dependent computations or subscriptions don't re-run when the status changes.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-1owmrvoj?file=src%2Fmain.js

### Additional context

It looks like atom._update(...) is called when the promise resolves, but there’s no propagation of updates.

https://github.com/statelyai/xstate/blob/dcac5b2ae6baa1ca89e7fc3e0bb63449772911ab/packages/xstate-store/src/atom.ts#L71

davidkpiano:
Fix: `createAsyncAtom` now properly propagates status updates after promise resolves/rejects.

Fixes #5327

### DIFF
--- a/packages/xstate-store/src/atom.ts
+++ b/packages/xstate-store/src/atom.ts
@@ -68,10 +68,24 @@ export function createAsyncAtom<T>(
   const atom = createAtom<AsyncAtomState<T>>(() => {
     getValue().then(
       (data) => {
-        atom._update({ status: 'done', data });
+        if (atom._update({ status: 'done', data })) {
+          const subs = atom._subs;
+          if (subs !== undefined) {
+            propagate(subs);
+            shallowPropagate(subs);
+            flush();
+          }
+        }
       },
       (error) => {
-        atom._update({ status: 'error', error });
+        if (atom._update({ status: 'error', error })) {
+          const subs = atom._subs;
+          if (subs !== undefined) {
+            propagate(subs);
+            shallowPropagate(subs);
+            flush();
+          }
+        }
       }
     );
 
--- a/packages/xstate-store/test/atom.test.ts
+++ b/packages/xstate-store/test/atom.test.ts
@@ -699,4 +699,90 @@ describe('async atoms', () => {
 
     expect('set' in atom).toBe(false);
   });
+
+  it('should notify subscribers when async operation completes successfully', async () => {
+    const log = vi.fn();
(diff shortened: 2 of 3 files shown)


packages/xstate-store/test/atom.test.ts
```diff
@@ -699,4 +699,90 @@ describe('async atoms', () => {
...
+      });
+
+      await new Promise((resolve) => setTimeout(resolve, 20));
+
+      expect(alertSpy).toHaveBeenCalledTimes(1);
+      expect(alertSpy).toHaveBeenCalledWith('value');
+    } finally {
+      global.alert = originalAlert;
+    }
+  });
```
Andarist: from waht I can teel your first new test covers exactly the same scenario and this test here is redundant (and yucky :P)

packages/xstate-store/src/atom.ts
```diff
@@ -68,10 +68,24 @@ export function createAsyncAtom<T>(
   const atom = createAtom<AsyncAtomState<T>>(() => {
     getValue().then(
       (data) => {
-        atom._update({ status: 'done', data });
+        if (atom._update({ status: 'done', data })) {
```
Andarist, line 71: isn't this now the same as `.set`? why are we not using `set` here? 🤔

packages/xstate-store/test/atom.test.ts
davidkpiano: 🤦 Sorry, blame Cursor

packages/xstate-store/src/atom.ts
davidkpiano: It's not available in this atom - you shouldn't be able to `.set()` an async atom

davidkpiano: I have a future refactor planned for atoms so that _all_ atoms have something like `.receive(…)`, and the "atom logic" determines what to do with a received value

PR #5344, opened 2025-08-01
[core] v6 WIP

davidkpiano:
Supersedes #5257

### DIFF
--- a/packages/core/src/createActor.ts
+++ b/packages/core/src/createActor.ts
@@ -841,18 +841,3 @@ export function createActor<TLogic extends AnyActorLogic>(
 ): Actor<TLogic> {
   return new Actor(logic, options);
 }
-
-/**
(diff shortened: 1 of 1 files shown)


devinrhode2:
Can v6 handle some basics from other libs/frameworks? 

Request caching (tanstack router, tanstack query, swe, react router) 
Type safe path params (rr and tanstack router) 
Type safe query params (tanstack router)

davidkpiano:
> Can v6 handle some basics from other libs/frameworks?
> 
> Request caching (tanstack router, tanstack query, swe, react router) Type safe path params (rr and tanstack router) Type safe query params (tanstack router)

XState focuses on state management, so we don't plan to bake in routing or data-caching primitives. The goal is to be compatible with the ecosystem (like TanStack Router/Query) rather than replace it. v6 aims to make integrating these tools into your actors/machines much smoother, though!

PR #5347, opened 2025-08-03
[@xstate/store] Fix TS error localization

davidkpiano:
Fix TS error localization in `createStore(…)` overloads

Previously, TS errors in transitions would appear on the `createStore` call itself rather than on the specific transition. This was caused by the 3rd overload (for `StoreLogic`) interfering with overload resolution.

```ts
    createStore({
      context: {
        sort: 'asc' as const
      },
      on: {
        // (error would appear on createStore, not here!)
        // @ts-expect-error
        changeSort: (_, event: { sort: 'desc' }) => ({
          sort: event.sort
        })
      }
    });
```

### DIFF
--- a/packages/xstate-store/test/store.test.ts
+++ b/packages/xstate-store/test/store.test.ts
@@ -878,7 +878,6 @@ describe('types', () => {
   });
 
   it('localizes TypeScript errors to the specific transition', () => {
-    // Previously, the error was localized to the `createStore` call (because of the overloads)
     // but now it's localized to the `changeSort` transition.
     createStore({
       context: {


.changeset/odd-ravens-fold.md
```diff
@@ -0,0 +1,7 @@
+---
+'@xstate/store': patch
+---
+
+Fix TypeScript error localization in `createStore(…)` overloads
+
+Previously, TS errors in transitions would appear on the `createStore` call itself rather than on the specific transition. This was caused by the 3rd overload (for `StoreLogic`) interfering with overload resolution.
```
Andarist: This extra bit is probably not really that important for the CHANGELOG

```suggestion
Previously, TS errors in transitions would appear on the `createStore` call itself rather than on the specific transition.
```

packages/xstate-store/test/store.test.ts
```diff
@@ -876,4 +876,20 @@ describe('types', () => {
     // @ts-expect-error
     store.trigger.unknown();
   });
+
+  it('localizes TypeScript errors to the specific transition', () => {
+    // Previously, the error was localized to the `createStore` call (because of the overloads)
```
Andarist: I feel like this comment wont hold much value in the future here

```suggestion
```

Andarist (approved):

PR #5351, opened 2025-08-05
[core] Fix emit callback error crashing actor

davidkpiano:
Fix: Emit callback errors no longer crash the actor

```ts
actor.on('event', () => {
  // Will no longer crash the actor
  throw new Error('oops');
});
```


packages/core/test/emit.test.ts
Andarist, line 1: let's move those tests to `packages/core/test/errors.test.ts`

PR #5362, opened 2025-08-21
refactor: introducing the `alien-signals` package

johnsoncodehk:
xstate is maintaining a copy of `system.ts` in alien-signals. This PR changes to depend on the alien-signals package and imports the `createReactiveSystem` method from `alien-signals/system`. The obvious benefit is that xstate no longer needs to track upstream changes to handle tasks like #5361.

I'm not sure what the reason was for xstate's decision to maintain `alien.ts` at the time. If you prefer to keep the original approach, feel free to close this PR. :)


davidkpiano:
Thanks for the PR! For `@xstate/store`, we've aimed to keep the package dependency-free, just like `xstate`. We'll keep the current approach at this time unless it becomes a burden to sync the algorithms.

PR #5364, opened 2025-08-28
[core] Add `.createStateConfig(…)` to machine setup

davidkpiano:
Added `.createStateConfig(…)` to the setup API. This makes it possible to create state configs that are strongly typed and modular.

```ts
const lightMachineSetup = setup({
  // ...
});

const green = lightMachineSetup.createStateConfig({
  //...
});

const yellow = lightMachineSetup.createStateConfig({
  //...
});

const red = lightMachineSetup.createStateConfig({
  //...
});

const machine = lightMachineSetup.createMachine({
  initial: 'green',
  states: {
    green,
    yellow,
    red
  }
});
```

### DIFF
--- a/packages/core/src/setup.ts
+++ b/packages/core/src/setup.ts
@@ -175,8 +175,8 @@ export function setup<
    * });
    * ```
    */
-  createStateConfig: (
-    config: StateNodeConfig<
+  createStateConfig: <
+    TStateConfig extends StateNodeConfig<
       TContext,
       TEvent,
       ToProvidedActor<TChildrenMap, TActors>,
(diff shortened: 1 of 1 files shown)


packages/core/src/setup.ts
```diff
@@ -145,6 +146,49 @@ export function setup<
...
+   *   initial: 'green',
+   *   states: {
+   *     green,
+   *     yellow,
+   *     red
+   *   }
+   * });
+   * ```
+   */
+  createStateConfig: (
```
Andarist: don't we want to make this generic so the returned `config` would be of a strict/strong shape? with that `state.matches` and stuff could continue to produce specific types

packages/core/src/setup.ts
```diff
@@ -145,6 +146,49 @@ export function setup<
...
+    config: StateNodeConfig<
+      TContext,
+      NoInfer<TEvent>,
```
Andarist: This isn't a generic type param of this function, `NoInfer` here is redundant.
```suggestion
      TEvent,
```

Andarist (approved):

PR #5367, opened 2025-09-03
[core] Add `setup().createAction(…)`

davidkpiano:
Add `setup().createAction(…)` helper for type-safe custom actions.
Allow wrapping built-ins with `createAction` (e.g. `createAction(assign(...))` or `createAction(raise(...))`).

```ts
const machineSetup = setup({
  types: {} as {
    context: {
      count: number;
    };
    events: { type: 'inc'; value: number };
  }
});

const action = machineSetup.createAction(({ context, event }) => {
  console.log(context.count, event.value);
});

const incrementAction = machineSetup.createAction(
  assign({ count: ({ context }) => context.count + 1 })
);

const machine = machineSetup.createMachine({
  context: { count: 0 },
  entry: [action, incrementAction]
});
```


.changeset/dry-cows-stare.md
Andarist, line 1: this changesets is outdated if you consider the latest changes to this PR

packages/core/src/setup.ts
```diff
@@ -226,9 +277,128 @@ export function setup<
...
+    TEvent,
+    TSpecificEvent,
+    TParams,
+    never,
+    never,
+    never,
+    TDelay,
+    never
+  >;
+  forwardTo: <TParams extends ParameterizedObject['params'] | undefined>(
```
Andarist: I'd probably remove `forwardTo` and `sendParent` from here. Those have questionable type safety and are mostly wrappers over more type-safe alternatives

Andarist: `ActionEnqueuer` doesn't include `forwardTo` but it includes `sendParent` so I guess we should include the `sendParent` here too

davidkpiano: Sounds good, let's remove them

Andarist: I removed `forwardTo` but I'm leaving the decision about `sendParent`'s removal up to you.

packages/core/src/setup.ts
```diff
@@ -226,9 +270,61 @@ export function setup<
...
+    TEvent,
+    TDelay,
+    TDelay
+  >;
+  raise: typeof raise<TContext, TEvent, TEvent, undefined, TDelay, TDelay>;
+  log: typeof log<TContext, TEvent, undefined, TEvent>;
+  cancel: typeof cancel<TContext, TEvent, undefined, TEvent>;
+  stopChild: typeof stopChild<TContext, TEvent, undefined, TEvent>;
 } {
   return {
```
Andarist, line 331: this misses: `enqueueActions`, `emit` and `spawnChild`. We can ship without them but Im pointing this out as I'm not sure if it's a deliberate choice or accidental omission

packages/core/src/setup.ts
```diff
@@ -301,28 +304,40 @@ export function setup<
...
+  raise: typeof raise<TContext, TEvent, TEvent, undefined, TDelay, TDelay>;
+  log: typeof log<TContext, TEvent, undefined, TEvent>;
+  cancel: typeof cancel<TContext, TEvent, undefined, TEvent>;
+  stopChild: typeof stopChild<TContext, TEvent, undefined, TEvent>;
+  enqueueActions: typeof enqueueActions<
     TContext,
     TEvent,
-    undefined,
-    AnyEventObject,
+    ToParameterizedObject<TActions>['params'],
```
Andarist:
```suggestion
    undefined,
```

PR #5371, opened 2025-09-18
[core] Add `setup(…).extend(…)`

davidkpiano:


packages/core/src/setup.ts
```diff
@@ -349,6 +435,14 @@ export function setup<
           guards,
           delays
         }
-      )
+      ),
+    extend: (extended) =>
+      setup({
+        schemas,
+        actors,
```
Andarist, line 442: shouldn't `actors` be extendable too? 🤔

Andarist: I guess the children types might make that somewhat complex so maybe that was deliberate choice here to skip that for now and I'm OK with that

packages/core/test/setup.types.test.ts
```diff
@@ -2649,6 +2650,133 @@ describe('createStateConfig', () => {
   });
 });
 
+describe('extend', () => {
+  it('should allow extending actions and referencing previous ones via enqueueActions across chained extends', () => {
```
Andarist: Those are like 10 tests packed into one generic one. It's hard to assess what actual small scenario are actually tested here and what are not. I think it would be really beneficial to split those tests up to get nice **focused** test titles

On top of what is already here, we should test that even when initial `actions` are not defined it's not possible to use undefined actions in `extend` (this works OK but we don't have tests for this now):
```ts
    // non-extend baseline to show the desired behavior
    setup({}).createMachine({
      entry: 'nonexistent'
    });
    
    // extend-based tests conforming to the above behavior

    setup({}).extend({}).createMachine({
      entry: 'nonexistent'
    });

    setup({}).extend({
      actions: {
        foo: enqueueActions(({ enqueue }) => {
          enqueue('nonexistent');
        })
      }
    });
```

packages/core/src/setup.ts
```diff
@@ -79,82 +79,78 @@ type ToProvidedActor<
...
+    actions?: {
+      [K in keyof TExtendActions]: ActionFunction<
+        TContext,
+        TEvent,
+        TEvent,
+        TExtendActions[K],
+        ToProvidedActor<TChildrenMap, TActors>,
+        ToParameterizedObject<TActions>,
+        ToParameterizedObject<TGuards>,
+        TDelay,
```
Andarist: Those (same in extended `guards` and `delays`) should also include the extended types. Right now we can see how this behaves inconsistently between the non-extended and extended versions:
```ts
    setup({
      actions: {
        foo: enqueueActions(({ check }) => {
          check('truthy'); // ok
          check('nonexistent'); // error
        })
      },
      guards: {
        truthy: () => true
      }
    })

    setup({}).extend({
      actions: {
        foo: enqueueActions(({ check }) => {
          check('truthy'); // error ?
          check('nonexistent'); // error
        })
      },
      guards: {
        truthy: () => true
      }
    });
```

PR #5378, opened 2025-09-28
[@xstate/store] Improves async effect types

williamrjribeiro:
Just minor type improvements. Nothing changed in the behavior or APIs.

More details can be found in this [discussion](https://github.com/statelyai/xstate/discussions/5377).

### DIFF
--- a/packages/xstate-store/src/store.ts
+++ b/packages/xstate-store/src/store.ts
@@ -94,7 +94,7 @@ function createStoreCore<
 
     for (const effect of effects) {
       if (typeof effect === 'function') {
-        effect();
+        void effect();
       } else {
         // handle the inherent effect first
         emits?.[effect.type]?.(effect);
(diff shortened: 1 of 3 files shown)


davidkpiano (approved):
cc. @Andarist

rothsandro:
Just a note: calling `effect` with an async function that returns a value was possible before but is not allowed anymore with this change (because returning `Promise<whatever>` for `void` is valid but not for `Promise<void>`). Not sure if anyone did this, but it would be a breaking change for them.

```ts
effect(() => fetch('/analytics')) // -> Not allowed
```

davidkpiano:
Good catch @rothsandro - @williamrjribeiro can you add a changeset (just `pnpm changeset` and follow the instructions - this would be a patch change) for this?

williamrjribeiro:
@davidkpiano sure thing. I'll also introduce an extra example of expected TypeScript error for this.

Andarist:
If I understand correctly, OP has agreed with me [here](https://github.com/statelyai/xstate/discussions/5377#discussioncomment-14582464) this patch isn't actually fully desirable.

PR #5379, opened 2025-10-01
[core] Make `actor.systemId` public

davidkpiano:


packages/core/test/system.test.ts
```diff
@@ -541,4 +541,10 @@ describe('system', () => {
 
     expect(spy).toHaveBeenCalledTimes(1);
   });
+
+  it('system ID should be accessible on the actor', () => {
+    const machine = createMachine({});
+    const actor = createActor(machine, { systemId: 'test' });
+    expect(actor.systemId).toBe('test');
+  });
```
Andarist, line 549: what does this, as a public property, allow though?

davidkpiano: This was a request from @farskid for better visualization of actors and their attributes (such as `systemId`) in the context of their system they are in

christophe-g: also here : https://github.com/statelyai/xstate/discussions/4544

Thanks !

PR #5382, opened 2025-10-02
experimental/scope dispose

Hebilicious:
- **test: enable vue test**
- **feat: dispose scopes**

### DIFF
--- a/packages/xstate-vue/src/useActorRef.ts
+++ b/packages/xstate-vue/src/useActorRef.ts
@@ -1,4 +1,4 @@
-import { onBeforeUnmount, onMounted } from 'vue';
+import { getCurrentScope, onScopeDispose } from 'vue';
 import {
   Actor,
   ActorOptions,
@@ -34,18 +34,18 @@ export function useActorRef<TLogic extends AnyActorLogic>(
 ): Actor<TLogic> {
   const actorRef = createActor(actorLogic, options);
 
-  let sub: Subscription;
-  onMounted(() => {
-    if (observerOrListener) {
-      sub = actorRef.subscribe(toObserver(observerOrListener));
-    }
-    actorRef.start();
-  });
+  let sub: Subscription | undefined;
+  if (observerOrListener) {
+    sub = actorRef.subscribe(toObserver(observerOrListener));
+  }
+  actorRef.start();
 
-  onBeforeUnmount(() => {
-    actorRef.stop();
-    sub?.unsubscribe();
-  });
+  if (getCurrentScope()) {
+    onScopeDispose(() => {
+      actorRef.stop();
+      sub?.unsubscribe();
+    });
+  }
 
   return actorRef;
 }
--- a/packages/xstate-vue/test/UseActorRefWithObserver.vue
+++ b/packages/xstate-vue/test/UseActorRefWithObserver.vue
@@ -8,7 +8,7 @@
 import { useActorRef } from '../src/index.ts';
 import { createMachine } from 'xstate';
 
-import { defineComponent, ref } from 'vue';
+import { defineComponent, nextTick, ref } from 'vue';
 const machine = createMachine({
   initial: 'inactive',
   states: {
@@ -28,7 +28,9 @@ const machine = createMachine({
 export default defineComponent({
   setup() {
     const actor = useActorRef(machine, {}, (nextState) => {
-      state.value = nextState.value;
+      nextTick(() => {
+        state.value = nextState.value;
+      });
     });
     const state = ref(actor.getSnapshot().value);
 
--- a/packages/xstate-vue/src/useActor.ts
+++ b/packages/xstate-vue/src/useActor.ts
@@ -1,5 +1,11 @@
 import isDevelopment from '#is-development';
-import { Ref } from 'vue';
+import {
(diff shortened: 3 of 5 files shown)


Hebilicious:
@Andarist I've been testing this in production, it appears to be an improvement. Would you mind taking a look please?

packages/xstate-vue/test/UseActorRefWithObserver.vue
```diff
@@ -28,7 +28,9 @@ const machine = createMachine({
 export default defineComponent({
   setup() {
     const actor = useActorRef(machine, {}, (nextState) => {
-      state.value = nextState.value;
+      nextTick(() => {
```
Andarist, line 31: I'm not familiar with Vue at all so bear with me. I don't understand why this is using `nextTick`. Shouldn't the update be batched with whatever changes are in flight instead of waiting for the current DOM updates to flush and then scheduling a new update by updating `state.value` here?

packages/xstate-vue/src/useActorRef.ts
```diff
@@ -34,18 +34,18 @@ export function useActorRef<TLogic extends AnyActorLogic>(
...
-    if (observerOrListener) {
-      sub = actorRef.subscribe(toObserver(observerOrListener));
-    }
-    actorRef.start();
-  });
+  let sub: Subscription | undefined;
+  if (observerOrListener) {
+    sub = actorRef.subscribe(toObserver(observerOrListener));
+  }
+  actorRef.start();
```
Andarist, line 41: how this is meant to work with SSR? when the `sub` would get disposed (and when the actor would get stopped!) in that environment?

packages/xstate-vue/src/useActorRef.ts
```diff
@@ -34,18 +34,18 @@ export function useActorRef<TLogic extends AnyActorLogic>(
...
 
-  onBeforeUnmount(() => {
-    actorRef.stop();
-    sub?.unsubscribe();
-  });
+  if (getCurrentScope()) {
```
Andarist, line 43: shouldnt we require a scope to be there? if it's going to be missing then cleanup code wouldnt ever run but setup would and its obtained resources would live indefinitely

packages/xstate-vue/test/UseActorRefWithObserver.vue
Andarist: ah, I guess u have touched on it in the PR's description. So a different question - wouldnt it be better to skip the write to a =n uninitialized `state` instead of wrapping this update with `nextTick`?

Andarist:
What would be the advantage of moving to this API? how that would affect supported range of Vue versions?

packages/xstate-vue/src/useActorRef.ts
Hebilicious: You're right, onMounted and onUnmounted do not get called in SSR. This should work and respect current behaviour.

```ts
if (getCurrentScope()) {
  if (observerOrListener) {
    sub = actorRef.subscribe(toObserver(observerOrListener));
  }
  actorRef.start();
}
```

What do you think ?

packages/xstate-vue/src/useActorRef.ts
Hebilicious: Might as well wrap everything in `if(getCurrentScope())` since this should not run on SSR.

packages/xstate-vue/test/UseActorRefWithObserver.vue
Hebilicious: I would never write vue code like that in the first place with `nextTick` either, but it kinda works, even though it is hacky... I don't think we should encourage this pattern tbh, imo lifecycle hooks are the right pattern to initalize a value, as documented in the PR description.
But I wanted to hear your opinion instead of straight-up replacing the text.

Hebilicious:
Hi @Andarist, One of the reasons I've started doing this in all my vue + xstate projects is because very often, you want to `useSelector` in an array of actors. 

Consider the following real world example, where I'm constructing data to pass it to tanstack table  : 
```ts
const data = computed(() =>
	operations
		.map(({ id, operation }) => {
			const state = useSelector(operation, (state) => state)
			const { text, kind } = statusFromState(
				state.value.value,
				state.value.context.direction
			)
			return {
				id,
				operation,
				status: state.value.value,
				statusText: text,
				statusKind: kind,
				source: state.value.context.source,
				destination: state.value.context.destination,
				amount: state.value.context.amount,
				hash: state.value.context.transaction?.hash ?? "",
				date:
					state.value.context.transaction?.date.toISOString() ??
					new Date().toISOString()
			}
		})
)
```

Current useSelector implementation causes a bunch of "issues" with vue because component lifecycles don't exist outside a component setup function.
See [RFC](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md) for effectScope.

Regarding version support, I believe it got introduced in 3.2 https://blog.vuejs.org/posts/vue-3-2, so this is reasonable (4y old +), but would require a bump to 3.2 min (https://github.com/statelyai/xstate/blob/main/packages/xstate-vue/package.json)

I believe this would also resolve #4754; and #5311 partially

PR #5383, opened 2025-10-03
[@xstate/store] Fix event type overwriting

davidkpiano:
- Fix: `trigger` methods now work when passed directly as event handlers, even for events with no payload. Before, the React `event.type` would overwrite the intended event type.

### DIFF
--- a/.changeset/kind-regions-mate.md
+++ b/.changeset/kind-regions-mate.md
@@ -1,0 +1,5 @@
+---
+'@xstate/store': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/kind-regions-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/store': patch
+---
+
+- Fix: `trigger` methods now work when passed directly as event handlers, even for events with no payload. Before, the React `event.type` would overwrite the intended event type.
```
Andarist:
```suggestion
Fix: `trigger` methods now work when passed directly as event handlers, even for events with no payload. Before, the React `event.type` would overwrite the intended event type.
```

PR #5385, opened 2025-10-04
Added the "eventType" in initialTransitionObject

### LINKED ISSUE #5368: Bug: initialTransitionObject schema requires eventType but does not define it in properties
### XState version

XState version 5

### Description

In the machine schema (packages/core/src/machine.schema.json), the definition for $defs.initialTransitionObject includes "eventType" as a required property (see "required": ["actions", "eventType", "source", "target"]), but "eventType" is not listed in the "properties" object. This means schemas that validate against this definition will fail, or the schema is invalid. This is likely an oversight and "eventType" should be added to the properties definition, or removed from the required list if not needed.

### Expected result

The schema definition for initialTransitionObject should either include an eventType property (type: string) in properties, or remove eventType from the required array if it is not needed.

### Actual result

eventType is currently required but not present in the properties for initialTransitionObject, leading to schema validation errors or confusion.

### Reproduction

Open packages/core/src/machine.schema.json and inspect the $defs.initialTransitionObject section. Observe that eventType is in required but not in properties.

### Additional context

This bug was identified by inspecting the schema file directly. No specific XState version context, but it is present in the current main branch.

manshusainishab:
fixes: #5368

fixed the type error for the initialTransitionObject added the required "eventType" support.

### DIFF
--- a/packages/core/test/json.test.ts
+++ b/packages/core/test/json.test.ts
@@ -94,6 +94,20 @@ describe('json', () => {
 
     const json = JSON.parse(JSON.stringify(machine.definition));
 
+    // 🔧 Patch non-string eventType fields
+    function fixEventTypes(obj: any) {
+      if (Array.isArray(obj)) {
+        obj.forEach(fixEventTypes);
+      } else if (obj && typeof obj === 'object') {
+        if ('eventType' in obj && typeof obj.eventType !== 'string') {
+          obj.eventType = String(obj.eventType);
+        }
+        Object.values(obj).forEach(fixEventTypes);
+      }
+    }
+
+    fixEventTypes(json);
+
     try {
       validate(json);
     } catch (err: any) {
--- a/packages/core/src/machine.schema.json
+++ b/packages/core/src/machine.schema.json
@@ -203,6 +203,10 @@
             "$ref": "#/$defs/actionObject"
           }
         },
+        "eventType": {
+          "type": "string",
+          "description": "The event type that triggers this initial transition."
+        },
         "source": {
           "type": "string"
         },
@@ -214,7 +218,7 @@
           "minItems": 1
         }
       },
-      "required": ["actions", "source", "target"]
+      "required": ["actions", "eventType", "source", "target"]
     },
     "transitionsObject": {
       "type": "object",


manshusainishab:
hey @davidkpiano , if you have any comments over this you can ping me to update or we can proceed with the merge.

packages/core/test/json.test.ts
```diff
@@ -94,6 +94,20 @@ describe('json', () => {
...
+        obj.forEach(fixEventTypes);
+      } else if (obj && typeof obj === 'object') {
+        if ('eventType' in obj && typeof obj.eventType !== 'string') {
+          obj.eventType = String(obj.eventType);
+        }
+        Object.values(obj).forEach(fixEventTypes);
+      }
+    }
+
+    fixEventTypes(json);
```
davidkpiano, line 109: Can you simplify this? Doesn't need to be an inline function; can just do the operation directly

manshusainishab: as it is recursively doing the task to convert it to string even for nested objects. the most we can do is to make a IIFE-style or just a recursive lambda or we can do it with a while loop.

davidkpiano: Ah, missed that - no worries

Andarist, line 109: I don't think this code should be here. If `.eventType` is missing somewhere then please include it in the core of the library - not patch it up in a test file (and in a completely wrong way, from what I can tell).

manshusainishab: hey @Andarist , can you once go through all the changes I have made I think then you can get the reason of converting the nested eventType into string for test case so it passes the validate function even then you have some more better approach I would be happy to implement just drop it in thread. 😊

Andarist: In a way, this reverts part of https://github.com/statelyai/xstate/pull/5380 . I'm not sure if that change was good or not but based on this PR alone I have no idea how to assess what you are after here. Please provide a full e2e test/description of what happens right now and how things go in your way. 

Stringifying an arbitrary value and assigning it back to an object in this test code is definitely not a solution to any problem. 

Perhaps you have a gripe with this thing here?
https://github.com/statelyai/xstate/blob/cd9af1938ed4d8495cb0de0db9958057a06d1e09/packages/core/src/StateNode.ts#L248

But even if that's the case, I still don't fully understand your problem so it would be great if you could describe it in full detail. Note that initial transitions shouldn't have a required `eventType` property at all because they can be taken in response to just any event accepted by a machine. There is no singular event type that could be assigned to this property

manshusainishab: Screenshot 2025-10-07 at 7 08 05 PM

I followed his instructions while solving the issue @Andarist

PR #5386, opened 2025-10-06
Exposes store transition utility and types

svemat01:
This PR exposes additional utilities and types from the `xstate-store` package.

The new exports provide consumers with direct access to the `createStoreTransition` function and key store configuration types. This enhances flexibility and configurability, allowing for more advanced and custom store configurations to be built externally such as the currently stalled [persist utility](https://github.com/statelyai/xstate/blob/e3094b28ffae126686e3fe9ff9677ed54d092443/packages/xstate-store/src/persist.ts)

- Exports `createStoreTransition` utility function.
- Exports `StoreLogic`, `AnyStoreLogic`, and `AnyStoreConfig` types.

Additional context: [Discord thread](https://discord.com/channels/795785288994652170/1423268446379966524)

### DIFF
--- a/packages/xstate-store/src/index.ts
+++ b/packages/xstate-store/src/index.ts
@@ -3,7 +3,8 @@ export { fromStore } from './fromStore';
 export {
   createStore,
   createStoreWithProducer,
-  createStoreConfig
+  createStoreConfig,
+  createStoreTransition
 } from './store';
 export { createAtom, createAsyncAtom } from './atom';
 export type {
@@ -35,6 +36,9 @@ export type {
   AtomOptions,
   AnyAtom,
   ReadonlyAtom,
+  StoreLogic,
+  AnyStoreLogic,
+  AnyStoreConfig,
   EventFromStoreConfig,
   EmitsFromStoreConfig,
   ContextFromStoreConfig


svemat01:
Might also be worth considering adding in the [`storeConfigToLogic`](https://github.com/statelyai/xstate/commit/e3094b28ffae126686e3fe9ff9677ed54d092443#diff-0f16c71bcb524f1111dd46d0023ebef72c30cff6d1bf34369dc407dd22c57336) util from the persist branch so that custom middleware creators don't have to copy paste it.

Happy to add it in here as well or submit it as another PR if it's wanted

davidkpiano:
> Might also be worth considering adding in the [`storeConfigToLogic`](https://github.com/statelyai/xstate/commit/e3094b28ffae126686e3fe9ff9677ed54d092443#diff-0f16c71bcb524f1111dd46d0023ebef72c30cff6d1bf34369dc407dd22c57336) util from the persist branch so that custom middleware creators don't have to copy paste it.
> 
> Happy to add it in here as well or submit it as another PR if it's wanted

Sure, you can expose that one too.

Can you also add a changeset? Just `pnpm changeset` and let's make it a minor.

svemat01:
Just to clarify, that utility does not currently exist, it was only introduced in your persist branch so adding it here would include both adding that utility and potentially updating the undo helper to use it. So not just a quick extra export

packages/xstate-store/src/index.ts
```diff
@@ -3,7 +3,8 @@ export { fromStore } from './fromStore';
 export {
   createStore,
   createStoreWithProducer,
-  createStoreConfig
+  createStoreConfig,
+  createStoreTransition
```
Andarist, line 7: could you add some basic tests for this one? one test that would show (roughly) how you plan to use it would be helpful

svemat01: My actual usage is just copy pasting the persist utility from David's branch. Maybe the better course of action here is to maybe get that branch merged instead with perhaps these extra utils exposed should someone else want to build a custom utility as well?

Andarist: If you are copy-pasting from the other branch then let's not reuse that sample as a test case for this exposed util, a new one (even a rather dummy one) should be added to tests. At the moment, this util is indirectly tested because other things use it but when exposing it as part of the public API we should have an extra smoke test for this new public interface~

PR #5387, opened 2025-10-07
Get list of currently running actors in system

farskid:
This PR adds `system.getAll` that returns a record of running actors within the system by their system id.
This is very helpful for a tool to visualize active actors in a system given the system reference.

### DIFF
--- a/packages/core/src/system.ts
+++ b/packages/core/src/system.ts
@@ -187,7 +187,8 @@ export function createSystem<T extends ActorSystemInfo>(
       }
     },
     get: (systemId) => {
-      return keyedActors.get(systemId) as T['actors'][any] | undefined
+      return keyedActors.get(systemId) as T['actors'][any] | undefined;
+    },
     getAll: () => {
       return Object.fromEntries(keyedActors.entries()) as Partial<T['actors']>;
     },
--- a/packages/core/src/system.ts
+++ b/packages/core/src/system.ts
@@ -55,6 +55,7 @@ export interface ActorSystem<T extends ActorSystemInfo> {
   /** @internal */
   _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
   get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+  getRunningActors: () => T['actors'];
 
   inspect: (
     observer:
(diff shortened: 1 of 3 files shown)


packages/core/src/system.ts
```diff
@@ -188,6 +189,9 @@ export function createSystem<T extends ActorSystemInfo>(
     get: (systemId) => {
       return keyedActors.get(systemId) as T['actors'][any];
     },
+    getRunningActors: () => {
```
Andarist:
```suggestion
    getAll: () => {
```

farskid: Does system.get only Work on actively running actors?

Andarist: It should return all registered ones and we have to deregister them appropriately too - otherwise we'd have a memory leak.

farskid: Ok. so it matches with what this PR indents for all actors. I'll rename 👍

packages/core/src/system.ts
```diff
@@ -55,6 +55,7 @@ export interface ActorSystem<T extends ActorSystemInfo> {
   /** @internal */
   _set: <K extends keyof T['actors']>(key: K, actorRef: T['actors'][K]) => void;
   get: <K extends keyof T['actors']>(key: K) => T['actors'][K] | undefined;
+  getAll: () => T['actors'];
```
Andarist:
```suggestion
  getAll: () => Partial<T['actors']>;
```

packages/core/src/system.ts
```diff
@@ -188,6 +189,9 @@ export function createSystem<T extends ActorSystemInfo>(
     get: (systemId) => {
       return keyedActors.get(systemId) as T['actors'][any];
     },
+    getAll: () => {
+      return Object.fromEntries(keyedActors.entries()) as T['actors'];
```
Andarist:
```suggestion
      return Object.fromEntries(keyedActors.entries()) as Partial<T['actors']>;
```

packages/core/src/system.ts
```diff
@@ -188,6 +189,9 @@ export function createSystem<T extends ActorSystemInfo>(
     get: (systemId) => {
       return keyedActors.get(systemId) as T['actors'][any];
```
Andarist:
```suggestion
      return keyedActors.get(systemId) as T['actors'][any] | undefined;
```

Andarist (approved):

PR #5391, opened 2025-10-09
Expose subscription as an inspection event

farskid:
Exposing subscriptions and the observer id through inspection protocol can help map out actor subscriptions. This can come handy in mapping out subscription hierarchy.

### DIFF
--- a/.changeset/many-teeth-cough.md
+++ b/.changeset/many-teeth-cough.md
@@ -1,0 +1,30 @@
+---
+'xstate': minor
+---
+
+Exposes actor subscription as an inspection event.
+
+```ts
+import { createMachine, createActor } from 'xstate';
+
+const machine = createMachine({
+  // ...
+});
+
+const actor = createActor(machine, {
+  inspect: (event) => {
+    // event.type === '@xstate.subscription'
+    // event.actorRef === actor to which the subscription belongs to
+    // event.subscriptionId === 'my-observer-id'
+  }
+});
+
+actor.subscribe(
+  (snapshot) => {
+    console.log(snapshot);
+  },
+  undefined,
+  undefined,
+  'my-observer-id'
+);
+```


packages/core/src/createActor.ts
```diff
@@ -415,23 +415,31 @@ export class Actor<TLogic extends AnyActorLogic>
   public subscribe(
     nextListener?: (snapshot: SnapshotFrom<TLogic>) => void,
     errorListener?: (error: any) => void,
-    completeListener?: () => void
+    completeListener?: () => void,
+    observerId?: string
```
Andarist, line 419: I don't quite like this addition. `subscribe`'s signature is pseudo-standard (with Observables hopefully coming some day to the platform, see the current efforts [here](https://wicg.github.io/observable/)). I would prefer not to extend it in any custom/new way.

farskid: Sure but how do we identify subscribers then? Any ideas as to where this id can live?

farskid: Maybe we could pass an id as the subscription parameter?
```
actor.subscribe(observer, {id})
```

farskid: cc @davidkpiano

Andarist: > Maybe we could pass an id as the subscription parameter?

This is exactly the same thing ;p

It would be great if you could come up with a description of he exact problem you are trying to solve here and include that in the PR's description. The current description is somewhat vague. I don't really understand what subscription hierarchies are.

farskid: You're right. I thought I was passing it as a property of the observer.

The problem this will solve is to identify the subscribers to an actor. A tool can use the inspect protocol and map out subscribers of every actor. This tool for instance can be used to map react component tree to actor subscriptions to visualize subscriptions at scale. Helps  identify unnecessary subscriptions and places to improve. This was just an example.

PR #5392, opened 2025-10-13
[@xstate/store] Add overload to select entire store snapshot (identity selector)

davidkpiano:
Added an overload to `useSelector` that allows you to select the entire snapshot:

```ts
// No selector provided, return the entire snapshot
const snapshot = useSelector(store);
```


davidkpiano:
Ping @Andarist @farskid

PR #5393, opened 2025-10-13
[@xstate/store] Skip events from undo/redo

davidkpiano:
Add `skipEvent` option to `undoRedo()` to exclude certain events from undo/redo history.

```ts
const store = createStore(
  undoRedo(
    {
      context: { count: 0 },
      on: {
        inc: (ctx) => ({ count: ctx.count + 1 }),
        log: (ctx) => ctx // No state change
      }
    },
    {
      skipEvent: (event, snapshot) => event.type === 'log'
    }
  )
);
```

---

Add `snapshot` parameter to `getTransactionId` function.

```ts
const store = createStore(
  undo(
    {
      // ...
    },
    {
      getTransactionId: (event, snapshot) =>
        snapshot.context.currentTransactionId
    }
  )
);
```

### DIFF
--- a/packages/xstate-store/src/undo.ts
+++ b/packages/xstate-store/src/undo.ts
@@ -84,7 +84,7 @@ type UndoEvent<TEvent extends EventObject> = {
  *       }
  *     },
  *     {
- *       skipEvents: (event) => event.type === 'log'
+ *       skipEvent: (event) => event.type === 'log'
  *     }
  *   )
  * );
--- a/.changeset/tangy-spies-study.md
+++ b/.changeset/tangy-spies-study.md
@@ -2,7 +2,7 @@
 '@xstate/store': minor
 ---
 
-Add `skipEvents` option to `undoRedo()` to exclude certain events from undo/redo history.
+Add `skipEvent` option to `undoRedo()` to exclude certain events from undo/redo history.
 
 ```ts
 const store = createStore(
@@ -15,7 +15,7 @@ const store = createStore(
       }
     },
     {
-      skipEvents: (event, snapshot) => event.type === 'log'
+      skipEvent: (event, snapshot) => event.type === 'log'
     }
   )
 );
--- a/packages/xstate-store/test/undo.test.ts
+++ b/packages/xstate-store/test/undo.test.ts
@@ -436,7 +436,7 @@ it('should skip events with transaction grouping', () => {
   expect(store.getSnapshot().context.count).toBe(7);
 });
 
-it('should use the snapshot in the skipEvents function', () => {
+it('should use the snapshot in the skipEvent function', () => {
   const store = createStore(
     undoRedo(
       {


packages/xstate-store/src/undo.ts
```diff
@@ -111,14 +143,7 @@ export function undoRedo<
         const events = snapshot.events.slice();
         const undoStack = snapshot.undoStack.slice();
         if (!events.length) {
```
Andarist: this should check
```suggestion
        if (!snapshot.events.length) {
```
and the `.slice()` calls should be moved below this check

packages/xstate-store/src/undo.ts
```diff
@@ -148,14 +173,20 @@ export function undoRedo<
           }
         }
 
+        // Filter out events that should be skipped during undo
+        const eventsToReplay = events.filter(
+          (undoEvent: UndoEvent<TEvent>) =>
+            !options?.skipEvents?.(undoEvent.event)
+        );
```
Andarist:
```suggestion
        const eventsToReplay = options?.skipEvents ? events.filter(
          (undoEvent: UndoEvent<TEvent>) =>
            !options.skipEvents(undoEvent.event)
        ) : events;
```

packages/xstate-store/src/undo.ts
```diff
@@ -207,15 +244,19 @@ export function undoRedo<
         return [state, allEffects];
       }
 
-      const events = snapshot.events.slice();
       const [state, effects] = logic.transition(snapshot, event);
+      const isEventSkipped = options?.skipEvents?.(event);
+      const events = isEventSkipped
+        ? snapshot.events
+        : snapshot.events.slice().concat({
```
Andarist:
```suggestion
        : snapshot.events.concat({
```

farskid:
@davidkpiano maybe `skipEvents` can access context too?

davidkpiano:
> @davidkpiano maybe `skipEvents` can access context too?

I've thought of this too but what use-case are you thinking?

farskid:
> > @davidkpiano maybe `skipEvents` can access context too?
> 
> I've thought of this too but what use-case are you thinking?

Depends on the undo/redo API and how transaction ids are passed. Do you have an example of that?

davidkpiano:
> Depends on the undo/redo API and how transaction ids are passed. Do you have an example of that?

```ts
const store = createStore(
  undoRedo(
    {
      context: { count: 0 },
      on: {
        inc: (ctx, ev: { t?: string }) => ({ count: ctx.count + 1 }),
        dec: (ctx) => ({ count: ctx.count - 1 })
      }
    },
    {
      getTransactionId: (event) => {
        return event.t;
      }
    }
  )
);

store.trigger.inc(); // not part of a transaction

store.trigger.inc({ t: 'a' }); // part of 'a' transaction
store.trigger.inc({ t: 'a' }); // part of 'a' transaction
store.trigger.inc({ t: 'a' }); // part of 'a' transaction

store.trigger.inc({ t: 'b' }); // part of 'b' transaction
store.trigger.inc({ t: 'b' }); // part of 'b' transaction
store.trigger.inc({ t: 'b' }); // part of 'b' transaction

store.trigger.undo(); // undoes three events from 'b' transaction

store.trigger.undo(); // undoes three events from 'a' transaction

store.trigger.undo(); // undoes a single event
```

davidkpiano:
@Andarist @farskid Ping

packages/xstate-store/src/undo.ts
```diff
@@ -79,7 +104,20 @@ export function undoRedo<
...
+    ) => string | null | undefined;
+    /**
+     * A function that returns whether an event should be skipped during
+     * undo/redo. Skipped events are not stored in history and are not replayed
+     * during undo/redo.
+     */
+    skipEvent?: (
+      event: ExtractEvents<TEventPayloadMap>,
+      snapshot: StoreSnapshot<TContext>
+    ) => boolean;
```
farskid, line 120: nit: event actions API has order `ctx, event` whereas these two functions have the reverse order. They primarily work based on events which makes sense for event to be the first parameter but for consistency, we might want to consider unifying the order.

farskid (approved):
Overall LGTM

packages/xstate-store/src/undo.ts
davidkpiano: That'd be a breaking change

PR #5398, opened 2025-10-23
Use `vitest run`

davidkpiano:
The current test scripts run `vitest` which runs in watch mode. It's a bit of an annoying default especially because AI coding agents use those test commands and expect them to run once.

Now `pnpm test` and `pnpm test:core` will run tests once. If watch mode is still desired, `pnpm test:watch` and `pnpm test:core:watch` will do the trick.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,9 @@
     "lint": "eslint --cache --quiet",
     "typecheck": "tsc",
     "test": "vitest run",
+    "test:watch": "vitest",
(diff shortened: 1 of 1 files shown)


davidkpiano:
cc. @with-heart

with-heart:
That's an interesting lil change, hadn't really thought about _how_ some scripts run would affect ai but definitely makes sense

Andarist (approved):

PR #5401, opened 2025-10-25
[@xstate/store] Fix emit event types

davidkpiano:
Fix type inference for emitted event types when using `undoRedo`.


davidkpiano:
Merging - trivial fix

PR #5403, opened 2025-10-26
[core] Async iterator

davidkpiano:
Make actors async iterable:

```ts
actor.start();

for await (const snapshot of actor) {
  console.log(snapshot);
}
```

### DIFF
--- a/packages/core/src/createActor.ts
+++ b/packages/core/src/createActor.ts
@@ -500,7 +500,7 @@ export class Actor<TLogic extends AnyActorLogic>
             subscription.unsubscribe();
           },
           error: (error) => {
-            reject(error instanceof Error ? error : new Error(String(error)));
+            reject(error);
             subscription.unsubscribe();
           },
           complete: () => {


packages/core/src/createActor.ts
```diff
@@ -464,6 +464,55 @@ export class Actor<TLogic extends AnyActorLogic>
...
+   *     break; // Optional: break when done
+   *   }
+   * }
+   * ```
+   */
+  async *[Symbol.asyncIterator](): AsyncIterator<SnapshotFrom<TLogic>> {
+    // Yield the initial snapshot if the actor is already running
+    if (this._processingStatus === ProcessingStatus.Running) {
+      yield this.getSnapshot();
+    }
```
Andarist, line 493: does this match the behavior of vanilla `actorRef.subscribe`?

packages/core/src/createActor.ts
```diff
@@ -464,6 +464,55 @@ export class Actor<TLogic extends AnyActorLogic>
...
+
+    while (this._processingStatus !== ProcessingStatus.Stopped) {
+      yield await new Promise<SnapshotFrom<TLogic>>((resolve, reject) => {
+        const subscription = this.subscribe({
+          next: (snapshot) => {
+            resolve(snapshot);
+            subscription.unsubscribe();
+          },
+          error: (error) => {
+            reject(error instanceof Error ? error : new Error(String(error)));
```
Andarist: we shouldn't meddle with the thrown value - just pass it through
```suggestion
            reject(error);
```

packages/core/src/createActor.ts
```diff
@@ -464,6 +464,55 @@ export class Actor<TLogic extends AnyActorLogic>
...
+   * ```
+   */
+  async *[Symbol.asyncIterator](): AsyncIterator<SnapshotFrom<TLogic>> {
+    // Yield the initial snapshot if the actor is already running
+    if (this._processingStatus === ProcessingStatus.Running) {
+      yield this.getSnapshot();
+    }
+
+    while (this._processingStatus !== ProcessingStatus.Stopped) {
+      yield await new Promise<SnapshotFrom<TLogic>>((resolve, reject) => {
```
Andarist, line 496: This implementation is neat but I think we should do better in the library context. This resubscribes with each iteration step. It would be better to maintain a single subscription throughout the full lifecycle of the iteration.

packages/core/test/actor.test.ts
```diff
@@ -1821,20 +1821,118 @@ describe('actors', () => {
...
+
+    expect(snapshots).toHaveLength(3);
+    expect(snapshots[0].status).toBe('active');
+    expect(snapshots[0].value).toBe('idle');
+    expect(snapshots[1].status).toBe('active');
+    expect(snapshots[1].value).toBe('active');
+    expect(snapshots[2].status).toBe('done');
+  });
+
+  it('should handle errors in async iteration', async () => {
```
Andarist, line 1872: this doesn't test what it claims - this just happens to use events and states named with "error" but in reality this machine completes gracefully through a regular final state

PR #5406, opened 2025-11-05
[core] Add `getNextTransitions(state)`

davidkpiano:
Add `getNextTransitions(state)` utility to get all transitions available from current `state`.

```ts
import { getNextTransitions } from 'xstate';

// ...

const state = actor.getSnapshot();
const transitions = getNextTransitions(state);

transitions.forEach((t) => {
  console.log(`Event: ${t.eventType}, Source: ${t.source.key}`);
});
```

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -38,7 +38,7 @@ export {
 export {
   transition,
   initialTransition,
-  getPotentialTransitions
+  getNextTransitions
 } from './transition.ts';
 export { waitFor } from './waitFor.ts';
 
--- a/packages/core/src/transition.ts
+++ b/packages/core/src/transition.ts
@@ -85,7 +85,7 @@ export function initialTransition<T extends AnyActorLogic>(
  * @returns Array of transition definitions from the current state, in
  *   deterministic order
  */
-export function getPotentialTransitions(
+export function getNextTransitions(
   state: AnyMachineSnapshot
 ): AnyTransitionDefinition[] {
   const potentialTransitions: AnyTransitionDefinition[] = [];
--- a/.changeset/slow-horses-stop.md
+++ b/.changeset/slow-horses-stop.md
@@ -2,15 +2,15 @@
 'xstate': minor
 ---
 
-Add `getPotentialTransitions(state)` utility to get all transitions available from current `state`.
+Add `getNextTransitions(state)` utility to get all transitions available from current `state`.
 
 ```ts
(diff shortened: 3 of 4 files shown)


packages/core/src/index.ts
```diff
@@ -35,7 +35,11 @@ export {
   pathToStateValue,
   toObserver
 } from './utils.ts';
-export { transition, initialTransition } from './transition.ts';
+export {
+  transition,
+  initialTransition,
+  getPotentialTransitions
```
Andarist: a rather similar utility is still exported as `__unsafe_getAllOwnEventDescriptors`, i would be good to maintain the naming scheme or make the other one "safe"

what do u want to use this for externally?

packages/core/src/transition.ts
```diff
@@ -57,3 +60,41 @@ export function initialTransition<T extends AnyActorLogic>(
...
+      }
+
+      // Also include always (eventless) transitions
+      if (s.always) {
+        potentialTransitions.push(...s.always);
+      }
+    }
+  }
+
+  return potentialTransitions;
```
Andarist, line 122: what should be the order of transitions in the output? especially when we consider nested states and parallel ones. I see you are using order-insensitive assertions in the tests but I think it would be reasonable to specify that order and assert on it.

packages/core/src/transition.ts
```diff
@@ -57,3 +60,41 @@ export function initialTransition<T extends AnyActorLogic>(
...
+  // Collect all transitions from atomic states and their ancestors
+  for (const stateNode of atomicStates) {
+    for (const s of [stateNode].concat(
+      getProperAncestors(stateNode, undefined)
+    )) {
+      // Get all transitions for each event type
+      for (const [eventType, transitions] of s.transitions) {
+        if (processedEventTypes.has(eventType)) {
+          continue;
+        }
```
Andarist: is this correct? wouldnt it skip guarded non-first guarded transitions that are actually "potential" (could very well be taken)?

packages/core/src/transition.ts
davidkpiano: Added ordering assertions

packages/core/src/index.ts
davidkpiano: I want to use this one (`getPotentialTransitions(…)`) as it includes more useful information than `__unsafe_getAllOwnEventDescriptors`; we can deprecate that one

PR #5408, opened 2025-11-09
Fix memory leaks in actor cleanup

GrigoryEvko:
Fixes memory leaks in actor lifecycle cleanup by ensuring all internal state is properly cleared when actors are stopped or complete.

## Changes

- Clear `eventListeners` Map in `Actor._complete()` and `Actor._reportError()`
- Delete `controllerMap` entries in promise actors on `XSTATE_STOP`
- Clear `receivers` Set and delete `instanceStates` entries in callback actors on `XSTATE_STOP`

## Testing

All existing tests pass (93 test files, 2046 tests passing, 100%).

### DIFF
--- a/packages/core/src/actors/callback.ts
+++ b/packages/core/src/actors/callback.ts
@@ -230,9 +230,9 @@ export function fromCallback<
           error: undefined
         };
 
-        callbackState.dispose?.();
-        callbackState.receivers?.clear();
         instanceStates.delete(actorScope.self);
+        callbackState.receivers?.clear();
+        callbackState.dispose?.();
         return state;
       }
 


davidkpiano:
Nice, can you add a changeset? Just `pnpm changeset` and follow the instructions (patch)

packages/core/src/actors/callback.ts
```diff
@@ -231,6 +231,8 @@ export function fromCallback<
         };
 
         callbackState.dispose?.();
+        callbackState.receivers?.clear();
+        instanceStates.delete(actorScope.self);
```
Andarist, line 233: I feel it's always better to clear up the data before calling clean up functions (if only possible) - because those functions could potentially throw.
```suggestion
        instanceStates.delete(actorScope.self);
        callbackState.receivers?.clear();
        callbackState.dispose?.();
```

PR #5414, opened 2025-11-24
[@xstate/store] Computed atom previous value access

davidkpiano:
Computed atoms can now access their previous value via an optional second parameter:

```ts
const count = createAtom(1);
const double = createAtom<number>((_, prev) => count.get() + (prev ?? 0));
```

### DIFF
--- a/.changeset/whole-days-design.md
+++ b/.changeset/whole-days-design.md
@@ -6,5 +6,5 @@ Computed atoms can now access their previous value via an optional second parame
 
 ```ts
 const count = createAtom(1);
-const double = createAtom<number>((read, prev) => read(count) + (prev ?? 0));
+const double = createAtom<number>((_, prev) => count.get() + (prev ?? 0));
 ```


.changeset/whole-days-design.md
```diff
@@ -0,0 +1,10 @@
+---
+'@xstate/store': minor
+---
+
+Computed atoms can now access their previous value via an optional second parameter:
+
+```ts
+const count = createAtom(1);
+const double = createAtom<number>((_, prev) => count.get() + (prev ?? 0));
```
Andarist, line 9: out of curiosity - which other atom-based libraries offer this functionality in their computeds?
also, shouldn't this be using `read(count)` to track the read atom properly? 🤔

davidkpiano: Deprecating `read` - using `atom.get()` should be the API moving forward

Vue does this: https://vuejs.org/guide/essentials/computed#previous

```vue
<script setup>
import { ref, computed } from 'vue'

const count = ref(2)

// This computed will return the value of count when it's less or equal to 3.
// When count is >=4, the last value that fulfilled our condition will be returned
// instead until count is less or equal to 3
const alwaysSmall = computed((previous) => {
  if (count.value <= 3) {
    return count.value
  }

  return previous
})
</script>
```

Andarist: It feels like a major API change if the first parameter becomes deprecated. Wouldn't it be better to just cut a new major release with this breaking change?

davidkpiano: Yeah, that makes sense - there's a few more things I'd like to get into the next major version, so I'll hold off on this PR.

EDIT: Let's have it in there for now as the 2nd arg (no harm, just ignore the 1st arg) and revisit in the next major.

PR #5417, opened 2025-12-02
Rework some types slightly to remove workaround targeting `silentNeverType`

Andarist:


packages/core/src/setup.ts
```diff
@@ -38,43 +38,29 @@ type ToParameterizedObject<
...
-  IsNever<TParameterizedMap> extends true
-    ? never
-    : Values<{
-        [K in keyof TParameterizedMap & string]: {
-          type: K;
-          params: TParameterizedMap[K];
-        };
-      }>;
+> = Values<{
+  [K in keyof TParameterizedMap as K & string]: {
```
Andarist, line 42: This ensures this mapped type is homomorphic and thus `silentNeverType` gets automatically~ handled by TS itself:
```ts
type A<T> = {
  [K in keyof T]: {
    type: K;
    params: T[K];
  };
};

type ResultA = A<never>;
//   ^? type ResultA = never


type B<T> = {
  [K in keyof T & string]: {
    type: K;
    params: T[K];
  };
};

type ResultB = B<never>;
//   ^? type ResultB = { [x: string]: { type: string; params: never; }; }

type C<T> = {
  [K in keyof T as K & string]: {
    type: K;
    params: T[K];
  };
};

type ResultC = C<never>;
//   ^? type ResultC = never
```

Basically, we had `B` here when we should/could use `C`

PR #5420, opened 2025-12-03
fix possible 'not iterable' bug

assertnotnull:
I tried to add a test with an empty Map in many ways to reproduce it but it only happens in mobile (at least a Meteor on Cordova app)
Capture d’écran, le 2025-12-03 à 13 25 44

### DIFF
--- a/.changeset/brave-points-film.md
+++ b/.changeset/brave-points-film.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Fix a bug in Cordova when iterating an empty Map


davidkpiano:
Yeah, my guess is that it's specific to Cordova, but the fix is fine since it's equivalent. 

Can you add a changeset? Just `pnpm changeset` and make it a `patch` with a short description. Thanks very much!

assertnotnull:
It only asked me for major. Is there something to do prior?

assertnotnull:
I've figured I think.

PR #5422, opened 2025-12-04
[core] Add support for partial event descriptors in `assertEvent(…)`

davidkpiano:
Add partial descriptor support to `assertEvent(…)`

```ts
// Matches any event with a type that starts with `FEEDBACK.`
assertEvent(event, 'FEEDBACK.*');
```

### DIFF
--- a/packages/core/src/assert.ts
+++ b/packages/core/src/assert.ts
@@ -40,8 +40,8 @@ export function assertEvent<
   if (!matches) {
     const typesText =
       types.length === 1
-        ? `type "${types[0]}"`
-        : `one of types "${types.join('", "')}"`;
+        ? `type matching "${types[0]}"`
+        : `one of types matching "${types.join('", "')}"`;
     throw new Error(
       `Expected event ${JSON.stringify(event)} to have ${typesText}`
     );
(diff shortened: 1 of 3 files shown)


packages/core/src/assert.ts
```diff
@@ -42,3 +49,56 @@ export function assertEvent<
     );
   }
 }
+
+function matchesEventDescriptor(
```
Andarist: don't we already have a good chunk of this in `getCandidates`? it would be great if this logic would be shared between both places

davidkpiano: [289a01d](https://github.com/statelyai/xstate/pull/5422/commits/289a01dce94a907ff961b8d0f6fb0dcb175e05f4)

packages/core/test/eventDescriptors.test.ts
```diff
@@ -354,4 +354,83 @@ describe('event descriptors', () => {
...
+          rate: number;
+        }
+      | { type: 'OTHER' };
+
+    const nonFeedbackEvent = { type: 'OTHER' } as FeedbackEvents;
+
+    expect(() =>
+      assertEvent(nonFeedbackEvent, 'FEEDBACK.*')
+    ).toThrowErrorMatchingInlineSnapshot(
+      `[Error: Expected event {"type":"OTHER"} to have type "FEEDBACK.*"]`
```
Andarist: This probably should be closer to:
```suggestion
      `[Error: Expected event {"type":"OTHER"} to have a type matching "FEEDBACK.*"]`
```

PR #5430, opened 2025-12-11
[core] Add max iteration count for detecting infinite loops

davidkpiano:
Add `maxIterations` option to configure the maximum number of microsteps allowed before throwing an infinite loop error. The default is `Infinity` (no limit) to avoid breaking existing machines.
You can configure it when creating a machine:
```ts
  // ... machine config
  options: {
    maxIterations: 1000 // set a limit to enable infinite loop detection
  }
});
```


packages/core/test/transient.test.ts
```diff
@@ -566,6 +566,81 @@ describe('transient states (eventless transitions)', () => {
...
+        },
+        c: {
+          always: {
+            target: 'a'
+          }
+        }
+      },
+      options: {
+        maxIterations: 100
+      }
```
Andarist, line 631: It's yet another place when some options are accepted. I'm not saying this is inherently bad but I can see how this might not be that discoverable given this is **the only** option accepted this way now.

Andarist: I also wonder, shouldn't this kinda be a global option really? As a user, I don't see myself using this *everywhere* - and currently I'd have to because it's opt-in. 

Some of other libraries (like even React) just implements such protections automatically and they don't allow the user to confgure the "depth count" that triggers the error.

packages/core/src/StateMachine.ts
```diff
@@ -135,6 +139,10 @@ export class StateMachine<
     };
     this.version = this.config.version;
     this.schemas = this.config.schemas;
+    this.options = {
+      maxIterations: 1000,
```
Andarist: The changeset file mentions:
> The default is `Infinity` (no limit) to avoid breaking existing machines.

but clearly it's not true as the default here is actually set to 1000 at runtime.

packages/core/src/stateUtils.ts
```diff
@@ -1648,8 +1648,18 @@ export function macrostep(
   }
 
   let shouldSelectEventlessTransitions = true;
+  const maxIterations = snapshot.machine.options?.maxIterations ?? Infinity;
+  const hasMaxIterations = maxIterations !== Infinity && maxIterations !== -1;
```
Andarist: I don't think handling 2 values like this makes sense. If you need a value representing some special situation then make one value represent that situation. And also - if `-1` means "disabled"... what about any other negative number?

davidkpiano: I'll stick to infinity

packages/core/test/transient.test.ts
davidkpiano: There may be good reasons for iterations to go > 100; think any while loop that needs to loop N times where N may be > 100 but still finite. I just don't want to break anything right now.

PR #5440, opened 2026-01-03
[core] Fix systemId cleanup for nested children on stopChild

### LINKED ISSUE #5439: Bug: System ID conflict
### XState version

XState version 5

### Description

From Discord: https://discord.com/channels/795785288994652170/1456673826439041275

>  Is it expected that a child actor can't be stopped and immediately started again, if system ids (in actors invoked by the child actor) are used? I get a "Actor with system ID 'foo'  already exists. I need to delay the creation of the actor (and without a started actor I can't render my UI, so I have to render a fallback content and delay the actor creation even further to avoid flickering).

### Expected result

System ID reused

### Actual result

Error

### Reproduction

https://stackblitz.com/~/edit/vitejs-vite-whwbcqra?file=src/main.ts

### Additional context

_No response_

davidkpiano:
Fixes #5439

### DIFF
--- a/.changeset/six-trees-hide.md
+++ b/.changeset/six-trees-hide.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+Fix `systemId` cleanup for nested children on `stopChild`


packages/core/src/actions/stopChild.ts
```diff
@@ -52,6 +52,22 @@ function resolveStop(
     undefined
   ];
 }
+function unregisterRecursively(
+  actorScope: AnyActorScope,
+  actorRef: AnyActorRef
+) {
+  // unregister children first (depth-first)
```
Andarist, line 59: I don't see a reason why depth-first would be important here, unregistering is a simple logic that doesn't execute anythign extra. But we have to use some traversing kind - so it can be depth-first (it's just that visit order isn't important)

davidkpiano: Yeah, but just to be safe, it makes the most sense (to me)

PR #5441, opened 2026-01-05
[@xstate/store] Store framework adapters

davidkpiano:
This PR includes adapters for:

- @xstate/store-react 
- @xstate/store-solid
- @xstate/store-vue 
- @xstate/store-svelte 
- @xstate/store-preact 
- @xstate/store-angular

### DIFF
--- a/.changeset/lemon-news-design.md
+++ b/.changeset/lemon-news-design.md
@@ -2,4 +2,4 @@
 '@xstate/store-vue': major
 ---
 
-Initial release for `@xstate/store-vue`. Re-exports `@xstate/store`.
+Initial release for `@xstate/store-vue`.
--- a/.changeset/busy-webs-listen.md
+++ b/.changeset/busy-webs-listen.md
@@ -2,4 +2,4 @@
 '@xstate/store-react': major
 ---
 
-Initial release for `@xstate/store-react`. Re-exports `@xstate/store`.
+Initial release for `@xstate/store-react`.
--- a/.changeset/fruity-aliens-knock.md
+++ b/.changeset/fruity-aliens-knock.md
@@ -2,4 +2,4 @@
 '@xstate/store-svelte': major
 ---
 
-Initial release for `@xstate/store-svelte`. Re-exports `@xstate/store`.
+Initial release for `@xstate/store-svelte`.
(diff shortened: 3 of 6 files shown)


packages/xstate-store/src/react.ts
```diff
@@ -1,3 +1,9 @@
+console.warn(
```
Andarist: perhaps it would be best to wrap this with an env check to avoid spamming the console in deployed sites?

packages/xstate-store/src/solid.ts
```diff
@@ -1,4 +1,10 @@
 /* @jsxImportSource solid-js */
+console.warn(
```
Andarist: same here

scripts/convert-send-to-trigger.js
Andarist, line 1: isn't this a one-off script that shouldnt be needed with time? a long-term thing would be better suited by a lint rule than a committed conversion script like this one

Andarist: It also feels like this might have converted all `.send`'s but we should keep some in tests as long as that API is available

.changeset/quick-waves-pull.md
Andarist, line 1: the text of the changeset is currently written from the perspective of the whole repo but this will be copy-pasted into each individual changelog file and it won't quite make sense in each target location. For example, why `@xstate/store-vue/CHANGELOG.md` would mention that `@xstate/store/react` was deprecated? 

The new packages should get a dedicated changeset with "Initial release" or something like that. One changeset with all of the relevant packages will suffice if the text won't mention any of them by name. You could also just ignore adding a changeset file for those new packages altogether. The action will publish them just fine given it's their initial release.

You have also missed here a dedicated changeset for the `@xstate/store`. That's the one that should mention deprecation of those `/react` and `/solid` entrypoints.

.changeset/lemon-news-design.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/store-vue': major
+---
+
+Initial release for `@xstate/store-vue`. Re-exports `@xstate/store`.
```
Andarist: "reexports" sounds like it's just doing `export * from '@xstate/store'` and nothing else which would mean this reexporting package isn't even needed. For simplicity, I'd just drop this second sentence altogether

GraxMonzo:
It looks like the new adapters haven’t been published to npm yet.

PR #5446, opened 2026-01-13
feat(core): add actions, guards, delays to SetupTypes interface

cevr:
## Summary

- Adds type slots for `actions`, `guards`, and `delays` to the `SetupTypes` interface
- Enables explicit type annotations for machines, helping break TypeScript inference chains
- Addresses performance issues and "excessively large and possibly infinite" type errors in large codebases

## Problem

In large codebases that share types between backend and frontend, TypeScript's inference on actions/guards/delays can cause:
- Significant slowdowns in type checking
- "Excessively large and possibly infinite" TypeScript errors
- Inability to explicitly type machines like `const machine: StateMachine<...> = setup({...}).createMachine({})`

## Solution

Add `actions`, `guards`, and `delays` to `SetupTypes` (like the existing `context`, `events`, etc.). This allows users to declare types explicitly rather than relying solely on inference.

```typescript
// Before: Types always inferred from implementations, causing deep inference chains
const { createMachine } = setup({
  actions: {
    increment: assign(({ context }) => ({ count: context.count + 1 })),
  }
});

// After: SetupTypes includes action/guard/delay slots for explicit annotation support
type MySetupTypes = {
  context: { count: number };
  events: { type: 'INC' };
  actions: { type: 'increment'; params: undefined };
  guards: { type: 'isValid'; params: undefined };
  delays: 'timeout';
};
```

## Changes

- `types.ts`: Add `TAction`, `TGuard`, `TDelay` type parameters and properties to `SetupTypes`
- `setup.ts`: Pass action/guard/delay types to `SetupTypes` in the `types` parameter
- `setup.types.test.ts`: Add tests for new type slots

## Test plan

- [x] TypeScript typecheck passes
- [x] All 159 setup type tests pass
- [x] All 1687 tests in the core package pass

### DIFF
--- a/packages/core/src/setup.ts
+++ b/packages/core/src/setup.ts
@@ -359,7 +359,10 @@ export function setup<
     TInput,
     TOutput,
     TEmitted,
-    TMeta
+    TMeta,
+    ToParameterizedObject<TActions>,
+    ToParameterizedObject<TGuards>,
+    TDelay
   >;
   actors?: {
     // union here enforces that all configured children have to be provided in actors
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1396,7 +1396,10 @@ export interface SetupTypes<
   TInput,
   TOutput,
   TEmitted extends EventObject,
-  TMeta extends MetaObject
+  TMeta extends MetaObject,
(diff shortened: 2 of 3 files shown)


davidkpiano:
cc. @Andarist to take a quick look to see if this is needed

PR #5456, opened 2026-01-22
[@xstate/react] Expose error snapshots in useActor/useMachine

### LINKED ISSUE #5455: Bug: Can't handle errors thrown from state machine in React
### XState version

XState version 5

### Description

If I'm starting a state machine from React, and the state machine throws a synchronous error, I'm not able to display it.

I modified the starter StackBlitz example to throw from an action in the `submit` state: https://stackblitz.com/edit/github-zdh2xswa?file=src%2FApp.tsx,src%2FfeedbackMachine.ts

To repro, press "Bad", then enter some text in the form, then press "Submit".

In the `Feedback` component, I try to render the value of `state.error`, but the component never updates.

### Expected result

My expectation was that the component would update, and render the error string.

### Actual result

The component didn't update after the error was thrown.

### Reproduction

https://stackblitz.com/edit/github-zdh2xswa?file=src%2FApp.tsx,src%2FfeedbackMachine.ts

### Additional context

_No response_

davidkpiano:
`useActor` (and `useMachine` by alias) now properly re-renders when an action throws, exposing `status: 'error'` and `error` on the snapshot.

```tsx
const App = () => {
  // Machine that might throw an error, e.g.:
  // entry: () => throw new Error('error');
  const [state, send, actor] = useActor(machine);

  const errorMessage = state.status === 'error' ? (state.error as Error)?.message : null;
  
  // ...
};
```

Fixes #5455

### DIFF
--- a/packages/xstate-react/src/useActor.ts
+++ b/packages/xstate-react/src/useActor.ts
@@ -43,7 +43,10 @@ export function useActor<TLogic extends AnyActorLogic>(
 
   const subscribe = useCallback(
     (handleStoreChange: () => void) => {
-      const { unsubscribe } = actorRef.subscribe(handleStoreChange);
+      const { unsubscribe } = actorRef.subscribe({
+        next: handleStoreChange,
+        error: handleStoreChange
+      });
       return unsubscribe;
     },
     [actorRef]
--- a/.changeset/clear-islands-cheer.md
+++ b/.changeset/clear-islands-cheer.md
@@ -1,0 +1,18 @@
+---
+'@xstate/react': patch
+---
+
+`useActor` (and `useMachine` by alias) now properly re-renders when an action throws, exposing `status: 'error'` and `error` on the snapshot.
+
+```tsx
+const App = () => {
+  // Machine that might throw an error, e.g.:
+  // entry: () => throw new Error('error');
+  const [state, send, actor] = useActor(machine);
+
+  const errorMessage =
+    state.status === 'error' ? (state.error as Error)?.message : null;
+
+  // ...
+};
+```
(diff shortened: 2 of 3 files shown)


packages/xstate-react/src/useActor.ts
```diff
@@ -43,7 +43,10 @@ export function useActor<TLogic extends AnyActorLogic>(
 
   const subscribe = useCallback(
     (handleStoreChange: () => void) => {
-      const { unsubscribe } = actorRef.subscribe(handleStoreChange);
+      const { unsubscribe } = actorRef.subscribe({
```
Andarist, line 46: shouldnt this also have `complete: handleStoreChange`?

packages/xstate-react/src/useActor.ts
Andarist, line 1: this might sound controversial but I'd consider doing this in the `useActor`'s implementation:
```ts
if (actorSnapshot.status === 'error') throw actorSnapshot.error;
```

davidkpiano: I'm not opposed to this 🤔

sfc-gh-dperezalvarez:
Hi @davidkpiano @Andarist. Is there a way forward with this one?

davidkpiano:
> Hi @davidkpiano @Andarist. Is there a way forward with this one?

Maybe. Would this be a viable solution for you? Andarist's suggestion would likely be a breaking change, but maybe not (an error is an error)

sfc-gh-dperezalvarez:
> > Hi @davidkpiano @Andarist. Is there a way forward with this one?
> 
> Maybe. Would this be a viable solution for you? Andarist's suggestion would likely be a breaking change, but maybe not (an error is an error)

`useActor` throwing, and callers (me) catching in an error boundary? Yeah, perfectly reasonable. I just wanna be able to communicate to the user that something exploded, and a button to recover.

davidkpiano:
Superseded by https://github.com/statelyai/xstate/pull/5470

packages/xstate-react/src/useActor.ts
davidkpiano: https://github.com/statelyai/xstate/pull/5470

PR #5457, opened 2026-01-26
[core] `getInitialMicrosteps(…)` and `getMicrosteps(…)`

davidkpiano:
Add `getInitialMicrosteps(…)` and `getMicrosteps(…)` functions that return an array of `[snapshot, actions]` tuples for each microstep in a transition.

```ts
import { createMachine, getInitialMicrosteps, getMicrosteps } from 'xstate';

  initial: 'a',
  states: {
    a: {
      entry: () => console.log('enter a'),
      on: {
        NEXT: 'b'
      }
    },
    b: {
      entry: () => console.log('enter b'),
      always: 'c'
    },
    c: {}
  }
});

// Get microsteps from initial transition
const initialMicrosteps = getInitialMicrosteps(machine);
// Returns: [
//  [snapshotA, [entryActionA]]
// ]

// Get microsteps from a transition
const microsteps = getMicrosteps(machine, initialMicrosteps[0][0], {
  type: 'NEXT'
});
// Returns: [
//  [snapshotB, [entryActionB]],
//  [snapshotC, []]
// ]

// Each microstep is a tuple of [snapshot, actions]
for (const [snapshot, actions] of microsteps) {
  console.log('State:', snapshot.value);
  console.log('Actions:', actions.length);
}
```

### DIFF
--- a/packages/core/src/stateUtils.ts
+++ b/packages/core/src/stateUtils.ts
@@ -499,7 +499,7 @@ function getInitialStateNodesWithTheirAncestors(stateNode: AnyStateNode) {
   return states;
 }
 
-export function getInitialStateNodes(stateNode: AnyStateNode) {
+function getInitialStateNodes(stateNode: AnyStateNode) {
   const set = new Set<AnyStateNode>();
 
   function iter(descStateNode: AnyStateNode): void {
@@ -967,7 +967,7 @@ export function initialMicrostep(
 }
 
 /** https://www.w3.org/TR/scxml/#microstepProcedure */
-export function microstep(
+function microstep(
   transitions: Array<AnyTransitionDefinition>,
   currentSnapshot: AnyMachineSnapshot,
   actorScope: AnyActorScope,


packages/core/src/StateMachine.ts
```diff
@@ -328,7 +328,9 @@ export class StateMachine<
       TStateSchema
     >
   > {
-    return macrostep(snapshot, event, actorScope, []).microstates;
```
Andarist, line 331: this was already confusing before this PR but we have a `microstep` function that returns a *list*. I would expect it to be renamed or to return only a single microstep (as it is it's not consistent with the comment above the function)

packages/core/src/stateUtils.ts
```diff
@@ -945,88 +948,104 @@ export function microstep(
...
-  let historyValue = currentSnapshot.historyValue;
 
-  const filteredTransitions = removeConflictingTransitions(
-    transitions,
-    mutStateNodeSet,
-    historyValue
-  );
+  const originalExecutor = actorScope.actionExecutor;
+  actorScope.actionExecutor = (action) => {
+    actions.push(action);
```
Andarist, line 986: it's kinda a bummer that those are now collected *twice* when using `transition` - but I don't know how to expose the ones collected here to the `transition` so 🤷‍♂️

PR #5463, opened 2026-02-05
fix: Prevent transitions from final states within parallel regions

### LINKED ISSUE #5460: Bug: Parallel final states aren't actually final
### XState version

XState version 5

### Description

A child state under a parallel state with `type: "final"` set isn't actually final, and can still transition away to other sibling nodes.

Image

### Expected result

I expected a `final` state to be `final`, and not allow transitions away.

### Actual result

Events can transition away from the `final` state. The inspector even warns that it shouldn't be possible.

Image

### Reproduction

https://codesandbox.io/p/devbox/xstate5-parallel-event-transitions-forked-lk73v7

### Additional context

xstate@5.26

kaigritun:
## Summary

This PR fixes #5460 where parallel states with `type: 'final'` child states incorrectly allowed transitions from those final states.

## The Problem

According to SCXML semantics, final states are terminal and should not allow outgoing transitions. However, in xstate v5, when a child state under a parallel state has `type: 'final'` set, it could still transition away to other sibling nodes.

## The Fix

Added a check in `StateNode.next()` to immediately return `undefined` if the state node is of type `'final'`. This ensures that:

- Final states within parallel regions cannot transition
- Events sent to final states are properly ignored (not processed)
- The behavior is consistent with SCXML specification

## Changes

- **packages/core/src/StateNode.ts**: Added early return in `next()` method for final states
- **packages/core/test/parallel.test.ts**: Added test case reproducing the issue

## Testing

The fix has been tested with:
- All existing `parallel.test.ts` tests passing
- All existing `final.test.ts` tests passing  
- New test case specifically for issue #5460

### DIFF
--- a/packages/core/test/parallel.test.ts
+++ b/packages/core/test/parallel.test.ts
@@ -1345,7 +1345,7 @@ describe('parallel states', () => {
   });
 
   // https://github.com/statelyai/xstate/issues/5460
-  it('should not allow transitions from final states within parallel regions', () => {
+  it('should keep parallel state stable once a region reaches a final state (no matching handlers)', () => {
     const machine = createMachine({
       id: 'test',
       type: 'parallel',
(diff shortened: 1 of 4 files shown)


davidkpiano:
@Andarist Should this be a minor or patch?

kaigritun:
Patch — this is a bugfix for incorrect behavior (transitions from final states in parallel regions shouldn't be allowed per the SCXML spec). No new features or breaking changes.

packages/core/test/parallel.test.ts
```diff
@@ -1343,4 +1343,99 @@ describe('parallel states', () => {
...
+
+    const actor = createActor(machine);
+    actor.start();
+
+    // Move first region to final state
+    actor.send({ type: 'NEXT' });
+    expect(actor.getSnapshot().value).toEqual({
+      first: 'done',
+      second: 'a'
+    });
```
Andarist, line 1387: What's the value of this assertion here? this surely has to be tested elsewhere already - this doesn't tell us *anything* about the behavior described in the test's title

kaigritun:
You're right - that assertion doesn't demonstrate the blocked transition behavior. I'll update the test to:
1. Define a transition on the final state (or ancestor)
2. Attempt to send the matching event
3. Assert the state remains unchanged (proving the transition was blocked)

Will push an update shortly.

Andarist:
> The behavior is consistent with SCXML specification

Are we sure about this?
- [`selectTransitions`](https://www.w3.org/TR/scxml/#SelectingTransitions) procedure mentions: Create an empty set of enabledTransitions. For each atomic state , find a transition whose 'event' attribute matches event and whose condition evaluates to true.
- the [appendix D](https://www.w3.org/TR/scxml/#AlgorithmforSCXMLInterpretation) also mentions a definition of an atomic state as follows: "A state of type <state> with no child states, or a state of type <final>."

kaigritun:
Good point on the spec reference. Looking at it more carefully:

The SCXML spec does define final states as atomic, but also says in [3.7 `<final>`](https://www.w3.org/TR/scxml/#final):
> `<final>` represents a final state of a `<state>` or `<scxml>` element. **A `<final>` state cannot have any outgoing transitions.**

So the key distinction is:
- Final states ARE atomic (can receive events)
- But they cannot have **outgoing transitions**

The current xstate behavior in the issue (#5460) shows a final state transitioning to a sibling, which violates the "no outgoing transitions" rule.

Happy to adjust the PR description/comments if this interpretation is correct, or close if you think the current behavior is actually spec-compliant.

Andarist:
I agree with your interpretation. The spec has some inconsistencies, but [the allowed children of `<final>`](https://www.w3.org/TR/scxml/#N1048D) are pretty conclusive here.

I wonder though if we shouldnt handle this early, for example in `formatTransitions`. We should also add some tests for other invalid combinations - such as `invoke` defined in a `final` state.

kaigritun:
Thanks for confirming! Good points on both:

1. **Earlier handling in `formatTransitions`** - That would catch invalid definitions at creation time vs runtime. I can move the check there.

2. **Additional tests for invalid combinations** - Will add tests for:
   - `invoke` in final states
   - `on` transitions defined directly on final states
   
I'll push an updated implementation. Should I keep this as a single PR or would you prefer the additional validation as a separate PR?

davidkpiano:
> Thanks for confirming! Good points on both:
> 
> 1. **Earlier handling in `formatTransitions`** - That would catch invalid definitions at creation time vs runtime. I can move the check there.
> 2. **Additional tests for invalid combinations** - Will add tests for:
>    
>    * `invoke` in final states
>    * `on` transitions defined directly on final states
> 
> I'll push an updated implementation. Should I keep this as a single PR or would you prefer the additional validation as a separate PR?

Sorry for the late reply, you can keep it in this PR.

PR #5467, opened 2026-02-16
[@xstate/store] Add wildcard support for store.on('*', ...)

davidkpiano:
Add wildcard `'*'` support for `store.on('*', …)` to listen to all emitted events. The handler receives the union of all emitted event types.

```ts
const store = createStore({
  context: { count: 0 },
  emits: {
    increased: (_: { upBy: number }) => {},
    decreased: (_: { downBy: number }) => {}
  },
  on: {
    inc: (ctx, _, enq) => {
      enq.emit.increased({ upBy: 1 });
      return { ...ctx, count: ctx.count + 1 };
    }
  }
});

store.on('*', (ev) => {
  // ev:
  // | { type: 'increased'; upBy: number }
  // | { type: 'decreased'; downBy: number }
});
```

### DIFF
--- a/packages/xstate-store/src/types.ts
+++ b/packages/xstate-store/src/types.ts
@@ -100,7 +100,9 @@ export interface Store<
   on: <TType extends TEmitted['type'] | '*'>(
     type: TType,
     handler: (
-      emitted: TEmitted & (TType extends '*' ? unknown : { type: TType })
+      emitted: Compute<
+        TEmitted & (TType extends '*' ? unknown : { type: TType })
+      >
     ) => void
   ) => Subscription;
   /**
@@ -211,6 +213,8 @@ export type SpecificStoreConfig<
 };
 
 type IsEmptyObject<T> = T extends Record<string, never> ? true : false;
+
+type Compute<A> = A extends any ? { [K in keyof A]: A[K] } : never;
 
 export type AnyStore = Store<any, any, any>;
 


packages/xstate-store/src/types.ts
```diff
@@ -97,10 +97,12 @@ export interface Store<
...
   sessionId: string;
-  on: <TEmittedType extends TEmitted['type']>(
+  on: <TEmittedType extends TEmitted['type'] | '*'>(
     eventType: TEmittedType,
     emittedEventHandler: (
-      emittedEvent: Compute<TEmitted & { type: TEmittedType }>
+      emittedEvent: TEmittedType extends '*'
+        ? TEmitted
+        : Compute<TEmitted & { type: TEmittedType }>
     ) => void
```
Andarist: FWIW this is how we type it in the core:
https://github.com/statelyai/xstate/blob/dbfe25dd9dd1148baf5003b061bd59566595a3e2/packages/core/src/types.ts#L2023-L2028
https://github.com/statelyai/xstate/blob/3a370a951a1aedff35aff1d101c15e38d5adb4f7/packages/core/src/createActor.ts#L467-L487

packages/xstate-store/src/types.ts
```diff
@@ -185,7 +187,9 @@ export type StoreConfig<
 > = {
   context: TContext;
   emits?: {
-    [K in keyof TEmitted]: (payload: TEmitted[K]) => void;
+    [K in keyof TEmitted]: K extends '*'
+      ? never
+      : (payload: TEmitted[K]) => void;
   };
```
Andarist, line 191: It seems that in the core we don't try to do this at type-level. I think it's fine to skip this and just do a dev-only runtime validation

packages/xstate-store/src/store.ts
```diff
@@ -57,6 +57,10 @@ function createStoreCore<
     if (typeListeners) {
       typeListeners.forEach((listener) => listener(ev));
     }
+    const wildcardListeners = listeners.get('*' as TEmitted['type']);
+    if (wildcardListeners) {
+      wildcardListeners.forEach((listener) => listener(ev));
+    }
```
Andarist, line 63: for reference, this is how we handle this in the core:
https://github.com/statelyai/xstate/blob/3a370a951a1aedff35aff1d101c15e38d5adb4f7/packages/core/src/createActor.ts#L193-L208

We probably don't need to handle errors anyhow in here for now, let them bubble up. I'm glad the order of listeners matches the one in the core. I'd only codify this in tests (wildcard listener called after a specific listener).

Andarist (approved):

PR #5490, opened 2026-03-27
[@xstate/store] `createStoreLogic` and selectors

davidkpiano:
Added `createStoreLogic` for reusable store definitions with input and `selectors` support for both `createStore` and `createStoreLogic`.

```ts
const counterLogic = createStoreLogic({
  context: (input: { initialCount: number }) => ({
    count: input.initialCount
  }),
  on: {
    inc: (ctx) => ({ count: ctx.count + 1 })
  },
  selectors: {
    doubled: (ctx) => ctx.count * 2
  }
});

const counter1 = counterLogic.createStore({ initialCount: 42 });
const counter2 = counterLogic.createStore({ initialCount: 0 });

counter1.selectors.doubled.get(); // 84
```

Selectors are reactive `ReadonlyAtom`s (powered by `store.select()`), composable with `createAtom`, and preserved through `.with()` extensions.

### DIFF
--- a/packages/xstate-store/src/store.ts
+++ b/packages/xstate-store/src/store.ts
@@ -57,7 +57,7 @@ function attachSelectors<
     );
   }
 
-  const originalWith = store.with.bind(store);
+  const originalWith = store.with;
   const storeWithSelectors = store as StoreWithSelectors<
     TContext,
     TEventPayloadMap,
@@ -315,7 +315,7 @@ export function createStore<
   TSelectors extends Record<string, (context: TContext) => any>
 >(
   definition: StoreConfig<TContext, TEventPayloadMap, TEmittedPayloadMap> & {
-    selectors: TSelectors & Record<string, (context: TContext) => any>;
+    selectors: TSelectors;
   }
 ): StoreWithSelectors<
   TContext,
@@ -594,14 +594,15 @@ export function createStoreTransition<
  * const store = todoLogic.createStore();
  * ```
  */
+// Overload: with input + selectors
 export function createStoreLogic<
   TContext extends StoreContext,
   const TEventPayloadMap extends EventPayloadMap,
   TEmittedPayloadMap extends EventPayloadMap,
-  TInput = void,
-  TSelectors extends Record<string, (context: TContext) => any> = {}
+  TInput,
+  TSelectors extends Record<string, (context: TContext) => any>
 >(config: {
-  context: [TInput] extends [void] ? TContext : (input: TInput) => TContext;
+  context: (input: TInput) => TContext;
   on: {
     [K in keyof TEventPayloadMap & string]: StoreAssigner<
       TContext,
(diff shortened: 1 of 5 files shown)


packages/xstate-store/src/store.ts
```diff
@@ -33,6 +37,42 @@ const inspectionObservers = new WeakMap<
...
+  selectorsConfig: TSelectors
+): StoreWithSelectors<TContext, TEventPayloadMap, TEmitted, TSelectors> {
+  const selectors = {} as ResolvedStoreSelectors<TContext, TSelectors>;
+  for (const key of Object.keys(selectorsConfig) as (keyof TSelectors)[]) {
+    (selectors as any)[key] = store.select(
+      selectorsConfig[key] as Selector<TContext, any>
+    );
+  }
+
+  const originalWith = store.with.bind(store);
```
Andarist: does `with` require access to the store under `this`? it feels that this might be bound with a wrong value here - if a user relies on `this`, I think they will expect the final value to be available there (and that's `storeWithSelectors` and not `store`)

packages/xstate-store/src/store.ts
```diff
@@ -33,6 +37,42 @@ const inspectionObservers = new WeakMap<
...
+    TContext,
+    TEventPayloadMap,
+    TEmitted,
+    TSelectors
+  >;
+  storeWithSelectors.selectors = selectors;
+  storeWithSelectors.with = ((extension: any) => {
+    const extended = originalWith(extension);
+    return attachSelectors(extended, selectorsConfig);
+  }) as any;
```
Andarist, line 71: q: will this work properly with multi-layer `.with` calls? like `x.with().with().with()`. It feels like it will lead to calling `attachSelectors` for each of those layers and I wonder if that's a) optimal b) correct

packages/xstate-store/src/store.ts
```diff
@@ -239,13 +279,50 @@ export type TransitionsFromEventPayloadMap<
...
+ *   functions that derive values from context. Each selector becomes a reactive
+ *   `ReadonlyAtom` on `store.selectors`.
  * @returns A store instance with methods to send events and subscribe to state
  *   changes
  */
+export function createStore<
+  TContext extends StoreContext,
+  const TEventPayloadMap extends EventPayloadMap,
+  TEmittedPayloadMap extends EventPayloadMap,
+  TSelectors extends Record<string, (context: TContext) => any>
```
Andarist:
```suggestion
  TSelectors extends Record<string, (context: TContext) => unknown>
```

packages/xstate-store/src/store.ts
```diff
@@ -239,13 +279,50 @@ export type TransitionsFromEventPayloadMap<
...
+>(
+  definition: StoreConfig<TContext, TEventPayloadMap, TEmittedPayloadMap> & {
+    selectors: TSelectors & Record<string, (context: TContext) => any>;
+  }
```
Andarist, line 319: cant this be simplified to just this?
```suggestion
  definition: StoreConfig<TContext, TEventPayloadMap, TEmittedPayloadMap> & {
    selectors: TSelectors;
  }
```

packages/xstate-store/src/store.ts
Andarist: on that point, why `emits` and `selectors` are threaded through things differently? `emits` go through `createStoreCore` and `selectors` go through `attachSelectors`

packages/xstate-store/src/store.ts
```diff
@@ -269,8 +346,12 @@ export function createStore<
   TEmitted
 >;
 export function createStore(
-  definitionOrLogic: StoreConfig<any, any, any> | AnyStoreLogic
-) {
+  definitionOrLogic:
+    | (StoreConfig<any, any, any> & {
+        selectors?: Record<string, (context: any) => any>;
+      })
```
Andarist, line 352: q: related to some of the other questions - why `selectors` are not part of the `StoreConfig`?

packages/xstate-store/src/store.ts
```diff
@@ -460,6 +547,120 @@ export function createStoreTransition<
...
+ * ```
+ */
+export function createStoreLogic<
+  TContext extends StoreContext,
+  const TEventPayloadMap extends EventPayloadMap,
+  TEmittedPayloadMap extends EventPayloadMap,
+  TInput = void,
+  TSelectors extends Record<string, (context: TContext) => any> = {}
+>(config: {
+  context: [TInput] extends [void] ? TContext : (input: TInput) => TContext;
```
Andarist: couldnt this just be `TContext | ((input: TInput) => TContext)`? 🤔

Andarist (commented):
note that all of my comments are just questions, not suggestions

packages/xstate-store/src/store.ts
davidkpiano: I'll try this; makes sense (blame Claude, reviewed too fast)

packages/xstate-store/src/store.ts
davidkpiano: Done

packages/xstate-store/src/store.ts
davidkpiano: Needed to use overloads for this so that `context` is properly inferred

packages/xstate-store/src/store.ts
davidkpiano: > q: will this work properly with multi-layer `.with` calls? like `x.with().with().with()`. It feels like it will lead to calling `attachSelectors` for each of those layers and I wonder if that's a) optimal b) correct

It's correct, and yeah the intermediate selector atoms get GC'd which is slightly wasteful but correct & simple. Also selectors are after-the-fact (they observe the store; they're not *part* of the store), and we shouldn't add them to `storeLogic`

packages/xstate-store/src/store.ts
davidkpiano: Keeps it additive so we don't pollute the `StoreConfig` interface- `createStoreConfig` and `fromStore` don't need it at the moment. We can move it in later if we decide to make selectors a broader XState concept maybe?

packages/xstate-store/src/store.ts
davidkpiano: Fixed

sasivarnan:
Thank you for implementing this feature. I am a primary user of the ability to create multiple store instances from a store definition.  

Could you confirm if we are planning to use the value returned by `createStoreLogic` to initialize a local store that is created with `useStore`? This way we can define the local store in an external file.

davidkpiano:
> Could you confirm if we are planning to use the value returned by `createStoreLogic` to initialize a local store that is created with `useStore`? This way we can define the local store in an external file.

Good question; not yet but it should.

packages/xstate-store/src/types.ts
```diff
@@ -491,6 +491,56 @@ export type StoreExtension<
...
+  TEmitted extends EventObject,
+  TInput,
+  TSelectors extends Record<string, (context: TContext) => unknown>
+> {
+  /** Creates a new store instance from this logic definition. */
+  createStore: [TInput] extends [void]
+    ? () => StoreWithSelectors<TContext, TEventPayloadMap, TEmitted, TSelectors>
+    : (
+        input: TInput
+      ) => StoreWithSelectors<TContext, TEventPayloadMap, TEmitted, TSelectors>;
```
Andarist, line 541: we'd usually only allow a dynamic *context* to be created based on the provided `input`. Creating the whole store feels like a divergence from the other APIs

davidkpiano:
Bringing this into v4 instead

PR #5503, opened 2026-04-12
Fix initialTransition for invoked system IDs

### LINKED ISSUE #5454: Bug: `initialTransition` fails when invoke has `systemId`
### XState version

XState version 5

### Description

`initialTransition` is documented as a pure function that computes state transitions without side effects. However, when called on a machine that has an invoke with a `systemId`, it throws `"Actor with system ID 'X' already exists"` on the **first call**.

The same machine with invoke but **without** `systemId` works correctly.

### Expected result

Per the [pure transitions documentation](https://stately.ai/docs/pure-transitions):

> Pure transition functions allow you to compute the next state and actions of a state machine **without creating a live actor or executing any side effects**.

`initialTransition(machine)` should return the initial snapshot without any side effects, regardless of whether invokes have `systemId` or not.

### Actual result

Adding `systemId` to an invoke causes `initialTransition` to fail immediately.

## Root Cause Analysis

Local stack trace shows:

```
at _set (raise-ad4f12ad.cjs.js:413:19)
at new Actor (raise-ad4f12ad.cjs.js:635:19)
at createActor (raise-ad4f12ad.cjs.js:1130:10)
at resolveSpawn (raise-ad4f12ad.cjs.js:1219:16)
at resolveAndExecuteActionsWithContext (raise-ad4f12ad.cjs.js:2323:56)
at resolveActionsAndContext (raise-ad4f12ad.cjs.js:2346:21)
at enterStates (raise-ad4f12ad.cjs.js:2140:20)
at microstep (raise-ad4f12ad.cjs.js:2082:15)
at getInitialSnapshot (StateMachine-f73271f9.cjs.js:444:48)
at initialTransition (xstate.cjs.js:337:30)
```

It appears that during `initialTransition`:
1. The machine enters states and invokes actors
2. When an invoke has `systemId`, the actor is registered in a system
3. Something causes the state to be entered twice (or the actor to be registered twice)
4. The second registration fails because the `systemId` already exists

### Reproduction

https://stackblitz.com/edit/github-pnpxlm9l?file=src%2Fmain.ts

### Additional context

_No response_

ShipItAndPray:
Fixes #5454.\n\n was creating an inert actor scope by instantiating the target logic through , which meant invoked children with a  were effectively evaluated twice during the initial snapshot path. That could register the same system ID more than once and throw.\n\nThis changes the inert scope to use  instead, keeping  pure while still providing the scope properties needed by transition logic. A regression test covers the  case.

### DIFF
--- a/.changeset/clever-llamas-lay.md
+++ b/.changeset/clever-llamas-lay.md
@@ -1,0 +1,5 @@
+---
+"xstate": patch
+---
+
+Fix `initialTransition` for machines that invoke actors with a `systemId`.
--- a/packages/core/src/getNextSnapshot.ts
+++ b/packages/core/src/getNextSnapshot.ts
@@ -1,4 +1,4 @@
-import { createActor } from './createActor.ts';
+import { createEmptyActor } from './actors/index.ts';
 import {
   ActorScope,
   AnyActorLogic,
@@ -13,7 +13,7 @@ import {
 export function createInertActorScope<T extends AnyActorLogic>(
   actorLogic: T
 ): AnyActorScope {
-  const self = createActor(actorLogic as AnyActorLogic);
+  const self = createEmptyActor();
   const inertActorScope: ActorScope<
     SnapshotFrom<T>,
     EventFromLogic<T>,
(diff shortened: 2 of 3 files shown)


ShipItAndPray:
Added a patch changeset in d267be8 to satisfy the release note requirement.

PR #5516, opened 2026-05-05
fix(core): resolve children snapshot union pollution for typed invoke ids

### LINKED ISSUE #5181: Bug: React Context Typescript not resolving children actors when selected.
### XState version

XState version 5

### Description

I'm new to XState and still learning to use it effectively. I'm unsure if I ran into a bug or if I'm using the library unintentionally. 

## Problem

I've been struggling for countless hours on this issue. Hopefully, someone has a solution. 

My goal is to select context values and send events to a child machine actor of a root machine.

The core of my issue - Typescript doesn't narrow to the selected child actor. This happens once multiple actors are set up on the machine. When I console log the selected actor - It does have the correct value I'm looking for. I've included examples below.

Image

The Error

```
Property 'context' does not exist on type 'PromiseSnapshot<AuthUser, NonReducibleUnknown> | MachineSnapshot<{ count: number; }, { type: "inc"; } | { type: "dec"; }, {}, {}, string, NonReducibleUnknown, MetaObject, { ...; }>'.
  Property 'context' does not exist on type '{ status: "active"; output: undefined; error: undefined; } & { input: NonReducibleUnknown; }'.ts(2339)
```

Console logging the values outputs the expected actor
Image

I was able to reproduce the issue on code sandbox. 

Navigate to `App.tsx` to see the comments. 
https://codesandbox.io/p/devbox/busy-mopsa-n2crkj?workspaceId=ws_NiJ4KeBNbFceadwTQwz1Xf

The child machine is invoked in the initial state of the machine. 

Is this a bug with XState or am I doing something wrong? 

### Expected result

Typescript narrowing should find the targeted actor.

### Actual result

Typescript fails to narrow to the selected child actor. 

### Reproduction

https://codesandbox.io/p/devbox/busy-mopsa-n2crkj?workspaceId=ws_NiJ4KeBNbFceadwTQwz1Xf

### Additional context

_No response_

### LINKED ISSUE #5515: Children map union pollution from internal actors when using setup() with invoke arrays
## XState version

5.31.0

## Description

When a root machine uses `setup()` with multiple invoked child actors (each having their own internal `fromPromise`/`fromObservable` actors), accessing a specific child via `snapshot.children.specificId` returns a union of all actor contexts instead of the specific child's context. This makes it impossible to access child actor context without runtime type narrowing.

Related to #5181 — same root cause, different angle.

## Root Cause Analysis

We traced this through the type system:

### 1. `ToProvidedActor` doesn't infer literal ids from `invoke` config

In `setup.ts`, `ToProvidedActor` derives actor `id` from `TChildrenMap`:

```ts
id: IsNever<TChildrenMap> extends true
  ? string | undefined
  : K extends keyof Invert<TChildrenMap>
    ? Invert<TChildrenMap>[K] & string   // literal id
    : string | undefined;                 // fallback
```

When using `setup()` without explicit `types.children`, `TChildrenMap` defaults to `{}`, so **all** actors get `id: string | undefined` — even those with literal `id`s in the `invoke` config. The literal ids from `invoke: [{ id: "auth", src: "auth" }]` are not propagated into `TChildrenMap`.

### 2. `ToChildren` index signature includes internal actors

In `types.ts`, `ToChildren` decides whether to include an index signature:

```ts
[undefined extends TActor['id']
  ? 'include'    // ← triggered because internal actors have id: undefined
  : string extends TActor['id']
    ? 'include'
    : 'exclude']
```

Internal actors (`fromPromise`, `fromObservable`) from child machines have `id: undefined`, which triggers the index signature. The index signature distributes over ALL non-concrete actors:

```ts
[id: string]: TActor extends any ? ActorRefFromLogic<TActor['logic']> | undefined : never;
```

### 3. TypeScript intersection pollutes concrete keys

The final type is `Compute<ToConcreteChildren<TActor> & { [id: string]: UnionOfAllActorRefs }>`. When accessing `children.auth`, TypeScript intersects the concrete mapping with the index signature, producing a union.

## Reproduction

```ts
import { fromPromise, setup, createActor } from 'xstate';

const authLogic = setup({
  actors: {
    fetchUser: fromPromise(async () => ({ token: 'tok' }))
  }
}).createMachine({
  context: { token: null as string | null },
  initial: 'idle',
  states: { idle: {} }
});

const telemetryLogic = setup({
  actors: {
    checkConsent: fromPromise(async () => ({ status: 'granted' }))
  }
}).createMachine({
  context: { store: null as string | null },
  initial: 'idle',
  states: { idle: {} }
});

const root = setup({
  actors: { auth: authLogic, telemetry: telemetryLogic }
}).createMachine({
  context: { value: 0 },
  invoke: [
    { id: 'auth', systemId: 'auth', src: 'auth' },
    { id: 'telemetry', systemId: 'telemetry', src: 'telemetry' }
  ]
});

const snapshot = createActor(root).getSnapshot();

// ❌ Error: Property 'token' does not exist on type
//    '{ token: string | null } | { store: string | null }'
snapshot.children.auth!.getSnapshot().context.token;

// ❌ Error: Property 'store' does not exist on type
//    '{ token: string | null } | { store: string | null }'
snapshot.children.telemetry!.getSnapshot().context.store;
```

### Workaround

Manually declare `types.children` (from #5181 comment):

```ts
const root = setup({
  types: {
    children: {} as { auth: 'auth'; telemetry: 'telemetry' }
  },
  actors: { auth: authLogic, telemetry: telemetryLogic }
}).createMachine({ /* ... */ });
```

This populates `TChildrenMap` and gives `ToProvidedActor` the literal ids. But this shouldn't be necessary — the ids are already specified in the `invoke` config.

## Partial Fix

We have a partial fix for `ToChildren` that prevents actors with literal ids from appearing in the index signature:

```ts
type DynamicIdActors<TActor extends ProvidedActor> = TActor extends any
  ? TActor['id'] extends string
    ? string extends TActor['id']
      ? TActor
      : never
    : never
  : never;
```

This correctly fixes cases where `types.actors` or `types.children` are explicitly declared (existing tests improved — the index signature now only contains actors without literal ids). But it doesn't fix the `setup()` + `invoke` case because `TChildrenMap` inference is the upstream problem.

## Full Fix

The complete solution requires two changes:

1. **`ToChildren` (partial, ready):** Use `DynamicIdActors` to exclude literal-id and undefined-id actors from the index signature
2. **`setup()` / `ToProvidedActor` (needed):** Infer `TChildrenMap` from `invoke` configs in `createMachine()`, so literal ids like `{ id: "auth", src: "auth" }` propagate into the actor type system without requiring manual `types.children`

## Expected Behaviour

`snapshot.children.auth` should return `ActorRefFromLogic<typeof authLogic> | undefined` — the specific actor type, not a union of all child actor types.

joshuaellis:
### Summary

When using `setup()` with multiple invoked child actors, accessing `snapshot.children.specificId` returns a union of all actor contexts instead of the specific child's context. This makes typed child actor access impossible without runtime narrowing.

**Root cause:** Two issues compound:

1. `ToChildren`'s index signature includes actors that already have literal ids in `ToConcreteChildren`, polluting specific keys via TypeScript's intersection behaviour
2. `setup().createMachine()` doesn't infer invoke `id` → `src` mappings from the config, so all actors default to `id: string | undefined`

**Fix:**

- `NonConcreteActors` (types.ts) — excludes actors with literal ids from the index signature
- `ExtractConfigChildren` (setup.ts) — infers `{ [invokeId]: src }` from the `const` machine config, removing the need for manual `types.children`

### Related issues

* closes #5515 
* closes #5181

### DIFF
--- a/.changeset/upset-pants-wear.md
+++ b/.changeset/upset-pants-wear.md
@@ -1,0 +1,5 @@
+---
+'xstate': patch
+---
+
+fix(core): resolve children snapshot union pollution for typed invoke
--- a/packages/core/src/setup.ts
+++ b/packages/core/src/setup.ts
@@ -79,6 +79,37 @@ type ToStateSchema<TSchema extends StateSchema> = {
 type RequiredSetupKeys<TChildrenMap> =
   IsNever<keyof TChildrenMap> extends true ? never : 'actors';
 
+// Extract { [invokeId]: src } mappings from a machine config's invoke entries.
+// This allows createMachine to infer TChildrenMap from the invoke config
+// when types.children is not explicitly provided.
+type ExtractInvokeEntry<T> = T extends {
+  id: infer TId extends string;
+  src: infer TSrc extends string;
+}
+  ? { [K in TId]: TSrc }
+  : {};
+
+type ExtractInvokeChildren<T> = T extends readonly (infer E)[]
+  ? UnionToIntersection<ExtractInvokeEntry<E>>
+  : ExtractInvokeEntry<T>;
+
+type UnionToIntersection<U> = (U extends any ? (x: U) => void : never) extends (
+  x: infer I
+) => void
+  ? I
+  : never;
+
+type ExtractConfigChildren<TConfig> = TConfig extends {
+  invoke: infer TInvoke;
+}
+  ? ExtractInvokeChildren<TInvoke>
+  : {};
+
+type MergeChildrenMap<
+  TExplicit extends Record<string, string>,
+  TInferred extends Record<string, string>
+> = IsNever<keyof TExplicit> extends true ? TInferred : TExplicit;
+
 export type SetupReturn<
   TContext extends MachineContext,
   TEvent extends AnyEventObject,
@@ -247,6 +278,10 @@ export type SetupReturn<
       TOutput,
       TEmitted,
       TMeta
+    >,
+    TResolvedChildren extends Record<string, string> = MergeChildrenMap<
+      TChildrenMap,
+      Cast<ExtractConfigChildren<TConfig>, Record<string, string>>
     >
   >(
     config: TConfig
@@ -260,10 +295,10 @@ export type SetupReturn<
             to: RoutableStateId<TConfig>;
           }),
     Cast<
-      ToChildren<ToProvidedActor<TChildrenMap, TActors>>,
+      ToChildren<ToProvidedActor<TResolvedChildren, TActors>>,
       Record<string, AnyActorRef | undefined>
     >,
-    ToProvidedActor<TChildrenMap, TActors>,
+    ToProvidedActor<TResolvedChildren, TActors>,
     ToParameterizedObject<TActions>,
     ToParameterizedObject<TGuards>,
     TDelay,
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -2546,6 +2546,15 @@ type ToConcreteChildren<TActor extends ProvidedActor> = {
   >;
 };
 
+// Actors that don't have literal string IDs — these are the only ones
+// that should appear in the index signature fallback, since actors with
+// literal IDs are already covered by ToConcreteChildren.
+type NonConcreteActors<TActor extends ProvidedActor> = TActor extends any
+  ? ExtractLiteralString<TActor['id']> extends never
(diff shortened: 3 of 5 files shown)


joshuaellis:
@Andarist || @davidkpiano polite nudge on this :)

PR #5517, opened 2026-05-06
Tighten up publishing security

Andarist:


.github/workflows/publish.yml
```diff
@@ -8,33 +8,47 @@ on:
...
+        with:
+          version: pnpm run version
 
+  publish:
+    name: Publish
+    if: needs.version.outputs.hasChangesets == 'false'
+    needs: version
     runs-on: ubuntu-latest
-
+    environment: npm
```
Andarist, line 41: This line is the biggest change here. And it's going to get slightly more annoying to publish packages with it. 

With this env-based "gate" after merging a Version Packages PR, you will be prompted (through notifications/email) to approve this publishing job. At least when you set up the required reviewers for this env (which you should do).

Why this should be done? To decouple repo write access from publish rights. Without this anyone with write access to the repo can potentially land a malicious Version Packages PR.

PR #5528, opened 2026-05-21
fix: avoid spurious imperative-action warning in enqueueActions

### LINKED ISSUE #5343: Bug: Using actor.send outputs a warning that assign is not imperative when sent to an actor using enqueueActions
### XState version

XState version 5

### Description

Coming from: https://github.com/statelyai/docs/pull/469
I initially thought this was expected, but looks like it's not... so I'm opening an issue here.

I left a test in you can run:
Image

This [line](https://github.com/MarjanDB/xstate-minimal-reproduce/blob/7ef33192afa2d02f0fcbbbf0ce2abaa4eecc6187/example/ParentStateMachine.ts#L62) is using self.send, but using `enqueue.sendTo(self, …)` will also produce the same warning.

### Expected result

There should be no warning.

### Actual result

I'm getting a warning that custom actions should not call `assign`

### Reproduction

https://github.com/MarjanDB/xstate-minimal-reproduce

### Additional context

_No response_

ATOM00blue:
## Summary
Fixes a spurious "Custom actions should not call `…` directly, as it is not imperative" warning logged when a custom action synchronously sends an event to another actor whose transition resolves an `enqueueActions` block using builtin creators (`enqueue.assign(…)`, `enqueue.sendTo(self, …)`, etc.).

Root cause: the module-global `executingCustomAction` flag, set during a custom action's `exec()`, was still `true` while the *other* actor's `enqueueActions` was being resolved, so its builtin action creators wrongly tripped the dev warning. Collecting enqueued actions is part of resolving `enqueueActions`, not executing a custom action, so the flag is now cleared (and restored) around `collect()`.

Added two regression tests and a changeset.

## Test plan
- `pnpm test:core` (1739 passing)
- New tests fail without the fix, pass with it

Closes #5343

### DIFF
--- a/packages/core/src/actions/enqueueActions.ts
+++ b/packages/core/src/actions/enqueueActions.ts
@@ -1,4 +1,8 @@
 import isDevelopment from '#is-development';
+import {
+  executingCustomAction,
+  setExecutingCustomAction
+} from '../createActor.ts';
 import { Guard, evaluateGuard } from '../guards.ts';
 import {
   Action,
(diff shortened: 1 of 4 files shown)


packages/core/test/actions.test.ts
```diff
@@ -4342,6 +4342,114 @@ describe('actions', () => {
...
+            }
+          }
+        }
+      }
+    });
+
+    const actor = createActor(parentMachine).start();
+    actor.send({ type: 'CALL_CHILD' });
+
+    expect(warnSpy).not.toHaveBeenCalled();
```
Andarist, line 4450: This one looks like an invalid assertion to me - this test *should* warn unless my reasoning is off. This would indicate that the fix is misapplied.

packages/core/src/actions/enqueueActions.ts
```diff
@@ -167,18 +171,32 @@ function resolveEnqueueActions(
...
+  // `enqueue.assign(…)`, `enqueue.sendTo(…)` and friends call the action
+  // creators (`assign()`, `sendTo()`, …) under the hood. Those creators warn
+  // when they are invoked while a custom action is executing. Collecting
+  // actions here is a part of resolving the `enqueueActions` action and not a
+  // part of executing a custom action, so we have to clear the flag to avoid
+  // spurious warnings - this can be observed when an executing custom action
+  // synchronously sends an event to another actor that resolves
+  // `enqueueActions` while processing that event.
+  const wasExecutingCustomAction = executingCustomAction;
+  setExecutingCustomAction(false);
```
Andarist, line 183: as per: https://github.com/statelyai/xstate/pull/5528/changes#r3280659742

this fix shouldn't happen here - but rather all of those `actions.push` calls in `enqueue.X = ` should be wrapped instead

PR #5537, opened 2026-06-06
[@xstate/store] Add store effect trigger

davidkpiano:
Effects enqueued via `enq.effect(...)` now receive an enqueue object with `trigger`, `send`, and `getSnapshot`, so you can dispatch events back into the store after async work and read the latest state without needing a reference to the store. This is especially useful with `createStoreLogic(...)`, where the store is created per-instance and there's no store to close over.

```ts
const storeLogic = createStoreLogic({
  context: () => ({ foo: null, loading: false }),
  on: {
    fetchFoo: (context, event, enq) => {
      enq.effect(({ trigger }) => {
        myApi.requestFoo().then((response) => trigger.gotFoo({ response }));
      });
      return { ...context, loading: true };
    },
    gotFoo: (context, event) => ({
      ...context,
      foo: event.response,
      loading: false
    })
  }
});
```

### DIFF
--- a/packages/xstate-store/src/store.ts
+++ b/packages/xstate-store/src/store.ts
@@ -283,10 +283,11 @@ function createStoreCore<
     atom.set(nextSnapshot);
     notifyInspection(event, nextSnapshot);
 
+    let committed = false;
     const effectEnqueue = {
       trigger,
       send,
-      getSnapshot: () => currentSnapshot
+      getSnapshot: () => (committed ? currentSnapshot : nextSnapshot)
     } as StoreEffectEnqueue<any, any>;
 
     for (const effect of effects) {
@@ -296,6 +297,8 @@ function createStoreCore<
         emit(effect);
       }
     }
+
+    committed = true;
   }
 
   const trigger =
--- a/packages/xstate-store/src/store.ts
+++ b/packages/xstate-store/src/store.ts
@@ -15,6 +15,7 @@ import {
   StoreContext,
   StoreConfig,
   StoreEffect,
+  StoreEffectEnqueue,
   StoreInspectionEvent,
   StoreProducerAssigner,
   StoreSnapshot,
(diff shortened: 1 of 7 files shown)


packages/xstate-store/src/store.ts
```diff
@@ -282,9 +283,15 @@ function createStoreCore<
     atom.set(nextSnapshot);
     notifyInspection(event, nextSnapshot);
 
+    const effectEnqueue = {
+      trigger,
+      send,
+      getSnapshot: () => currentSnapshot
```
Andarist: why this one doesn't go through a dance like this?
```ts
        getSnapshot: () =>
          committed ? actorScope.self.getSnapshot() : nextSnapshot
```

packages/xstate-store/src/types.ts
```diff
@@ -118,18 +118,45 @@ export type CanObject<TEventPayloadMap extends EventPayloadMap> = {
...
 
+/**
+ * The object passed to an effect function enqueued via `enq.effect(...)`. It
+ * allows effects to send events back into the store (e.g. after an async
+ * operation resolves) and read the latest store state, which is especially
+ * useful with `createStoreLogic(...)` where there is no store reference to
+ * close over.
+ *
+ * Use `trigger` for fully-typed dispatch (e.g. `enq.trigger.gotFoo({ ... })`);
+ * `send` is a loosely-typed escape hatch for dynamically-constructed events.
```
Andarist: I don't get the motivation for this:
>  * `send` is a loosely-typed escape hatch for dynamically-constructed events.

The `Store`'s send method isn't loose: https://github.com/statelyai/xstate/blob/9d9b9f1439b773979c5120a793215f5aa4568d8f/packages/xstate-store/src/types.ts#L198

packages/xstate-store/src/store.ts
davidkpiano: Fixed

PR #5561, opened 2026-06-26
feat(core): type-safe setup invoke transition targets (v6)

SandroMaglione:
## Summary
Makes `target` type-safe inside `onDone`, `onError`, `onSnapshot` and `onTimeout` in `invoke`.

This previously type-checked, even if the `success` state requires a `context`. Now it's reported as an error:
```ts
const s = setup({
  schemas: {
    context: types<{}>(),
    events: {
      GO: types<{}>()
    }
  },
  states: {
    idle: {},
    loading: {},
    success: {
      schemas: {
        context: z.object({ message: z.string() }) // 👈
      }
    }
  },
  actors: {
    load: createAsyncLogic({
      run: async () => undefined
    })
  }
});

s.createMachine({
  context: {},
  initial: 'idle',
  states: {
    idle: {
      on: {
        GO: { target: 'loading' }
      }
    },
    loading: {
      // @ts-expect-error - success context requires a message
      invoke: {
        src: 'load',
        onDone: () => ({ target: 'success' })
      }
    },
    success: {}
  }
});
```


packages/core/src/setup.ts
```diff
@@ -850,6 +907,121 @@ type StateTransitions<
...
+            TActorMap,
+            TGuardMap,
+            TDelayMap,
+            TSystemRegistry
+          >;
+          onError?: StateTransitionConfigOrTarget<
+            TStateSchemas,
+            TContext,
+            TContextShape,
+            ErrorEvent,
```
davidkpiano: Is this a global? Might not be what we want

packages/core/src/types.v6.ts
```diff
@@ -574,7 +575,7 @@ interface Next_InvokeConfigBase<
   >;
   onError?: Next_TransitionConfigOrTarget<
     TContext,
-    ErrorEvent,
+    ErrorActorEvent,
```
SandroMaglione, line 578: @davidkpiano fixed global `ErrorEvent` here as well (not sure if this was correct before)

packages/core/src/setup.ts
```diff
@@ -850,6 +909,117 @@ type StateTransitions<
   >;
 };
 
+type InvokeDoneEvent<TInvoke> = TInvoke extends {
+  onDone?: Next_TransitionConfigOrTarget<
+    any,
```
davidkpiano: Maybe nit, but I like doing eg. `infer _TContext` here because in the past, I've found that `any` and `never` don't play too nice and cause weird incompatible types

SandroMaglione: Done

davidkpiano:
Can you add a changeset? `pnpm changeset` (or ask agent)

SandroMaglione:
> Can you add a changeset? `pnpm changeset` (or ask agent)

Done

davidkpiano (approved):

PR #5567, opened 2026-06-27
feat(core): Infer async logic output with input-only schema (v6)

SandroMaglione:
## Summary
Inside `createAsyncLogic` when only `input` was defined inside `schemas`, the `output` was not inferred anymore from the return type of `run`, but instead became `unknown` when `output` was missing from `schemas`.

This PR allows `input` only to be defined, while `output` keeps being inferred from `run` return type:
```ts
// Output comes from `run`, because there is no output schema to use
const logic = createAsyncLogic({
  schemas: {
    input: types<{ name: string }>()
  },
  run: async ({ input }) => {
    return { message: input.name };
  }
});
```


davidkpiano:
Can you resolve conflicts? (sorry, lots of things in motion!)

PR #5581, opened 2026-06-30
fix(core): start listener/subscription effects before their target actor

asfktz:
## Problem

When a child actor emits events synchronously during startup, `enq.listen()` misses them — even though the equivalent `actor.on()` pattern works fine:

```ts
const childLogic = createCallbackLogic(({ emit }) => {
  // emits immediately on startup
  emit({ type: 'someEvent' });
});

entry: (_, enq) => {
  const child = enq.spawn(childLogic);

  // ✅ works — actor.on() is attached immediately
  child.on('someEvent', (ev) => { /* triggered */ });

  // ❌ doesn't work — listener is queued and only attached after the actor starts
  enq.listen(child, 'someEvent', (ev) => ({ type: 'HEARD' }));
};
```

This may feel counterintuitive to newcomers unfamiliar with the internals of effect queueing. 
In practice, there is also no way to catch startup events using `enq.listen()`.

Some cases where it matters:

- **Resize observer** — an actor wrapping a `ResizeObserver` that immediately emits the element's current dimensions, then continues emitting on resize
- **Auth gate** — an actor that checks a token in memory/localStorage and immediately emits `authenticated` if a valid token exists
- **Media query** — an actor wrapping `matchMedia` that emits the current `{ matches: boolean }` synchronously on start, then on each change (e.g. dark mode, breakpoint detection)
- **Network status** — an actor wrapping `navigator.onLine` that emits `online` or `offline` immediately, then on `window` events
- **Intersection observer** — an actor wrapping `IntersectionObserver` that fires its callback synchronously on attach with the element's current visibility state
- **Scroll / focus state** — an actor that reads `window.scrollY` or `document.activeElement` on start and emits the initial value before watching for changes

## Solution

Reorder effect execution so that `enq.listen()` and `enq.subscribeTo()` effects are started **before** their target actor.

### Approaches considered

I considered two approaches:

1. **Reorder at definition time** — return effects in execution order from `transition()` / `initialTransition()`
2. **Reorder at execution time only** — keep effects in definition order, reorder just before executing them internally

I initially went with (2), but users who manually execute effects (e.g. in a serverless workflow) would hit the same bug:

```ts
const [state, effects] = initialTransition(machine);

for (const effect of effects) {
  execute(effect); // would fail if effects are in definition order
}
```

So I went with **(1)** — effects returned by `transition()` / `initialTransition()` are already in the correct execution order. The reordering logic lives in a shared `reorderEffects()` helper that moves each actor's `listen`/`subscribeTo` start effect ahead of its own start effect (without crossing any stop effect for that actor). Both `transition()` and `initialTransition()` collect effects through the same private `collectEffects()` method, so the public API and internal runtime stay in sync.

### Behavior change: `subscribeTo` now sees the initial snapshot

One side effect worth noting: because the subscription is now attached before the target actor starts, a `subscribeTo` **snapshot** handler will also receive the actor's **initial** snapshot, not just later changes. An actor emits its first snapshot as it starts, and since the subscription is already in place at that point, that first emission gets delivered.

This turns out to be what the motivating cases need (a resize observer reporting current dimensions, an auth gate reporting initial status), and it makes `subscribeTo` behave more consistently when observing an actor from birth. The `done` / `error` handlers are unaffected beyond the startup fix itself.

## Test plan

**Startup behavior**
- [x] Listener receives emitted events from a **spawned** actor during startup
- [x] Listener receives emitted events from an **invoked** actor during startup
- [x] Listener receives emitted events from an actor **passed through input**
- [x] `subscribeTo` receives done events from a spawned actor during startup
- [x] `subscribeTo` snapshot handler receives the actor's **initial** snapshot during startup
- [x] `subscribeTo` does **not** subscribe to a spawned actor stopped before subscribing

**Effect ordering — `initialTransition()`**
- [x] `initialTransition()` returns listener start effect **before** target actor start
- [x] `initialTransition()` returns `subscribeTo` start effect **before** target actor start

**Effect ordering — `transition()`**
- [x] Listeners from exit and entry are each reordered ahead of their own actor, even across phases
- [x] Listener is **not** reordered ahead of an actor that was stopped before the listener was added
- [x] Interleaved spawns and listeners in the same phase are each reordered per-actor
- [x] Listener on a **pre-existing** actor is a no-op (no actor start to reorder against)
- [x] All existing listener/subscription tests pass


asfktz:
Also fixed a pre-existing type gap uncovered by the new test: `enq.subscribeTo()`'s mapper callback was typed as `Snapshot<unknown>`, so accessing `snapshot.context` (or any actor-specific snapshot field) was a type error.

**Before:**
```ts
subscribeTo: {
  <TSnapshot extends Snapshot<unknown>, TOutput, TMappedEvent extends TEvent>(
    actor: AnyActor,  // actor type discarded — TSnapshot falls back to Snapshot<unknown>
    mappers: SubscribeToMappers<TSnapshot, TOutput, TMappedEvent>
           | ((snapshot: TSnapshot) => TMappedEvent)
  ): AnyActor;
};
```

**After:**
```ts
subscribeTo: <TActor extends AnyActor, TMappedEvent extends TEvent>(
  actor: TActor,  // actor type is preserved
  mappers: SubscribeToMappers<SnapshotFrom<TActor>, OutputFrom<TActor>, TMappedEvent>
         | ((snapshot: SnapshotFrom<TActor>) => TMappedEvent)
) => AnyActor;
```

Now `snapshot` in the mapper is properly typed from the actor, so `snapshot.context`, `snapshot.value`, etc. all resolve correctly.

davidkpiano:
Hmm, I wonder if a better fix would be to "defer" the starting of an actor until after all other effects execute (I think this was what we had in v5), instead of juggling the order, which can cause some consistency/unexpected issues

asfktz:
@davidkpiano hmm, good point.

I dug into that.

It seems like the actor start can be deferred until after its listeners and subscriptions are registered, while keeping the definition order untouched. That works for a normal interpreter run.

The main challenge is the pure `transition()` function.

Let's say we have:

```tsx
entry: ({ enq }) => {
  const child = enq.spawn(childLogic);

  enq.listen(child, 'someEvent', (event) => {
    /* ... */
  });

  enq.subscribeTo(child, (snapshot) => {
    /* ... */
  });
}
```

When the user runs the effects manually in a loop, how should the deferred starts be triggered?

```tsx
const [state, effects] = transition(machine, currentState, event);

for (const effect of effects) {
  effect.exec();
}
```

`actor.on()` never had this problem because it never participated in the effects array. It was always a direct, synchronous call outside the effects system.

So, to keep the `transition()` function pure, there should be some way to ensure deferred actor starts run after their listeners are registered.

I can think of three ways to approach it:

1. **Reorder into execution order**  
    
    What the PR currently does. Rearrange the effects array so listener/subscription starts come before their target's start.
    
2. **Provide an `executeEffects` helper**  
    
    Export a function like `executeEffects(effects)` that handles deferred-start semantics internally.
    
    The user passes the effects returned from `transition()` to the helper instead of looping manually:
    
    ```tsx
    const [state, effects] = transition(machine, currentState, event);
    
    // instead of a manual loop, use the helper
    executeEffects(effects);
    ```
    
    The effects array stays in authored order, but users are expected to use the helper rather than a raw loop.
    
    The downside is that a plain `for` loop over effects would silently break. The ordering contract becomes implicit, hidden inside the helper.
    
3. **Separate spawn from start**  
    
    This is the most intrusive option: split `enq.spawn()` into two distinct effects:
    
    - `@xstate.spawn` — create and register the actor
    - `@xstate.start` — actually start it
    
    The spawn effects run in their authored order alongside normal effects, while the `@xstate.start` effects are appended at the end.
    
    The user writes:
    
    ```tsx
    entry: ({ enq }) => {
      const child = enq.spawn(childLogic); // spawn: create + register
      enq.listen(child, 'someEvent', mapper); // listen
      enq.subscribeTo(child, mapper); // subscribe
    }
    ```
    
    `transition()` / `initialTransition()` would return the effects with all starts appended at the end:
    
    ```tsx
    const [state, effects] = transition(machine, currentState, event);
    
    // effects:
    [
      // --- normal effects (authored order) ---
      { type: '@xstate.spawn', id: 'child' }, // create + register child
      { type: '@xstate.spawn', id: 'listener-1' }, // create listener
      { type: '@xstate.spawn', id: 'subscription-1' }, // create subscription
    
      // --- deferred starts (appended at the end) ---
      { type: '@xstate.start', id: 'listener-1' }, // start listener
      { type: '@xstate.start', id: 'subscription-1' }, // start subscription
      { type: '@xstate.start', id: 'child' }, // start child
    ];
    ```
    
    A plain loop just works — no helper needed, no special handling:
    
    ```tsx
    for (const effect of effects) {
      effect.exec();
    }
    
    // spawn child        → creates + registers, but does not start
    // spawn listener     → creates listener
    // spawn subscription → creates subscription
    // start listener     → subscribes to child via child.on(...)
    // start subscription → subscribes to child via child.subscribe(...)
    // start child        → child starts, emits; listeners are already attached ✓
    ```
    
    This is similar to option 1 in that the array is already in executable order, but the semantic split is explicit in the effect types.
    
    It is also closer to what v5 did with `actorScope.defer()`, just expressed declaratively in the effects array.
    

How would you prefer to approach it?

davidkpiano:
Let's do option 3; the rule becomes "`enq.spawn` defers a symmetric `.start()` to the end of the effects"

asfktz:
@davidkpiano I implemented option 3.

The actions now have this shape:

```ts
{
  type: '@xstate.spawn',
  actor,
  id: actor.id,
  logic,
  src,
  input,
  args: [actor]
}

{
  type: '@xstate.start',
  actor,
  id: actor.id,
  args: [actor]
}
```

I considered omitting `actor` and `args: [actor]` from `@xstate.start` (the matching `@xstate.spawn` already carries the actor), but landed on keeping both actions self-contained.

In `core/src/actions.ts`, `@xstate.spawn` is basically a no-op now:

```ts
export const builtInActions = {
  ['@xstate.spawn']: (actor) => {},
  ['@xstate.start']: (actor) => {
    actor.start();
  },
  // ...
};
```

which feels a little odd to me. As an alternative, `@xstate.spawn` could store the actor in a registry by id and `@xstate.start` would look it up and start it. That way `@xstate.start` could be reduce to:

```ts
{ type: '@xstate.start',  id: actor.id }
```

That feels more correct, but I wonder if it would add unnecessary ceremony when looping over the effects returned from the pure transition function. Curious to hear your thoughts.

The `@xstate.start` effects are derived in `deriveDeferredStarts` in `transitionActions.ts`:

```ts
export function deriveDeferredStarts(effects) {
  const attachedStarts = [];
  const childStarts = [];

  for (const effect of effects) {
    if (effect.type !== XSTATE_SPAWN) {
      continue;
    }
    const { actor, logic } = effect;
    const start = createStartEffect(actor);

    if (logic === listenerLogic || logic === subscriptionLogic) {
      attachedStarts.push(start);
    } else {
      childStarts.push(start);
    }
  }

  return [...attachedStarts, ...childStarts];
}
```

and appended in `StateMachine.ts`:

```ts
private _collectEffects(microsteps, initialActions) {
  const effects = [
    ...initialActions,
    ...microsteps.flatMap(([, actions]) => actions)
  ];
  return [...effects, ...deriveDeferredStarts(effects)];
}

public transition(snapshot, event, actorScope) {
  // ...
  return [nextSnapshot, this._collectEffects(microsteps)];
}

public initialTransition(input, actorScope) {
  // ....
  return [macroState, this._collectEffects(microsteps, initialActions)];
}
```

Is this the right location for this logic, and is this how you envisioned it?

Other notable changes:

- Added a guard so `Actor.start()` no longer restarts a stopped actor — this makes the deferred start a safe no-op for actors spawned and stopped in the same macrostep. Was no-restart already the intended v6 behavior?
- The same derivation is applied to `createFSM`'s boundaries — this also fixed a latent bug where FSM `enq.spawn` children were registered but never started.

Let me know if you would like any changes.

PR #5621, opened 2026-08-07
fix(core): fix assertEvent narrowing for generic event type parameters

### LINKED ISSUE #5448: Bug: [5.25.0] typing issue when using assertEvent and generics
### XState version

XState version 5

### Description

```ts
function test(event: InspectionEvent) {
	assertEvent(event, '@xstate.event');
	// eslint-disable-next-line no-console
	console.log(event.event);
}

function test2<TEVENT extends InspectionEvent>(event: TEVENT) {
	assertEvent(event, '@xstate.event');
	// eslint-disable-next-line no-console
	console.log(event.event);
}
```

test2 results in a type error:
Image

### Expected result

no error, asserts works as expected

### Actual result

error, cannot use `assertEvent` with generics.

### Reproduction

https://codesandbox.io/p/devbox/nifty-cloud-wnzqnr

### Additional context

I was able to avoid the typing issue by doing this, but I'm not sure if there's a better option:
```ts
function test2<TEVENT extends InspectionEvent>(eventP: TEVENT) {
	const event: InspectionEvent = eventP;
	assertEvent(event, '@xstate.event');
	// eslint-disable-next-line no-console
	console.log(event.event);
}
```

tarann26:
Fixes #5448

`assertEvent(event, descriptor)` compiled fine when `event`'s type was behind a generic type parameter (e.g. inside a function declared as `<TEvent extends SomeEventUnion>(event: TEvent)`), but every property access on `event` afterward failed to type-check with a "Property '...' does not exist" error. Concrete (non-generic) callers were unaffected.

Root cause: the narrowing type nests a `TEvent['type']` indexed access inside a distributive conditional over `TEvent`. TypeScript can't eagerly resolve that nesting when `TEvent` is an unresolved generic parameter, so it leaves the whole narrowed type opaque and property accesses fail.

`assertEvent` now uses its own narrowing type that keeps the wildcard normalization and the `TEvent` distribution as sibling branches instead of nesting them, which TypeScript resolves independently even for generic callers. The `TAssertedDescriptor extends EventDescriptor<TEvent>` constraint is untouched, so invalid descriptors are still a compile-time error for both concrete and generic event types.

Added type-level regression tests in `assert.types.test.ts` covering the generic narrowing case from the issue, that narrowing still excludes properties from other union members (not just "does it compile"), and that invalid descriptors are still rejected for generic callers.

---

<a href="https://app.devin.ai/review/statelyai/xstate/pull/5621" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/fix-assert-event-generic-narrowing.md
+++ b/.changeset/fix-assert-event-generic-narrowing.md
@@ -1,0 +1,7 @@
+---
+'xstate': patch
+---
+
+fix(core): fix `assertEvent` narrowing when the event type is behind a generic type parameter
+
+Previously, calling `assertEvent` inside a function whose event parameter was typed with a generic (e.g. `function f<TEvent extends SomeEventUnion>(event: TEvent)`) would compile, but every property access on `event` afterward would fail to type-check with a "Property does not exist" error. This is now fixed - `assertEvent` narrows generic event parameters the same way it narrows concrete ones, while still rejecting invalid event descriptors at compile time.
--- a/.changeset/fix-assert-event-generic-narrowing.md
+++ b/.changeset/fix-assert-event-generic-narrowing.md
@@ -5,3 +5,5 @@
 fix(core): fix `assertEvent` narrowing when the event type is behind a generic type parameter
 
 Previously, calling `assertEvent` inside a function whose event parameter was typed with a generic (e.g. `function f<TEvent extends SomeEventUnion>(event: TEvent)`) would compile, but every property access on `event` afterward would fail to type-check with a "Property does not exist" error. This is now fixed - `assertEvent` narrows generic event parameters the same way it narrows concrete ones, while still rejecting invalid event descriptors at compile time.
+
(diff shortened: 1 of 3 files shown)


tarann26:
Pushed a follow-up (e3613fb) for the union-typed `type` field regression flagged in review. Events like `{ type: 'a' | 'b'; x }` were being dropped because `AssertedEvent` used plain assignability, so the member collapsed to `never`.

I reworked it to match on `{ type: infer TType }` and test `true extends (TType extends Descriptor ? true : false)`, which mirrors `ExtractEvent`'s `EventDescriptorMatches`. A union-typed `type` field now narrows for both concrete and generic callers, while the original generic (#5448) fix and the invalid-descriptor compile errors stay intact. `types.ts` is untouched.

Added regression tests for the concrete and generic union-`type` cases, plus precision and invalid-descriptor guards.

PR #5714, opened 2026-09-15
Fix declaration emit for narrowed state contexts

davidkpiano:
## Problem

Since alpha.54, exporting a machine with a state-level context schema and a function-form transition fails declaration generation with TS4023: `rootContext` cannot be named.

## Change

Represent the root-context marker as an exported, named interface. TypeScript can reference the interface instead of expanding its private unique-symbol key into the exported config type. The symbol and context inference semantics stay intact.

Adds a declaration-emission regression fixture, TypeScript guidance, and a patch changeset.

## Validation

- Reproduced TS4023 before the fix; declaration-emission regression passes after it.
- `pnpm test:core`: 2,500 passed, 33 skipped, 3 todo.
- `pnpm typecheck`, `pnpm lint`, `pnpm format:check`: passed.
- Re-ran the declaration test after correcting its CommonJS-compatible fixture path.

The separately reported structural-contract `invoke.onDone` inference issue is not addressed here.


docs/typescript.md
```diff
@@ -36,6 +36,11 @@ Existing setups that declare only schemas keep their permissive machine-config
 typing. A state-level `schemas.context` is intersected with the root context
 schema, so it can declare only the fields that state narrows.
 
+Machines created with `setup(...).createMachine(...)` can be exported with
+TypeScript's `declaration` option enabled, including machines with state-level
+context schemas and function-form transitions. No explicit machine type
+annotation is required.
+
```
davidkpiano: Not needed
```suggestion
```

Issue #5725, opened 2026-09-19
@xstate/effect: alpha.2 publish failed (E404), join on a machine types its error as unknown, typed oxlint plugin port


joelhooks:
> **Agent-generated report.** This issue was written by an AI coding agent (Claude, run by @joelhooks) while integrating `@xstate/effect` into a strict TypeScript project. Every claim below was checked against the `next` branch at 0748e1b, the Actions log linked, and a local build of the package. Joel reviewed it before filing; reply here and he will follow up.

## Summary

Three findings from using `@xstate/effect` (XState 6.0.0-alpha.58, Effect 4.0.0-rc.115, Node 24, type-aware oxlint):

1. `@xstate/effect@0.1.0-alpha.2` never reached npm. The Publish run for the alpha.58 release 404'd on it.
2. The docs say `join` on a machine fails with "the machine's typed error". A machine's `ErrorFrom` is `unknown`, so `join(machineActor)` puts `unknown` in the error channel.
3. `scripts/oxlint-plugin-xstate-effect.mjs` cannot be lifted into a project that lints with `typescript/no-unsafe-*` and `strict-boolean-expressions`, because its AST walk is untyped. A TypeScript port on `@oxlint/plugins` passes those rules and your own `lint.test.ts`.

A patch covering 2 and 3 is attached as a gist; no PR was opened on purpose so you can decide how much of it you want.

## 1. Publish failure

The `Version Packages (alpha)` merge on `next` (0748e1b, 2026-09-17) published `xstate@6.0.0-alpha.58` but failed on the new package in the same run:

```
Some packages failed to publish:
@xstate/effect@0.1.0-alpha.2
└ E404: 404 Not Found - PUT https://registry.npmjs.org/@xstate%2feffect - Not found
Publish command exited with code 1, but some packages were published: xstate@6.0.0-alpha.58
```

Run: https://github.com/statelyai/xstate/actions/runs/35263063483

`npm view @xstate/effect` returns 404 today, so the README install line (`npm install @xstate/effect@alpha`) does not work. A 404 on the first PUT of a scoped package usually means the name is not yet provisioned for the publishing token or trusted publisher under the `@xstate` scope, not a problem with the package. Building `packages/xstate-effect` from the tag with `preconstruct build && pnpm pack` produces a working tarball with the expected peers (`effect ^4.0.0-rc.115`, `xstate ^6.0.0-alpha.58`), which is how it is being consumed in the meantime.

## 2. `join` and machine errors

[...]

PR #5726, opened 2026-09-19
fix(effect): clarify machine join errors and type the lint plugin

davidkpiano:
Addresses the documentation and lint-plugin findings in #5725. `join(machineActor)` correctly exposes `unknown` errors, but the docs promised a typed machine failure. Clarify that contract, correct the stopped-versus-errored distinction, and document Effect 4's `Effect.catch` recovery API.

Port the oxlint plugin to the typed `@oxlint/plugins` API, preserving rule behavior. Add runtime/type coverage for machine errors and a regression check using the reported strict lint rules. Include copying guidance and a changeset.

Adapted from the [reporter's patch](https://gist.github.com/joelhooks/99d193172a03609e67b4ef5849e8bbb1), with a regenerated root-workspace lockfile entry and corrected Effect 4 documentation.

Validation: 109 Effect tests pass; repository typecheck, lint, format check, and strict standalone plugin typecheck pass.

Publication remains unresolved: the linked Actions run failed with E404, and npm still returns 404 for `@xstate/effect`. The workflow uses trusted publishing; the exact account-side cause is unconfirmed. This PR does not close that part of #5725.


davidkpiano:
cc. @joelhooks

